PR #2079, opened 2021-11-23
small bug fix in Distributed Poisson Solver

### LINKED ISSUE #2074: Distributed Poisson Solver
Seems like the Distributed Poisson solver would work only when `grid.Nx == grid.Ny` and `ranks[1] == 1`. Should we maybe implement a transpose to allow different grid sizes in Nx and Ny and 2D parallelization in case of vertically stretched domains?

On the other hand, this might hinder the performance quite a bit...

simone-silvestri:
eigenvaluies correctly distributes

partially resolves #2074 (`Nx` can be different than `Ny`)


francispoulin (approved):
Makes sense.

simone-silvestri:
Finally found the correct settings after discussing with @ali-ramadhan 
Added test for `Nx != Ny` I'll merge when tests pass

glwagner (approved):

PR #2081, opened 2021-11-24
Better `show(io, ::AbstractGrid)`

### LINKED ISSUE #2064: Using `Base.summary` instead of `short_show` and better show for grids and fields
Right now at the REPL we get

```julia
julia> using Oceananigans

julia> grid = RectilinearGrid(size=3, z=(0, 1), topology=(Flat, Flat, Bounded))
RectilinearGrid{Float64, Flat, Flat, Bounded} on the CPU()
                   domain: x ∈ [1.0, 1.0], y ∈ [1.0, 1.0], z ∈ [0.0, 1.0]
                 topology: (Flat, Flat, Bounded)
        size (Nx, Ny, Nz): (1, 1, 3)
        halo (Hx, Hy, Hz): (0, 0, 1)
grid in x: Flattened
grid in y: Flattened
grid in z: Regular, with spacing 0.3333333333333333
```

so `grid in x:` isn't aligned with the rest of the "data table".

Here's what we get when `z` is stretched:

```julia
julia> grid = RectilinearGrid(size=3, z=[0.0, 0.2, 0.5, 1.0], topology=(Flat, Flat, Bounded))
RectilinearGrid{Float64, Flat, Flat, Bounded} on the CPU()
                   domain: x ∈ [1.0, 1.0], y ∈ [1.0, 1.0], z ∈ [0.0, 1.0]
                 topology: (Flat, Flat, Bounded)
        size (Nx, Ny, Nz): (1, 1, 3)
        halo (Hx, Hy, Hz): (0, 0, 1)
grid in x: Flattened
grid in y: Flattened
grid in z: Stretched, with spacing min=0.2, max=0.5
```

So here's a few ideas:

#### Include "spacing" information under the `domain` row?

One idea would be to show something like

```julia
x: Flat
y: Flat
z: [0.0, 1.0], Δz: 0.33
```

instead of `domain:` (for the regularly-spaced case), and

```julia
x: Flat
y: Flat
z: [0.0, 1.0], extrema(Δz): (0.2, 0.5)
```

for the stretched case. We could also increase the length of the table and put the spacings on their own lines, eg

```julia
x: Flat
y: Flat
z: [0.0, 1.0]
extrema(Δz): (0.2, 0.5)
```

We might also just hide flattened directions (we already express that the directions are `Flat` so why print this again?)

#### Compactify how we show the size + halo. I've always found the way we write `(Nx, Ny, Nz)` to negatively impact readability. 

I think it's ok to just write `size`, because we want to move towards providing a function-based API rather than expecting users to access struct properties directly. Eg we could have just

```julia
size: (1, 1, 1)
halo: (0, 0, 1)
```

What do others think?

navidcy:
This PR fixes the alignment issue in `show(io, ::AbstractGrid)`.

Partially addresses #2064.

### DIFF
--- a/docs/src/model_setup/grids.md
+++ b/docs/src/model_setup/grids.md
@@ -27,9 +27,9 @@ RectilinearGrid{Float64, Periodic, Periodic, Bounded} on the CPU()
                  topology: (Periodic, Periodic, Bounded)
         size (Nx, Ny, Nz): (32, 64, 256)
         halo (Hx, Hy, Hz): (1, 1, 1)
-grid in x: Regular, with spacing 4.0
-grid in y: Regular, with spacing 4.0
-grid in z: Regular, with spacing 2.0
+                grid in x: Regular, with spacing 4.0
+                grid in y: Regular, with spacing 4.0
+                grid in z: Regular, with spacing 2.0
 ```
 
 !!! info "Default domain"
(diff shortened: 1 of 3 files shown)


navidcy:
Feel free to modify further as per discussion in #2064.

tomchor (approved):
lgtm

tomchor:
I don't know if this should close https://github.com/CliMA/Oceananigans.jl/issues/2064 though, since other chances are being discussed there.

navidcy:
I'll merge to fix the alignment and we open a new one for the rest of what is discussed in #2064.

PR #2082, opened 2021-11-24
Enhance docstrings + add models & modules in Docs/Appendix/Library

### LINKED ISSUE #2080: Use Doctest filtering in doctests
We'd like to convert this example:

https://github.com/CliMA/Oceananigans.jl/blob/1f7792ac8be223a0b32bf38d0d8954ea57e9d294/src/Advection/weno_fifth_order.jl#L79-L89

into a doctest. But to do so, we need to [filter out] from the output the two lines related to the `@warn ...`.

I tried to do that, but I had trouble properly creating the regular expression to capture both these lines.

E.g.,

<pre>
<code>
```@meta
DocTestFilters = [Regex("Warning: defaulting"), Regex("Oceananigans.Advection")"]
```

```jldoctest
julia> using Oceananigans

julia> WENO5()
┌ Warning: defaulting to uniform WENO scheme with Float64 precision, use WENO5(grid = grid) if this was not intended
└ @ Oceananigans.Advection ~/Research/OC2/src/Advection/weno_fifth_order.jl:90
WENO5 advection scheme with:
    ├── X regular
    ├── Y regular
    └── Z regular
```
</code>
</pre>

didn't work because, apparently, I was only capturing the "Warning: defaulting" and "Oceananigans.Advection" from each line. What I'd like is something like `* "Warning: defaulting" *` but I don't know how to write a `Regex` to do so. Anyone could help?

cc @tomchor

navidcy:
This PR simplifies the Docs/Appandix/Library and also adds missing modules in it. 

Furthermore, it enhances formatting in several docstrings and converts some examples to doctests.

Closes  #2080.

For example, regarding 2 vs 4 spaces in markdown formatting,

**before this PR**

Screen Shot 2021-11-25 at 4 03 01 pm

while **after this PR**

Screen Shot 2021-11-25 at 4 02 41 pm

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
@@ -19,26 +19,27 @@ struct ImplicitFreeSurface{E, G, B, I, M, S}
 end
 
 """
-    ImplicitFreeSurface(; solver=:PreconditionedConjugateGradient, gravitational_acceleration=g_Earth, solver_settings...) =
+    ImplicitFreeSurface(; solver_method=:Default, gravitational_acceleration=g_Earth, solver_settings...)
 
 The implicit free surface equation is
 
 ```math
-(∇ʰ ⋅ H ∇ʰ - 1 / (g Δt²)) ηⁿ⁺¹ = ∇ʰ ⋅ Q★ / (g Δt) - ηⁿ / (g Δt²)
+(∇ʰ ⋅ H ∇ʰ - 1 / (g Δt²)) ηⁿ⁺¹ = ∇ʰ ⋅ Q★ / (g Δt) - ηⁿ / (g Δt²) ,
 ```
 
 where ``H`` is depth, ``g`` is gravitational acceleration, ``Δt`` is time step, and
 ``Q★`` is the barotropic volume flux associated with the predictor velocity field.
 
-This can be solved in general using the `PreconditionedConjugateGradientSolver`.
+This equation be solved in general using the `PreconditionedConjugateGradientSolver`.
 
 In the case that ``H`` is constant, we divide through to obtain
 
 ```math
-(∇² - 1 / (g H Δt²)) ηⁿ⁺¹ = 1 / (g H Δt) * (∇ʰ ⋅ Q★ - ηⁿ / Δt)
+(∇² - 1 / (g H Δt²)) ηⁿ⁺¹ = 1 / (g H Δt) * (∇ʰ ⋅ Q★ - ηⁿ / Δt) ,
 ```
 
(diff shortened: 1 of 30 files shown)


src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
```diff
@@ -19,26 +19,27 @@ struct ImplicitFreeSurface{E, G, B, I, M, S}
...
 ```math
-(∇ʰ ⋅ H ∇ʰ - 1 / (g Δt²)) ηⁿ⁺¹ = ∇ʰ ⋅ Q★ / (g Δt) - ηⁿ / (g Δt²)
+(∇ʰ ⋅ H ∇ʰ - 1 / (g Δt²)) ηⁿ⁺¹ = ∇ʰ ⋅ Q★ / (g Δt) - ηⁿ / (g Δt²) ,
 ```
 
 where ``H`` is depth, ``g`` is gravitational acceleration, ``Δt`` is time step, and
 ``Q★`` is the barotropic volume flux associated with the predictor velocity field.
 
-This can be solved in general using the `PreconditionedConjugateGradientSolver`.
+This equation be solved in general using the `PreconditionedConjugateGradientSolver`.
```
tomchor, line 33:
```suggestion
This equation can be solved in general using the `PreconditionedConjugateGradientSolver`.
```

src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
```diff
@@ -19,26 +19,27 @@ struct ImplicitFreeSurface{E, G, B, I, M, S}
 end
 
 """
-    ImplicitFreeSurface(; solver=:PreconditionedConjugateGradient, gravitational_acceleration=g_Earth, solver_settings...) =
+    ImplicitFreeSurface(; solver_method=:Default, gravitational_acceleration=g_Earth, solver_settings...)
```
tomchor, line 22: I wonder if there's a way to get the function signature automatically like python does, so these little typos (that you just fixed) don't happen.

tomchor (approved):

src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
navidcy: Nice point. Perhaps.

navidcy: https://juliadocs.github.io/DocStringExtensions.jl/latest/#DocStringExtensions.SIGNATURES ?

tomchor: I like this. Seems like it would at least help and bring some consistency.

PR #2083, opened 2021-11-28
Fix WENO5 doctest

navidcy:
Seems that "\n" cannot be part of regular expressions for doctest filters. Otherwise, something goes wrong with the parsing and

```
DocTestFilters = [Regex(".*┌ Warning.*
"), Regex(".*└ @ Oceananigans.*
")]
```
appears in the docstring.

Compare 

https://clima.github.io/OceananigansDocumentation/dev/appendix/library/#Oceananigans.Advection.WENO5

to

https://clima.github.io/OceananigansDocumentation/previews/PR2083/appendix/library/#Oceananigans.Advection.WENO5

### DIFF
--- a/src/Advection/weno_fifth_order.jl
+++ b/src/Advection/weno_fifth_order.jl
@@ -63,8 +63,8 @@ Keyword arguments
 =================
 
   - `grid`: (defaults to `nothing`)
-  - `stretched_smoothness`: When `true` it results in the coefficients for the smoothness indicators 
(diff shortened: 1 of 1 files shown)


tomchor:
Just so I understand this filtering thing. It doesn't remove the regex-filtered lines from the output; it just doesn't test those lines against the "prediction", right?

glwagner (approved):

navidcy:
> Just so I understand this filtering thing. It doesn't remove the regex-filtered lines from the output; it just doesn't test those lines against the "prediction", right?

Yeap!

PR #2087, opened 2021-11-30
Allow NaNChecker.erroring

### LINKED ISSUE #2086: Option for `NaNChecker` to exit with error
Before the recent overhaul to callbacks, `NaNChecker` exited with an error. While I agree that exiting without an error is convenient at times, it would be good to have the built-in option to keep the original behavior.

For example, consider a code that runs a simulation and then writes a checkpointer after it's done running. The legacy behavior was that if the simulation blew up, `run!` threw an error and the rest of the script (including the checkpointing) wouldn't happen. That would allow the user to restart the simulation from the latest checkpointer (possibly with a lower cfl threshold). Now if the simulation blows up, a checkpointer (full of NaNs) is still written, possibly overwriting the previous checkpointer.

I know we can cook up a callback that manually throws an error, but this seems like a common enough case that there should be a built-in option for it. (Or maybe it's just me?)

glwagner:
This PR adds a field `erroring` to the `NaNChecker`, and also makes the `NaNChecker` mutable. This means that the default `NaNChecker` can be converted to an erroring `NaNChecker` by writing

```julia
simulation.callbacks[:nan_checker].func.erroring = true
```

or

```julia
erroring_NaNChecker!(simulation)
```

TODO

- [x] Test

Resolves #2086 .


tomchor:
I checked and this seems to work:

```julia
julia> using Oceananigans

julia> model = NonhydrostaticModel(grid=grid)
NonhydrostaticModel{CPU, Float64}(time = 0 seconds, iteration = 0) 
├── grid: RectilinearGrid{Float64, Periodic, Periodic, Bounded}(Nx=4, Ny=4, Nz=4)
├── tracers: ()
├── closure: Nothing
├── buoyancy: Nothing
└── coriolis: Nothing

julia> noise(x, y, z) = 100+randn()
noise (generic function with 1 method)

julia> set!(model, u=noise)

julia> simulation = Simulation(model, Δt=100, stop_iteration=200)
Simulation{typename(NonhydrostaticModel){typename(CPU), Float64}}
├── Model clock: time = 0 seconds, iteration = 0
├── Next time step: 1.667 minutes
├── Stop time: Inf years
├── Stop iteration : 200.0
├── Wall time limit: Inf
├── Callbacks: typename(OrderedCollections.OrderedDict) with 4 entries:
│   ├── stop_time_exceeded => typename(Callback)
│   ├── stop_iteration_exceeded => typename(Callback)
│   ├── wall_time_limit_exceeded => typename(Callback)
│   └── nan_checker => typename(Callback)
├── Output writers: typename(OrderedCollections.OrderedDict) with no entries
└── Diagnostics: typename(OrderedCollections.OrderedDict) with no entries

julia> simulation.callbacks[:nan_checker].func.erroring = true
true

julia> run!(simulation)
[ Info:     ... simulation initialization complete (454.946 μs)
[ Info:     ... initial time step complete (3.370 ms).
ERROR: time = 10000.0, iteration = 100: NaN found in field u. Aborting simulation.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] (::Oceananigans.Simulations.NaNChecker{NamedTuple{(:u,), Tuple{Field{Face, Center, Center, CPU, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, RectilinearGrid{Float64, Period....
....
```

tomchor (approved):
Thanks for this PR!

glwagner:
Great! Now we just need a test in CI to make sure it doesn't break in the future, and we can merge this PR.

navidcy:
@glwagner, merge?

PR #2088, opened 2021-11-30
AllSchedule for combining scheduling criteria and avoiding checkpointing with NaNs

glwagner:
This PR adds a new schedule called `AllSchedule` that allows scheduling based on multiple criterion.

It also adds a `hasnan(model::AbstractModel)` that checks for a NaN in a model's first field.

Together this allows schedules that avoid actuating if NaNs are detected via

```julia
schedule = AllSchedule(TimeInterval(100), !hasnan)
```

for example. Note that `all` is short-circuiting, so order matters. Putting `!hasnan` second means that `hasnan` will not be called unless `TimeInterval(100)(model)` returns `true` first (this matters for avoiding `hasnan` calls on every time-step).

TODO:

- [x] better name than `MultiSchedule` perhaps `AllSchedules`?
- [x] Also have `AnySchedule` that actuates if _any_ schedules actuate?
- [x] Tests

This feature was discussed on #2086 .


navidcy:
`AnySchedule` -> `SensitiveSchedule`?

navidcy:
Could you resolve the conflicts @glwagner?

glwagner:
> `AnySchedule` -> `SensitiveSchedule`?

`EitherSchedule`? (but "either" implies there are just two whereas here there can be many).

navidcy:
> > `AnySchedule` -> `SensitiveSchedule`?
> 
> `EitherSchedule`? (but "either" implies there are just two whereas here there can be many).

`AnySchedule` is good! It just took me a while to get used to :)

navidcy:
> > > `AnySchedule` -> `SensitiveSchedule`?
> > 
> > 
> > `EitherSchedule`? (but "either" implies there are just two whereas here there can be many).
> 
> `AnySchedule` is good! It just took me a while to get used to :)

Also I just noticed that these are documented and that's great!

glwagner:
We could also use `AndSchedule` (rather than `AllSchedule`) and `OrSchedule` (rather than `AnySchedule`). "and" / "or" refer to the underlying conditionals that are used in the reductions `all` / `any`. Is this easier to understand (I think it might be).

tomchor (approved):
Great addition

tomchor:
> We could also use `AndSchedule` (rather than `AllSchedule`) and `OrSchedule` (rather than `AnySchedule`). "and" / "or" refer to the underlying conditionals that are used in the reductions `all` / `any`. Is this easier to understand (I'm not sure)?

I think the `AndSchedule` and `OrSchedule` notation is clearer (or at least more explicit), so I'm leaning toward that one. But I don't feel strongly anyway so feel free to merge this whenever you're happy!

glwagner:
I like "and" / "or" better as well, after having slept on it.

PR #2093, opened 2021-12-01
Avoid computing hydrostatic pressure when z is Flat

### LINKED ISSUE #2092: Avoid updating hydrostatic pressure for Flat z dimensions
This is related to https://github.com/CliMA/Oceananigans.jl/blob/2e29790bbe95a9b4d431751578612289a4929e8f/src/Fields/field_tuples.jl#L94-L105 (which was left out from PR #1994).

Possibly also related to #1919?

navidcy:
Closes #2092

### DIFF
--- a/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
+++ b/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
@@ -16,6 +16,12 @@ the `buoyancy_perturbation` downwards:
     end
 end
 
+update_hydrostatic_pressure!(model) = update_hydrostatic_pressure!(model.grid, model)
+
+update_hydrostatic_pressure!(::AbstractGrid{<:Any, <:Any, <:Any, <:Flat}, model) = nothing
+
+update_hydrostatic_pressure!(grid, model) = update_hydrostatic_pressure!(model.pressures.pHY′, model.architecture, model.grid, model.buoyancy, model.tracers)
+
 function update_hydrostatic_pressure!(pHY′, arch, grid, buoyancy, tracers)
     pressure_calculation = launch!(arch, grid, :xy, _update_hydrostatic_pressure!,
                                    pHY′, grid, buoyancy, tracers,
@@ -26,6 +32,3 @@ function update_hydrostatic_pressure!(pHY′, arch, grid, buoyancy, tracers)
 
     return nothing
 end
-
-update_hydrostatic_pressure!(pHY′, arch, grid::AbstractGrid{<:Any, <:Any, <:Any, <:Flat}, args...) = nothing
-
--- a/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
+++ b/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
@@ -30,5 +30,6 @@ function update_hydrostatic_pressure!(pHY′, arch, grid, buoyancy, tracers)
     # Fill halo regions for pressure
     wait(device(arch), pressure_calculation)
 
+    fill_halo_regions!(pHY′, arch)
     return nothing
 end
--- a/src/Models/NonhydrostaticModels/velocity_and_tracer_tendencies.jl
+++ b/src/Models/NonhydrostaticModels/velocity_and_tracer_tendencies.jl
@@ -5,15 +5,27 @@ using Oceananigans.Operators
 using Oceananigans.StokesDrift
 using Oceananigans.TurbulenceClosures: ∂ⱼ_τ₁ⱼ, ∂ⱼ_τ₂ⱼ, ∂ⱼ_τ₃ⱼ, ∇_dot_qᶜ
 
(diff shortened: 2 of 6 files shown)


src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
```diff
@@ -24,8 +30,6 @@ function update_hydrostatic_pressure!(pHY′, arch, grid, buoyancy, tracers)
     # Fill halo regions for pressure
     wait(device(arch), pressure_calculation)
 
+    fill_halo_regions!(pHY′, arch)
```
navidcy, line 33: @glwagner, is this OK here? It was outside in the `update_state!` but we only want it when z is *not* flat.

navidcy:
Hm...

Now we get into trouble with `calculate_tendencies!(model::NonhydrostaticModel)` which passes on `model.pressures.pHY′`...

navidcy:
I'm leaning towards rewriting `PressureFields` so that
```Julia
function PressureFields(arch, grid::AbstractGrid{<:Any, <:Any, <:Any, <:Flat}, bcs=NamedTuple())
    default_pressure_boundary_conditions =
        (pHY′ = FieldBoundaryConditions(grid, (Center, Center, Center)),
         pNHS = FieldBoundaryConditions(grid, (Center, Center, Center)))

    bcs = merge(default_pressure_boundary_conditions, bcs)
    pNHS = CenterField(arch, grid, bcs.pNHS)

    return (; pHY′=nothing, pNHS=pNHS)
end
```
and then adding methods, e.g.,
```Julia
∂xᶠᶜᵃ(i, j, k, grid, ::nothing) = 0
```

src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
glwagner: It's probably best here, rather than in `update_state!`.

src/Operators/derivative_operators.jl
```diff
@@ -94,3 +94,27 @@
...
+@inline ∂yᵃᶠᵃ(i, j, k, grid, ::Nothing, args...) = 0
+
+@inline ∂yᶜᶜᵃ(i, j, k, grid, ::Nothing, args...) = 0
+@inline ∂yᶠᶜᵃ(i, j, k, grid, ::Nothing, args...) = 0
+
+@inline ∂yᶠᶠᵃ(i, j, k, grid, ::Nothing, args...) = 0
+@inline ∂yᶜᶠᵃ(i, j, k, grid, ::Nothing, args...) = 0
+
+@inline ∂zᵃᵃᶜ(i, j, k, grid, ::Nothing, args...) = 0
+@inline ∂zᵃᵃᶠ(i, j, k, grid, ::Nothing, args...) = 0
```
glwagner, line 120: Another solution (slightly less code) is to define a method like `hydrostatic_pressure_gradient_x(i, j, k, grid, p)`.

Then `p::Nothing` returns 0 and otherwise returns `∂xᶠᶜᵃ(i, j, k, grid, p)`.

What do you think?

glwagner (approved):
Looks great. Just proposed an alternative to defining new derivative methods that might help make code more clear / avoid silent bugs when differentiating `nothing` (also will increase our test coverage since we won't have to add tests for all the derivative methods).

src/Operators/derivative_operators.jl
navidcy: That's  much better I think.

I was also a bit skeptical for `∂x` to return 0 when given nothing. We may not always want that.

navidcy:
Is there any other model that may be impacted from this?

src/Models/NonhydrostaticModels/velocity_and_tracer_tendencies.jl
```diff
@@ -5,15 +5,27 @@ using Oceananigans.Operators
 using Oceananigans.StokesDrift
 using Oceananigans.TurbulenceClosures: ∂ⱼ_τ₁ⱼ, ∂ⱼ_τ₂ⱼ, ∂ⱼ_τ₃ⱼ, ∇_dot_qᶜ
 
+"return the ``x``-gradient of hydrostatic pressure"
+hydrostatic_pressure_gradient_x(i, j, k, grid, hydrostatic_pressure) = ∂xᶠᶜᵃ(i, j, k, grid, hydrostatic_pressure)
+hydrostatic_pressure_gradient_x(i, j, k, grid, ::Nothing) = 0
```
glwagner, line 10:
```suggestion
hydrostatic_pressure_gradient_x(i, j, k, grid, ::Nothing) = zero(eltype(grid))
```

might be a bit safer

glwagner:
> Is there any other model that may be impacted from this?

As mentioned in a DM I don't think `HydrostaticFreeSurfaceModel` supports vertically-flat configurations. This could be worth visiting when we have a nonlinear free surface.

PR #2094, opened 2021-12-02
a little change to run checkpointers with IBG

simone-silvestri:


src/OutputWriters/checkpointer.jl
```diff
@@ -201,8 +201,13 @@ function set!(model, filepath::AbstractString)
         # Validate the grid
         checkpointed_grid = file["grid"]
 
-        model.grid == checkpointed_grid ||
-            error("The grid associated with $filepath and model.grid are not the same!")
+	if model.grid isa ImmersedBoundaryGrid
+            model.grid.grid == checkpointed_grid ||
```
glwagner, line 205: 🤦 

We should change this property name to `underlying_grid` or something (future PR...)

src/OutputWriters/checkpointer.jl
```diff
@@ -201,8 +201,13 @@ function set!(model, filepath::AbstractString)
         # Validate the grid
         checkpointed_grid = file["grid"]
 
-        model.grid == checkpointed_grid ||
-            error("The grid associated with $filepath and model.grid are not the same!")
+	if model.grid isa ImmersedBoundaryGrid
+            model.grid.grid == checkpointed_grid || 
+            error("The grid associated with $filepath and the underlying ib grid are not the same!")
```
glwagner, line 206:
```suggestion
            error("The grid associated with $filepath and the underlying `ImmersedBoundaryGrid.grid` are not the same!")
```

glwagner (approved):

src/OutputWriters/checkpointer.jl
simone-silvestri: yeah that would be usefull

PR #2097, opened 2021-12-03
Fixes for implicit vertical diffusion with immersed boundaries

### LINKED ISSUE #2049: Should vertical spacings `Δzᵃᵃᶜ` and `Δzᵃᵃᶠ` depend on horizontal location?
They don't right now, even though right now the lateral areas `Ax` and `Ay` do depend on horizontal location:

https://github.com/CliMA/Oceananigans.jl/blob/24e766481cebbc8f61099b386623d175218acedb/src/Operators/spacings_and_areas_and_volumes.jl#L151

Though our underlying grids are "extruded" in the vertical (put another way, we make the thin shell approximation for spherical shell grids), vertical spacings may in principle need to depend on horizontal location to accurately represent bathymetry.

Another way to see this is that the vertical spacings for `GridFittedBottom` (implemented in #2023) are

https://github.com/CliMA/Oceananigans.jl/blob/8eadc493e04002066448323573edcfde046b9c30/src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl#L56-L62

thus conditioning on the function `is_immersed`. But `is_immersed` is evaluated at `(Center, Center, Center)`. But more general operators such as `Δzᶠᶜᶜ` may be needed for correct bathymetry representation.

Note that we might need this note only to compute depths correctly, but also for correct diagnostics.

cc @sandreza @jm-c

### LINKED ISSUE #2065: bug in vertically implicit diffusion solver ?
It seems that the ivd_diagonal (main diagnonal) has not the right 1/ Δz factor:
idv_diagonal(k) should only be function of  Δz_c(k) and also  Δz_f(k) &  Δz_f(k+1); I think by switching k_f & k_c in the definition of "kappa_ Δz^2" it should fix this (to check).
This problem should only shows up with vertically streched grid.

For the 2 other diagonals ((upper & lower), it's not clear to me because I don't really know what this comment: "# Shift to adjust for Tridiagonal indexing convenction" means: Does this means that we don't pick one of the 2 simpler conventions for indexing the tri-diagonal coeff, i.e, either by line/row or by columns ?

### LINKED ISSUE #2116: Vertically-implicit diffusion needs to know about field location (or whether field is momentum or tracer)
Right now the functions that calculate implicit vertical diffusion coefficients don't know about a field's horizontal location:

https://github.com/CliMA/Oceananigans.jl/blob/11727d66587e01067bf1ac24e2655e593e98fe9e/src/TurbulenceClosures/vertically_implicit_diffusion_solver.jl#L41-L44

except implicitly through `κ⁻⁻ᶠ`. However, when using an immersed boundary, we need the _horizontal_ location in order to correctly mask the coefficients of the implicit vertical diffusion tridiagonal matrix.

So, rather than passing the function `κ⁻⁻ᶠ`, we should pass the field's location here. This will also mean that we only need one `ivd_upper_diagonal`, and we can dispatch on the special case that a field is located at cell interfaces in the vertical.

### LINKED ISSUE #2158: Do we need `Ax`, `Ay`, `Az` and `volume` from `Operators.jl`?
These are just functions which retrieve `Ax`, `Ay`, `Az` and `V` on the location of interest. On the other hand, they conflict with the same variables exported from `AbstractOperations`. I think they should be exported only by the latter because as `AbstractOperations` they have a unique functionality whereas from `Operators` they are just a renaming of `Axᵃᵃᵃ`, `Ayᵃᵃᵃ`, `Azᵃᵃᵃ`and `Vᵃᵃᵃ`

glwagner:
The tridiagonal coefficients in the solver for the implicit vertical diffusion correction step need to be modified when there's an immersed boundary.

WIP with @jm-c 

Closes #2065
Closes #2158 
Closes #2116
Closes #2049


src/TurbulenceClosures/vertically_implicit_diffusion_solver.jl
```diff
@@ -40,46 +40,49 @@ implicit_diffusion_solver(::ExplicitTimeDiscretization, args...; kwargs...) = no
 
 @inline function ivd_upper_diagonalᵃᵃᶜ(i, j, k, grid, clock, Δt, κ⁻⁻ᶠ, κ)
     κᵏ⁺¹ = κ⁻⁻ᶠ(i, j, k+1, grid, clock, κ)
-    return - Δt * κ_Δz²(i, j, k, k+1, grid, κᵏ⁺¹)
+
+    return ifelse(k < 1,
+                  zero(eltype(grid)),
+                  - Δt * κ_Δz²(i, j, k, k+1, grid, κᵏ⁺¹))
```
jm-c, line 46: I think the special cases (k=1 & k=Nz) should be switched, i.e., we should take care of k = 0 in lower-diag function and take care of k=Nz in upper-diag function.

glwagner: since there's no test I think we are free to switch. @jm-c do you want me to do it or do you want to do it?

jm-c: I am currently checking other index related things here and will try to run it after that. From there I could push changes to his PR.

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
```diff
@@ -73,3 +79,31 @@ const GFBIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:GridFit
 
 Adapt.adapt_structure(to, ib::GridFittedBottom) = GridFittedBottom(adapt(to, ib.bottom))     
 
+#####
+##### Implicit vertical diffusion
+#####
+
+# Tracers and horizontal velocities at cell centers in z
+
+@inline function ivd_upper_diagonalᵃᵃᶜ(i, j, k, ibg::GFIBG, clock, Δt, κ⁻⁻ᶠ, κ)
```
jm-c, line 88: there is a problem here since GFIBG is not defined. I tried to use GFBIBG instead but still getting an error that seems to be related to changes in this file.

glwagner: Sorry about that, I'll fix it.

jm-c:
@glwagner Thanks for fixing this, I can now run it.

src/Grids/grid_generation.jl
```diff
@@ -59,16 +59,9 @@ function generate_coordinate(FT, topology, N, H, coord, architecture)
...
-    # Seems needed to avoid out-of-bounds error in viscous dissipation
-    # operators wanting to access ΔF[N+2].
-    # ΔF = OffsetArray(ΔF, -H)
-    # ΔF = OffsetArray(cat(ΔF[0], ΔF..., ΔF[N], dims=1), -H-1)
-    # ΔF = OffsetArray(arch_array(architecture, ΔF.parent), ΔF.offsets...)
 
-    ΔC = OffsetArray(ΔC, -H)
-    ΔC = OffsetArray(arch_array(architecture, ΔC.parent), ΔC.offsets...)
+    ΔC = OffsetArray(arch_array(architecture, ΔC), -H)
+    ΔF = OffsetArray(arch_array(architecture, ΔF), -H-1)
```
glwagner, line 64: You're going to have merge conflicts with #2150 ... you probably want to merge in those changes rather than making them both places !

src/TurbulenceClosures/vertically_implicit_diffusion_solver.jl
```diff
@@ -34,7 +34,7 @@ implicit_diffusion_solver(::ExplicitTimeDiscretization, args...; kwargs...) = no
 ##### Note: "ivd" stands for implicit vertical diffusion.
 #####
 
-@inline κ_Δz²(i, j, kᶠ, kᶜ, grid, κ) = κ / Δzᵃᵃᶜ(i, j, kᶜ, grid) / Δzᵃᵃᶠ(i, j, kᶠ, grid)
+@inline κ_Δz²(i, j, kᶜ, kᶠ, grid, κ) = κ / Δzᵃᵃᶜ(i, j, kᶜ, grid) / Δzᵃᵃᶠ(i, j, kᶠ, grid)
```
jm-c, line 37: This is addressing issue #2065 which can be close once this PR is merged in.

glwagner: @jm-c I'll add a statement to the PR description so this happens automatically.

simone-silvestri:
Should be ready to go when all tests pass. We can wait for PR #2121 to finish and then merge this in

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
```diff
@@ -104,13 +104,13 @@ end
...
                   zero(eltype(ibg.grid)),
-                  ivd_upper_diagonalᵃᵃᶜ(i, j, k, ibg.grid, clock, Δt, νᶜᶜᶜ, κ))
+                  ivd_upper_diagonalᵃᵃᶠ(i, j, k, ibg.grid, clock, Δt, νᶜᶜᶜ, κ))
 end
 
 @inline function ivd_lower_diagonalᵃᵃᶠ(i, j, k, ibg::GFIBG, clock, Δt, νᶜᶜᶜ, κ)
     return ifelse(solid_node(Center(), Center(), Center(), i, j, k+1, ibg),
                   zero(eltype(ibg.grid)),
-                  ivd_lower_diagonalᵃᵃᶜ(i, j, k, ibg.grid, clock, Δt, νᶜᶜᶜ, κ))
+                  ivd_lower_diagonalᵃᵃᶠ(i, j, k, ibg.grid, clock, Δt, νᶜᶜᶜ, κ))
```
glwagner, line 113: Looks like this function and the one above are identical?

simone-silvestri: the one above is the upper diagonal and this one is the lower? Or you mean the change?

glwagner: Ahh, I was focused on the three letter code and missed that one was `upper` and one was `lower`.

simone-silvestri:
Update: we have to regenerate the Hydrostatic regression data because of a bug where
`@inline znode(::Center, k, grid::LatitudeLongitudeGrid) = @inbounds grid.zᵃᵃᶠ[k]` instead of
`@inline znode(::Center, k, grid::LatitudeLongitudeGrid) = @inbounds grid.zᵃᵃᶜ[k]`

for the rest, it should be ready to go. I am trying now a 1/4 degree run and it seems to keep.

navidcy:
bump patch release?

simone-silvestri:
Maybe! I actually still have a couple test to fix before it's ready 😅

navidcy (approved):

navidcy:
bump patch release?

src/ImmersedBoundaries/masked_reductions.jl
```diff
@@ -0,0 +1,79 @@
...
+import Statistics: norm
+
+# ###
+# ###  reduction operations involving immersed boundary grids exclude the immersed region 
+# ### (we exclude also the values on the faces of the immersed boundary with `solid_interface`)
+# ###
+
+const ImmersedField = AbstractField{<:Any, <:Any, <:Any, <:ImmersedBoundaryGrid}
+
+struct MaskedObject{LX, LY, LZ, O, G, M, T} <: AbstractArray{T, 3}
```
glwagner, line 12:
```suggestion
struct MaskedOperation{LX, LY, LZ, O, G, M, T} <: AbstractOperation{LX, LY, LZ, G, T} 
```

note that `AbstractOperation` subtypes `AbstractArray` (via `AbstractField`).

src/ImmersedBoundaries/masked_reductions.jl
```diff
@@ -0,0 +1,79 @@
...
+
+# ###
+# ###  reduction operations involving immersed boundary grids exclude the immersed region 
+# ### (we exclude also the values on the faces of the immersed boundary with `solid_interface`)
+# ###
+
+const ImmersedField = AbstractField{<:Any, <:Any, <:Any, <:ImmersedBoundaryGrid}
+
+struct MaskedObject{LX, LY, LZ, O, G, M, T} <: AbstractArray{T, 3} 
+   obj  :: O
```
glwagner, line 13: `operand` or `parent`?

src/ImmersedBoundaries/masked_reductions.jl
```diff
@@ -0,0 +1,79 @@
...
+@inline masked_object(LX, LY, LZ, obj, grid, mask) = MaskedObject{LX, LY, LZ}(obj, grid, mask)
+@inline masked_object(obj::ImmersedField, mask)    = masked_object(location(obj)..., obj, obj.grid, mask)
+
+# One-directional masking
+
+
+Adapt.adapt_structure(to, mo::MaskedObject{LX, LY, LZ}) where {LX, LY, LZ} =
+            MaskedObject{LX, LY, LZ}(adapt(to, mo.obj), adapt(to, mo.grid), adapt(to, mo.mask))
+
+@inline Base.size(mo::MaskedObject) = Base.size(mo.obj)
```
glwagner, line 32: If we subtype `AbstractOperation` we don't need to define this.

src/ImmersedBoundaries/masked_reductions.jl
```diff
@@ -0,0 +1,79 @@
...
+
+function Statistics._mean(f, c::ImmersedField; dims)
+    r = sum(f, c; dims = dims)
+    n = sum(f, masked_object(c / c, 0); dims = dims)
+
+    @show r, n
+    return r ./ n
+end
+
+Statistics._mean(c::ImmersedField; dims) = Statistics._mean(identity, c, dims=dims)
```
glwagner, line 79: None of the above should be necessary!

glwagner, line 79: Any way to avoid this big copy/paste?

glwagner: Yes, I think there should be a way to avoid this. I think a slight modification to the definition of reductions in `field.jl` will permit introducing masks with just a few extra lines.

simone-silvestri: The new commit changes fields. I have to change some stuff still though

src/Fields/field.jl
```diff
@@ -350,14 +350,16 @@ function reduced_location(loc; dims)
     end
 end
 
-## Allow support for ImmersedFields
-@inline masked_object(a, mask) = a
+## Allow support for MaskedOperations
```
glwagner, line 353:
```suggestion
## Interface to support specialization for masked reductions
```

src/Fields/field.jl
```diff
@@ -350,14 +350,16 @@ function reduced_location(loc; dims)
...
+
 get_neutral_mask(::Union{AllReduction, AnyReduction}) = true
 get_neutral_mask(::MinimumReduction) =   Inf
 get_neutral_mask(::MaximumReduction) = - Inf
 get_neutral_mask(::ProdReduction)    =   1
 get_neutral_mask(::SumReduction)     =   0
 
+@inline mask_operation(operand, mask) = operand
```
glwagner, line 361:
```suggestion
@inline mask_operand(operand, mask) = operand
```

src/ImmersedBoundaries/immersed_fields_reductions.jl
```diff
@@ -0,0 +1,43 @@
...
+const ImmersedField = AbstractField{<:Any, <:Any, <:Any, <:ImmersedBoundaryGrid}
+
+@inline immersed_condition(i, j, k, mo::MaskedObject{LX, LY, LZ}) = solid_interface(LX(), LY(), LZ(), i, j, k, mo.grid) 
+
+@inline function masked_operation(obj::ImmersedField, mask) 
+    return masked_object(location(obj)..., obj, obj.grid, mask, immersed_condition)
+end
+
+@inline masked_length(c::AbstractField) = length(c)
+@inline masked_length(c::ImmersedField) = sum(masked_object(c / c, 0))
```
glwagner, line 22: Are we worried about redefining `length`?

src/AbstractOperations/masked_operations.jl
```diff
@@ -0,0 +1,39 @@
...
+
+@inline function Base.getindex(mo::MaskedOperation{LX, LY, LZ}, i, j, k) where {LX, LY, LZ}
+    return ifelse(condition(i, j, k, mo), 
+                  get_mask(mo.mask, i, j, k), 
+                  getindex(mo.operand, i, j, k))
+end
+
+@inline get_mask(mask::Number       , i, j, k) = mask
+@inline get_mask(mask::AbstractArray, i, j, k) = mask[i, j, k]
+@inline get_mask(mask::Base.Callable, i, j, k) = mask(i, j, k)
```
glwagner, line 39: beautiful 🌅

glwagner: Maybe include `grid` among the arguments to the callable?

glwagner: That's what we would need if people want to mask as function of `x, y, z` for example

src/AbstractOperations/conditional_operations.jl
```diff
@@ -0,0 +1,36 @@
...
+     function ConditionalOperation{LX, LY, LZ}(operand::O, grid::G, condition::C, mask::M) where {LX, LY, LZ, O, G, C, M}
+         T = eltype(operand)
+         return new{LX, LY, LZ, O, G, C, M, T}(operand, grid, condition, mask)
+     end
+end
+ 
+@inline condition_operand(operand::AbstractField, condition, mask)    = condition_operand(location(operand)..., operand.grid, condition, mask)
+@inline condition_operand(LX, LY, LZ, operand, grid, condition, mask) = ConditionalOperation{LX, LY, LZ}(operand, grid, condition, mask)
+
+@inline conditional_length(c::ConditionalOperation)       = sum(c / c)
```
glwagner, line 20: Doesn't this produce `NaN` when the `mask=0`?

simone-silvestri: yeah, I am trying different solutions for the length. I think using `set!` we can set an auxiliary field to sum where `condition == true`

glwagner: we need to sum `OneField` with a zero mask (a field that is 1 everywhere)

glwagner: it's fast too cause `OneField` is trivial, so it all boils down to evaluating the conditional everywhere (that expense can't be avoided)

simone-silvestri:
To summarize the updates, we can make use of the type `ConditialOperation` to perform conditional sums with a mask. Example:

```
julia> grid = RectilinearGrid(arch, size = (4, 1, 1), extent = (1, 1, 1))
RectilinearGrid{Float64, Periodic, Periodic, Bounded}
             architecture: CPU()
                   domain: x ∈ [0.0, 1.0], y ∈ [0.0, 1.0], z ∈ [-1.0, 0.0]
                 topology: (Periodic, Periodic, Bounded)
        size (Nx, Ny, Nz): (4, 1, 1)
        halo (Hx, Hy, Hz): (1, 1, 1)
             spacing in x: Regular, with spacing 0.25
             spacing in y: Regular, with spacing 1.0
             spacing in z: Regular, with spacing 1.0

julia> field = Field{Center, Center, Center}(grid)
Field located at (Center, Center, Center)
├── data: OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, size: (4, 1, 1)
├── grid: RectilinearGrid{Float64, Periodic, Periodic, Bounded}(Nx=4, Ny=1, Nz=1)
└── boundary conditions: west=Periodic, east=Periodic, south=Periodic, north=Periodic, bottom=ZeroFlux, top=ZeroFlux, immersed=ZeroFlux

julia> set!(field, [1, 2, 3, 4])

julia> sum(field)
10.0

julia> sum(field, condition = field .> 2)
7.0

julia> sum(field, condition = field .> 2, mask = 100)
207.0

julia> sum(field, condition = (i, j, k, grid, field) -> grid.xᶜᵃᵃ[i] < 0.3)
1.0
```

This type is used to automatically exclude the immersed region in the reduction of `AbstractField{<:Any, <:Any, <:Any, <:ImmersedBoundaryGrid}`

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
-version = "0.68.2"
+version = "0.68.3"
```
navidcy, line 3:
```suggestion
version = "0.68.4"
```

src/AbstractOperations/conditional_operations.jl
```diff
@@ -1,6 +1,7 @@
+using Oceananigans.Fields: OneField
 import Oceananigans.Fields: condition_operand, conditional_length, set!
 
-# For conditional reductions such as mean(condition_operand(u * v, 0, u > 0))
+# For conditional reductions such as mean(u * v, condition = u .> 0))
```
glwagner, line 4: This will change to `condition = u > 0` after #2169 !

src/Fields/field_reductions.jl
```diff
@@ -66,7 +66,7 @@ Base.show(io::IO, field::ReducedComputedField) =
           "├── data: $(typeof(field.data)), size: $(size(field))\n",
           "├── grid: $(summary(field.grid))\n",
           "├── operand: $(summary(field.operand))\n",
-          "└── status: ", show_status(field.status))
+          "└── status: ", field.status)
```
glwagner, line 69:
```suggestion
          "└── status: ", summary(field.status))
```

right?

src/AbstractOperations/metric_field_reductions.jl
```diff
@@ -40,7 +40,7 @@ function Reduction(avg::Average, field::AbstractField; dims)
         # Construct summand of the Average
         L⁻¹_field_dx = field * dx / L
 
-        return Reduction(sum!, L⁻¹_field_dx; dims)
+        return Reduction(sum!, L⁻¹_field_dx; condition, mask, dims)
```
glwagner, line 43: We can't wrap the operation itself? Eg

```suggestion
        return Reduction(sum!, ConditionalOperation(L⁻¹_field_dx; condition, mask), dims)
```

Then we don't have to change `Reduction` and there's no boiler plate (this implementation ties `ConditionalOperation` and `Reduction` together intimately which could be a headache later).

src/Fields/field_reductions.jl
```diff
@@ -2,9 +2,11 @@
 ##### Reductions of AbstractField
 #####
 
-struct Reduction{R, O, D}
+struct Reduction{R, O, D, C, M}
     reduce! :: R
     operand :: O
+    condition :: C
+    mask :: M
```
glwagner, line 9: ❌  no copy paste!

src/AbstractOperations/conditional_operations.jl
```diff
@@ -25,8 +25,8 @@ end
 @inline condition_operand(operand::ConditionalOperation, ::Nothing, mask) = condition_operand(location(operand)..., operand, operand.grid, operand.condition, mask)
 @inline condition_operand(operand::ConditionalOperation, condition, mask) = condition_operand(location(operand)..., operand, operand.grid, condition, mask)
 
-@inline conditional_length(c::ConditionalOperation)       = sum(condition_onefield(c, 0))
-@inline conditional_length(c::ConditionalOperation, dims) = sum(condition_onefield(c, 0); dims = dims)
+@inline conditional_length(c::ConditionalOperation)       = Int.(sum(condition_onefield(c, 0)))
+@inline conditional_length(c::ConditionalOperation, dims) = Int.(sum(condition_onefield(c, 0); dims = dims))
```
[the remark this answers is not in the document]
glwagner: Wait you shouldn't need this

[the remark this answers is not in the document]
simone-silvestri: `conditional_length(c)` is an integer but `conditional_length(c, dims)` is a field

src/AbstractOperations/conditional_operations.jl
```diff
@@ -15,16 +17,21 @@ struct ConditionalOperation{LX, LY, LZ, O, G, C, M, T} <: AbstractOperation{LX,
      end
 end
 
-@inline condition_operand(operand::AbstractField, condition, mask)    = condition_operand(location(operand)..., operand, operand.grid, condition, mask)
+const FieldOrConditional = Union{AbstractField, ConditionalOperation}
+
 @inline condition_operand(LX, LY, LZ, operand, grid, condition, mask) = ConditionalOperation{LX, LY, LZ}(operand, grid, condition, mask)
+@inline condition_operand(operand::AbstractField, condition, mask)    = condition_operand(location(operand)..., operand, operand.grid, condition, mask)
+@inline function condition_operand(operand::AbstractField, condition::AbstractArray, mask) 
+    condition = arch_array(architecture(operand.grid), condition)
```
glwagner, line 25: true

PR #2099, opened 2021-12-05
Make possible to have string(a::WENO5) = "WENO5"

### LINKED ISSUE #2098: `WENO5` is very different from other advection schemes
To get the string name of an advection scheme I can do something like the following:

```
scheme1 = UpwindBiasedThirdOrder()
string(scheme1)[1:end-2]
```
which produces "UpwindBiasedThirdOrder".  I am doing this to get titles in figures to compare the results of the different schemes.

However, when I try doing something similar with `WENO5`, it fails because it is designed very differenetly:

```
scheme2 = WENO5()
string(scheme2)
```

which instead produces

"WENO5 advection scheme with: \n    ├── X regular \n    ├── Y regular \n    └── Z regular"

I am not suggesting anything is wrong but can someone suggest how I might be able to get string names from these advection schemes more easily?  

@simone-silvestri ?

simone-silvestri:
Extended the print function so that 
string(::AdvectionScheme) = "AdvectionScheme"

closes #2098

### DIFF
--- a/src/Advection/centered_fourth_order.jl
+++ b/src/Advection/centered_fourth_order.jl
@@ -2,12 +2,16 @@
 ##### Centered fourth-order advection scheme
 #####
 
+import Base: print 
+
 """
     struct CenteredFourthOrder <: AbstractCenteredAdvectionScheme{1}
 
 Centered fourth-order advection scheme.
 """
 struct CenteredFourthOrder <: AbstractCenteredAdvectionScheme{1} end
+
+Base.print(io::IO, ::CenteredFourthOrder) = print(io, "CenteredFourthOrder")
 
 const C4 = CenteredFourthOrder
 const centered_fourth_order = C4()
--- a/src/Advection/centered_second_order.jl
+++ b/src/Advection/centered_second_order.jl
@@ -2,12 +2,16 @@
 ##### Centered second-order advection scheme
 #####
 
+import Base: print
+
 """
     struct CenteredSecondOrder <: AbstractAdvectionScheme{0}
 
 Centered second-order advection scheme.
 """
 struct CenteredSecondOrder <: AbstractAdvectionScheme{0} end
+
+Base.print(io::IO, ::CenteredSecondOrder) = print(io, "CenteredSecondOrder")
 
 boundary_buffer(::CenteredSecondOrder) = 0
 
--- a/src/Advection/upwind_biased_fifth_order.jl
+++ b/src/Advection/upwind_biased_fifth_order.jl
@@ -2,12 +2,16 @@
 ##### Upwind-biased 3rd-order advection scheme
 #####
 
+import Base: print 
+
 """
     struct UpwindBiasedFifthOrder <: AbstractUpwindBiasedAdvectionScheme{2}
 
 Upwind-biased fifth-order advection scheme.
 """
 struct UpwindBiasedFifthOrder <: AbstractUpwindBiasedAdvectionScheme{2} end
+
+Base.print(io::IO, ::UpwindBiasedFifthOrder) = print(io, "UpwindBiasedFifthOrder")
 
 const U5 = UpwindBiasedFifthOrder
 
--- a/src/Advection/upwind_biased_first_order.jl
+++ b/src/Advection/upwind_biased_first_order.jl
@@ -9,6 +9,8 @@ Upwind-biased first-order advection scheme.
 """
 struct UpwindBiasedFirstOrder <: AbstractUpwindBiasedAdvectionScheme{1} end
 
+Base.print(io::IO, ::UpwindBiasedFirstOrder) = print(io, "UpwindBiasedFirstOrder")
+
 const U1 = UpwindBiasedFirstOrder
 
 @inline boundary_buffer(::U1) = 1
--- a/src/Advection/upwind_biased_third_order.jl
+++ b/src/Advection/upwind_biased_third_order.jl
@@ -2,12 +2,15 @@
 ##### Upwind-biased 3rd-order advection scheme
 #####
 
+import Base: print
 """
     struct UpwindBiasedThirdOrder <: AbstractUpwindBiasedAdvectionScheme{1}
 
 Upwind-biased third-order advection scheme.
 """
 struct UpwindBiasedThirdOrder <: AbstractUpwindBiasedAdvectionScheme{1} end
+
+Base.print(io::IO, ::UpwindBiasedThirdOrder) = print(io, "UpwindBiasedThirdOrder")
 
 const U3 = UpwindBiasedThirdOrder
 
--- a/src/Advection/weno_fifth_order.jl
+++ b/src/Advection/weno_fifth_order.jl
@@ -180,7 +180,7 @@ end
 const JSWENO = WENO5{<:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:Nothing}
 const ZWENO  = WENO5{<:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:Any}
 
-Base.print(io::IO, a::WENO5) = print(io, "WENO5")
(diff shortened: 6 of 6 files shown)


francispoulin (approved):
It all looks good, thanks again for putting this together so quickly.

navidcy:
I'm skeptical whether this is needed.

See https://github.com/CliMA/Oceananigans.jl/issues/2098#issuecomment-986380458

navidcy:
Not that having `Base.print()` methods for advection schemes hurts to be honest. :)

glwagner:
Previously, `print` fell back to `show`?

francispoulin:
I agree with what @glwagner suggested, before this PR `WENO5` had a show but not a print.  

This PR adds the print to `WENO5`, which all other schemes have.  I think this is an improvement and allows us to deal with all advection schemes similarly. Is there any disadvantage to this?

glwagner:
There's a disadvantage if we need to use `print` for an application other than the one in #2098. Then this PR creates work for us in the future. We can't really predict whether that will occur (but it could come up in the context of log-file creation... ?)

I think the purpose of `print` is to remove "decorations" and "Julia-specific details", rather than important numerical information such as whether the WENO5 coefficients are grid-stretching-dependent:

```julia
help?> print
search: print println printstyled sprint isprint prevind parentindices precision

  print([io::IO], xs...)

  Write to io (or to the default output stream stdout if io is not given) a canonical
  (un-decorated) text representation. The representation used by print includes minimal
  formatting and tries to avoid Julia-specific details.

  print falls back to calling show, so most types should just define show. Define print if
  your type has a separate "plain" representation. For example, show displays strings with
  quotes, and print displays strings without quotes.

  string returns the output of print as a string.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> print("Hello World!")
  Hello World!
  julia> io = IOBuffer();

  julia> print(io, "Hello", ' ', :World!)

  julia> String(take!(io))
  "Hello World!"
```

Here, it looks like `print` returns the "base name" of the advection scheme's type, eg similar to `summary(scheme) = string(typeof(scheme))`, but throwing away type parameter information:

```julia
help?> summary
search: summary MethodSummary

  summary(io::IO, x)
  str = summary(x)

  Print to a stream io, or return a string str, giving a brief description of a value. By
  default returns string(typeof(x)), e.g. Int64.

  For arrays, returns a string of size and type info, e.g. 10-element Array{Int64,1}.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> summary(1)
  "Int64"

  julia> summary(zeros(2))
  "2-element Vector{Float64}"
```

@navidcy suggested `string(typeof(string).name.wrapper)` which seems to do the job of removing type parameter information from a type's string?

francispoulin:
Thanks for the explanation @glwagner .  I now understand what @navidcy was suggesting, and thank you both.  I agree that this works and I can certainly use this for what I needed to do.

However, I would suggest that we probably want the same strcuture for all of the advection schemes.  I'm not saying which is better, but uniformity is probably desired, right?

glwagner:
Doesn't `string(typeof(scheme).name.wrapper)` return something similar for all advection schemes? Or am I misunderstanding what you're saying?

francispoulin:
Sorry I was unclear @glwagner , yes, it certainly does that and I will use this moving forward.

In my previous comment, I was referring to my inital observation that when we use `string` for `WENO5` and any other scheme, we get very different results.  I'm not sure why we would not different results for different advection schemes.

glwagner:
I think something that we haven't fully grasped in the past (and I am largely to blame) is how to properly display information to users in the REPL and log files (eg by extending existing methods so the interface for displaying `Oceanangians` types is the same as other julia types). There are a few functions available:

* `show`
* `print` (seems similar to `show`, except used where fancy formatting such as colors, etc may not be appropriate or available)
* `summary`

We also want even more minimal representations than those produced by `summary`, like `string(typeof(obj).name.wrapper)`. There may also be a julia function for this, but I didn't find one after a cursory search.

We have (unfortunately) introduced what appears to be a needless alternative to `summary` in the source code; we call this `short_show`. So we already have sizeable technical debt...

I think it's helpful to follow convention as much as possible, especially if this package is used by other packages (eg it will be embedded in a climate model at some point in the future, which will be a separate package...)

glwagner:
> Sorry I was unclear @glwagner , yes, it certainly does that and I will use this moving forward.
> 
> In my previous comment, I was referring to my inital observation that when we use `string` for `WENO5` and any other scheme, we get very different results. I'm not sure why we would not different results for different advection schemes.

It makes sense that different advection schemes return different things from `string(scheme)` because some are more complicated than others, and thus require more characters to display type and value information. For example, `CenteredSecondOrder` is a singleton type; it has no values at all. But `WENO5` has values, because it contains coefficients that may be precomputed.

This is true for other objects; eg `UnitRange{Int64}` versus `Vector{Int64}`:

```julia
julia> string(1:3)
"1:3"

julia> string([1, 2, 3])
"[1, 2, 3]"
```

francispoulin:
Thanks for the explanation @glwagner .  No problem that you didn't understand this perfectly in the past as I am still learning the difference between `print`, `show`, and `summary`.

It seems that all the advection schemes have the same structure, except `WENO5` but this is not a problem and therefore I am happy to close #2098 .  This issue perhaps better addressed in a future PR, that focuses on the structure we would want in general.  Agreed?

simone-silvestri:
Ok you convinced me with the `print`. Better to use @navidcy's solution. On the other hand it feels a little clunky, so maybe we can define a `nameof` function which returns `string(typeof(scheme1).name.wrapper)`.

francispoulin:
`nameof` is a lot friendlier than the current approach. I would vote yes for this option.

glwagner:
> Ok you convinced me with the `print`. Better to use @navidcy's solution. On the other hand it feels a little clunky, so maybe we can define a `nameof` function which returns `string(typeof(scheme1).name.wrapper)`.

`nameof` is a cool idea. Is there a similar julia function?

francispoulin:
I checked the help and indeed there is!

```
help?> nameof
search: nameof

  nameof(m::Module) -> Symbol

  Get the name of a Module as a Symbol.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> nameof(Base.Broadcast)
  :Broadcast

  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  nameof(t::DataType) -> Symbol

  Get the name of a (potentially UnionAll-wrapped) DataType (without its parent module) as a symbol.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> module Foo
             struct S{T}
             end
         end
  Foo
  
  julia> nameof(Foo.S{T} where T)
  :S

  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  nameof(f::Function) -> Symbol

  Get the name of a generic Function as a symbol. For anonymous functions, this is a compiler-generated name. For
  explicitly-declared subtypes of Function, it is the name of the function's type.

```

glwagner:
Ok so I don't think we need `nameof` the way to write this is `string(nameof(typeof(scheme)))`, eg:

```julia
julia> struct Test{T}
       a :: T
       end

julia> t = Test(1.1)
Test{Float64}(1.1)

julia> nameof(typeof(t))
:Test

julia> string(nameof(typeof(t)))
"Test"
```

francispoulin:
Agreed.  I just tried it for a couple of schemes and agree that this works, and is maybe slightly easier to remember than the previous suggestion.

glwagner:
It's also always preferred to use a function (which represents a user interface) rather than directly referencing a `struct` implementation (eg referencing `.name.wrapper`) that's subject to change.

francispoulin:
Good to know.  

I am happy to close #2098 unless anyone has other concerns about this?

simone-silvestri:
Ok, if it's ok for everyone I'll also close this PR

simone-silvestri:
Ok, if it's ok for everyone I'll also close this PR

PR #2100, opened 2021-12-06
Implements a `ConsecutiveIterations` schedule

glwagner:
This PR implements a new schedule called `ConsecutiveIterations`. This schedule is constructed with a "parent schedule" as an argument and an optional second argument specifying the "number" of consecutive iterations to actuate for:

```julia
schedule = ConsecutiveIterations(TimeInterval(100))

schedule = ConsecutiveIterations(IterationInterval(4))

schedule = ConsecutiveIterations(IterationInterval(20), 3)
```

The schedule will then actuated (ie, triggering output writing or callback execution) at _both_ the moment specified by the parent schedule, and for `N` iterations consecutive to that one, where `N=1` by default.

This PR also adds a property `offset` to `IterationInterval`. Together, this allows one to compute time-derivatives (either forward derivative, or, if using `IterationInterval(N, offset=-1)`, a centered difference).

cc @whitleyv 

TODO

- [ ] Test

This could also maybe be called `SubsequentIterations`... ?


glwagner:
@whitleyv let us know if this feature allows you to compute the time-derivative of outputted quantities, and there's any changes or additional features that might help with its usability.

tomchor (approved):
Assuming this has been tested, I think this is a really nice solution, so I'm already approving it :)

Is it worth giving the option to compute a backward time derivative? (Meaning outputting the step immediately before, not after, the parent schedule? Or maybe even a central one?) 
@whitleyv do you think that would be useful or that it wouldn't produce a meaningful difference for reasonable values of Δt?

glwagner:
> Assuming this has been tested, I think this is a really nice solution, so I'm already approving it :)
> 
> Is it worth giving the option to compute a backward time derivative? (Meaning outputting the step immediately before, not after, the parent schedule? Or maybe even a central one?) @whitleyv do you think that would be useful or that it wouldn't produce a meaningful difference for reasonable values of Δt?

I think it's pretty worthwhile! The problem is that, except for `IterationInterval`, we usually can't _guarantee_ that a schedule will always actuate on the _next iteration_ (since the time-step can be modified by a callback, or by an alignment requirement associated with another schedule present in the simulation).

Users can implement a custom "just before iteration interval" schedule to achieve a backward difference (with respect to other quantities that are output on `IterationInterval` by writing

```julia
primary_schedule = IterationInterval(10)
just_before_iteration_interval(model) = model.clock.iteration + 1 % primary_schedule.interval == 0
schedule = ConsecutiveIterations(just_before_iteration_interval)
```

Thinking about it a bit, I think that we do want to calculate a forward difference for most purposes. The reason is that a time-stepping scheme is largely _explicit_, in the sense that it looks like

```
uⁿ⁺¹ = uⁿ + ∫ Gⁿ dt
```

Thus, the quantities of interest are `Gⁿ`, `uⁿ`, and `uⁿ⁺¹`.

glwagner:
> Assuming this has been tested

It hasn't; that's one of the todo's that must be completed before merging.

navidcy:
We can add a test. E.g. in the 1D diffusion equation using Euler we can use the `ConsecutiveIterations(TimeInterval(..))` to reproduce the RHS of the equation?

navidcy (approved):

src/Utils/schedules.jl
```diff
@@ -149,6 +149,37 @@ function specified_times_str(st)
...
+    parent :: S
+    previous_parent_actuation_iteration :: Int
+end
+
+"""
+    ConsecutiveIterations(parent_schedule)
+
+Return a `schedule::ConsecutiveIterations` that actuates both when `parent_schedule`
+actuates, and at iterations immediately following the actuation of `parent_schedule`.
+"""
```
navidcy, line 166:
```suggestion
actuates, and at iterations immediately following the actuation of `parent_schedule`.
This can be used, for example, when one wants to use output to numerically reconstruct the
time-derivative of a quantity.
"""
```

navidcy (approved):
So with this schedule one can reconstruct forward or backwards first-order time-derivative approximations. Can we reconstruct a centered first-order time-derivative? Seems like this can be easily generalized to save the `N` consecutive iterations following the actuation of the `parent_schedule` (instead of just 1). But we can leave this for a future PR. :)

glwagner:
> So with this schedule one can reconstruct forward or backwards first-order time-derivative approximations. Can we reconstruct a centered first-order time-derivative? Seems like this can be easily generalized to save the `N` consecutive iterations following the actuation of the `parent_schedule` (instead of just 1). But we can leave this for a future PR. :)

That's a good idea.

We can also generalize `IterationInterval` with an "offset" that could be used to center the output on a particular iteration, if using `IterationInterval`.

navidcy:
@glwagner, can you resolve conflicts? perhaps we merge and add an issue for test to be implemented in the future?

PR #2103, opened 2021-12-07
More tests for boundary conditions

navidcy:
Closes #2091.

PR #2110, opened 2021-12-09
Quality-of-life improvement for grid constructors

glwagner:
This PR adds a new constructor for `RectilinearGrid` that looks something like

```julia
grid = RectilinearGrid(Float64, size = (1, 1, 1), ...)
```

that defaults to `architecture = CPU()`.

This avoids the error observed on #2103. Since users _rarely_ change the floating point type, this alternative constructor probably won't be used all that much. So it's not crucial, but might help a few people.

If others (@navidcy, @simone-silvestri ?) think this is a good idea, I'll add the same for other grid constructors.

If we don't merge this PR, we should probably still validate the positional inputs to grid constructors to make sure they are valid (eg `architecture` needs to be `AbstractArchitecture`, etc).

### DIFF
--- a/src/Grids/rectilinear_grid.jl
+++ b/src/Grids/rectilinear_grid.jl
@@ -289,8 +289,9 @@ function RectilinearGrid(architecture::AbstractArchitecture = CPU(),
         Nx, Ny, Nz, Hx, Hy, Hz, Lx, Ly, Lz, Δxᶠᵃᵃ, Δxᶜᵃᵃ, xᶠᵃᵃ, xᶜᵃᵃ, Δyᵃᶜᵃ, Δyᵃᶠᵃ, yᵃᶠᵃ, yᵃᶜᵃ, Δzᵃᵃᶠ, Δzᵃᵃᶜ, zᵃᵃᶠ, zᵃᵃᶜ)
 end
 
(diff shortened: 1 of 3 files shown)


navidcy (approved):

simone-silvestri:
All for it!

src/Grids/rectilinear_grid.jl
```diff
@@ -253,7 +253,7 @@ RectilinearGrid{Float64, Periodic, Bounded, Bounded}
                 grid in z: Stretched, with spacing min=0.6826950100338962, max=1.8309085743885056
 ```
 """
-function RectilinearGrid(architecture = CPU(),
+function RectilinearGrid(architecture::AbstractArchitecture = CPU(),
                          FT = Float64;
```
navidcy, line 257:
```suggestion
                         FT::DataType = Float64;
```

glwagner: It's not needed for dispatch, but could help prevent bugs (though this usage violates YAS...)

Though typing on `architecture` also violates YAS (it's also not needed for dispatch).

glwagner: @navidcy ok to merge without this?

navidcy: > @navidcy ok to merge without this?

Sure!!

PR #2122, opened 2021-12-16
Add possible option to test with one file

christophernhill:
This lets me do 

export TEST_GROUP="none"
export TEST_FILE="test_matrix_poisson_solver.jl"

so I don't have to edit runtest to try a single file test.

Not sure if its good idea???

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -29,7 +29,6 @@ include("dependencies_for_runtests.jl")
             include("test_batched_tridiagonal_solver.jl")
             include("test_preconditioned_conjugate_gradient_solver.jl")
             include("test_poisson_solvers.jl")
-            include("test_matrix_poisson_solver.jl")
         end
     end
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -1,6 +1,11 @@
 include("dependencies_for_runtests.jl")
 
 @testset "Oceananigans" begin
+    if test_file != :none
+      @testset "Single file test" begin
+            include(String(test_file))
+      end
+    end
```
glwagner, line 8:
```suggestion
    if test_file != :none
        @testset "Single file test" begin
            include(String(test_file))
        end
    end
    
```

glwagner (approved):
Good --- however I recommend using `include("file_to_test.jl")` because this avoids recompilation. This allows a bit faster workflow via Revise. However, I'm fine to have this feature...

PR #2140, opened 2022-01-05
Split tests into smaller bits and pieces

### LINKED ISSUE #2139: GPU tests take way too long
Some of the buildkite pipelines take way too long. Specifically the `time_stepping_2`, `regression` and `solver` tests.

I suggest we further split them?

navidcy:
Closes #2139.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -86,10 +86,10 @@ steps:
 ##### Solver tests
 #####
 
-  - label: "🦅 gpu solver tests"
+  - label: "🦅 gpu poisson solvers tests"
     env:
       JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
-      TEST_GROUP: "solvers"
+      TEST_GROUP: "poisson-solvers"
     commands:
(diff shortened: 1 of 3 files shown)


navidcy:
Perhaps we should split the regression tests? A natural split can be done for each grid type, e.g., see:
https://github.com/CliMA/Oceananigans.jl/blob/59a8a81f8ff1a0835f9d549428d61d7295c6c5f6/test/test_regression.jl#L73

navidcy:
OK.

1. I don't understand why with https://github.com/CliMA/Oceananigans.jl/blob/b7a25e8b2bb31831cc8d8d355be9637961b9ea19/test/runtests.jl#L111-L114
the tests fail to recognise `grid_type`. See, e.g., https://buildkite.com/clima/oceananigans/builds/4915#25771fef-ddef-4095-b7cb-61ded273cbe0/27-308

2. Do we *really* want to do the regression tests both on regular grid and on a vertically **unstretched** grid? They are both the same grids... And there are tests in the `tests_grids.jl` I think that they tests that these grids are "equal". See https://github.com/CliMA/Oceananigans.jl/blob/affa548de5cfbad9d8f835a180cd69b617aadd95/test/test_grids.jl#L280-L288
So, I suggest we only tests for **one** grid type. What do you think?

simone-silvestri:
I agree, it would make everything more readable

glwagner:
I suggest we merge #2121 first.

PR #2141, opened 2022-01-10
Adds a correct simplified Inverse preconditioner

simone-silvestri:
This PR implements the formulation of the default preconditioner for GPU following Marshall J. et al., "Finite-volume, incompressible Navier Stokes model for studies of the ocean on parallel computers" (here called `SimplifiedInversePreconditioner`)

Also, the `Jacobi` preconditioner has been removed as it gives no benefits in terms of speed


glwagner:
What does `Simplified` mean? Perhaps there's a better term?

simone-silvestri:
I called it `Simplified` because it tries to mimic an inverse from heuristic rather than strict mathematical reasoning (but it still works very good!) so it doesn't quite look like a sparse inverse but it is applied in the same way. Maybe we can call it `ApproximateInversePreconditioner`?

benchmark/benchmark_hydrostatic_model.jl
```diff
@@ -103,10 +103,8 @@ grid_types = [
 free_surface_types = [
     :MatrixImplicitFreeSurface,
     :ExplicitFreeSurface,
-    # ImplicitFreeSurface doesn't yet work on MultiRegionGrids like the ConformalCubedSphereGrid:
-    :FFTImplicitFreeSurface, 
     :ImplicitFreeSurface,
-    # :ImplicitFreeSurface
+    :NoneImplicitFreeSurface
```
glwagner, line 107: Can we use `nothing` rather than adding a new type?

glwagner:
> I called it `Simplified` because it tries to mimic an inverse from heuristic rather than strict mathematical reasoning (but it still works very good!) so it doesn't quite look like a sparse inverse but it is applied in the same way. Maybe we can call it `ApproximateInversePreconditioner`?

Nice! "Approximate" is better I think!

I was under the impression that good preconditioners are _usually_ "approximate inverses" --- perhaps that's not true?

I think the more specific the name, the better.

Something to think about --- it looks like this stuff is in `Solvers`. If, someday, the `Solvers` module is moved outside Oceananigans, do all of these names still make sense? Or do these names only make sense in the context of a specific inverse problem?

benchmark/benchmark_hydrostatic_model.jl
simone-silvestri: you mean `nothingImplicitFreeSurface` here? Or switch `preconditioner_method = :None` to `preconditioner_method = nothing`?

simone-silvestri:
yes, they are always approximate inverses. I wanted to differentiate between the `SparseInverse` which is constructed by mathematical reasoning and `SimplifiedInverse` which is build on heuristic reasoning

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
-version = "0.66.3"
+version = "0.67.1"
```
navidcy, line 3: why is this considered change? version is 0.67.1 already?
@simone-silvestri perhaps bring this branch up to date with main?

navidcy (changes requested):

Project.toml
simone-silvestri: Done!

src/Models/HydrostaticFreeSurfaceModels/pcg_implicit_free_surface_solver.jl
```diff
@@ -189,7 +189,7 @@ end
 """
     _implicit_free_surface_precondition!(P_rⁿ⁺¹, grid, r, ∫ᶻ_Axᶠᶜᶜ, ∫ᶻ_Ayᶜᶠᶜ, g, Δt)
 
-Return the simplified inverse preconditioner applied to the residuals in the form
+Return the heuristic inverse preconditioner applied to the residuals in the form
```
glwagner, line 192:
```suggestion
Return the asymptotic inverse preconditioner applied to the residuals in the form
```

I think a good long description of this preconditioner is that its an asymptotic expansion of the inverse for diagonally-dominant systems.

glwagner (approved):

Project.toml
simone-silvestri: Changes are done! I'll merge

PR #2150, opened 2022-01-13
Bug fix in stretched grid!

simone-silvestri:
At the moment 

`ΔF[i-1] = C[i] - C[i-1]` where `C` and `F` are the center and face coordinate, respectively. On the other hand, the derivative on the face `[i]` is calculated with `(c[i] - c[i-1]) / ΔF[i]` (where `c` is the value of the derived variable at the centers)

therefore, it has to be that `ΔF[i] = C[i] - C[i-1]`

Because of how we test, this bug was miraculously eluding all testing

### DIFF
--- a/src/Grids/grid_generation.jl
+++ b/src/Grids/grid_generation.jl
@@ -55,15 +55,13 @@ function generate_coordinate(FT, topology, N, H, coord, architecture)
 
     ΔC = [F[i + 1] - F[i] for i = 1:TF-1]
 
-    ΔF = OffsetArray(ΔF, -H)
-    ΔC = OffsetArray(ΔC, -H)
+    ΔF = [ΔF[1], ΔF..., ΔF[end]]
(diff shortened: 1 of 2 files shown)


glwagner:
I definitely agree that `ΔF[i] = C[i] - C[i-1]`!

glwagner:
But this is confusing notation... can we update to use the triple notation that we use throughout the code?

The confusion is that MITgcm uses the notation `ΔzF` to denote the _difference of_ `zF`. This is _opposite_ the meaning of our superscripts, which indicate the _location_. In our notation `Δz^aaf` is the _difference of_ `z^aac`. 

I suggest we just avoid these capital letters and use only superscripts...

glwagner (approved):
I approve, but let's use `Δᶠ` instead of `ΔF`. It's too easy to confuse `ΔF` with MITgcm notation.

PR #2151, opened 2022-01-15
Split tests into smaller bits and pieces (take #2)

### LINKED ISSUE #2139: GPU tests take way too long

navidcy:
This is a second take on #2140. Closes #2139.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -370,10 +370,10 @@ steps:
 ##### Regression
 #####
 
-  - label: "🐫 gpu regression tests"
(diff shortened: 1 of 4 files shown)


glwagner (approved):
Nice!

I think we should also move some GPU tests to tartarus, but we can do this here or later.

navidcy:
Do as you please!

navidcy:
If tests pass then I suggest we merge and leave further work for future PR (perhaps after OSM?)

PR #2152, opened 2022-01-18
Small fix for single column model mode

glwagner:
PR #2121 changed some function signatures, which broke functionality provided by `single_column_model_mode.jl` that sets `model.free_surface = nothing`. This PR restores that functionality.

I think we could add a test, but the experimental nature of `single_column_model_mode.jl` could also mean it'd be better to wait.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
-version = "0.68.0"
+version = "0.68.2"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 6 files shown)


navidcy:
Something's still not right... :(

navidcy:
Do we need a `with_halo` method for `LatitudeLongitudeGrid`?

simone-silvestri:
A with_halo for lat lon grid can be very usefull

glwagner:
Huh, not sure how `with_halo` for the lat lon grid got deleted... I'll put it back.

navidcy (approved):

navidcy:
I think the previews folder in `gh-pages` has overgrown again... Will deal with it! :)

navidcy:
ok, merge?

PR #1732, opened 2021-06-01
Validate field boundary conditions

glwagner:
This long overdue PR implementations a validation step in the lowest-level outer constructor for `Field`.

This resolves a number of issues but I need to find them.

Couple todo's:
- [ ] tests
- [x] expand error message to help users choose valid boundary conditions

PR #2163, opened 2022-01-20
Validate boundary conditions on Fields

### LINKED ISSUE #419: Check boundary condition arrays are CuArrays if arch=GPU
The code segfaults otherwise.

### LINKED ISSUE #890: Check for 'invalid' boundary conditions
Incorrect behavior results if, for example, a `NormalFlow` boundary condition is applied to a tracer or non-wall-normal velocity component. Conversely, only the `NormalFlow` boundary condition type is correct for a wall-normal velocity component.

We may want to validate the boundary condition types in the constructors for `XFaceField`, `YFaceField`, `ZFaceField`, and `CellField`. I think this might save some users a bit of pain.

glwagner:
This PR adds a step in the outer `Field` constructor that validates boundary conditions. We check three things:

1. That boundary conditions are compatible with the topology. Namely, we are restricted to default choices in `Periodic` or `Flat` directions; anything but default throws an error.

2. If a topology is `Bounded`, we check that boundary conditions are compatible with field location. Mostly we cannot support flux, value, or gradient boundary conditions for fields at faces, and we only support `nothing` for fields in `Flat` directions.

3. That boundary condition arrays are on the right architecture.

TODO:

- [x] Test

Resolves #419
Resolves #890

Supercedes #1732

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -38,6 +38,7 @@ steps:
       TEST_GROUP: "init"
       CUDA_VISIBLE_DEVICES: "-1"
       JULIA_BINDIR: "$TARTARUS_HOME/julia-$JULIA_VERSION/bin"
+      TMPDIR: "$TARTARUS_HOME/tmp"
     commands:
       # Download julia binaries
       - "wget -N -P $TARTARUS_HOME https://julialang-s3.julialang.org/bin/linux/x64/$JULIA_MINOR_VERSION/julia-$JULIA_VERSION-linux-x86_64.tar.gz"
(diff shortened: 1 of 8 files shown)


navidcy:
Is tartarus out of disk space?
https://buildkite.com/clima/oceananigans/builds/5250#86d62c38-a78a-487f-8f1f-5c5ae4a55a08/19-577

navidcy (approved):

glwagner:
> Is tartarus out of disk space? https://buildkite.com/clima/oceananigans/builds/5250#86d62c38-a78a-487f-8f1f-5c5ae4a55a08/19-577

Yes although #2164 was supposed to fix that

glwagner:
Ah, but we need to also change `TMPDIR` for the documentation build!

glwagner:
> Is tartarus out of disk space? https://buildkite.com/clima/oceananigans/builds/5250#86d62c38-a78a-487f-8f1f-5c5ae4a55a08/19-577

Also just to clarify, tartarus has plenty of space but the root directory `/` is only 50 GB large, and thus prone to filling up. We build everything on another disk, but julia uses `/tmp` as a "temporary directory" by default. So we just need to ask julia to use a different temporary directory on a partition with more storage.

PR #2169, opened 2022-01-21
Binary comparison operators and more robust support for Fields-as-conditionals

glwagner:
This PR adds comparison operators to our supported operators, and also cleans up `Field` constructors for cases where computations or reductions return non-floating-point elements (such as Booleans).

This support is progress towards allowing users to specify things like conditional averages as diagnostics, using `AbstractOperation`s.

It's still a WIP because we need tests, and probably should also wait for a few other PRs before merging (eg #2097).


glwagner:
We can return to this kind of feature when it's needed.

PR #2174, opened 2022-01-23
Let regridding work on non-`SingleColumnGrid` +  better `total_length` on `Flat` grids

### LINKED ISSUE #2067: Refactor `regrid!` so that it works with non-flat `x` and `y` dimensions?
Currently we only support regrinding for `SingleColumnGrid` (i.e., grids in which `x` and `y` dimensions are `Flat`).

https://github.com/CliMA/Oceananigans.jl/blob/02f4d9dc17293a399ca3adf68b6de71ee7d59f16/src/Fields/regridding_fields.jl#L60

But the same code could also work as is, only for regrinding in `z` even if `x` and `y` dimensions weren't `Flat`. Isn't it?

cc: @glwagner

glwagner:
This PR checks whether two grids are only different in the vertical dimension using runtime information, rather then relying on both fields being `SingleColumnGrid`.

It also extends `total_length` for directions that are `Flat`, but have non-zero grid points (for simulations representing ensembles of simulations).

Closes #2067

### DIFF
--- a/src/Fields/regridding_fields.jl
+++ b/src/Fields/regridding_fields.jl
@@ -47,27 +47,38 @@ output_field[1, 1, :]
 """
 regrid!(a, b) = regrid!(a, a.grid, b.grid, b)
 
-function regrid!(a, target_grid, source_grid, b)
-    msg = """Regridding
-             $(summary(b)) on $(summary(source_grid))
-             to $(summary(a)) on $(summary(target_grid))
-             is not supported."""
+function we_can_regrid(a, target_grid, source_grid, b)
+    # Only 1D regridding is supported, so check that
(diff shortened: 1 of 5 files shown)


src/Fields/regridding_fields.jl
```diff
@@ -47,28 +47,39 @@ output_field[1, 1, :]
 """
 regrid!(a, b) = regrid!(a, a.grid, b.grid, b)
 
+function we_can_regrid(a, target_grid, source_grid, b)
+    # Only 1D regridding is supported, so check that
```
navidcy, line 51:
```suggestion
    # Only 1D regridding in the vertical is supported, so check that
```

navidcy (approved):

PR #2175, opened 2022-01-24
Update packages

navidcy:


simone-silvestri (approved):

francispoulin (approved):
I glanced over the changes and nothing seems to be out or order.  If tests pass then we should be good to go.

glwagner:
Valiant effort. Looks like some matrix solver tests don't pass --- @simone-silvestri perhaps you can take a look when you have time.

simone-silvestri:
I ll take a look

navidcy:
Seems that it passed when I rerun.

PR #2181, opened 2022-01-25
Cleanup in the implicit free surface tests + add `matrix_implicit_free_surface_solver_tests`

navidcy:
Some cleaning up in the implicit free surface solvers tests.

Resolves #2184.

PR #2186, opened 2022-01-26
Change regridding criterion to check field size rather than grid size

glwagner:
~Also bumps to 0.68.8.~

The application for this kind of regridding is from a field that, while living on a 3D grid, is reduced in x and y. Because the field is reduced in x and y it's a valid source field for a target field that's on a 1D grid.

We should probably add a test to this PR (regridding from a reduced field on a 3D grid to a non-reduced field on a 1D grid).

cc @adelinehillier @navidcy

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
-version = "0.68.7"
+version = "0.68.8"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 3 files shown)


navidcy:
I'll add a test.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
-version = "0.68.7"
+version = "0.68.8"
```
navidcy, line 3:
```suggestion
version = "0.68.7"
```

navidcy:
I brought this back to 0.68.7.

Let's wait for this PR and also for #2181 before bumping up minor release.

navidcy (approved):

glwagner:
Thanks so so much @navidcy

Issue #2185, opened 2022-01-26
Mean of stretched dimension using `WindowedSpatialAverage`


tomchor:
I can see through the definition of `Reduction` that averages in stretched grids are correctly calculated: https://github.com/CliMA/Oceananigans.jl/blob/2f9269df54e836117d9ac8b9790d1c3383bd6384/src/AbstractOperations/metric_field_reductions.jl#L29-L46

However, the same doesn't happen for `WindowedSpatialAverage`: https://github.com/CliMA/Oceananigans.jl/blob/2f9269df54e836117d9ac8b9790d1c3383bd6384/src/Diagnostics/windowed_spatial_average.jl#L44-L50

What's the preferred way to include this into `WindowedSpatialAverage`? Or maybe it would be best to abandon `WindowedSpatialAverage` and add the slicing functionality to `Reduction`?

tomchor:
Actually, I'm just noticing that `Reduction` has a option `mask`. Is that already enough to mimic the windowing behavior?

glwagner:
For reducing any `AbstractField` over "windows", `ConditionalOperation` is sufficient. The task there is to define `condition` appropriately (`mask` is chosen as the "neutral" value for the particular reduction). I guess there isn't an ordinary constructor for `ConditionalOperand`, but there is a function `condition_operand`...

https://github.com/CliMA/Oceananigans.jl/blob/b5bf45ed51587954f61ca7492e946fef69ff72ee/src/AbstractOperations/conditional_operations.jl#L23

(@simone-silvestri why is this function not simply a constructor for `ConditionalOperand`?) This feature is new so the API might change. We also need doc strings...

## On views of `Field`...

We may also eventually want to implement the concept of `view`s of `Field`s. I think if that were implemented properly, we'd be able to reduce those objects. 

#2177 is related; the location `Level` proposed there could handle slices across a whole dimension (but not necessarily "windowing"...)

This is for another issue, but field "windows" might be easy to support by using a `view` for `data`. Cause then:

```julia
julia> a = rand(4, 4)
4×4 Matrix{Float64}:
 b0.720962  0.484984  0.935646  0.453511
 0.41186   0.805359  0.105536  0.724929
 0.738268  0.287272  0.24803   0.782426
 0.777353  0.999748  0.515442  0.321921

julia> b = view(a, 2:3, 2:3)
2×2 view(::Matrix{Float64}, 2:3, 2:3) with eltype Float64:
 0.805359  0.105536
[...]

PR #2193, opened 2022-01-26
Removes `WindowedSpatialAverage`

tomchor:
It seems like the behavior of `WindowedSpatialAverage` can now be achieved by using `Field(Average(field, dims=dims, condition=condition))`. It might be a bit slower, but it's more flexible, so I think we should get rid of `WindowedSpatialAverage`.

Closes https://github.com/CliMA/Oceananigans.jl/issues/2185


navidcy:
This is a breaking change, so a patch release is in order.

glwagner (approved):

PR #2195, opened 2022-01-27
Update docs/docstrings

### LINKED ISSUE #2194: `regrid!` is (a bit) more general than it advertises
After #2186 I believe that the warning in the docstring of `regrid!` should be amended.

https://github.com/CliMA/Oceananigans.jl/blob/e90b22bcfcf3cba53ed61e46024d3a759f35ef56/src/Fields/regridding_fields.jl#L9-L16

navidcy:
A few updates.

Closes #2194.

### DIFF
--- a/src/Fields/field_reductions.jl
+++ b/src/Fields/field_reductions.jl
@@ -11,7 +11,8 @@ end
 """
     Reduction(reduce!, operand; dims)
 
-Return a `Reduction` of `operand` with `reduce!`, along `dims`.
+Return a `Reduction` of `operand` with `reduce!`, along `dims`. Note that `Reduction`
+expects a `reduce!` operation that works in-place.
 
 Field elements for which `condition(i ,j ,k) == true` are masked with a `mask`.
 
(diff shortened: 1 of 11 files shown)


glwagner (approved):

src/Fields/field_reductions.jl
```diff
@@ -13,21 +13,35 @@ end
 
 Return a `Reduction` of `operand` with `reduce!`, along `dims`.
 
-elements for which `condition(i ,j ,k) == true` are masked with a `mask`
+Field elements for which `condition(i ,j ,k) == true` are masked with a `mask`.
 
 Example
```
tomchor, line 18: Should we mention somewhere here that the reduction must happen in-place? (Or is that not the case?)

navidcy: you mean that the reduction operator must be a function that works 'in place'?

tomchor: Yes!

navidcy: Done via  9713191

tomchor (approved):

simone-silvestri (approved):

PR #2196, opened 2022-01-27
Fixes `FreeSurface` constructor for `SingleColumnGrid` and adds a test

glwagner:
I thought I had fixed this previously but it looks like not. So I've added a test that `isnothing(model.free_surface)` when `grid isa SingleColumnGrid`.

For the sake of a verbose explanation:

When `grid isa SingleColumnGrid`, the constructor-helper `FreeSurface` should return `nothing` rather than some other free surface object, like `ExplicitFreeSurface`, etc.

THEN, when `isnothing(model.free_surface)`, the handy function `fields(model)` should _omit_ the free surface displacement `η` from the `NamedTuple` of model fields.

A bug crept into the code because we didn't have a test during #2121 when some of the function signatures changed (because we don't need `arch` _and_ `grid` if `arch` is stored in `grid`).

cc @navidcy 
cc @adelinehillier


navidcy (approved):

navidcy:
@glwagner, is e1b6c9b what you meant to do?

glwagner:
> @glwagner, is [e1b6c9b](https://github.com/CliMA/Oceananigans.jl/commit/e1b6c9bda59ebab7595017e206080c504e1aaee8) what you meant to do?

Yes... `:η` is the "name" of the field.

PR #2200, opened 2022-01-27
Changes to `ConditionalOperation` 

### LINKED ISSUE #2188: User interface to `ConditionalOperand`
We need a user interface to `ConditionalOperand`, so it can be user-facing.

#2185 is related.

cc @simone-silvestri

simone-silvestri:
In case we have reductions on a function (e.g `maximum(abs, f, condition = (i, j, k, f.grid, f) -> i < 3)`) we want to be able to apply the function to the field _first_ and mask the field where `condition = true`  _after_

_to be more specific:_
When a reduction is called with a function as an argument such as
```
sum(sin, a::AbstractField)
```
the function will be evaluated element-wise with `mapreduce`. i.e., the function is applied _first_ and the reduction is applied _after_
```
mapreduce(x -> sin(x), +, a)
```
For this reason the `neutral_element` of a "functional" reduction is not the `neutral_element` of its basic operation (as we implemented) but the neutral element of `basic_operation(f())`. The neutral element is found by inverting this operation `f⁻¹(neutral_element(basic_operation))` In the example above it would be
```
neutral_element = sin⁻¹(neutral_element(+)) = sin⁻¹(0)
```

In principle, we should mask the field with `f⁻¹(neutral_element(reduction))` so that applying `f` and then the reduction would effectively exclude the masked areas

Unfortunately, we do not know a-priori the inverse of the function. Therefore, this PR implements a workaround:

- `ConditionalOperation` has now a property `func`,  which is a unary function applied to the elements of `operand` in `Base.getindex`
- functional reductions are implemented as `sum(f, c; condition) = sum(identity, conditional_operand(f, c, condition))` 

this PR also introduces a docstring and a `show` method for `ConditionalOperation` 
closes #2188

### DIFF
--- a/src/AbstractOperations/conditional_operations.jl
+++ b/src/AbstractOperations/conditional_operations.jl
@@ -120,7 +120,7 @@ end
 @inline get_condition(condition::AbstractArray, i, j, k, grid, args...) = condition[i, j, k]
 
 
-Base.summary(operation::ConditionalOperation) = string("Conditioned field at ", show_location(operation))
+Base.summary(c::ConditionalOperation) = string("ConditionalOperand of ", summary(c.operand), " with condition ", summary(c.condition))
 
 Base.show(io::IO, operation::ConditionalOperation) =
     print(io,
--- a/src/AbstractOperations/conditional_operations.jl
+++ b/src/AbstractOperations/conditional_operations.jl
@@ -90,7 +90,7 @@ end
 @inline truefunc(args...) = true
 
 @inline condition_onefield(c::ConditionalOperation{LX, LY, LZ}, mask) where {LX, LY, LZ} =
-                              ConditionalOperation{LX, LY, LZ}(OneField(), identity, c.grid, c.condition, mask)
+                              ConditionalOperation{LX, LY, LZ}(OneField(Int), identity, c.grid, c.condition, mask)
 
 @inline conditional_length(c::ConditionalOperation)       = sum(condition_onefield(c, 0))
 @inline conditional_length(c::ConditionalOperation, dims) = sum(condition_onefield(c, 0); dims = dims)
--- a/src/AbstractOperations/conditional_operations.jl
+++ b/src/AbstractOperations/conditional_operations.jl
@@ -40,6 +40,9 @@ Keyword arguments
 
 - `mask`: the scalar mask
 
+`condition_operand` is the function used to construct a `ConditionalOperation`
+
+`condition_operand(func::Function, operand::AbstractField, condition, mask) = ConditionalOperation(operand; func, condition, mask)`
 
 Example
 =======
--- a/src/AbstractOperations/conditional_operations.jl
+++ b/src/AbstractOperations/conditional_operations.jl
@@ -12,20 +12,19 @@ struct ConditionalOperation{LX, LY, LZ, O, F, G, C, M, T} <: AbstractOperation{L
     condition :: C
     mask :: M
 
-    @doc """
-        ConditionalOperation{LX, LY, LZ}(operand, func, grid, condition, mask)
-
-    Returns a masking procedure whereas indexing a `ConditionalOperation` returns
-     `func(getindex(operand, args...))` where `condition == true` and `mask` where `condition == false`.
-    """
      function ConditionalOperation{LX, LY, LZ}(operand::O, func::F, grid::G, condition::C, mask::M) where {LX, LY, LZ, O, F, G, C, M}
          T = eltype(operand)
          return new{LX, LY, LZ, O, F, G, C, M, T}(operand, func, grid, condition, mask)
      end
 end
 
-"""
-    ConditionalOperation(operand; func=identity, condition=nothing, mask=0)
+@doc """
+ConditionalOperation{LX, LY, LZ}(operand, func, grid, condition, mask)
+
+Returns a masking procedure whereas indexing a `ConditionalOperation` returns
+`func(getindex(operand, args...))` where `condition == true` and `mask` where `condition == false`.
+
+ConditionalOperation(operand; func=identity, condition=nothing, mask=0)
 
 Returns a `ConditionalOperation` which allows masked reductions over the `AbstractField` `operand`.
 
--- a/src/AbstractOperations/conditional_operations.jl
+++ b/src/AbstractOperations/conditional_operations.jl
@@ -54,14 +54,16 @@ julia> using Oceananigans.Fields: condition_operand
 
 julia> c = CenterField(RectilinearGrid(size=(2, 1, 1), extent=(1, 1, 1)));
 
-julia> d = condition_operand(c, (i, j, k, grid, c) -> i < 2, 10)
-Conditioned Field at (Center, Center, Center)
+julia> f(i, j, k, grid, c) = i < 2; d = condition_operand(cos, c, f, 10)
+Conditioned field at (Center, Center, Center)
+├── operand: 2×1×1 Field{Center, Center, Center} on RectilinearGrid on CPU
 ├── grid: 2×1×1 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 1×1×1 halo
-└── tree: 
-    Conditioned Field at (Center, Center, Center)
+├── func: typeof(cos)
+├── condition: typeof(f)
+└── mask: 10
 
 julia> d[1, 1, 1]
-0.0
+1.0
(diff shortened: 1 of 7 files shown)


glwagner:
Can we also use `mask = f(neutral_value)` or no?

simone-silvestri:
what is the `neutral_value`? I think if we don't change `mapreduce` we want to have something like `mask = f⁻¹(get_neutral_value(reduction!))`, now, how to implement this can get a bit sketchy

glwagner:
> what is the `neutral_value`? If there is one implemented in Julia that's cool. I saw something about a `GPUarrays.neutral_element` (https://github.com/JuliaGPU/GPUArrays.jl/blob/master/src/host/mapreduce.jl) don't know if we can use it
> 
> I think if we don't want to change `mapreduce` we have to have something like `mask = f⁻¹(get_neutral_value(reduction!))`, which might be very difficult as not all the functions are readily invertible.
> 
> For the moment I added a function as a property of `ConditionalOperation` such that `getindex` of `c::ConditionalOperation` is `c.func(getindex(c.operand, args...))` when `c.condition = true`. Then a reduction like `sum(f, c, condition=condition)` will be called on a `ConditionalOperation` with `func = f` and `condition = condition`. I am thinking about other methods which do not require adding a property to the type though

I mean changing

https://github.com/CliMA/Oceananigans.jl/blob/d6a515e2ce5de7c0b10dc29427f96d777dafcf01/src/Fields/field.jl#L422-L423

to

```julia
Base.$(reduction!)(f::Function, r::ReducedField, a::AbstractArray;
                           condition = nothing, mask = f(get_neutral_mask(Base.$(reduction!))), kwargs...)
```

so we pass `get_neutral_mask(r)` through `f`.

glwagner:
This also affects initialization, which I think we handle:

https://github.com/CliMA/Oceananigans.jl/blob/d6a515e2ce5de7c0b10dc29427f96d777dafcf01/src/Fields/field.jl#L377-L378

simone-silvestri:
but then the function is applied in `mapreduce` (basically you sum `f(vector[i])`) so in this case, if we pass `f(get_neutral_mask(Base.$(reduction!))` to mask places we want to exclude, then we would sum over `f(f(get_neutral_mask(Base.$(reduction!)))` which is not the same as `f(f⁻¹(get_neutral_mask(Base.$(reduction!))) == get_neutral_mask(Base.$(reduction!)` which is what we want.

As a practical example if we use `mask = f(get_neutral_mask(Base.$(reduction!))` in case of a sum and `f(x) = x +1` we would have `sum(f, c, mask = 1, condition = (i, j, k, grid, c) -> i < Nx/2)`. Then we would subsitute `1` where `condition == true` and this would effectively add `2` for each `i >= Nx/2` (`f(mask)`). We would like to have `mask = -1` which is `f⁻¹(x)` but that is not really possible as we don't know `f⁻¹(x)` for each `f(x)` a-priori.

Or maybe I haven't interpreted correctly what you wrote?

glwagner:
> but then the function is applied in `mapreduce` (basically you sum `f(vector[i])`) so in this case, if we pass `f(get_neutral_mask(Base.$(reduction!))` to mask places we want to exclude, then we would sum over `f(f(get_neutral_mask(Base.$(reduction!)))` which is not the same as `f(f⁻¹(get_neutral_mask(Base.$(reduction!))) == get_neutral_mask(Base.$(reduction!)` which is what we want.
> 
> As a practical example if we use `mask = f(get_neutral_mask(Base.$(reduction!))` in case of a sum and `f(x) = x +1` we would have `sum(f, c, mask = 1, condition = (i, j, k, grid, c) -> i < Nx/2)`. Then we would subsitute `1` where `condition == true` and this would effectively add `2` for each `i >= Nx/2` (`f(mask)`). We would like to have `mask = -1` which is `f⁻¹(x)` but that is not really possible as we don't know `f⁻¹(x)` for each `f(x)` a-priori.
> 
> Or maybe I haven't interpreted correctly what you wrote?

No that's a good explanation. The "neutral element" for an arbitrary reduction with element-wise transformation is inverse transformation applied to the identity neutral element. Since we are unable (probably...) to obtain the inverse of arbitrary transformations, we need to implement a workaround. I think its a good idea to add some version of this explanation to the PR description.

src/AbstractOperations/metric_field_reductions.jl
```diff
@@ -32,7 +32,7 @@ function Reduction(avg::Average, field::AbstractField; condition = nothing, mask
 
     if all(d in regular_dimensions(field.grid) for d in dims)
         # Dimensions being reduced are regular; just use mean!
-        return Reduction(mean!, condition_operand(field, condition, mask); dims)
+        return Reduction(mean!, condition_operand(identity, field, condition, mask); dims)
```
glwagner, line 35: Might save some typing if we have

```julia
condition_operand(operand::AbstractField, condition, mask) = condition_operand(identity, operand, condition, mask)
```

src/AbstractOperations/conditional_operations.jl
```diff
@@ -5,45 +5,109 @@ import Oceananigans.Fields: condition_operand, conditional_length, set!
...
+function ConditionalOperation(operand::AbstractField; func = identity, condition = nothing, mask = 0)
+    grid = operand.grid
+    LX, LY, LZ = location(operand)
+    return ConditionalOperation{LX, LY, LZ}(operand, func, grid, condition, mask)
+end
+
+function ConditionalOperation(c::ConditionalOperation; func = nothing, condition = nothing, mask = nothing)
+    LX, LY, LZ = location(c)
+    return ConditionalOperation{LX, LY, LZ}(c.operand, c.func, c.grid, c.condition, c.mask)
+end
```
glwagner, line 81:
```suggestion
ConditionalOperation(c::ConditionalOperation) = c
```

src/AbstractOperations/conditional_operations.jl
```diff
@@ -5,45 +5,109 @@ import Oceananigans.Fields: condition_operand, conditional_length, set!
...
-# If we reduce a ConditionalOperation with no condition, we keep the original condition!
-@inline condition_operand(operand::ConditionalOperation, ::Nothing, mask) = condition_operand(location(operand)..., operand, operand.grid, operand.condition, mask)
+@inline condition_operand(func::typeof(identity), c::ConditionalOperation, ::Nothing, mask) = ConditionalOperation(c; mask)
+@inline condition_operand(func::Function, c::ConditionalOperation, ::Nothing, mask) = ConditionalOperation(c; func, mask)
+
+@inline truefunc(args...) = true
 
 @inline condition_onefield(c::ConditionalOperation{LX, LY, LZ}, mask) where {LX, LY, LZ} =
-                              ConditionalOperation{LX, LY, LZ}(OneField(), c.grid, c.condition, mask)
+                              ConditionalOperation{LX, LY, LZ}(OneField(), identity, c.grid, c.condition, mask)
```
glwagner, line 93:
```suggestion
                              ConditionalOperation{LX, LY, LZ}(OneField(Int), identity, c.grid, c.condition, mask)
```

src/AbstractOperations/conditional_operations.jl
```diff
@@ -58,3 +119,14 @@ end
 @inline get_condition(condition, i, j, k, grid, args...)                = condition(i, j, k, grid, args...)
 @inline get_condition(condition::AbstractArray, i, j, k, grid, args...) = condition[i, j, k]
 
+
+Base.summary(operation::ConditionalOperation) = string("Conditioned field at ", show_location(operation))
```
glwagner, line 123:
```suggestion
Base.summary(c::ConditionalOperation) = string("ConditionalOperand of ", summary(c.operand), " with condition ", summary(c.condition))
```

src/AbstractOperations/conditional_operations.jl
```diff
@@ -58,3 +119,14 @@ end
 @inline get_condition(condition, i, j, k, grid, args...)                = condition(i, j, k, grid, args...)
 @inline get_condition(condition::AbstractArray, i, j, k, grid, args...) = condition[i, j, k]
```
glwagner, line 120:
```suggestion
@inline get_condition(condition::AbstractArray, i, j, k, grid, args...) = @inbounds condition[i, j, k]
```

glwagner (approved):

glwagner:
@simone-silvestri tests haven't passed!

simone-silvestri:
Ouch

PR #1260, opened 2020-12-07
Add support for interpolating abstract operations

ali-ramadhan:
Need to add some tests before merging though.

cc @tomchor

PR #2204, opened 2022-01-31
Generalize `interpolate(field, x, y, z)`

glwagner:
Looks like this only worked for `Field`. Not it should work for `AbstractOperations` too. Supercedes #1260 .


francispoulin (approved):
Makes sense

PR #2205, opened 2022-01-31
Include `diffusive_cfl` in `TimeStepWizard`'s docstring

tomchor:
This PR includes the `diffusive_cfl` keyword argument in the docstring (specifically the constructor's signature) of `TimeStepWizard`. The only arguments that aren't listed now are the advective and diffusive time scales. I chose to keep them this way in the PR because it seems like those options aren't user-facing.

### DIFF
--- a/src/Simulations/time_step_wizard.jl
+++ b/src/Simulations/time_step_wizard.jl
@@ -12,7 +12,7 @@ end
 infinite_diffusion_timescale(args...) = Inf # its not very limiting
 
 """
-    TimeStepWizard(cfl=0.2, max_change=1.1, min_change=0.5, max_Δt=Inf, min_Δt=0.0)
+    TimeStepWizard(cfl=0.2, diffusive_cfl=Inf, max_change=1.1, min_change=0.5, max_Δt=Inf, min_Δt=0.0)
 
 Callback for adapting simulation time-steps `Δt` to maintain the advective
 Courant-Freidrichs-Lewy (`cfl`) number, the `diffusive_cfl`, while maintaining


glwagner (approved):
> I chose to keep them this way in the PR because it seems like those options aren't user-facing.

Yes, and also hopefully deprecated soon when we implement a more generic CFL calculator that's valid (meaning we don't need a special function for that).

Are you using `diffusive_cfl`? Do you have cases where the time-step is limited by diffusive CFL?

tomchor:
> Are you using `diffusive_cfl`? Do you have cases where the time-step is limited by diffusive CFL?

Yes and yes. It's kinda rare that the diffusivity is the limiting factor (and it generally happens in test runs rather in production-ready code) but it happens often enough that I always set-it.

In fact, now that I think about it, I'd advocate that we set `diffusive_cfl=0.9` (ot even 1) by default.

glwagner:
I think having a default but non-trivial CFL would be surprising and therefore undesirable, even if it's convenient for some users.

I was hoping we might be able to deprecate `diffusive_cfl` because it's a big maintenance burden, but let's keep it if its useful.

tomchor:
> I think having a default but non-trivial CFL would be surprising and therefore undesirable, even if it's convenient for some users.
> 
> I was hoping we might be able to deprecate `diffusive_cfl` because it's a big maintenance burden, but let's keep it if its useful.

I wouldn't setting the `diffusive_cfl` is surprising. After all the diffusive cfl is as much of a necessary condition for convergence as the advective cfl, it just is a limiting-factor less often for oceanic simulations. But I'm okay with keeping the current behavior as well.

glwagner:
> > I think having a default but non-trivial CFL would be surprising and therefore undesirable, even if it's convenient for some users.
> > I was hoping we might be able to deprecate `diffusive_cfl` because it's a big maintenance burden, but let's keep it if its useful.
> 
> I wouldn't setting the `diffusive_cfl` is surprising. After all the diffusive cfl is as much of a necessary condition for convergence as the advective cfl, it just is a limiting-factor less often for oceanic simulations. But I'm okay with keeping the current behavior as well.

I certainly don't disagree that it's convenient; the question is more whether the benefit justifies the cost. Users also want a code that's easy to maintain and extend because then they get more features! For constant diffusivities, the CFL constraint can be pre-calculated. For some common closures the calculation is simple, but this is not _generically_ the case for all turbulence closures. It's really only for variable diffusivities that we _need_ a feature for on-the-fly calculation.

We could also use something like `diffusive_cfl=nothing` which completely avoids the calculation. Then we can justify supporting a diffusive CFL calculation for selected turbulence closures. That might be a good route (discussion for another issue).

PR #2212, opened 2022-02-02
Saves actual `solver_method` name in `ImplicitFreeSurface` (not just `:Default`)

navidcy:
This is a workaround to avoid printing `solver: Default` and instead print `solver: FFTImplicitFreeSurfaceSolver`.

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
@@ -59,8 +59,7 @@ function FreeSurface(free_surface::ImplicitFreeSurface{Nothing}, velocities, gri
     barotropic_y_volume_flux = Field{Center, Face, Nothing}(grid)
     barotropic_volume_flux = (u=barotropic_x_volume_flux, v=barotropic_y_volume_flux)
 
-    solver_method = free_surface.solver_method
-    solver = build_implicit_step_solver(Val(solver_method), grid, gravitational_acceleration, free_surface.solver_settings)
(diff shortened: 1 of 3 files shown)


glwagner:
Huh, shouldn't we fill in `solver_method` correctly when we construct `ImplicitFreeSurface`?

navidcy:
> Huh, shouldn't we fill in `solver_method` correctly when we construct `ImplicitFreeSurface`?

Good point!

navidcy:
I think here is where solver is set to `Default`.
https://github.com/CliMA/Oceananigans.jl/blob/931c6d507789c3a86cd4526c1279ecf93c94c063/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl#L46

navidcy:
Can't figure out where 
https://github.com/CliMA/Oceananigans.jl/blob/931c6d507789c3a86cd4526c1279ecf93c94c063/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl#L45-L46
are used.

navidcy:
@glwagner, see changes after f3c959c

simone-silvestri:
Solver `Default` is modified once the model is instantiated

simone-silvestri (approved):

navidcy:
@glwagner, some of the tests fail. I don't see why... Any ideas?

simone-silvestri:
I don't know if it might help but in `src/Solvers/preconditioned_conjugate_gradient_solver.jl` line 232 there is a typo in `show` (`solver.q` should be `solver.grid`). I don't know about the iterate, maybe because you are trying to return two different object from the same function and it is trying to index into them? (I thought that was allowed though)

navidcy:
@simone-silvestri, how does it look now?

PR #2213, opened 2022-02-02
Fixes bug in `TimeStepWizard`'s diffusive timescale calculation

tomchor:
The clause that controls the definition of `cell_diffusion_timescale` in `TimeStepWizard` had a bug. I re-wrote in a way that's clearer to understand and that fixes the bug.

Consider the MWE below:

```julia
using Oceananigans

grid = RectilinearGrid(size=(8, 8, 8), extent = (1,1,1),
                       topology=(Periodic, Periodic, Bounded))

closure = IsotropicDiffusivity(ν=1)
model = NonhydrostaticModel(grid = grid,
                            closure = closure)
@info "" model
                            
wizard = TimeStepWizard(diffusive_cfl=0.1,)
```

Before this PR:

```julia
julia> wizard.cell_diffusion_timescale(model)
Inf

julia> @which wizard.cell_diffusion_timescale(model)
infinite_diffusion_timescale(args...) in Oceananigans.Simulations at /home/tomas/.julia/packages/Oceananigans/IHPoE/src/Simulations/time_step_wizard.jl:12
```

After this PR:

```julia
julia> wizard.cell_diffusion_timescale(model)
0.015625

julia> @which wizard.cell_diffusion_timescale(model)
cell_diffusion_timescale(model) in Oceananigans.TurbulenceClosures at /home/tomas/repos/Oceananigans.jl/src/TurbulenceClosures/turbulence_closure_diagnostics.jl:15
```

I'll also add a test to catch this in the future, since there are no tests for `diffusive_cfl` that I could tell.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
-version = "0.69.2"
+version = "0.69.3"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/Simulations/time_step_wizard.jl
+++ b/src/Simulations/time_step_wizard.jl
@@ -1,3 +1,6 @@
+using Oceananigans: TurbulenceClosures
+
+
 mutable struct TimeStepWizard{FT, C, D}
                          cfl :: FT
                diffusive_cfl :: FT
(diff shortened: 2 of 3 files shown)


glwagner (approved):

glwagner:
I would add a minimal test given this feature could change, and hopefully make sure the test doesn't increase CI cost much.

francispoulin (approved):

tomchor:
> I would add a minimal test given this feature could change, and hopefully make sure the test doesn't increase CI cost much.

Added the most minimal test I could think of and bumped the patch version since this is a bug fix. If everyone's okay with it, I'll merge if the tests pass.

simone-silvestri (approved):

PR #2214, opened 2022-02-02
3D spacing and derivatives

### LINKED ISSUE #2209: Conditional fluxes for Biharmonic operator and Immersed Boundaries
Biharmonic operator requires more halo cells than the classical Laplacian, so the `conditional_flux` method for the immersed boundary must be adapted to shut down fluxes from biharmonic diffusion up to 2 cells distance from an immersed boundary

Or maybe find a more sustainable solution, for example refactoring the biharmonic diffusivity to be calculated `∇²κ∇²` (which might require some work)

simone-silvestri:
This PR is motivated by the need to specify higher order operators on `ImmersedBoundaries` #2209 

Since all operators we use depend on first order derivatives, it is enough to cancel all derivatives inside the immersed domain, but, to do that, all operators have to specify a 3D location

So the changes that this PR introduces are: 
- fully 3D metric functions (i.e., `Δ, A, V` for each 3D combination of `ᶜ` and `ᶠ`). 1D and 2D spacings are still defined (as well as 2D areas for grids that support them) but are not explicitly exported by `Oceananigans.Operators`. This is because maybe we want to use them in the future but they should be used only when _strictly_ necessary. As an example, in a function which is known to be 2D or 1D (as a function exclusive to reduced fields)
- Operators are now _only_ 3D with an exception of interpolations and differences that do not involve metric terms (I would also remove the `ᵃ` on the interpolations because it does not make much sense in my opinion, what do you think?)
- I removed `Ax_u, Ax_v, Ax_ζ` and so on in favour of a general `Ax_q` (q for quantity) as the different character did not add anything to the operator so in my opinion it was a bit redundant and confusing (all information is stored in the direction of the area and the location) Let me know if you prefer something different (like maybe `Ax_mul_q` or `Ax_var`) or if you liked the `Ax_u, Ax_v, Ax_ζ` formulation
- Last point (which motivated all this): derivatives are canceled if either of the points in the stencil are fully immersed

closes #2209
closes #2218


navidcy:
@simone-silvestri I can review. 
But could you add a small description?

tomchor (approved):
Can we bump a minor version before merging since this is a breaking change?
Otherwise looks good! (Provided the tests pass)

src/ImmersedBoundaries/immersed_grid_metrics.jl
```diff
@@ -57,61 +57,24 @@ using Oceananigans.AbstractOperations: GridMetricOperation
...
-
-               :Vᶜᶜᶜ, 
-               :Vᶠᶜᶜ,
-               :Vᶜᶠᶜ,
-               :Vᶜᶜᶠ,
-              )
+# Defining all the metrics for Immersed Boundaries
+
+for LX in (:ᶜ, :ᶠ), LY in (:ᶜ, :ᶠ), LZ in (:ᶜ, :ᶠ)
+    for dir in (:x, :y, :z), operator in (:Δ, :A)
```
glwagner, line 63: It's so beautiful 🥺

glwagner:
This is an amazing refactor.

glwagner (approved):
Just amazing

I think it's hard to test everything now, but hopefully as we build infrastructure and testing for immersed boundaries we'll be able to fully exercise the 3D metrics.

And now we can write grids that might work for deep atmospheres... (though we don't have the physics :-P)

francispoulin (approved):
An incredible change for the better.  Well done @simone-silvestri !

simone-silvestri:
Thanks for the review, I'll merge!

tomchor:
Shouldn't this be a minor version bump instead of a patch one? After all it's a breaking change if anyone is using the old operators...

simone-silvestri:
Oh sorry I thought you meant patch version bump. I ll rectify that

Issue #1462, opened 2021-03-14
Simulation with non-dimensional units


navidcy:
In the [Bickley jet example](https://clima.github.io/OceananigansDocumentation/dev/generated/Bickley_jet_shallow_water/), where we run the simulation we get the message:
```julia
[ Info: Simulation is stopping. Model time 2.500 minutes has hit or exceeded simulation stop time 2.500 minutes.
```
This is because we construct the simulation
```julia
Simulation(model, Δt = 1e-2, stop_time = 150)
```
it's assumed that 150 is 150 seconds. That could be a bit confusing. Is there is a way to tell the `Simulation` constructor that our units are nondimensional? Should we have such option?

ali-ramadhan:
Yes this is indeed confusing. We probably do want a way of specifying the unit system of the simulation, also for NetCDF attributes. We discussed this a bit in #1362.

Seems like using Unitful.jl to define a unit for lengths (`grid.units`?) and a unit for time (`clock.units`?) might be enough?

I guess we might need a unit for mass, and a unit for each tracer as well?

But for this issue, a unit for time should be enough.

glwagner:
Maybe `Simulation` needs a `metadata` property where we can put things like this.

glwagner:
I think if we're going to keep track of units, we should store this in the grid (somehow)...

glwagner:
Just like to discuss this a little more.

I think it'd be nice to have some awareness of units in the code. Some places we assume units, like `prettytime`. Other places we don't mention units, like `@show grid`.

The two important cases are SI units and "no units". Maybe we can just have `grid.units = u"m"` by default, where

```
julia> using Unitful

julia> typeof(u"m")
Unitful.FreeUnits{(m,), 𝐋, nothing
```

is the `Unitful.jl` way of saying "meters".

From the length of grid, we can infer the units for every other quantity. I think we should start unambitious and just have `grid.units` and `clock.units`; if `grid.units === u"m"` then we'll set `clock.units = u"s"`. We'll then use these units in things like `prettytime` and `@show grid`.

In principle, we also need `field.units`; field quantities have units that cannot be derived from `grid`. But for Oceananigans, units are really just sugar for showing and displaying things. So it may not make sense to add units to `field`.

glwagner:
Thinking about it more I think we should have `grid.units = nothing` as default.

[...]

PR #2231, opened 2022-02-07
Adds `longform` option to `prettytime`

glwagner:
Useful if we want to print a lot of time values in an array.


test/test_utils.jl
```diff
@@ -1,3 +1,5 @@
+include("dependencies_for_runtests.jl")
+
```
navidcy, line 2: `runtests.jl` includes that and then `test_utils.jl` re-includes it?
what about function re-definitions? I think in Julia 1.7 we may be getting warnings for that.

navidcy (approved):
Can you put in the first comment an example of what an output with `longform` looks like? Is it just `hrs -> hours`, etc?

test/test_utils.jl
glwagner: Dang

glwagner: The purpose here is to include the test file at the REPL so we can debug a test with `Revise`. Maybe there's a better way. Or, we can remove all of these, and put them in just when we need to do that debugging.

glwagner:
> Can you put in the first comment an example of what an output with `longform` looks like? Is it just `hrs -> hours`, etc?

Yaa. Longform is default, retains current behavior. Shortform could be used for showing an array of times, eg

```julia
julia> t = [0, 10minutes, 3days, 5years]
4-element Vector{Float64}:
      0.0
    600.0
 259200.0
      1.5768e8

julia> prettytime.(t)
4-element Vector{String}:
 "0 seconds"
 "10 minutes"
 "3 days"
 "5 years"

julia> prettytime.(t, false)
4-element Vector{String}:
 "0 s"
 "10 m"
 "3 d"
 "5 yrs"
```

glwagner:
We can also use the tools in `Unitful.jl` for this kind of thing, so maybe we'll transition to that if we resolve #1462 .

PR #2232, opened 2022-02-08
Subtle reformulation of `CATKEVerticalDiffusivity`

glwagner:
This PR subtly reformulates CATKE in an attempt to disentangle some of its free parameters and make them easier to calibrate.


src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/mixing_length.jl
```diff
@@ -136,7 +136,7 @@ end
     return ifelse(N² == 0, zero(FT), N² / (∂z_u² + ∂z_v²))
 end
 
-@inline step(x, c, w) = (1 + tanh((x - c) / w)) / 2
+@inline step(x, c, w) = (1 + tanh(x / w - c)) / 2
```
navidcy, line 139: is this what you wanted?
I guess you want this so that c is non dimensional?

navidcy (approved):

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/mixing_length.jl
glwagner: `x`, `c`, and `w` are all non-dimensional.

However, this allows us to control the shape of `step` in a slightly different. Written this way, `c` sets `step(x=0)`; eg choosing `c=2` means that `step(x=0) = 0.04`. Then changing `w` controls how steeply `step` increases from `~0`. This decreases the number of parameters in the calibration.

PR #2235, opened 2022-02-08
Cute compute! for AbstractOperation

glwagner:
Might be useful at the REPL for interactive stuff cause you can write

```julia
julia> N² = compute!(α * ∂z(T) - β * ∂z(S))
```

for example.

If people like, I'll add docs and a test or two.

I guess the equivalent one-liner right now is

```julia
julia> N² = @compute Field(α * ∂z(T) - β * ∂z(S))
```

and the equivalent two-liner is

```julia
julia> N² = Field(α * ∂z(T) - β * ∂z(S))
julia> compute!(N²)
```

### DIFF
--- a/src/AbstractOperations/computed_field.jl
+++ b/src/AbstractOperations/computed_field.jl
@@ -76,3 +76,6 @@ end
     @inbounds data[i, j, k] = operand[i, j, k]
 end
 
+# It's cute
+compute!(op::AbstractOperand) = compute!(Field(op))
+


glwagner:
Bah, I realized this won't work because we have to call `compute!` on all leaves of an expression tree.

We can support this if we change that interface, eg if we add something `compute_leaf!`. Then we can adapt `compute!` for public use and update the private methods like `compute_leaf!` as needed.

tomchor:
> I guess the equivalent one-liner right now is
> 
> ```julia
> julia> N² = @compute Field(α * ∂z(T) - β * ∂z(S))
> ```

Honestly I think this is already pretty handy and enforces the idea that in order to compute something it needs to be a field. When I was starting with Oceananigans it took me a while to understand that I couldn't put abstract operations into the output writers because the computation would fail. I think something like you're proposing would have added to my confusion.

Also, I only recently became aware that this existed. Maybe we should document the `@compute` macro somewhere in the docs?

glwagner:
> I think something like you're proposing would have added to my confusion.

Isn't the confusion a problem with the output writers API? I think it sounds like a great idea to support `AbstractOperation` output. It's kind of logical. Doing this even allows us to do some clever stuff behind the scenes, like using one underlying array to store the results of multiple computations performed serial (thus saving memory).

glwagner:
I'll close this PR because, in addition to it being unpopular, there's too much work we need to do to support it. Maybe something will come out of #2242 though.

tomchor:
> > I think something like you're proposing would have added to my confusion.
> 
> Isn't the confusion a problem with the output writers API? I think it sounds like a great idea to support `AbstractOperation` output. It's kind of logical. Doing this even allows us to do some clever stuff behind the scenes, like using one underlying array to store the results of multiple computations performed serial (thus saving memory).

The confusion is that I wasn't aware that that a `Field` has `data`, and therefore takes up memory, which allows it to store values. While an `AbstractOperation` is just instructions on how to compute things. This extended to the output writers, but I wouldn't say output writers were the source of confusion.

If we make it so that users don't have to know what a `Field` is and we can use abstract operations everywhere, then my comment is moot.

PR #2238, opened 2022-02-09
Minor change to Ocean mixing and convection example

tomchor:
This is a very minor update just to make the plotting command in the Ocean mixing and convection example a bit more general and self-contained. This makes it easier for people to copy-paste that snippet and have it work out of box (I do that frequently for example).

### DIFF
--- a/examples/ocean_wind_mixing_and_convection.jl
+++ b/examples/ocean_wind_mixing_and_convection.jl
@@ -61,7 +61,7 @@ grid = RectilinearGrid(size = (32, 32, Nz),
 
 # We plot vertical spacing versus depth to inspect the prescribed grid stretching:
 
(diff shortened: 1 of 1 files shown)


francispoulin:
Looks like a good idea but I see that some tests were not successful?

glwagner (approved):
Makes sense!

glwagner:
> Looks like a good idea but I see that some tests were not successful?

Tartarus is on the struggle bus it seems

navidcy:
We need some PlotRecipes for fields...

(In general, not for this PR!)

glwagner:
> We need some PlotRecipes for fields...
> 
> (In general, not for this PR!)

We need that badly...

I think we should take the approach in `ClimaCore.jl` which is to develop an "external" package that lives in `/lib`. Perhaps `Makinanigans.jl` or `Vizinanigans.jl`. Performing new releases is a little more arduous with that setup but its probably worth it.

PR #2243, opened 2022-02-12
Deploy docs with `forcepush = true`

navidcy:
The docs repo is running out of space. In an attempt to remedy that I'm trying here `forcepush = true` for deploying docs.

https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Documenter.deploydocs


glwagner:
> forcepush a boolean that specifies the behavior of the git-deployment. The default (forcepush = false) is to push a new commit, but when forcepush = true the changes will be combined with the previous commit and force pushed, erasing the Git history on the deployment branch.

glwagner (approved):

Issue #2242, opened 2022-02-11
Automagically wrap `AbstractOperation` and `Reduction` in `Field` inside `OutputWriters`?


glwagner:
This came up on #2235.

The idea is to support syntax like

```julia
outputs = (ζ = ∂x(v) - ∂y(u), U = Average(u, dims=1))
output_writer = JLD2OutputWriter(model, outputs, ... )
```

rather than requiring things to be wrapped in `Field` to work.

This wasn't easy to support previously, before we consolidate all the `Field` implementations (we had `ComputedField`, `Field`, `AveragedField`...). But now we can support it and it would reduce boilerplate (eg everything is `Field(this)`, `Field(that)` by quite a bit for the majority of use cases.

I think the downside is that users aren't immediately exposed to `Field` infrastructure. A good "Fields tutorial" in the docs might help with that though.

glwagner:
> Pardon if this question is naive but why is it important that users to exposed to the Field infrastructure?

Because that way they are prepared to do sophisticated post-processing! But I think you're right, maybe it's not a real "downside".

glwagner:
Done

PR #2246, opened 2022-02-14
`Field.indices` for windowed output, reductions, diagnostics, and post-processing

### LINKED ISSUE #1080: Convenience args for FieldSlicer
It'd be nice if `FieldSlicer` accepted arguments that indicate physical positions to slice at, eg

```julia
horizontal_slicer = FieldSlicer(grid, z=-64)
```

This would not interpolate, but would instead just slice field at a `k`-index determined by the kwarg `z=-64`. We can use a convention whereby indices assigned according to cell-centered locations via

```julia
k = searchsortedfirst(znodes(Cell, grid), z)
```

A more satisfying solution to slices will be provided by `InterpolatedField`, but this comes at the cost of some computation (and its not certain when precisely we'll implement it).

### LINKED ISSUE #1647: Support `interior(f, i, j, k)` to create custom `view`s into `parent(f)` with correct indexing
It might be pretty nice to support syntax like `interior(f::AbstractField, i, j, k)` that returns `view(parent(f), i′, j′, k′)`, where `i′, j′, k′` are offset to account for the halo regions.

One reason this is useful is because `view(f.data, i, j, k)` is a _doubly-wrapped_ array (it's a view of an `OffsetArray`, which is wrapped around either `Array` or `CuArray`) and therefore cannot be used in broadcasting, reductions, etc on the GPU.

cc @mukund-gupta

### LINKED ISSUE #2177: Index for Reduced fields
Where is a reduced field located in the `Nothing` direction?

It is natural to locate the free surface at the surface,  (`k = grid.Nz`), but for other fields which are Averages or sums the choice is not that natural

This is important in the context of Immersed boundaries to check if a Reduced Field is immersed or not (and eventally mask it). For the free surface it is enough to check `solid_interface(i, j, grid.Nz, grid)`, for averaged fields it might be necessary to check the whole column `all(solid_interface(i, j, 1:grid.Nz, grid))`

glwagner:
This PR adds a new property `indices` to `Field`.

By default `indices = (:, :, :)`. Using `UnitRange` for a dimension's indices will "window" a field in that direction.

This permits diagnostics to be "windowed" for selected indices; for example a surface field might be computed with `indices = (:, :, grid.Nz:grid.Nz)`.

Another important feature is the extension of `FieldTimeSeries` to "sliced" fields. This means we can save just the surface velocity fields (rather than the 3D velocity fields), and in post-processing compute the surface vorticity.

This PR is still very much WIP, but hopefully will be fully featured by the time it's merged, and we'll convert free surface displacements from reduced fields to windowed fields.

Closes #2177
Closes #1080
Closes #1647


tomchor:
If I understand correctly this would make the `field_slicer` argument for output writers obsolete, no?

glwagner:
> If I understand correctly this would make the `field_slicer` argument for output writers obsolete, no?

I missed this! Yes, that's correct. We _could_ keep it, however I think it's nicer to remove because it's one of the best ways to ensure that the "windowing infrastructure" implemented in this PR is performant and easy to use. In place of `field_slicer`, we will add some convenience arguments that allow output fields to be "constructed" from what's given to `OutputWriter`. This will also allow us to close #2242 (but I haven't implemented that yet here).

src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -32,6 +32,20 @@ mutable struct FieldBoundaryConditions{W, E, S, N, B, T, I}
...
 end
 
+function FieldBoundaryConditions(indices::Tuple, west, east, south, north, bottom, top, immersed)
+    # Turn bcs in windowed dimensions into nothing
+    west, east   = window_boundary_conditions(indices[1], west, east)
+    south, north = window_boundary_conditions(indices[2], south, north)
+    bottom, top  = window_boundary_conditions(indices[3], bottom, top)
+    return FieldBoundaryConditions(west, east, south, north, bottom, top, immersed)
+end
+
```
simone-silvestri, line 42: with something like this we can support also different bc for the same field (i.e., open on half of the domain and bounded in the other half). 

It will be useful whenever we can use FFTs (otherwise we can just put an immersed boundary)

navidcy:
I think we should add shortcuts so `(:, :, grid.Nz)` is treated as `(:, :, grid.Nz:grid.Nz)`.

glwagner:
> I think we should add shortcuts so `(:, :, grid.Nz)` is treated as `(:, :, grid.Nz:grid.Nz)`.

It's there!

https://github.com/CliMA/Oceananigans.jl/blob/36f4f037dbb8e1fc226593376680f68f4b9ad653/src/Fields/field.jl#L93

navidcy:
bump version?

glwagner:
> bump version?

right before merger

navidcy (approved):

PR #2247, opened 2022-02-14
Diffusivity refactor

### LINKED ISSUE #2113: Replace `AnisotropicDiffusivity` and `AnisotropicBiharmonicDiffusivity` with their curvilinear-friendly counterparts
It'd be simpler to maintain just one 2nd order and 4th order diffusivity implementation. Also, `HorizontallyCurvilinearAnisotropicBiharmonicDiffusivity` is _truly_ biharmonic, whereas `AnisotropicBiharmonicDiffusivity` is not.

Deletion of the existing schemes and renaming of the curvilinear-friendly schemes will easily resolve this issue.

### LINKED ISSUE #2239: `TurbulenceClosures` consolidation of diffusivities
@simone-silvestri brainstormed today about ways to clean up `TurbulenceClosures`.

One idea that came up is to combine `IsotropicDiffusivity` and `HorizontallyCurvilinearAnisotropicDiffusivity` into a single `ScalarDiffusivity` (where isotropy in various directions is _implied_). We'll also simplify the "anisotropic" diffusivity in this process. The user interface could look something like

```julia
three_dimensional_diffusivity = ScalarDiffusivity(ThreeDimensional(), ν=1)
horizontal_diffusivity = ScalarDiffusivity(Horizontal(), ν=1)
vertical_diffusivity = ScalarDiffusivity(Vertical(), ν=1)
```

so, no more `νh` and `νz`.

The advantage is that we eliminate / share much more code. Arguably the API is a bit cleaner as well. (We can discuss aliases and defaults like `HorizontalScalarDiffusivity`, etc.)

Under the hood we'll add the `Directionality` type to the type parameters of `ScalarDiffusivity`:

```julia
struct ScalarDiffusivity{TimeDisc, Nu, Kappa, Directionality} <: AbstractScalarDiffusivity{TimeDisc, Dir}
    ν :: Nu
    κ :: Kappa

    function IsotropicDiffusivity{TimeDisc}(dir::Dir, ν::Nu, κ::Ka) where {TimeDisc, Dir, Nu, Ka}
        return new{TimeDisc, Nu, Ka, Dir}(ν, κ)
    end
end
```

Then `abstract_isotropic_diffusivity.jl` becomes `abstract_scalar_diffusivity.jl`, and will implement the necessary abstractions for 3D isotropic scalar diffusivities, horizontally-isotropic diffusivities (a la MITgcm), and vertical diffusivities. 

Consolidating code is an important precursor to a larger refactor that might

1) support `AbstractOperation` viscosities and diffusivities at all locations on the grid
2) eliminate `diffusivity_fields` and
3) perhaps also involve fixing `TwoDimLeith` and implementing a biharmonic Leith.

simone-silvestri:
Refractors the `IsotropicDiffusivity`, `AnisotropicDiffusivity` and `HorizontallyCurvilinearAnisotropicDiffusivity` so that only one `ScalarDiffusivity` is defined.

with this PR, the following API changes are implemented
- `closure = IsotropicDiffusivity(ν = 1, κ = 1) -> closure = ScalarDiffusivity(ν = 1, κ = 1, direction=ThreeDimensional()) (default)`
- `closure = HorizontallyCurvilinearAnistropicDiffusivity(νh = 2, κh = 2, νh = 1, κh = 1) -> 
horizontal_closure = ScalarDiffusivity(ν = 2, κ = 2, direction=Horizontal()); 
vertical_closure = ScalarDiffusivity(ν = 1, κ = 1, direction=Vertical())`

For biharmonic diffusivity
- Bug fixes in `HorizontallyCurvilinearAnisotropicBiharmonicDiffusivity` which is kept as the standard method for biharmonic diffusivity and renamed `ScalarBiharmonicDiffusivity` so now
`closure =HorizontallyCurvilinearAnisotropicBiharmonicDiffusivity(ν = 1, κ = 1) -> closure = ScalarBiharmonicDiffusivity(ν = 1, κ = 1, direction=Horizontal()) (default)`
- Also for biharmonic we have a `ThreeDimensional`, `Horizontal` and `Vertical` mode
- Nuking of `AnisotropicBiharmonicDiffusivity`

Additionally
- Implementation of a `discrete` mode for viscosity and diffusivity functions (with `discrete_diffusivity = true`) to be able to have viscosity and diffusivity depend on grid metrics so that we can have
 ```
using Oceananigans.Operators: Δx

νh(lx, ly, lz, i, j, k, grid, clock) =  Δx(i, j, k, grid, lx, ly, lz)^4  / (3600*24)
closure = ScalarDiffusivity(ν = νh, discrete_diffusivity = true)
```
- Implementation of `solid_node` and `solid_interface` for Bounded grids. Used in `Biharmonic` operators to enforce zero flux and zero laplacian boundary conditions

closes #2239
closes #2113

### CODE REVIEW

**Comment by francispoulin** [comment]
Great idea!

**Comment by glwagner** [comment]
Can we use types rather than symbols? `ThreeDimensional()`, `Horizontal()`, etc.

**Comment by simone-silvestri** [comment]
> Can we use types rather than symbols? `ThreeDimensional()`, `Horizontal()`, etc.

They are converted into types in the constructor `return ScalarDiffusivity{TD, eval(direction)}(FT(ν), κ)` (and are effectively types in the type signature. I personally prefer it this way because it allows to build closures in the API avoiding to have to import additional types...

But, I don't feel so strongly about it... I can also change it

File: src/Operators/derivative_operators.jl [hunk 1]
```diff
@@ -55,19 +55,19 @@ for dir in (:x, :y, :z), L1 in (:ᶜ, :ᶠ), L2 in (:ᶜ, :ᶠ)
...
         $der_face_3(i, j, k, grid, f::Function, args...) = $der_face_1(i, j, k, grid, $der_center_2, f::Function, args...)
-        $der_face_4(i, j, k, grid, f::Function, args...) = $der_face_2(i, j, k, grid, $der_center_2, f::Function, args...)
+        $der_face_4(i, j, k, grid, f::Function, args...) = $der_face_2(i, j, k, grid, $der_face_2,   f::Function, args...)
 
-        $der_center_2(i, j, k, grid, f::Function, args...) = $der_center_1(i, j, k, grid, $der_face_1, f::Function, args...)
-        $der_center_3(i, j, k, grid, f::Function, args...) = $der_center_1(i, j, k, grid, $der_face_2, f::Function, args...)
-        $der_center_4(i, j, k, grid, f::Function, args...) = $der_center_2(i, j, k, grid, $der_face_2, f::Function, args...)
+        $der_center_2(i, j, k, grid, f::Function, args...) = $der_center_1(i, j, k, grid, $der_face_1,   f::Function, args...)
+        $der_center_3(i, j, k, grid, f::Function, args...) = $der_center_1(i, j, k, grid, $der_face_2,   f::Function, args...)
+        $der_center_4(i, j, k, grid, f::Function, args...) = $der_center_2(i, j, k, grid, $der_center_2, f::Function, args...)
```
glwagner, line 70:
```suggestion
        $fourth_derivative_center(i, j, k, grid, f::Function, args...) = $der_center_2(i, j, k, grid, $der_center_2, f::Function, args...)
```

glwagner: Otherwise it's really not possible to understand these lines in a diff...

src/TurbulenceClosures/turbulence_closure_implementations/scalar_biharmonic_diffusivity.jl
```diff
@@ -0,0 +1,56 @@
...
+
+  - `νz`: Vertical viscosity. `Number`, `AbstractArray`, or `Function(x, y, z, t)`.
+
+  - `κh`: Horizontal diffusivity. `Number`, `AbstractArray`, or `Function(x, y, z, t)`, or
+          `NamedTuple` of diffusivities with entries for each tracer.
+
+  - `κz`: Vertical diffusivity. `Number`, `AbstractArray`, or `Function(x, y, z, t)`, or
+          `NamedTuple` of diffusivities with entries for each tracer.
+"""
+function ScalarBiharmonicDiffusivity(FT=Float64; ν=0, κ=0, direction = :Horizontal)
```
glwagner, line 37:
```suggestion
function ScalarBiharmonicDiffusivity(FT=Float64; ν=0, κ=0, direction = Horizontal())
```

src/TurbulenceClosures/turbulence_closure_implementations/scalar_biharmonic_diffusivity.jl
```diff
@@ -0,0 +1,56 @@
...
+    ν = convert_diffusivity(FT, ν)
+    κ = convert_diffusivity(FT, κ)
+    return ScalarBiharmonicDiffusivity{eval(direction)}(FT(ν), κ)
```
glwagner, line 40:
```suggestion
    return ScalarBiharmonicDiffusivity{Dir}(FT(ν), κ)
```

src/TurbulenceClosures/turbulence_closure_implementations/scalar_diffusivity.jl
```diff
@@ -0,0 +1,55 @@
...
+Returns parameters for an isotropic diffusivity model with viscosity `ν`
+and thermal diffusivities `κ` for each tracer field in `tracers`
+`ν` and the fields of `κ` may be constants, arrays, fields, or
+functions of `(x, y, z, t)`.
+
+`κ` may be a `NamedTuple` with fields corresponding to each tracer, or a
+single number to be a applied to all tracers.
+"""
+function ScalarDiffusivity(FT=Float64;
+                              ν=0, κ=0, direction=:ThreeDimensional, time_discretization::TD = ExplicitTimeDiscretization()) where {TD, Dir}
```
glwagner, line 25:
```suggestion
                              ν=0, κ=0, direction=ThreeDimensional(), time_discretization::TD = ExplicitTimeDiscretization()) where {TD, Dir}
```

also given this "abbreviated" syntax, we may want to change `ExplicitTimeDiscretization` to just `Explicit`.

src/TurbulenceClosures/turbulence_closure_implementations/scalar_diffusivity.jl
```diff
@@ -0,0 +1,55 @@
...
+
+    if direction == :Horizontal && time_discretization == VerticallyImplicitTimeDiscretization()
+        throw(ArgumentError("The Implicit discretization is supported only for vertical directions"))
```
glwagner, line 28:
```suggestion
        throw(ArgumentError("VerticallyImplicitTimeDiscretization is only supported for diffusivities in `Horizontal` or `ThreeDimensional` directions"))
```

glwagner:
For consistency I think we should also either use `HorizontalDirection()`, `ThreeDimensionalDirection()` -- OR -- we should use `time_discretization = Explicit()`, etc. I think right now I favor the latter since it's perfectly meaningful but more concise.

glwagner:
> > Can we use types rather than symbols? `ThreeDimensional()`, `Horizontal()`, etc.
> 
> They are converted into types in the constructor `return ScalarDiffusivity{TD, eval(direction)}(FT(ν), κ)` (and are effectively types in the type signature. I personally prefer it this way because it allows to build closures in the API avoiding to have to import additional types...
> 
> But, I don't feel so strongly about it... I can also change it

I think the most important thing is a uniform API that's easy to remember. For a few exceptions (that we will hopefully address eventually) we use types throughout. I think if we have an API that's half types, half symbols it will get hard to remember when we use what...

src/TurbulenceClosures/turbulence_closure_implementations/scalar_diffusivity.jl
glwagner: (which we can do later)

src/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -68,18 +66,16 @@ include("turbulence_closure_utils.jl")
...
 include("closure_tuples.jl")
 include("isopycnal_rotation_tensor_components.jl")
 
 # Implementations:
 include("turbulence_closure_implementations/nothing_closure.jl")
-include("turbulence_closure_implementations/isotropic_diffusivity.jl")
-include("turbulence_closure_implementations/anisotropic_diffusivity.jl")
-include("turbulence_closure_implementations/horizontally_curvilinear_anisotropic_diffusivity.jl")
-include("turbulence_closure_implementations/horizontally_curvilinear_anisotropic_biharmonic_diffusivity.jl")
-include("turbulence_closure_implementations/anisotropic_biharmonic_diffusivity.jl")
```
glwagner, line 82: 💪

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -0,0 +1,147 @@
...
+    abstract type AbstractScalarDiffusivity <: AbstractTurbulenceClosure end
+
+Abstract type for closures with *isotropic* diffusivities.
+"""
+abstract type AbstractScalarDiffusivity{TD, Dir} <: AbstractTurbulenceClosure{TD} end
+
+"""
+    struct ThreeDimensional end
+
+Corresponds to a closure which is three-dimensionally isotropic
```
glwagner, line 11:
```suggestion
Specifies a three-dimensionally isotropic `ScalarDiffusivity`.
```

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -0,0 +1,147 @@
...
+@inline viscous_flux_vy(i, j, k, grid, closure::AID, clock, U, args...) = - 2 * ν_σᶜᶜᶜ(i, j, k, grid, clock, viscosity(closure, args...), Σ₂₂, U.u, U.v, U.w)
+@inline viscous_flux_wy(i, j, k, grid, closure::AID, clock, U, args...) = - 2 * ν_σᶜᶠᶠ(i, j, k, grid, clock, viscosity(closure, args...), Σ₃₂, U.u, U.v, U.w)
+
+@inline viscous_flux_ux(i, j, k, grid, closure::AHD, clock, U, args...) = - ν_δᶜᶜᶜ(i, j, k, grid, clock, closure.ν, U.u, U.v)   
+@inline viscous_flux_vx(i, j, k, grid, closure::AHD, clock, U, args...) = - ν_ζᶠᶠᶜ(i, j, k, grid, clock, closure.ν, U.u, U.v)
+@inline viscous_flux_uy(i, j, k, grid, closure::AHD, clock, U, args...) = + ν_ζᶠᶠᶜ(i, j, k, grid, clock, closure.ν, U.u, U.v)   
+@inline viscous_flux_vy(i, j, k, grid, closure::AHD, clock, U, args...) = - ν_δᶜᶜᶜ(i, j, k, grid, clock, closure.ν, U.u, U.v)
+# are we sure about this?? This might not be consistent for AHD
+@inline viscous_flux_wx(i, j, k, grid, closure::AHD, clock, U, args...) = - ν_σᶠᶜᶠ(i, j, k, grid, clock, viscosity(closure, args...), ∂xᶠᶜᶠ, U.w)
+@inline viscous_flux_wy(i, j, k, grid, closure::AHD, clock, U, args...) = - ν_σᶜᶠᶠ(i, j, k, grid, clock, viscosity(closure, args...), ∂yᶜᶠᶠ, U.w)
```
glwagner, line 64: I guess this strikes the heart of what we mean by `Horizontal`...

What's the stress tensor for a "Horizontal" scalar diffusivity?

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -0,0 +1,147 @@
...
+    viscous_flux_u = Symbol(:viscous_flux_u, dir)
+    viscous_flux_v = Symbol(:viscous_flux_v, dir)
+    viscous_flux_w = Symbol(:viscous_flux_w, dir)
+    @eval begin
+        @inline $diffusive_flux(i, j, k, grid, closure::$closure, c, c_idx, clock, args...) = zero(eltype(grid))
+        @inline $viscous_flux_u(i, j, k, grid, closure::$closure, c, clock, U, args...)     = zero(eltype(grid))
+        @inline $viscous_flux_v(i, j, k, grid, closure::$closure, c, clock, U, args...)     = zero(eltype(grid))
+        @inline $viscous_flux_w(i, j, k, grid, closure::$closure, c, clock, U, args...)     = zero(eltype(grid))
+    end
+end
```
glwagner, line 97: 👏

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -0,0 +1,147 @@
...
+    struct ThreeDimensional end
+
+Corresponds to a closure which is three-dimensionally isotropic
+"""
+struct ThreeDimensional end
+
+"""
+    struct Horizontal end
+
+Corresponds to a closure which is horizontally isotropic and calculated in a VectorInvariant form
```
glwagner, line 18:
```suggestion
Specifies a horizontally-isotropic, `VectorInvariant`, `ScalarDiffusivity`.
```

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -0,0 +1,147 @@
...
+"""
+struct Horizontal end
+
+"""
+    struct Vertical end
+
+Corresponds to a closure which calculates only the vertical direction
```
glwagner, line 25:
```suggestion
Specifies a `ScalarDiffusivity` acting only in the vertical direction.
```

src/TurbulenceClosures/turbulence_closure_diagnostics.jl
```diff
@@ -23,102 +31,29 @@ maximum_numeric_diffusivity(::Nothing) = 0
...
+    Δ = min_Δxyz(grid, Dir)
+    max_κ = maximum_numeric_diffusivity(closure.κ)
+    max_ν = maximum_numeric_diffusivity(closure.ν)
+    return min(Δ^4/ max_ν, Δ^4 / max_κ)
 end
 
 function cell_diffusion_timescale(closure::SmagorinskyLilly{FT, TD, P, <:NamedTuple{()}},
                                   diffusivities, grid) where {FT, TD, P}
-    Δ = min_Δxyz(grid)
+    Δ = min_Δxyz(grid, ThreeDimensional)
```
glwagner, line 50:
```suggestion
    Δ = min_Δxyz(grid, direction(closure))
```

elsewhere we need

```julia
direction(::AbstractScalarDiffusivity{TD, Dir}) where {TD, Dir} = Dir()
```

src/TurbulenceClosures/turbulence_closure_diagnostics.jl
```diff
@@ -23,102 +31,29 @@ maximum_numeric_diffusivity(::Nothing) = 0
...
     max_ν = maximum(diffusivities.νₑ.data.parent)
     return Δ^2 / max_ν
 end
 
 function cell_diffusion_timescale(closure::SmagorinskyLilly, diffusivities, grid)
-    Δ = min_Δxyz(grid)
+    Δ = min_Δxyz(grid, ThreeDimensional)
```
glwagner, line 56:
```suggestion
    Δ = min_Δxyz(grid, direction(closure))
```

src/TurbulenceClosures/turbulence_closure_diagnostics.jl
```diff
@@ -127,20 +62,20 @@ end
 
 function cell_diffusion_timescale(closure::AnisotropicMinimumDissipation{FT, TD, PK, PN, <:NamedTuple{()}},
                                   diffusivities, grid) where {FT, TD, PK, PN}
-    Δ = min_Δxyz(grid)
+    Δ = min_Δxyz(grid, ThreeDimensional)
```
glwagner, line 65:
```suggestion
    Δ = min_Δxyz(grid, direction(closure))
```

src/TurbulenceClosures/turbulence_closure_diagnostics.jl
```diff
@@ -23,102 +31,29 @@ maximum_numeric_diffusivity(::Nothing) = 0
 # As the name suggests, we give up in the case of a function diffusivity
 maximum_numeric_diffusivity(κ::Function) = 0
 
-function cell_diffusion_timescale(closure::IsotropicDiffusivity, diffusivities, grid)
-    Δ = min_Δxyz(grid)
+function cell_diffusion_timescale(closure::ScalarDiffusivity{TD, Dir}, diffusivities, grid) where {TD, Dir}
+    Δ = min_Δxyz(grid, Dir)
```
glwagner, line 35:
```suggestion
    Δ = min_Δxyz(grid, direction(closure))
```

This will make it easy to change the position of the type parameter later if we need to.

glwagner: ("Only use type parameters for dispatch" - YASGuide)

src/TurbulenceClosures/turbulence_closure_diagnostics.jl
```diff
@@ -4,13 +4,21 @@
 
 using Oceananigans.Grids: topology, min_Δx, min_Δy, min_Δz
 
-function min_Δxyz(grid)
+function min_Δxyz(grid, ::Type{ThreeDimensional})
```
glwagner, line 7:
```suggestion
function min_Δxyz(grid, ::ThreeDimensional)
```

This will be possible if we have `direction(closure)` function. (Or maybe `directionality(closure)` or `isotropy(closure)`...)

src/TurbulenceClosures/turbulence_closure_implementations/scalar_diffusivity.jl
glwagner: also should we call this `direction` or `isotropy`?

glwagner: (or something else...)

glwagner (approved):
So good!

I think there's a few other diffusivities that can subtype like `CATKEVerticalDiffusivity <: AbstractScalarDiffusivity{TD, Vertical}` and also `ConvectiveAdjustmentDiffusivity`. But we can fix those after this PR.

Needs a HUGE version bump. Ok maybe just to 0.71.

glwagner:
> * Implementation of zero laplacians on the immersed boundaries to allow a biharmonic closure with immersed boundaries

Do we also need to zero the Laplacian on _non-immersed_ (ie ordinary) boundaries?

I think we also need to throw an error if the _only_ closure if `ScalarBiharmonicDiffusivity` and the boundary condition is anything _but_ `FluxBoundaryCondition`. It's kind of complex because it's still ok to have biharmonic and non-flux, provided that there's _another_ closure (in a closure tuple) present...

src/TurbulenceClosures/turbulence_closure_implementations/scalar_diffusivity.jl
simone-silvestri: I think isotropy is a good name, then the input to the constructor will be `isotropy::Iso=Threedimensional()`

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
simone-silvestri: To be consistent and symmetric (as vertical derivatives of horizontal velocities are not accounted for in the vertical stress of the horizontal momentum) we neglect the vertical shear in the horizontal stress of vertical momentum. 

This formulation might not conserve angular momentum, but if the grid is highly anisotropic it is not a problem (it would be a problem to include the additional terms)

simone-silvestri:
I think the flux boundary condition implementation is correct (we ensure zero laplacian by mirroring our field). The problem is in the `Open` boundary conditions, we have to extend those to ensure 0 (or whatever value is correct) laplacian on the boundary

PR #2005, opened 2021-10-11
Update CUDA

navidcy:
This PR uses the latest CUDA version which includes the fix to https://github.com/JuliaGPU/CUDA.jl/issues/1169 (see https://github.com/JuliaGPU/CUDA.jl/pull/1174)

PR #2035, opened 2021-11-02
Halo optimization and solving the race condition on corner nodes

simone-silvestri:
This PR is created to solve the boundary condition race condition that occurs when trying to fill in the halo corner nodes.

It is basically just an integration of PR #1985 that runs the halo filling sequentially to avoid unsynchronized execution and PR #1923 to fill left and right halos together which should increase performance.

IMPORTANT NOTE:

This PR will fail the ocean large eddy simulation regression tests when lines (50-53) in BoundaryConditions/fill_halo_regions.jl are uncommented.

The only thing that these lines do is a reordering of halo filling order such that the periodic boundary conditions are calculated after all the "Bounded type" boundary conditions.

This is needed since periodic boundary conditions require previous evaluation of boundary conditions in other directions to fill the corner nodes (see PR #1985).

It is possible that the ocean-large-eddy-simulation-regression-data has been generated with periodic boundary conditions (west-east, north-south) which are evaluated before the bounded ones (top-bottom) (see the fill_halo_regions! function in main). As such the issue with the test failing when uncommenting lines (50-53) would not be a bug in the code but a bug in the previous generated regression data. This error is quite small (as the corners would be updated based on the previous step "bounded type" boundary conditions) but still enough to make the test fail. (I leave this to you to confirm :))

Resolves #1179

PR #2253, opened 2022-02-18
Multi-region module for MPI-less multi-GPU computation 

### LINKED ISSUE #1582: `Connected` topologies
For distributed parallelism and cubed spheres it would be useful to introduce the idea of `Connected` topologies: dimensions along which the model is connected to another domain or another model.

They behave very much like `Periodic` dimensions in that you only need to store the first N faces along the dimension, which face N+1 being shared with the domain/model it's connected to. Contrast this with `Bounded` dimensions in which all N+1 faces are stored.

The boundary conditions associated with a `Connected` topology will be some sort of communication, e.g. `HaloCommunication` or `CubedSphereExchange` right now.

For MPI we will also need `LeftConnected` and `RightConnected` for wall-bounded cases where you have a wall on your left and are communicating with another rank on your right (or vice versa).

PR #1522 just adds `const Connected = Periodic` for now since it behaves exactly like `Periodic` for now except with different boundary conditions. We may want to dedicate a separate PR to fleshing out this idea of `Connected` topologies.

simone-silvestri:
todo

- [x] Implement the preconditioner for free surface Implicit solver
- [x] Allow for flux boundary conditions with `Array` type parameters
- [ ] Implement multi-region output writers (and eventually readers?)
- [x] Adapt `AbstractOperations` to `MultiRegionFields`
- [x] Add more tests for implicit solvers
- [x] Benchmark results

Closes #1582 
Closes #2255 
closes #2005


glwagner:
it's WORKING

francispoulin:
I am trying to follow the code, and sorry that it's not obvious to me what's goipartition = XPartition([3, 3, 3, 3])ng on.

Because you have `devices=(0,1)`, I presume that means you are running on two GPUs?

But you also have `partition = XPartition([3, 3, 3, 3])`, which seems like you are partitioning it onto 4 regions, each of length 3, which sum to 12 (the length of the domain).  

If you have two devices, wouldn't you normally have two partitions and not four?

glwagner:
> I am trying to follow the code, and sorry that it's not obvious to me what's goipartition = XPartition([3, 3, 3, 3])ng on.
> 
> Because you have `devices=(0,1)`, I presume that means you are running on two GPUs?
> 
> But you also have `partition = XPartition([3, 3, 3, 3])`, which seems like you are partitioning it onto 4 regions, each of length 3, which sum to 12 (the length of the domain).
> 
> If you have two devices, wouldn't you normally have two partitions and not four?

We want to allow general partitioning across devices. For example the cubed sphere is a 6 region domain, but we won't always have 6 devices.

The question to ask is what our "usual" case is, but rather what the most general case that we want to support looks like.

simone-silvestri:
I tried to keep a certain generality, such that partitions and devices are disconnected. This will play well the cubed sphere implementation.

example:
```
julia> mrg = MultiRegionGrid(grid, partition = XPartition(3), devices = (0, 1, 2))
MultiRegionGrid partitioned on GPU(): 
├── grids: 4×1×1 RectilinearGrid{Float64, Periodic, Periodic, Flat} on GPU with 1×1×0 halo 
├── partitioning: Equal partitioning in X (3 regions) 
└── architecture: Oceananigans.Architectures.MultiGPU() (3 devices)

julia> mrg.devices
(CUDA.CuDevice(0), CUDA.CuDevice(1), CUDA.CuDevice(2))

julia> mrg = MultiRegionGrid(grid, partition = XPartition(3), devices = (2, 1, 1))
MultiRegionGrid partitioned on GPU(): 
├── grids: 4×1×1 RectilinearGrid{Float64, Periodic, Periodic, Flat} on GPU with 1×1×0 halo 
├── partitioning: Equal partitioning in X (3 regions) 
└── architecture: Oceananigans.Architectures.MultiGPU() (2 devices)

julia> mrg.devices
(CUDA.CuDevice(2), CUDA.CuDevice(1), CUDA.CuDevice(1))

julia> mrg = MultiRegionGrid(grid, partition = XPartition(3), devices = nothing)
MultiRegionGrid partitioned on CPU(): 
├── grids: 4×1×1 RectilinearGrid{Float64, Periodic, Periodic, Flat} on CPU with 1×1×0 halo 
├── partitioning: Equal partitioning in X (3 regions) 
└── architecture: CPU() 

julia> mrg.devices
(CPU(), CPU(), CPU())

julia> mrg = MultiRegionGrid(grid, partition = XPartition(3), devices = (0, 3))
MultiRegionGrid partitioned on GPU(): 
├── grids: 4×1×1 RectilinearGrid{Float64, Periodic, Periodic, Flat} on GPU with 1×1×0 halo 
├── partitioning: Equal partitioning in X (3 regions) 
└── architecture: Oceananigans.Architectures.MultiGPU() (2 devices)

julia> mrg.devices
(CUDA.CuDevice(0), CUDA.CuDevice(0), CUDA.CuDevice(3))

julia> 

```

glwagner:
> I tried to keep a certain generality, such that partitions and devices are disconnected. This will play well the cubed sphere implementation.
> 
> In the end each partition will have a device associated with it specified in `mrg.devices` How you assign these devices is up to you
> 
> example:
> 
> ```
> julia> mrg = MultiRegionGrid(grid, partition = XPartition(3), devices = (0, 1, 2))
> MultiRegionGrid partitioned on GPU(): 
> ├── grids: 4×1×1 RectilinearGrid{Float64, Periodic, Periodic, Flat} on GPU with 1×1×0 halo 
> ├── partitioning: Equal partitioning in X (3 regions) 
> └── architecture: Oceananigans.Architectures.MultiGPU() (3 devices)
> 
> julia> mrg.devices
> (CUDA.CuDevice(0), CUDA.CuDevice(1), CUDA.CuDevice(2))
> 
> julia> mrg = MultiRegionGrid(grid, partition = XPartition(3), devices = (2, 1, 1))
> MultiRegionGrid partitioned on GPU(): 
> ├── grids: 4×1×1 RectilinearGrid{Float64, Periodic, Periodic, Flat} on GPU with 1×1×0 halo 
> ├── partitioning: Equal partitioning in X (3 regions) 
> └── architecture: Oceananigans.Architectures.MultiGPU() (2 devices)
> 
> julia> mrg.devices
> (CUDA.CuDevice(2), CUDA.CuDevice(1), CUDA.CuDevice(1))
> 
> julia> mrg = MultiRegionGrid(grid, partition = XPartition(3), devices = nothing)
> MultiRegionGrid partitioned on CPU(): 
> ├── grids: 4×1×1 RectilinearGrid{Float64, Periodic, Periodic, Flat} on CPU with 1×1×0 halo 
> ├── partitioning: Equal partitioning in X (3 regions) 
> └── architecture: CPU() 
> 
> julia> mrg.devices
> (CPU(), CPU(), CPU())
> 
> julia> mrg = MultiRegionGrid(grid, partition = XPartition(3), devices = (0, 3))
> MultiRegionGrid partitioned on GPU(): 
> ├── grids: 4×1×1 RectilinearGrid{Float64, Periodic, Periodic, Flat} on GPU with 1×1×0 halo 
> ├── partitioning: Equal partitioning in X (3 regions) 
> └── architecture: Oceananigans.Architectures.MultiGPU() (2 devices)
> 
> julia> mrg.devices
> (CUDA.CuDevice(0), CUDA.CuDevice(0), CUDA.CuDevice(3))
> ```
> 
> In the above example
> 
> ```
> julia> mrg.devices
> (CUDA.CuDevice(0), CUDA.CuDevice(0), CUDA.CuDevice(1), CUDA.CuDevice(1))
> ```

Does it make sense to add a `device` to `Architectures.GPU`? Eg

```julia
struct GPU{D}
    device :: D
end
```

?

This could also prepare `GPU` for supporting AMD since then we have `GPU{<:CuDevice}`, etc.

tomchor:
This seems great! Although I gotta say, I don't see how that code achieves communication between GPUs. I'm guessing since this doesn't use MPI, we'll only be able to use the GPUs in one node and a single core of that node, correct?

Do you have any resources that you can point me to so I can read about this?

glwagner:
> This seems great! Although I gotta say, I don't see how that code achieves communication between GPUs. I'm guessing since this doesn't use MPI, we'll only be able to use the GPUs in one node and a single core of that node, correct?
> 
> Do you have any resources that you can point me to so I can read about this?

@tomchor this blog post advertises some of the CUDA features that make a single-node, multi-GPU implementation possible:

https://juliagpu.org/post/2022-01-28-cuda_3.5_3.8/

src/MultiRegion/multi_region_field_bcs.jl
```diff
@@ -25,46 +43,96 @@ const CBC  = BoundaryCondition{<:Connected}
...
+fill_halo_regions!(f::MultiRegionField, args...; kwargs...) = fill_halo_regions!(f.data, f.boundary_conditions, architecture(f), f.grid, args...; kwargs...)
+
+function fill_halo_regions!(f::MultiRegionObject, bcs, arch, mrg::MultiRegionGrid, args...; kwargs...)
+    
+    # Apply top and bottom boundary conditions as usual
+    z_event  = apply_regionally(fill_bottom_and_top_halo!, f, bcs.bottom, bcs.top, arch, device_event(arch), mrg, args...; kwargs...) 
+    
+    # Find neighbour and pass it to the fill_halo functions
+    x_neighb = apply_regionally(find_neighbours, bcs.west,  bcs.east, f.regions)
+    y_neighb = apply_regionally(find_neighbours, bcs.south, bcs.north, f.regions)
```
glwagner, line 72:
```suggestion
    x_neighb = apply_regionally(find_neighbors, bcs.west,  bcs.east, f.regions)
    y_neighb = apply_regionally(find_neighbors, bcs.south, bcs.north, f.regions)
```

glwagner:
@simone-silvestri discussed the pros and cons of implementing this "single process version" of `MultiRegion`.

### Cons

The main con is the difficulty in maintaining code for `MultiRegionGrid` and `MultiRegionObject`, plus `apply_regionally!`. The difficulties we face here are identical to the issues that have plagued our maintenance of `CubedSphereGrid`: the need to redispatch and `apply_regionally!` a huge number of methods on `MultiRegionGrid` creates extensive cross dependencies between "local" implementations and their multi-region versions. This is unbelievably annoying to maintain when refactoring or adding features to existing local implementations.

We can mitigate the above issue with (i) clever and careful code design and (ii) potentially a bit of metaprogramming, similar to how `@kernel` and `adapt(to, obj)` achieve CPU/GPU polymorphism.

### Pros

We identified 3 main pros:

1. We can run problems distributed across many GPUs, or cubed sphere problems on GPU or CPU, on a single process without an MPI dependency. This is a huge win for user experience, interactivity, and ease-of-use (users don't need to invoke `mpiexec`).

2. Related to 1, we can debug complex multiregion applications (like the cubed sphere) without MPI. This is important, given the complexity of the cubed sphere (and extensions / generalizations of the quasi-unstructured cubed sphere algorithm).

3. CUDA provides functionality that works on multi-GPU nodes: CUDA FFT, matrix algebra, etc. Being able to launch these utilities from a single Julia process may greatly simplify the implementation of GPU-distributed elliptic solvers.

francispoulin:
Briefly, it makes it friendlier for the user but not so friendly for the maintainer.  

If it's easier for the user then it seems like a great idea, but it would be nice that it is not so difficult to maintain, which I gather is the case here.

glwagner:
Note that we may need to resolve https://github.com/JuliaGPU/KernelAbstractions.jl/issues/289 to merge this.

simone-silvestri:
Update:

the module is almost functional for `ExplicitFreeSurface` `Periodic` models in x. 
There is still to address the `Face` fields in a `Bounded` direction that i am going to deal with by introducing a `Connected{LeftTopology, RightTopology}` topology. 

The code is not at its maximum efficiency because we would have to restructure the `fill_halo_regions` function such that if we pass a tuple of fields they are done all at the same time (i.e., `fill_halo_west_east`, `north_south` and `top_bottom` for all the fields instead of doing them sequentially per field)

This said on a fairly large grid (`512²×30`) I get an efficiency of 78% going from 1 to 2 GPUs

glwagner:
> The code is not at its maximum efficiency because we would have to restructure the `fill_halo_regions` function such that if we pass a tuple of fields they are done all at the same time (i.e., `fill_halo_west_east`, `north_south` and `top_bottom` for all the fields instead of doing them sequentially per field)

perhaps we pursue this after merge? It doesn't seem all that difficult.

We also need this "tupled kernel" pattern for diagnostics. Perhaps we can write a little utility that unrolls loops over tuples. ?

simone-silvestri:
We could, but I think it is quite crucial to the success of this module. I think there is still a long way to go before we can merge this so I would rather try out the options we have and see the results before merging than having to restructure code after

glwagner:
> We could, but I think it is quite crucial to the success of this module. I think there is still a long way to go before we can merge this so I would rather try out the options we have and see the results before merging than having to restructure code after

What if we implement "combined halo filling" in a separate PR? Then we can merge into this one and see.

simone-silvestri:
Sure! I can open a new PR!

src/BoundaryConditions/boundary_condition.jl
```diff
@@ -87,6 +87,7 @@ const HBC  = BoundaryCondition{<:Halo}
     PeriodicBoundaryCondition() = BoundaryCondition(Periodic, nothing)
       NoFluxBoundaryCondition() = BoundaryCondition(Flux,     nothing)
 ImpenetrableBoundaryCondition() = BoundaryCondition(Open,     nothing)
+        HaloBoundaryCondition() = BoundaryCondition(Halo,     nothing)
```
glwagner, line 90: `CommunicationBoundaryCondition`? (All boundary conditions _involve_ the halos, so `HaloBoundaryCondition` is a little ambiguous)

simone-silvestri: That can work!

validation/bickley_jet/bickley_jet.jl
```diff
@@ -80,96 +78,108 @@ function run_bickley_jet(; output_time_interval = 2, stop_time = 200, arch = CPU
...
-#     filepath = experiment_name * ".jld2"
+    @info "Making a fun movie about an unstable Bickley jet..."
 
-#     ζ_timeseries = DiskTimeSeries(:ζ, filepath)
-#     c_timeseries = DiskTimeSeries(:c, filepath)
+    filepath = experiment_name * ".jld2"
 
-#     grid = c_timeseries.grid
+    ζ_timeseries = DiskTimeSeries(:ζ, filepath)
+    c_timeseries = DiskTimeSeries(:c, filepath)
```
glwagner, line 142:
```suggestion
    ζ_timeseries = FieldTimeSeries(filepath. "ζ")
    c_timeseries = FieldTimeSeries(filepath, "c")
```

simone-silvestri:
if the grid has a value `FT` is substituted with `eltype(grid)` but we need a default for `WENO5(nothing)`?
The other option is to just request a grid all the times even if all directions are uniform

glwagner:
> if the grid has a value `FT` is substituted with `eltype(grid)` but we need a default for `WENO5(nothing)`?
> The other option is to just request a grid all the times even if all directions are uniform

Ah, I understand.

I'm hoping to "regularize" the advection schemes (eg always provide `grid`) via the design proposed in #2388 so maybe doesn't matter.

glwagner:
@simone-silvestri I noticed that a few validation tests were changed to use `MultiRegion`. What do you think about adding a new category of validation tests, maybe `validation/multi_region/`, and leaving the existing validation tests unchanged? If you think this is a good idea, I'll copy the changed tests into a new directory, and revert the existing tests to their original form.

glwagner:
I implemented a simple new validation test that runs a two-dimensional barotropic turbulence problem with `ExplicitFreeSurface`: https://github.com/CliMA/Oceananigans.jl/blob/ss/multi_region/validation/multi_region/multi_region_turbulence.jl

Here's some notes so far for 128x128x1 resolution:

* 1000 time-steps with `WENO5()` and in a "single region" takes about 3.9 seconds on my laptop
* 1000 time-steps with `WENO5(vector_invariant=VelocityStencil())` takes about 2.8 seconds on my laptop. Note that when we write "WENO5(vector_invariant=stencil)" we mean that we are using the vector invariant formulation of momentum with a WENO reconstruction for vorticity, using either "velocity" or "vorticity" in the WENO smoothness metric. The "WENO, Vector Invariant" scheme is probably faster because it has fewer WENO interpolations (just one per momentum component rather than 2).
* 1000 time-steps on a two-region grid with `WENO5()` takes 7.4 seconds
* `WENO5(vector_invariant=VelocityStencil())` blows up.

glwagner:
Update: https://github.com/CliMA/Oceananigans.jl/pull/2253/commits/a3c297bec11791e4080f62f1d7649c4f880f7c3e brings the 128^2 multi-region time down to 6.8 s, but I think there's still a ways to go. Next up is to figure out whether communication matters at all for this timing.

simone-silvestri:
If the solutions are the same, you can try removing the `fill_halo` to benchmark the implementation

francispoulin:
Very interesting results @glwagner and it's great to see this progress!  

A couple of questions.

When you use the `MultiRegionGrid`, it's still being run in serial, right?

Also, I noticed that the halos are of length `4`.  I thought `WENO` only needed `3`, but maybe I'm wrong?

simone-silvestri:
@francispoulin yes, it is serial on the CPUs. The halos are indeed 3 for the `WENO5`, but if we use a vector invariant weno formulation (so we are interpolating vorticity) we need 4 halos because calculating vorticity itself requires one halo.

@glwagner there is a type inference issue with `getregion(model, 1)` (you can see it in your validation test also). I think, solving that, the only overhead will be the `fill_halo` (and launching multiple kernels which maybe will not affect too much)

francispoulin:
Pretty cool to see that the `MultiRegionGrid` method can actually be faster!  But I imagine that is more by chance.  On average I would guess it would be as fast, which is I think what we want.  

Well done @simone-silvestri !

glwagner:
Now the only thing we have to do is fix 234823749 merge conflicts and we're good to go!

francispoulin:
The results in the table above were run on one `GPU` I gather.

Are we planning to do some tests on 2 and 4 GPU's to see how fast they are?

simone-silvestri:
Actually, this was on 1 CPU. I ll do some more in depth benchmarking today.

simone-silvestri:
I seem to have a problem with the GPU cubed sphere tests
```
ERROR: Out-of-bounds array access.
ERROR: a exception was thrown during kernel execution.
  | Run Julia on debug level 2 for device stack traces.
```
When running the tests on tartarus, both normally and with debug level 2, I cannot reproduce the error

glwagner:
> I seem to have a problem with the GPU cubed sphere tests
> 
> ```
> ERROR: Out-of-bounds array access.
> ERROR: a exception was thrown during kernel execution.
>   | Run Julia on debug level 2 for device stack traces.
> ```
> 
> When running the tests on tartarus, both normally and with debug level 2, I cannot reproduce the error

We've seen this issue before where sverdrup tests fail but tartarus passes. Can anyone remember where we saw that?

simone-silvestri:
I think it might have been with the regression tests in #2035, I am not sure how we solved it

simone-silvestri:
Finally all type issues have been resolved using three different fixes
(1) Ensure that `apply_regionally!` and `construct_regionally` are inlined always
(2) Avoid any Generator in the `getregion` function but explicitly pass arguments to it through metaprogramming the correct expressions in
(3) Use alternating inner and outer `getregion` to force compilation

glwagner:
Nice! Performance differences could be noise too. Is it possible to say what the changes to fill halo regions were? Just in case #2477 takes longer than expected?

simone-silvestri:
> So now the hold up is just the pencil arrays stuff?

and the cubed sphere issue

simone-silvestri:
On Satori using a stupidly large mesh `(1024 × 1024 × 100)` gives 89% efficiency going from 1 to 2 GPU for the `multi_region_turbulence.jl` benchmark (3.36 minutes for `RectilinearGrid` and 1.89 minutes for `MultiRegionGrid` on 2 devices)

There is still a lot of low hanging fruits I feels like because of the grid is smaller the performance is quite bad

glwagner:
> On Satori using stupidly large meshes gives 85% - 89% efficiency going from 1 to 2 GPU for the `multi_region_turbulence.jl` benchmark (Note `1440×600×48` is the size of the 1/4 degree simulation) Unfortunately the efficiency decreases on a larger number of GPUs... we definitely have to fix the scaling
> 
> #### Strong Scaling
> Grid size	Grid	GPUs	wall time	efficiency
> `1024×1024×100`	`RectilinearGrid`	1	3.4 minutes	100%
> `1024×1024×100`	`MultiRegionGrid`	2	1.9 minutes	89.5%
> `1440×600×48`	`RectilinearGrid`	1	1.4 minutes	100%
> `1440×600×48`	`MultiRegionGrid`	2	49.2 seconds	85.4%
> `1440×600×48`	`MultiRegionGrid`	3	38.8 seconds	72.2%
> Going to smaller meshes than these hampers the efficiency incredibly. I think there might be a lot of low hanging fruits to optimize multi GPU

Nice results though! Is this for implicit or explicit free surface?

simone-silvestri:
This is for explicit free surface

glwagner:
> This is for explicit free surface

Well dang. Ya work to do then.

simone-silvestri:
This is the same benchmark performed with `ImplicitFreeSurface`, by imposing a divergent velocity `u(x, y, z) = x / 10` to make sure the implicit solver iterates. Looking at the results it seems like it doesn't iterate too much... (probably WENO cleans up?) And it is very weird that the `RectilinearGrid` version is not affected by the FreeSurface calculation? (I have double checked that the free surface solver is correct)

#### Strong Scaling
| Grid size | Grid | GPUs | wall time | efficiency |
| -- | -- | -- | -- | -- |
| `1440×600×48`| `RectilinearGrid` | 1 | 1.37 minutes | 100% |
| `1440×600×48`| `MultiRegionGrid` | 2 | 1.05 seconds | 65.2% |

glwagner:
> This is the same benchmark performed with `ImplicitFreeSurface`, by imposing a divergent velocity `u(x, y, z) = x / 10` to make sure the implicit solver iterates. Looking at the results it seems like it doesn't iterate too much... (probably WENO cleans up?) And it is very weird that the `RectilinearGrid` version is not affected by the FreeSurface calculation? (I have double checked that the free surface solver is correct)
> 
> #### Strong Scaling
> Grid size	Grid	GPUs	wall time	efficiency
> `1440×600×48`	`RectilinearGrid`	1	1.37 minutes	100%
> `1440×600×48`	`MultiRegionGrid`	2	1.05 minutes	65.2%

THATS CALLED A SPEED UP 🍻

test/test_cubed_spheres.jl
```diff
@@ -18,72 +18,77 @@ using Oceananigans.Models.HydrostaticFreeSurfaceModels: VerticalVorticityField
...
-                                            closure = nothing,
-                                            tracers = :c,
-                                            buoyancy = nothing)
-
-        @testset "Constructing a grid from file [$(typeof(arch))]" begin
-            @test grid isa ConformalCubedSphereGrid
-        end
+        # These tests cause an underfined `Bound Acces Error` on GPU() CI with the new CUDA version.
+        # This error is not reproducible on neither on Tartarus not Sverdrup.
+        # Therefore, they are excised for the moment (PR #2253)
```
glwagner, line 23: Smells like progress

glwagner (approved):

test/test_cubed_spheres.jl
simone-silvestri: always improving :)

simone-silvestri:
Using 
```
grid     = RectilinearGrid(arch, size=(Nx, Ny, Nz), halo=(4, 4, 4), x=(0, 2π), y=(0, 2π), z=(0, 1), topology=(Periodic, Periodic, Bounded))

bottom = zeros(Nx, Ny)
bound  = Int.(Ny/2-40:Ny/2+40)
bottom[bound, bound] .= 0.5

mrg  = MultiRegionGrid(grid, partition=XPartition(2), devices = 2)
```
and 
```
u(x, y, z) = (x * z) / 10
```

(with `ImplicitFreeSurface`) we get

#### Strong Scaling
| Grid size | Grid | GPUs | wall time | efficiency |
| -- | -- | -- | -- | -- |
| `1440×600×48`| `RectilinearGrid` | 1 | 1.848 minutes | 100% |
| `1440×600×48`| `MultiRegionGrid` | 2 | 1.12 minutes | 82.5% |

Scaling gets better?? at this point I a little confused...
Bottomline... 
there is still a bunch of optimization and more systematic benchmarking to be done, I'll merge this PR and then we can think about improving the scaling, the first things that come in mind are
- Ensure that all the `apply_regionally!` and `construct_regionally` calls are asynchronous. This might not be the case if there are memory copies inside function calls. That would serialize the execution of part of the code. To ensure this we require a more in-depth profiling using ***nsys***
- remove all `fill_halo_regions!` that are not `Periodic` or `Communication` which will allow asynchronous execution of halo filling across different direction (luckily already being done in #2477)
- Bundle together the halo passing in a single boundary buffer to allow sending field tuples together (depends on #2509)

Additional work to do on `MultiRegion` is 
- Perform more systematic benchmarking
- Design correct `OutputWriters` and `OutputReaders` for `MultiRegionFields`

Maybe not immediate priorities but definitely important
- Adapt `RungeKutta3` to `Multiregion` through `@apply_regionally`
- Implement a multi-region version of the Nonhydrostatic pressure solver
- Overlap computation and communication (for this we require _non-blocking_ `fill_halo_regions!` with returning `events`)
- Optimize the distributed implicit solver (maybe using `PETSc.jl`? Might not be needed when the implicit solve fits in one GPU)

tomchor:
That's awesome @simone-silvestri. Great work!

A couple of questions:

> - Design correct OutputWriters and OutputReaders for MultiRegionFields

Does this mean that the code with multi-grid doesn't have the ability to output things at the moment? Or does it output to one file per region/GPU?

> - Implement a multi-region version of the Nonhydrostatic pressure solver

So I take that a nonhydrostatic pressure solver for multi-grid isn't defined yet? Or is it just non-optimized (but will run)?

simone-silvestri:
@tomchor, I haven't implemented `MultiRegion` for `NonhydrostaticModel` yet. It would be quite easy to extent it if it weren't for the pressure solver which is done in two different ways and both of them are global (`FFT` and `TridiagonalFFT`). A quick and dirty solution would be to do as I have done with the `ImplicitFreeSurface` and reserve the calculation on 1 GPU only. The problem is that the pressure solve is 3D and not 2D so we would probably like to use the `cuFFT` multi-gpu functionality (in particular the cuFFTxt library, which I hope is implemented in `CUDA.jl` although I haven't checked yet) I ll open a PR to start implementing it

Regarding the `OutputWriters`, they are implemented in jld2 trivially with `reconstruct_global_field`. It runs, but there is a weird bug for which the fields are not updated. I think it is solvable easily but I have to take a deeper look. On the other hand, to anticipate the `CubedSphere` refactor, we probably want to have `OutputWriters` which output all regions separately (as `reconstruct_global_field` does not mean much in a cubed sphere context).
The `OutputReaders`, on the other hand, are completely undefined

glwagner:
Seems maybe that we want to output to a single file for every MultiRegionGrid, but in some special way by extending stuff in

https://github.com/CliMA/Oceananigans.jl/blob/main/src/OutputWriters/fetch_output.jl

Whatever we do there can then be interpretable by `FieldTimeSeries` to make sense of things.

"reconstructing" output writers can probably be implemented with function output right now.

glwagner:
That'd be just amazing if cuFFTxt works out of the box eh?

tomchor:
> I ll open a PR to start implementing it

That's awesome! Thank you! Let me know how I can help out.

src/Fields/broadcasting_abstract_fields.jl
```diff
@@ -31,13 +31,6 @@ const BroadcastedArrayOrCuArray = Union{Broadcasted{<:DefaultArrayStyle},
     end
 end
 
-# TODO: make this support Field that are windowed in _only_ 1 or 2 dimensions.
-# Right now, this may only produce expected behavior (re: dimensionality) for
-# WindowedField that are windowed in three-dimensions. Of course, broadcasting with
-# scalar `bc` is no issue.
-@inline Base.Broadcast.materialize!(dest::WindowedField, bc::BroadcastedArrayOrCuArray) =
-    Base.Broadcast.materialize!(parent(dest), bc)
-
```
glwagner, line 28: What went wrong?

glwagner: I guess this affects that minor case when you wabt to broadcast a field and an Array (because the Array has non-shifted indexing, but the `WindowedField` has the "global" indices)

simone-silvestri: it was a duplicated at 45-46?

glwagner: oh 🤦

PR #2255, opened 2022-02-19
Update CUDA and CUDAKernels take #2

navidcy:


vchuravy:
The new Kernel abstractions dropped support for Julia 1.6

navidcy:
All right!

@glwagner perhaps it's time for Julia v1.7?

PR #2262, opened 2022-02-20
Adds a sentence to the Anisotropic diffusivity section of the docs

tomchor:
I think the docs weren't 100% clear before on how to actually use different horizontal and vertical components in a single closure. I also couldn't figure out how to set `νx` different from `νy`, but I'm assuming that possible. So if anyone knows, I think we could also add that there.

### DIFF
--- a/docs/src/model_setup/turbulent_diffusivity_closures_and_les_models.md
+++ b/docs/src/model_setup/turbulent_diffusivity_closures_and_les_models.md
@@ -45,6 +45,9 @@ isotropy: Oceananigans.TurbulenceClosures.Vertical
 
 ```
 
+After that you can set `closure = (horizontal_closure, vertical_closure)` when constructing the model so that
+all components will be taken into account when calculating the diffusivity term.
+
 ## Smagorinsky-Lilly
 
 To use the Smagorinsky-Lilly LES closure, no parameters are required


glwagner:
> I also couldn't figure out how to set `νx` different from `νy`, but I'm assuming that possible.

I don't think this is possible anymore, right @simone-silvestri ?

glwagner (approved):

simone-silvestri:
You are right, there is no more support for different `νx` and `νy`. 

I see no problem with adding `XDirection`, `YDirection` and `ZDirection` `ScalarDiffusivities`

simone-silvestri (approved):

tomchor:
I'm gonna merge this as is then, since there's nothing else to add.

> You are right, there is no more support for different `νx` and `νy`.
> 
> I see no problem with adding `XDirection`, `YDirection` and `ZDirection` `ScalarDiffusivities`

My vote goes to having this. I think this generally is welcome, and most of the code seems to be already there anyway.

glwagner:
I noticed we have `ZDirection` in `Oceananigans.Grids` --- so we can reuse that for `ScalarDiffusivity`, and define `XDirection` and `YDirection`.

> [](https://github.com/tomchor)My vote goes to having this. I think this generally is welcome, and most of the code seems to be already there anyway.

What's the use case?

tomchor:
I didn't have a use-case in mind when suggesting this. It'll be rare for oceanography that two horizontal directions need to have different eddy diffusivities (though I'm sure that would be reasonable in some cases (balanced jet?)), but I think it's worth to implement that  capability if we're trying to make Oceananigans a flexible, general-purpose-ish CFD package. Plus, given that parallelization capabilities are limited to only some directions, I can envision cases where users will rotate their domain in a way where the "horizontal" directions are actually `y`, `z`, and then `HorizontalDirections` won't work as intended.

In any case, I personally will probably never use this feature, and it was only a suggestions for possible future uses. So feel free to ignore it if you think it's not worth it :)

glwagner:
That's fair. I think it's preferred generally to implement features when there's a strong need for it. The reason to wait is because code features (and their tests) often (not always) impose a maintenance burden on developers. So, given that supporting some features might slow down the pace of code development, it's in the interest of both users and developers to keep low-priority features out of the code for as long as possible / forever.

Rotating a domain is an interesting use case. Other solutions are certainly cleaner, but pragmatism may force us in that direction.

Issue #2254, opened 2022-02-18
Export `Horizontal` and `Vertical` diffusivity isotropies from top-level?


glwagner:
We shouldn't have to import them into user-scripts as in

https://github.com/CliMA/Oceananigans.jl/blob/80f328b9d94bb4f13af8347f06356b93127638cc/examples/eady_turbulence.jl#L123

if we have conflicts then we might need more verbose names like `HorizontalIsotropy`, `VerticalIsotropy`. Should resolve ASAP if we need a name change.

tomchor:
I agree with exporting them by default. I also think if we change notation to `ZDirection`, `XDirection` etc., it will probably decrease the chance for name conflicts.

navidcy:
Yeap, export!

glwagner:
Stale

Issue #2261, opened 2022-02-20
Is there a better word than `isotropy` to use for the new ScalarDiffusivity?


tomchor:
I'm just now reading about the new `ScalarDiffusivity` and trying to use it. I think it's a great idea and I love how it simplifies things, but the use of the word "isotropy" kinda threw me off a bit.

To me setting a scalar diffusivity as `ScalarDiffusivity(..., isotropy=Horizontal())`, implies that the diffusivity is the same in the horizontal directions, but that you can also set a different diffusivity in the vertical direction. Which you can't in this instance of the constructor; you have to create a new one.

There's an extra issue I think if you set `ScalarDiffusivity(..., isotropy=Vertical())`. Since by definition isotropy is [uniformity in all orientations](https://en.wikipedia.org/wiki/Isotropy#:~:text=Isotropy%20is%20uniformity%20in%20all,the%20prefix%20an%2C%20hence%20anisotropy.), I don't think it's accurate to say something is isotropic over one single direction.

Maybe a better word would be "directions"? Or directionality? (if that's even a word...)

glwagner:
We can consider changing this kwarg to `directions`. I don't think we had much a conversation about it, but the two we considered originally were `direction` and `isotropy`. One motivation for using `isotropy` is to smoothly transition from "IsotropicDiffusivity" (with three-dimensionality implied) and "AnisotropicDiffusivity" (meaning, different in the horizontal and vertical directions) to a single struct. But having a clear meaning for most people is more important than that continuity.

Isotropy in _one_ direction is accurate in my opinion, though I agree its the trivial case (like a "set" with one element) and therefore reads a little awkwardly. Diffusion can act in the "horizontal directions" while being horizontally-anisotropic (ie the diffusion of a vector depends on its orientation), so the meaning is a bit different. But, we can document that `ScalarDiffusivity` is isotropic in its `directions` for vectors like momentum.

> To me setting a scalar diffusivity as `ScalarDiffusivity(..., isotropy=Horizontal())`, implies that the diffusivity is the same in the horizontal directions, but that you can also set a different diffusivity in the vertical direction.

We can indeed use a second closure to set a vertical diffusivity.

rafferrari:
I tend to agree that "direction" is a better descriptor. Isotropy implies same kappa in different directions, so it requires at least two directions to be meaningful.

glwagner:
[...]

PR #2266, opened 2022-02-22
Tentative changes to `ScalarDiffusivity`

tomchor:
This PR is an attempt at unifying the directions used by `TurbulenceClosures` a bit, in addition to a few minor things. In summary, this PR

- Nukes the `Vertical` definition at `abstract_scalar_diffusivity_closure.jl` and uses the already-defined `ZDirection` instead. I think this decreases the amount of code but is also more accurate (since for rotated domains the `z` direction of the code isn't necessarily the "physical" vertical direction)
- Renames `ThreeDimensional`, `Horizontal` to `XYZDirections` and `XYDirections` and moves their definition to `Grids`. This was done for consistency with `ZDirection` and also follows the same argument about rotated domains.
- Exports `XYZDirections`, `XYDirections` and `ZDirection` at the top level
- Makes minor improvements to documentation

I could also change the notation from `isotropy` to `dimensions`, according to https://github.com/CliMA/Oceananigans.jl/issues/2261, depending on how people feel. I haven't done that yet because there doesn't appear to be a consensus on what to do there as of writing this.

Closes https://github.com/CliMA/Oceananigans.jl/issues/2254

### DIFF
--- a/docs/src/model_setup/turbulent_diffusivity_closures_and_les_models.md
+++ b/docs/src/model_setup/turbulent_diffusivity_closures_and_les_models.md
@@ -17,36 +17,39 @@ julia> closure = ScalarDiffusivity(ν=1e-2, κ=1e-2)
 ScalarDiffusivity:
 ν=0.01, κ=0.01
 time discretization: Oceananigans.TurbulenceClosures.Explicit()
-isotropy: Oceananigans.TurbulenceClosures.ThreeDimensional
+isotropy: Oceananigans.TurbulenceClosures.XYZDirections
 ```
 
 ## Constant anisotropic diffusivity
 
 To specify constant values for the horizontal and vertical viscosities, ``\nu_h`` and ``\nu_z``, and horizontal and vertical
-diffusivities, ``\kappa_h`` and ``\kappa_z``, you can use [`ScalarDiffusivity(isotropy = Horizontal())`](@ref) and
-[`ScalarDiffusivity(isotropy = Vertical())`](@ref)
+diffusivities, ``\kappa_h`` and ``\kappa_z``, you can use [`ScalarDiffusivity(isotropy = XYDirections())`](@ref) and
+[`ScalarDiffusivity(isotropy = ZDirection())`](@ref)
 
 ```jldoctest
 julia> using Oceananigans.TurbulenceClosures
 
-julia> using Oceananigans.TurbulenceClosures: Horizontal, Vertical
+julia> using Oceananigans.TurbulenceClosures: XYDirections, ZDirection
 
-julia> horizontal_closure = ScalarDiffusivity(ν=1e-3, κ=2e-3, isotropy=Horizontal())
+julia> horizontal_closure = ScalarDiffusivity(ν=1e-3, κ=2e-3, isotropy=XYDirections())
 ScalarDiffusivity:
 ν=0.001, κ=0.002
 time discretization: Oceananigans.TurbulenceClosures.Explicit()
-isotropy: Oceananigans.TurbulenceClosures.Horizontal
+isotropy: Oceananigans.TurbulenceClosures.XYDirections
 
-julia> vertical_closure = ScalarDiffusivity(ν=1e-3, κ=2e-3, isotropy=Vertical())
+julia> vertical_closure = ScalarDiffusivity(ν=1e-3, κ=2e-3, isotropy=ZDirection())
 ScalarDiffusivity:
 ν=0.001, κ=0.002
 time discretization: Oceananigans.TurbulenceClosures.Explicit()
-isotropy: Oceananigans.TurbulenceClosures.Vertical
+isotropy: Oceananigans.TurbulenceClosures.ZDirection
 
 ```
 
-After that you can set `closure = (horizontal_closure, vertical_closure)` when constructing the model so that
-all components will be taken into account when calculating the diffusivity term.
+After that you can set `closure = (horizontal_closure, vertical_closure)` when constructing the
+model so that all components will be taken into account when calculating the diffusivity term. Note
+that `ScalarDiffusivity`s with `isotropy=XYDirections()` are implemented using a [different
+scheme](https://mitgcm.readthedocs.io/en/latest/algorithm/algorithm.html#horizontal-dissipation)
+from `isotropy=XYZDirections()` and `isotropy=ZDirections()` that conserves potential vorticity.
 
 ## Smagorinsky-Lilly
 
--- a/docs/src/model_setup/turbulent_diffusivity_closures_and_les_models.md
+++ b/docs/src/model_setup/turbulent_diffusivity_closures_and_les_models.md
@@ -29,8 +29,6 @@ diffusivities, ``\kappa_h`` and ``\kappa_z``, you can use [`ScalarDiffusivity(is
 ```jldoctest
 julia> using Oceananigans.TurbulenceClosures
 
-julia> using Oceananigans.TurbulenceClosures: XYDirections, ZDirection
-
 julia> horizontal_closure = ScalarDiffusivity(ν=1e-3, κ=2e-3, isotropy=XYDirections())
 ScalarDiffusivity:
 ν=0.001, κ=0.002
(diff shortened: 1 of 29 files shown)


glwagner:
I prefer `HorizontalDirections` to `XYDirections`. How do others feel?

glwagner:
As for `ZDirection`, I'm ok with this. However, it's worth noting that we use "vertically-implicit" (a term in widespread usage) to denote an implicit treatment of the "z-direction" terms in the diffusivity. Food for thought...

simone-silvestri:
I am also ok with `Vertical` and `Horizontal`, but would refrain from calling them `Directions`, since it implies a bit more than just a laplacian in the vertical or horizontal direction. I think something on the line of `VerticalFormulation`, `HorizontalFormulation` and `ThreeDimensionalFormulation` is better.

tomchor:
@glwagner and I talked yesterday and here are some points that emerged (@glwagner feel free to chime in if I'm forgetting/misrepresenting something).

We agreed that the important points in this PR are two. (1) We should export all directions/formulations (`ZDirection`, `XYDirections`, `ThreeDimensional` or whatever we end up calling them) at the top level, and (2) we shouldn't reintroduce a `Vertical` direction in `TurbulenceClosures` when there already exists a `ZDirection` defined in `Grids` that's used for buoyancy purposes.

However, we need to agree on a notation. A few main options emerged from our talk:

### Option 1

First option is we keep calling the directions `Vertical` and `Horizontal`, since those names are pretty intuitive and play well with  already-defined terms like `VerticallyImplicit` (which are also intuitive and widely used in the literature). The downside to this option, if we were to follow (2), appears as an inconsistency whenever we tilt the domain and need to use `buoyancy = Buoyancy(model=BuoyancyTracer(), vertical_unit_vector=ĝ)` (`ĝ` being a unit vector). So we could end up with a case where the vertical direction is implicitly defined as being tilted 45 degrees (by means of `Buoyancy`), but for `ScalarDiffusivity` `Vertical` will still mean the `z` direction, which is confusing.

Furthermore, the default option for buoyancy would be called `buoyancy = Buoyancy(model=BuoyancyTracer(), vertical_unit_vector=Vertical())`, which is not helpful.

### Option 2

Second option is to change the notation to what I proposed here: `XYDirection` etc., which avoids this inconsistency. The downside is that it doesn't play well with terms like `VerticallyImplicit` anymore. Something we either would have to live with, or we'd need to change those terms as well (something like `ZImplicit` isn't wrong, but it's not what people generally call it).


### Option 3

We could also keep the names `Vertical` etc., but change the interface of `Buoyancy` so that it doesn't implicitly define that the vertical direction is. So instead of having the interface be

 `buoyancy = Buoyancy(model=BuoyancyTracer(), vertical_unit_vector=ĝ)`

it would be something like

 `buoyancy = Buoyancy(model=BuoyancyTracer(), direction_antiparallel_to_gravity=ĝ)`

which is equally correct and doesn't assume that "vertical" is connected to the direction of gravity. The issue here is that the name `direction_antiparallel_to_gravity` is super long.


### Option 4

@simone-silvestri raised an interesting point that these aren't just "directions" they are also different formulations. In light of this, it might make sense to not do (2). That way we'd leave `ZDirection` in `Grids`, and define `VerticalFormulation`, `HorizontalFormulation` and `ThreeDimenionalFormulation` in `TurbulenceClosures`. Since these latter aren't "directions" anymore, there wouldn't be an overlap with `ZDirection` (which literally indicates only a direction).

Thoughts?

glwagner:
I think option 4 -- `VerticalFormulation`, `HorizontalFormulation` -- probably makes the most sense for closures. I think it's important that users realize that these types indicate more than just the direction. If we use the name of a "direction" like vertical, horizontal, zdirection, etc, then this distinction is lost.

I also think in terms of API there are a few choices. The most verbose is something like

```julia
closure = ScalarDiffusivity(ν=1, κ=1, formulation=VerticalFormulation()) # kwarg
closure = ScalarDiffusivity(VerticalFormulation(), ν=1, κ=1) # positional argument
closure = ScalarDiffusivity{VerticalFormulation}(ν=1, κ=1) # explicit type parameter specification
closure = VerticalScalarDiffusivity(ν=1, κ=1) # alias
```

I think the last form above is best.

Separately, this discussion make me think we should use `gravity_unit_vector` instead of `vertical_unit_vector` in the API for `Buoyancy` (to avoid conflicting defintions between "vertical" in `Buoyancy` and `TurbulenceClosures`). We can define `NegativeZDirection()` as the default for `gravity_unit_vector`.

simone-silvestri:
I agree with switching `Vertical` or `Horizontal` to a positional argument if it has to be verbose, but then also switch the time discretization to a positional argument and `VerticallyImplicitTimeDiscretization` and `ExplicitTimeDiscretization`

I also like the idea of changing the name for the vector along which buoyancy acts which I agree is not consistent with the term `Vertical`, `gravity_unit_vector` or `buoyancy_unit_vector` are fine, and for that we can keep the current default

simone-silvestri (approved):

tomchor:
Great, I also think option 4 is the best. I prefer the kwarg formulation, but I don't feel strongly about it.

The issue atm is that I think I won't have time to finish this PR by myself for a few days (I couldn't figure out why some tests aren't passing, so it'll probably take me a fair amount of time to have everything done). So if this needs to be ready for AGU OSM I'd need some help getting the tests to pass (it might actually be better to start over in another PR).

simone-silvestri:
I can help you out today if you want, if you change to the required formulation I can make the tests pass. Just let me know when it is ready to fix the bugs

glwagner:
> Great, I also think option 4 is the best. I prefer the kwarg formulation, but I don't feel strongly about it.
> 
> The issue atm is that I think I won't have time to finish this PR by myself for a few days (I couldn't figure out why some tests aren't passing, so it'll probably take me a fair amount of time to have everything done). So if this needs to be ready for AGU OSM I'd need some help getting the tests to pass (it might actually be better to start over in another PR).

I can open a new PR if you're ok with that.

I guess `buoyancy_unit_vector` is in fact the right name (rather than `vertical_unit_vector`! `gravity_unit_vector` (which is antiparallel to `buoyancy_unit_vector`) might be more familiar though.

glwagner:
> I can help you out today if you want, if you change to the formulation we decided I can try to fix the tests. Just let me know when it is ready to fix the bugs

Or we can fix this up!

glwagner:
I feel like the alias is the least confusing (and prettiest) of the options, since it enforces the notion that these are completely distinct objects. We can define both a type alias _and_ constructor.

simone-silvestri:
ok so we can have positional argument and alias
```
const VerticalScalarDiffusivity(args...; kwargs...) = ScalarDiffusivity(VerticalFormulation(), args...; kwargs...)
```

tomchor:
> > Great, I also think option 4 is the best. I prefer the kwarg formulation, but I don't feel strongly about it.
> > The issue atm is that I think I won't have time to finish this PR by myself for a few days (I couldn't figure out why some tests aren't passing, so it'll probably take me a fair amount of time to have everything done). So if this needs to be ready for AGU OSM I'd need some help getting the tests to pass (it might actually be better to start over in another PR).
> 
> I can open a new PR if you're ok with that.
> 
> I guess `buoyancy_unit_vector` is in fact the right name (rather than `vertical_unit_vector`! `gravity_unit_vector` (which is antiparallel to `buoyancy_unit_vector`) might be more familiar though.

I think starting over in another PR is probably the way to go unfortunately. It'll probably take too long to revert some of the things I initially did here. I'll try to help out there however I can.

glwagner:
> ```julia
> ```julia
> closure = VerticalScalarDiffusivity(ν=1, κ=1) # alias
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> ```

Yes perfect!

glwagner:
Put `time_discretization` first in the arguments

simone-silvestri:
Ok then I ll open a new PR and change the following:

### Scalar Diffusivity

`formulation` and `time_disc` as positional arguments, defaults: 3D and Explicit
from `Vertical` to `VerticalFormulation`
from `Explicit` to `ExplicitTimeDiscretization` (again :sweat_smile:)

aliases for `VerticalScalarDiffusivity`, `HorizontalScalarDiffusivity`

all the above exported to the public API

### Buoyancy

change to `gravity_unit_vector` 

if we all agree I proceed

PR #2268, opened 2022-02-22
Initialize schedules at iteration 0

glwagner:
This PR tweaks `initialize_simulation!` so that the schedules get called when `iteration == 0` --- in addition to running diagnostics, executing callbacks, and writing output.

This is a tiny step in the direction of having a robust initialization procedure. I think we still need to work on the abstractions a bit though, since it'd be nice to initialize schedules / callbacks more explicitly rather than implicitly relying on `iteration == 0`. This change will affect `SpecifiedTimes` and `WallTimeInterval` the most, I think.

TODO:

- [ ] Test for initialization of `SpecifiedTimes`?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
-version = "0.71.1"
+version = "0.71.2"
 
 [deps]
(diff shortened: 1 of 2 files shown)


francispoulin:
I like the new set up.  Very clean, and following this structure will be nice moving forward.

navidcy (approved):

glwagner:
I'm going to merge this, then hopefully combine a few more commits into a new minor version.

PR #2273, opened 2022-02-22
CATKE with an improved model for shear-driven mixing

### LINKED ISSUE #2272: Update on CATKE default parameter values?
Recent calibration developments may have produced better defaults, e.g., here:
https://github.com/CliMA/Oceananigans.jl/blob/b3ddbc84c8f35aaf5b93fbbfdb4cffcada5c6533/src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/CATKEVerticalDiffusivities.jl#L108
and
https://github.com/CliMA/Oceananigans.jl/blob/b3ddbc84c8f35aaf5b93fbbfdb4cffcada5c6533/src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/mixing_length.jl#L69-L85

Should we update these values?

(I'd like to add CATKE closure to the channel...)

glwagner:
This PR updates CATKE with an improved model for how the presence of shear modulates the mixing length and turbulent mixing.

A few preliminary calibration results suggest that these changes are very promising, especially at coarse resolutions. A short summary is that, originally, CATKE was flexible enough to model both convection and shear-driven mixing --- just not at the same time (because TKE levels are much higher for shear-driven mixing). Adding a more explicit model for the modulation of the mixing length in the presence of shear allows CATKE to capture both convection and shear-driven regimes.

We can update the default values for CATKE in this PR as well.

This figure is a little out of context, but it shows agreement between LES (gray line) and the "best parameters" from a calibration run (blue line) at 8m resolution with five physical scenarios (from LESbrary's 4 day suite). The "other line" is the "worst" parameters in the calibration ensemble (so that one should be ignored).

image

Closes #2272


navidcy:
Yes. Yes. Yes.

navidcy:
@glwagner could you resolve the conflicts?

navidcy (approved):

glwagner:
> @glwagner could you resolve the conflicts?

For sure, I originally added this to Oceanganigans 0.70.3 so I need to merge `main`.

glwagner:
I don't think we should resolve the conflicts traditioonally, we should just completely replace the old file with the new...

navidcy:
> I don't think we should resolve the conflicts traditioonally, we should just completely replace the old file with the new...

Well I didn't necessarily imply a traditional resolve method!

Nice!!

glwagner:
Ok @navidcy I've updated the parameters.

For the sake of documentation: these parameters were obtained by calibrating CATKE against 5 LES from LESbrary's "4 day suite" at 8m resolution. The last iteration of the calibration produced these plots:

image

The TKE is a bit off, but the buoyancy profiles are very good. The shear is slightly worse.

I have some ideas for improving CATKE further, but it's going to take a bit of time.

I think moving forward the right approach is to always use CATKE's default configuration for experiments. Then, whenever we update CATKE we'll take a new version. That way we can figure out what CATKE flavor we used by referencing the Oceananigans version.

glwagner:
This is good to go so I'll merge.

PR #2275, opened 2022-02-24
Adds drag to grid-fitted Immersed boundaries

tomchor:
The main goal of this PR (done jointly with @whitleyv) is to add drag as an option to grid-fitted Immersed boundaries.

This is draft for now since first we want to get something that is working properly without worrying to much about user-interface.


glwagner:
@simone-silvestri and I discussed how to implement this, and our tentative plan is to add a new term to the tendency kernels:

https://github.com/CliMA/Oceananigans.jl/blob/main/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl

https://github.com/CliMA/Oceananigans.jl/blob/main/src/Models/NonhydrostaticModels/velocity_and_tracer_tendencies.jl

The reason for this approach is to avoid entangling the immersed boundary condition calculations with the existing closure kernel functions. We're not totally sure what the immersed boundary condition stuff will entail and we think this approach will be easier to maintain and extend in the near future.

Hopefully we can adapt some of the ideas and code in this PR to that eventual implementation. I think this implementation will prove useful for users who just want to use the hard-coded boundary conditions.

tomchor:
> Hopefully we can adapt some of the ideas and code in this PR to that eventual implementation. I think this implementation will prove useful for users who just want to use the hard-coded boundary conditions.

I think so. Let's definitely try!

Is it useful to have a general `Drag` boundary condition in general (in addition to `Periodic`, etc.), that would be used not only for immersed boundaries, but also for actual boundaries?

tomchor:
CC @wenegrat

glwagner:
> > Hopefully we can adapt some of the ideas and code in this PR to that eventual implementation. I think this implementation will prove useful for users who just want to use the hard-coded boundary conditions.
> 
> I think so. Let's definitely try!
> 
> Is it useful to have a general `Drag` boundary condition in general (in addition to `Periodic`, etc.), that would be used not only for immersed boundaries, but also for actual boundaries?

We have `Flux`, which is the same.

glwagner:
> Quick general question: with this interface IBM immediately goes to drag BC, right?

That's correct. There's no way to run IBM without a drag BC with this implementation.

tomchor:
> We have `FluxBoundaryCondition` already which is the same. The difference in terms of user input is that an immersed boundary condition _function_ has to be called with three indices `i, j, k` rather than two like all the other boundary conditions. I don't think we need any new types in the source code.

I wouldn't say it's the same. But it is much more general though, and a drag can be applied with `Flux`. But the reason I ask is that it's much harder to impose an _arbitrary_ flux on an immersed boundary. I think it'd be complex to translate betweewn `u`, `v`, `w` to tangential and normal directions of the solid (both for the user and on the developer side).

simone-silvestri:
Nice work! 

Just as food for thought, another way to specify drag (in this case linear) is as a forcing 

```
# Linear bottom drag:
μ = 0.001 # ms⁻¹

# Tells you whether to apply the drag or not on the last cell
@inline is_immersed_drag_u(i, j, k, grid) =
     Int(solid_interface(Face(), Center(), Center(), i, j, k-1, grid) & !solid_node(Face(), Center(), Center(), i, j, k, grid))

# Keep a constant linear drag parameter [s⁻¹] independent of vertical level
@inline u_immersed_bottom_drag(i, j, k, grid, clock, fields, μ) =
                  @inbounds - μ * is_immersed_drag_u(i, j, k, grid) * fields.u[i, j, k] / Δzᶠᶜᶜ(i, j, k, grid)

Fu = Forcing(u_immersed_bottom_drag, discrete_form = true, parameters = μ)
```

This is just the same as how flux BC are treated (implicitly) in `apply_flux_bc` and (as said by @glwagner) was the direction we were thinking to go in the future for adding fluxes (immersed boundaries and boundaries).

But surely this is a good exploration! Eventually we should decide how to treat fluxes and unify all formulations

whitleyv:
> Just as food for thought, another way to specify drag (in this case linear) is as a forcing
> ```
> @inline u_immersed_bottom_drag(i, j, k, grid, clock, fields, μ) =
>                   @inbounds - μ * is_immersed_drag_u(i, j, k, grid) * fields.u[i, j, k] / Δzᶠᶜᶜ(i, j, k, grid)
> ```

I was implementing this way, but was unsure if dividing by Δz, or whichever direction, is always going to give the divergence correctly, such as when it's not linear. Are there cases where it would be more complicated and difficult to formulate in this way? Or am I wrong on the role of dividing by Δz? This is why we went directly for inputting it as the flux instead of as a forcing.

glwagner:
> > Just as food for thought, another way to specify drag (in this case linear) is as a forcing
> > ```
> > @inline u_immersed_bottom_drag(i, j, k, grid, clock, fields, μ) =
> >                   @inbounds - μ * is_immersed_drag_u(i, j, k, grid) * fields.u[i, j, k] / Δzᶠᶜᶜ(i, j, k, grid)
> > ```
> 
> I was implementing this way, but was unsure if dividing by Δz, or whichever direction, is always going to give the divergence correctly, such as when it's not linear. Are there cases where it would be more complicated and difficult to formulate in this way? Or am I wrong on the role of dividing by Δz? This is why we went directly for inputting it as the flux instead of as a forcing.

There's no mathematical distinction between a "flux divergence" and a "forcing" (both contribute to the volume averaged tendency in a cell).

If you use a flux, then note the divergence will be taken by multiplying by cell area and dividing by volume.

If you implement a flux divergence as a forcing as @simone-silvestri then you have to do the math yourself to multiple by the appropriate cell area and divide by volume.

Since currently all our grids are extruded in the vertical, vertical area / volume is equal to the vertical cell spacing. That's why this factor appears in @simone-silvestri vertical flux.

src/ImmersedBoundaries/conditional_fluxes.jl
```diff
@@ -44,25 +97,25 @@ const ATD = AbstractTimeDiscretization
 
 # dx(uu), dy(vu), dz(wu)
 # ccc,    ffc,    fcf
-@inline _advective_momentum_flux_Uu(i, j, k, ibg::GFIBG, args...) = conditional_flux_ccc(i, j, k, ibg, advective_momentum_flux_Uu, args...)
-@inline _advective_momentum_flux_Vu(i, j, k, ibg::GFIBG, args...) = conditional_flux_ffc(i, j, k, ibg, advective_momentum_flux_Vu, args...)
-@inline _advective_momentum_flux_Wu(i, j, k, ibg::GFIBG, args...) = conditional_flux_fcf(i, j, k, ibg, advective_momentum_flux_Wu, args...)
+@inline _advective_momentum_flux_Uu(i, j, k, ibg::GFIBG, closure, U, args...) = conditional_flux_ccc(i, j, k, ibg, advective_momentum_flux_Uu, closure, U, args...)
+@inline _advective_momentum_flux_Vu(i, j, k, ibg::GFIBG, closure, U, args...) = conditional_flux_ffc(i, j, k, ibg, advective_momentum_flux_Vu, closure, U, args...)
+@inline _advective_momentum_flux_Wu(i, j, k, ibg::GFIBG, closure, U, args...) = conditional_flux_fcf(i, j, k, ibg, advective_momentum_flux_Wu, closure, U, args...)
```
tomchor, line 102: Depending on how I define the arguments of this function, I get either an error when calling `div_𝐯u` (with a looong stack trace and no obvious error) or this just calls the default `_advective_momentum_flux_Wu(...) = advective_momentum_flux_Wu(...)` method...

I'm not sure what I'm getting wrong here, but suggestions are welcome.

src/ImmersedBoundaries/conditional_fluxes.jl
```diff
@@ -44,25 +97,25 @@ const ATD = AbstractTimeDiscretization
...
 
 # dx(uv), dy(vv), dz(wv)
 # ffc,    ccc,    cff
-@inline _advective_momentum_flux_Uv(i, j, k, ibg::GFIBG, args...) = conditional_flux_ffc(i, j, k, ibg, advective_momentum_flux_Uv, args...)
-@inline _advective_momentum_flux_Vv(i, j, k, ibg::GFIBG, args...) = conditional_flux_ccc(i, j, k, ibg, advective_momentum_flux_Vv, args...)
-@inline _advective_momentum_flux_Wv(i, j, k, ibg::GFIBG, args...) = conditional_flux_cff(i, j, k, ibg, advective_momentum_flux_Wv, args...)
+@inline _advective_momentum_flux_Uv(i, j, k, ibg::GFIBG, closure, disc::ATD, clock, U, args...) = conditional_flux_ffc(i, j, k, ibg, advective_momentum_flux_Uv, closure, disc::ATD, clock, U, args...)
+@inline _advective_momentum_flux_Vv(i, j, k, ibg::GFIBG, closure, disc::ATD, clock, U, args...) = conditional_flux_ccc(i, j, k, ibg, advective_momentum_flux_Vv, closure, disc::ATD, clock, U, args...)
```
simone-silvestri, line 107: Do you need `closure` and `disc` in the advective fluxes?

src/ImmersedBoundaries/conditional_fluxes.jl
simone-silvestri: hmmm weird, this should call 
`advective_momentum_flux_Wu(i, j, k, grid, scheme, W, u)` (so I guess your `closure` here is the `scheme`) or zero if on a `solid_interface`

tomchor: Thanks for the help, @simone-silvestri. It's much appreciated. Here are some comments:

> conditional_flux should work both for viscous_flux and advective_flux. So every specific argument to one of the two should go after the function specification, just i, j, k, grid before

The reason these arguments are specified is because we need `U` to be passed to the drag function as well. (@whitleyv can you confirm?)

And yes, you're right about the drag right now being counted twice. My plan was to deal with this after I got everything working, but your comment made me realize that separating `conditional_flux` into two will probably help with implementation right now since I can pick one whose arguments will be passed as `args...` :)

src/ImmersedBoundaries/conditional_fluxes.jl
tomchor: You're right, I don't. Just removed it

tomchor:
With this last commit the bottom drag should be working for `u` at least, which shows that the function calls are working. Expanding it to other components and directions is now straightforward. 

Here's a video comparing this IBM with drag BC with two non IBM models (one with no-slip BC, and another with drag BC):

https://user-images.githubusercontent.com/13205162/159105365-7ace908b-6617-4d43-abdd-8d8eb201f06b.mp4

src/ImmersedBoundaries/conditional_fluxes.jl
```diff
@@ -1,37 +1,93 @@
...
+####
+#### Drag functions
+####
+
+const κVK = 0.4 # van Karman's const
+const z₀ = 0.02 # roughness length (meters), user defined in future?
+
+@inline west_drag_const(i, j, k, grid)   = @inbounds -(κVK / log(0.5 * Δxᶠᶜᶜ(i, j, k, grid) / z₀))^2 
+@inline south_drag_const(i, j, k, grid)  = @inbounds -(κVK / log(0.5 * Δyᶜᶠᶜ(i, j, k, grid) / z₀))^2 
+@inline bottom_drag_const(i, j, k, grid) = @inbounds -(κVK / log(0.5 * Δzᶜᶜᶠ(i, j, k, grid) / z₀))^2
```
tomchor, line 20: In to be accurate with the constant for stretched grids I think we need to substute `Δzᶜᶜᶠ(i, j, k, grid)` with something like

```julia
znodes(Center, grid)[k] - znodes(Face, grid)[k]
```
Is there a preferred way to make that work for GPUs? (I think ideally we'd compute this ahead of time instead of on the fly since the constant isn't changing...)

francispoulin:
> With this last commit the bottom drag should be working for `u` at least, which shows that the function calls are working. Expanding it to other components and directions is now straightforward.
> 
> Here's a video comparing this IBM with drag BC with two non IBM models (one with no-slip BC, and another with drag BC):
> 
>  u_drag.mp4

So the blue and dashed green curves should coincide, which they do for most of the domain, but there's the begging part of the dashed curve that doesn't seem to have blue.  Is that a plotting issue or is there something I'm missing?

tomchor:
> Thanks @tomchor , and very nice.
> 
> What would the difference of those two curves look like in the region where there is no masking? Are they identicial or just similar?

In the simulations without the IBM that region does't exist! The domain in those cases starts at `z=0`. And in the simulation with IBM that region exists, but lies inside the immersed solid.

francispoulin:
> > Thanks @tomchor , and very nice.
> > What would the difference of those two curves look like in the region where there is no masking? Are they identicial or just similar?
> 
> In the simulations without the IBM that region does't exist! The domain in those cases starts at `z=0`. And in the simulation with IBM that region exists, but lies inside the immersed solid.

I meant in the physical region.  The curves look like they overlap but if you plotted the difference you could see how much they differ in the physical domain.  Just an idea.

glwagner:
I'm curious whether anyone has looked at whether this drag produces noticably different momentum loss for steppy / sloping boundaries? 

https://www.tandfonline.com/doi/pdf/10.3402/tellusa.v50i1.14514 found that pressure-induced form drag overwhelms viscous drag over steppy coastlines in large-scale models. I've also found that steppy boundaries introduce a ton of form drag in some idealized simulations. But I'm not sure whether that form drag is comparable, more than, or negligible compared to what we get from viscous drag like the one implemented here.

src/ImmersedBoundaries/conditional_fluxes.jl
whitleyv: Yes. There might be an easier way to get those fields, but if I didn't have all the args written out up to the velocity, it wouldn't recognize it as the velocity needed in the drag function down the line.

glwagner:
@tomchor @whitleyv what's the ultimate vision for this PR? @simone-silvestri and I discussed a method for adding immersed fluxes that would introduce a new term into the tendency kernels, eg here:

https://github.com/CliMA/Oceananigans.jl/blob/30ec599de5e3813d57b46e8d8fa4196bcdd250e0/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl#L44

adding something like

```julia
+ ∂ⱼ_τ₁ⱼ_immersed(i, j, k, grid, closure, diffusivities, immersed_bc, velocities, tracers, clock, buoyancy)
```

This would allow the `immersed` boundary condition specified by the user to make its way into the tendency calculation. We think this might be simpler (in terms of the code) than passing `immersed_bc` into ` ∂ⱼ_τ₁ⱼ` (where we would throw it away unless we are on `ImmersedBoundaryGrid`.)

Another advantage is that we can then isolate the code that calculates flux divergences for `Value` and `Gradient` (eg for no-slip boundary conditions), which could get a little complicated.

I'm also trying to figure out if that would cover this case. I think we can use `ContinuousBoundaryCondition` for this kind of drag boundary condition. So in the user script we'd write something like

```julia
u_drag(x, y, z, t, u, v, w, Cd) = - Cd * u * sqrt(u^2 + v^2 + w^2)
v_drag(x, y, z, t, u, v, w, Cd) = - Cd * v * sqrt(u^2 + v^2 + w^2)
w_drag(x, y, z, t, u, v, w, Cd) = - Cd * w * sqrt(u^2 + v^2 + w^2)

u_immersed_bc = FluxBoundaryCondition(u_drag, parameters=Cd, field_dependencies=(:u, :v, :w))
v_immersed_bc = FluxBoundaryCondition(v_drag, parameters=Cd, field_dependencies=(:u, :v, :w))
w_immersed_bc = FluxBoundaryCondition(w_drag, parameters=Cd, field_dependencies=(:u, :v, :w))
```

and then under the hood, if we implement fluxes across immersed boundaries correctly, we'll get the right thing? Note the above code assumes that fluxes are along the "inward facing normal" --- one of the complications of the user API for this...

The corresponding code for no-slip would be

```julia
u_immersed_bc = ValueBoundaryCondition(0)
u_immersed_bc = ValueBoundaryCondition(0)

**Comment by whitleyv** [comment]
> @tomchor @whitleyv what's the ultimate vision for this PR

I think we just want to add a more "precise" drag option to IB by using the `conditional_flux` functions, so that a drag can be added at the exact locations that need it for the grid fitted version, while also avoiding the user having to call the correct conditional for each location. (For context, when adding drag to IB I have been just adding drag as a forcing over a cell-width above the boundary. So I'd prefer something a little more exact.) Hard coding this option in is mostly a test to see how it would go while using the available framework, before creating a full method that users would be able to implement nicely. 

Implementing immersed boundary conditions as a separate term would work, I believe! A `value` immersed option wouldn't involve as much effort (as we've previously worked out for no slip), besides getting the direction correct. I think, one of our concerns was the interpolating of field dependencies within a functional form of a boundary condition, and how they would interact with the IB, could it be generalized to any input.

**Comment by glwagner** [comment]
> > @tomchor @whitleyv what's the ultimate vision for this PR
> 
> I think we just want to add a more "precise" drag option to IB by using the `conditional_flux` functions, so that a drag can be added at the exact locations that need it for the grid fitted version, while also avoiding the user having to call the correct conditional for each location. (For context, when adding drag to IB I have been just adding drag as a forcing over a cell-width above the boundary.

For sure! A forcing function could serve as a prototype for the new tendency kernel function that I'm proposing, right? That was my original motivation for proposing the forcing function --- because we would be able to re-use that code (eg all the right conditionals, etc) when we went to implement immersed boundary conditions in the source code. The new tendency kernel function that @simone-silvestri and I are suggesting would be identical to such a forcing function.

> I think, one of our concerns was the interpolating of field dependencies within a functional form of a boundary condition, and how they would interact with the IB, could it be generalized to any input.

That's true we have to take care with interpolation. I think the first question is what user API we'd like to support; then we can generate source code to match.

**Comment by whitleyv** [comment]
We have fully implemented a hard coded drag into all of the momentum viscous fluxes. Since this implementation of the drag has different values for "symmetric" fluxes such as `viscous_flux_uy` and `viscous_flux_vx` that are located at the same point, the `conditional_flux_fcc` location-separated fluxes were not enough. We split these into functions such as `conditional_flux_fcc_uy` and `conditional_flux_fcc_vx`. It'll be easier to address these issues directly in the immersed `viscous_flux` once we have a separate term in the tendency equations with its own functions. 

These functions then check whether the interface between solid and fluid is at the top of a cell or the bottom of a cell, applying the correctly directioned drag for that side. The way these conditionals work, the fully immersed cells that aren't on a boundary would be calculating fluxes as usual, except that all those terms are masked. So the fluxes inside the solid will remain zero.

When the solid-fluid interface is at the top or right hand side of a cell, the velocities for that `(i,j,k)` location will all be within the solid. If you use them in your drag functions, it would just be zero flux. So, the drag at a top flux interface, for instance, located at `(i,j,k)` will use the velocities at `(i,j,k-1)`, and so on for the other directions. This makes them match the usual non-immersed drag BC for the 1D case as seen below. 

https://user-images.githubusercontent.com/67593861/160839128-37ab95fa-f28b-43c0-8bbf-3c850ae754f2.mp4

https://user-images.githubusercontent.com/67593861/160839127-912ee3d0-75e9-4536-a494-2ec3c6c2b006.mp4

https://user-images.githubusercontent.com/67593861/160839125-bef17105-d8ac-4999-a3f0-fbbb55e8498a.mp4

@tomchor has also tested them with 3D setups, which worked as well!   There's still a lot of room for streamlining and speeding up this set up, which I believe will be easier to do in the full implementation.

**Comment by tomchor** [comment]
Okay so there's some bad news: this IBM implementation seems to be pretty slow. 

I ran two identical simulations with the same number of points on a GPU, but using a regular grid on one and an immersed boundary grid on the other (with the immersed solid boundaries set to coincide with the domain walls). Apparently for a 16×4×4 grid the IBM version is about 10 times slower than the non IBM one. For a 128×32×32 the factor is of about 8 times , so it doesn't appear to be getting better with increasing grid sizes. The difference seems to be similar (albeit a bit smaller) for CPUs.

I'm not sure of what the best way is to speed things up. We could try changing how we implement the if-else clauses (right now we're using nested `ifelse()` functions, which always evaluates all arguments). 

Alternatively instead of testing whether or not we're inside or outside the solid (in addition to fluid-solid boundaries) every time we calculate fluxes, we could store the indices that correspond to solid boundaries and solid interior points and IBM launch kernels only on those indices (although I'm not sure how to achieve that for GPUs).

**Comment by simone-silvestri** [comment]
Isn't this because of the implicit solve? What grid/model combination are you using?

**Comment by glwagner** [comment]
> Isn't this because of the implicit solve? What grid/model combination are you using?
> 
> You can try testing again using a matrix solver for the non immersed model (if you are using the hydrostatic model). The timings should be similar.

They are probably using the FFT-based solver + `NonhydrostaticModel`.

It might be worthwhile benchmarking this problem on `main`. That's a more direct measurement of the cost of the specific changes in this PR.

I also suggest setting up a test that evaluates whether the drag boundary condition produces a measurable difference in a simulation that has a steppy boundary (like a constant slope, rather than a flat wall). You want to determine the relative contribution of this explicit drag compared to form drag for your setup of interest before investing lots of effort here.

**Comment by tomchor** [comment]
> > Isn't this because of the implicit solve? What grid/model combination are you using?
> > You can try testing again using a matrix solver for the non immersed model (if you are using the hydrostatic model). The timings should be similar.
> 
> They are probably using the FFT-based solver + `NonhydrostaticModel`.

Exactly.

> 
> It might be worthwhile benchmarking this problem on `main` with immersed boundaries but no drag. That's a more direct measurement of the cost of the drag law implemented here over no drag law, but still immersed boundary.

We can do that, but I'm keeping this branch up-to-date with `main` for now so (at least for now) it won't change results.


> I also suggest setting up a test that evaluates whether the drag boundary condition produces a measurable difference in a simulation that has a steppy boundary (like a constant slope, rather than a flat wall). You want to determine the relative contribution of this explicit drag compared to form drag for your setup of interest before investing lots of effort here.

That's a good point. Although I'd have to discuss this with @whitleyv since she's the main developer here. My thinking is that if indeed the form drag dominates, we can probably upgrade this to a partial cell grid-fitted IBM (much like @francispoulin intends to do in https://github.com/CliMA/Oceananigans.jl/pull/2306). If my understanding is correct, you'd only need to change the definition of the volumes and areas to account for the partial step and it seems to significantly decrease form drag.

File: src/ImmersedBoundaries/conditional_fluxes.jl [hunk 4]
```diff
@@ -1,37 +1,127 @@
...
+####
+#### Drag functions
+####
+
+const κᵥₖ = 0.4 # van Karman's const
+const z₀ = 0.02 # roughness length (meters), user defined in future?
+
+@inline west_drag_const(i, j, k, grid)   = @inbounds -(κᵥₖ / log(0.5 * Δxᶜᶜᶜ(i, j, k, grid) / z₀))^2 
+@inline south_drag_const(i, j, k, grid)  = @inbounds -(κᵥₖ / log(0.5 * Δyᶜᶜᶜ(i, j, k, grid) / z₀))^2 
+@inline bottom_drag_const(i, j, k, grid) = @inbounds -(κᵥₖ / log(0.5 * Δzᶜᶜᶜ(i, j, k, grid) / z₀))^2
```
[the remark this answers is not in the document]
whitleyv: That seems like a good idea!

glwagner:
> > > Isn't this because of the implicit solve? What grid/model combination are you using?
> > > You can try testing again using a matrix solver for the non immersed model (if you are using the hydrostatic model). The timings should be similar.
> > 
> > 
> > They are probably using the FFT-based solver + `NonhydrostaticModel`.
> 
> Exactly.
> 
> > It might be worthwhile benchmarking this problem on `main` with immersed boundaries but no drag. That's a more direct measurement of the cost of the drag law implemented here over no drag law, but still immersed boundary.
> 
> We can do that, but I'm keeping this branch up-to-date with `main` for now so (at least for now) it won't change results.
> 
> > I also suggest setting up a test that evaluates whether the drag boundary condition produces a measurable difference in a simulation that has a steppy boundary (like a constant slope, rather than a flat wall). You want to determine the relative contribution of this explicit drag compared to form drag for your setup of interest before investing lots of effort here.
> 
> That's a good point. Although I'd have to discuss this with @whitleyv since she's the main developer here. My thinking is that if indeed the form drag dominates, we can probably upgrade this to a partial cell grid-fitted IBM (much like @francispoulin intends to do in #2306). If my understanding is correct, you'd only need to change the definition of the volumes and areas to account for the partial step and it seems to significantly decrease form drag.

Oh for sure, and hopefully we will also eventually have shaved cells which will reduce form drag even more. But you could also calculate that by the time we have partial cells working well with the FFT-based nonhydrostatic solver, et cetera, we might also have a more general implementation for boundary fluxes.

But from a science point of view, if we simply need "some drag" to produce qualitatively useful results, then relying on form drag for `GridFittedBottom` is a simple and effective strategy to make initial progress. If we find that the numerical value of the drag exerts a strong control on the simulation results --- warranting an effort to implement a high-quality drag model, then the next question is whether the form drag with `GridFittedBottom` is _less_ than what we think is realistic. If there is indeed too little drag then we have good justification for hard coding a drag boundary condition into `conditional_flux`.

On the other hand, if there's too much drag, then the first thing to work on is an alternative representation of bathymetry. We can then repeat the form drag analysis with the new bathymetry representation.

whitleyv:
> @whitleyv Should we close this PR? I think some parts of it already got used by @glwagner's PR that got merged and the rest of it is now obsolete (due to the same PR), no?

Sounds good!

tomchor:
That region doesn't exist in the domain without IBM.

The fluid region starts at z=0 in all cases, but for the IBM case, z<0 is
inside the immersed solid, while for the cases without IBM functionality
it's simply outside the domain.

PR #2283, opened 2022-02-28
Small adjustment to Turbulence Closures section of docs

tomchor:
I believe the [Physics/Turbulence Closures](https://clima.github.io/OceananigansDocumentation/stable/physics/turbulence_closures/) section of docs had a minor inconsistency caused by a minus sign.

Also (I haven't changed this yet because I haven't had the time to check in the code) I believe there might be a factor of 2 missing from the τ_ij equation (not numbered), which is the first Eq in [this subsection](https://clima.github.io/OceananigansDocumentation/stable/physics/turbulence_closures/#Smagorinsky-Lilly-turbulence-closure)). 

Other pairs of eyes here would be appreciated, but if not I'll double check everything at a later date.

### DIFF
--- a/docs/src/physics/turbulence_closures.md
+++ b/docs/src/physics/turbulence_closures.md
@@ -66,7 +66,7 @@ Each tracer may have a unique diffusivity components ``\kappa_h`` and ``\kappa_v
 In the turbulence closure proposed by [Lilly62](@cite) and [Smagorinsky63](@cite),
 the subgrid stress associated with unresolved turbulent motions is modeled diffusively via
 ```math
-\tau_{ij} = \nu_e \Sigma_{ij} \, ,
+\tau_{ij} = - \nu_e \Sigma_{ij} \, ,
 ```
 where ``\Sigma_{ij} = \tfrac{1}{2} \left ( v_{i, j} + v_{j, i} \right )`` is the resolved
 strain rate.
@@ -97,7 +97,7 @@ where ``\Delta x``, ``\Delta y``, and ``\Delta z`` are the grid spacing in the
 
 The effect of subgrid turbulence on tracer mixing is also modeled diffusively via
 ```math
-\boldsymbol{q}_c = \kappa_e \boldsymbol{\nabla} c \, ,
+\boldsymbol{q}_c = - \kappa_e \boldsymbol{\nabla} c \, ,
 ```
 where the eddy diffusivity ``\kappa_e`` is
 ```math
@@ -124,7 +124,7 @@ constant isotropic background viscosity and diffusivities for each tracer. The e
 predictor is
 ```math
     \begin{equation}
-    \nu_e^\dagger = -C \Delta_f^2
+    \nu_e^\dagger = C \Delta_f^2
     \frac
         {(\hat{\partial}_k \hat{v}_i) (\hat{\partial}_k \hat{v}_j) \hat{\Sigma}_{ij}
         + C_b \hat{\delta}_{i3} (\hat{\partial}_k \hat{v_i}) (\hat{\partial}_k b)}
@@ -136,7 +136,7 @@ while the eddy diffusivity predictor for tracer ``c`` is
 ```math
     \begin{equation}
     \label{eq:kappa-dagger}
-    \kappa_e^\dagger = -C \Delta_f^2
+    \kappa_e^\dagger = C \Delta_f^2
     \frac
         {(\hat{\partial}_k \hat{v}_i) (\hat{\partial}_k c) (\hat{\partial}_i c)}
         {(\hat{\partial}_l c) (\hat{\partial}_l c)} \, .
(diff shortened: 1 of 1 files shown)


docs/src/physics/turbulence_closures.md
```diff
@@ -97,7 +97,7 @@ where ``\Delta x``, ``\Delta y``, and ``\Delta z`` are the grid spacing in the
 
 The effect of subgrid turbulence on tracer mixing is also modeled diffusively via
 ```math
-\boldsymbol{q}_c = \kappa_e \boldsymbol{\nabla} c \, ,
+\boldsymbol{q}_c = - \kappa_e \boldsymbol{\nabla} c \, ,
```
navidcy, line 100: this change sounds correct!

docs/src/physics/turbulence_closures.md
```diff
@@ -66,7 +66,7 @@ Each tracer may have a unique diffusivity components ``\kappa_h`` and ``\kappa_v
 In the turbulence closure proposed by [Lilly62](@cite) and [Smagorinsky63](@cite),
 the subgrid stress associated with unresolved turbulent motions is modeled diffusively via
 ```math
-\tau_{ij} = \nu_e \Sigma_{ij} \, ,
+\tau_{ij} = - \nu_e \Sigma_{ij} \, ,
```
navidcy, line 69: this change sounds correct!

navidcy (approved):
Also the factor 2 that you mention seems like it's actually missing since as it stands now:

∂ⱼ τᵢⱼ = ∂ⱼ ½(uᵢ,ⱼ + uⱼ,ᵢ) = ½(uᵢ,ⱼⱼ + (uⱼ,ⱼ),ᵢ) = ½ ∇²uᵢ

docs/src/physics/turbulence_closures.md
glwagner: Most definitely.

docs/src/physics/turbulence_closures.md
glwagner: This is missing a factor `2`?

glwagner, line 69:
```suggestion
\tau_{ij} = - 2 \nu_e \Sigma_{ij} \, ,
```

tomchor: I couldn't find this line in the code yet (although I haven't had much time to look for it yet). I just wanna make sure that the code is programmed that way and that it didn't get absorbed into `ν_e` before making this change in the docs.

tomchor:
I found the calculation of the sgs flux with the factor of 2 for scalar diffusivities here: https://github.com/CliMA/Oceananigans.jl/blob/4f1fe0fa7edd9489ba20875be64b69daa0a77b95/src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl#L57-L74

But I couldn't find similar lines for non-scalar diffusivities. I'm just trying to make sure that the implementation matches the docs and that a factor of 2 didn't get absorved into the definition of some viscosities when coding before merging this.

@glwagner @simone-silvestri @navidcy can any of you point me to the relevant lines for non-scalar viscosities?

simone-silvestri:
@tomchor the relevant fluxes are in the respective files for an `AbstractTurbulenceClosure` (for example `ConvectiveAdjustment` or `IsopycnalSkewSymmetricDiffusivity`, while `AbstractEddyClosures` (for example `AnisotropicMinimumDissipation` or `SmagorinskyLilly` use the functions defined in `AbstractScalarDiffusivity` (for example both `AnisotropicMinimumDissipation` and `SmagorinskyLilly` are only `ThreeDimensional`)

tomchor:
> @tomchor the relevant fluxes are in the respective files for an `AbstractTurbulenceClosure` (like for `ConvectiveAdjustment` or `IsopycnalSkewSymmetricDiffusivity`), while an `AbstractEddyClosure` (for example `AnisotropicMinimumDissipation` or `SmagorinskyLilly`) uses the functions defined in `AbstractScalarDiffusivity` (for example both `AnisotropicMinimumDissipation` and `SmagorinskyLilly` are only `ThreeDimensional`)

Ah, I see. We subtype one with the other here https://github.com/CliMA/Oceananigans.jl/blob/4f1fe0fa7edd9489ba20875be64b69daa0a77b95/src/TurbulenceClosures/abstract_eddy_viscosity_closure.jl#L7

Then in this case I think the code will match the docs after this PR. I'll merge as soon as tests pass.

Just a note, do we wanna consider modifying the subtyping here in the future? The recent changes in the turbulence closures were great, but I'd argue that there's still room for improvement. For example (if I understand correctly) right now `SmagorisnkyLilly` is an `AbstractEddyViscosityClosure`, which itself is a subtype of `AbstractScalarDiffusivity`, even though the diffusivity in a `SmagLilly` isn't a scalar, which is confusing (this is what threw me off at first). Maybe it's worth considering defining an `AbstractViscosityClosure`, and then `AbstractScalarDiffusivity` and `AbstractEddyViscosityClosure` coudl be separate subtypes of it?

I'm happy to open an issue to discuss this more.

glwagner:
The LES closures subtype `AbstractScalarDiffusivity` and I think this is what the docs are referring to, so that section of the code is indeed the correct reference for all relevant cases.

glwagner:
> even though the diffusivity in a SmagLilly isn't a scalar, which is confusing (this is what threw me off at first).

What do you mean?

tomchor:
Sorry my initial comment [here](https://github.com/CliMA/Oceananigans.jl/pull/2283#issuecomment-1055659167) stemmed from a momentary misunderstanding of the terminology! I agree that the subtyping makes sense.

PR #2284, opened 2022-02-28
Fix bug for horizontal diffusivity with Flat horizontal dimension + updates baroclinic adjustment validation script

navidcy:
This PR updates the baroclinic adjustment validations scripts.

Closes #2291

PR #2288, opened 2022-03-01
Fix inconsistent variable definition in `ScalarDiffusivity` constructor

navidcy:
The main contribution of this PR is that it corrects the `time_discretization` positional argument in the `ScalarDiffusivity` constructor. It was sometimes referred to as `time_discretization` and other times as `time_disc`.


simone-silvestri (approved):

glwagner (approved):
Hard work by unsung heros

navidcy:
`test_dynamics.jl` fail.

Is this correct here?
https://github.com/CliMA/Oceananigans.jl/blob/ad110ecdc635192c110e229df4fbfa04fc7a2266/src/TurbulenceClosures/turbulence_closure_implementations/scalar_diffusivity.jl#L39

glwagner:
> `test_dynamics.jl` fail.
> 
> Is this correct here?
> 
> https://github.com/CliMA/Oceananigans.jl/blob/ad110ecdc635192c110e229df4fbfa04fc7a2266/src/TurbulenceClosures/turbulence_closure_implementations/scalar_diffusivity.jl#L39-L41
> 
> I guess it should be `if formulation == VerticallFormulation() and ...`?

No that's correct --- we can't use vertically implicit time-stepping with `HorizontalFormulation`.

The reason is that `HorizontalFormulation` has no vertical terms.

glwagner:
This should probably wait until #2284

navidcy:
> > `test_dynamics.jl` fail.
> > Is this correct here?
> > https://github.com/CliMA/Oceananigans.jl/blob/ad110ecdc635192c110e229df4fbfa04fc7a2266/src/TurbulenceClosures/turbulence_closure_implementations/scalar_diffusivity.jl#L39-L41
> > 
> > I guess it should be `if formulation == VerticallFormulation() and ...`?
> 
> No that's correct --- we can't use vertically implicit time-stepping with `HorizontalFormulation`.
> 
> The reason is that `HorizontalFormulation` has no vertical terms.

Then the error message should be `only supported for `VerticalFormulation or ThreeDimensionalFormulation"`, right?

navidcy:
OK fixed it. Apparently tests were passing before because the `time_discretization` argument was not working properly.

PR #2289, opened 2022-03-01
Better local variable name: `area` -> `grid_metric`

### LINKED ISSUE #2285: Local definition of `area` sometimes is `Δx`?
So here a local variable `area` is defined but `area` could be `Δx`. This is confusing... right?

https://github.com/CliMA/Oceananigans.jl/blob/4f1fe0fa7edd9489ba20875be64b69daa0a77b95/src/Operators/products_between_fields_and_grid_metrics.jl#L8

@simone-silvestri

navidcy:
Closes #2285.


glwagner (approved):
Because

```
metric in (:Δ, :A)
```

francispoulin (approved):

src/Operators/products_between_fields_and_grid_metrics.jl
```diff
@@ -1,14 +1,14 @@
 #####
-##### Operators of the form A * q where A is an area and q is some quantity.
+##### Operators of the form grid_metric * q, where grid_metric is a grid metric and q is some variable
```
glwagner, line 2:
```suggestion
##### Operators of the form grid_metric * q
```

Since "some variable" isn't anymore specific than simply writing "q"

tomchor (approved):

PR #2294, opened 2022-03-01
Use verbose names in `LinearEquationOfState`

### LINKED ISSUE #453: Use verbose names in LinearEquationOfState
Rather than using `α` and `β` to denote the thermal expansion and saline contraction coefficients, we should probably just use `thermal_expansion` and `haline_contraction`. The syntax would then be

```julia
eos = LinearEquationOfState(thermal_expansion=α, haline_contraction=β)
```

glwagner:
This PR changes `α` and `β` in `LinearEquationOfState` to `thermal_expansion` and `haline_contraction`.

Closes #453


francispoulin (approved):
This makes our syntax longer, but much clearer and I would argue friendier.  Good idea!

simone-silvestri:
Don't we have functions with that name?

glwagner:
> Don't we have functions with that name? Maybe it gets confusing when developing?

The names are similar but not identical:

https://github.com/CliMA/Oceananigans.jl/blob/35d93732248e03718086c7f80c3c2d46571f3c6d/src/BuoyancyModels/linear_equation_of_state.jl#L47-L55

It doesn't seem too bad to me (straightforward even?) but let me know what you think, we don't have to merge!

navidcy (approved):

PR #2295, opened 2022-03-01
Better subtyping of `AbstractScalarDiffusivity`

glwagner:
This PR eliminates `AbstractEddyDiffusivity` in favor of simply defining `viscosity` and `diffusivity` directly for the LES closures. It also converts the boundary layer parameterizations to subtype `AbstractScalarDiffusivity{TD, VerticalFormulation}`. This change eliminates much code duplication.


src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -65,7 +65,8 @@ const AVD = AbstractScalarDiffusivity{<:Any, <:VerticalFormulation}
 @inline viscous_flux_vx(i, j, k, grid, closure::AHD, clock, U, args...) = - ν_ζᶠᶠᶜ(i, j, k, grid, clock, closure.ν, U.u, U.v)
 @inline viscous_flux_uy(i, j, k, grid, closure::AHD, clock, U, args...) = + ν_ζᶠᶠᶜ(i, j, k, grid, clock, closure.ν, U.u, U.v)   
 @inline viscous_flux_vy(i, j, k, grid, closure::AHD, clock, U, args...) = - ν_δᶜᶜᶜ(i, j, k, grid, clock, closure.ν, U.u, U.v)
-# are we sure about this?? This might not be consistent for AHD
+
+# TODO: are we sure about this?? This might not be consistent for AHD
```
simone-silvestri, line 69: I convinced myself that this is, indeed, consistent in case of AHD for highly anisotropic grids, that is the case in which we would use this formulation...

since the vertical fluxes of horizontal velocities do not have horizontal gradients of the vertical velocity, the horizontal flux of vertical velocity should not contain vertical gradients of horizontal velocity...

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -129,25 +130,25 @@ const VITD = VerticallyImplicitTimeDiscretization
...
 @inline function viscous_flux_uz(i, j, k, grid::VerticallyBoundedGrid, ::VITD, closure::Union{AID, AVD}, args...)
-    return ifelse(k == 1 || k == grid.Nz+1, 
+    return ifelse(k == 1 | k == grid.Nz+1, 
                   viscous_flux_uz(i, j, k, grid, ExplicitTimeDiscretization(), closure, args...),
                   ivd_viscous_flux_uz(i, j, k, grid, closure, args...))
 end
 
 @inline function viscous_flux_vz(i, j, k, grid::VerticallyBoundedGrid, ::VITD, closure::Union{AID, AVD}, args...)
-    return ifelse(k == 1 || k == grid.Nz+1, 
+    return ifelse(k == 1 | k == grid.Nz+1,
```
simone-silvestri, line 139:
```suggestion
    return ifelse((k == 1) | (k == grid.Nz+1), 
```

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -129,25 +130,25 @@ const VITD = VerticallyImplicitTimeDiscretization
...
 @inline function viscous_flux_vz(i, j, k, grid::VerticallyBoundedGrid, ::VITD, closure::Union{AID, AVD}, args...)
-    return ifelse(k == 1 || k == grid.Nz+1, 
+    return ifelse(k == 1 | k == grid.Nz+1, 
                   viscous_flux_vz(i, j, k, grid, ExplicitTimeDiscretization(), closure, args...),
                   ivd_viscous_flux_vz(i, j, k, grid, closure, args...))
 end
 
 @inline function viscous_flux_wz(i, j, k, grid::VerticallyBoundedGrid{FT}, ::VITD, closure::Union{AID, AVD}, args...) where FT
-    return ifelse(k == 1 || k == grid.Nz+1, 
+    return ifelse(k == 1 | k == grid.Nz+1,
```
simone-silvestri, line 145:
```suggestion
    return ifelse((k == 1) | (k == grid.Nz+1), 
```

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -129,25 +130,25 @@ const VITD = VerticallyImplicitTimeDiscretization
...
 @inline function viscous_flux_wz(i, j, k, grid::VerticallyBoundedGrid{FT}, ::VITD, closure::Union{AID, AVD}, args...) where FT
-    return ifelse(k == 1 || k == grid.Nz+1, 
+    return ifelse(k == 1 | k == grid.Nz+1, 
                   viscous_flux_wz(i, j, k, grid, ExplicitTimeDiscretization(), closure, args...),
                   zero(FT))
 end
 
 @inline function diffusive_flux_z(i, j, k, grid::VerticallyBoundedGrid{FT}, ::VITD, closure::Union{AID, AVD}, args...) where FT
-    return ifelse(k == 1 || k == grid.Nz+1, 
+    return ifelse(k == 1 | k == grid.Nz+1,
```
simone-silvestri, line 151:
```suggestion
    return ifelse(((k == 1) | (k == grid.Nz+1), 
```

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
glwagner: I think I agree, but what do you mean that it's dependent on grid anisotropy? There's no check for grid anisotropy here?

src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl
```diff
@@ -7,38 +7,32 @@ Parameters for the "anisotropic minimum dissipation" turbulence closure for larg
...
+    function AnisotropicMinimumDissipation{TD}(Cν::PN, Cκ::PK, Cb::PB) where {TD, PN, PK, PB}
+        return new{TD, PK, PN, PB}(Cν, Cκ, Cb)
     end
 end
 
 const AMD = AnisotropicMinimumDissipation
 
-Base.show(io::IO, closure::AMD{TD, FT}) where {TD, FT} =
-    print(io, "AnisotropicMinimumDissipation{$FT} turbulence closure with:\n",
+@inline viscosity(::AMD, diffusivities, args...) = diffusivities.νₑ
```
simone-silvestri, line 22: is this not a `[i, j, k]` field?

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
glwagner: It feels like the issue is that the formulation of some terms really depends on whether or not the model is hydrostatic or not. Coriolis forces are the same; if we are non-hydrostatic we include some terms that we wouldn't otherwise. I think we might need a type to describe whether a model is hydrostatic (or we can use a numeric multiplier like MITgcm)

src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl
glwagner: It is, but this function just returns the _object_; it's indexed into later.

I think we can further simplify the interface...

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
simone-silvestri: No there is not, but when the grid is very anisotropic it is better to use the `Horizontal` formulation that the `ThreeDimensional` formulation.

talking with @jm-c we came to the conclusion that is not desired to have the horizontal flux of vertical velocity dominated by the vertical gradients of horizontal velocity (which might be very large!) which would be the case in terms of a very anisotropic grid.

src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl
glwagner: See here:

https://github.com/CliMA/Oceananigans.jl/blob/cb97515799b6889b955fdc36474e47f97c0a68d1/src/TurbulenceClosures/vertically_implicit_diffusion_solver.jl#L139-L176

We also would use this function for immersed boundary fluxes.

I might try to clean this up now though.

simone-silvestri:
A lot of code simplification, nice!

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
glwagner: But we don't have any information about grid anisotropy here?

glwagner: I'm not sure I agree with the argument.

What we want is an interface that allows users to define the PDE they want to solve numerically. We should definitely avoid psychology / assumptions about what users are trying to do (eg "if users select `HorizontalScalarDiffusivity` then _of course_ they are using an anisotropic grid... right?!?)

If there are two legitimate choices then we need a way to express the difference between the two in our API...

glwagner: Since we've already appended our types with the word "formulation" we can maybe leverage that to add a "horizontal diffusion" that would be appropriate for nearly isotropic grids, right? It's not a high priority...

glwagner: > If the grid is quite isotropic (Lx ≈ Ly ≈ Lz) probably it is better to use the ThreeDimensional formulation as maybe the Horizontal formulation removes an important part of the dissipation. I was maybe worried about that... (probably we should write it somewhere in the docs)

What if we really do want diffusion only in the horizontal, but with an isotropic grid? That is the question.

navidcy:
Merge `main` and resolve conflicts?

glwagner:
> Merge `main` and resolve conflicts?

Yessir! I want to get #2246 in first though.

tomchor:
I didn't quite understand what the bug in AMD was. Was it important?

glwagner:
A previous commit on this PR created a bug in one of the constructors so it wasn't possible to instantiate AMD. The subsequent commit fixed it.

tomchor:
> A previous commit on this PR created a bug in one of the constructors so it wasn't possible to instantiate AMD. The subsequent commit fixed it.

Thanks. I thought there was a bug in AMD since before this PR that I wasn't aware about

glwagner:
One remaining question is a user interface for extracting viscosities and diffusivities, especially when we have closure tuples.

I think we want a function based interface. Take for example the case

```julia
closure = (AnisotropicMinimumDissipation(), ScalarDiffusivity(ν = 1.05e-6, κ=1e-7))
```

We could have some API functions that look like

```julila
viscosity(model) # returns 1.05e-6 + νₑ
diffusivity(model, :b) # returns 1e-7 + κₑ
diffusivities(model) # returns a named tuple with all tracer diffusivities
```

This could work if _all_ closures are _either_ `AbstractScalarDiffusivity` or `AbstractBiharmonicDiffusivity`.

For models with heterogeneous closures we need a bit more. For example with

```julia
closure = (ScalarBiharmonicDiffusivity(ν = 1e2, κ=1e1), CATKEVerticalDiffusivity())
```

we might have something like

```julia
viscosity(closure[1], model)
viscosity(closure[2], model)
```

which returns the viscosity associated with `closure[n]` (here, either biharmonic diffusivity or CATKE's nonlinear diffusivity).

The issue is that we don't want users picking apart `model.diffusivity_fields` since this will make scripts more fragile, prone to breakage when code changes, and difficult to understand. (I also think we need to change this to `closure_fields` at some point). 

Thoughts or other ideas?

glwagner:
> In addition, I am convinced of the simplicity of keeping the `diffusivity_fields`, nesting too many structs might make code too complex (thinking also at future development of `MultiRegion`)

Ok to clarify you're saying we _should keep_ `diffusivity_fields` / `closure_fields`?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/CATKEVerticalDiffusivities.jl
```diff
@@ -199,141 +198,18 @@ end
...
-
-const VerticallyBoundedGrid{FT} = AbstractGrid{FT, <:Any, <:Any, <:Bounded}
-
-@inline diffusive_flux_z(i, j, k, grid, ::VITD, closure::CATKEVD, args...) = zero(eltype(grid))
-@inline viscous_flux_uz(i, j, k, grid, ::VITD, closure::CATKEVD, args...) = zero(eltype(grid))
-@inline viscous_flux_vz(i, j, k, grid, ::VITD, closure::CATKEVD, args...) = zero(eltype(grid))
-
-@inline function diffusive_flux_z(i, j, k, grid::VerticallyBoundedGrid{FT}, ::VITD, closure::CATKEVD, args...) where FT
-    return ifelse(k == 1 || k == grid.Nz+1, # on boundaries, calculate fluxes explicitly
-                  diffusive_flux_z(i, j, k, grid, ExplicitTimeDiscretization(), closure, args...),
```
simone-silvestri, line 307: :boom:

src/TurbulenceClosures/turbulence_closure_implementations/convective_adjustment_vertical_diffusivity.jl
```diff
@@ -147,90 +151,6 @@ end
...
-
-@inline function viscous_flux_vz(i, j, k, grid, closure::CAVD, clock, velocities, diffusivities, args...)
-    ν = νᶜᶠᶠ(i, j, k, grid, clock, diffusivities.ν)
-    return - ν * ∂zᶜᶠᶠ(i, j, k, grid, velocities.v)
-end
-
-@inline function viscous_flux_wz(i, j, k, grid, closure::CAVD, clock, velocities, diffusivities, args...)
-    ν = νᶜᶜᶜ(i, j, k, grid, clock, diffusivities.ν)
-    return - ν * ∂zᶜᶜᶜ(i, j, k, grid, velocities.w)
-end
```
simone-silvestri, line 232: :boom: :boom:

navidcy:
@glwagner is this ready to merge? I believe so. If @simone-silvestri likes it I'll merge it ;)

src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
```diff
@@ -17,38 +32,38 @@ const AIBD = AbstractScalarBiharmonicDiffusivity{<:ThreeDimensionalFormulation}
...
+@inline viscous_flux_vx(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶠᶠᶜ(i, j, k, grid, clo, K, clk, biharmonic_mask_x, ∂xᶠᶠᶜ, ∇²ᶜᶠᶜ, U.v)
+@inline viscous_flux_wx(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶠᶜᶠ(i, j, k, grid, clo, K, clk, biharmonic_mask_x, ∂xᶠᶜᶠ, ∇²ᶜᶜᶠ, U.w)
+@inline viscous_flux_uy(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶠᶠᶜ(i, j, k, grid, clo, K, clk, biharmonic_mask_y, ∂yᶠᶠᶜ, ∇²ᶠᶜᶜ, U.u)
+@inline viscous_flux_vy(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, ∂yᶜᶜᶜ, biharmonic_mask_y, ∇²ᶜᶠᶜ, U.v)
+@inline viscous_flux_wy(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶜᶠᶠ(i, j, k, grid, clo, K, clk, biharmonic_mask_y, ∂yᶜᶠᶠ, ∇²ᶜᶜᶠ, U.w)
+@inline viscous_flux_uz(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶠᶜᶠ(i, j, k, grid, clo, K, clk, biharmonic_mask_z, ∂zᶠᶜᶠ, ∇²ᶠᶜᶜ, U.u)
+@inline viscous_flux_vz(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶜᶠᶠ(i, j, k, grid, clo, K, clk, biharmonic_mask_z, ∂zᶜᶠᶠ, ∇²ᶜᶠᶜ, U.v)
+@inline viscous_flux_wz(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, ∂zᶜᶜᶜ, biharmonic_mask_z, ∇²ᶜᶜᶠ, U.w)
+
```
navidcy, line 44:
```suggestion

```

src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
```diff
@@ -17,38 +32,38 @@ const AIBD = AbstractScalarBiharmonicDiffusivity{<:ThreeDimensionalFormulation}
...
+@inline viscous_flux_wz(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, ∂zᶜᶜᶜ, biharmonic_mask_z, ∇²ᶜᶜᶠ, U.w)
+                                                                                                                      
+@inline viscous_flux_ux(i, j, k, grid, clo::AHBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, δ★ᶜᶜᶜ, U.u, U.v)   
+@inline viscous_flux_vx(i, j, k, grid, clo::AHBD, K, U, C, clk, b) = + ν_σᶠᶠᶜ(i, j, k, grid, clo, K, clk, ζ★ᶠᶠᶜ, U.u, U.v)
+@inline viscous_flux_wx(i, j, k, grid, clo::AHBD, K, U, C, clk, b) = + ν_σᶠᶜᶠ(i, j, k, grid, clo, K, clk, biharmonic_mask_x, ∂xᶠᶜᶠ, ∇²ᶜᶜᶠ, U.w)
+@inline viscous_flux_uy(i, j, k, grid, clo::AHBD, K, U, C, clk, b) = - ν_σᶠᶠᶜ(i, j, k, grid, clo, K, clk, ζ★ᶠᶠᶜ, U.u, U.v)   
+@inline viscous_flux_vy(i, j, k, grid, clo::AHBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, δ★ᶜᶜᶜ, U.u, U.v)
+@inline viscous_flux_wy(i, j, k, grid, clo::AHBD, K, U, C, clk, b) = + ν_σᶜᶠᶠ(i, j, k, grid, clo, K, clk, biharmonic_mask_y, ∂yᶜᶠᶠ, ∇²ᶜᶜᶠ, U.w)
+
```
navidcy, line 51:
```suggestion

```

src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
```diff
@@ -17,38 +32,38 @@ const AIBD = AbstractScalarBiharmonicDiffusivity{<:ThreeDimensionalFormulation}
...
-
-@inline diffusive_flux_x(i, j, k, grid, closure::AHBD, c, ::Val{tracer_index}, clock, args...) where tracer_index = κ_σᶠᶜᶜ(i, j, k, grid, clock, closure.κ[tracer_index], biharmonic_mask_x, ∂x_∇²h_cᶠᶜᶜ, c)
-@inline diffusive_flux_y(i, j, k, grid, closure::AHBD, c, ::Val{tracer_index}, clock, args...) where tracer_index = κ_σᶜᶠᶜ(i, j, k, grid, clock, closure.κ[tracer_index], biharmonic_mask_y, ∂y_∇²h_cᶜᶠᶜ, c)
-@inline diffusive_flux_z(i, j, k, grid, closure::AVBD, c, ::Val{tracer_index}, clock, args...) where tracer_index = κ_σᶜᶜᶠ(i, j, k, grid, clock, closure.κ[tracer_index], biharmonic_mask_z, ∂³zᶜᶜᶠ, c)
+@inline diffusive_flux_x(i, j, k, grid, clo::AIBD, K, ::Val{id}, U, C, clk, b) where id = κ_σᶠᶜᶜ(i, j, k, grid, clo, K, Val(id), clk, biharmonic_mask_x, ∂xᶠᶜᶜ, ∇²ᶜᶜᶜ, C[id])
+@inline diffusive_flux_y(i, j, k, grid, clo::AIBD, K, ::Val{id}, U, C, clk, b) where id = κ_σᶜᶠᶜ(i, j, k, grid, clo, K, Val(id), clk, biharmonic_mask_y, ∂yᶜᶠᶜ, ∇²ᶜᶜᶜ, C[id])
+@inline diffusive_flux_z(i, j, k, grid, clo::AIBD, K, ::Val{id}, U, C, clk, b) where id = κ_σᶜᶜᶠ(i, j, k, grid, clo, K, Val(id), clk, biharmonic_mask_z, ∂zᶜᶜᶠ, ∇²ᶜᶜᶜ, C[id])
+
```
navidcy, line 63:
```suggestion

```

glwagner:
Please do merge

tomchor (approved):

PR #2296, opened 2022-03-02
Update friction velocity definition in `stratified_couette_flow.jl`

### LINKED ISSUE #597: Typo in stratified Couette flow verification  experiment
@jm-c also pointed out that I forgot to update the indices in computing the friction velocity in `stratified_couette_flow.png` after the k index was reversed.

glwagner:
Resolves #597

### DIFF
--- a/validation/stratified_couette_flow/stratified_couette_flow.jl
+++ b/validation/stratified_couette_flow/stratified_couette_flow.jl
@@ -15,7 +15,7 @@ function uτ(model, Uavg, U_wall)
     compute!(Uavg)
     U = Array(interior(Uavg))  # Exclude average of halo region.
 
-    # Use a finite difference to calculate dU/dz at the top and bottomtom walls.
+    # Use a finite difference to calculate dU/dz at the top and bottom walls.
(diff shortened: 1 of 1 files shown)


validation/stratified_couette_flow/stratified_couette_flow.jl
```diff
@@ -18,8 +18,8 @@ function uτ(model, Uavg, U_wall)
     # Use a finite difference to calculate dU/dz at the top and bottomtom walls.
```
francispoulin, line 18: Fix typo?  `bottomtom`

francispoulin (approved):
As long as `U[Nz]` is adjacent to the top wall and `U[1]` is adjacent to the bottom wall, then I am happy.

validation/stratified_couette_flow/stratified_couette_flow.jl
glwagner:
```suggestion
    # Use a finite difference to calculate dU/dz at the top and bottom walls.
```

PR #2306, opened 2022-03-03
Partial cell immersed boundaries

francispoulin:
This creates a partial cell immersed boundary method, following [Adroft et al. (1997)](https://journals.ametsoc.org/view/journals/mwre/125/9/1520-0493_1997_125_2293_rotbsc_2.0.co_2.xml).

Things to be done still:

- [ ] test for tracer advection (done but needs to be added to tests)
- [ ] test with momentum integrtion, both in rectilinear and lat lon grids (needs to be done)


src/ImmersedBoundaries/partial_cell_immersed_boundaries.jl
```diff
@@ -0,0 +1,48 @@
...
+
+const PCIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:PartialCellBottom}
+
+bottom_cell(i, j, k, ibg::PCIBG) = !is_immersed(i, j, k, ibg.grid, ibg.immersed_boundary) & is_immersed(i, j, k-1, ibg.grid, ibg.immersed_boundary)
+
+@inline function Δzᵃᵃᶜ(i, j, k, ibg::PCIBG)
+    x, y, z = node(c, c, c, i, j, k, underlying_grid)
+    Δzᵃᵃᶜ = ibg.immersed_boundary.bottom_height(x, y)
+    return Δzᵃᵃᶜ
+end
```
glwagner, line 48: We need to add the other `Δz` metrics, right?

glwagner, line 48:
```suggestion
@inline function Δzᶜᶜᶜ(i, j, k, ibg::PCIBG)
    x, y, z = node(c, c, c, i, j, k, underlying_grid)
    Δzᶜᶜᶜ = ibg.immersed_boundary.bottom_height(x, y)
    return Δzᶜᶜᶜ
end
```

francispoulin: Yes, you are correct.  This works for tracer advection but we need more metrics for momentum advection.

glwagner: I may have been a little hasty here. We need to check `minimum_fractional_partial_Δz` right?

Do you mind if I commit to this PR?

glwagner: I don't think it works correctly for tracer advection either, because the side-walls of the tracer cells are not correct?

glwagner:
@francispoulin do you mind if I commit to this PR?

francispoulin:
Yes, please do!

glwagner:
The discussion on #2410 reminded me: to complete this implementation, we are going to need to make some modification to the hydrostatic pressure integral.

There are two possibilities. We can either

1) use the MITgcm method, which _avoids_ using the modified grid spacing in the bottom cell for the hydrostatic pressure integral. Using "underlying" grid spacing avoids pressure gradient errors that would otherwise induce flow even for horizontally uniform density stratification.

2) Implement Lin 1997 to evaluate the hydrostatic pressure integral: https://rmets.onlinelibrary.wiley.com/doi/abs/10.1002/qj.49712354214?casa_token=_4v_YF-yx0cAAAAA:vC9wnmpvE_2HqoKudORNBfVCkxK4_YgRfp5_Lh2hIFFNc0ZyST-D7nwEqt1No4FculdvrohhKVF2wT9r

francispoulin:
Quickly looking at Lin 1997, they compare the results of their method with a finite volume (FV) method.  Their method does better.  I wonder if their FV method is at all similar to what the MITgcm uses?  

Is it fair to say the first is easier and the second has the potential to be more accurate?

glwagner:
> Quickly looking at Lin 1997, they compare the results of their method with a finite volume (FV) method. Their method does better. I wonder if their FV method is at all similar to what the MITgcm uses?
> 
> Is it fair to say the first is easier and the second has the potential to be more accurate?

MITgcm does not use Lin 1997. Instead, they use vertical spacing from the _underlying_ grid (not the partial cell grid) to calculate the hydrostatic pressure. This somehow reduces the accuracy of the hydrostatic pressure calculation, but results in an approximation that does not have spurious pressure gradient errors. It's simple and effective.

Lin 1997 _is_ a finite volume method.

francispoulin:
Sorry that I was unclear.  Yes, Lin's method is a finite volume method.

In Lin's paper he has plots his results and compares his method with another method, which he titles finite volume.  I know that his method is a specialized finite volume method.

If Lin's approach has the potential to be more accurate than the method used by the MITgcm, it sounds like a fruitful direction to try.

glwagner:
> Sorry that I was unclear. Yes, Lin's method is a finite volume method.
> 
> In Lin's paper he has plots his results and compares his method with another method, which he titles finite volume. I know that his method is a specialized finite volume method.
> 
> If Lin's approach has the potential to be more accurate than the method used by the MITgcm, it sounds like a fruitful direction to try.

Ah sorry! I don't know what that other method is. One major advantage of Lin 1997 aside from accuracy is that it may work with shaved cells and perhaps (mild) generalized vertical coordinates (though I suppose we can continue to use the "MITgcm method" with shaved cells too).

francispoulin:
It sounds like this method is certainly worth trying to see how it does compared to other approaches.

glwagner:
First results from the "stratified ocean remains at rest" validation:

image

Short answer --- with partial cells, a resting ocean does not remain at rest! So, we need to implement either the "MITgcm method" for hydrostatic pressure, or Lin 1997 (or both).

src/ImmersedBoundaries/partial_cell_immersed_boundaries.jl
```diff
@@ -0,0 +1,91 @@
...
+
+const PCIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:PartialCellBottom}
+
+bottom_cell(i, j, k, ibg::PCIBG) = !is_immersed(i, j, k, ibg.grid, ibg.immersed_boundary) & is_immersed(i, j, k-1, ibg.grid, ibg.immersed_boundary)
+
+@inline function Δzᶜᶜᶜ(i, j, k, ibg::PCIBG)
+    underlying_grid = ibg.grid
+    ib = ibg.immersed_boundary
+    # Get node at face above
+    x, y, z = node(c, c, c, i, j, k+1, underlying_grid)
```
glwagner, line 59: @fadaie91 this is the line you believe has a typo right? Can you paste your sketch into this PR?

francispoulin: @fadaie91 : Can you confirm we are proposing that in line 59, we should have '(c, c, f)'?  

That would certainly work with the comment but it's been a while so I might be misremembering.

fadaie91: Yes we should have (c, c, f)

glwagner:
As of https://github.com/CliMA/Oceananigans.jl/pull/2306/commits/c3e276fc6ac2abffcdb73727258dddd050d08cc2 we have, with `minimum_fractional_partial_Δz=1.0` (in this limit, `PartialCellBottom` and `GridFittedBottom` should be identical):

image

But with `minimum_fractional_partial_Δz=0.2`:

image

so there's a bug!

glwagner:
Ok, fixed an issue dispatching on a grid with partial cells in the hydrostatic pressure integral (turns out things go very wrong if we naively integrate to find hydrostatic pressure with partial cells!) here: https://github.com/CliMA/Oceananigans.jl/pull/2306/commits/9cc91dafe140ce4d5124a4d80e8a35496e45ad7e

Things look ok now, and a resting ocean remains at rest:

image

glwagner:
@fadaie91 is going to run some more validation tests, but I think this PR is basically ready to go.

@sandreza this may interest you as well for your 1 deg simulation.

src/ImmersedBoundaries/partial_cell_immersed_boundaries.jl
```diff
@@ -0,0 +1,98 @@
+using Adapt
+using CUDA: CuArray
+using Oceananigans.Fields: fill_halo_regions!
+using Oceananigans.Architectures: arch_array
+using Printf
+
+import Oceananigans.TurbulenceClosures: ivd_upper_diagonal,
```
simone-silvestri, line 7: Are you changing the `ivd_upper_diagonal`?

glwagner: I don't think so --- that's a mistake!

tomchor (approved):

tomchor:
@francispoulin @glwagner do you guys mind if I try to fix the docs and the conflict?

src/ImmersedBoundaries/partial_cell_immersed_boundaries.jl
tomchor: I just removed it since it indeed wasn't being used

tomchor:
Assuming the tests pass, are there still tests to be done or is this ready to merge/use?

francispoulin:
Thanks @tomchor for fixing this PR so that the tests pass.

So far what we have only tested the case when there is buoyancy with no motion.  It would be nice to have a problem with non-zero velocities work.   

I see there is this script [here](https://github.com/CliMA/Oceananigans.jl/blob/main/validation/immersed_boundaries/flow_over_hills.jl) that we could play with but it does not seem like a minimal working example.  Do you know of a simple script that looks at flow over a bump we could look at?

@glwagner also suggested we test this with lat-lon grids and create tests.

tomchor:
> So far what we have only tested the case when there is buoyancy with no motion. It would be nice to have a problem with non-zero velocities work.

Agreed. I most likely will be using IBMs starting now so I probably will be testing IBMs on a variety of settings.

> I see there is this script [here](https://github.com/CliMA/Oceananigans.jl/blob/main/validation/immersed_boundaries/flow_over_hills.jl) that we could play with but it does not seem like a minimal working example. Do you know of a simple script that looks at flow over a bump we could look at?

I don't have anything ready but that script looks pretty good to me. We just need to tweak it a bit? What would be the goal here as a validation script?

> @glwagner also suggested we test this with lat-lon grids and create tests.

I think that should be the case for lat-lon and rectilinear grids, no? At the moment I think we're kinda limited with IBM tests:

```
~/repos/Oceananigans.jl/test $ grep -r "\<Immersed" * 
test_boundary_conditions_integration.jl:            immersed_rectilinear_grid(topology) = ImmersedBoundaryGrid(RectilinearGrid(arch; topology, z=(-Lz, Lz), grid_kw...), ib)
test_boundary_conditions_integration.jl:                        if grid isa ImmersedBoundaryGrid && side == :bottom
test_boundary_conditions_integration.jl:                        if grid isa ImmersedBoundaryGrid && side == :bottom
test_boundary_conditions_integration.jl:            # Omit ImmersedBoundaryGrid from vertically-periodic test
test_conditional_reductions.jl:using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBoundary
test_conditional_reductions.jl:using Oceananigans.ImmersedBoundaries: conditional_length
test_conditional_reductions.jl:        @info "    Testing Reductions on Immersed fields"
test_conditional_reductions.jl:        ibg  = ImmersedBoundaryGrid(grid, GridFittedBoundary((x, y, z) -> (x < 0.5)))
test_dynamics.jl:using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBoundary, GridFittedBottom, mask_immersed_field!
test_dynamics.jl:    grid            = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom((x, y) -> 0))
test_dynamics.jl:    grid            = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bathymetry))
test_dynamics.jl:            # Immersed grid cases
test_dynamics.jl:            immersed_vertical_grid = ImmersedBoundaryGrid(RectilinearGrid(arch,
test_dynamics.jl:            ImmersedBoundaryGrid(RectilinearGrid(arch,
test_hydrostatic_free_surface_immersed_boundaries_implicit_solve.jl:using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBottom
test_hydrostatic_free_surface_immersed_boundaries_implicit_solve.jl:@testset "Immersed boundaries test divergent flow solve with hydrostatic free surface models" begin
test_hydrostatic_free_surface_immersed_boundaries_implicit_solve.jl:        grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(B))
test_hydrostatic_free_surface_immersed_boundaries.jl:using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBoundary, GridFittedBottom
test_hydrostatic_free_surface_immersed_boundaries.jl:@testset "Immersed boundaries with hydrostatic free surface models" begin
test_hydrostatic_free_surface_immersed_boundaries.jl:            grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBoundary(bump))
test_hydrostatic_free_surface_immersed_boundaries.jl:            @info "  Testing surface boundary conditions with ImmersedBoundaries in HydrostaticFreeSurfaceModel [$arch_str]..."
test_hydrostatic_free_surface_immersed_boundaries.jl:            grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bathymetry))
test_hydrostatic_free_surface_immersed_boundaries.jl:            grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(B))
test_multi_region_unit.jl:using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBottom, GridFittedBoundary
test_multi_region_unit.jl:                ibg = ImmersedBoundaryGrid(grid, immersed_boundary)
test_shallow_water_models.jl:using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBoundary
test_shallow_water_models.jl:    @testset "ShallowWaterModels with ImmersedBoundaryGrid" begin
test_shallow_water_models.jl:            @testset "ShallowWaterModels with ImmersedBoundaryGrid [$arch]" begin
test_shallow_water_models.jl:                @info "Testing ShallowWaterModels with ImmersedBoundaryGrid [$arch]"
test_shallow_water_models.jl:                grid_with_bump = ImmersedBoundaryGrid(grid, GridFittedBoundary(bump))
```

Although maybe added tests could be done in a separate PR?

glwagner:
We should add tests judiciously because our CI resources are currently a little strained.

tomchor:
@francispoulin @glwagner I took the liberty of fixing two things in the past commits:

- A trivial bug where `show()` wasn't defined for `PartialCellBottom`
- `on_architecture()` wasn't defined for `PartialCellBottom`, which led to a bug (both on CPUs and GPUs). So I generalized the types for `on_architecture()` in `grid_fitted_bottom.jl`

tomchor:
Sorry, I had mistaken `minimum_fractional_Δz` for its complementary part! Now that I have the correct meaning in mind, the value  makes perfect sense. Thanks for the clarification @francispoulin and @glwagner

PR #2310, opened 2022-03-06
Update Docs authors

navidcy:


docs/make.jl
```diff
@@ -154,7 +154,7 @@ format = Documenter.HTML(
 
 makedocs(bib,
   sitename = "Oceananigans.jl",
-   authors = "Ali Ramadhan, Gregory Wagner, John Marshall, Jean-Michel Campin, Chris Hill",
+   authors = "Climate Modeling Alliance and co-contributors",
```
tomchor, line 157: I like this change ;)
But why co-contributors and not contributors?

```suggestion
   authors = "Climate Modeling Alliance and contributors",
```

glwagner (approved):
Much better

PR #2320, opened 2022-03-08
`Field` Stokes shear and tendency with `UniformStokesDrift`

glwagner:
This PR provides a convenience constructor for building `UniformStokesDrift` with four `Field`s for the Stokes shear and tendency, rather than functions. It also adds kernel functions for the field case, so users now have the choice between functions of `(z, t)`, `Field`s, or `nothing`.

This is useful when calculating the Stokes profile is a relatively expensive or involved computation (eg, involving integration over a spectrum of waves). This permits two optimizations:

1. Stationary Stokes shear profiles can be precomputed.
2. Time-dependent Stokes shear profiles and Stokes tendencies can be computed in a `Callback`. This saves computation time for 3D runs because the Stokes profiles are 1D.
 
co-authored with @qingli411

TODO: 

- [x] add a test


navidcy (approved):

glwagner:
Thanks @navidcy for the help with this one!

navidcy:
What happened here? All tests were OK and one day disaster happened...

glwagner:
Damn, I thought we were about to merge too

navidcy:
> Damn, I thought we were about to merge too

It stops at `CATKEVerticalDiffusivity`...

glwagner:
🤞

navidcy:
I'm really confused with this one...

glwagner:
me too

navidcy:
Should we close this PR? If no, could someone resolve the conflicts/

glwagner:
I can resolve because it should be easy.

navidcy:
@glwagner should we try again for this PR?

navidcy:
All tests pass (for the first time!) but docs fail with `Out of memory` error. This error started appearing in all CI........

glwagner:
> All tests pass (for the first time!) but docs fail with `Out of memory` error. This error started appearing in all CI........

I think the fix is just to re-run. Maybe we need to reduce the number of agents on tartarus so there aren't so many builds running at the same time. We have a ton of PRs open I guess.

glwagner:
Another solution could be add a tag similar to what we have on ClimaOcean where the documentation is only built when we want it to.

navidcy:
I tried rerunning several times… 
I also tried to be mindful to not have a lot of PRs running their CI the same time. Nothing came to fruition.. I’ll try again.

navidcy:
> Another solution could be add a tag similar to what we have on ClimaOcean where the documentation is only built when we want it to.

The out of memory errors appear everywhere btw. Not only in docs build. 

https://buildkite.com/clima/oceananigans/builds/20430#0194cd33-3e4f-49ca-abc1-9cc36b428559/28-510

glwagner:
so chill

PR #2324, opened 2022-03-09
Split timestep tests

### LINKED ISSUE #2325: GPU timestep tests are taking too long
See, e.g., https://buildkite.com/clima/oceananigans/builds/6032#50bccc68-1818-409b-800c-f742a1dd44c1

navidcy:
Timestep tests gave yet again become too long. (~3hrs on gpu)
This PR attempts to split them. Closes #2325.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -234,6 +234,33 @@ steps:
     depends_on: "init_cpu"
 
 #####
+##### NonhydrostaticModel and time stepping (part 3)
+#####
+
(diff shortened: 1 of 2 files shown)


glwagner (approved):
Which new tests pushed us over the limit?

navidcy:
I think the new diffusion/cosine tests? Can't think of something else...

glwagner:
> I think the new diffusion/cosine tests? Can't think of something else...

Ah indeed.

It's also possible that some PRs caused a compile-time regression. Hopefully not a performance regression though. We should test that.

PR #2329, opened 2022-03-09
Adds back code coverage reporting via Codecov

navidcy:
This PR attempts to enable Codecov code coverage reports from tests.


francispoulin:
Mind if I ask what this PR does?  Sorry that i don't quite follow from looking at the code.

glwagner:
> Mind if I ask what this PR does? Sorry that i don't quite follow from looking at the code.

Not sure about the code but judging from the PR name I think it's going to generate an estimate of code coverage (eg the percentage of lines of code in the package that are tested during CI).

navidcy:
I'm trying to set up back Codecov for Oceananigans.jl, e.g., we'll be able to see which parts of the code are covered by the tests and which bits are not. Something like https://app.codecov.io/gh/CliMA/OceanLearning.jl

navidcy:
@maleadt do you know how I can add Codecov coverage in our [buildkite pipeline](https://github.com/CliMA/Oceananigans.jl/blob/ncc/adds-codecov/.buildkite/pipeline.yml)?

From what I read at [`JuliaGPU/buildkite/README`](https://github.com/JuliaGPU/buildkite/blob/main/README.md) here, I get that I have to add:

```Julia
    plugins:
      - JuliaCI/julia#v1:
          version: "1"
      - JuliaCI/julia-test#v1: ~
      - JuliaCI/julia-coverage#v1:
          codecov: true
```

in *every* step in our buildkite pipeline. Is this correct?

maleadt:
Correct. Buildkite doesn't support any templating last I looked.The `codecov` variable is currently unused though, so you can leave that out.

glwagner:
Is it a good idea then? Is it worth the effort? We'll have to keep it up?

navidcy:
> Is it a good idea then? Is it worth the effort? We'll have to keep it up?

I'm almost there. I just can't seem to be able to submit the coverage reports!
We don

glwagner:
> > Is it a good idea then? Is it worth the effort? We'll have to keep it up?
> 
> I'm almost there. I just can't seem to be able to submit the coverage reports!
> 
> We don't need to keep anything up! It's just for our own information. Well, if, e.g., we notice from the Codecov report that a big chunk of a submodule is not tested then we can have that in mind and either write some tests or inform users to use that module with caution?

Oh, I thought it required extensive changes in our yml files.

glwagner:
Is there any way to obtain code coverage on demand rather than in every PR?

navidcy:
@maleadt, I tried all combinations I could think of but I couldn't get `Codecov.submit()` to work using the `SECRET_CODECOV_TOKEN`. It was only when I added the actual token as a kwarg that worked, e.g.,
https://github.com/CliMA/Oceananigans.jl/blob/ed1fe321199fd08c28052cd23a8c1b9d68368345/.buildkite/pipeline.yml#L540

Any ideas how to avoid using the token per se?

maleadt:
The whole `SECRET_VAR` decryption scheme only works with our workers, which run specific hooks to decode those values: https://github.com/JuliaGPU/buildkite/blob/main/image/hooks/environment.

navidcy:
> The whole `SECRET_VAR` decryption scheme only works with our workers, which run specific hooks to decode those values: https://github.com/JuliaGPU/buildkite/blob/main/image/hooks/environment.

Gotcha!

PR #2338, opened 2022-03-12
Update architecture in distributed benchmark scripts

johnryantaylor:


francispoulin (approved):

benchmark/distributed_nonhydrostatic_model_mpi.jl
```diff
@@ -11,8 +11,6 @@ using Benchmarks
 
 Logging.global_logger(OceananigansLogger())
 
-MPI.Init()
-
```
glwagner, line 15: I think this is necessary --- or at least I can't run this script without it, because I get something like

```julia
[2022/03/13 13:00:26.020] INFO  Benchmarking weak scaling nonhydrostatic model with Slab decomposition [N=(128, 128, 16), ranks=(1, 1, 1)]...
Invalid MIT-MAGIC-COOKIE-1 keyInvalid MIT-MAGIC-COOKIE-1 keyNo protocol specified
[ Info: Oceananigans will use 24 threads
┌ Warning: Package Benchmarks does not have Oceananigans in its dependencies:
│ - If you have Benchmarks checked out for development and have
│   added Oceananigans as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Benchmarks
└ Loading Oceananigans into Benchmarks from project dependency, future warnings for Benchmarks are suppressed.
[ Info: Oceananigans will use 24 threads
*** The MPI_Comm_rank() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
```

glwagner:
After adding `MPI.Init()` back to `distributed_nonhydrostatic_model_mpi.jl`, the script seems to run (it's still running...)

Here's a few idle thoughts (we don't need to resolve in this PR, but just leaving here for future reference):

* The "Benchmarks" module in `/benchmarks/src` causes a few issues like:
```julia
┌ Warning: Package Benchmarks does not have Oceananigans in its dependencies:
│ - If you have Benchmarks checked out for development and have
│   added Oceananigans as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Benchmarks
└ Loading Oceananigans into Benchmarks from project dependency, future warnings for Benchmarks are suppressed.
```
It'd be nice to fix that eventually.

* The output is kind of annoying:

```julia
[2022/03/13 13:17:49.875] INFO  Benchmarking weak scaling nonhydrostatic model with Slab decomposition [N=(128, 128, 32), ranks=(1, 2, 1)]...
Invalid MIT-MAGIC-COOKIE-1 keyInvalid MIT-MAGIC-COOKIE-1 keyNo protocol specified
[ Info: Oceananigans will use 24 threads
[ Info: Oceananigans will use 24 threads
┌ Warning: Package Benchmarks does not have Oceananigans in its dependencies:
│ - If you have Benchmarks checked out for development and have
│   added Oceananigans as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Benchmarks
└ Loading Oceananigans into Benchmarks from project dependency, future warnings for Benchmarks are suppressed.
┌ Warning: Package Benchmarks does not have Oceananigans in its dependencies:
│ - If you have Benchmarks checked out for development and have
│   added Oceananigans as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Benchmarks
└ Loading Oceananigans into Benchmarks from project dependency, future warnings for Benchmarks are suppressed.
[ Info: Oceananigans will use 24 threads
[ Info: Oceananigans will use 24 threads
```
We don't need to be told `N` times how many threads Oceananigans is using.

* I think the `README.md` could be clearer and the scripts could have better names (eg which script runs the benchmark versus which script sets up the model)

benchmark/distributed_nonhydrostatic_model_mpi.jl
johnryantaylor: Thanks for catching that - Init_MPI() should definitely stay!

glwagner: Ok, I put it back in, hope that's ok!

glwagner:
Thanks @johnryantaylor !

PR #2346, opened 2022-03-13
Serialize grids with CPU architecture

### LINKED ISSUE #2179: Serialize grid with `CPU` architecture?
I think it might make sense to serialize grids in JLD2 files with `CPU` architecture by default; this way they can be loaded on systems without a GPU. 

The downside is that users with GPUs would have to write

```julia
cpu_grid = file["serialized/grid"]
grid = on_architecture(GPU(), cpu_grid)
```

to load their grid on the GPU.

glwagner:
Closes #2179

### DIFF
--- a/src/OutputWriters/output_writer_utils.jl
+++ b/src/OutputWriters/output_writer_utils.jl
@@ -53,7 +53,11 @@ serializeproperty!(file, location, p::AbstractArray) = saveproperty!(file, locat
 serializeproperty!(file, location, p::Function) = @warn "Cannot serialize Function property into $location"
 serializeproperty!(file, location, p::ContinuousBoundaryFunction) = @warn "Cannot serialize ContinuousBoundaryFunction property into $location"
 
+# Serializing grids:
+serializeproperty!(file, location, p::AbstractGrid) = file[location] = on_architecture(CPU(), p.grid)
+
 function serializeproperty!(file, location, p::ImmersedBoundaryGrid)
+    # TODO: convert immersed boundary grid to array representation in order to save.
     # Note: when we support array representations of immersed boundaries, we should save those too.
     @warn "Cannot serialize ImmersedBoundaryGrid; serializing underlying grid instead."
     file[location] = p.grid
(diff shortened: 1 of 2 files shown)


navidcy:
We get a bunch of warnings from:
https://github.com/CliMA/Oceananigans.jl/blob/484f9a2868023918b78e494f34671849d2b44ea5/src/OutputWriters/jld2_output_writer.jl#L209

src/OutputWriters/output_writer_utils.jl
```diff
@@ -27,7 +28,7 @@ saveproperty!(file, location, p::Tuple) =
 saveproperty!(file, location, p) =
     [saveproperty!(file, location * "/$subp", getproperty(p, subp)) for subp in propertynames(p)]
 
-saveproperty!(file, location, p::ImmersedBoundaryGrid) = saveproperty!(file, location, p.grid)
+saveproperty!(file, location, p::ImmersedBoundaryGrid) = saveproperty!(file, location, grid)
```
navidcy, line 31: I don't understand where from the `saveproperty!` knows `grid`.

src/OutputWriters/output_writer_utils.jl
```diff
@@ -53,10 +54,14 @@ serializeproperty!(file, location, p::AbstractArray) = saveproperty!(file, locat
 serializeproperty!(file, location, p::Function) = @warn "Cannot serialize Function property into $location"
 serializeproperty!(file, location, p::ContinuousBoundaryFunction) = @warn "Cannot serialize ContinuousBoundaryFunction property into $location"
 
+# Serializing grids:
+serializeproperty!(file, location, p::AbstractGrid) = file[location] = on_architecture(CPU(), grid)
```
navidcy, line 58: same here, I don't see where from `serializeproperty!` knows about `grid`

src/OutputWriters/output_writer_utils.jl
glwagner: Thats a typo / bug.

src/OutputWriters/output_writer_utils.jl
glwagner: Bug!

navidcy: 🐞🐛🐜

navidcy (approved):

PR #2335, opened 2022-03-11
Tupled halo fillings

simone-silvestri:
changes in this PR

instead of doing individual halo filling for each fill, ex:
```
fill_halo_regions(<:Tuple, args...)
for field in fields 
     fill_halo_regions!(field, args...)
end
```

we'll lump all the fields together in a kernel

PR #2362, opened 2022-03-16
Bugfix in `fill_halo` for julia < 1.7

### LINKED ISSUE #2358: Fill Halo not working in julia < 1.7
In my last PR (#2335) I introduced tupled halo filling. 

This PR required inferring the length of a Tuple of different element types on the GPU (which is not necessarily easy).
I found a hack to do it, but I was working in julia 1.7.

Today, running a simulation with several different functional boundary conditions (in `discrete_form`) with julia 1.6, I found out that the compiler does not like my solution and does not infer the Tuple length (spitting out an `Invalid LLVM` error). This error is not there in julia-1.7

Unfortunately, I already merged my PR because I stupidly did not think to test all possible BC with julia-1.6.

Now, to revert to a non-tupled halo filling which works across julia distribution it's a one-liner (in `field_tuples.jl`)
```
@inline fill_halo_regions_field_tuple!(full_fields, grid, args...; kwargs...) = 
        fill_halo_regions!(extract_field_data.(full_fields), extract_field_bcs.(full_fields), grid, args...; kwargs...)
```
to
```
@inline fill_halo_regions_field_tuple!(full_fields, grid, args...; kwargs...) = 
   for field in full_fields
      fill_halo_regions!(field, args...; kwargs...)
   end
end
```

What should we do? We could change this line and leave the infrastructure intact for when we update to julia-1.7 (then we would't have tupled halo filling till then) or find a completely different solution (which will require a lot of restructuring)
Probably we also need more tests for boundary conditions with different BC on different sides.

I see that the tests that run on julia-1.6 do not return an error though.  Do other people encounter the same error?

simone-silvestri:
closes #2358

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
-version = "0.73.0"
+version = "0.73.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/BoundaryConditions/fill_halo_regions.jl
+++ b/src/BoundaryConditions/fill_halo_regions.jl
@@ -114,6 +114,7 @@ end
 @kernel function _fill_west_and_east_halo!(c::NTuple, west_bc, east_bc, grid, args...)
     j, k = @index(Global, NTuple)
     ntuple(Val(length(west_bc))) do n
+        Base.@_inline_meta
         @inbounds begin
             _fill_west_halo!(j, k, grid, c[n], west_bc[n], args...)
             _fill_east_halo!(j, k, grid, c[n], east_bc[n], args...)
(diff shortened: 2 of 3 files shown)


navidcy:
Bump patch release?
Perhaps a test?

simone-silvestri:
It's weird... time stepping itself should be a test because of the multiple `fill_halo_regions!` both on individual fields and tuples. Here it seems that everything passes...

navidcy:
> It's weird... time stepping itself should be a test because of the multiple `fill_halo_regions!` both on individual fields and tuples. Here it seems that everything passes...

Let's bump a patch release and merge then.

navidcy:
> It's weird... time stepping itself should be a test because of the multiple `fill_halo_regions!` both on individual fields and tuples. Here it seems that everything passes also without this fix...

without this fix?

simone-silvestri:
yeah, it helped me getting it working on julia 1.6 on my computer but tests which use julia 1.6 passed already when I merged https://github.com/CliMA/Oceananigans.jl/pull/2335

navidcy (approved):

navidcy:
Merge?

tomchor (approved):

PR #2375, opened 2022-03-21
Fewer warnings from output writers

glwagner:
This PR removes some of the warnings that output writers throw when trying to save or serialize functions to disk.

These warnings are outdated in a way because they are really aimed at users trying to restore models from a checkpoint. Previously, we used to require that an entire model could be deserialized from disk when restoring from a checkpoint. This created special headaches for models that involved user-defined functions. 

But now we have a much better and simpler way for restoring models from a checkpoint. As a result we rarely care whether or not boundary conditions are serialized to disk if they are functions.


navidcy (approved):

navidcy:
the outputwriters were a bit bubbly I admit ;)

PR #2377, opened 2022-03-22
Fix bug for vertically-implicit diffusion with multiple vertically-implicit closures

glwagner:
This PR fixes a bug that crept in during #2295 that affects closure tuples with more than one vertically-implicit closure. The bug was an incorrect calculation of the diagonal coefficient of the tridiagonal matrix that's solved to effect an implicit time-step. This PR also adds a test to the good ol' `test_diffusion_cosine` for a tuple of two vertically-implicit `VerticalScalarDiffusivity`.

### DIFF
--- a/src/TurbulenceClosures/closure_tuples.jl
+++ b/src/TurbulenceClosures/closure_tuples.jl
@@ -18,8 +18,8 @@ end
 ##### Kernel functions
 #####
 
-funcs     = [:∂ⱼ_τ₁ⱼ, :∂ⱼ_τ₂ⱼ, :∂ⱼ_τ₃ⱼ, :∇_dot_qᶜ, :_ivd_upper_diagonal, :_ivd_lower_diagonal, :_ivd_diagonal]
-alt_funcs = [:∂ⱼ_τ₁ⱼ, :∂ⱼ_τ₂ⱼ, :∂ⱼ_τ₃ⱼ, :∇_dot_qᶜ, :ivd_upper_diagonal, :ivd_lower_diagonal, :ivd_diagonal]
+funcs     = [:∂ⱼ_τ₁ⱼ, :∂ⱼ_τ₂ⱼ, :∂ⱼ_τ₃ⱼ, :∇_dot_qᶜ, :_ivd_upper_diagonal, :_ivd_lower_diagonal]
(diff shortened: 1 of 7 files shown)


simone-silvestri (approved):

src/TurbulenceClosures/closure_tuples.jl
```diff
@@ -18,8 +18,8 @@ end
 ##### Kernel functions
 #####
 
-funcs     = [:∂ⱼ_τ₁ⱼ, :∂ⱼ_τ₂ⱼ, :∂ⱼ_τ₃ⱼ, :∇_dot_qᶜ, :_ivd_upper_diagonal, :_ivd_lower_diagonal, :_ivd_diagonal]
-alt_funcs = [:∂ⱼ_τ₁ⱼ, :∂ⱼ_τ₂ⱼ, :∂ⱼ_τ₃ⱼ, :∇_dot_qᶜ, :ivd_upper_diagonal, :ivd_lower_diagonal, :ivd_diagonal]
+funcs     = [:∂ⱼ_τ₁ⱼ, :∂ⱼ_τ₂ⱼ, :∂ⱼ_τ₃ⱼ, :∇_dot_qᶜ, :_ivd_upper_diagonal, :_ivd_lower_diagonal]
+alt_funcs = [:∂ⱼ_τ₁ⱼ, :∂ⱼ_τ₂ⱼ, :∂ⱼ_τ₃ⱼ, :∇_dot_qᶜ, :ivd_upper_diagonal, :ivd_lower_diagonal]
```
simone-silvestri, line 22: why do we need to have `_ivd_lower_diagonal` and `_ivd_upper_diagonal`?

glwagner: It's really just a convenience so that we can dispatch solely on `closure::Tuple` here. Otherwise, we have to disambiguate for `LZ::Center` and `LZ::Face` (and other possible specific versions). Not sure if it's better to disambiguate or use "interface" functions like the "`_`" versions?

glwagner: Basically it simplifies the code just below that line. But the cost is maybe that it's a little confusing why we need it (hence your question) so maybe not worth it...

simone-silvestri: Clear! For me it's ok to keep it like this.
Maybe a more verbose option would be to call them `tupled_ivd_*_diagonal`, like we did for immersed boundaries with `immersed_ivd_*_diagonal`, but it's not necessary

glwagner: mm yes except it would be `maybe_tupled_*`... that's a good idea.

PR #2379, opened 2022-03-23
Implements RiBasedVerticalDiffusivity

glwagner:
This PR implements a closure that's similar in spirit to [Pacanowski-Philander](https://journals.ametsoc.org/view/journals/phoc/11/11/1520-0485_1981_011_1443_povmin_2_0_co_2.xml) but should also work for convective mixing.

In short, there are two mixing coefficients: a viscosity for momentum and a diffusivity for tracers. They both have the form

```julia
κ = κ₀ * step_down(Ri, Ri₀, Riᵟ)
```

where ` κ₀` is the diffusivity when the buoyancy gradient is negative and there's no shear (ie, convection), and `Ri` is the Richardson number. `step_down` is a function that's `1` for `Ri < Ri₀`, `0` for `Ri > Ri₀ + Riᵟ`, and linearly varying between `1` and `0` otherwise.

Another way to view this closure is as a generalization of convective adjustment to cases with shear.

This closure has 6 parameters, 3 each for viscosity and diffusivity.

I'd like to calibrate parameters to the LESbrary before merging, so this PR is just a draft for now.


glwagner:
Just a proof of concept...

https://user-images.githubusercontent.com/15271942/159792588-cd6475aa-5d26-45b4-a14a-e46e7bfadb87.mp4

simone-silvestri (approved):

PR #2380, opened 2022-03-23
Return `field` from `compute!(field::ReducedComputedField)`

glwagner:
Just a little sugary convenience so we can write one-liners like

```Julia
U = compute!(Field(Average(u, dims=1)))
```

(Note that other `compute!` already do this, we just have a few lingering inconsistencies.)


navidcy (approved):

francispoulin (approved):
Good idea!

PR #2388, opened 2022-03-29
[WIP] High-order WENO stencil for horizontal divergence

glwagner:
This PR investigates the possibility of using a convex combination of high-order stencils to calculate horizontal divergence. The goal is to use a stencil for horizontal divergence that tames the omnipresent grid-scale noise in the vertical velocity field in large-scale hydrostatic models.

TODO:

- [ ] Add `divergence` property to `HydrostaticFreeSurfaceModel` and pass stencil info into kernels
- [ ] Update implicit free surface to use the vertical integral of horizontal divergence, rather than vertical integrating the velocity fields separately
- [ ] Fuse `Advection` module into `Operators`? The module `Advection` just defines stencils after all...
- [ ] Implement `UpwindWENO4` and `CenteredWENO5` horizontal divergence stencils


francispoulin:
Very interesting idea. How do you plan to combine the different schemes to get a smooth result?

glwagner:
> Very interesting idea. How do you plan to combine the different schemes to get a smooth result?

We'll use the WENO method, which weights candidate stencils based on a smoothness metric.

francispoulin:
> > Very interesting idea. How do you plan to combine the different schemes to get a smooth result?
> 
> We'll use the WENO method, which weights candidate stencils based on a smoothness metric.

Makes sense, thanks!

glwagner:
> > > Very interesting idea. How do you plan to combine the different schemes to get a smooth result?
> > 
> > 
> > We'll use the WENO method, which weights candidate stencils based on a smoothness metric.
> 
> Makes sense, thanks!

The non-trivial part of the method is to use a non-trivial stencil to reconstruct `u` at the interfaces of tracer cells. For both advection and continuity we use the (trivial) second-order reconstruction

```
u_ijk = u[i, j, k]
```

Yet high-order reconstructions are possible; for example a fourth-order reconstruction (in the x-direction only) is

```
u_ijk = 13/12 * u[i, j, k] - 1/24 * u[i-1, j, k] - 1/24 * u[i+1, j, k]
```

francispoulin:
Seems straightforward and a great idea!

glwagner:
I propose we change the API for vector invariant with WENO interpolation to read something like

```julia
momentum_advection = VectorInvariant(WENO5(), VelocitySmoothness())
```

or something like that. In other words I think the nesting should be reversed: the _first_ question is whether we are using the vector invariant or flux form formulation (flux form is default right now). The _second_ question is how we interpolate vorticity. The _third_ question is the field used in the smoothness metric.

I also think we should disentangle the smoothness indicators (eg velocity and vorticity) from the WENO5 struct; we can express this concept simply by re-writing the functions to accept two fields: one for reconstruction, and one for smoothness. The default uses the reconstructed field for smoothness. `VectorInvariant` will handle passing vorticity or velocity into those kernels.

simone-silvestri:
I think smoothness will be applied only to WENO so probably it makes sense to keep it inside the advection scheme?

glwagner:
> I think smoothness will be applied only to WENO so probably it makes sense to keep it inside the advection scheme?

good point

simone-silvestri:
Not so sure about that. It is a quite experimental idea so I think it should be a feature which is not to be abused. Also, in case of a `VelocityStencil` (or how we want to call it), we also have a different way of calculating the stencil that requires different boundary treatment (see `conditional_fluxes.jl`), so I would be cautious to generalize this

glwagner:
> Not so sure about that. It is a quite experimental idea so I think it should be a feature which is not to be abused.

Hopefully this change will make it easier to remove the feature?

glwagner:
Inspired by @simone-silvestri 's concerns, I'm wondering if what we need is actually a more robust system for specifying advection schemes in general (not just `VectorInvariant`), and controlling the behavior of the advective term. Perhaps something like:

```julia
""" Specifies a reconstruction scheme for fluxes of the form u * c,
where u is the `advecting_velocity` and c is the `advected_quantity` """
struct FluxReconstruction
    advecting_velocity
    advected_quantity
end

struct AdvectionScheme
    formulation # `FluxForm` or `VectorInvariant`
    mass_flux_reconstruction # "cell" or "colocated" reconstruction"; default CenteredSecondOrder
    momentum_flux_reconstruction
    tracer_reconstruction # NamedTuple for each tracer individually.
end
```

glwagner:
> That might be a good way to specify the advection of the complete set of coupled equations. The formulation applies only to momentum, but it is not a problem

I think you're right! Maybe...

```julia
""" Specifies a reconstruction scheme for fluxes of the form u * c,
where u is the `advecting_velocity` and c is the `advected_quantity` """
struct FluxForm
    advecting_velocity
    advected_quantity
end

""" Specifies a reconstruction scheme for the vector invariant formulation of
momentum advection."""
struct VectorInvariant
     vorticity_stencil
     vorticity_reconstruction
     # velocity_stencil?
end

struct UpwindWENOVorticityReconstruction
    scheme
    smoothness
end

struct AdvectionScheme
    mass # TBD... only support CenteredSecondOrder now
    momentum # FluxForm or VectorInvariant
    tracers # NamedTuple of FluxForm
end
```

glwagner:
I'm closing this. I'll open an issue to discuss an advection scheme refactor.

PR #2389, opened 2022-03-29
Adds MultipleForcings and AdvectiveForcing

### LINKED ISSUE #2136: Specifying multiple forcings for a model field
Mentioned in #2134, I want to include both a `Relaxation` and `Forcing` for a velocity component, specifically a sponge layer at the top and bottom of the domain and a forcing function within the non-sponge domain.

One potential syntax for this suggested by @glwagner is passing a tuple to the model constructor:  
```
forcing = (; u = (u_relaxation, u_sponge)),
```

This seems like a straightforward syntax from a user perspective since it's similar to other model constructor syntax, e.g., specifying tracers.

Since a `Relaxation` can be called from within a `Forcing` function, it might be easier to simple provide an example of that syntax to create a combined `Forcing`. The calling order of the forcings is explicit in that case since the user is writing the function and it doesn't require any additional development (well, other than an example).

glwagner:
This PR adds two features:

1. The ability to specify multiple forcing functions with a tuple (eg #2136)
2. A new forcing function type `AdvectiveForcing` for adding advective terms as a forcing function. A few users want to model buoyant or sinking particles so we need a nice way to support this that doesn't require importing low-level Oceananigans kernel functions. This was discussed on Slack and in #2385 .

Some notes:

* `AdvectiveForcing` can also model advection of momentum, but this will only work for `NonhydrostaticModel` at the moment (because the kernel functions for momentum advection have different names for nonhydrostatic vs hydrostatic)
* "forcing tuples" are wrapped in a lightweight `MultipleForcings` wrapper; this is just an internal convenience. However it does mean that we can more easily support alternative ways to "combine" forcings (other than just adding them) in the future if need be.

### Example of multiple forcings

From #2136, in model constructors:

```julia
forcing = (; u = (u_relaxation, u_sponge))
```

### Example of `AdvectiveForcing`

From the docstring:

```julia
# Physical parameters
gravitational_acceleration          = 9.81     # m s⁻²
ocean_density                       = 1026     # kg m⁻³
mean_particle_density               = 2000     # kg m⁻³
mean_particle_radius                = 1e-3     # m
ocean_molecular_kinematic_viscosity = 1.05e-6  # m² s⁻¹

# Terminal velocity of a sphere in viscous flow
Δb = gravitational_acceleration * (mean_particle_density - ocean_density) / ocean_density
ν = ocean_molecular_kinematic_viscosity
R = mean_particle_radius

w_Stokes = - 2/9 * Δb / ν * R^2

settling = AdvectiveForcing(WENO5(), w=w_Stokes)
```

Closes #2136 

TODO:
- [x] Tests
- [x] Validation case with simple sinking, reacting particle that combines `MultipleForcing` and `AdvectiveForcing`?


glwagner:
@iuryt do you want to help set up a validation case for this feature?

glwagner:
@johnryantaylor this might interest you!

iuryt:
> @iuryt do you want to help set up a validation case for this feature?

Yep. I will be working on the NP model with a P sinking velocity this week. Any idea for an experiment that changes MLD with time? 
Maybe making it 1D for simplicity? But I would like to test for 3D to make sure I am programming it correctly. Maybe buoyancy=nothing and diffuse a tanh-like temp profile? But would be better to have a shallowing mld instead.

glwagner:
> > @iuryt do you want to help set up a validation case for this feature?
> 
> Yep. I will be working on the NP model with a P sinking velocity this week. Any idea for an experiment that changes MLD with time? Maybe making it 1D for simplicity? But I would like to test for 3D to make sure I am programming it correctly. Maybe buoyancy=nothing and diffuse a tanh-like temp profile? But would be better to have a shallowing mld instead.

I think a 2D example with turbulence + reactions + sinking would be neat. Probably best to keep it simple and avoid an MLD calculation. Now I'm thinking maybe just a sinking tracer with a source near the surface would be good.. 🤔

iuryt:
What about this?

```julia
using Oceananigans
using Oceananigans.units

# local depth
const H = -10meters
# constant sinking velocity
const w_siking = -1 # m/s

# 1D grid
grid = RectilinearGrid(size=128, z=(H, 0), topology=(Flat, Flat, Bounded))

# constant diffusivity
closure = ScalarDiffusivity(κ=1)

# cools down at z=0
@inline cooling_func(x, y, z, t) = ifelse(z==0,-0.1,0)
cooling = Forcing(cooling_func)

# use WENO to advect down the tracer
sinking = AdvectiveForcing(WENO5(), w=w_siking)

model = NonhydrostaticModel(grid=grid, closure=closure, buoyancy=nothing, tracers=:T, forcing = (; T = (cooling, sinking)))

simulation = Simulation(model, Δt = 0.01 * H/w_sinking, stop_iteration = 100)

run!(simulation)
```

glwagner:
> I tried to run, but can't find `AdvectiveForcing()`
> 
> But I noticed that this is listed in https://github.com/CliMA/Oceananigans.jl/blob/glw/forcing-features/src/Forcings/Forcings.jl
> 
> It also doesn't find `ContinuousForcing`, for instance.

`ContinuousForcing` isn't exported by `using Oceananigans`; we have to get that name specifically: `using Oceananigans.Forcings: ContinuousForcing`.

glwagner:
> ```julia
> using Oceananigans
> using Oceananigans.Units
> 
> # local depth
> const H = -10meters
> # constant sinking velocity
> const w_siking = -1 # m/s
> 
> # 1D grid
> grid = RectilinearGrid(size=128, z=(H, 0), topology=(Flat, Flat, Bounded))
> 
> # constant diffusivity
> closure = ScalarDiffusivity(κ=1)
> 
> # cools down at z=0
> @inline cooling_func(x, y, z, t) = ifelse(z==0,-0.1,0)
> cooling = Forcing(cooling_func)
> 
> # use WENO to advect down the tracer
> sinking = AdvectiveForcing(WENO5(), w=w_siking)
> 
> model = NonhydrostaticModel(grid=grid, closure=closure, buoyancy=nothing, tracers=:T, forcing = (; T = (cooling, sinking)))
> 
> simulation = Simulation(model, Δt = 0.01 * H/w_sinking, stop_iteration = 100)
> 
> run!(simulation)
> ```

Nice, thanks! I'm gonna edit this to be 2D and have a source of turbulence, I'll send a result soon!

glwagner:
Here's what I got now!

```julia
using Oceananigans
using Oceananigans.Units
using GLMakie

grid = RectilinearGrid(size=(128, 128), x=(0, 128), z=(-64, 0), topology=(Periodic, Flat, Bounded))

b_bcs = FieldBoundaryConditions(top=FluxBoundaryCondition(1e-8))

@inline growth_func(x, y, z, t, p) = 1 / p.τ * exp(z / p.h)
growth = Forcing(growth_func, parameters=(τ=1hour, h=4.0))
sinking = AdvectiveForcing(WENO5(), w=-1)

                            tracers = (:b, :P),
                            buoyancy = BuoyancyTracer(),
                            boundary_conditions = (; b=b_bcs),
                            forcing = (; P = (growth, sinking)))

bᵢ(x, y, z) = 1e-5 * z + 1e-9 * rand()
set!(model, b=bᵢ)

Δz = grid.Δzᵃᵃᶜ
Δt = 0.1 * Δz # for a sinking velocity w=1
simulation = Simulation(model; Δt, stop_iteration = 1000)


fig = Figure()
ax = Axis(fig[1, 1])

P = model.tracers.P
hm = heatmap!(ax, interior(P, :, 1, :))

function update!(sim)
    hm.input_args[1][] = interior(P, :, 1, :)
    return nothing
end

simulation.callbacks[:plot] = Callback(update!, IterationInterval(100))

display(fig)

run!(simulation)
```

src/Forcings/advective_forcing.jl
```diff
@@ -0,0 +1,58 @@
+using Oceananigans.Advection: UpwindBiasedFifthOrder
```
iuryt, line 1: add `div_Uc`

src/Forcings/advective_forcing.jl
```diff
@@ -0,0 +1,58 @@
...
+function AdvectiveForcing(scheme=UpwindBiasedFifthOrder(); u=ZeroField(), v=ZeroField(), w=ZeroField())
+    velocities = (; u, v, w)
+    return AdvectiveForcing(velocities, scheme, nothing, nothing) # stub
+end
+    
+function regularize_forcing(af::AdvectiveForcing, field, field_name, model_field_names)
+    kernel_function =
+        field_name === :u ? div_𝐯u :
+        field_name === :v ? div_𝐯v :
+        field_name === :w ? div_𝐯w : div_Uc
```
iuryt, line 49: Trying to understand the syntax.

if `field_name` is `u`, then `kernel_function` is `div_vu`,
but if `field_name` is `v`, then `kernel_function` is `div_vv`,
and if `field_name` is `w`, then `kernel_function` is `div_vw`
if is none of that, use `div_Uc`?

Isn't is better to use `ifelse` instead? Does this runs on GPU?

iuryt (commented):
Some comments.

src/Forcings/advective_forcing.jl
glwagner: No, this function only runs once when the model is constructed, so here we prioritize readability.

src/Forcings/advective_forcing.jl
glwagner: Ah sorry... there were two commits I neglected to push

glwagner:
So @iuryt it looks like the tracer NaNs fairly quickly in the example.

I am wondering if we are implementing this right. The confusion here is whether we should advect tracer across the boundary. For all the cases we've considered up to know, the advection velocity is 0 across a boundary in a `Bounded` direction. This is different, because we impose a constant vertical velocity everywhere including at the top and bottom boundary. As a result tracer is not conserved. Is this what we want? Should we still use the "flux form" for the advection term here? There's no guarantee that the imposed velocity field is non-divergent, for example, An alternative would be to write the advection term as u_dot_grad_c. However in that case we cannot use high-order stencils (at least not the same ones that we use for advection by the "resolved" velocity field). Any idea what the right thing to do it?

glwagner:
> This is great, thanks @glwagner and @iuryt! Si Chen has been testing the code with a buouyant tracer added through forcing like this and he is running a comparison with Jenny Dingwall's simulations with Diablo. The extra `slip' velocity needs to vanish at the boundaries in order to conserve tracer, but then adding the advection term to the RHS forcing like @glwagner suggested seems to work and conserves tracer. We did something like this: 

```julia
lambda = 1 # decay scale for slip velocity in meters
for k=0:Nz+2
    slip_vel.w[:,:,k].+=(tanh(max(-grid.zᵃᵃᶠ[k]/lambda,0))*tanh(max((grid.zᵃᵃᶠ[k]+Lz)/lambda,0)))
end
```

Makes sense!

Since there's a `dw/dz`, this means that the "additional" velocity field has divergence and thus tracer still not exactly conserved, right? But perhaps the divergence is small enough that it's almost conserved.

Should we implement the advection differently (ie not using flux form) so that it's correct even when the velocity field is divergent? Another possibility is to have a flag, something like

```julia
forcing = AdvectiveForcing(w=prescribed_w, flux_form=true)
```

Or, perhaps `divergent=true`, or something like that. Whatever's least surprising / intuitive...

johnryantaylor:
Its ok for there to be a divergence in the slip velocity.  In the case of the ocean surface or seafloor that's exactly what you want since the particles will then accumulate at those surfaces depending on whether they are light or dense.  Tracer is conserved in an integrated sense since the flux through the boundary vanishes.  In some cases the user might want a flux across the boundary, e.g. for dense particles sinking through an open boundary at the bottom of the domain, so we probably want the profile of the slip velocity in the user's control.  Also, note that for very buoyant tracers there will be numerical issues if the tracer all accumulates at the surface.  Finer grid spacing can help a bit, but I think this is a drawback to using buoyant tracers and particles work better in those cases (Jenny discusses this in her draft paper on microplastics which is under review).

glwagner:
> Its ok for there to be a divergence in the slip velocity. In the case of the ocean surface or seafloor that's exactly what you want since the particles will then accumulate at those surfaces depending on whether they are light or dense. Tracer is conserved in an integrated sense since the flux through the boundary vanishes. In some cases the user might want a flux across the boundary, e.g. for dense particles sinking through an open boundary at the bottom of the domain, so we probably want the profile of the slip velocity in the user's control. Also, note that for very buoyant tracers there will be numerical issues if the tracer all accumulates at the surface. Finer grid spacing can help a bit, but I think this is a drawback to using buoyant tracers and particles work better in those cases (Jenny discusses this in her draft paper on microplastics which is under review).

That makes physical sense!

I believe supporting divergent slip velocities means we cannot use the "flux form" of the advection term, ie we cannot rewrite

```
u_slip ⋅ ∇c = ∇ ⋅ (u_slip c)
```

correct? This means that we cannot use native Oceananigans operators for this advection term, since the operators we've implemented all assume that the advecting velocity field is non-divergent.

This is not much of an issue since we can easily write a second-order discretization of `u_slip ⋅ ∇c`. In addition, we can support a flag in which users might optionally declare the velocity field as non-divergent, in which case we can use higher-order discretizations of `∇ ⋅ (u_slip c)`.

Perhaps `SlipVelocity` is a better name than `AdvectiveForcing` since it parallels the literature?

johnryantaylor:
For this application.  I'm not sure that it makes too much difference whether you write it in conservative form or not.  The decrease in slip velocity at the surface is always going to be pretty arbitrary and I think the important thing is just making sure that there is no flux through the boundary.  Regarding naming, SlipVelocity is more descriptive for this application, but maybe there are other reasons to add this type of forcing where you wouldn't call it `slip', in which case maybe AdvectiveForcing is more appropriate for a broader suite of applications.

iuryt:
The explanations that @johnryantaylor gave are worth for a textbook. I was just about to write that we usually assume no flux at the boundaries in order to conserve the tracer, but he gave a new perspective to this.

Taking an example from [PSOM](https://github.com/PSOM/V1.0/blob/master/code/NA/src/tracersource_nut.f90), they forced phytoplankton flux to zero at the boundaries, but using a first-order upwind scheme.

```fortran
            flux(0)= 0.0
            flux(NK)= 0.0
            do k=NK-1,1,-1
!     Kz is at cell faces
               dpdz= (Tr(1,i,j,k+1,0)-Tr(1,i,j,k,0))*wz(i,j,k)*DLinv
               flux(k) = KzmaxTr*Kz(i,j,k)*dpdz + wsink*Tr(1,i,j,k+1,0)
            end do
```

What if we leave for the users to give the velocity in which it will be advected?
Does that makes sense to simply give `w` as a function of `z` in which relax to 0 a the boundaries?
But you mentioned that Oceananigans assume non-divergent velocity field, but that's for the forcing, so it should be fine, right?

iuryt:
Another question:

A tracer can accumulate only if its buoyancy is different from the fluid buoyancy, right?
If so, can we define the buoyancy of the tracer and then the sinking/rising will be computed indirectly?
This is beyond what I was thinking for the use of `AdvectiveForcing`, but could be interesting, right?
What if we have a tracer that its buoyancy is the same as an intermediate level in the ocean and can accumulate there?
Pretty sure there are some chemical tracers that could act like that.

johnryantaylor:
Good point @iuryt.  The slip velocity is proportional to the difference between the particle and fluid densities.  Its common to assume that the slip velocity is constant for simplicity, but there are plenty of applications where the user might want the slip velocity to vary in space and time with the fluid density.  I think that should be straightforward when adding the extra advection as a forcing term since you could make the slip velocity a function of the fluid density.

iuryt:
So then should we simply use `SlipVelocity` for the sake of the generality and leave the user to define a function for that velocity that could depend on field variables.
Should we also leave for users to define the way it will diverge at the boundary?
What could be the simplest example we could use for this validation, @johnryantaylor ?

tomchor:
It could also be interesting to add this slip velocity to the convecting plankton example.

iuryt:
> It could also be interesting to add this slip velocity to the convecting plankton example.

I am developing an NPP model to run on Oceananigans, maybe instead of giving a new example, a could simply add the other terms on the plankton example. What do you think?

glwagner:
> @glwagner ,
> 
> So then, based on the code snippet given by @johnryantaylor and our discussion, this validation could be adapted to something like:
> 
> ```julia
> lamb = 1 # decay scale for slip velocity in meters
> @inline w(x,y,z,t) = (tanh(max(-z/lamb,0))*tanh(max((z+H)/lamb,0)))
> sinking = SlipVelocity(WENO5(), w=w)
> ```
> 
> The way the code is, does it accept functions for velocity inputs? Is that something I could do to help you?

You can do this a few ways:

```julia
# Vertical velocity function
const lamb = # something
const H = # something
@inline w_func(x, y, z) = tanh(max(-z / lamb, 0.0)) * tanh(max((z + H) / lamb, 0.0))

# Field
w = ZFaceField(grid)
set!(w, w_func)
sinking = AdvectiveForcing(w = w)

# Function field (avoids allocating memory)
w = FunctionField((Center, Center, Face), w_func, grid)
sinking = AdvectiveForcing(w = w)
```

We could also add the feature to `AdvectiveForcing` that, if an input is a function, we assume it's a function of `x, y, z, t` and deal with it appropriately (by wrapping in a `FunctionField` and adding reference to the model `clock`).

glwagner:
Here's a new 1D validation case that tests a "buoyant particle" model using a dynamic slip velocity that depends on buoyancy. The particle concentration field migrates to it's neural buoyancy (here in the center of the domain), until the particle concentration reaches a steady-state in a competition between diffusion and buoyant slippage.

https://user-images.githubusercontent.com/15271942/161167309-bfedffdb-6f2b-4552-8f49-1ec23ce7b4ed.mp4

validation/biogeochemistry/settling_diffusion.jl
```diff
@@ -0,0 +1,74 @@
...
+ν = 1e-4
+κ_particle = 1e-3
+R_particle = 1.0
+Δt = 1e-2
+
+grid = RectilinearGrid(size=128, z=(-4, 4), topology=(Flat, Flat, Bounded))
+
+no_penetration = ImpenetrableBoundaryCondition()
+slip_bcs = FieldBoundaryConditions(grid, (Center, Center, Face),
+                                   top=no_penetration, bottom=no_penetration)
```
iuryt, line 19: Awesome! Want to use this for the NP model.

glwagner: nice

iuryt:
After passing the checks, It just need a review and will be pushed?

glwagner:
> After passing the checks, It just need a review and will be pushed?

I think we should add a small test.

glwagner:
Here's a video from the `sediment_entrainment.jl` validation case:

https://user-images.githubusercontent.com/15271942/161605652-7e1596b0-f76e-4076-902f-6b4e7c78cfae.mp4

johnryantaylor:
looks really cool - nice work!

iuryt:
That's so cool!!!!

glwagner:
Last validation for now I think... two tracers that react (a - > b), one sinking, one rising:

https://user-images.githubusercontent.com/15271942/161640083-9df98a3c-c6cb-4ccf-bf04-d396b4e4ddc5.mp4

glwagner:
I'm a bit stumped. The error is

```
Test threw exception
--
  | Expression: advective_and_multiple_forcing(arch)
  | TaskFailedException
  |  
  | nested task error: BoundsError
```

But I can't reproduce this locally.

I can fathom a `BoundsError` when we use a high-order advection scheme and our halos aren't big enough. Right now we check the "model advection scheme" and the grid halos to catch this issue, but we don't check the forcings.

On the other hand, if I time step a model for which I _know_ the halos are too small (locally), I don't get an error (even though I'd like to). So maybe there _is_ something fishy...

iuryt:
> I'm a bit stumped. The error is
> 
> ```
> Test threw exception
> --
>   | Expression: advective_and_multiple_forcing(arch)
>   | TaskFailedException
>   |  
>   | nested task error: BoundsError
> ```
> 
> But I can't reproduce this locally.
> 
> I can fathom a `BoundsError` when we use a high-order advection scheme and our halos aren't big enough. Right now we check the "model advection scheme" and the grid halos to catch this issue, but we don't check the forcings. However, the test uses `halo = (3, 3, 3)` right now:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/791bb83e4c49386cc31292bf391762f1cd96bdee/test/test_forcings.jl#L115
> 
> On the other hand, if I time step a model for which I _know_ the halos are too small (locally), I don't get an error (even though I'd like to). So maybe there _is_ something fishy...

I don't know how to help, but I am following to see if can learn something and maybe next time I can be a better helper.

glwagner:
> I don't know how to help, but I am following to see if can learn something and maybe next time I can be a better helper.

Can you reproduce the error in the test locally? You can try navigating to `/test`, starting with `julia --project`, and then

```julia
julia> include("test_forcings.jl")
```

you might need to add a few packages to your global environment like `Test` to make this work.

Also running

```
TEST_GROUP=time_stepping_2 julia --project -e 'using Pkg; Pkg.test()'
```

will run the forcing tests (and a few others). If you do that, do you see an error like in the CI or do the tests pass?

Also running just the problematic test in isolation can be useful:

https://github.com/CliMA/Oceananigans.jl/blob/791bb83e4c49386cc31292bf391762f1cd96bdee/test/test_forcings.jl#L114-:130

docs/src/model_setup/forcing_functions.md
```diff
@@ -296,14 +296,13 @@ using Oceananigans
 r_sediment = 1e-4 # "Fine sand"
 ρ_sediment = 1200 # kg m⁻³
 ρ_ocean = 1026 # kg m⁻³
-Δb = - 9.81 * (ρ_sediment - ρ_ocean) / ρ_ocean
+Δb = 9.81 * (ρ_ocean - ρ_sediment) / ρ_ocean
```
iuryt, line 299: What's the difference here?

glwagner: easier to understand

glwagner: The only thing that matters...

glwagner:
> > I don't know how to help, but I am following to see if can learn something and maybe next time I can be a better helper.
> 
> Can you reproduce the error in the test locally? You can try navigating to `/test`, starting with `julia --project`, and then
> 
> ```julia
> julia> include("test_forcings.jl")
> ```
> 
> you might need to add a few packages to your global environment like `Test` to make this work.
> 
> Also running
> 
> ```
> TEST_GROUP=time_stepping_2 julia --project -e 'using Pkg; Pkg.test()'
> ```
> 
> will run the forcing tests (and a few others). If you do that, do you see an error like in the CI or do the tests pass?
> 
> Also running just the problematic test in isolation can be useful:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/791bb83e4c49386cc31292bf391762f1cd96bdee/test/test_forcings.jl#L114-L130

@iuryt this is no longer needed since I fixed the issue (now we wrap constant velocities in `ConstantField` and that seems to work). The remaining error is a GPU compilation issue...

iuryt:
Sorry, I was out. I can try running on GPU to see if I can catch the bug. Any specific test I could run?

glwagner:
> Sorry, I was out. I can try running on GPU to see if I can catch the bug. Any specific test I could run? I commented on the code, but did not see your post asking me to run the test. When I clicked on the notification it jumped directly to the code, sorry.

No, nothing to do! Just sorting out details now and hopefully will be done soon.

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
 └── field dependencies: (:T,)
 ```
 
+## `AdvectiveForcing`
+
+`AdvectiveForcing` defines a forcing function that represents advection by
+a separate or "slip" velocity relative to the prognostic model velocity field.
+`AdvectiveForcing` is implemented with native Oceananigans advection operators,
+which means that tracers advected by the "flux form" advection term
+``∇ ⋅ u⃗ˢˡⁱᵖ c``. Caution is advised when ``u⃗ˢˡⁱᵖ`` is not divergence free.
```
iuryt, line 289: I don't know if it is something with my browser, but the unicode characters look strange here. Is this suppose to look like this from here, but will be fine on the documentation?

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+
+no_penetration = ImpenetrableBoundaryCondition()
+slip_bcs = FieldBoundaryConditions(grid, (Center, Center, Face),
+                                   top=no_penetration, bottom=no_penetration)
+
+w_slip = ZFaceField(grid, boundary_conditions=slip_bcs)
+sinking = AdvectiveForcing(WENO5(; grid), w=w_slip)
+
+# output
+AdvectiveForcing with the WENO5 scheme:
```
iuryt, line 334: If the idea is to be consistent, we could use `UpwindBiasedFifthOrder()` here as well, but if the idea is to show that users can choose other schemes, we can keep as it is.

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+├── u: ZeroField{Int64}
+├── v: ZeroField{Int64}
+└── w: 32×32×33 Field{Center, Center, Face} on RectilinearGrid on CPU
```
iuryt, line 337: Isn't it better to specify the architecture while creating the grid? Then it would be obvious why the `AdvectiveForcing` object is on CPU.

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+
+```jldoctest sinking
+using Oceananigans.BoundaryConditions: fill_halo_regions!
+
+model = NonhydrostaticModel(; grid, tracers=(:b, :P), forcing=(; P=sinking))
+simulation = Simulation(model; Δt=1, stop_iteration=100)
+
+# Build abstract operation for slip velocity
+b_particle = - 1e-4 # relative buoyancy depends on reference density and initial buoyancy condition
+b = model.tracers.b
```
iuryt, line 351: Is that any difference here if we define `b = model.tracers.b` or `b = simulation.model.tracers.b`?

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+R = 1e-3 # mean particle radius
```
iuryt, line 352: in meters

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+ν = 1.05e-6 # molecular kinematic viscosity of water
```
iuryt, line 353: in m^2/s

src/Forcings/Forcings.jl
```diff
@@ -1,13 +1,15 @@
...
 
 using Oceananigans.Fields
 
 include("continuous_forcing.jl")
 include("discrete_forcing.jl")
 include("relaxation.jl")
+include("advective_forcing.jl")
 include("forcing.jl")
 include("model_forcing.jl")
+include("multiple_forcings.jl")
```
iuryt, line 13: Is that a reason for including files in this order? If not, should we have a criterium for choosing the order based on the programming style?

src/Forcings/advective_forcing.jl
```diff
@@ -0,0 +1,84 @@
+using Oceananigans.Advection: UpwindBiasedFifthOrder, div_Uc, div_𝐯u, div_𝐯v, div_𝐯w
+using Oceananigans.Fields: ZeroField, ConstantField
+using Adapt
+
+maybe_constant_field(u) = u
+maybe_constant_field(u::Number) = ConstantField(u)
```
iuryt, line 6: Defining the function twice?

src/Forcings/model_forcing.jl
```diff
@@ -28,7 +28,7 @@ regularize_forcing(::Nothing, field::AbstractField, field_name, model_field_name
 """
     model_forcing(model_fields; forcings...)
 
-Return a named tuple of forcing functions for each field in `model_fields`, wrapping
+Return a NamedTuple of forcing functions for each field in `model_fields`, wrapping
```
iuryt, line 31: Maybe using backticks (`NamedTuple`)?

iuryt (commented):
I tried to do my best on reviewing it.

docs/src/model_setup/forcing_functions.md
glwagner: The idea is to show we can use different schemes.

docs/src/model_setup/forcing_functions.md
glwagner: They look a little strange to me too unfortunately.

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+sinking = AdvectiveForcing(UpwindBiasedFifthOrder(), w=w_sediment)
+
+# output
+AdvectiveForcing with the UpwindBiasedFifthOrder scheme:
+├── u: ZeroField{Int64}
+├── v: ZeroField{Int64}
+└── w: ConstantField(-0.00352102)
+```
+
+The first argument to `AdvectiveForcing` is the advection scheme (here `WENO5()`).
```
glwagner, line 312:
```suggestion
The first argument to `AdvectiveForcing` is the advection scheme (here `UpwindBiasedFifthOrder()`).
```

docs/src/model_setup/forcing_functions.md
glwagner: I think its ok to use defaults --- that's why they exist. Well-chosen defaults increase code readability and understandability, I think.

docs/src/model_setup/forcing_functions.md
glwagner: If `simulation` was built with `model`, then they are the same.

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+```jldoctest sinking
+using Oceananigans.BoundaryConditions: fill_halo_regions!
+
+model = NonhydrostaticModel(; grid, tracers=(:b, :P), forcing=(; P=sinking))
+simulation = Simulation(model; Δt=1, stop_iteration=100)
+
+# Build abstract operation for slip velocity
+b_particle = - 1e-4 # relative buoyancy depends on reference density and initial buoyancy condition
+b = model.tracers.b
+R = 1e-3 # mean particle radius
```
glwagner, line 352:
```suggestion
R = 1e-3 # [m] mean particle radius
```

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+using Oceananigans.BoundaryConditions: fill_halo_regions!
+
+model = NonhydrostaticModel(; grid, tracers=(:b, :P), forcing=(; P=sinking))
+simulation = Simulation(model; Δt=1, stop_iteration=100)
+
+# Build abstract operation for slip velocity
+b_particle = - 1e-4 # relative buoyancy depends on reference density and initial buoyancy condition
+b = model.tracers.b
+R = 1e-3 # [m] mean particle radius
+ν = 1.05e-6 # molecular kinematic viscosity of water
```
glwagner, line 353:
```suggestion
ν = 1.05e-6 # [m² s⁻¹] molecular kinematic viscosity of water
```

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+As an example, consider a model for sediment settling at a constant rate:
+
+```jldoctest
+using Oceananigans
+
+r_sediment = 1e-4 # "Fine sand"
+ρ_sediment = 1200 # kg m⁻³
+ρ_ocean = 1026 # kg m⁻³
+Δb = 9.81 * (ρ_ocean - ρ_sediment) / ρ_ocean
+ν_molecular = 1.05e-6
```
glwagner, line 300:
```suggestion
ν_molecular = 1.05e-6 # m² s⁻¹
```

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+
+As an example, consider a model for sediment settling at a constant rate:
+
+```jldoctest
+using Oceananigans
+
+r_sediment = 1e-4 # "Fine sand"
+ρ_sediment = 1200 # kg m⁻³
+ρ_ocean = 1026 # kg m⁻³
+Δb = 9.81 * (ρ_ocean - ρ_sediment) / ρ_ocean
```
glwagner, line 299:
```suggestion
Δb = 9.81 * (ρ_ocean - ρ_sediment) / ρ_ocean m s⁻²
```

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+ν_molecular = 1.05e-6
+w_sediment = 2/9 * Δb / ν_molecular * r_sediment^2
```
glwagner, line 301:
```suggestion
w_sediment = 2/9 * Δb / ν_molecular * r_sediment^2 # m s⁻¹
```

src/Forcings/advective_forcing.jl
glwagner: There are two methods: one "fallback" which just passes `u` through, and a second function that converts `u` to `ConstantField(u)` when `u` is an integer. This allows `u` to be indexed into in the advection scheme functions. (A `ConstantField` is a field that returns a constant when we write `u[i, j, k]`).

src/Forcings/Forcings.jl
iuryt: I was asking because you didn't add the files at the end, but agree that shouldn't change anything.

src/Forcings/advective_forcing.jl
iuryt: I see, so basically the function just pass u if u is not a number (e.g. a Field). If u is a number, it will return a constant field.

docs/src/model_setup/forcing_functions.md
```diff
@@ -277,3 +280,88 @@ ContinuousForcing{Nothing} at (Center, Center, Center)
...
+sinking = AdvectiveForcing(UpwindBiasedFifthOrder(), w=w_sediment)
+
+# output
+AdvectiveForcing with the UpwindBiasedFifthOrder scheme:
+├── u: ZeroField{Int64}
+├── v: ZeroField{Int64}
+└── w: ConstantField(-0.00352102)
+```
+
+The first argument to `AdvectiveForcing` is the advection scheme (here `WENO5()`).
```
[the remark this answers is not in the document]
iuryt: Haven't tried it yet. Will try in the next version if I find something.

iuryt (commented):
My humble review. haha

I also have a question. Should we have test cases for `AdvectiveForcing` in `u` and `v`?

iuryt (approved):

simone-silvestri (approved):

francispoulin (approved):

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.74.0"
+version = "0.74.1"
```
tomchor, line 4: I'm not 100% clear on the versioning we do. Since this adds a feature should it be a minor version bump? Or are those only for breaking changes?

tomchor (approved):

Project.toml
glwagner: We've been bumping the minor version for API changes while bumping just the patch release for new features or bug fixes.

PR #2390, opened 2022-03-29
add div_Uc

iuryt:
Running the example from #2389  it returned this error
`ERROR: LoadError: UndefVarError: div_Uc not defined`

Hope this suggestion makes sense.


iuryt:
@glwagner 
I should have reviewed your code instead, right?

This is now complaining about `af` variable
`nested task error: UndefVarError: af not defined`

glwagner:
I sent a flurry of updates once I had the validation case coded up! It should work now

iuryt:
> I sent a flurry of updates once I had the validation case coded up! It should work now

Wait, what should I do to test that? haha

Issue #2391, opened 2022-03-29
`Fields()` throws an error when `indices` keyword is used


tomchor:
The call to `Field()` doesn't appear to be working with just the `indices` keyword argument:

```julia
julia> using Oceananigans

julia> grid = RectilinearGrid(size=(1, 1, 1), extent=(1, 1, 1))
1×1×1 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 1×1×1 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=1.0
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=1.0
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=1.0

julia> model = NonhydrostaticModel(; grid)
NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
├── grid: 1×1×1 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 1×1×1 halo
├── timestepper: QuasiAdamsBashforth2TimeStepper
├── tracers: ()
├── closure: Nothing
├── buoyancy: Nothing
└── coriolis: Nothing

julia> Field(model.velocities.u; indices=(:,:,1))
[...]

PR #2392, opened 2022-03-29
Fixes and tests `Field(u::Field; indices)` and `Field(u::AbstractOperation; indices)`

tomchor:
Closes https://github.com/CliMA/Oceananigans.jl/issues/2391

### DIFF
--- a/src/Fields/field.jl
+++ b/src/Fields/field.jl
@@ -648,16 +648,23 @@ end
 function fill_halo_regions!(field::Field, args...; kwargs...)
     reduced_dims = reduced_dimensions(field)
 
-    if !(field.indices isa typeof(default_indices(3))) # filter bcs for non-default indices
-        maybe_filtered_bcs = FieldBoundaryConditions(field.indices, field.boundary_conditions)
-    else
-        maybe_filtered_bcs = field.boundary_conditions
+    # To correctly fill the halo regions of fields with non-default indices, we'd have to
+    # offset indices in the fill halo regions kernels.
+    # For now we punt and don't support filling halo regions on windowed fields.
+    # Note that `FieldBoundaryConditions` _can_ filter boundary conditions in
+    # windowed directions:
+    #
+    #   filtered_bcs = FieldBoundaryConditions(field.indices, field.boundary_conditions)
+    #  
+    # which will be useful for implementing halo filling for windowed fields in the future.
+    if field.indices isa typeof(default_indices(3))
+        fill_halo_regions!(field.data,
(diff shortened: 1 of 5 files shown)


glwagner:
Mind if I commit to this PR?

test/test_field.jl
```diff
@@ -218,6 +218,8 @@ end
                     field_sz = test_field_sizes[t]
                     parent_sz = test_parent_sizes[t]
                     f_view = view(f, indices...)
+                    f_sliced = Field(f, indices)
+                    f2_sliced = Field(f^2, indices)
```
glwagner, line 222: There's no other test for abstract operations in this file... should we put this in `test_abstract_operations.jl`? The other test is perfect.

glwagner: Or actually I think `test_computed_field.jl` is the ideal place to test this.

tomchor: Yeah, move those around however you think is necessary. I never know what the proper place for tests are; that was just by best guess

tomchor:
> Mind if I commit to this PR?

Commit away!

test/test_field.jl
glwagner: It's subjective!

glwagner (approved):
Hopefully that does it. We were missing 1) `validate_indices` inside the computed `Field` constructor and 2) a translator from `Field(f::Field; indices) = view(f, indices...)`.

src/Fields/field.jl
```diff
@@ -157,7 +157,7 @@ function Field(loc::Tuple,
 end
     
 Field(z::ZeroField; kw...) = z
-Field(f::Field) = f # hmm...
+Field(f::Field; indices=f.indices) = view(f, indices...) # hmm...
```
tomchor, line 160: :+1: this is probably all that was needed to fix the bug

tomchor:
```suggestion
Field(f::Field; indices=f.indices) = view(f, indices...)
```

glwagner: There's still a few uncertain things here, like whether we should respect / include other keyword arguments (boundary conditions, data, etc...)

tomchor: Then can we include a more descriptive comment please?

Also, doesn't removing the line `Field(f::Field) = f` mean that `Field(u)` won't work anymore? Maybe we wanna keep that line?

glwagner: `Field(u)` works; the kwarg is default `u.indices` (eg the same indices)

PR #2394, opened 2022-03-30
On architecture for immersed boundary

simone-silvestri:


glwagner (approved):
a work of art

tomchor:
May I ask what are the practical implications of this PR? IBM seemed to be working on GPUs and CPUs well before this.

simone-silvestri:
It's just a small detail...

if you had an `ImmersedBoundary` on the `GPU` and wanted to move it on the `CPU` (in case the immersed boundary was an abstractarray) the `immersed_boundary` would remain a `CuArray` even after calling `on_architecture(CPU(), ibg)`. 

This was a small bug which never really affected us (you never really need to shift a grid from the `GPU` to the `CPU` in practice) but I needed this because when I do a `MultiRegionGrid` of a `ImmersedBoundary` and split the `bottom` in different `GPU`s, I need to pass that bottom correctly to the `CPU` otherwise we have `CUDA illegal access` problems (i.e. you cannot just split an from one GPU to others without taking care of it adequately)

tomchor:
> It's just a small detail...
> 
> if you had an `ImmersedBoundary` on the `GPU` and wanted to move it on the `CPU` (in case the immersed boundary was an abstractarray) the `immersed_boundary` would remain a `CuArray` even after calling `on_architecture(CPU(), ibg)`.
> 
> This was a small bug which never really affected us (you never really need to shift a grid from the `GPU` to the `CPU` in practice and most of the time that bottom is a function, not an array). On the other hand I needed this because when I do a `MultiRegionGrid` of a `ImmersedBoundary` and split the `bottom` in different `GPU`s, I need to pass that bottom correctly to the `CPU` otherwise we have `CUDA illegal access` problems (i.e. you cannot just split an array from one GPU to others without taking care of it adequately)
> 
> So I was getting an error related to this when I wanted to split a simulation with "realistic" bathymetry onto 2 GPUs

Makes sense. Thanks!

PR #2402, opened 2022-03-31
Adds flow_over_hills validation case

glwagner:
@tomchor and @whitleyv this PR adds a script that we can use to _estimate_ the relative effect of bottom drag vs form drag in buoyancy-stratified simulations over hilly bathymetry. This script only produces an estimate, because the bottom drag is implemented in a flat bottom setup, while form drag requires bathymetry. I'd hypothesis that we get more drag when there are both hills and bottom drag, because 1) greater shear at the bottom and 2) stronger downwards mixing of momentum due to strong turbulence.

We could use #2275 to extend the results obtained here to cases with both drag and hills.

I think it'll also be interesting if more accurate bathymetry representations like the one in #2306 lead to a reduction in form drag (cc @francispoulin). Possibly the spurious form drag is a nice metric to use to judge developments in our representation of bathymetry...

I'll run these simulations at high resolution and report some results, hopefully soon.

Side note: it's annoying that we don't serialized the immersed boundary so I might try to fix that here too...


glwagner:
I want to merge this then use it to develop the user API for immersed boundary fluxes.

Also, did anyone use this?

Partly I am wondering whether we may actually need shaved cells for accurate LES computations... otherwise our boundaries will always have "effective roughness", and form drag may dominate. It is actually possible that form drag dominance is ok as a model for rough boundaries. But it'd be nice to have the possibility of "smooth" boundaries where fluxes can be controlled via parameterizations.

tomchor:
> I want to merge this then use it to develop the user API for immersed boundary fluxes.
> 
> Also, did anyone use this?

I haven't used this, but I can test it if you want.

> Partly I am wondering whether we may actually need shaved cells for accurate LES computations in complex domains with non-straight immersed boundaries) ... otherwise our boundaries will always have "effective roughness", and form drag may dominate. It is actually possible that form drag dominance is ok as a model for rough boundaries. But it'd be nice to have the possibility of "smooth" boundaries where fluxes can be controlled via parameterizations.
> 
> It's also interesting that PALM, which uses an almost identical numerical method that we use, is primarily used for atmospheric LES in urban environments. So their boundaries _are mostly flat_ (the sides of buildings, etc...)

Shaved cells are definitely preferred in terms of accuracy (and convergence with resolution), but might not be necessary in a number of cases. My intuition from seeing results from papers (and from previous LES+IBM at my previous group) is that partial step cells would be almost as accurate as shaved cells (at least for oceanic cases with mild bathymetry slopes) but significantly simpler to implement. Since we already have the full-cell grid-fitted IBM it might be better to just focus on that for now and later modify it to make it partial-step when everything is working well (I think we'd just need to modify the calculation of vertical areas and volumes, no?).

Also, when running ocean LES the aspect ratios are generally isotropic (Δz/Δx~1) but the bathymetry slopes are very small, which means that for some cases even full-step grid-fitted IBM might be enough. Only some tests will tell for sure though.

validation/immersed_boundaries/flow_over_hills.jl
```diff
@@ -0,0 +1,160 @@
...
+                          save_interval = 0.1,
+                          architecture = CPU(),
+                          name = "flow_over_hills")
+
+    underlying_grid = RectilinearGrid(architecture,
+                                      size = (Nx, Nz),
+                                      halo = (3, 3),
+                                      x = (0, 2π),
+                                      z = (0, 2π),
+                                      topology = (Periodic, Flat, Bounded))
```
tomchor, line 25: I think in a 2D domain the effect of form drag will probably be exacerbated in comparison with a 3D simulation. I think this is find since this is just a test, but something to keep in mind

validation/immersed_boundaries/flow_over_hills.jl
```diff
@@ -0,0 +1,160 @@
...
+stop_time = 100.0
+reference_name = "bottom_drag_reference"
+#=
+reference_sim = hilly_simulation(; stop_time, Nx, ϵ=0, name=reference_name, bottom_drag=true)
+run!(reference_sim)
+δU_ref, t_ref = momentum_time_series(name * ".jld2")
+=#
+
+experiments = []
+for ϵ = [0.02, 0.05, 0.1]
```
tomchor, line 111: I'd suggest a more descriptive name for the height than `ϵ` if possible

validation/immersed_boundaries/flow_over_hills.jl
```diff
@@ -0,0 +1,160 @@
...
+        hills = @. ϵ * (1 + sin(x)) / 2
+        grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(hills))
+    else # no hills
+        grid = underlying_grid
+    end
+
+    closure = isfinite(Re) ? ScalarDiffusivity(ν=1/Re, κ=1/Re) : nothing
+
+    if bottom_drag
+        Δz = 2π / Nz
```
tomchor, line 38: Maybe use `grid.Δzaac` here to be more general?

tomchor (approved):

validation/immersed_boundaries/flow_over_hills.jl
glwagner: Ok!

validation/immersed_boundaries/flow_over_hills.jl
glwagner: It's a fractional height -- but the domain is non-dimensional (height 1).

validation/immersed_boundaries/flow_over_hills.jl
glwagner: Hmm, I feel this is a qualitative test so it doesn't matter?

glwagner: But ok.

glwagner: But wait. Using `grid.Δzᵃᵃᶜ` isn't generic because it doesn't work on a stretched grid. For things to be "correct" we need to do a lot more work (eg evaluate spacing following the hills, since spacing would change esp with `PartialBottomGrid`. Do you mind if we just keep the simplicity of this? We can do the hard work of generalizing it later if we need that.

validation/immersed_boundaries/flow_over_hills.jl
glwagner: I guess the height is actually 2 pi.

glwagner: "if possible" 👀

validation/immersed_boundaries/flow_over_hills.jl
tomchor: It's more in case you change the height on the domain down the line. But it's up you!

validation/immersed_boundaries/flow_over_hills.jl
tomchor: Isn't the height 2π?

glwagner:
> Since we already have the full-cell grid-fitted IBM it might be better to just focus on that for now and later modify it to make it partial-step when everything is working well (I think we'd just need to modify the calculation of vertical areas and volumes, no?).

Not sure what you mean in terms of "focus". I wasn't proposing any work, just speculating about the potential advantages of shaved cells. Do you mean focus on full cell IBM for this particular test? I agree with that --- we don't have shaved cells so we can't test it ?

validation/immersed_boundaries/flow_over_hills.jl
tomchor: But granted, I see it's a ratio. I think the PR is good enough to merge already, I just think this would make things clearer.

glwagner:
> Also, when running ocean LES the aspect ratios are generally isotropic (Δz/Δx~1) but the bathymetry slopes are very small, which means that for some cases even full-step grid-fitted IBM might be enough. Only some tests will tell for sure though.

Doesn't this depend on the setup? The actual ocean bottom is littered with boulders and other features of various sizes, and there's plenty of work in very steep canyons, ie https://en.wikipedia.org/wiki/Scripps_Canyon

validation/immersed_boundaries/flow_over_hills.jl
glwagner: Ok, lets plan to generalize this script when we need to!

validation/immersed_boundaries/flow_over_hills.jl
glwagner: Is it ok to leave as is and just merge?

glwagner: I'll test this more extensively on the immersed boundaries PR and probably update it a bit. I'll also be adding fluxes using the new API there. I can change "epsilon" to "h" there.

tomchor:
> > Since we already have the full-cell grid-fitted IBM it might be better to just focus on that for now and later modify it to make it partial-step when everything is working well (I think we'd just need to modify the calculation of vertical areas and volumes, no?).
> 
> Not sure what you mean in terms of "focus". I wasn't proposing any work, just speculating about the potential advantages of shaved cells. Do you mean focus on full cell IBM for this particular test? I agree with that --- we don't have shaved cells so we can't test it ?

Sorry for not being clear. I meant focus for development. For example, I see you're planning on implementing a more general BC interface for IBMs in https://github.com/CliMA/Oceananigans.jl/pull/2437. My point is that I think it would probably be easier to focus efforts on that kind of stuff first, make the full-step grid-fitted IBM (and its interface) functional, to only then try to improve the IBM implementation itself.

Hope that clarifies it.

validation/immersed_boundaries/flow_over_hills.jl
tomchor: > Is it ok to leave as is and just merge?

Totally. I already approved it :)

glwagner:
> > > Since we already have the full-cell grid-fitted IBM it might be better to just focus on that for now and later modify it to make it partial-step when everything is working well (I think we'd just need to modify the calculation of vertical areas and volumes, no?).
> > 
> > 
> > Not sure what you mean in terms of "focus". I wasn't proposing any work, just speculating about the potential advantages of shaved cells. Do you mean focus on full cell IBM for this particular test? I agree with that --- we don't have shaved cells so we can't test it ?
> 
> Sorry for not being clear. I meant focus for development. For example, I see you're planning on implementing a more general BC interface for IBMs in #2437. My point is that I think it would probably be easier to focus efforts on that kind of stuff first, make the full-step grid-fitted IBM (and its interface) functional, to only then try to improve the IBM implementation itself.
> 
> Hope that clarifies it.

Oh for sure, we hope to have that stuff merged soon. Shaved cells are eons away (unless some intrepid external contrib wants to focus it) --- we can't look at that within Climate Modeling Alliance until we have realistic global solutions.

Issue #884, opened 2020-08-29
Inconsistency in output writers interface


glwagner:
The `NetCDFOutputWriter` requires the keyword argument `filename`:

https://github.com/CliMA/Oceananigans.jl/blob/ed73ad36c99b61f24c37c73e4a2d26a12c7601d9/src/OutputWriters/netcdf_output_writer.jl#L147-L163

while `JLD2OutputWriter` requires `prefix` (which is optional, even though the default `prefix=""` would produce a file called `.jld2`:

https://github.com/CliMA/Oceananigans.jl/blob/fbbb5f81b851bb838c8796efba36a3021cb89f0e/src/OutputWriters/jld2_output_writer.jl#L56-L60

There are also some minor differences: `filename` is used for the NetCDF writer, while `filepath` is used for the JLD2 writer.

We should probably even out these differences:

* I think `filename` is more descriptive than `prefix`
* I also think its redundant to require the user to specify the file suffix (`.jld2` or `.nc`) in addition to specifying `JLD2` or `NetCDF` in the output writer name. So we need to weigh the two above considerations in coming to a decision. 
* `filepath` may be a better name for the struct member if we are also going to allow the keyword argument `dir` (this is not part of `NetCDFOutputWriter` either)

In general I think we should strive to make all the keywords as identical as possible so that users can switch between the writers by changing `NetCDF` to `JLD2` and vice versa.

ali-ramadhan:
Yeah definitely agree the two should have a consistent interface. I'm also in favour of a single `filepath` keyword argument that encompasses both `filename` and `dir`.

glwagner:
@tomchor let's discuss here, we could use a refresh...

I think the main issue is to figure out how to merge the interface. JLD2 supports file "splitting" which is why a `prefix` and `dir` makes more sense than specifying the whole filepath. We could try to infer `prefix` and `dir` from a `filepath` too, just need to decide if we are ok that users write `filepath` but then the files end up in different places (because of the file splitting).

Changing `force` to `mode` seems totally fine.

tomchor:
> Changing `force` to `mode` seems totally fine.

Agreed.


[...]

PR #2416, opened 2022-04-06
Unifies interfaces in `NetCDFOutputWriter`, `JLD2OutputWriter` and `Checkpointer`

tomchor:
- [x] Changes `mode` and `force` to overwrite existing
- [x] Change `prefix` to something clearer (maybe `filename`)
- [x] Separate keyword `filepath` in `NetcdfOutputWriter` into `dir` and `filename` (or whatever is replacing `prefix`)
- [x] Maybe add a couple more tests for `NetcdfWriter`

Closes https://github.com/CliMA/Oceananigans.jl/issues/884
Closes https://github.com/CliMA/Oceananigans.jl/issues/2427


glwagner:
I agree with these!

I think `filename` (including the extension) is a good choice. I also agree its convenient to use `dir` and `filename` rather than `filepath`.

tomchor:
> I agree with these!
> 
> I think `filename` (including the extension) is a good choice. I also agree its convenient to use `dir` and `filename` rather than `filepath`.
> 
> For `JLD2`, we need to "validate" the `filename` to ensure that it ends in `jld2`. This is needed because we have to identify the "prefix" (the part without the extension) in order to do file splitting.

Ah, yes, I forgot about that...

So `JLD2` requires that the splitting is done with the `.jld2` extension, correct?

glwagner:
> > I agree with these!
> > I think `filename` (including the extension) is a good choice. I also agree its convenient to use `dir` and `filename` rather than `filepath`.
> > For `JLD2`, we need to "validate" the `filename` to ensure that it ends in `jld2`. This is needed because we have to identify the "prefix" (the part without the extension) in order to do file splitting.
> 
> Ah, yes, I forgot about that...
> 
> So `JLD2` requires that the splitting is done with the `.jld2` extension, correct?

No, we don't have to require that. But we do need to know what the extension is in order to produce new filenames. We can either "infer" the extension, or assume that it's `.jld2`. Do we want to allow other extensions? What's the benefit of that? It might be nice to enforce uniformity.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -287,22 +287,27 @@ function NetCDFOutputWriter(model, outputs; filepath, schedule,
...
+        if isfile(filepath) && !overwrite_existing
+            @warn "$filepath already exists and `overwrite_existing = false`. Mode will be set to append to existing file. " *
+                  "You might experience errors when writing output if the existing file belonged to a different simulation!"
 
-    elseif !isfile(filepath) && !overwrite_existing
-        @warn "File $filepath cannot be found nd `overwrite_existing = false`." *
-              "This may cause errors when creating $filepath"
+        elseif isfile(filepath) && overwrite_existing
+            @warn "$filepath already exists and `overwrite_existing = true`. " *
+                  "Please make sure this setup was intentional as file $filepath will be overwritten!"
```
glwagner, line 308:
```suggestion
            @warn "Overwriting existing $filepath."
```

tomchor:
Leaving the command I used for the previous commit here to make my life easier in the future:

`find . -type f -name "*.jl" -exec sed -i "s/\<force\> = true/overwrite_existing = true/g" {} \;`

tomchor:
Any reason why `AbstractOutputWriter` is defined in `Oceananigans.jl` and not in `OutputWriters.jl`?

https://github.com/CliMA/Oceananigans.jl/blob/b945655a8363f60d03077253edcd77a56936c5cc/src/Oceananigans.jl#L153-L158

docs/src/model_setup/output_writers.md
```diff
@@ -178,7 +177,7 @@ c_avg = Field(Average(model.tracers.c, dims=(1, 2)))
 
 # Note that model.velocities is NamedTuple
 simulation.output_writers[:velocities] = JLD2OutputWriter(model, model.velocities,
-                                                          prefix = "some_more_data",
+                                                          filename = = "some_more_data.jld2",
```
glwagner:
```suggestion
                                                          filename = "some_more_data.jld2",
```

glwagner: Similarly going down in this file.

tomchor: Oops. Thanks!

tomchor: Actually I'd fixed that already. I think you made a comment on an old version...

test/test_output_writers.jl
```diff
@@ -90,6 +90,40 @@ function test_dependency_adding(model)
     return nothing
 end
 
+function test_creating_and_appending(model, output_writer)
+
+    simulation = Simulation(model, Δt=1, stop_iteration=5)
+    output = fields(model)
+    filename = "test_caa.h"
```
glwagner, line 97: <3

tomchor:
> @tomchor are we checking that we use `jld2` extension? Either we need to check this or generalize this line:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/992475546fba59ad07e584c77d5592b1882d607a/src/OutputWriters/jld2_output_writer.jl#L303
> 
> to work with any extension, I think.

I think it's good to keep it as general as the `NetCDFWriter`, but I don't have a strong opinion. What's your vote?

test/test_output_writers.jl
```diff
@@ -88,27 +90,61 @@ function test_dependency_adding(model)
...
+
+    # Create a simulation with `overwrite_existing = true` and run it
+    simulation.output_writers[:writer] = output_writer(model, output,
+                                                       filename = filename,
+                                                       schedule = IterationInterval(1),
+                                                       overwrite_existing = true)
```
tomchor, line 103: Having both writers be extension-agnostic lets us write general stuff like this :)

glwagner:
> > @tomchor are we checking that we use `jld2` extension? Either we need to check this or generalize this line:
> > https://github.com/CliMA/Oceananigans.jl/blob/992475546fba59ad07e584c77d5592b1882d607a/src/OutputWriters/jld2_output_writer.jl#L303
> > 
> > to work with any extension, I think.
> 
> I think it's good to keep it as general as the `NetCDFWriter`, but I don't have a strong opinion. What's your vote?

I'm ok with generality, but just out of curiosity what is the purpose of that? It seems "nice" to enforce `.jld2` and `.nc` because it helps human communication (eg lots of software does this, from Microsoft Word to Adobe Lightroom), and it's a little easier for us to deal with in the code too. But if it's really helpful to allow other extensions I think that's fine.

test/test_output_writers.jl
glwagner: That's how most software works for example google docs:

image

gives `DocumentName.docx`, `DocumentName.odt`, etc.

tomchor:
> I'm ok with generality, but just out of curiosity what is the purpose of that? It seems "nice" to enforce `.jld2` and `.nc` because it helps human communication (eg lots of software does this, from Microsoft Word to Adobe Lightroom for various formats), and it's a little easier for us to deal with in the code too. It's common in other contexts too to infer information from the extension (eg when we "save" an image, the extension dictates the compression algorithm, format, etc). But if it's really helpful to allow other extensions I think that's fine.

I think mostly it's windows and mac that enforce these extensions. But I agree with your point about clarity and making coding easier.

test/test_output_writers.jl
tomchor: > I guess if we append the extension automatically then we'd write
> 
> ```julia
>     filename = "test_caa"
> ```
> 
> and attain equal generality?

I think if we're going to enforce the extension, it's best that we ask for the full filename and check if the extension is appropriate, rather than asking for the prefix and appending the extension. I think it's clearer for the user and makes for shorter code too.

src/OutputWriters/jld2_output_writer.jl
```diff
@@ -158,33 +158,36 @@ JLD2OutputWriter scheduled on TimeInterval(20 minutes):
...
-                                 force = false,
+                    overwrite_existing = false,
                                   init = noinit,
                              including = default_included_properties(model),
                                verbose = false,
                                   part = 1,
                                jld2_kw = Dict{Symbol, Any}())
 
+    # Enforce that extensions need to be `jld2`
+    @assert filename[end-4:end] == ".jld2" "`filename` needs to have `.jld2` extension"
```
glwagner, line 175:
```suggestion
    filename[end-4:end] == ".jld2" || throw(ArgumentError("`filename` needs to have `.jld2` extension"))
```

tomchor: What's wrong with assert? Do you want an `ArgumentError`?

glwagner: Just a little more friendly... `AssertionError` is sometimes mysterious (the purpose of `@assert` is to provide a correctness check that might be undesirable for performance, so gets removed at a certain optimization level)

glwagner: Here's the docstring for `@assert`:

```julia
help?> @assert
  @assert cond [text]

  Throw an AssertionError if cond is false. Preferred syntax for writing assertions. Message text is optionally displayed upon assertion failure.

  │ Warning
  │
  │  An assert might be disabled at various optimization levels. Assert should therefore only be used as a debugging tool and not used for authentication verification (e.g., verifying passwords), nor
  │  should side effects needed for the function to work correctly be used inside of asserts.
```

glwagner: `ArgumentError` means

```julia
help?> ArgumentError
search: ArgumentError

  ArgumentError(msg)

  The parameters to a function call do not match a valid signature. Argument msg is a descriptive error string.
```

Here we are telling the user that the argument `filename` is incorrect.

glwagner: You might envision here that somebody attempts to run a model at `O3` optimization, with a bug in `filename`. Then hours into the simulation they get an error (maybe) because we assume `.jld2`.

tomchor:
If the tests pass, I think this is ready to merge. Took the liberty to bump minor version since this is a breaking API change.

docs/src/model_setup/output_writers.md
```diff
@@ -21,8 +21,7 @@ The `Checkpointer` is discussed on a separate documentation page.
    periodic output according to the simulation time, simulation interval, or "wall time" (the physical time
    according to a clock on your wall). A fourth `schedule` called `AveragedTimeInterval` specifies
    periodic output that is time-averaged over a `window` prior to being written.
-4. The filename and directory. Currently `NetCDFOutputWriter` accepts one `filepath` argument, while
-   `JLD2OutputWriter` accepts a filename `prefix` and `dir`ectory.
+4. The `filename` and `dir`ectory.
```
glwagner, line 24: <3

examples/eady_turbulence.jl
```diff
@@ -292,8 +292,8 @@ u, v, w = model.velocities # unpack velocity `Field`s
 
 simulation.output_writers[:fields] = JLD2OutputWriter(model, (; ζ, δ),
                                                       schedule = TimeInterval(4hours),
-                                                        prefix = "eady_turbulence",
-                                                         force = true)
+                                                        filename = "eady_turbulence.jld2",
+                                                         overwrite_existing = true)
```
glwagner, line 296:
```suggestion
                                                        filename = "eady_turbulence.jld2",
                                                        overwrite_existing = true)
```

test/regression_tests/hydrostatic_free_turbulence_regression_test.jl
```diff
@@ -77,8 +77,8 @@ function run_hydrostatic_free_turbulence_regression_test(grid, free_surface; reg
         simulation.output_writers[:fields] = JLD2OutputWriter(model, outputs,
                                                               dir = directory,
                                                          schedule = IterationInterval(stop_iteration),
-                                                           prefix = output_prefix,
-                                                            force = true)
+                                                         filename = output_filename,
+                                                            overwrite_existing = true)
```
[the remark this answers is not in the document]
glwagner: Actually there's some alignment pain in this expression that I can't resolve via suggestions

glwagner (approved):
Looks great!

Main two questions are:

1. Should we enforce `.nc` in `NetCDFOutputWriter`? I think it's nice to enforce this.
2. Related to 1., we may want to provide an interface that omits the extension entirely (eg `filename` has no extension). This leads to an interface that is truly format-independent --- just change `JLD2` to `NetCDF` (and the file extension is even changed for you, which is nice I think!)

Both can be resolved later. Btw the bump is very important because this fixes broken docs!

tomchor:
> 1. Should we enforce `.nc` in `NetCDFOutputWriter`? I think it's nice to enforce this.

I think the benefit of enforcing it for NetCDF is less clear than for JLD2, so I chose to not place any requirements that NCDatasets.jl doesn't. But again I don't have a strong opinion here.

> 2. Related to 1., we may want to provide an interface that omits the extension entirely (eg `filename` has no extension). This leads to an interface that is truly independent --- just change `JLD2` to `NetCDF` and you change the file format (and the extension is updated for you, which is nice!)
 
Yeah I like that. Although then it wouldn't be able to be called `filename` imo, since you're not giving the filename anymore :)

> Both can be resolved later. Btw the bump is very important because this fixes broken docs!

Can we deal with these in other PRs though? I won't be able to work on these anymore for a couple of weeks and I don't want this PR to go stale (especially because it has doc fixes, like you mentioned).

glwagner:
> Although then it wouldn't be able to be called filename imo, since you're not giving the filename anymore :)

What do you mean? What's your definition of "name"?

PR #2417, opened 2022-04-06
Updates Benchmarks environment

glwagner:
`benchmarks/Manifest.toml` had CUDA updated to 3.7 which caused problems when benchmarking Oceananigans restricted to 3.3.6. This PR adds a `compat` to `benchmarks/Project.toml` and downgrades `CUDA`.

I think maybe a better solution would be to add Oceananigans explicitly to `benchmarks/Project.toml`. But then to run the benchmarks on the current repo we'd have to update the environment every time.

### DIFF
--- a/benchmark/Manifest.toml
+++ b/benchmark/Manifest.toml
@@ -24,10 +24,10 @@ uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
 uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
 
 [[BFloat16s]]
-deps = ["LinearAlgebra", "Printf", "Random", "Test"]
-git-tree-sha1 = "a598ecb0d717092b5539dbbe890c98bac842b072"
+deps = ["LinearAlgebra", "Test"]
+git-tree-sha1 = "4af69e205efc343068dc8722b8dfec1ade89254a"
 uuid = "ab4f0b2a-ad5b-11e8-123f-65d77653426b"
-version = "0.2.0"
+version = "0.1.0"
 
 [[BSON]]
 git-tree-sha1 = "ebcd6e22d69f21249b7b8668351ebf42d6dc87a1"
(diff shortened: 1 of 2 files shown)


tomchor:
> `benchmarks/Manifest.toml` had CUDA updated to 3.7 which caused problems when benchmarking Oceananigans restricted to 3.3.6. This PR adds a `compat` to `benchmarks/Project.toml` and downgrades `CUDA`.
> 
> I think maybe a better solution would be to add Oceananigans explicitly to `benchmarks/Project.toml`. But then to run the benchmarks on the current repo we'd have to update the environment every time.

Is it possible to automatically pull compat entries from the repo's `Project.toml`? (If it is, should we?)

tomchor:
Maybe a better question (which might be silly): is there any reason why benchmarks need to have a different benchmark from the main repo?

glwagner:
> Maybe a better question (which might be silly): is there any reason why benchmarks need to have a different benchmark from the main repo?

Because our benchmarking scripts have additional dependencies that are unrelated to the source code, like `BenchmarkTools`, `BSON`, `PyPlot`, `PrettyTables`, `PkgBenchmark`, `Plots`, DataFrames`.

tomchor:
> It's possible to do anything really, the `.toml` files are ordinary text files. The second question is the important one --- what's the best way to do this that causes least headaches.

Yeah, my question was in the sense of "is there Julia functionality to do that". If it requires us to write a script to manipulate the text, then I agree that it'd be too much of a headache...

glwagner:
> > It's possible to do anything really, the `.toml` files are ordinary text files. The second question is the important one --- what's the best way to do this that causes least headaches.
> 
> Yeah, my question was in the sense of "is there Julia functionality to do that". If it requires us to write a script to manipulate the text, then I agree that it'd be too much of a headache for something rather hacky...

I guess what `Pkg` provides is: if we add `Oceananigans` as a dependency, compat for `Oceananigans` is imposed on `benchmarks/Project.toml`. But this complicates the purpose of the benchmarks, which is to execute the local code.

Maybe we can add lines at the top of the benchmark script like

```julia
using Pkg; Pkg.dev("..")
```

Then we add the local version of Oceananigans to the `Project.toml` before the script is executed.

PR #2420, opened 2022-04-06
Bickley jet, continued

glwagner:
This PR splits `validation/bickley_jet.jl` into two scripts (one for the "immersed bickley jet" and one for the "ordinary" bickley jet) and somewhat improves the script for the ordinary bickley jet.

@simone-silvestri I left the "immersed bickley jet" unchanged so you can continue to use it as is.


francispoulin:
> And it's still the same ol Bickley jet after al
> 
>  bickley_jet_Nh_128_WENO5.mp4

And still looking good!

francispoulin (approved):

PR #2426, opened 2022-04-09
Semi-implicit treatment of TKE dissipation for CATKE

glwagner:
This PR extends the capability of the "implicit diffusion solver" to handle an "additional linear term". This feature is then used to treat the TKE dissipation term semi-implicitly in attempt to alleviate CATKE's time-step restrictions.

I use quotes because it's not exactly a "diffusion solver" anymore. So a little name shuffling might be in order if this turns out to work well.

Thanks to @yairchn for help in formulating this potential solution for CATKE's inclination toward instability.

### DIFF
--- a/benchmark/Manifest.toml
+++ b/benchmark/Manifest.toml
@@ -24,10 +24,10 @@ uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
 uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
 
 [[BFloat16s]]
-deps = ["LinearAlgebra", "Printf", "Random", "Test"]
-git-tree-sha1 = "a598ecb0d717092b5539dbbe890c98bac842b072"
+deps = ["LinearAlgebra", "Test"]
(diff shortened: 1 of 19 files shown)


glwagner:
I think this is working!

These simulations were all run with a 10 minute time step:

https://user-images.githubusercontent.com/15271942/162590019-f990bda8-ca7d-4bf4-9f0d-5354b3d4d5fa.mp4

Before this PR, this case had to use a 30 second time-step for CATKE to run stably.

glwagner:
@simone-silvestri I refactored the tuple halo filling (partly to shorten the code but also to filter more stuff from the tuples, and to avoid filling halos for the same fields twice --- sometimes a field can appear twice in a named tuple...) But it looks like my changes broke some stuff, so I might need some help fixing that up.

glwagner:
@simone-silvestri @navidcy @sandreza I propose we merge this, and I will open an issue about cleaning up the names.

simone-silvestri (approved):

PR #2431, opened 2022-04-12
[WIP] Implement "MEKE" parameterization with prognostic mesoscale eddy kinetic energy

### LINKED ISSUE #2422: Interface for closures (and other model terms) that introduce auxiliary prognostic variables
@simone-silvestri were discussing implementing some new closures such as K-epsilon / "Generic Length Scale", and the OSMOSIS scheme used by the NEMO ocean model. Both these closures, and `CATKEVerticalDiffusivity`, introduce new prognostic variables (such as turbulent kinetic energy, dissipation, or mixed layer depth) that must be evolved alongside the model state.

For CATKE we have hacked in a turbulent kinetic energy equation that gets used when i) the tracer is named `:e` and ii) one of the closures is `CATKEVerticalDiffusivity`:

https://github.com/CliMA/Oceananigans.jl/blob/9e620637670cdd4defa32c07c51fd5574520de73/src/Models/HydrostaticFreeSurfaceModels/calculate_hydrostatic_free_surface_tendencies.jl#L99-L105

this works but the design is not sustainable. There are a few issues: first, we can't continue to abuse `model.tracers` to add new prognostic variables. Some prognostic variables shouldn't be called "tracers" --- like mixed layer depth. Also, we probably want a design in which `model.tracers` is unaffected by the closure being used. This will make it easier to switch between different closures in, for example, a large-scale ocean simulation. The second issue is that we need to design an interface so that the "prognostic dynamics" of a closure's auxiliary prognostic variable can be implemented and confined to the file where the closure is defined (ie, `HydrostaticFreeSurfaceModel` shouldn't know anything about `CATKE` in particular).

One solution is to add a new model property called `auxiliary_prognostic_fields`. This might behave similarly to `model.diffusivity_fields`; eg every `closure` can populate this object in the model constructor. Then we'll add tendency fields for those terms, and time-step them forward. The "tendency kernel function" that get's called will also have be provided by the closure.

Just to make things more logical, we could also rename `diffusivity_fields` to `auxiliary_diagnostic_fields` when we do this.

This would all have to come with an API for accessing the data. Right now we have `fields(model)`. We could divide this into two functions: `prognostic_fields(model)` and `diagnostic_fields(model)`, and then also `fields(model)` which combines the two. One decision there is whether to "flatten" the lists, or to maintain the hierarchical structure that we currently have in `diffusivity_fields`. 

I also think we need an interface for extracting mixing coefficients / diffusivities (rather than requiring that users spelunk into the model properties to find the information they want). We have a prototype with `viscosity` and `diffusivity` functions but it hasn't been used much and isn't really tested. So that would have to be expanded upon as well.

glwagner:
This PR implements the "MEKE" mesoscale parameterization proposed by [Jansen et al (2015)](https://www.sciencedirect.com/science/article/pii/S1463500315000967) (see also [Kong and Jansen (2020)](https://journals.ametsoc.org/view/journals/phoc/51/3/JPO-D-20-0142.1.xml)).

To implement this PR we add a property to `HydrostaticFreeSurfaceModel` called `auxiliary_prognostic_fields` which provides a container for closure-specific prognostic fields that need to be stepped forward alongside the rest of the model's prognostic state.

Another change is that we introduce `AbstractSkewSymmetricDiffusivity` that provides an interface for implementing mesoscale closures with skew and symmetric diffusivities.

The "MEKE" parameterization has a two-dimensional prognostic eddy kinetic energy variable that this feature supports.

It's WIP now, but a few notes are:

* We need a slightly more descriptive name than MEKE or `MesoscaleEddyKineticEnergy` I think (though it does pronounce well "mee-key")... maybe `PrognosticMEKEDiffusivity` or something? We may want to distinguish from (or better, combine with?) a similar parameterization with 2D prognostic MEKE called ["GEOMETRIC"](https://journals.ametsoc.org/view/journals/phoc/48/10/jpo-d-18-0017.1.xml)

* There's probably a way to improve the `auxiliary_prognostic_fields` design... in particular, I'm wondering if we should use a more hierarchical structure for auxiliary fields that also encompasses `diffusivity_fields`, something like

```julia
struct AuxiliaryFields
    user_defined
    diagnostic_closure_fields
    prognostic_closure_fields
end
```

The downside is that eddy diffusivities for LES are then buried in `model.auxiliary_fields.diagnostic_closure_fields`. BUT we can also design an interface for extracting these like `eddy_diffusivity(model)`. Curious what people think about that (@tomchor you've had opinions). Basically it's easier to separate the user API (here, functions that extract properties) from the model struct design (which is motivated more by internal considerations). But it's tradeoffs all the way down so good to discuss.

I also did a little clean up and changed the internal function `TurbulenceClosures.DiffusivityFields` to `TurbulenceClosures.diffusivity_fields`. The code is inconsistent about the use of TitleCase (is it a constructor / struct? is it a function?) and there's a bit of clean up to do...

Resolves #2422


glwagner:
@noraloose

NoraLoose:
Hi @glwagner, thanks for starting this PR! Sorry for chiming in so late - I'm finally getting started with Oceananigans.jl (but am already loving it :tada:).

I don't know the code well enough (yet!) to comment on your code-related questions above. But I can give some general input concerning the MEKE scheme.

1. **References**: The most up-to-date references for MEKE are [Jansen et al. (2019)](https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/2019MS001750) and [Kong and Jansen (2020)](https://journals.ametsoc.org/view/journals/phoc/51/3/JPO-D-20-0142.1.xml); in the latter mostly Appendices A b,c. 
2. **Naming convention**: The prognostic MEKE equation is used to inform the GM coefficient via `kappa_GM = c * sqrt(2 MEKE) * L_mix` with a mixing length `L_mix`, but also to energetically constrain the backscatter in the momentum equation. See this schematic, modified from Figure 1, Jansen et al. (2019):
MEKE_schematic
So how about `PrognosticMEKEDiffusivity` and `PrognosticMEKEBackscatter`? Probably good to keep these 2 separated to be able to switch on/off different terms in the MEKE budget. For instance, as of now, MOM6 OM4 only uses the `PrognosticMEKEDiffusivity` (and no backscatter), but the CPT and others are looking into the energetically-constrained backscatter term intensively these days!
3. **MEKE vs. GEOMETRIC**: As you mention above, the two schemes are very similar. There are maybe three notable differences: 
  a) **energy type**: MEKE solves a prognostic equation for `EKE`, whereas GEOMETRIC solves a prognostic equation for `EKE + EPE`. To inform GM, the two schemes then use the variable `2 * EKE` and `EKE + EPE`, respectively. [Bachman (2017)](10.1016/j.ocemod.2016.12.004) found that in practice, kappa_GM is not sensitive to whether one solves for `EKE` (as MEKE does) or `EKE + EPE` (as GEOMETRIC does). So we could probably start by following the MEKE approach, and just solve for `EKE`.
  b) **GM/energy source term**: In MEKE, it is left to the user how to specify the mixing length scale `L_mix` in `kappa_GM = c * sqrt(2 MEKE) * L_mix` and in the inferred MEKE source term. (Options are: the deformation scale, Rhines scale, grid scale, frictional halting scale, Eady length scale, or any combination of those, see [here](https://github.com/NOAA-GFDL/MOM6/blob/0e8acd90b46656ccb7fbc73bc5911ef2733a345c/src/parameterizations/lateral/MOM_MEKE.F90#L995-L1012).) In GEOMETRIC, the mixing length scale is essentially chosen as the Eady length scale. So if you use MEKE with the Eady length scale `L_e` (see Appendix Ab in Kong and Jansen, 2021), you recover the GEOMETRIC source term.
  c) **sub-grid dissipation**: MEKE assumes only bottom drag, GEOMETRIC uses one "catch-all" linear damping term as the energy sink term in the sub-grid energy budget. As highlighted in [Mak et al. (2022)](https://onlinelibrary.wiley.com/doi/abs/10.1029/2021GL097259), this sub-grid dissipation term is probably the big elephant in the room, which the solution is very sensitive to but noone knows what the term should look like.

glwagner:
@NoraLoose my turn to apologize for a very slow response --- thank you for those explanations, they are very helpful! I think this PR is going a bit stale so we may have to restart it, but when that happens I'll port your comments over to the new PR. Probably we need to wait for #2477, which among other things implements a more stable discretization for GM+Redi a la [Griffies et al 1998](https://journals.ametsoc.org/view/journals/phoc/28/5/1520-0485_1998_028_0805_idiazc_2.0.co_2.xml).

glwagner:
It's way stale!

PR #2432, opened 2022-04-12
Updates `distributed_nonhydrostatic_model_mpi.jl`

tomchor:
This PR tries to make the script a bit easier to run based on my experiences with it. For example `Benchmark` and `@gpu_sync` aren't needed in this case and we can delete it.

So far I've only been able to run this with one core (see https://github.com/CliMA/Oceananigans.jl/issues/2433).

Closes https://github.com/CliMA/Oceananigans.jl/issues/2433

### DIFF
--- a/benchmark/Manifest.toml
+++ b/benchmark/Manifest.toml
@@ -515,7 +515,7 @@ uuid = "38a345b3-de98-5d2b-a5d3-14cd9215e700"
 version = "2.36.0+0"
 
 [[LinearAlgebra]]
-deps = ["Libdl", "libblastrampoline_jll"]
+deps = ["Libdl"]
 uuid = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 
 [[LogExpFunctions]]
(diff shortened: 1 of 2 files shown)


glwagner (approved):

francispoulin:
Thanks @tomchor for doing this.

I remember last year, Henry needed to use `@gpu_sync` when he was running stuff on multiple cores.  Maybe why it is not necessary for a single core?

I presume you are going to try and run this on more than one core?  If you try it and get an error I'd be curious to know what it is.

tomchor:
> Thanks @tomchor for doing this.
> 
> I remember last year, Henry needed to use `@gpu_sync` when he was running stuff on multiple cores. Maybe why it is not necessary for a single core?

I think in this case it's not necessary because it's a multi-CPU testing script, so no GPUs involved. And that allows us to get rid of the `Benchmarks` dependence and make the script more portable.
 
> I presume you are going to try and run this on more than one core? If you try it and get an error I'd be curious to know what it is.

I did! And I'm running into issues. Please check https://github.com/CliMA/Oceananigans.jl/issues/2433! Any help there is appreciated

PR #2435, opened 2022-04-13
Updates output writers interface to append extension automatically

glwagner:
cc @tomchor

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.74.1"
+version = "0.74.2"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.74.0"
+version = "0.74.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/OutputWriters/jld2_output_writer.jl
+++ b/src/OutputWriters/jld2_output_writer.jl
@@ -172,17 +172,16 @@ function JLD2OutputWriter(model, outputs; filename, schedule,
                                   part = 1,
                                jld2_kw = Dict{Symbol, Any}())
 
-    base_filepath = auto_extension(filename, ".jld2")
-
+    mkpath(dir)
+    filename = auto_extension(filename, ".jld2")
+    filepath = joinpath(dir, filename)
+    overwrite_existing && isfile(filepath) && rm(filepath, force=true)
+    
     outputs = NamedTuple(Symbol(name) => construct_output(outputs[name], model.grid, indices, with_halos)
                          for name in keys(outputs))
 
     # Convert each output to WindowedTimeAverage if schedule::AveragedTimeWindow is specified
     schedule, outputs = time_average_outputs(schedule, outputs, model)
-
-    mkpath(dir)
-    filepath = joinpath(dir, base_filepath)
-    overwrite_existing && isfile(filepath) && rm(filepath, force=true)
 
     initialize_jld2_file!(filepath, init, jld2_kw, including, outputs, model)
     
--- a/src/OutputWriters/jld2_output_writer.jl
+++ b/src/OutputWriters/jld2_output_writer.jl
@@ -57,10 +57,10 @@ Keyword arguments
 
   ## Filenaming
 
-  - `filename` (required): Descriptive filename. ".jld2" is appended to `filename` in the file path.
+  - `filename` (required): Descriptive filename. ".jld2" is appended to `filename` in the file path
+                           if `filename` does not end in ".jld2".
 
-  - `dir`: Directory to save output to.
-           Default: "." (current working directory).
+  - `dir`: Directory to save output to. Default: "." (current working directory).
 
   ## Output frequency and time-averaging
 
--- a/src/OutputWriters/jld2_output_writer.jl
+++ b/src/OutputWriters/jld2_output_writer.jl
@@ -57,7 +57,7 @@ Keyword arguments
 
   ## Filenaming
 
-  - `filename` (required): Descriptive filename including extension.
+  - `filename` (required): Descriptive filename. ".jld2" is appended to `filename` in the file path.
 
   - `dir`: Directory to save output to.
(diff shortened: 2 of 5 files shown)


tomchor:
I think I still prefer to not append the extension and leave things the way they are mainly for three reasons:

- If you don't include the extension, then (using [usual jargon](https://www.computerhope.com/jargon/f/filename.htm)) it's not a complete file name anymore since it doesn't uniquely identify the file. To be correct I think we'd need to go back to `prefix` which isn't super clear imo
- It leads to less clear code if you're postprocessing the output in the name script. Like this pseudo code:

```julia
simulation.output_writers[:writer] = JLD2OutputWriter(model, outputs, filename=filename)

run!(simulation)
real_filename = filename * ".jld2"
jldopen(real_filename)
```

tomchor:
Also what we'd call `filename` would be different than what is called file name in the docs for [`JLD2`](https://docs.juliahub.com/JLD2/O1EyT/0.4.22/internals/#JLD2.jldopen) and [`NCDatasets`](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#NCDatasets.NCDataset-Tuple{NCDatasets.CFVariable}), which might be a source of confusion in new users.

glwagner:
> * If you don't include the extension, then (using [usual jargon](https://www.computerhope.com/jargon/f/filename.htm)) it's not a complete file name anymore since it doesn't uniquely identify the file.

That link you provided says

> A file name may also describe only the first portion of the file.

glwagner:
> * Like this pseudo code:
> 
> ```julia
> simulation.output_writers[:writer] = JLD2OutputWriter(model, outputs, filename=filename)
> 
> run!(simulation)
> real_filename = filename * ".jld2"
> jldopen(real_filename)
> ```

I appreciate this feedback but need to understand --- how is this code "unclear"?

I think there _is_ a detriment to convenience --- which we have been accepting (without any loss of productivity, if we can be real...) since time eternal with the `prefix` kwarg, before #2416). I do think it'd be nice to solve this with `FieldTimeSeries` so we can write

```julia
simulation.output_writers[:writer] = JLD2OutputWriter(model, outputs, filename=filename)
run!(simulation)
u = FieldTimeSeries(filename, "u")
```

This is easy...

Balanced against the convenience detriment, I think the _cost_ of allowing arbitrary extensions is relatively higher (sharing data, figuring out the format of other people's data...). We also may have many more output writers in the future (Zarr, BSON, whatever), and I think it's friendly to relieve users from the burden of changing the filename.

glwagner:
The other possibility which achieves the same goal (and which I'm fine with) is to enforce `.nc` in `filename` for `NetCDFOutputWriter`. We'd have to change a lot fewer tests for that too...

glwagner:
Also to elaborate a bit: file extensions are _always_ arbitrary. The extension is just for people reading the contents of a directory, so they know what kind of data is contained in a file. For example this is fine:

```bash
$ echo 'println("Hello world!")' >> test.txt                                                                                                                                        
$ julia --project test.txt                                                                                                                                                          [11:28:26]
Hello world!
```

It's just poor communication to put your julia code in files appended with `.txt`. And your text editor won't highlight it correctly 😱 

So I think we really should somehow enforce extensions.

tomchor:
> > * Like this pseudo code:
> > 
> > ```julia
> > simulation.output_writers[:writer] = JLD2OutputWriter(model, outputs, filename=filename)
> > 
> > run!(simulation)
> > real_filename = filename * ".jld2"
> > jldopen(real_filename)
> > ```
> 
> I appreciate this feedback but can you clarify why this code is "unclear"?
 
Maybe unclear isn't the right word to use here, but I'm just pointing out that you'd end up with two things that a human would call a file name: `filename` (without extension) and `real_filename` (with extension).

tomchor:
> The other possibility which achieves the same goal (and which I'm fine with) is to enforce `.nc` in `filename` for `NetCDFOutputWriter`. We'd have to change a lot fewer tests for that too...

I personally would prefer that option instead of having the extension be automatically appended. Mostly because the name you pass to the writer (`something.jld2` or whatever it is) uniquely identifies the file that gets generated on your computer.

tomchor:
This may be beyond the point of this PR, but I feel like lots of these discussions that ultimately mostly impact user interface and experience would hugely benefit from, well, user input! 

This fact, coupled with the fact that Oceananigans probably now has a decently-sized user-base, may mean that we could start using polls and directly asking users about user-interface decisions. That way we can get more hard data on what makes the code easier and clearer to most users and weight that more accurately against other aspects of the decision.

Just a thought...

francispoulin:
I have read through this discussion and don't have a strong feeling either way.  Whether the whole file name or the root file name is called the `file name`, is a matter of semantics.  

As a user, I would ideally just want to define the root file name and then have the ending added as need be.

I realize this means that when we read the file in, what we called the filename is not truly the file name, and we need to add the ending.  I have not been bothered by this and am happy to do whatever is easier for the average user.

simone-silvestri:
My two cents on this:

1) I think that output writers and readers always have to work with the correct extension (jld2 with `.jld2` and netcdf with `.nc`) and this should be always enforced 
2) It would be a good idea in my opinion to support either options of passing the prefix or the filename (as many softwares do) so that we get the best of both options while always preserving the extension which I believe is our priority. (Just implement a check on the last relevant characters, if there is a correct extension leave it as is otherwise append it)

glwagner:
> 2\. It would be a good idea in my opinion to support either options of passing the prefix or the filename (as many softwares do) so that we always preservw the extension which I believe is our priority. (Just implement a check on the last relevant characters, if there is a correct extension leave it as is otherwise append it)

I guess we can do

```julia
dotindex = findlast('.', filename)
possible_ext = isnothing(dotindex) ? "" : filename[dotindex+1:end]
possible_ext == ".nc" || filename *= ".nc"
```

We can put similar logic into `FieldTimeSeries`.

tomchor:
> > 2. It would be a good idea in my opinion to support either options of passing the prefix or the filename (as many softwares do) so that we always preservw the extension which I believe is our priority. (Just implement a check on the last relevant characters, if there is a correct extension leave it as is otherwise append it)
> 
> I think this is an excellent suggestion!
> 
> We can do
> 
> ```julia
> dotindex = findlast('.', filename)
> possible_ext = isnothing(dotindex) ? "" : filename[dotindex+1:end]
> possible_ext == ".nc" || filename *= ".nc"
> ```

Agreed. I think that's a great compromise.

glwagner:
@tomchor do you approve? Are there any further changes you'd like to see ?

tomchor:
> @tomchor do you approve? Are there any further changes you'd like to see ?

I was waiting for the checks to pass to review. Let me review it right now.

tomchor (approved):
lgtm!

glwagner:
> > @tomchor do you approve? Are there any further changes you'd like to see ?
> 
> I was waiting for the checks to pass to review. Let me review it right now.

Got it. PR is complete so only minor debugging left (if anything -- every test passed except the one that used a `.jld2` filename)

Issue #1277, opened 2020-12-16
Possible inconsistency in the eddy definition of diffusivities/closures in the docs


tomchor:
Maybe I'm missing something, but when I look at the equations in the [closures](https://clima.github.io/OceananigansDocumentation/stable/numerical_implementation/turbulence_closures/) page it seems that `ν_eddy` does not include the molecular viscosity `ν`. However looking at this [closure page](https://clima.github.io/OceananigansDocumentation/stable/physics/turbulence_closures/) in the model physics it seems like the molecular viscosity is included in the Smagorisnky and AMD eddy viscosities.

Looking at `model.diffusivities.ν_e` it seems like that the lowest value is always the molecular (background) viscosity, so I'm guessing that the convention used by the model is the latter one. Should those be consistent with each other?

Thanks!

glwagner:
You're right that the molecular viscosity is added to `νₑ`:

https://github.com/CliMA/Oceananigans.jl/blob/81db22f4a26396142e8cd5b5a4c50c75790c1d50/src/TurbulenceClosures/turbulence_closure_implementations/verstappen_anisotropic_minimum_dissipation.jl#L122-L136

It seems nicer to me to keep these separate in the docs, but it's less expensive / convenient to combine them in the code.

We could define `νˢᵍˢ` and `νₑ = νˢᵍˢ + ν`.

It's also been pointed out that making the default value of `ν` to be a particular number for a particular temperature water is weird. We should probably have default `ν = 0`.

glwagner:
@tomchor I think you should feel free to edit the docs (and also make other appropriate clarifying changes if they are appropriate). I think we can call `νₑ` the "effective diffusivity" and `νˢᵍˢ` the subgrid or subfilter diffusivity.

tomchor:
> We could define `νˢᵍˢ` and `νₑ = νˢᵍˢ + ν` in the docs to solve this issue.

On the surface I definitely agree with this. Although I prefer to keep calling `ν_e` the eddy diffusivity, since that's what everyone calls it and it's supposed to model the action of eddies. The caveat here is that this doesn't match the convention that the code uses, right? Which I think can be pretty confusing.

> It's also been pointed out that making the default value of `ν` to be a particular number for a particular temperature water is weird. We should probably have default `ν = 0`.

This seems like a matter of preference, but I kinda like that the default is a reasonable value. Saves me the trouble of looking up what the molecular values are if I'm simulating something :)

[...]

PR #2437, opened 2022-04-14
Support boundary conditions on immersed boundaries

glwagner:
This PR will implement support boundary conditions on immersed boundaries.

The design is to add a new function to the tendency kernel function for each field that represents the contribution to the viscous / diffusive flux divergence in boundary-adjacent cells due to the user-defined immersed boundary condition.

This PR will support `Flux` boundary conditions for sure.

If it turns out to be simple, this PR will also support `Value` / `Gradient` boundary conditions for `ScalarDiffusivity`. There may be some challenges when `ScalarDiffusivity` is embedded in a closure tuple (ie the case we get for wall-resolved LES). Hopefully the solution to that is not too complicated.

Another problem to solve is how to "regularize" `ContinuousBoundaryFunction`. Immersed boundary conditions are 3D, so function of `i, j, k`. Thus we need a slightly different interface for them. `discrete_form=true` will also have a different signature (`i, j, k, grid, ...` rather than `i, j, grid, ...`).


whitleyv:
Everything useful that I have is probably in the draft PR #2275 and the associated branch! They aren't forcing functions, but we hard coded in some drag functions for the flux in the IB case that could be used, and the sandbox has the boundary condition versions of those same functions as a test. It wouldn't take much to turn them into divergences!

glwagner:
> #2275

Is there a forcing function that computes the "immersed boundary normal" direction with associated logic?

whitleyv:
> Is there a forcing function that computes the "immersed boundary normal" direction with associated logic?

Not exactly, but if you're determining a grid fitted normal direction then these boolean functions determine if the location is an interface with fluid on the left and solid on the right or vice versa for all the options. They are only true on a boundary face, not if its a solid-solid or fluid-fluid wall.

https://github.com/CliMA/Oceananigans.jl/blob/8b988b34abcfbfd2e0453aedec5916c22b8c35c7/src/ImmersedBoundaries/conditional_fluxes.jl#L48

I don't think it would be hard to use these to set a directional sign for the normal direction, depending on which way you're doing that. If you want the normal direction to be based on the true IB rather than grid fitted, then I don't have anything recent.

glwagner:
Thanks! 

> If you want the normal direction to be based on the true IB rather than grid fitted, then I don't have anything recent.

I believe with the methods we are using, we only ever need the normal direction of the "numerical" or discrete IB. Actually I think we should call _that_ the "true" IB; if there's a continuous function from which the discrete specification is derived, maybe we can call that the "IB generator"? Right not this "normal" direction is always aligned with a cardinal direction of the grid (east, west, etc...). In the future we _could_ entertain shaved cells, which would introduced IB faces with non-trivial normals (but still discrete, and calculable from the discrete representation assuming, for example, that the faces of the shaved cells are flat planes).

glwagner:
But I realized that a second way to implement this immersed flux is to re-use the conditional flux functions, but reverse the logic. Not sure what is best yet.

whitleyv:
> But I realized that a second way to implement this immersed flux is to re-use the conditional flux functions, but reverse the logic. Not sure what is best yet.

We did run into the problem with the conditional flux functions, that fluxes located on the same spot (like `fcc`), won't have the same faces to check all the time (`uy` will be checking the south-north interface and `vx` would be east-west). Depending on the function, the symmetric fluxes might not be assigned as the same either, esp. when it comes to interpolation. We just made new separate conditional flux functions for each viscous flux combo, but that might not be the way to go long term.

glwagner:
> > But I realized that a second way to implement this immersed flux is to re-use the conditional flux functions, but reverse the logic. Not sure what is best yet.
> 
> We did run into the problem with the conditional flux functions, that fluxes located on the same spot (like `_fcc`), won't be checking the same faces for an interface/ normal direction (`_uy` will be checking the south-north interface and `_vx` would be east-west). Depending on the function, the symmetric fluxes might not be assigned as the same either, esp. when it comes to interpolation. We just made new separate conditional flux functions for each viscous flux combo, but that might not be the way to go long term.

Right, we need an API for immersed fluxes. I think for continuous functions we can build objects under the hood for each possible location (so six `ContinuousBoundaryFunction` for east, west, south, north, top, bottom). For `discrete_form` we'll need to ponder because in that case to be correct the user must provide six functions.

glwagner:
Here's a summary of the current design:

I've introduced a new type called `ImmersedBoundaryCondition` to represent boundary conditions on immersed boundaries. The type has 6 fields for each direction, west, east, south, north, bottom, top.

I believe this covers the general case in which we might have different fluxes on any face of a cell that's the boundary between a wet node and an immersed node. For example, we can implement "bottom drag" for large scale ocean models, with no "side drag". In practice this is really just a performance optimization rather than a physical model (the horizontal drag has no effect on the solution), but it's probably important that we can support it. This low-level, but fully general interface allow advanced users to specify any kind of immersed flux / boundary condition they need to.

The next challenge is to design a convenient API that interprets user-input and builds `ImmersedBoundaryCondition` appropriately under the hood. For constant `ValueBoundaryCondition` (eg no-slip) or constant `FluxBoundaryCondition`, this is simple --- just copy the boundary condition into every direction. For `FluxBoundaryCondition` we orient the flux along the inward normal (eg positive for `west, south, bottom` and negative for `east, north, top`. This breaks existing convention for fluxes... but we really don't have a choice. We may want to change the convention for the _other_ boundaries eventually.

`ContinuousBoundaryFunction` is perhaps the most important case since we'll use this to implement drag boundary conditions. For this I believe we can use the existing infrastructure for `ContinuousBoundaryFunction` (eg we have a way to properly interpolate fields in a manner consistent with the application of drag boundary conditions, etc). It's functionality will have to be extended to accept functions of `i, j, k` (the way it works now is that the "third" index (normal to the boundary) is stored inside the object; for immersed boundary conditions we don't need that).

`DiscreteBoundaryFunction` won't be very convenient for most people for immersed boundary conditions, but we can treat it the same way we treat constant `Value` or `Flux`.

I wasn't planning to implement support for `Gradient` just yet, though this can be done too.

Finally, one remaining task is to define or generalize the existing `conditional_flux` so that it adds fluxes only on _immersed_ boundaries. It looks like right now it's too eager and will add fluxes on non-immersed boundaries. Associated with this development I'd like to refactor our naming scheme for the boolean tools we have. Right now these are `solid_node` and solid_interface`. I'd like to change `solid_node` to `external_node` (ie external to the active domain) and perhaps something liek `external_or_boundary_node` for `solid_interface` to represent _both_ nodes that are completely external, or which lie _on_ the boundary. A better name is welcome... Finally we want to have `boundary_node` which is true only for cells that are _on_ or _adjacent_ to the boundary.

glwagner:
Major progress here, and it looks like `ValueBoundaryCondition` for tracers may be working:

image

Since I had to dig deep into the immersed boundary implementation, I made a few more long-needed changes:

* Finally, we directly serialize `ImmersedBoundaryGrid`
* I changed `ImmersedBoundaryGrid.grid` (eminently confusing!) to `ImmersedBoundaryGrid.underlying_grid` to emphasize that there are "two grids" --- one underlying rectangular grid, and an overlying "immersed grid"
* I cleaned up the output writers utilities a bit as well.
* I built out and hopefully clarified out "node condition" functionality. There are now a few functions:
    - `external_cell` identifies whether a _tracer cell_ (or perhaps the "reference cell" --- the one located at center, center, center) is outside the `Bounded` domain (this has no effect in `Periodic` directions)
    - `external_node` identifies whether a _location_ is external to the `Bounded` domain. This can be used for cells _or_ cell interfaces. `external_node` is `false` for locations that "touch" wet cells.
    - `peripheral_node` identifies whether a location is either _on or external_ to the `Bounded` domain. So this intersects with `external_node` in the exterior, but additionally contains locations that touch wet cells.
    - `immersed_peripheral_node` identifies locations that are on or external to the immersed domain, but _not_ external to the underlying grid domain (we need this so that we don't apply immersed boundary conditions at all peripheral nodes).

I think there's still a fair amount of work to do on this interface but hopefully this is a good step forward. Here's a few more thoughts:

* It seems that we should really use `Bounded` topology to be synonymous with "impenetrable". This might help us disentangle a few issues (eg satisfying impenetrability, correct divergences, etc.) I think we are discovering that in some cases its better to hard code certain "assumed boundary conditions" directly into operators, such as impenetrable conditions (considering complex domains is the key to realize the right abstractions here, since there are many simple solutions for non-complex domains and no obvious way to distinguish their tradeoffs).
* Related to the above, when we support "open" boundaries, I think we may actually want an "Open" topology that functions similarly to the topologies we use for distributed models work (perhaps, in fact, we can reuse those abstractions). I think the design assumed (though it was never implemented) that we could use "open boundary condition" for open boundaries, rather than changing the topology. But with complex domains, it's helpful to _assume_ certain boundary conditions based on _grid topology_ (rather than boundary conditions applied to fields). ImmersedBoundaries are hard-coded as "effectively `Bounded` now, as well.

All that said I am pretty excited about this progress. I think the code and abstractions are shaping up nicely.

glwagner:
It's a joyous day...

https://user-images.githubusercontent.com/15271942/163684303-d05b97d6-69d6-447a-b214-0f618a780dfb.mp4

This puts a regular value boundary condition at the top, and an immersed value condition at the bottom (opposite signs). Looks good to me.

Next is to figure out the shenanigans to get `ContinuousBoundaryFunction` to work. Though we coudl merge this now / soon without that, since we can support no-slip with this and it also allows value boundary condition (on non-immersed boundaries) with immersed boundary grid.

tomchor:
Wow, awesome work!

Here are a few thoughts/comments:

> * I changed `ImmersedBoundaryGrid.grid` (eminently confusing!) to `ImmersedBoundaryGrid.underlying_grid` to emphasize that there are "two grids" --- one underlying rectangular grid, and an overlying "immersed grid"

:+1: much clearer

> * I built out and hopefully clarified out "node condition" functionality. There are now a few functions:
>   
>   * `external_cell` identifies whether a _tracer cell_ (or perhaps the "reference cell" --- the one located at center, center, center) is outside the `Bounded` domain (this has no effect in `Periodic` directions)
>   * `external_node` identifies whether a _location_ is external to the `Bounded` domain. We use the term "node" because this can be used for cells _or_ cell interfaces. `external_node` is `false` for locations that "touch" wet cells.

So `external_cell` is just a special case `external_node`, correct? It also seems to me that this for now assumes a full-step grid-fitted IBM, no?

>   * `peripheral_node` identifies whether a location is either _on or external_ to the `Bounded` domain. So this intersects with `external_node` in the exterior, but additionally contains locations that touch wet cells.

Not sure I understand this. Is this trying to check whether a location is _adjacent_ to an IBM in addition to a location that's far (i.e. more than one grid point away) from an IBM?

>   * `immersed_peripheral_node` identifies locations that are on or external to the immersed domain, but _not_ external to the underlying grid domain (we need this so that we don't apply immersed boundary conditions at all peripheral nodes).

Also a bit confused. Sounds to me that this is the same as the previous one but excluding halo regions, which probably isn't right...

>   * `immersed_cell` is now the name of the function `is_immersed` --- immersed boundary implementations define this to obtain correct logic for boundary conditions. We then extend `external_cell` to cover both the external cells of the underlying bounded domain and the immersed cells.
> 
> I think there's still a fair amount of work to do on this interface but hopefully this is a good step forward. Here's a few more thoughts:

Quick question: Have you checked the performance here? I ask because the model was running pretty slowly when I was testing https://github.com/CliMA/Oceananigans.jl/pull/2275...

glwagner:
The concept of "external cell" is independent of the immersed boundary implementation. For _any_ grid that's bounded in one or more direction, with or without immersed boundary, there are cells that are "external" to the prognostic state.

The difference between "full cell" and other immersed boundaries does not lie with any of these Boolean functions. The differences instead lie in how grid metrics (and also cell face normal vectors) are modified in cells on the periphery of the "internal" domain.

For example, a partial cell method changes the vertical spacing in the bottom most cell. But this "modified" cell is _not_ external. Otherwise, it would not matter that we modified its metrics / geometry. The modified cell is part of the prognostic state.

glwagner:
I haven't looked at performance / GPU compilation yet, but that will happen before it's merged. I do think there is a type inference issue somewhere. But we have a bag of tricks to fix it so hopefully it won't be too hard.

glwagner:
Peripheral node can be used for conditioning fluxes. For example vorticity has non trivial values in the periphery, and is part of the domain. However a flux evaluated at the same location represents a flux across a boundary. So we need different functions for masking fields than for conditioning fluxes.

External cell is identical to external node evaluated at tracer cell centers.

glwagner:
Immersed peripheral node isolates the peripheral nodes exclusive to the immersed boundary and excludes peripheral nodes of the underlying grid. We need this because the immersed boundary condition applies only to immersed peripheral nodes and not all peripheral nodes.

glwagner:
Some progress:

The type infererence issue seems to be fixed. For `flow_over_hills.jl` the cost per time-step is approximately the same on my CPU for the "reference case" (no immersed boundary), and either free-slip or no-slip immersed boundary.

I updated `flow_over_hills.jl` to introduce an initial perturbation that goes turbulent immediately.

Here's a visualization of vorticity alongside total momentum loss for a 64x64 resolution:

https://user-images.githubusercontent.com/15271942/163830810-8f15244b-27e2-4bf0-a7b2-e35d73bd8c79.mp4

The "reference" case has a no-slip boundary but no bathymetry.

The momentum loss between reference and free-slip case are similar. No-slip and immersed boundary has the most momentum loss.

The effect of the pressure errors can be seen plainly that there is transient _injection_ of momentum into the system, which isn't physically possible. However over long periods of time the momentum goes in the right direction.

I'm wondering if anyone has experience with PALM, which I believe uses a very similar immersed boundary method (with no pressure solver correction)... ? They must have similar "leakage" errors.

Overall I'm happy. I'd like to implement a few more convenient features and then I think this implementation is complete.

There's still a fair amount of work to do to write tests and documentation.

francispoulin:
@glwagner : is it the separation of vorticity from the topography in the two immersed boundary cases that are the transient _injections_ that you are referring to?

If we were to try this with the `ShallowWaterModel`, where instead of having flow over a bumpy bottom, we had flow along a bumpy coastline, we would not get pressure errors since we don't solve for the pressure.  How difficult would it be to set this up for that scenario?  I'd be happy to run it and see what happens but it would take em a while to set things up.  

The motvation is to assess how much of this is due to pressure errors versus other stuff.

tomchor:
> > Not sure I understand this. Is this trying to check whether a location is adjacent to an IBM in addition to a location that's far (i.e. more than one grid point away) from an IBM?
> 
> Are you confused by the _definition_, or the way it's used? `peripheral_node` and `immersed_peripheral_node` are needed to condition fluxes. `external_node` is needed for masking. For example, vorticity is valid _on_ the boundary, and thus peripheral values of vorticity should not be masked (I don't think...). But we do need to condition fluxes on peripheral nodes.

I was a bit confused as to both. But your comments clarified quite a bit, thanks! I think the rest of my doubts will be answered once I have time to read the code more carefully...

tomchor:
> @glwagner : is it the separation of vorticity from the topography in the two immersed boundary cases that are the transient _injections_ that you are referring to?

@francispoulin I think he's talking about the fact that the graph on the right (the domain average/integrated momentum) isn't monotonic for IBM cases. Overall it has the correct tendency (loss of momentum) but has small scale fluctuations that aren't physical (i.e. there's no physical mechanism increasing the avg momentum in the domain, so the fluctuations must be numerical in origin).

tomchor:
> The effect of the pressure errors can be seen plainly that there is transient _injection_ of momentum into the system, which isn't physically possible. However over long periods of time the momentum goes in the right direction.
> 
> I'm wondering if anyone has experience with PALM, which I believe uses a very similar immersed boundary method (with no pressure solver correction)... ? They must have similar "leakage" errors.

I can't help there, but I agree it's a good idea. For now I think we should test that these errors are decreasing reasonably fast with resolution. (But that can be done in another PR I think.)

glwagner:
> @glwagner : is it the separation of vorticity from the topography in the two immersed boundary cases that are the transient _injections_ that you are referring to?
> 
> If we were to try this with the `ShallowWaterModel`, where instead of having flow over a bumpy bottom, we had flow along a bumpy coastline, we would not get pressure errors since we don't solve for the pressure. How difficult would it be to set this up for that scenario? I'd be happy to run it and see what happens but it would take em a while to set things up.
> 
> The motvation is to assess how much of this is due to pressure errors versus other stuff.

I'm looking at the time-series of total momentum --- the line goes down, but also _up_! So momentum is _increasing_ inside the domain, transiently.

I didn't try to connect this with the vorticity field but maybe you can draw some conclusions!

tomchor:
> The type infererence issue seems to be fixed. For `flow_over_hills.jl` the cost per time-step is approximately the same on my CPU for the "reference case" (no immersed boundary), and either free-slip or no-slip immersed boundary.

That's awesome! Just out of curiosity (since I also investigated this I couldn't figure it out), how did you fix the type inference issue?

cc @wenegrat @whitleyv

glwagner:
> @glwagner : is it the separation of vorticity from the topography in the two immersed boundary cases that are the transient _injections_ that you are referring to?
> 
> If we were to try this with the `ShallowWaterModel`, where instead of having flow over a bumpy bottom, we had flow along a bumpy coastline, we would not get pressure errors since we don't solve for the pressure. How difficult would it be to set this up for that scenario? I'd be happy to run it and see what happens but it would take em a while to set things up.
> 
> The motvation is to assess how much of this is due to pressure errors versus other stuff.

I think it would work flawlessly in that case!

This PR is all about viscous fluxes across immersed boundaries, so when we implement a viscous closure for `ShallowWaterModel` we can try to setup flow around islands or something and do a similar analysis as we've done here!

src/ImmersedBoundaries/conditional_fluxes.jl
```diff
@@ -16,17 +15,18 @@ Return either
 
 This can be used either to condition intrinsic flux functions, or immersed boundary flux functions.
 """
-@inline conditional_flux(i, j, k, ibg, ℓx, ℓy, ℓz, qᴮ, qᴵ, nc=ipn) = ifelse(nc(ℓx, ℓy, ℓz, i, j, k, ibg), qᴮ, qᴵ)
+@inline conditional_flux(i, j, k, ibg, ℓx, ℓy, ℓz, qᴮ, qᴵ) =
+    ifelse(immersed_peripheral_node(ℓx, ℓy, ℓz, i, j, k, ibg), qᴮ, qᴵ)
```
glwagner, line 19: @tomchor this is the key change to fix type inference (and then there is a cascade of associated changes in all the functions that use this operator).

Early in this PR I generalized `conditional_flux` to take an final optional positional argument `nc` (stands for `node_conditional`) to which one might pass either `immersed_peripheral_node` or `peripheral_node`.

I believe what was happening is that the compiler then refused to _specialize_ on this final argument (not sure why --- sometimes the compiler refuses to specialize on _function arguments_ because every function is it's own type, meaning that the relative advantage of specializing on an object which can only ever have just one instance is low). So the type inference algorithm stopped at this function. This would then prevent the _entire_ tendency calculation from being inlined, leading to a dramatic loss in performance.

glwagner:
> > The type infererence issue seems to be fixed. For `flow_over_hills.jl` the cost per time-step is approximately the same on my CPU for the "reference case" (no immersed boundary), and either free-slip or no-slip immersed boundary.
> 
> That's awesome! Just out of curiosity (since I also investigated this I couldn't figure it out), how did you fix the type inference issue?
> 
> cc @wenegrat @whitleyv

Here the issue was introduced in an early commit on this PR --- so it may not be exactly the same issue you faced. On the other hand, I think it happened with the same function that you modified (`conditional_flux`) so there may be a lesson you can glean. I also changed the design of the function (previously it took a function argument, `flux`, now it simply takes the scalar value of the flux). That subtle change might have made an important difference, since in the first case a function is being passed into another function (requiring the compiler to specialize on a function argument, sometimes a tricky thing), and another case it was just a floating point number being passed in (relatively vanilla in comparison). Did you test that kind of change in your PR?

francispoulin:
> > @glwagner : is it the separation of vorticity from the topography in the two immersed boundary cases that are the transient _injections_ that you are referring to?
> > If we were to try this with the `ShallowWaterModel`, where instead of having flow over a bumpy bottom, we had flow along a bumpy coastline, we would not get pressure errors since we don't solve for the pressure. How difficult would it be to set this up for that scenario? I'd be happy to run it and see what happens but it would take em a while to set things up.
> > The motvation is to assess how much of this is due to pressure errors versus other stuff.
> 
> I'm looking at the time-series of total momentum --- the line goes down, but also _up_! So momentum is _increasing_ inside the domain, transiently.
> 
> I didn't try to connect this with the vorticity field but maybe you can draw some conclusions!

Does the total energy do the same thing or is it just the momentum?

glwagner:
> Does the total energy do the same thing or is it just the momentum?

I'll add that to `flow_over_hills.jl`!

francispoulin:
> > Does the total energy do the same thing or is it just the momentum?
> 
> I'll add that to `flow_over_hills.jl`!

Thanks.  

Given the fact that the reference case flows over a flat bottom it doesn't suprise me that the mometum decreases monotonically.  

The 2nd and 3rd case are flowing over topography and there should be a periodic effect of the topography on the momentum. No?  Total momentum must vary because of the forcing and I don't know exactly what that should look like.  But if the energy is not monotonic that would be a clear sign that something is going wrong.

glwagner:
A few other things that this PR inspired me to think about:

1. I think it'd be nice if `FieldBoundaryConditions` accepted a positional argument to specify the "default" boundary condition in `Bounded` directions. This would simplify scripts where every boundary is no-slip (for example) since we could write

```julia
no_slip = ValueBoundaryCondition(0)
u_bcs = FieldBoundaryConditions(no_slip)
```

Since the positional argument specifies the default we can also override it:

```julia
u_bcs = FieldBoundaryConditions(no_slip, top=FluxBoundaryCondition(nothing))
```

to have no-slip on all boundaries except the top, where we have free-slip / zero flux.

2. I think we do really need a convenience constructor for drag boundary conditions because a) this is friendly for a very common case and b) the drag boundary condition involves uncertain free parameters (Von Karman, roughness...) which we may want to calibrate via ParameterEstimocean and c) similar to the last point, it's crucial that these parameters are obviated in user scripts for reproducibility; defining an interface for drag will help.

I think some simple struct called `QuadraticDrag`, which `regularize_boundary_condition` knows how to deal with, will suffice for this.

francispoulin:
I like the proposed idea for how to specify `FieldBoundaryConditions`.  That seems very clearn and friendly.

src/BoundaryConditions/continuous_boundary_function.jl
```diff
@@ -77,53 +80,93 @@ function regularize_boundary_condition(bc::BoundaryCondition{C, <:ContinuousBoun
...
+const XBoundaryFunction{LY, LZ, S} = BoundaryCondition{<:Any, <:ContinuousBoundaryFunction{Nothing, LY, LZ, S}} where {LY, LZ, S}
+const YBoundaryFunction{LX, LY, S} = BoundaryCondition{<:Any, <:ContinuousBoundaryFunction{LX, Nothing, LZ, S}} where {LX, LZ, S}
+const ZBoundaryFunction{LX, LY, S} = BoundaryCondition{<:Any, <:ContinuousBoundaryFunction{LX, LY, Nothing, S}} where {LX, LY, S}
 
-    i′ = boundary_index(i)
-
-    return bc.func(ynode(Face(), LY(), LZ(), i′, j, k, grid),
-                   znode(Face(), LY(), LZ(), i′, j, k, grid),
-                   clock.time, args...)
+# Return ContinuousBoundaryFunction on east or west boundaries.
```
tomchor, line 105:
```suggestion
"Return ContinuousBoundaryFunction on east or west boundaries."
```

Maybe turn these comments into docstring? Equality descriptive but more easily accessible...

src/Grids/exterior_node.jl
```diff
@@ -0,0 +1,85 @@
...
+
+#####
+##### Exterior node and peripheral node
+#####
+
+"""
+    exterior_cell(i, j, k, grid)
+
+Return `true` when the tracer cell is outside the boundaries of the domain.
+Return `false` either when the tracer cell at index `i, j, k` is "wet".
```
tomchor, line 23: Is there something missing in this sentence?
Also, can we clarify what outside the boundaries of the domain mean, please? (I understand this as being halo regions or inside immersed solid, but not sure if that's exactly the case)

src/Grids/exterior_node.jl
```diff
@@ -0,0 +1,85 @@
...
+
+@inline exterior_cell(i, j, k, grid::XYZBoundedGrid) = (i < 1) | (i > grid.Nx) |
+                                                       (j < 1) | (j > grid.Ny) |
+                                                       (k < 1) | (k > grid.Nz)
+
+"""
+    exterior_node(LX, LY, LZ, i, j, k, grid)
+
+Return `true` when the location `(LX, LY, LZ)`, interpreted as _either_ a
+cell interface, or the cell itself, is "entirely" exterior to a `Bounded` domain.
```
tomchor, line 43: Same comment here as above. Does this include include immersed solid?

src/ImmersedBoundaries/immersed_boundary_condition.jl
```diff
@@ -0,0 +1,290 @@
...
+    return 2 * (cⁱʲᵏ - cᵇ) / Δ
+end
+
+# Metric and index gymnastics for the 6 facets of the cube
+
+@inline function west_ib_flux(i, j, k, ibg, bc::VBC, (LX, LY, LZ), c, closure::ASD, K, id, clock, fields)
+    Δ = Δx(index_left(i, LX), j, k, ibg, LX, LY, LZ)
+    κ = h_diffusivity(i, j, k, ibg, flip(LX), LY, LZ, closure, K, id, clock)
+    ∇c = left_gradient(i, j, k, ibg, κ, Δ, bc, c, clock, fields)
+    return - κ * ∇c
```
glwagner, line 115: @simone-silvestri take a look at this code and also `right_gradient` and `left_gradient` to see how we satisfy `ValueBoundaryCondition` for immersed boundaries

raphaelouillon:
@glwagner I tested the no-slip boundary condition on the classic case of a lock-release gravity current in a rectilinear channel (see video) and compared the position of the front of the dense current as a function of time for the case of an immersed boundary with the no-slip condition, a real boundary with the no-slip condition, and an immersed boundary with the slip condition. 

https://user-images.githubusercontent.com/31293515/164257859-0deb6cf1-f6b7-414a-a86d-8b47d194b98a.mp4


Xfr

src/Grids/exterior_node.jl
glwagner: The domain is divided into "wet" cells (where we evolve prognostic fields) and "dry" cells. There are dry cells even without the immersed grid.

I think the docstring should be short, but we can include a more involved introduction to the concept of halo regions + immersed cells versus "wet" prognosed cells somewhere and reference it here.

src/Grids/exterior_node.jl
glwagner: This function defines `exterior_node` for "native" or "underlying" grids. Later, in `ImmersedBoundaries`, this function is extended to also be correct when an `ImmersedBoundaryGrid` is wrapped around an underlying grid.

src/BoundaryConditions/continuous_boundary_function.jl
glwagner: I can, however these are not user-facing so I don't think we should add docstrings to the documentation library (since we already have a huge number of functions in there).

src/Grids/exterior_node.jl
glwagner: I wonder if

```
inactive_node
```

and 

```
inactive_cell
```

are better names than "exterior". What do you think @tomchor @simone-silvestri ?

I'll keep "peripheral".

glwagner:
> @glwagner I tested the no-slip boundary condition on the classic case of a lock-release gravity current in a rectilinear channel (see video) and compared the position of the front of the dense current as a function of time for the case of an immersed boundary with the no-slip condition, a real boundary with the no-slip condition, and an immersed boundary with the slip condition. 
> 
> 
> 
> https://user-images.githubusercontent.com/31293515/164257859-0deb6cf1-f6b7-414a-a86d-8b47d194b98a.mp4
> 
> 
> 
> 
> 
> Xfr
> 
> 
> 
> 
> 
> 

@raphaelouillon I think I introduced a bug when I introduced support for GradientBoundaryCondition :-/ I'll let you know when that's fixed.

src/Grids/exterior_node.jl
tomchor: I think inactive cell is more intuitive :+1:

src/BoundaryConditions/continuous_boundary_function.jl
tomchor: What's the problem of adding docstrings there?

glwagner:
@raphaelouillon I think at least that major bug is fixed --- I find some differences with no slip now:

https://user-images.githubusercontent.com/15271942/164346072-fa355a60-f595-4af2-b83e-f7d9885b3f20.mp4

glwagner:
I'm seeing some significant slow down with the boundary condition

```julia
# Monin-Obukhov drag coefficient
z₀ = 1e-4 # Charnock roughness
κ = 0.4 # Von Karman constant
Cᴰ(Δz) = (κ / log(Δz / 2z₀))^2

@inline bottom_drag_u(x, y, t, u, w, Cᴰ) = - Cᴰ * u * sqrt(u^2 + w^2)
@inline bottom_drag_w(x, y, t, u, w, Cᴰ) = - Cᴰ * w * sqrt(u^2 + w^2)
@inline bottom_drag_u(x, y, z, t, u, w, Cᴰ) = - Cᴰ * u * sqrt(u^2 + w^2)
@inline bottom_drag_w(x, y, z, t, u, w, Cᴰ) = - Cᴰ * w * sqrt(u^2 + w^2)

Δz = 1 / Nz
Δx = 2π / Nz
u_drag_bc = FluxBoundaryCondition(bottom_drag_u, field_dependencies=(:u, :w), parameters=Cᴰ(Δz))
w_drag_bc = FluxBoundaryCondition(bottom_drag_w, field_dependencies=(:u, :w), parameters=Cᴰ(Δx))
u_bcs = FieldBoundaryConditions(bottom=u_drag_bc, immersed=u_drag_bc)
w_bcs = FieldBoundaryConditions(immersed=w_drag_bc)
```

This is the basic way to implement a quadratic drag from the interface in this PR.

In this case what happens under the hood is that we create 4 `ContinuousBoundaryFunction`. So there could be a type instability compiling all of those (which have been notoriously fickle to compile in the past).

We clearly need to hard code quadratic drag though, because for stretched grids and partial cells (and other types of immersed boundaries in the future) we have to do precompute the logarithm of the grid metrics (to use in a Monin-Obukhov-type model) in each direction independently, as well as the logarithm of the roughness.

It's too much for this PR though, so I think we should just document how to specify no-slip on immersed boundaries (which appears to be performant), and add a few tests. Then in a future PR we can add a `QuadraticDrag` utility (I have a prototype for this object; others are welcome to collaborate on implementing the necessary functions to support it).

glwagner:
Here's some results for reference (flat bottom no slip), "no slip", "free slip", and "bottom drag" for fractional heights h=0.1 and h=0.2:

# h = 0.1
https://user-images.githubusercontent.com/15271942/164373427-e614311e-7ff0-492c-a0de-6156eadd24ae.mp4

# h = 0.2
https://user-images.githubusercontent.com/15271942/164373538-08dcc7ff-820c-4492-9a75-a4a406d2c416.mp4

Energy always decreases (rapidly). Momentum is more squirrely, but fortunately we do find that specifying drag / no-slip seems to extract more momentum than otherwise (eg than with free slip boundary conditions). The amount of momentum lost with a free-slip boundary condition and hills of size `h=0.2` is roughly equivalent to the flat bottom case with no-slip. There's also transient periods with `h=0.2` where the case with no-slip boundary conditions has more momentum than free-slip.

I'm going to run a few more cases at higher resolution. We should also have a more quantitative validation test.

tomchor:
> We clearly need to hard code quadratic drag though, because for stretched grids and partial cells (and other types of immersed boundaries in the future) we have to do precompute the logarithm of the grid metrics (to use in a Monin-Obukhov-type model) in each direction independently, as well as the logarithm of the roughness.

That's what @whitleyv and I found also on #2275. I agree that may be better left for another PR...

francispoulin:
Movies are looking very good @glwagner ! 

Great to see that the kinetic energy is only decreasing.  

The momentum oscillating seems very reasonable to me.  The topography does act like a force can change the momentum, but I don't have a reference to site.

glwagner:
> > We clearly need to hard code quadratic drag though, because for stretched grids and partial cells (and other types of immersed boundaries in the future) we have to do precompute the logarithm of the grid metrics (to use in a Monin-Obukhov-type model) in each direction independently, as well as the logarithm of the roughness.
> 
> That's what @whitleyv and I found also on #2275. I agree that may be better left for another PR...

Ah, can't you link to the relevant results? It's good to keep track of all this.

glwagner:
Also update on correctness / testing: there are now tests that fluxes are correctly imposed:

https://github.com/CliMA/Oceananigans.jl/blob/22739d6a5d410317b0a6fa5cf538ba43dabe16bf/test/test_boundary_conditions_integration.jl#L226

I don't believe there are any simple tests that `Value` / `Gradient` boundary conditions are imposed correctly --- even without an immersed boundary. Instead, we have relied on validation tests to check that. I think it does make sense though to have some tests in that direction... maybe something like a test that 1D couette flow is maintained (for momentum) and that a simple linear tracer gradient is maintained (for tracers). Those would go here: https://github.com/CliMA/Oceananigans.jl/blob/glw/immersed-bcs/test/test_dynamics.jl

Since we don't have that yet we should probably add in another PR. It'd be nice therefore to have a few validation experiments in hand before merging that indicate no-slip "seems to work".

francispoulin:
Interesting that the `h=0.2` case has kinetic energy increasing for the case of free-slip.  I don't suspect that is physical but should go away with increased resolution.  The movies are very fun to watch!

raphaelouillon:
@glwagner I am running into issues when setting up a 3D simulation. I am not sure where the error is coming from but I suspect it comes from a typo in `/Oceananigans/MuMNB/src/Operators/divergence_operators.jl` line 62, where `Ayᵂ ` is defined but not used. Should it be `Ayˢ`?

raphaelouillon:
@glwagner I tested the fix locally, seems to fix the issue. I was also catching up with a past discussion on turbulence closure (https://github.com/CliMA/Oceananigans.jl/issues/1277).  I tried setting up a simulation with `closure =  (AnisotropicMinimumDissipation(), ScalarDiffusivity(ν=κ,κ=κ))` but am getting errors with `no method matching south_ib_flux`. Could this be specific to immersed boundary or has the turbulence closure definition of the molecular values of kinematic viscosity and diffusivity changed since https://github.com/CliMA/Oceananigans.jl/issues/1277?

glwagner:
> @glwagner I tested the fix locally, seems to fix the issue. I was also catching up with a past discussion on turbulence closure (#1277). I tried setting up a simulation with `closure = (AnisotropicMinimumDissipation(), ScalarDiffusivity(ν=κ,κ=κ))` but am getting errors with `no method matching south_ib_flux`. Could this be specific to immersed boundary or has the turbulence closure definition of the molecular values of kinematic viscosity and diffusivity changed since #1277?

Ah, a closure tuple! We might need to add more code to support that. Can you paste the whole error?

glwagner:
> @glwagner I am running into issues when setting up a 3D simulation. I am not sure where the error is coming from but I suspect it comes from a typo in `/Oceananigans/MuMNB/src/Operators/divergence_operators.jl` line 62, where `Ayᵂ ` is defined but not used. Should it be `Ayˢ`?

Nice catch! Thank you!!! Indeed, all the cases I was running were `Flat` in `y` and so I didn't catch that bug. 

PS you might already know but just in case, here's some tips for github: you can generate a link by browsing the code and clicking on the line number to bring up a prompt:

image

when you paste the link in a PR like this one it will show the line:

https://github.com/CliMA/Oceananigans.jl/blob/22286264ced8707fcdcb3d983b0b53e7f73907d2/src/Operators/divergence_operators.jl#L67

In addition, you can actually go into "Files changed" on this PR and _suggest_ the change directly: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request

glwagner:
@tomchor if you would like to help with docs development do you think you can look over the changes that this PR makes to the boundary conditions section and let me know what you think? Feel free to add more information if you think it would be useful.

glwagner:
> @glwagner I tested the fix locally, seems to fix the issue. I was also catching up with a past discussion on turbulence closure (#1277). I tried setting up a simulation with `closure = (AnisotropicMinimumDissipation(), ScalarDiffusivity(ν=κ,κ=κ))` but am getting errors with `no method matching south_ib_flux`. Could this be specific to immersed boundary or has the turbulence closure definition of the molecular values of kinematic viscosity and diffusivity changed since #1277?

Here's a question. Doing wall-resolved LES is a bit tricky here. In the way the code is written, the diffusivity for wall fluxes with value or gradient boundary condition is derived from the model closure. For closure tuples, the logical thing to do is to add all the diffusivities together to come up with the immersed flux. However, this means that to do wall resolved LES we have to solve a tricky problem which is how to specify that the LES diffusivity has a value of 0 on the boundary.

An alternative approach might attach a diffusivity / viscosity to the immersed boundary condition, rather than extracting it from the closure.

docs/src/model_setup/boundary_conditions.md
```diff
@@ -1,62 +1,173 @@
 # [Boundary conditions](@id model_step_bcs)
 
-A boundary condition is applied to each field, dimension, and endpoint. There are left and right boundary conditions
-for each of the x, y, and z dimensions so each field has 6 boundary conditions. Each of these boundary conditions may
-be specified individually. Each boundary condition can be specified via a constant value, an array, or a function.
-
-The left and right boundary conditions associated with the x-dimension are called west and east, respectively. For the
-y-dimension, left and right are called south and north. For the z-dimension, left and right are called bottom and top.
+Boundary conditions are intimately related to the grid topology, and only
+need to be considered in directions with `Bounded` topology.
```
tomchor, line 4: That's not true if there are immersed solids and IBM is used, no?

docs/src/model_setup/boundary_conditions.md
```diff
@@ -1,62 +1,173 @@
...
-5. [`Open`](@ref)
+There are three primary boundary condition classifications:
 
-Boundary conditions are constructed using the classification as a prefix: `FluxBoundaryCondition`, `ValueBoundaryCondition`, and so on.
+1. [`FluxBoundaryCondition`](@ref) specifies fluxes directly.
 
-## Starter tips
+   For example, sunlight absorbed at the ocean surface imparts a temperature flux that heats near-surface fluid.
+   If there is a known `diffusivity`, you can express `FluxBoundaryCondition(flux)`
+   using `GradientBoundaryCondition(-flux / diffusivity)` (aka "Neumann" boundary condition).
```
tomchor, line 112: I like this explanation using `GradientBoundaryCondition`, but you haven't introduced `GradBoundaryCondition` yet. Maybe you could keep the text but the order could be value, gradient, flux? This also has the advantage of being in increasing order of complexity IMO.

docs/src/model_setup/boundary_conditions.md
glwagner: I was thinking that immersed boundaries could only be used with `Bounded` but you're right that we do have `GridFittedBoundary` for 3D stuff which can be used with `Periodic`. I'll update.

docs/src/model_setup/boundary_conditions.md
```diff
@@ -1,62 +1,173 @@
...
-3. [`Value`](@ref) (Dirchlet)
-4. [`Gradient`](@ref) (Neumann)
-5. [`Open`](@ref)
+There are three primary boundary condition classifications:
 
-Boundary conditions are constructed using the classification as a prefix: `FluxBoundaryCondition`, `ValueBoundaryCondition`, and so on.
+1. [`FluxBoundaryCondition`](@ref) specifies fluxes directly.
 
-## Starter tips
+   For example, sunlight absorbed at the ocean surface imparts a temperature flux that heats near-surface fluid.
```
tomchor, line 110: The approximation of sunlight absorption as a BC is not always valid. Often you do need to consider it as heating the near-surface fluid, as you said. So maybe change the example to outgoing longwave radiation from the ocean surface which cools the boundary?

docs/src/model_setup/boundary_conditions.md
```diff
@@ -1,62 +1,173 @@
...
    * Prescribe a velocity tangent to a boundary as in a driven-cavity flow (for example), where the top boundary is moving.
      Momentum will flux into the domain do the difference between the top boundary velocity and the interior velocity, and the prescribed viscosity.
 
-4. `FluxBoundaryCondition` _directly_ prescribes the flux of a quantity across a boundary rather than calculating it given a viscosity or diffusivity.
-   For example, sunlight absorbed at the ocean surface imparts a temperature flux that heats near-surface fluid.
-   If there is a known `diffusivity`, you can express `FluxBoundaryCondition(flux)` using `GradientBoundaryCondition(-flux / diffusivity)` (aka "Neumann" boundary condition).
-   But when `diffusivity` is not known or is variable (as for large eddy simulation, for example), it's convenient and more straightforward to apply `FluxBoundaryCondition`.
+   _Note_: Do not use `ValueBoundaryCondition` on a wall-normal velocity component.
+   `ImpenetrableBoundaryCondition` is internally enforced and thus only needs to be specified
+   for "additional" fields created outside model constructors.
```
tomchor, line 130: I think statement may be a bit confusing. I myself can't think of a situation where the avg user needs to know/use this. So maybe if there's no such situation we could remove the "additional fields" comments? If there are such situations, maybe mention a quick example? (No code necessary though)

docs/src/model_setup/boundary_conditions.md
```diff
@@ -1,62 +1,173 @@
...
+├── bottom: DefaultBoundaryCondition (ValueBoundaryCondition: 0)
+├── top: FluxBoundaryCondition: Nothing
+└── immersed: DefaultBoundaryCondition (ValueBoundaryCondition: 0)
+```
 
 ## Boundary condition structures
 
 Oceananigans uses a hierarchical structure to express boundary conditions:
 
 1. Each boundary has one [`BoundaryCondition`](@ref)
```
tomchor, line 168: Each boundary of each field?

docs/src/model_setup/boundary_conditions.md
```diff
@@ -322,3 +424,61 @@ julia> model.tracers.c
...
+julia> model = NonhydrostaticModel(; grid, boundary_conditions=(u=velocity_bcs, v=velocity_bcs, w=velocity_bcs));
+
+julia> model.velocities.w.boundary_conditions.immersed
+ImmersedBoundaryCondition:
+├── west: ValueBoundaryCondition: 0
+├── east: ValueBoundaryCondition: 0
+├── south: ValueBoundaryCondition: 0
+├── north: ValueBoundaryCondition: 0
+├── bottom: Nothing
+└── top: Nothing
```
tomchor, line 457: Shouldn't these be also `ValueBoundaryCondition: 0`? Or at least the top BC (since it's a grid-fitted bottom)

docs/src/model_setup/boundary_conditions.md
glwagner: I think we should avoid statements about what models should be used in what scenario and focus exclusively on how the code could be used.

docs/src/model_setup/boundary_conditions.md
glwagner: We've seen this error numerous times; that's why I'm mentioning it.

glwagner: Hm, I wasn't sure what you meant by "this". We often see the error that people attempt to enforce impenetrability with `ValueBoundaryCondition(0)` so I'm trying to be proactive in preventing it (and I see why one might make that mistake).

As for using impenetrable boundary condition, it comes up when creating slip velocity fields that are used with `AdvectiveForcing`.

glwagner: I noticed that we include a more detailed statement about `AdvectiveForcing` just below so I simplified this line.

tomchor: Sorry, just to clarify. This part

```
   _Note_: Do not use `ValueBoundaryCondition` on a wall-normal velocity component.
   `ImpenetrableBoundaryCondition` is internally enforced
```

Is important and I agree that it should be there.

This part:

```
only needs to be specified for "additional" fields created outside model constructors.
``` 
is the one that I think may be confusing without a quick mention of an example. It confused me, and I'm a regular contributor. But maybe I just don't know enough about field BCs...

docs/src/model_setup/boundary_conditions.md
glwagner: I changed it to be more bland and mention just "heating" and "cooling"

docs/src/model_setup/boundary_conditions.md
glwagner: Ok, I moved this text into `GradientBoundaryCondition`.

docs/src/model_setup/boundary_conditions.md
glwagner: This is code produced by the doctest, so it here because that's default behavior!

As for what's going on --- the default boundary condition on velocity components that are normal to cell interfaces is `nothing`. (Here this is a vertical velocity.) As explained above, we cannot use `ValueBoundaryCondition(0)` for wall-normal velocity components.

docs/src/model_setup/boundary_conditions.md
tomchor: > I think we should avoid statements about what models should be used in what scenario / opinions about what models are valid and focus exclusively on how the code can be used.

I agree with this. The issue I raised is because the example you gave might not always be valid depending on general model set-up. I think it's good now

docs/src/model_setup/boundary_conditions.md
tomchor: Ah, I missed the fact that this is a vertical velocity. Good point :+1:

docs/src/model_setup/boundary_conditions.md
```diff
@@ -1,62 +1,177 @@
...
+├── top: FluxBoundaryCondition: Nothing
+└── immersed: DefaultBoundaryCondition (ValueBoundaryCondition: 0)
+```
 
 ## Boundary condition structures
 
 Oceananigans uses a hierarchical structure to express boundary conditions:
 
-1. Each boundary has one [`BoundaryCondition`](@ref)
+1. Each boundary of reach field has one [`BoundaryCondition`](@ref)
```
tomchor, line 172:
```suggestion
1. Each boundary of each field has one [`BoundaryCondition`](@ref)
```

docs/src/model_setup/boundary_conditions.md
glwagner: I think if its confusing you should ignore it (it's only relevant to people building `AdvectiveForcing`). How do we make that clear?

glwagner:
Thanks @tomchor and @raphaelouillon with your help on this PR! I implemented preliminary support for closure tuples, but it still needs to be tested. There's still some work to do to support proper both wall-modeled and wall-resolved LES in complex domains but I think this is a good start.

docs/src/model_setup/boundary_conditions.md
```diff
@@ -1,62 +1,177 @@
...
+       * constant velocity tangential to a boundary as in a driven-cavity flow (for example), 
+         where the top boundary is moving. Momentum will flux into the domain do the difference
+         between the top boundary velocity and the interior velocity, and the prescribed viscosity.
 
-4. `FluxBoundaryCondition` _directly_ prescribes the flux of a quantity across a boundary rather than calculating it given a viscosity or diffusivity.
-   For example, sunlight absorbed at the ocean surface imparts a temperature flux that heats near-surface fluid.
-   If there is a known `diffusivity`, you can express `FluxBoundaryCondition(flux)` using `GradientBoundaryCondition(-flux / diffusivity)` (aka "Neumann" boundary condition).
-   But when `diffusivity` is not known or is variable (as for large eddy simulation, for example), it's convenient and more straightforward to apply `FluxBoundaryCondition`.
+   _Note_: Do not use `ValueBoundaryCondition` on a wall-normal velocity component.
+   `ImpenetrableBoundaryCondition` is internally enforced for fields created inside the model constructor.
```
tomchor, line 132:
```suggestion
   _Note_: Do _not_ use `ValueBoundaryCondition` on a wall-normal velocity component. For that, `ImpenetrableBoundaryCondition` is internally enforced for fields created inside the model constructor. The user only needs to change the boundary condition on a wall-normal velocity component if an inflow is desired (this should be done carefully, in order to not violate mass conservation). 
```
Two notes: 

- After the advective forcing PR we need to change this
- Nowhere do we mention `OpenBoundaryConditoin` here. I think we should, especially because it appears in the output of many code snippets and it's bound to leave some readers curious.

glwagner: Well, `ImpenetrableBoundaryCondition` is an alias for `OpenBoundaryCondition{nothing}`, and we don't support any other kind of open boundary condition. I also think we might need to rethink how this works soon (eg an "open" boundary should be specified via grid topology, and `Bounded` should be synonymous with impenetrable for velocity fields, similar to how `Periodic` topology is synonymous with periodic boundary condition).

So I think we can maybe take your comment to mean that we should update `summary(::OpenBoundaryCondition{nothing})` so that it displays as "ImpenetrableBoundaryCondition"?

tomchor: I didn't know we didn't support `OpenBoundaryCondition`. I modified my suggestion to fit that, so now it only mentions `AdvectiveForcing`. 

May I ask what's missing to support it? I've tried it in a couple of toy models and it seemed to work as intended...

> So I think we can maybe take your comment to mean that we should update summary(::OpenBoundaryCondition{nothing}) so that it displays as "ImpenetrableBoundaryCondition"?

I think that would help, but I don't feel strongly about it. I only mentioned it right now because, as it is, `OpenBoundaryCondition` appears on the docs section about boundary conditions but isn't explained anywhere there.

glwagner: It only works for second order advection. The pressure solver isn't right if the velocity is non-constant in time. If it's constant, it doesn't provide any features additional to `PeriodicBoundaryCondition` with a sponge layer. We're discontinuing this boundary condition imminently; but hopefully we will support proper support for open boundary conditions eventually, for sure.

glwagner: there's a comment a few lines down that says:

> Note however that impenetrability is internally enforced for model velocity components,
so that `ImpenetrableBoundaryCondition` is only used for _additional_ velocity components
that are not evolved by a model, such as a velocity component used for (`AdvectiveForcing`)[@ref].

Ill try to combine your suggestion with this.

tomchor (approved):

PR #2440, opened 2022-04-15
Adds HorizontalDivergenceFormulation for scalar diffusions

glwagner:
This allows us to implement "divergence damping" for large scale ocean models via `ScalarDiffusivity`:

```julia
using Oceananigans.TurbulenceClosures: HorizontalDivergenceFormulation
divergence_damping = ScalarDiffusivity(HorizontalDivergenceFormulation(), ν=1)
```

We probably also want vorticity damping. There also might be other names, for vorticity damping the coefficient is often referred to as the "rotational viscosity".

A few other things:

* Possibly one should throw an error if users specify kappa, since this closure has no tracer diffusion.
* I added an abstract type for the formulations
* Should we use flux=0 fallbacks? We have a substantial amount of code devoted to emitting zero fluxes for many cases (eg zero tracer flux in this case).

### DIFF
--- a/src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
+++ b/src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
@@ -5,33 +5,35 @@ Abstract type for closures with *isotropic* diffusivities.
 """
 abstract type AbstractScalarDiffusivity{TD, F} <: AbstractTurbulenceClosure{TD} end
 
+abstract type AbstractDiffusivityFormulation end
+
 """
     struct ThreeDimensionalFormulation end
 
 Specifies a three-dimensionally-isotropic `ScalarDiffusivity`.
 """
-struct ThreeDimensionalFormulation end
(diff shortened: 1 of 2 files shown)


simone-silvestri (approved):

glwagner:
It's also worth noting that "divergence damping" is common in other GCMs. For example from the CAM-SE 2.0 paper:

image

They have different viscous coefficients for vortical and divergent flow components.

https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1029/2017MS001257

simone-silvestri:
Top: biharmonic diffusion. Bottom: WENO vector invariant with "divergence damping" with nu = 100 m^2 / s

image

francispoulin:
I am confused about the formula.  The Laplacian of the velocity is divided into two parts.  The first is the gradient of the divergence and the second is the curl of the curtl of the velocity.

The first should be zero but is not because of numerical error.  Is this really the term that we are using to diffuse?  Just wanted to confirm that I understand this correctly.

glwagner:
It's only the horizontal divergence!

Issue #2445, opened 2022-04-18
Possible issue with distributed architecture


tomchor:
I've been trying to run [this code](https://github.com/CliMA/Oceananigans.jl/blob/00c98a72943cfaaa3b034770561b7ed6a408de40/benchmark/distributed_nonhydrostatic_model_mpi.jl), and I get an error depending on the number of points and ranks I choose in each direction. For example I noticed that when `Nx*Rx == Ny*Ry == Nz*Rz` the code runs successfully. But if that condition isn't satisfied (for example if I set `Nx=Ny=Nz=8`, `Rx=Rz=1` and `Ry=2`) I get an error like this:

```
ERROR: ERROR: LoadError: LoadError: DimensionMismatch(DimensionMismatch("arrays could not be broadcast to a common size; got a dimension wi"arrays could not be broadcast to a common size; got a dimension with lengths 8 andth lengths 8 and 4") 4")
Stacktrace:
  [1] 
Stacktrace:
  [1] _bcs1_bcs1
    @ ./
    @ ./broadcast.jl:broadcast.jl:501 [inlined]
501 [inlined]
  [2]   [2] _bcs(_bcs(shape::shape::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}, newshape::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}}, newshape::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}})
    @ Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}, Base.OneTo{Int64}})
    @ Base.Broadcast ./broadcast.jl:Base.Broadcast ./broadcast.jl:495
  [3]495
  [3] broadcast_shape
    @  broadcast_shape
    @ ././broadcast.jl:broadcast.jl:489489 [inlined]
 [inlined]
  [4]  [4] combine_axes combine_axes
    @ 
    @ ././broadcast.jl:broadcast.jl:484 [inlined]484 [inlined]

  [5]   [5] _axes_axes
    @ ./
    @ ./broadcast.jl:209broadcast.jl:209 [inlined]
  [6] [inlined]
  [6] axes
    @  axes
    @ ././broadcast.jl:207broadcast.jl:207 [inlined]
 [inlined]
[...]

Issue #2347, opened 2022-03-13
Error from benchmark/test distributed nonhydrostatic model scripts


johnryantaylor:
I am getting the following error from benchmark/distributed_nonhydrostatic_model.jl where I have set threaded = false and decomposition = Slab().  I get a very similar error from the distributed nonhydrostatic script when I run the tests script, but the distributed shallow water model runs fine.  In case it matters, I'm running mpich/3.0.4 (a bit old) on a linux system running Ubuntu.

```
Oceananigans v0.71.5
Julia Version 1.6.5
Commit 9058264a69 (2021-12-19 12:30 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Xeon(R) CPU           E5645  @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, westmere)
[2022/03/13 19:01:45.515] INFO  Benchmarking weak scaling nonhydrostatic model with Slab decomposition [N=(128, 128, 16), ranks=(1, 1, 1)]...
[2022/03/13 19:02:09.310] INFO  Setting up distributed nonhydrostatic model with N=(128, 128, 16) grid points and ranks=(1, 1, 1) on rank 0...
[2022/03/13 19:02:52.561] INFO  Warming up distributed nonhydrostatic model on rank 0...
[2022/03/13 19:04:45.955] INFO  Benchmarking distributed nonhydrostatic model on rank 0...
[2022/03/13 19:04:50.814] INFO  Done benchmarking on rank 0. Median time: 72.806 ms
[2022/03/13 19:04:57.400] INFO  Benchmarking weak scaling nonhydrostatic model with Slab decomposition [N=(128, 128, 32), ranks=(1, 2, 1)]...
[2022/03/13 19:05:21.386] INFO  Setting up distributed nonhydrostatic model with N=(128, 128, 32) grid points and ranks=(1, 2, 1) on rank 1...
[2022/03/13 19:05:21.430] INFO  Setting up distributed nonhydrostatic model with N=(128, 128, 32) grid points and ranks=(1, 2, 1) on rank 0...
[2022/03/13 19:06:04.003] INFO  Warming up distributed nonhydrostatic model on rank 1...
[2022/03/13 19:06:04.004] INFO  Warming up distributed nonhydrostatic model on rank 0...
[...]

PR #2447, opened 2022-04-20
Update dependency packages

navidcy:


glwagner:
Thanks @navidcy !

navidcy:
Any idea why [distributed tests](https://buildkite.com/clima/oceananigans/builds/6805#646493ef-a91c-418c-b51f-4cafeced8ff1) fail?

glwagner:
Possibly related to discussion on #2347 ? Broadcasting ordinary arrays with PencilArrays is fragile. It needs to be fixed...

tomchor:
> Possibly related to discussion on #2347 ? Broadcasting ordinary arrays with PencilArrays is fragile. It needs to be fixed...

I think that's correct. A way to test this is to change the following lines: 
https://github.com/CliMA/Oceananigans.jl/blob/3ba95e07a27d93129a87f266aecc3c11d20aac14/test/test_distributed_models.jl#L471-L472

to 

```julia
 arch = MultiArch(ranks=(1, 4, 1)) 
 grid = RectilinearGrid(arch, topology=topo, size=(8, 2, 8), extent=(1, 2, 3)) 
```

And it should make tests pass. For some reason I found that the condition `Rx*Nx == Ry*Ny == Rz*Nz` needs to be satisfied in order to make tests pass (see https://github.com/CliMA/Oceananigans.jl/issues/2445).

I don't necessarily recommend merging that way, since this doesn't fix the issue, instead it circumvents it, but I'm pretty sure we won't be able to update the packages (including updating to julia 1.7) before fixing this issue.

On the other hand, this test passes right now on main only because of a coincidence, and there's most likely a bug in code (based on answer by @glwagner and @jipolanco), so the test is not doing its intended purpose. That might be an argument for removing that test until we have time to fix the issue, since it's keeping us from updating other packages.

glwagner:
We could change to `@test_broken`.

glwagner:
Another possibility is to write some tests using `HydrostaticFreeSurfaceModel` with `ExplicitFreeSurface`, which is entirely explicit and thus requires only halo filling via MPI (with no PencilArrays dependency)

glwagner (approved):

navidcy:
All tests pass except the cleanup cpu. 
i’ll merge.

navidcy:
We need to make tests shorter… :(

PR #2461, opened 2022-04-22
Update CATKE parameters for new implicit time-stepping

glwagner:
This PR updates CATKE parameters, which was needed after the semi-implicit time-stepping for the TKE equation was introduced in #2426. This PR is all @adelinehillier's work!

The results were generated by training with a 2 minute time-step (previously we had to use time-steps as small as 1 second during training), and are pretty promising:

image

The next step is to calibrate and test CATKE at different vertical resolutions. But this is a good start.


navidcy (approved):

navidcy:
@glwagner, merge?

glwagner:
We haven't updated the parameters yet --- @adelinehillier any news on that front?

navidcy:
oh I see :)

glwagner:
@adelinehillier reports the following optimal values for CATKE to match the LESbrary:

image

Note that these are optimal values for an _older_ version of CATKE that calculated diffusivities at cell centers:

https://github.com/CliMA/Oceananigans.jl/blob/00c98a72943cfaaa3b034770561b7ed6a408de40/src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/CATKEVerticalDiffusivities.jl#L140-L143

Therefore work still remains to recalibrate CATKE for current numerics.

glwagner:
@sandreza do you want to take this on? It could be fun to test CATKE with the updated parameters in the 1 deg simulation too.

sandreza:
I'll try it out in the 1 degree. I will provide no guarantees that the parameters will be accurate since they'll be determined by eyeball

glwagner:
👁️

glwagner:
@simone-silvestri have updated the parameters. However, the scale of the plots doesn't allow us to accurately infer some very small, but very-importantly-non-zero parameter values (like `Cᵇc`, `Cᴷcʳ`, and `Cᵂu★`).

PR #2464, opened 2022-04-23
A bit more details on the GPU section

tomchor:
This PR adds a bit more detail to the GPU section of the docs, mentioning a few places where a user can find info about distributed models and contributing to this effort (to maybe get more hands on deck?) and adds a line about the GPU hours that come NSF grants (which apparently not many people are aware of).

### DIFF
--- a/docs/src/using_gpus.md
+++ b/docs/src/using_gpus.md
@@ -7,7 +7,11 @@ See the [performance benchmarks](@ref performance_benchmarks) for more details.
 
 See [Architecture](@ref) for instructions on setting up a model on a GPU.
 
-Oceananigans does not yet support distributed parallelism (multi-CPU or multi-GPU).
+Distributed parallelism (multi-CPU or multi-GPU) is currently work in progress, so even
+though some distributed simulations are current possible, expect a few bugs along the way.
+If you'd like to contibute or read more about it, please check [this PR](https://github.com/CliMA/Oceananigans.jl/pull/2253
+for multi-GPU), [this issue for multi-CPU](https://github.com/CliMA/Oceananigans.jl/issues/2347),
+and [this discussion](https://github.com/CliMA/Oceananigans.jl/discussions/2345).
 
 !!! tip "Running on GPUs"
     If you are having issues with running Oceananigans on a GPU or setting things up,
@@ -34,7 +38,8 @@ performance although Oceananigans can be used with 32-bit floats. Most recent ga
 should work but might have poor 64-bit float performance.
 
 If you have access to any supercomputer clusters, check to see if they have any GPUs.
-See also this Stack Overflow post:
+NSF grants automatically give access to [NCAR's Casper Cluster](https://arc.ucar.edu/knowledge_base/70549550),
+which has several high-end GPUs. See also this Stack Overflow post:
 [Where can I get access to GPU cluster for educational purpose?](https://scicomp.stackexchange.com/questions/8508/where-can-i-get-access-to-gpu-cluster-for-educational-purpose)
 
 Cloud computing providers such as Google Cloud and Amazon EC2 allow you to rent GPUs per


tomchor:
I'm not sure if the PRs/discussions/issues I linked to are the best choice, but I had to start somewhere. Suggestions are welcome :)

glwagner:
If we all agree, maybe we should add some of that info that I just mentioned to the documentation (ie a statement of purpose for the documentation, and the places to look for other kinds of information). How about that?

tomchor:
> If we all agree, maybe we should add some of that info that I just mentioned to the documentation (ie a statement of purpose for the documentation, and the places to look for other kinds of information). How about that?

I was about to write that :)

tomchor:
> > If we all agree, maybe we should add some of that info that I just mentioned to the documentation (ie a statement of purpose for the documentation, and the places to look for other kinds of information). How about that?
> 
> I was about to write that :)

But mostly, to elaborate, I think the docs can be "shaped" and curated in many ways and still be super useful. What you're proposing is definitely one good way to do it, and I agree especially with your third point. But I think no matter what we decide, we should make that clear somewhere that's relatively easy to find.

Following your proposal, we exclude the references to the PRs and issues and only link the discussion in this PR. Also we should probably create a discussion for multi-GPUs and link that as well since the distributed discussion that exists rn only focuses on MPI.

glwagner:
> > > If we all agree, maybe we should add some of that info that I just mentioned to the documentation (ie a statement of purpose for the documentation, and the places to look for other kinds of information). How about that?
> > 
> > 
> > I was about to write that :)
> 
> But mostly, to elaborate, I think the docs can be "shaped" and curated in many ways and still be super useful. What you're proposing is definitely one good way to do it, and I agree especially with your third point. But I think no matter what we decide, we should make that clear somewhere that's relatively easy to find.
> 
> Following your proposal, we should exclude the references to the PRs and issues and only keep the link to the discussion in this PR. Also we should probably create a discussion for multi-GPUs and link that as well since the distributed discussion that exists rn only focuses on MPI.
> 
> How does that sounds @glwagner ?

That sounds good, except that there's nothing to discuss yet about multi-GPU so we probably don't need to create a discussion about it.

I'd also propose not linking to any specific discussions in the docs. The reason is that the discussions are organic and different threads may rise or fall in importance over time. We don't want to have to update the docs when that happens.

glwagner:
I think you have a different vision of what the "docs" should be perhaps. Is there maybe another solution that we can entertain? It's just important to understand the ultimate form that this object I am calling the "documentation" might have to take. For example check out the documentation for ECMWF's Integrated Forecast System:

https://www.ecmwf.int/en/publications/ifs-documentation

https://zenodo.org/record/6334656#.YmNicS-B2gQ

We probably will want the docs to be citable, and may also want to submit parts of them as a paper. I'm trying to reduce the amount of work we'll have to do when the time comes around to make that transformation.

But I also think the information you want to add to the "knowledge ecosystem" is very useful. We just have to find the right place to put it.

francispoulin:
Very good discussion here!

I agree that the docs should be dedicated for the stable features, and the discussions can be for the bleeding edge.

As an aside, we are still calling the `HydrostaticModel` experimental but it seems pretty well developed.  Maybe even more so than the `NonhydrostaticModel` now, and it has a lot more features.  Do we want to continue calling it experimental or do we want to call some aspects of the model experimental?  

This question should probably go elsewhere but it might be a one word answer, no, so I thought I would mention it here.

glwagner:
I think for idealized stuff the hydrostatic model is fairly well tested, but if one wants to do complex domains, problems requiring mixing parameterizations, etc, then our capability is still a little green I think.

francispoulin:
I agree @glwagner .  Maybe we can say that with non-uniform coastlines and/or topography, the model is experimental?  Rather awkard to say I know, but it seems true.

PR #2474, opened 2022-04-25
Update docs to point to discussions, wiki, etc

glwagner:
This PR updates the docs to provide stronger links to the [new wiki](https://github.com/CliMA/Oceananigans.jl/wiki).

There's a few more updates I'd like to make, and perhaps we can completely migrate "using GPUs" and "installation instructions" over to the new wiki? @tomchor and @navidcy are the docs aficionados so let's see what they think.

~I'd also like to resolve #2471 .~

Closes #2471

PR #2484, opened 2022-04-27
Bump to 0.75.3

glwagner:
Captures immersed boundary condition API #2437 docs update #2474

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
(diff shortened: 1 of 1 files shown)


tomchor:
@glwagner I think the bump here isn't necessary. You bumped the version with the IBM Boundary condition PR, now we just have to tag that version, no?

glwagner:
oh sorry you mean we just didn't register it?

glwagner:
Also theres a warning so I'm bumping again with #2482

tomchor:
> oh sorry you mean we just didn't register it?

Yes, sorry if that wasn't clear!

I thought registering and tagging could be used interchangeably since the registrator automatically registers _and_ tags a new version, no? I'll try to be more precise with the language in the future

PR #2481, opened 2022-04-27
Implement the actually-Cox (aka the anti-anti-Cox) stencil for GM

glwagner:
This PR refactors the calculation of `isopycnal_rotation_components` to _actually_ use the "Cox stencil" described by [Griffies et al 1998](https://journals.ametsoc.org/view/journals/phoc/28/5/1520-0485_1998_028_0805_idiazc_2.0.co_2.xml).

~~It looks to me like MITgcm uses this so-called anti-Cox stencil.~~ It looks like MITgcm uses the Cox stencil. For example, when calculating the "Gradient of Sigma at rVel points" --- which appears in `isopycnal_rotation_tensor_xz_ccf` --- MITgcm uses

https://github.com/MITgcm/MITgcm/blob/9bcb4977016b9cc234328aa357b105d227417e43/pkg/gmredi/gmredi_calc_tensor.F#L712

which reconstructs the x-component of the density gradient by _averaging the gradient_ (with a four point average). On the other hand, we use

https://github.com/CliMA/Oceananigans.jl/blob/001ec56debf0447103d9b75c81de4036a8d6a94a/src/TurbulenceClosures/isopycnal_rotation_tensor_components.jl#L68

which instead computes the _gradient of the average_ (first by averaging with 4 points, then calculating the gradient of that).

~~But Griffies 1998 describers the Cox87 discretization as~~ EDIT: we were confused by Griffies' notation.

NOW, this PR changes all the buoyancy gradient calculations so they are similar to MITgcm (average of the gradient) --- the "actually-Cox stencil". We'll see what that does.

PR #2487, opened 2022-04-28
VerticallyImplicitTimeDiscretization for symmetric isoneutral diffusion

glwagner:
This PR adds the capability for treating the "33" component of the symmetric isoneutral diffusive flux implicitly.

cc @sandreza would be great if you test this.

This depends on #2481


glwagner:
Not working yet --- we have to be able to calculate the isopycnal slope inside the vertically-implicit diffusion solver, so need some refactoring.

glwagner:
Ok, I think this is basically working now.

https://user-images.githubusercontent.com/15271942/165753191-b9ef8480-72a8-49ef-809e-56fa13d853a3.mp4

For GM, the remaining issue is what happens at horizontal boundaries.

@sandreza let me know if you get satisfactory results from this PR and we will merge.

navidcy (approved):

navidcy:
`VerticallyImplicitTimeDiscretization` will be the default, right?

glwagner:
I made it default!

PR #2488, opened 2022-04-28
Properly pin MPICH_jll

navidcy:
According to https://pkgdocs.julialang.org/v1/compatibility/, `MPICH_jll = "4.0.1"` means `[4.0.1 - 5.0.0)` and it's not what we wanted!

(And indeed, when I called `using Pkg; Pkg.update()` MPICH_jll got updated to 4.0.2.)


glwagner:
Thanks @navidcy

navidcy:
approve?

glwagner (approved):

glwagner:
> approve?

was on my phone 🥺

glwagner:
I'll merge when tests pass!

navidcy:
cool, do that... and register a new version!

navidcy:
I [paused](https://github.com/JuliaRegistries/General/pull/59278) the release of 0.75.3 when I realised the pinning was not done correctly.

glwagner:
> I [paused](https://github.com/JuliaRegistries/General/pull/59278) the release of 0.75.3 when I realised the pinning was not done correctly.

wow, I didn't know you could do that

vchuravy:
I don't think we should pin a indirect dependency. Using Julia 1.7 with Mac M1 is decidedly unsupported and the user can do the pinning in their `Project.toml`

glwagner:
> I don't think we should pin a indirect dependency. Using Julia 1.7 with Mac M1 is decidedly unsupported and the user can do the pinning in their `Project.toml`

Ah, I didn't realize this was specifically a Mac M1 issue.

I think we'll continue to get issues and questions about this error if we don't do something about it (and its very inconvenient for this important subset of users). What's the best way to solve this problem?

navidcy:
There is a Mac M1 version of Julia here so people will download that and try to use Oceananigans with it...

glwagner:
Maybe @vchuravy's point is that it's labeled "experimental" so problems are expected. So we either live with inconvenience (experimental=inconvenience expected) or use temporary dirty solutions like this.

I guess from "our perspective", 1.7 "basically works" except for this. So while it's experimental on M series in general, it seems we "want" to label it as supported via this PR. Maybe we shouldn't.

vchuravy:
Any code using threading is eventually just going to hang on M1. My point is that Oceanigans should not be working around this for the user since it declares an unecessary dependency for everyone else. Users can pin mpich_jll locally if needed.

navidcy:
> Any code using threading is eventually just going to hang on M1.  

Interesting. I didn't know that!

vchuravy:
In particular https://github.com/JuliaLang/julia/issues/41820

navidcy:
Closing this in favor of https://github.com/CliMA/Oceananigans.jl/pull/2496

PR #2463, opened 2022-04-22
Near global one degree setup

glwagner:
This PR adds a near-global one degree setup to `validation/near_global_lat_lon`. This is mostly @sandreza and @simone-silvestri 's work. The setup spans -75S to 75N and is implemented on a regular `LatitudeLongitudeGrid` with 1 degree horizontal resolution and 50 vertical levels.

It's WIP right now partly because @sandreza is still sorting out some difficulties with including a `IsopycnalSkewSymmetricDiffusivity` closure to this setup. In addition, I'm going to update the `closure_tuples.jl` implementation because this setup may require a tuple of 5 closures.

As a teaser, the file `validation/near_global_lat_lon/one_degree_setups/plot_bathymetry.jl` produces

image

The bathymetry file is small (not much larger than few dozen long text files), so I've uploaded it directly into the repo from this PR. If we don't want to do this, I can reopen this PR with the same `.jl` files and download the bathymetry from `OceananigansArtifacts.jl` instead.

I'm thinking that we should add at least two one degree setups in this PR:

1. A "realistic" setup with initial conditions and surface relaxation or fluxes loaded from file (which @sandreza is working on)

2. An "idealized" setup that uses idealized and constant momentum and temperature forcing (perhaps with `constant_salinity=35.0`)

and possibly others.

Based on this experience, we can go in a similar direction with the 1/4 degree setup.

PR #2495, opened 2022-04-28
Biharmonic divergence damping

simone-silvestri:
This PR introduces a "Biharmonic" divergence damping.

In line with #2440, `ScalarBiharmonicDiffusivity(HorizontalDivergenceFormulation(), ν = 1)`, is a biharmonic diffusion that acts only on the divergent component of the horizontal velocity field.

This allows us to diffuse the remaining horizontally rotational component through the use of a WENO scheme (which does not act on vertical motions, hence does not dissipate noise generated through divergent motions)

### DIFF
--- a/src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
+++ b/src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
@@ -31,6 +31,7 @@ const ccc = (Center(), Center(), Center())
 
 const AIBD = AbstractScalarBiharmonicDiffusivity{<:ThreeDimensionalFormulation}
 const AHBD = AbstractScalarBiharmonicDiffusivity{<:HorizontalFormulation}
+const ADBD = AbstractScalarBiharmonicDiffusivity{<:HorizontalDivergenceFormulation}
 const AVBD = AbstractScalarBiharmonicDiffusivity{<:VerticalFormulation}
 
 @inline viscous_flux_ux(i, j, k, grid, clo::AIBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, ∂xᶜᶜᶜ, biharmonic_mask_x, ∇²ᶠᶜᶜ, U.u)
@@ -52,6 +53,9 @@ const AVBD = AbstractScalarBiharmonicDiffusivity{<:VerticalFormulation}
 @inline viscous_flux_vz(i, j, k, grid, clo::AVBD, K, U, C, clk, b) = + ν_σᶜᶠᶠ(i, j, k, grid, clo, K, clk, biharmonic_mask_z, ∂zᶜᶠᶠ, ∂²zᶜᶠᶜ, U.v)
 @inline viscous_flux_wz(i, j, k, grid, clo::AVBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, ∂zᶜᶜᶜ, biharmonic_mask_z, ∂²zᶜᶜᶠ, U.w)
 
+@inline viscous_flux_ux(i, j, k, grid, clo::ADBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, δ★ᶜᶜᶜ, U.u, U.v)   
+@inline viscous_flux_vy(i, j, k, grid, clo::ADBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, δ★ᶜᶜᶜ, U.u, U.v)
+
 #####
 ##### Diffusive fluxes
 #####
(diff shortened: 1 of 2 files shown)


src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -139,6 +139,23 @@ const C = Center
...
+    viscous_flux_u = Symbol(:viscous_flux_u, dir)
+    viscous_flux_v = Symbol(:viscous_flux_v, dir)
+    viscous_flux_w = Symbol(:viscous_flux_w, dir)
+    @eval begin
+        @inline $diffusive_flux(i, j, k, grid, args...) = zero(grid)
+        @inline $viscous_flux_u(i, j, k, grid, args...) = zero(grid)
+        @inline $viscous_flux_v(i, j, k, grid, args...) = zero(grid)
+        @inline $viscous_flux_w(i, j, k, grid, args...) = zero(grid)
+    end
+end
```
glwagner, line 158: Probably belongs at the top module level in `TurbulenceClosures.jl`

src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
```diff
@@ -52,6 +53,9 @@ const AVBD = AbstractScalarBiharmonicDiffusivity{<:VerticalFormulation}
 @inline viscous_flux_vz(i, j, k, grid, clo::AVBD, K, U, C, clk, b) = + ν_σᶜᶠᶠ(i, j, k, grid, clo, K, clk, biharmonic_mask_z, ∂zᶜᶠᶠ, ∂²zᶜᶠᶜ, U.v)
 @inline viscous_flux_wz(i, j, k, grid, clo::AVBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, ∂zᶜᶜᶜ, biharmonic_mask_z, ∂²zᶜᶜᶠ, U.w)
 
+@inline viscous_flux_ux(i, j, k, grid, clo::ADBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, δ★ᶜᶜᶜ, U.u, U.v)   
+@inline viscous_flux_vy(i, j, k, grid, clo::ADBD, K, U, C, clk, b) = + ν_σᶜᶜᶜ(i, j, k, grid, clo, K, clk, δ★ᶜᶜᶜ, U.u, U.v)
```
glwagner, line 57: Is this biharmonic? Don't we need a Laplacian somewhere? (Do we need `biharmonic_mask` too?)

glwagner (approved):

francispoulin:
Silly question but, in what line (or lines) do you compute the gradient of the divergence? 

I looked at the code and it was not obvious to me because I'm not very familiar with these terms.

simone-silvestri:
You're right that it's a little bit misleading. It is not technically a "divergence damping", because the actual operator is
```
∇(ν(∇ ⋅ Δu))
```
So it damps the laplacian of the velocity (here we switched the divergence and the laplacian operators)

francispoulin:
> You're right that it's a little bit misleading. It is not technically a "divergence damping", because the actual operator is
> 
> ```
> ∇(ν(∇ ⋅ Δu))
> ```
> 
> So it damps the laplacian of the velocity (here we switched the divergence and the laplacian operators)

Thanks!  And now I get why this is called biharmonic divergence damping.  

We could also write the term as 

``
∇ ( ν Δ (∇ ⋅ u))
``

to show that it depends explicitly on the divergence of the velocity.  If we ever get a divergence free solver then this would be exactly zero.  But the damping occurs here because of the error of the divergence.  

Not a problem, but I just wanted to make sure I understand this as I didn't know about this before today.

Incidently, we don't have the rotational equivalent that was mentioned before?  I presume for that we would just use the Laplacian, which is simpler.

francispoulin:
Thanks @glwagner . Yes, the fact that this is horizontal, makes a bit difference.

simone-silvestri:
Videos produced with a WENO vector invariant and Horizontal Biharmonic divergence damping for the momentum of the form
```
(1 / (1 / Δx² + 1 / Δy² ))² / 5days
```
(No other form of horizontal dissipation for velocity nor for the tracers)

Quarter degree simulation

https://user-images.githubusercontent.com/33547697/166839872-a07d8325-7844-433d-bedd-354f7b1a9cea.mp4

One degree simulation

https://user-images.githubusercontent.com/33547697/166840229-e6e751af-1087-447a-abb6-2170ed206168.mp4

francispoulin:
Wow! It's amazing how much more mesoscale activity we get in the western boundary current in the 1/4 of a degree compared to the 1 degree. A beautiful simulation!

simone-silvestri:
Just realized that the one degree is initialized with a linear stratification everywhere instead of "realistic" conditions from ECCO! That is why we have no equatorial currents and the results look so different! I ll rerun the one degree with more "realistic" (always in quote 😅 ) initial conditions

glwagner:
> Just realized that the one degree is initialized with a linear stratification everywhere instead of "realistic" conditions from ECCO! That is why we have no equatorial currents and the results look so different! I ll rerun the one degree with more "realistic" (always in quote 😅 ) initial conditions

True! You mean the idealized stratification from #2463 ?

simone-silvestri:
Exactly!

glwagner:
Damn

simone-silvestri:
The one degree simulation initiated with "realistic" initial conditions


https://user-images.githubusercontent.com/33547697/167052211-80f15339-a9b2-4dd4-9b48-9e1376b5c936.mp4

PR #2498, opened 2022-04-29
Remove spurious extra calculate_diffusivities!

glwagner:


glwagner:
The version bump captures vertically-implicit diffusion for GM in #2487

francispoulin (approved):
Nice change

PR #2510, opened 2022-05-03
Remove vertical advection from vector invariant WENO

simone-silvestri:
I found out that treating with WENO the vertical advection in a vector invariant formulation (`w⋅∂z(u)` and `w⋅∂z(u)`) might be detrimental because

- The noise is not propagated by the `u` and `v` fields which are weno-interpolated but by the `w` field which is just centrally-interpolated
- The wall-time increases substantially without a benefit
- In global simulations (both 1 degree and 1/4 degree) the solution seems to be less stable

This PR comments out the vertical WENO advection for the moment returning to a 2nd order version so we can still use a beneficial version of the WENO vector invariant. The vertical advection term requires further study....


glwagner (approved):
gud.

glwagner:
the people want to see the results

PR #2513, opened 2022-05-03
Permute PencilFFT temporary array dimensions

glwagner:
This PR permutes the dimensions of the _temporary array_ that's used by the `DistributedFFTBasedPoissonSolver` (and `PencilFFTs`) compared to Oceananigans data layout to look like (z, x, y). This permits "pencil" decompositions in x, y via PencilFFTs, which cannot partition along the first dimension.

This PR should probably expand the tests to cover these cases, which weren't tested previously because it wasn't supported.

Co-developed with @johnryantaylor.


glwagner:
@simone-silvestri distributed tests passed so this should be good to merge soon

simone-silvestri:
Perfect!

simone-silvestri (approved):

PR #2518, opened 2022-05-05
Drop direct dependency to GPUArrays and BenchmarkTools

navidcy:
Supersedes #2511.


simone-silvestri (approved):

simone-silvestri:
should also take off `BenchmarkTools` that I added by mistake to `deps` in #2253? (closing #2517)

tomchor (approved):

PR #2524, opened 2022-05-05
OutputWriters for MultiRegion

simone-silvestri:
this PR fixes a bug in the `OutputWriters` for which the output fields are not updated when writing

### DIFF
--- a/src/MultiRegion/multi_region_output_writers.jl
+++ b/src/MultiRegion/multi_region_output_writers.jl
@@ -5,27 +5,34 @@ import Oceananigans.OutputWriters:
                         construct_output,
                         serializeproperty!
 
-# This is not working at the moment?
+# This is working just fine at the moment?
 function fetch_output(mrf::MultiRegionField, model)
     field = reconstruct_global_field(mrf)
     compute_at!(field, time(model))
     return parent(field)
-  end
+end
   
 function construct_output(mrf::MultiRegionField, grid, user_indices, with_halos)
-  user_output = reconstruct_global_field(mrf)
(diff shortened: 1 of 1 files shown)


src/MultiRegion/multi_region_output_writers.jl
```diff
@@ -6,6 +6,8 @@ import Oceananigans.OutputWriters:
                         serializeproperty!
 
 # This is working just fine at the moment?
+# But it will be veeeeery slow, as reconstruct_global_field is not 
+# a performant operation
```
glwagner, line 10: Good!

src/MultiRegion/multi_region_output_writers.jl
```diff
@@ -6,6 +6,8 @@ import Oceananigans.OutputWriters:
...
 function fetch_output(mrf::MultiRegionField, model)
     field = reconstruct_global_field(mrf)
     compute_at!(field, time(model))
```
glwagner, line 13: Another possibility is to `@apply_regionally` to `fetch_output` right?

Then maybe we also want to extend `convert_output` to handle a `MultiRegionObject` or something (eg just save as a tuple, converting each element).

glwagner:
You can also considering transferring to the CPU first, and then carrying on with GPU calculations while the CPU does `reconstruct_global_field` and then saves to disk.

simone-silvestri:
Let s merge this to have at least an output for multi region (I would need it to debug the multi gpu global simulation).

Then we can design a new method that works in the context of cubed spheres also

glwagner (approved):

Issue #2509, opened 2022-05-03
Interface for allocating the prognostic fields and diagnostic fields at the same time


glwagner:
Efficiently distributing a model (especially across GPUs) requires minimizing the number of messages. `update_state!(model)` looks like:

1. Fill halos for the prognostic variables (requires message passing)
2. Computation of diagnostic variables
3. Fill halos for diagnostic variables (requires message passing)

We can thus minimize message passing by allocating two buffers: one for the prognostic fields, and one for the diagnostic fields.

@simone-silvestri and I propose a new utility `field_tuple` for building tuples of fields to solve this.

francispoulin:
This sounds interesting but I'm not sure if I complete understand.  

Can we think of an example of a prognistic field as velocity: `𝐯`, and a diagnostic field as vorticity: ζ?

If yes, I see that you need to fill halos to evolve the velocity and then you need to fill halos when you compute the vorticity.  Are you simply saying we allocate the space for the two sets of halos at the same time?

glwagner:
Yes, except that we don't allocate memory for vorticity at all; we compute it on the fly.

The prognostic fields are velocity components and tracers. For hydrostatic, w is diagnostic and free surface is prognostic. For nonhydrostatic, w is prognostic.

The diagnostic fields include pressure, w if hydrostatic, and any diagnostics fields associated with the closure.

We will also have a user interface for closure prognoetic fields at some point too.

glwagner:
I don't think we're going to implement this

PR #2525, opened 2022-05-05
[WIP] Model allocation interface and model-wise buffer communication

simone-silvestri:
This PR introduces the idea expressed in issue #2509 as well as introducing two unique buffers for _prognostic_ and _diagnostic_ fields which allow the parallelization of memory transfer


glwagner:
Is this stale @simone-silvestri ?

simone-silvestri:
Yep

PR #2477, opened 2022-04-27
[WIP] Implements homogeneous difference and interpolation operators

glwagner:
This major PR changes the way difference operators work so that across boundaries (either the end of the domain in `Bounded` directions, or across immersed boundaries), differences always return 0 and interpolation always returns the value on the "active" or "wet" side of the domain.

This change means the halos do not need to be filled for `Value / Gradient` boundary conditions. Instead, value and gradient boundary conditions will be enforced across `Bounded` the same way they are enforced across immersed boundaries. Moreover these boundary conditions will only be supported for `AbstractScalarDiffusivity`.

This leads to considerable code reduction, because it unifies the algorithms for non-immersed and immersed boundaries. This may decrease compile times. In addition, the reduction in halo filling will hopefully lead to further performance increases, as we have found that halo filling is inefficient especially on the GPU and should be avoided when possible. Finally, avoiding halo filling is important for distributed models, since it eliminates the dependency that halos must be filled across `Bounded` before they are sent between processes across a `Communication` topology. Evaluating performance changes will be an important part of this PR.

There's still a bit todo:

- [x] Refactor halo filling and `apply_flux_x` to enforce value / gradient boundary conditions
- [x] Implement homogeneous interpolation operators
- [ ] Possibly eliminate halo filling for `OpenBoundaryCondition`, leaving halo filling only for periodic and communication bcs

PR #2532, opened 2022-05-07
Improve support for immersed boundaries with `Flat`

glwagner:
Might not be enough...

Should add a test too.

### DIFF
--- a/src/Grids/inactive_node.jl
+++ b/src/Grids/inactive_node.jl
@@ -82,15 +82,19 @@ region of the grid.
 """
 @inline inactive_node(LX, LY, LZ, i, j, k, grid) = inactive_cell(i, j, k, grid)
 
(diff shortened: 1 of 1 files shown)


simone-silvestri:
very fast, good!

glwagner:
Yes for sure. I think we'll also need the 1D versions too --- because in #2477 these functions will be called for all grids, not just immersed.

glwagner:
Oh I realized I screwed it up because `Flat` is not a location, it's a topology. doh! I'll fix it

glwagner:
I can't do it right away so if you're inspired @simone-silvestri tonfix my mess feel free

glwagner:
Ok, looking at the code more I see we don't need to update `inactive_node` or peripheral_node` for `Flat`. The reason is they don't do any array access on their own (except potentially through `immersed_cell`). So the right fix here is to modify `immersed_cell`.

I took the liberties to scope creep and also added unrelated comments and did some clean up in `inactive_node`...

PR #2536, opened 2022-05-08
Support two-dimensional distributed FFT-based Poisson solves

glwagner:
As well as more general decompositions, and greatly expand the docstring.

Much help was provided by @jipolanco on this journey.

Note right now we depend on `PencilFFTs#master` but hopefully that'll change with https://github.com/jipolanco/PencilFFTs.jl/pull/47.

This required a bit of thinking and logic about pencil decompositions. What I learned is hopefully well-summarized in the docstring:

```
"""
    DistributedFFTBasedPoissonSolver(global_grid, local_grid)

Return a FFT-based solver for the Poisson equation,

∇²x = b

for `MultiArch`itectures.

Supported configurations
========================

We support two "modes":

    1. Two-dimensional decompositions in (x, y) for problems with either
       `Nz > Rx` or `Nz > Ry` (therefore, three-dimensional).

    2. One-dimensional decompositions in either x or y for problems that are
       either two-dimensional, or have limited dimensionality in z.

Above, `Nz = size(global_grid, 3)` and `Rx, Ry, Rz = architecture(local_grid).ranks`.

Other configurations that are decomposed in (x, y) but have too few Nz,
or any configuration decomposed in z, are not supported.

Algorithm for two-dimensional decompositions
============================================

When decomposition in two-dimensions for a three-dimensional problem,
there are three forward transforms, three backward transforms,
and 4 transpositions requiring MPI communication. In the following schematic, the first
dimension is always the local dimension. The following algorithm requires that `Nz > Rx`,
where `Nz` is the number of vertical cells and `Rx` is the number of ranks in x.
For `Nz < Rx` but `Nz > Ry`, a similar algorithm applies with x and y swapped:

1. `first(storage)` is initialized with layout (z, x, y).
2. Transform along z.
3  Transpose + communicate to storage[2] in layout (x, z, y),
   which is distributed into `(Rx, Ry)` processes in (z, y).
4. Transform along x.
5  Transpose + communicate to last(storage) in layout (y, x, z),
   which is distributed into `(Rx, Ry)` processes in (x, z).
6. Transform in y.

At this point the three in-place forward transforms are complete, and we
solve the Poisson equation by updating `last(storage)`.
Then the process is reversed to obtain `first(storage)` in physical
space and with the layout (z, x, y).

Restrictions
============

The algorithm for two-dimensional decompositions requires that `Nz = size(global_grid, 3)` is larger
than either `Rx = ranks[1]` or `Ry = ranks[2]`, where `ranks` are configured when building `MultiArch`.
If `Nz` does not satisfy this condition, we can only support a one-dimensional decomposition.

Algorithm for one-dimensional decompositions
============================================

This algorithm requires a one-dimensional decomposition with _either_ `Rx = 1`
_or_ `Ry = 1`, and is important to support two-dimensional transforms.

For one-dimensional decompositions, we place the decomposed direction _last_.
If the number of ranks is `Rh = max(Rx, Ry)`, this algorithm requires that
_both_ `Nx > Rh` _and_ `Ny > Rh`. It remains somewhat of a mystery why this
succeeds (ie, why the last transform is correctly decomposed).
"""
```


glwagner:
This PR also provides _experimental_ (untested) supported for Bounded domains: @johnryantaylor @raphaelouillon

Future work / PRs can add tests for these cases, and perhaps some validation experiments.

I also think we've unlocked the requirements for FFT + tridiagonal solve, but that is yet again for a future PR.

glwagner:
@simone-silvestri some of these lessons could carry over to FFTs for `MultiRegion`. We should also investigate whether there is scope for non-MPI transposes via PencilArrays...

src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -1,115 +1,183 @@
...
+
+    1. Two-dimensional decompositions in (x, y) for problems with either
+       `Nz > Rx` or `Nz > Ry` (therefore, three-dimensional).
+
+    2. One-dimensional decompositions in either x or y for problems that are
+       either two-dimensional, or have limited dimensionality in z.
+
+Above, `Nz = size(global_grid, 3)` and `Rx, Ry, Rz = architecture(local_grid).ranks`.
+
+Other configurations that are decomposed in (x, y) but have too few Nz,
```
navidcy, line 48:
```suggestion
Other configurations that are decomposed in ``(x, y)`` but have too few `Nz`,
```

src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -1,115 +1,183 @@
...
+```
+
+for `MultiArch`itectures.
+
+Supported configurations
+========================
+
+We support two "modes":
+
+    1. Two-dimensional decompositions in (x, y) for problems with either
```
navidcy, line 40:
```suggestion
    1. Two-dimensional decompositions in ``(x, y)`` for problems with either
```

src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -1,115 +1,183 @@
...
+       `Nz > Rx` or `Nz > Ry` (therefore, three-dimensional).
+
+    2. One-dimensional decompositions in either x or y for problems that are
```
navidcy, line 43:
```suggestion
    2. One-dimensional decompositions in either ``x`` or ``y`` for problems that are
```

src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -1,115 +1,183 @@
...
+       either two-dimensional, or have limited dimensionality in z.
```
tomchor, line 44: What is meant by limited dimensionality in `z`? Could we clarify this a bit?

src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -1,115 +1,183 @@
...
+
+for `MultiArch`itectures.
+
+Supported configurations
+========================
+
+We support two "modes":
+
+    1. Two-dimensional decompositions in (x, y) for problems with either
+       `Nz > Rx` or `Nz > Ry` (therefore, three-dimensional).
```
tomchor, line 41:
```suggestion
    1. Vertical pencil decompositions: two-dimensional decompositions in (x, y) for three dimensional problems (either `Nz > Rx` or `Nz > Ry`).
```

src/Distributed/distributed_fft_based_poisson_solver.jl
glwagner: It's defined precisely just above. I'll mention that.

src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -1,115 +1,182 @@
...
+
+This algorithm requires a one-dimensional decomposition with _either_ `Rx = 1`
+_or_ `Ry = 1`, and is important to support two-dimensional transforms.
+
+For one-dimensional decompositions, we place the decomposed direction _last_.
+If the number of ranks is `Rh = max(Rx, Ry)`, this algorithm requires that 
+_both_ `Nx > Rh` _and_ `Ny > Rh`. The resulting flow of transposes and transforms
+is similar to the two-dimensional case. It remains somewhat of a mystery why this
+succeeds (ie, why the last transform is correctly decomposed).
+"""
```
tomchor, line 94: Really nice docstring :+1:

src/Models/NonhydrostaticModels/solve_for_pressure.jl
```diff
@@ -3,16 +3,26 @@ using Oceananigans.Solvers: FFTBasedPoissonSolver, FourierTridiagonalPoissonSolv
...
 
+using PencilArrays: Permutation
+
 #####
 ##### Calculate the right-hand-side of the non-hydrostatic pressure Poisson equation.
 #####
 
+const ZXYPermutation = Permutation{(3, 1, 2), 3}
+const ZYXPermutation = Permutation{(3, 2, 1), 3}
+
```
tomchor, line 14: These were also defined in `distributed_fft_based_poisson_solver.jl`. Is this double definition necessary or can we define this in only one place to reduce code?

tomchor (approved):

src/Models/NonhydrostaticModels/solve_for_pressure.jl
glwagner: There's often a tradeoff between code reuse and code locality. Here I chose to "redefine" code so that this file could be understood without having to refer to `distributed_fft_based_poisson_solver.jl`. I reasoned that code reuse in this case held almost no advantage (`ZXYPermutation` is about as simple as it gets so there's little cost to repeating it), but that the gains in code locality were significant.

glwagner: There's no mathematical principle to deduce when code locality trumps reuse, or vice versa. I think it often makes to liberally redefine constants. For example, I could easily dispatch on `Permutation{(3, 1, 2), 3}` rather than `ZXYPermutation`. I just think that `ZXYPermutation` is more readable. On the other hand, if I import `ZXYPermutation` from elsewhere then I risk defeating the original purpose of making the code more readable, since the reader might have to find that other file.

But for anything but the most trivial functions, I think it usually makes sense to reuse rather than repeat.

tomchor: Agreed

PR #2537, opened 2022-05-08
Make JLD2 look before leaping

glwagner:
This PR "enhances" `JLD2OutputWriter` so it first _checks_ whether an iteration number exists in the file it's about to save to. If the iteration number does exist, it emits warning, but does not fail. Previously, it would fail. This helps when restoring a simulation from a checkpoint when output is saved at higher frequency than the simulation is checkpointed.

TODO: probably test makes sense.

### DIFF
--- a/src/OutputWriters/jld2_output_writer.jl
+++ b/src/OutputWriters/jld2_output_writer.jl
@@ -255,7 +255,7 @@ function write_output!(writer::JLD2OutputWriter, model)
 
     # Look before you leap, as they say
     jldopen(path, "r+"; writer.jld2_kw...) do file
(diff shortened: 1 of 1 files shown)


tomchor (approved):

navidcy (commented):
Are there breaking changes? If so, then bump patch release?

glwagner:
You could classify this as a bug fix, so I think patch is appropriate. If we go now it'll capture both this and #2536. Or we can wait for #2538

navidcy:
Wait for both. There is no rush!

glwagner:
#2538 is an opus

PR #1910, opened 2021-07-29
Don't precompute hydrostatic pressure in NonhydrostaticModel

glwagner:
This PR inserts the vertical buoyant acceleration directly into the vertical momentum equation in `NonhydrostaticModel`, rather than integrating to find the hydrostatic pressure and inserting its gradient into the horizontal momentum equations.

I just did the easiest thing right because I'm curious if regression tests pass. If they do, we can refactor `NonhydrostaticModel` to eliminate hydrostatic pressure, speed up the model, and reduce its memory footprint without too much pain (🎉)

If the regression tests don't pass, we will unfortunately have slightly more pain ahead of us in refactoring the regression tests. In addition to the above advantages, we also need to eliminate vertical integrals in `NonhydrostaticModel` to permit 2D distributed memory parallelization. This is because `PencilFFTs` only allows parallelization across dimensions higher than 1 (or in other words, we cannot parallelize in `x`). Thus with a vertical integral, we can only parallelize easily across `y`. If we eliminate the vertical integral, we'll be able to parallelize in `y, z`.

PR #2538, opened 2022-05-08
Distributed mixed FFT / vertical tridiagonal solver

glwagner:
This PR builds off #2536 and implements a distributed Poisson solver that users horizontal FFTs and a vertical tridiagonal solve, with more help from @jipolanco.

When distributed in (x, y), this kind of solver is more expensive than a pure FFT-based solver, because it requires 4 additional transpositions + communication.

For problems that are only distributed in x _or_ y (eg, slab decomposition), we can avoid the additional transpositions. ~~Implementing that optimization is TODO for this PR.~~

Some of the details are discussed on https://github.com/jipolanco/PencilFFTs.jl/issues/44.

Future work, which would require abstracting the implementation of hydrostatic pressure in `NonhydrostaticModel` (and, for friendliness, forbidding the use of VerticallyImplicitTimeDiscretization), could in principle support a more efficient version of this solver with pencil decomposition in (y, z) or (x, z). This memory layout would increase performance for very large problems that require a 2D domain decomposition, since decomposing in (y, z) or (x, z) reduces the number of transposes needed by 4 over (x, y). This feature is easy to code, but might take some time to test. We've already noticed on #1910 that lumping hydrostatic and nonhydrostatic pressure produces different (perhaps lower quality) solutions.

TODO:
- [x] Implement a more efficient algorithm for 1D "slab" decompositions
- [x] Add tests


src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -1,134 +1,339 @@
...
+    @. x̂ = - b̂ / (λ[1] + λ[2] + λ[3])
 
     # Set the zeroth wavenumber and volume mean, which are undetermined
     # in the Poisson equation, to zero.
-    if MPI.Comm_rank(solver.communicator) == 0 #MPI.Comm_size() - 1
-        x̂[1, 1, 1] = 0
+    multi_arch = architecture(solver.local_grid)
+
+    if MPI.Comm_rank(multi_arch.communicator) == 0
+        # This is an assumption: we *hope* PencilArrays allocates in this way
```
jipolanco, line 216: This should usually be true. You can verify it using `range_local`:

```julia
rng = PencilArrays.range_local(x̂)  # returns local range (i1:i2, j1:j2, k1:k2) in the global grid
@assert all(first.(rng) .== 1)  # check that i1 == j1 == k1 == 1
```

glwagner: Nice, thank you!

glwagner: Is there a way to change the condition to always be true? Ie, can we diagnose which rank has (1, 1, 1) (cheaply?)

glwagner:
@jipolanco I just send up another big commit --- I realized that we could use `extra_dims` when we use a 1D process grid + tridiagonal to save much communication:

https://github.com/CliMA/Oceananigans.jl/blob/e1cac85ff8fdd9032549b1a3c32569bc71a92c1e/src/Distributed/distributed_fft_based_poisson_solver.jl#L307-L314

and

https://github.com/CliMA/Oceananigans.jl/blob/e1cac85ff8fdd9032549b1a3c32569bc71a92c1e/src/Distributed/distributed_fft_based_poisson_solver.jl#L162-L183

Let me know what you think.

src/Distributed/distributed_fft_based_poisson_solver.jl
glwagner: I guess you provided precisely what we need for that:

```julia
ii, jj, kk = PencilArrays.range_local(x̂)
if (ii[1], jj[1], kk[1]) === (1, 1, 1)
    # zero it out
end
```

jipolanco:
> @jipolanco I just send up another big commit --- I realized that we could use `extra_dims` when we use a 1D process grid + tridiagonal to save much communication:

I can't look at the details right now, but that sounds like a good option. To be honest, I haven't really used `extra_dims` and I was thinking about actually removing it :smile: But if you find it useful then we'll keep it there. Let me know if you find any issues.

glwagner:
> > @jipolanco I just send up another big commit --- I realized that we could use `extra_dims` when we use a 1D process grid + tridiagonal to save much communication:
> 
> I can't look at the details right now, but that sounds like a good option. To be honest, I haven't really used `extra_dims` and I was thinking about actually removing it 😄 But if you find it useful then we'll keep it there. Let me know if you find any issues.

Ok! It seems convenient for 1D process grids / slab decomposition that use a tridiagonal solve along the third dimension rather than a transform. But we'll see...

glwagner:
@simone-silvestri this may pass

simone-silvestri (approved):

src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -38,32 +38,35 @@ infer_transform(::Flat) = PencilFFTs.Transforms.NoTransform!()
 Return a solver for the Poisson equation,
 
 ```math
-∇²x = b
+∇²φ = b
```
glwagner, line 41: We call it `x` in `solve!` -- should we change it there too? Trying to use common notation all over...

glwagner: Hmm I guess this is ok. We can't read it if we use `x` here because we also use `x` for the coordinate.

src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -38,32 +38,35 @@ infer_transform(::Flat) = PencilFFTs.Transforms.NoTransform!()
...
-We support two "modes":
+    1. Vertical pencil decompositions: two-dimensional decompositions in ``(x, y)``
+    for three dimensional problems that satisfy either `Nz > Rx` or `Nz > Ry`.
 
-    1. Vertical pencil decompositions: two-dimensional decompositions in (x, y)
-       for three dimensional problems that satisfy either `Nz > Rx` or `Nz > Ry`.
+    2. One-dimensional decompositions in either ``x`` or ``y``.
 
-    2. One-dimensional decompositions in either x or y.
+    Above, `Nz = size(global_grid, 3)` and `Rx, Ry, Rz = architecture(local_grid).ranks`.
```
glwagner, line 66: Maybe put this a little higher as in

> We support two modes for `Nz = size(global_grid, 3)` and `Rx, Ry, Rz = architecture(local_grid).ranks`:

tomchor (approved):

glwagner:
There's a less trivial error here: https://buildkite.com/clima/oceananigans/builds/7523#311535ff-f56d-410e-8571-c3b1d9757daf

I'll try to restart the whole build and see what happens.

tomchor:
Just realized the distributed tests have been running for 6 days. I guess it's fair to say there's still something to fix lol

Just killed it to save resources

glwagner:
@tomchor are you able to test locally? I believe these passed locally for me, so the problem might be relatively easy to solve.

tomchor:
> @tomchor are you able to test locally? I believe these passed locally for me, so the problem might be relatively easy to solve.

I've never tested anything in parallel locally, but I can definitely try

tomchor:
@glwagner  I ran the tests and they got stuck in the same where [this test](https://buildkite.com/clima/oceananigans/builds/8208#01821729-56d9-42df-8776-730055a3332e) got stuck. So it appears that there's something to be fixed here...

navidcy:
@glwagner how do you run locally? do you use `mpiexecjl`?

glwagner:
This is stale and has been superceded

PR #2540, opened 2022-05-09
WIP: Work on AMD gpu portability

christophernhill:
PR for discussion of changes to main. There are some things that can go in soon, others may require more work. 

Note 

- Julia AMD GPU stack (AMDGPU.jl, KernelAbstractions.jl) needs 1.7 Julia at least.
- Currently AMDGPU.jl works against ROCM 4.2. More recent ROCM 5 has LLVM that is ahead of Julia 1.7, so that breaks things! 
- people who are working on relevant Julia support include @jpsamaroo @vchuravy @luraess and @matinraayai ....

What fun! 

A buildkite test is here (   https://github.com/CliMA/Oceananigans.jl/blob/e4340a7f2ef10391b46e4b71bf3f74ffbd4bd945/.buildkite/pipeline.yml#L496 ) which gives some clues on what is needed to run something on MI50.


glwagner:
I think some stuff in `broadcast_abstract_field.jl` may need to be extended:

https://github.com/CliMA/Oceananigans.jl/blob/e4340a7f2ef10391b46e4b71bf3f74ffbd4bd945/src/Fields/broadcasting_abstract_fields.jl

specifically

https://github.com/CliMA/Oceananigans.jl/blob/e4340a7f2ef10391b46e4b71bf3f74ffbd4bd945/src/Fields/broadcasting_abstract_fields.jl#L18

which helps us identify when we're broadcasting between arrays and fields so that we hit 

https://github.com/CliMA/Oceananigans.jl/blob/e4340a7f2ef10391b46e4b71bf3f74ffbd4bd945/src/Fields/broadcasting_abstract_fields.jl#L42

https://github.com/CliMA/Oceananigans.jl/blob/e4340a7f2ef10391b46e4b71bf3f74ffbd4bd945/src/Fields/broadcasting_abstract_fields.jl#L46

Also

https://github.com/CliMA/Oceananigans.jl/blob/e4340a7f2ef10391b46e4b71bf3f74ffbd4bd945/src/Fields/broadcasting_abstract_fields.jl#L24-L28

where we take a shortcut for things like `u .= a` where `u isa Field` and a is some kind of array.

PR #2545, opened 2022-05-10
Update `baroclinic_adjustment.jl` in validations exps

francispoulin:
This script is a bit out of date and these few changes make it run.

### DIFF
--- a/validation/mesoscale_turbulence/baroclinic_adjustment.jl
+++ b/validation/mesoscale_turbulence/baroclinic_adjustment.jl
@@ -48,7 +48,7 @@ coriolis = BetaPlane(latitude = -45)
 κz = 𝒜 * κh # [m² s⁻¹] vertical diffusivity
 νz = 𝒜 * νh # [m² s⁻¹] vertical viscosity
 
-diffusive_closure = VerticalScalarDiffusivity(VerticallyImplicitTimeDiscretization, ν = νz, κ = κz)
+diffusive_closure = VerticalScalarDiffusivity(VerticallyImplicitTimeDiscretization(), ν = νz, κ = κz) 
 
 horizontal_closure = HorizontalScalarDiffusivity(ν = νh, κ = κh)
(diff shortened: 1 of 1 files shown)


validation/mesoscale_turbulence/baroclinic_adjustment.jl
```diff
@@ -137,35 +137,33 @@ end
...
+           bottom = (:, :, 1),
+           top = (:, :, grid.Nz))
 
 for side in keys(slicers)
     field_slicer = slicers[side]
 
     simulation.output_writers[side] = JLD2OutputWriter(model, fields(model),
                                                        schedule = TimeInterval(save_fields_interval),
-                                                       field_slicer = field_slicer,
                                                        filename = filename * "_$(side)_slice",
```
simone-silvestri, line 152: you maybe want to add to the output `indices = field_slicer`?

francispoulin: I honestly don't know how this works but that sounds like a good idea.

navidcy:
Do we even need the script since it's now included in the examples that are literated in the docs?

francispoulin:
> Do we even need the script since it's now included in the examples that are literated in the docs?

Good point. I'm also happy to remove it, which is easier.

validation/mesoscale_turbulence/baroclinic_adjustment.jl
simone-silvestri: the `indices` keyword tells you what field "slice" to save, so in case of `indices=(:, :, grid.Nz)` you are saving only the surface

navidcy:
Does it include anything else not included in the docs example? If not let’s remove it!

francispoulin:
I compared them and I don't see anything that is in the validation that we don't have in the example.  I am going to close this PR and create a new one that simply removes that file.

PR #2547, opened 2022-05-11
Synchronous intra-device copy

simone-silvestri:
This fixes a bug in `MultiRegion` for which `copyto!` is by default asynchronous.

This is not a problem until very large meshes (which unfortunately is what we want with multi GPU) 
I still have to test performance... and maybe we also want another solution


simone-silvestri:
This PR also fixes some memory leakage problems with the Implicit solver which would fill up the memory in case of large (1e8) meshes quickly and slow down execution to wait for GC

src/Architectures.jl
```diff
@@ -123,6 +123,22 @@ function unified_array(::GPU, arr::AbstractArray)
     return vec
 end
 
+## Only for contiguous data!! (i.e. the offset is always 1)
```
glwagner, line 126: Do `CuArray` have offsets?

glwagner: It might help to add that info to the comment

Issue #2548, opened 2022-05-12
Should we check the size of grids in stretched directions?


tomchor:
At the moment this is possible:

```julia
julia> using Oceananigans

julia> grid = RectilinearGrid(size=(4, 4, 4),
                              x = 1:1:10, y=1:1:10,
                              z = 1:1:10, 
                              halo=(3,3,3),
                              )
4×4×4 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── Periodic x ∈ [1.0, 5.0) variably spaced with min(Δx)=1.0, max(Δx)=1.0
├── Periodic y ∈ [1.0, 5.0) variably spaced with min(Δy)=1.0, max(Δy)=1.0
└── Bounded  z ∈ [1.0, 5.0] variably spaced with min(Δz)=1.0, max(Δz)=1.0

julia> grid.Nz
4

julia> length(grid.zᵃᵃᶜ)
10
```
Meaning, you can pass a certain size to the constructor, say `Nx`, and also pass an array for that direction (to "stretch it") that has a different size than `Nx`.

I haven't investigated the consequences of this, but this feels like it could cause problems. Should we throw a warning/error in these cases?

francispoulin:
Good observation and I would vote for yes.  I would prefer if we check and if the sizes don't match the user gets an error and is asked to fix it.

simone-silvestri:
Agreed! Happened to me many times and a warning would be helpful!

glwagner:
The length is correct -- 4 points plus 3 halo points on either side.

glwagner:
Just to clarify that the output was expected (z^aac is an OffsetArray that includes halo points). But I agree a warning would be nice if the length of a coordinate's interfaces does not match the requested cells + 1.

tomchor:
> The length is correct -- 4 points plus 3 halo points on either side.

You're right! That was a bad example. But like you mentioned later, an error/warning is desired

PR #2550, opened 2022-05-13
Throws error when size of dimension is mismatched when constructing RectilinearGrid

tomchor:
Closes https://github.com/CliMA/Oceananigans.jl/issues/2548

Before this PR:

```julia
julia> using Oceananigans

julia> grid = RectilinearGrid(size=(4, 4, 4),
                              x = (0,1), y=(0,1),
                              z = LinRange(0,1,20), 
                              halo=(3,3,3),
                              )
4×4×4 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)      regularly spaced with Δx=0.25
├── Periodic y ∈ [0.0, 1.0)      regularly spaced with Δy=0.25
└── Bounded  z ∈ [0.0, 0.210526] variably spaced with min(Δz)=0.0526316, max(Δz)=0.0526316

julia> grid.zᵃᵃᶜ
10-element OffsetArray(::Vector{Float64}, -2:7) with eltype Float64 with indices -2:7:
 -0.13157894736842105
 -0.07894736842105263
 -0.02631578947368421
  0.02631578947368421
  0.07894736842105263
  0.13157894736842105
  0.18421052631578946
  0.23684210526315788
  0.2894736842105263
  0.3421052631578947

julia> collect(grid.xᶜᵃᵃ)
10-element Vector{Float64}:
 -0.625
 -0.375
 -0.125
  0.125
  0.375
  0.625
  0.875
  1.125
  1.375
  1.625
```

After this PR:

```julia
julia> using Oceananigans

julia> grid = RectilinearGrid(size=(4, 4, 4),
                              x = (0,1), y=(0,1),
                              z = LinRange(0,1,20), 
                              halo=(3,3,3),
                              )
ERROR: ArgumentError: `length(z)` must be equal to size passed to `size` argument.
Stacktrace:
 [1] validate_dimension_specification(T::Type, ξ::LinRange{Float64}, dir::Symbol, size_dir::Int64, FT::Type)
   @ Oceananigans.Grids ~/repos/Oceananigans.jl/src/Grids/input_validation.jl:108
 [2] validate_rectilinear_domain(TX::Type, TY::Type, TZ::Type, FT::Type, size::Tuple{Int64, Int64, Int64}, extent::Nothing, x::Tuple{Int64, Int64}, y::Tuple{Int64, Int64}, z::LinRange{Float64})
   @ Oceananigans.Grids ~/repos/Oceananigans.jl/src/Grids/input_validation.jl:100
 [3] validate_rectilinear_grid_args(topology::Tuple{DataType, DataType, DataType}, size::Tuple{Int64, Int64, Int64}, halo::Tuple{Int64, Int64, Int64}, FT::Type, extent::Nothing, x::Tuple{Int64, Int64}, y::Tuple{Int64, Int64}, z::LinRange{Float64})
   @ Oceananigans.Grids ~/repos/Oceananigans.jl/src/Grids/rectilinear_grid.jl:284
 [4] RectilinearGrid(architecture::CPU, FT::Type; size::Tuple{Int64, Int64, Int64}, x::Tuple{Int64, Int64}, y::Tuple{Int64, Int64}, z::LinRange{Float64}, halo::Tuple{Int64, Int64, Int64}, extent::Nothing, topology::Tuple{DataType, DataType, DataType})
   @ Oceananigans.Grids ~/repos/Oceananigans.jl/src/Grids/rectilinear_grid.jl:259
 [5] top-level scope
   @ REPL[2]:1
 [6] top-level scope
   @ ~/.julia/packages/CUDA/Uurn4/src/initialization.jl:52
```

I decided to throw an error instead of a warning because I don't think we want the user be able to proceed in a situation like this.

### DIFF
--- a/src/Distributed/distributed_grids.jl
+++ b/src/Distributed/distributed_grids.jl
@@ -1,6 +1,5 @@
 using MPI
 using Oceananigans.Grids: topology, size, halo_size, architecture, pop_flat_elements
-using Oceananigans.Grids: validate_rectilinear_domain, validate_topology
 using Oceananigans.Grids: validate_rectilinear_grid_args, validate_lat_lon_grid_args
 using Oceananigans.Grids: generate_coordinate, with_precomputed_metrics
 using Oceananigans.Grids: cpu_face_constructor_x, cpu_face_constructor_y, cpu_face_constructor_z
--- a/src/Grids/input_validation.jl
+++ b/src/Grids/input_validation.jl
@@ -65,7 +65,7 @@ end
 
 coordinate_name(i) = i == 1 ? "x" : i == 2 ? "y" : "z"
 
-function validate_dimension_specification(T, ξ, dir, FT)
+function validate_dimension_specification(T, ξ, dir, size_dir, FT)
 
     isnothing(ξ)         && throw(ArgumentError("Must supply extent or $dir keyword when $dir-direction is $T"))
     length(ξ) == 2       || throw(ArgumentError("$dir length($ξ) must be 2."))
@@ -75,7 +75,7 @@ function validate_dimension_specification(T, ξ, dir, FT)
     return FT.(ξ)
 end
 
-function validate_rectilinear_domain(TX, TY, TZ, FT, extent, x, y, z)
+function validate_rectilinear_domain(TX, TY, TZ, FT, size, extent, x, y, z)
 
     # Find domain endpoints or domain extent, depending on user input:
     if !isnothing(extent) # the user has specified an extent!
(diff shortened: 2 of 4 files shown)


simone-silvestri (approved):

glwagner:
I'm not sure I agree with an error --- what do others think? It seems like you could do this intentionally and we wouldn't want to forbid that. But I don't feel strongly.

glwagner:
Another though: I think we should limit errors to when we _know_ something is incorrect. Here we restricting what users can do because of our "opinion" about how users should write code, which I think is a misguided philosophy. A warning is appropriate because this is a mistake in the majority of cases, however.

tomchor:
I couldn't think of any situation where a user would intentionally want to mismatch a coordinate and its size, hence the error. But I also don't feel particularly strong about it, so feel free to change it to a warning.

glwagner:
Well, the usage is limited here because we always "count from 1". For an atmospheric simulation with a sponge layer on top, you might have a stretched coordinate as the domain gets higher and higher. Then you could keep your levels fixed and reduce the height of the domain by changing Nz.

I think that's why it's good to have a guideline like "throw an error when we _know_ things are incorrect". Because if we get into the business of trying to "predict what users want to do", we'll quickly find our imaginations are limited!

glwagner:
There's definitely a trade-off in this case, as in many...

glwagner:
Also, shouldn't we add the same behavior to `LatitudeLongitudeGrid`? What's the reasoning only to add this to `RectilinearGrid`?

PR #2551, opened 2022-05-13
Fixes bug in IBM with Value or Gradient boundary conditions

tomchor:


simone-silvestri:
nice catch!

simone-silvestri (approved):

tomchor:
If everyone's okay with this, I'll register a new version after merging since I'd say this bug is important (if affect every time value or gradient BCs are applied to IBMs).

navidcy:
Perhaps we rename all variables to `closures` when they are tuples of 2 elements or more? Just a suggestion, feel free to ignore.

navidcy (approved):

francispoulin (approved):

PR #2554, opened 2022-05-14
Adds tracer-specific diffusivity subsection to `model_setup`

tomchor:
@zhazorken and I couldn't find information easy on the docs.


navidcy (approved):
nice

glwagner:
Nice!

I would omit the last sentence, better to specifically list closures this works for I think. For example it's not valid for the experimental "divergence damping" closures, and there could be closures added in the future that make this sentence false.

I would also mention that all tracers have to be assigned diffusivities if using this method (at least right now).

tomchor:
> I would also mention that all tracers have to be assigned diffusivities if using this method (at least right now).

Hadn't thought of that. I only included horizontal and vertical since I don't know how "official" other closures are. Feel free to add more closures to the text

PR #2522, opened 2022-05-05
Vector Invariant Formulation of the `ShallowWaterModel`

simone-silvestri:
As for now the `ShallowWaterModel` is implemented in a conservative form. 

This PR attempts to introduce a non-conservative form of the equation with a vector invariant formulation of advection

such that we can treat the advection with high order WENO reconstruction.
We should also think about introducing bathymetric terms (  and  ) and viscous term (#2403)

with @francispoulin

Issue #2454, opened 2022-04-21
Modular and unified user interface for advection schemes


glwagner:
We need to overhaul the way we specify advection schemes, whose current limitations were exposed when "WENO vector invariant" was implemented. I propose that we implement a hierarchical specification system using a new object called `AdvectionScheme` that specifies

1. The "form" of the advection operator for momentum and tracers (either `VectorInvariant` or `FluxForm` for momentum, always `FluxForm` for tracers)
2. For `FluxForm`, the reconstruction scheme for the advected component 
3. For `VectorInvariant`, the stencil for _computing_ vorticity, as well as the method for _reconstructing_ vorticity at the momentum locations
4. For `VectorInvariant` with WENO reconstruction, the type of the smoothness stencil (vorticity or velocity)
5. The method for reconstruction velocities in mass conservation (continuity) as well as tracer advection

We also may want to specify how we reconstruction _advecting_ velocities for flux form momentum advection (right now we use centered advection with an accuracy one order lower than the accuracy of the advected component reconstruction --- ie with `UpwindBiasedFifthOrder()` we reconstruct advecting components with a centered fourth order scheme).

Finally, we need to allow different schemes for every tracer.

This modular design will allow more detailed specification of an advection scheme and also us to remove the option for WENO reconstruction for vector invariant more easily, which @simone-silvestri pointed out is experimental.

We also would like to explore alternative discretization for the continuity equation. For this we need a "mass" reconstruction scheme, and the advecting scheme for tracers needs to reconstruct `advected_velocity` in a way that's consistent with the mass reconstruction.

Here's a sketch of what these new types might look like:

```julia
""" Specifies a reconstruction scheme for fluxes of the form u * c,
where u is the `advecting_velocity` and c is the `advected_quantity` """
struct FluxFormAdvection
    advected_field_reconstruction
    advecting_velocity_reconstruction
end

""" Specifies a reconstruction scheme for the vector invariant formulation of
momentum advection."""
struct VectorInvariantMomentumAdvection
     vorticity_stencil
     vorticity_reconstruction
     # velocity_reconstruction?
end

struct UpwindWENOVorticityReconstruction
    weno
    smoothness
end

struct AdvectionScheme
    mass # TBD... only support CenteredSecondOrder now
    momentum # FluxForm or VectorInvariant
    tracers # NamedTuple of FluxForm
end
```

francispoulin:
[...]

PR #2557, opened 2022-05-16
Maximum-principle-satisfying WENO scheme

simone-silvestri:
Following https://royalsocietypublishing.org/doi/epdf/10.1098/rspa.2011.0153 (section 5)

At the moment this is just a scheme check, so I have implemented it in 1D (and only in the horizontal direction) 

For the moment this script
```
using Oceananigans

arch = CPU()
grid = RectilinearGrid(arch, size = (100, 100, 1), x = (0, 2π), y = (0, 2π), z=(0, 1), topology = (Periodic, Periodic, Bounded))

free_surface = ImplicitFreeSurface()

model = HydrostaticFreeSurfaceModel(; grid, free_surface,
                                    momentum_advection = WENO5(),
                                    tracer_advection = WENO5(bounds = (0, 1)),
                                    buoyancy = nothing,
                                    tracers = :c)

@inline function linear_u_profile(x, y, z) 
    if x < π/2 || x > 3 * π/2
        return 0.0
    else
        if y > 7/4*π || y < π/4
            return 0.0
        elseif y < 5/4 * π && y > 3/4 * π
            return π/2
        elseif y >= π/4 && y <= 3/4 * π
            return y - π/4
        else
            return π/2 - (y - 5/4 * π)
        end
    end
end

@inline function rectangle_c(x, y, z)
    if x > 2/3 * π && x < 4/3 * π
        if x > 2/3 * π && x < 4/3 * π
            return 1.0
        else
            return 0.0
        end
    else
        return 0.0
    end
end

set!(model, u = linear_u_profile, c = rectangle_c)

Δt = grid.Lx / grid.Nx / maximum(abs, model.velocities.u) * 0.2

for step in 1:10000
    time_step!(model, Δt)
    @info "step $step, time $(model.clock.time)"
end
```
returns 
```
julia> model.tracers.c
100×100×1 Field{Center, Center, Center} on RectilinearGrid on CPU
├── grid: 100×100×1 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
└── data: 106×106×7 OffsetArray(::Array{Float64, 3}, -2:103, -2:103, -2:4) with eltype Float64 with indices -2:103×-2:103×-2:4
    └── max=1.03919, min=-0.0214752, mean=0.351983
```
for usual `WENO5` 
and 
```
julia> model.tracers.c
100×100×1 Field{Center, Center, Center} on RectilinearGrid on CPU
├── grid: 100×100×1 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
└── data: 106×106×7 OffsetArray(::Array{Float64, 3}, -2:103, -2:103, -2:4) with eltype Float64 with indices -2:103×-2:103×-2:4
    └── max=0.999906, min=1.09505e-5, mean=0.352073
```
for `WENO5(bounds = (0, 1))`

So it seems like this scheme, even in its 1D formulation, might  help with our tracer bound issues.


simone-silvestri:
This is a heatmap of the tracer after 10000 time steps. The top is `WENO5`, the bottom `WENO5(positivity_preserving = true)`. The out-of-bounds cells are substituted with a NaN

positivity_preserving

simone-silvestri:
Same plot obtained with a second order centered scheme for tracers

positivity_preserving-2

francispoulin:
This is great to see!

I imagine we could use this for the height field in `ShallowWaterModel` to allow for wetting and drying for a single layer.  And also to prevent layer collapse in the case of multiple layers.

simone-silvestri:
yeah, I guess in that case you don't want to set an upper bound (`bounds = (0, Inf)`)

glwagner:
Just a small plea to make this implementation compatible with / moves us towards the refactor envisioned in #2454

src/Advection/positivity_preserving_tracer_advection_operators.jl
```diff
@@ -0,0 +1,89 @@
...
+@inline function div_Uc(i, j, k, grid, advection::BoundPreservingScheme, U, c)
+
+    div_x = bounded_tracer_flux_divergence_x(i, j, k, grid, advection, U.u, c)
+    div_y = bounded_tracer_flux_divergence_y(i, j, k, grid, advection, U.v, c)
+    div_z = bounded_tracer_flux_divergence_z(i, j, k, grid, advection, U.w, c)
+
+    return 1/Vᶜᶜᶜ(i, j, k, grid) * (div_x + div_y + div_z)
+end
+
+@inline function bounded_tracer_flux_divergence_x(i, j, k, grid, advection::BoundPreservingScheme, u, c)
```
francispoulin, line 19: This looks great!  It seems like these three funcionts are almost identical except for u -> v -> w.  Is that true?

If yes, I wonder if there is away to merge these to one, rather than having three functions that do the same thing?

src/Advection/weno_fifth_order.jl
```diff
@@ -26,7 +30,7 @@ Weighted Essentially Non-Oscillatory (WENO) fifth-order advection scheme.
 
 $(TYPEDFIELDS)
 """
-struct WENO5{FT, XT, YT, ZT, XS, YS, ZS, VI, WF} <: AbstractUpwindBiasedAdvectionScheme{2}
+struct WENO5{FT, XT, YT, ZT, XS, YS, ZS, VI, WF, PP} <: AbstractUpwindBiasedAdvectionScheme{2}
```
glwagner, line 33: Is this the design we will want when we do #2454 ? If we can avoid digging ourselves into a deeper hole we should, but maybe it's not possible without a lot of effort...

glwagner (approved):

glwagner:
To support different bounds for different tracers, let's allow the advection scheme to be a `NamedTuple`. Then pass the advection scheme specific to each tracer into to the tracer tendency kernel.

simone-silvestri:
I think it is already like this, we can specify
```
model = HydrostaticFreeSurfaceModel(tracer_advection = (c = WENO5(bounds = (0, 1), T = WENO5()), tracers = (:c, :T))
```
I guess the nonhydrostatic model does not do the same

simone-silvestri:
I guess it is better to change the `ShallowWaterModel` in #2522 as a lot is changing already

src/Advection/weno_fifth_order.jl
simone-silvestri: I am not still completely sure how #2454 will go along, this change only modifies the `div_Uc` so I guess it might be easy to adapt in the future

simone-silvestri:
I like it! I changed the NonhydrostaticModel to be the same

glwagner:
> I like it! I changed the NonhydrostaticModel to be the same.
> 
> A way to facilitate users is that we leave the option to specify a global `advection` (so that we keep the keyword) or two separate `momentum_advection` and `tracer_advection`. I would rather just change it to be consistent with the HydrostaticFreeSurfaceModel directly.
> 
> what do you guys think?

I think we should do that in a different PR because it's a major, major breaking change to the most popular model!

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -50,7 +55,8 @@ end
 """
     NonhydrostaticModel(;     grid,
                               clock = Clock{eltype(grid)}(0, 0, 1),
-                          advection = CenteredSecondOrder(),
+                 momentum_advection = CenteredSecondOrder(),
+                   tracer_advection = CenteredSecondOrder(),
```
glwagner, line 59: I prefer advection, and we should do this in a different PR.

glwagner (changes requested):

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
simone-silvestri: Ok, then we can change it in a different PR (for the moment then separate `bounds` can be used only in the `HydrostaticFreeSurfaceModel`)

src/Advection/positivity_preserving_tracer_advection_operators.jl
```diff
@@ -0,0 +1,95 @@
...
+
+    c₊ᴸ =  _left_biased_interpolate_xᶠᵃᵃ(i+1, j, k, grid, advection, c)
+    c₊ᴿ = _right_biased_interpolate_xᶠᵃᵃ(i+1, j, k, grid, advection, c)
+    c₋ᴸ =  _left_biased_interpolate_xᶠᵃᵃ(i,   j, k, grid, advection, c)
+    c₋ᴿ = _right_biased_interpolate_xᶠᵃᵃ(i,   j, k, grid, advection, c)
+
+    p̃   =  (cᵢⱼ - ω̂₁ * c₋ᴿ - ω̂ₙ * c₊ᴸ) / (1 - 2ω̂₁)
+    M   = max(p̃, c₊ᴸ, c₋ᴿ) 
+    m   = min(p̃, c₊ᴸ, c₋ᴿ) 
+    θ   = min(abs((upper_limit - cᵢⱼ)/(M - cᵢⱼ + ε₂)), abs((lower_limit - cᵢⱼ)/(m - cᵢⱼ + ε₂)), 1.0)
```
glwagner, line 40:
```suggestion
    θ   = min(abs((upper_limit - cᵢⱼ)/(M - cᵢⱼ + ε₂)), abs((lower_limit - cᵢⱼ)/(m - cᵢⱼ + ε₂)), one(grid))
```

src/Advection/positivity_preserving_tracer_advection_operators.jl
```diff
@@ -0,0 +1,95 @@
...
+
+    c₊ᴸ =  _left_biased_interpolate_yᵃᶠᵃ(i, j+1, k, grid, advection, c)
+    c₊ᴿ = _right_biased_interpolate_yᵃᶠᵃ(i, j+1, k, grid, advection, c)
+    c₋ᴸ =  _left_biased_interpolate_yᵃᶠᵃ(i, j,   k, grid, advection, c)
+    c₋ᴿ = _right_biased_interpolate_yᵃᶠᵃ(i, j,   k, grid, advection, c)
+
+    p̃   =  (cᵢⱼ - ω̂₁ * c₋ᴿ - ω̂ₙ * c₊ᴸ) / (1 - 2ω̂₁)
+    M   = max(p̃, c₊ᴸ, c₋ᴿ) 
+    m   = min(p̃, c₊ᴸ, c₋ᴿ) 
+    θ   = min(abs((upper_limit - cᵢⱼ)/(M - cᵢⱼ + ε₂)), abs((lower_limit - cᵢⱼ)/(m - cᵢⱼ + ε₂)), 1.0)
```
glwagner, line 64:
```suggestion
    θ   = min(abs((upper_limit - cᵢⱼ)/(M - cᵢⱼ + ε₂)), abs((lower_limit - cᵢⱼ)/(m - cᵢⱼ + ε₂)), one(grid))
```

src/Advection/positivity_preserving_tracer_advection_operators.jl
```diff
@@ -0,0 +1,95 @@
...
+
+    c₊ᴸ =  _left_biased_interpolate_zᵃᵃᶠ(i, j, k+1, grid, advection, c)
+    c₊ᴿ = _right_biased_interpolate_zᵃᵃᶠ(i, j, k+1, grid, advection, c)
+    c₋ᴸ =  _left_biased_interpolate_zᵃᵃᶠ(i, j, k,   grid, advection, c)
+    c₋ᴿ = _right_biased_interpolate_zᵃᵃᶠ(i, j, k,   grid, advection, c)
+
+    p̃   =  (cᵢⱼ - ω̂₁ * c₋ᴿ - ω̂ₙ * c₊ᴸ) / (1 - 2ω̂₁)
+    M   = max(p̃, c₊ᴸ, c₋ᴿ) 
+    m   = min(p̃, c₊ᴸ, c₋ᴿ) 
+    θ   = min(abs((upper_limit - cᵢⱼ)/(M - cᵢⱼ + ε₂)), abs((lower_limit - cᵢⱼ)/(m - cᵢⱼ + ε₂)), 1.0)
```
glwagner, line 88:
```suggestion
    θ   = min(abs((upper_limit - cᵢⱼ)/(M - cᵢⱼ + ε₂)), abs((lower_limit - cᵢⱼ)/(m - cᵢⱼ + ε₂)), one(grid))
```

src/Advection/weno_fifth_order.jl
glwagner: In #2454 we will disentangle vector invariant from WENO5, so `VI` will go away. It might still make sense to blend `WENO5` with flux-limited WENO5, however. It's just that if we have nth order WENO, we would also have to figure out how to do nth order flux limited WENO if they are entangled.

glwagner (approved):

src/Advection/weno_fifth_order.jl
simone-silvestri: I think the limiter will be independent on the order of the original scheme. You always want to blend your scheme into a first order upwind (when crossing bounds)

glwagner: ok, great!

PR #2561, opened 2022-05-18
Completely trivial update to a comment about continguous CuArrays

glwagner:


src/Architectures.jl
```diff
@@ -123,7 +123,7 @@ function unified_array(::GPU, arr::AbstractArray)
     return vec
 end
 
-## Only for contiguous data!! (i.e. the offset is always 1)
+## Only for contiguous data!! (i.e. only if the offset for pointer(dst::CuArrat, offset::Int) is 1)
```
simone-silvestri, line 126: you want the add the offset as an argument?

glwagner: I can't claim to understand what's going on here (I added the comment as a hint to my future self, when I do have to figure this out) --- but if that makes this function more generally valid I don't see what the harm is. Might help us in the future and seems pretty minor...

navidcy (approved):

navidcy:
Merge?

PR #2568, opened 2022-05-21
Implement lastindex for FieldTimeSeries

glwagner:
This makes it so

```julia
bt = FieldTimeSeries(filename, "b")
b = bt[end]
```

returns the last element in the `FieldTimeSeries`.


glwagner:
There seems to be a pattern with the failing tests 🤔 🤔 🤔 🤔 🤔 @simone-silvestri

navidcy (approved):
nice

PR #2572, opened 2022-05-23
Fix matrix solver test take 2

simone-silvestri:
Just realized I left a bug... the solution was initialized with random numbers and as such was not converging some times. This finally solves the issue of the test not passing (hopefully)


navidcy (approved):

navidcy:
merge?

PR #2576, opened 2022-05-24
Beautify examples a bit more

glwagner:
This PR builds on @navidcy's excellent work to convert all the examples to Makie. I tried to unify the coding style in the examples as best I could. Also, there was a bit too much logging / plotting messages in the examples, which polluted the docs.

TODO:

- [x] We should use `Axis3` in the baroclinic adjustment animation.
- [x] The stretched grid is not displayed in the tilted bottom boundary layer (we could also get rid of this, because we already visualize a similar stretched grid in another example)
- [x] Remove logging noise from shallow water bickley jet, horizontal convection, and tilted bottom boundary layer examples

### DIFF
--- a/examples/tilted_bottom_boundary_layer.jl
+++ b/examples/tilted_bottom_boundary_layer.jl
@@ -167,15 +167,15 @@ simulation.callbacks[:progress] = Callback(progress_message, IterationInterval(1
 #
 # We add outputs to our model using the `NetCDFOutputWriter`,
 
-u, v, w = model.velocities
+u, v′, w = model.velocities
 b = model.tracers.b
-B = model.background_fields.tracers.b
+B∞ = model.background_fields.tracers.b
 
-total_b = b + B
-total_v = v + V∞
-ω_y = ∂z(u) - ∂x(w)
+B = b + B∞
+V = v + V∞
+ωy = ∂z(u) - ∂x(w)
 
-outputs = (; u, total_v, w, total_b, ω_y)
+outputs = (; u, V, w, B, ωy)
 
 simulation.output_writers[:fields] = NetCDFOutputWriter(model, outputs;
                                                         filename = joinpath(@__DIR__, "tilted_bottom_boundary_layer.nc"),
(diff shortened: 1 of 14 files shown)


glwagner:
@francispoulin the shallow water Bickley jet example claims to be non-dimensional,

https://github.com/CliMA/Oceananigans.jl/blob/45931bd231e1ffc0bb2dd8decf2f28e0c8101275/examples/shallow_water_Bickley_jet.jl#L43

yet chooses g = 9.8,

https://github.com/CliMA/Oceananigans.jl/blob/45931bd231e1ffc0bb2dd8decf2f28e0c8101275/examples/shallow_water_Bickley_jet.jl#L48

what's going on?

francispoulin:
> @francispoulin the shallow water Bickley jet example claims to be non-dimensional,
> 
> https://github.com/CliMA/Oceananigans.jl/blob/45931bd231e1ffc0bb2dd8decf2f28e0c8101275/examples/shallow_water_Bickley_jet.jl#L43
> 
> yet chooses g = 9.8,
> 
> https://github.com/CliMA/Oceananigans.jl/blob/45931bd231e1ffc0bb2dd8decf2f28e0c8101275/examples/shallow_water_Bickley_jet.jl#L48
> 
> what's going on?

Sorry about that.  The shallow water model is dimensional.  We pick $g = 9.81$ to model the interaction of water below air, but we could replace $g$ with $g\prime$ that allows for a reduced gravity.  I suggest we say dimensional.

navidcy:
> * [ ]  The stretched grid is not displaced in the tilted bottom boundary layer (we could also get rid of this, because we already visualize a similar stretched grid in another example)

displaced or displayed?

navidcy:
> > > * [ ]    The stretched grid is not displaced in the tilted bottom boundary layer (we could also get rid of this, because we already visualize a similar stretched grid in another example)
> > 
> > 
> > displaced or displayed?
> 
> sorry --- displayed!
> 
> see below --- the plot is built, but not displayed in the docs.
> 
> image

dealt by [f1ccee0](https://github.com/CliMA/Oceananigans.jl/pull/2576/commits/f1ccee097f3502fc567c63b6121c9fed8912729c)

navidcy (approved):

navidcy:
I'm not sure how to use `Axis3`.. Is Axis3 approxiate for the baroclinic adjustment example? Seems like it (from the name).

navidcy:
Actually there is something wrong with the shallow water example. Will look at it later.

https://clima.github.io/OceananigansDocumentation/previews/PR2576/generated/shallow_water_Bickley_jet/

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -65,6 +65,8 @@ lines(grid.Δzᵃᵃᶜ[1:Nz], grid.zᵃᵃᶜ[1:Nz],
 
 scatter!(grid.Δzᵃᵃᶜ[1:Nz], grid.zᵃᵃᶜ[1:Nz])
 
+current_figure() # hide
```
glwagner, line 68: Is this the same as `display(fig)`?

navidcy:
So, shallow water example blows up now. The only change I can see is that g=9.8 -> g=1 which would imply that the CFL restriction is even better, right?

What's the fluid depth that the model actually feels? @francispoulin?

francispoulin:
> So, shallow water example blows up now. The only change I can see is that g=9.8 -> g=1 which would imply that the CFL restriction is even better, right?
> 
> What's the fluid depth that the model actually feels? @francispoulin?

Yes, this is not a CFL condition.  The problem is that with a weaker gravity, the displacements in the free-surface become much bigger.  I suspect that the fluid depth is going to zero and that then causes a numerical instability.

I would suggest making `Lz = 10`, and see if it is stable.

If it is, I would also try increasing dt by a factor of 2.  If it is stable, then it should make the run twice as fast.

navidcy:
ok, the problem is that

```
U = 1 # Maximum jet velocity
 f = coriolis.f
 g = gravitational_acceleration
 Δη = f * U / g  # Maximum free-surface deformation as dictated by geostrophy
```
now implies that `Δη = 1` which is same as `Lz`! That we have a vanishing layer at some points which makes the CFL criterion break down...

navidcy:
oh @francispoulin I now saw your message!

francispoulin:
> oh @francispoulin I now saw your message!

Great minds think alike, and at the same time it seems! ;)

navidcy (approved):

PR #2577, opened 2022-05-25
To enable discrete functions on "ImmersedBoundaryConditions"

simone-silvestri:


glwagner (approved):

glwagner:
image

Depends on who "we" is

PR #2585, opened 2022-05-30
Fix admonition

navidcy:


src/Solvers/fft_based_poisson_solver.jl
```diff
@@ -69,8 +69,8 @@ on a staggered grid and for periodic or Neumann boundary conditions.
 In-place transforms are applied to ``b``, which means ``b`` must have complex-valued
 elements (typically the same type as `solver.storage`).
 
-!!! info "Alternative names for "generalized" Poisson equation
-    Equation ``(∇² + m) ϕ = b`` is sometimes called the "screened Poisson" equation
+!!! info "Alternative names for 'generalized' Poisson equation"
+    Equation ``(∇² + m) ϕ = b`` is sometimes referred to as the "screened Poisson" equation
```
francispoulin, line 73: Isn't it more commonly called a Helmholtz equation?  I have never heard it called the "screened Poisson" equation, so that seems less common to me.

navidcy: Damn. I merged and then saw this. 

I think @glwagner wrote this. From wiki or something.

francispoulin: No problem.  Maybe this is common in other circles?  We can always change later if we are so inclined.

glwagner: They are different -- helmholtz equation is an eigenvalue problem with oscillatory solutions, "screened Poisson" has decaying solutions. The key is the sign of the non-differentiated term in the operator.

glwagner: By "eigenvalue problem" I mean that either (1) `k` is unknown or (2) it has no solutions. Do you agree? We can't handle either case with this solver.

glwagner: But the docstring does mention Helmholtz anyways:

```
!!! info "Alternative names for 'generalized' Poisson equation"
    Equation ``(∇² + m) ϕ = b`` is sometimes referred to as the "screened Poisson" equation
    when ``m < 0``, or the Helmholtz equation when ``m > 0``.
```

What do we need to change?

glwagner: Possibly some confusion was created by looking at the diff (which just displays 2 lines) rather than the whole code. Here's the source code:

https://github.com/CliMA/Oceananigans.jl/blob/d5e898453041b47017bfb0950c74101d8f0b62ce/src/Solvers/fft_based_poisson_solver.jl#L72

francispoulin: Thanks for pointing that out, and sorry that I didn't look at the entire text.  Looks great!

PR #2589, opened 2022-05-31
Fix fatal interaction between FieldTimeSeries construction and input validation

### LINKED ISSUE #2588: Input validation for stretched grid dimensions breaks `FieldTimeSeries`
This input validation:

https://github.com/CliMA/Oceananigans.jl/blob/d14b6d222dc06fb520635902563af98ed2699246/src/Grids/input_validation.jl#L108

breaks the "naive" construction of grid domain that's used in `FieldTimeSeries`:

https://github.com/CliMA/Oceananigans.jl/blob/d14b6d222dc06fb520635902563af98ed2699246/src/OutputReaders/field_time_series.jl#L143

There, an `OffsetArray` with correct indexing (but incorrect size, according to the new restrictions on grid construction) is used to build a grid for `FieldTimeSeries` (when deserialization fails; typically this is because data was created with an old version of Oceananigans or a different Julia version).

glwagner:
Closes #2588

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.76.3"
+version = "0.76.4"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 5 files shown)


glwagner:
I really would rather emit a warning here rather than an error because I don't think an error is appropriate when input is legitimate (which was the case prior to this PR). @tomchor @simone-silvestri is that ok with you two?

tomchor:
> I really would rather emit a warning here rather than an error because I don't think an error is appropriate when input is legitimate (which was the case for what `FieldTimeSeries` did, prior to this PR). @tomchor @simone-silvestri is that ok with you two?

Sure

tomchor (approved):

PR #2596, opened 2022-06-02
Check for CUDA-enabled device on `grid` construction rather than in model constructors

### LINKED ISSUE #2595: Unnecessary GPU check?
Is this redundant? If there is no CUDA-enabled GPU then how could a `grid` on GPU be provided as constructor argument?

https://github.com/CliMA/Oceananigans.jl/blob/d675d8b9f7359219941fa81ce098eb51b7c5ac8f/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl#L125-L127

navidcy:
Closes #2595.

### DIFF
--- a/src/Grids/Grids.jl
+++ b/src/Grids/Grids.jl
@@ -14,6 +14,7 @@ export offset_data, new_data
 export on_architecture
 
 using CUDA
+using CUDA: has_cuda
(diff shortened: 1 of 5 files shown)


francispoulin (approved):
Looks good!

PR #2600, opened 2022-06-06
[WIP] Enhance `ShallowWater` model to allow for multiple fluid layers

### LINKED ISSUE #2507: Multi-layer shallow water model
Hi all,

I'd be very keen to implement a multi-layer version of the shallow water model. For example, here's the 2-layer version:

Screen Shot 2022-05-03 at 10 15 14 am

My ultimate ambition is to solve this set of equations that include a diapycnal term `S_int`:

Screen Shot 2022-05-03 at 10 17 48 am

where `Q_net` is a prescribed heat forcing at the surface. So, from what I understand and from what I discussed with @navidcy, if we have a multi-layer shallow water model like eqs. 56 above then I can implement the set of eqs with the diapycnal terms added forcing to each individual fluid layers. Does this sound right?

Is enhancing the `ShallowWaterModel` to have multiple layers feasible?

cc @francispoulin, @glwagner, @navidcy, @AndyHoggANU, @rmholmes

navidcy:
This PR enhances the `ShallowWaterModel` to allow for multiple fluid layers.

The equations are summarized in [this overleaf doc](https://www.overleaf.com/read/mtyjxnnrjpqv).

Closes #2507.


navidcy:
There is a nonlocal contribution in the vertical. The pressure gradient at fluid layer $j$ depends on all fluid thicknesses of that layer and **all** layers above that. Does this impose any constraint on the technical design (in terms of whether we write 3D kernels or `Nz` 2D kernels, where `Nz` the number of fluid layers...)

francispoulin:
Hmmm, I can appreciate that if we set the topology in the `z` direction to be `Flat`, then we have one layer.  I presume that if we set it to be something else, say `Layered`, then we will have layered? 

I like that idea a lot, however, I think we still need to throw an error if the user sets the topography to be `Bounded` or `Periodic`.  Do you agree @navidcy ?

glwagner:
> Hmmm, I can appreciate that if we set the topology in the `z` direction to be `Flat`, then we have one layer. I presume that if we set it to be something else, say `Layered`, then we will have layered?
> 
> I like that idea a lot, however, I think we still need to throw an error if the user sets the topography to be `Bounded` or `Periodic`. Do you agree @navidcy ?

I suggested that `Flat` means single layer grid, `Bounded` refers to multi layer.

francispoulin:
Bounded is certainly better than Periodic.  Is there an option to use another word that is more descriptive?

navidcy:
Yeap, I agree.

I just did a first commit to open the PR. Will wait for @dhruvbhagtani to add to it ;)

dhruvbhagtani:
Sorry for the slow reply - I think that using `Layered` is better, because to me, `Bounded` is less descriptive. Plus, we might have a different way of dealing with boundaries in the vertical than `Bounded`, so I would be inclined towards starting with `Layered`, so we have flexibility in choosing what to do with boundaries.

@navidcy, unfortunately I will not be able to start working on the PR before late July, as I’m travelling and will not be able to work in the next couple of months. But I’ll aim to get acquainted with `Oceananigans` in this period.

glwagner:
`Flat` and `Bounded` refer to the topology of the grid though -- not the physics. This keyword is used for all grids, for all models. So this keyword is not being used to specify physics, it's being used to build the grid. The proposal is therefore about how `ShallowWaterModel` solves equations on grids with different topologies.

francispoulin:
Right, and we are essentially looking for a keyword that the user can use to specify they want to use layers.  This keyword can be anything we want, right?  

I was advocating for something that seemed intuitive to me but any layered shallow water model is a plus for me.

glwagner:
> Right, and we are essentially looking for a keyword that the user can use to specify they want to use layers. This keyword can be anything we want, right?
> 
> I was advocating for something that seemed intuitive to me but any layered shallow water model is a plus for me.

I'm talking about the relationship between the vertical grid topology and the algorithm used by `ShallowWaterModel` to solve a problem on that grid.

navidcy:
@dhruvbhagtani I messed this up. I'll close it and open a new PR whenever you are ready :)

PR #2603, opened 2022-06-08
High-order advection stencils, overhaul of boundary treatment, and bugfix in advection reconstruction

simone-silvestri:
In this PR:
- Refactoring of advection to have three different reconstruction schemes: `Centered`, `UpwindBiased` and `WENO`
- advection schemes can be constructed with an `order` keyword argument which goes up to 12 for `Centered` and 11 for upwind schemes
- Reimplementation of boundary treatment: reconstruction method `_symmetric_`, `_left_biased_` and `_right_biased` have their own boundary treatment which entails checking the points for each method differently
- advection scheme have a `boundary_scheme` field which will is used to reconstruct values near the boundary. Boundary schemes use the same method as the _parent_ scheme but with a lower order:
     For example, when using a WENO 9th order, closing in near the boundaries the order will be progressively decreased to 7th, 5th, 3rd and finally 1st.

It also solves various bugs associated with advection and immersed boundary.

PR #2616, opened 2022-06-23
Moves masking fields from model update to output writers when using ImmersedBoundary

tomchor:
This PR removes the masking of tracers when using IBM.

At the moment all tracers are masked inside immersed solids. It's my understanding that these masked values aren't being used in any calculations except for the pressure solver.

~The issue is that having active scalars masked can mess up the pressure solver and lead to some numerical artifacts in some cases. I've noticed some small effects like that in my simulation, but @whitleyv had some more serious spurious dynamics pop up as a result in her simulations that only went away by not masking buoyancy. I'll let @whitleyv herself explain/show what was happening.~

~For now @whitleyv and I just commented out the lines doing the masking in the nonhydrostatic model, but if we do decide to move forward with this PR I'll clean the code more properly. I was hoping we could discuss a bit before doing that.~

Updated description:

When I originally opened this PR, the masking was creating an issue with the dynamics. However, this turned out to be a bug and was solved by #2603. Now the current situation is that masking is still done for all tracers and velocities in the main branch, but this has zero effect on the results.

This PR removes masking when updating `NonhydrostaticModel` and instead masks each output field only when writing to disk, saving computational resources (although just a bit).

CC @simone-silvestri @glwagner @wenegrat


src/Models/NonhydrostaticModels/update_nonhydrostatic_model_state.jl
```diff
@@ -14,9 +14,9 @@ Update peripheral aspects of the model (halo regions, diffusivities, hydrostatic
 function update_state!(model::NonhydrostaticModel)
     
     # Mask immersed tracers
-    tracer_masking_events = Tuple(mask_immersed_field!(c) for c in model.tracers)
+    #tracer_masking_events = Tuple(mask_immersed_field!(c) for c in model.tracers)
```
navidcy, line 17:
```suggestion
```

glwagner:
How does tracer masking affect the pressure solver?

glwagner:
If tracer masking changes the pressure solver, it is likely a bug. The pressure solver finds the solution to the Poisson pressure equation, which is a Poisson equation with a RHS proportional to the divergence of the predictor velocity.

Tracer masking should have no effect on dynamics and is purely a convenience for output (so we should probably move the masking step to output writers to avoid the unnecessary computational expense). However, there is a bug with immersed boundaries and high-order advection schemes (solved by #2603) that means the value of tracer fields in the immersed boundaries _does_ (spuriously and incorrectly) influence the outcome of a simulation.

In addition to that, interpolation / reconstruction across immersed boundaries can in principle touch tracer values within immersed boundaries. However, as far as I know we do not ever reconstruct across immersed boundaries in any core routines except the experimental `IsopycnalSkewSymmetricDiffusivity`. #2477 solves this problem by implementing "homogeneous" operators that do not touch tracer values within the immersed boundaries (ie differences return 0, reconstruction returns values from "active" cells only).

This PR could move tracer masking to output writing to save the computation. However, if the tracer masking affects the outcome of a simulation, then there is a deeper issue at hand (perhaps the issue that's solved by #2603?)

Perhaps a more specific question is: why does removing the tracer masking step "improve" simulation outcomes?

whitleyv:
In the past, problems seem to arise when masking buoyancy as a tracer. Here is an example of a slope at rest with a uniform stratification in a 2D domain run o n the main branch. This is using the nonhydrostatic model, but I've plotted the hydrostatic pressure and it's horizontal derivative because it shows how masking buoyancy with a strong stratification results in really strong horizontal pressure gradients. These gradients seem to leak out along the boundary creating the numerical effects you can see in the velocity.  

https://user-images.githubusercontent.com/67593861/175611052-1aa81aa0-0aa1-4e9a-bfbd-85369b07d99a.mp4

When I did this analysis a while ago, this problem seemed to be fixed by unmasking the tracers,  but on the most up to date version of Oceananigans, it doesn't seem to fix the problem. As @glwagner points out, that makes it sound more like a bug. Running it with https://github.com/CliMA/Oceananigans.jl/pull/2603, seems to fix the issue, at least the issue I noticed before. I believe @tomchor has tried both https://github.com/CliMA/Oceananigans.jl/pull/2603 and unmasking and seemed to see a slight improvement with unmasking?

glwagner:
Okay, thank you very much for the clarification about hydrostatic vs. nonhydrostatic pressure! The hydrostatic pressure integral:

https://github.com/CliMA/Oceananigans.jl/blob/e76deefc1fdf115e05856a18977c593d0b5d3b0d/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl#L15

does indeed use tracers; in addition it involves a reconstruction / interpolation that is might be affecting results here. Bug perhaps?

Note, this issue is not fixed by "unmasking" tracers (it's just that the bug / error is different, and perhaps less innocuous). This bug is also not fixed by #2603.

simone-silvestri:
I m wondering if this issue is solved in the new version

tomchor:
> I m wondering if this issue is solved in the new version

It is! Although I'll move the masking from the model update to the output construction since it doesn't change model runs and it saves time.

Tests should be passing, although apparently the GPU runs aren't running currently

tomchor:
@navidcy @simone-silvestri can someone review this PR? I believe it may be ready to merge.

navidcy:
I don't know the details here to provide useful feedback. Sorry... :(

tomchor:
> I don't know the details here to provide useful feedback. Sorry... :(

That's a good point. This PR changed from when it was first introduced and it's much simpler in scope, so I updated the PR description. If that's enough for you to give me feedback, feel free to do so :)

I'm only modifying the `NonhydrostaticModel` for now, as I'm not familiar with the hydrostatic one. (Not sure if eventually the same thing should be that for that...)

glwagner:
We should make the changes needed for both nonhydrostatic and hydrostatic models.

navidcy:
> We should make the changes needed for both nonhydrostatic and hydrostatic models.

For sure.

tomchor:
> We should make the changes needed for both nonhydrostatic and hydrostatic models.

@glwagner the reason why I hadn't removed masking from the hydrostatic model is that I'm really not familiar with it at all. So it's harder for me to figure out where it's okay to remove it, and testing it.

I just did a best guess and removed it from shallow water and nonhydrostatic models in the places I thought appropriate. It would be good to have some feedback from someone more familiar with the code though on that, though.

src/OutputWriters/output_construction.jl
```diff
@@ -16,6 +17,24 @@ end
...
+
+
+function _construct_output(output, grid, kwargs...)
+    if output isa AbstractOperation
+        output = Field(output)
+    end
+
+    if grid isa ImmersedBoundaryGrid
+        mask_field!(output)
+    end
```
navidcy, line 33: @glwagner, would this be better/faster?

```suggestion
    output isa AbstractOperation && output = Field(output)

    grid isa ImmersedBoundaryGrid && mask_field!(output)
```

simone-silvestri:
@tomchor, if you want I can remove it from the hydrostatic model, it should be the same as for the nonhydrostatic

glwagner:
We have to keep masking for velocities in the hydrostatic model.

In the nonhydrostatic model, the equivalent masking procedure acts on the predictor velocity field prior to solving for pressure:

https://github.com/CliMA/Oceananigans.jl/blob/ea6826fd2ffaed3f0df330cc952667ec878deb6a/src/Models/NonhydrostaticModels/pressure_correction.jl#L12-L14

tomchor:
> We have to keep masking for velocities in the hydrostatic model.
> 
> In the nonhydrostatic model, the equivalent masking procedure acts on the predictor velocity field prior to solving for pressure:

Yes, I kept that in the nonhydrostatic model. Although I think I removed it on the hydrostatic one so I'll put it back in.

tomchor:
@simone-silvestri @glwagner I think tests should be passing now. Ready for reviews

tomchor:
Also, I don't understand why we're calling masking events at update for every `NonhydrostaticModel` regardless of it having an `ImmersedGrid` or not:

https://github.com/CliMA/Oceananigans.jl/blob/d4a45adf21918f95ca4d23ec7167e720c44c501b/src/Models/NonhydrostaticModels/update_nonhydrostatic_model_state.jl#L14-L19

The same goes for the hydrostatic model

Why not something like

```julia
if model.grid isa ImmersedBoundaryGrid
    # call masking
end
```
?

glwagner:
The code does just that because `mask_immersed_field(c) = NoneEvent()` if `c.grid isa ImmseredBoundaryGrid` (via multiple dispatch).

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
```diff
@@ -36,8 +36,12 @@ end
...
 function masking_actions!(model)
     η = displacement(model.free_surface)
-    masking_events = Any[mask_immersed_field!(field)
-                         for field in merge(model.auxiliary_fields, prognostic_fields(model)) if field !== η]
+    prog = prognostic_fields(model)
+
+    # selects u, v, if they're prognostic variables
+    prognostic_vels = Tuple( val for (key, val) in zip(keys(prog), prog) if Symbol(key) in (:u, :v) )
+
+    masking_events = Any[mask_immersed_field!(field) for field in prognostic_vels]
```
glwagner, line 44:
```suggestion
    mask_velocities_event = mask_immersed_velocities!(model.velocities, architecture(model), model.grid)
```

some changes needed below. Also, may need to dispatch on `mask_immersed_velocities!(::PrescribedVelocityFields, args...) = tuple(NoneEvent())` perhaps.

src/Models/ShallowWaterModels/update_shallow_water_state.jl
```diff
@@ -10,11 +9,6 @@ Fill halo regions for `model.solution` and `model.tracers`.
 """
 function update_state!(model::ShallowWaterModel)
 
-    # Mask immersed fields
-    masking_events = Tuple(mask_immersed_field!(field) for field in model.solution)
-
-    wait(device(model.architecture), MultiEvent(masking_events))
-
```
glwagner, line 17: Do we need to mask velocities for the shallow water model? I think we might.

glwagner: The above lines don't touch tracers (just transport and height) so it might be fine to keep that.

src/OutputWriters/output_construction.jl
```diff
@@ -16,6 +17,24 @@ end
...
 
+function mask_field!(output)
+    field_masking_event = mask_immersed_field!(output)
+    wait(field_masking_event)
+end
+
+
+function _construct_output(output, grid, kwargs...)
+    if output isa AbstractOperation
+        output = Field(output)
```
glwagner, line 28: This will allocate memory for an abstract operation output _every time_ output is written. We don't want this --- instead, we want to allocate memory when the output writer is constructed.

glwagner (changes requested):

src/OutputWriters/output_construction.jl
glwagner: Sorry, I thought that the usage of `construct_output` was changed, but I see now that it's not so I revoke this comment. But there are still changes needed below.

src/OutputWriters/output_construction.jl
glwagner, line 33: Since `construct_output` is only called when the output is constructed, the masking is only performed for the first iteration, rather than every time we ask for output.

glwagner (commented):
This PR shouldn't touch `construct_output` which is only called when an output writer is initialized. Instead, we should mask fields before they are written to disk. Output is "fetched" (eg copied from the GPU, or, if already on the CPU, then extracted from a view or other reference, after being computed) by `fetch_and_convert_output`. There are a few options for masking. One option is to mask in this function, after computing:

https://github.com/CliMA/Oceananigans.jl/blob/cb8ac3e0bab2c25ab6c09afd6e73356ce86406a3/src/OutputWriters/fetch_output.jl#L12-L15

This is the most conservative approach, but could launch many superfluous kernels when a lot of computations are requested.

There's another issue at play: users may not always want masking. For example, the nonhydrostatic model does not exactly conserve mass. Users may want to output calculations on the boundaries to diagnose the effect of leakage, and too-aggressive masking would be annoying. We may want to add a flag to the output writers like `mask_immersed_output`, or something...

It may be trivial, but masking costs _could_ be reduced by performing them concurrently, and then `wait`ing at the end. This would require reordering loops in the output writer, because currently we compute-then-fetch each output one after the other serially. To make this change we would have to change the algorithm to something like

1. Compute every output
2. Mask every output (concurrently)
3. "fetch" and convert the outputs one after the other (without computing, now)

glwagner:
Note that `fetch_and_convert_output` is called here:

https://github.com/CliMA/Oceananigans.jl/blob/cb8ac3e0bab2c25ab6c09afd6e73356ce86406a3/src/OutputWriters/jld2_output_writer.jl#L259-L260

for JLD2OutputWriter, and here:

https://github.com/CliMA/Oceananigans.jl/blob/cb8ac3e0bab2c25ab6c09afd6e73356ce86406a3/src/OutputWriters/netcdf_output_writer.jl#L420

for the NetCDF output writer.

src/Models/ShallowWaterModels/update_shallow_water_state.jl
tomchor: You're saying I should put those lines back on?

src/OutputWriters/output_construction.jl
tomchor: Sounds good. I'll mark this as resolved

src/Models/ShallowWaterModels/update_shallow_water_state.jl
glwagner: Yes, I think we need to

src/OutputWriters/fetch_output.jl
```diff
@@ -2,15 +2,26 @@ using CUDA
...
+    wait(field_masking_event)
+end
 
 # Needed to support `fetch_output` with `model::Nothing`.
 time(model) = model.clock.time
 time(::Nothing) = nothing
 
-fetch_output(output, model) = output(model)
+function fetch_output(output, model)
+    model.grid isa ImmersedBoundaryGrid && mask_field!(output)
```
glwagner, line 17: It's better to use multiple dispatch:

```suggestion
    field_masking_event = mask_immersed_field!(output)
    wait(field_masking_event)
```

src/OutputWriters/fetch_output.jl
```diff
@@ -2,15 +2,26 @@ using CUDA
...
+    return output(model)
+end
```
glwagner, line 19: I don't think we want to mask output that isn't `Field`, right? What are we trying to support here?

src/OutputWriters/fetch_output.jl
```diff
@@ -2,15 +2,26 @@ using CUDA
...
 
 function fetch_output(field::AbstractField, model)
     compute_at!(field, time(model))
+    model.grid isa ImmersedBoundaryGrid && mask_field!(output)
```
glwagner, line 23: Two changes here:

First, we should rely on the dispatch in `mask_immersed_field!`. Second, we should only wait on the device where the masking is being performed. For example, we don't need the CPU to wait while masking is performed via GPU kernel.

```suggestion
    field_masking_event = mask_immersed_field!(output)
    wait(device(output), field_masking_event)
```

tomchor:
@glwagner I tested this locally, tests are passing, and I believe I addressed all your comments, so this should be ready for re-review. That said, github tells me there's a requested revision by you but I can't figure out what to do about it. Could you please take a look?

@simone-silvestri @navidcy also feel free to review

glwagner:
I think we will need to do some more testing to make sure this won't break our near global hydrostatic setups, which are unfortunately in a tenuous position because they rely on some untested features (and we don't have regression tests for some important cases). @simone-silvestri what do you think? We may want to wait for a few more PRs (perhaps containing some of those tests) to go in first.

tomchor:
> I think we will need to do some more testing to make sure this won't break our near global hydrostatic setups, which are unfortunately in a tenuous position because they rely on some untested features (and we don't have regression tests for some important cases). @simone-silvestri what do you think? We may want to wait for a few more PRs (perhaps containing some of those tests) to go in first.

@glwagner what is the status of the global hydrostatic setup tests? Do you think we've reached a point where we can merge this PR?

glwagner:
I don't think we have regression tests yet but not sure

glwagner:
@tomchor I think we should revisit this PR. Rather than "moving" the masking, I think we should start by adding a feature to both output writers that defines a "mask value" for immersed regions. What do you think about that? I would like to open a new PR for that. And before that, I'm going to open an issue so we can discuss the user interface.

Once we have a nice feature for masking output, we can consider whether or not to also mask (or not) within the time-stepping. These are separate questions, I think.

tomchor:
Sure. I'm okay with that

PR #2627, opened 2022-06-29
Docs library cleanup

navidcy:


src/Models/ShallowWaterModels/shallow_water_advection_operators.jl
```diff
@@ -115,14 +117,16 @@ end
 @inline v(i, j, k, grid, solution) = @inbounds solution.vh[i, j, k] / ℑyᵃᶠᵃ(i, j, k, grid, solution.h)
 
 """
-    c_div_U(i, j, k, grid, advection, U)
+    c_div_U(i, j, k, grid, solution, c, formulation)
```
tomchor, line 120: Do we want to try to fill these signatures automatically? Like with [DocStringExtensions.jl](https://juliadocs.github.io/DocStringExtensions.jl/latest/)?

Relying on humans to manually match the function signatures on the docstrings is definitely error-prone and more upkeep for us...

tomchor (approved):

glwagner (approved):

PR #2629, opened 2022-06-29
Implement device-aware GPU architecture

glwagner:
This PR associates the GPU architecture with a specific device. This helps make the device that a model / grid is located on more explicit, so we don't have to "infer" the device based on where a particular piece of memory is allocated.

cc @simone-silvestri

### DIFF
--- a/src/Architectures.jl
+++ b/src/Architectures.jl
@@ -60,19 +60,25 @@ Abstract supertype for Distributed architectures supported by Oceananigans.
 abstract type AbstractMultiArchitecture <: AbstractArchitecture end
 
 """
-    CPU <: AbstractArchitecture
+    CPU()
 
-Run Oceananigans on one CPU node. Uses multiple threads if the environment
-variable `JULIA_NUM_THREADS` is set.
+Returns a CPU architecture for building Oceananigans models on a single CPU node.
 """
 struct CPU <: AbstractArchitecture end
 
-"""
-    GPU <: AbstractArchitecture
 
-Run Oceananigans on a single NVIDIA CUDA GPU.
+struct GPU{D} <: AbstractArchitecture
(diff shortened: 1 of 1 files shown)


src/Architectures.jl
```diff
@@ -60,19 +60,25 @@ Abstract supertype for Distributed architectures supported by Oceananigans.
...
-    GPU <: AbstractArchitecture
+    GPU(device)
+
+Returns a GPU architecture for building Oceananigans models on a GPU `device`.
 
-Run Oceananigans on a single NVIDIA CUDA GPU.
+Currently, only Nvidia CUDA GPUs are supported.
 """
-struct GPU <: AbstractArchitecture end
+GPU() = GPU(CUDA.device())
```
simone-silvestri, line 81:
```suggestion
GPU() = CUDA.has_cuda_gpu() ? GPU(CUDA.device()) : GPU(nothing)
```

src/Architectures.jl
```diff
@@ -60,19 +60,31 @@ Abstract supertype for Distributed architectures supported by Oceananigans.
...
 
-Run Oceananigans on a single NVIDIA CUDA GPU.
+Currently, only Nvidia CUDA GPUs are supported.
 """
-struct GPU <: AbstractArchitecture end
+function GPU()
+    if CUDA.has_cuda_gpu() 
+        return GPU(CUDA.device()) 
+    else 
+        @warn "No cuda capable device found! Reconsider running with architecture = CPU()"
```
glwagner, line 84: @simone-silvestri isn't "cuda" capitalized? We capitalize it everywhere else, and the Julia package is "CUDA.jl". Let us know...

glwagner:
This is easily revived but is currently stale.

PR #2634, opened 2022-06-30
small bugfix for `ContinuousBoundaryFunction`

simone-silvestri:


navidcy (approved):

glwagner:
I guess this isn't tested because otherwise it would throw an error?

PR #2642, opened 2022-07-05
[WIP] Advection Refactor

simone-silvestri:
This PR sets out to refactor advection along the lines proposed in #2454
A final form for the new advection scheme is still to be decided though


glwagner:
Does this branch have the bugfix for high-order advection with immersed boundaries?

simone-silvestri:
It's in #2603 which I am planning to merge before this

PR #2646, opened 2022-07-07
Minor changes in Langmuir and Ocean wind mixing examples

tomchor:
This PR:
- The Langmuir example currently claims that the buoyancy flux in the 2021 paper is `2.307e-9 # m² s⁻³, surface buoyancy flux` and sets up the buoyancy flux accordingly. But the paper actually sets up a flux of `2.307e-8` so I changed that (I think it was a typo). @glwagner can you please confirm that I'm not missing anything?

- Updates the CPU -> GPU functionality done via grid constructor instead via model in the Ocean wind mixing example.

Docs should render here: It should render here: https://clima.github.io/OceananigansDocumentation/previews/PR2646

### DIFF
--- a/examples/langmuir_turbulence.jl
+++ b/examples/langmuir_turbulence.jl
@@ -47,7 +47,7 @@ grid = RectilinearGrid(size=(32, 32, 32), extent=(128, 128, 64))
 using Oceananigans.BuoyancyModels: g_Earth
 
  amplitude = 0.8 # m
-wavelength = 60 # m
+wavelength = 60  # m
 wavenumber = 2π / wavelength # m⁻¹
  frequency = sqrt(g_Earth * wavenumber) # s⁻¹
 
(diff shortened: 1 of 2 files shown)


navidcy:
Hm. It’s only an example so we can put whatever buoyancy flux we like. But it doesn’t hurt we follow a paper. Wondering whether 10x more Q will require smaller timestep though.

navidcy (approved):
Let’s make sure that we see the docs preview and that all is good before merging.

navidcy:
I think previews appear at the second commit. So I did a commit yo see ;)

tomchor:
> I think previews appear at the second commit. So I did a commit yo see ;)

Hmm. It's still not rendering for me. The url appears to be correct. https://github.com/CliMA/Oceananigans.jl/pull/2648 is rendering fine so I don't know what's wrong here

navidcy:
It renders fine. Perhaps you tried before tge docs were deployed?

tomchor:
> It renders fine. Perhaps you tried before tge docs were deployed?

Yeah, I think you're right. It's rendering now. The new video looks good to me!

francispoulin (approved):

navidcy:
@tomchor I also pushed a change in the ocean mixing example. Could you have a look? Thought I combine the two in this PR...

tomchor:
> @tomchor I also pushed a change in the ocean mixing example. Could you have a look? Thought I combine the two in this PR...

I like it! Should I merge? Or should we wait for @glwagner since his paper is the one that's cited for the Langmuir example?

PR #2651, opened 2022-07-12
Bugfix for `DiscreteDiffusionFunction` + passes prognostic + auxiliary fields to forcing functions

simone-silvestri:
small bugfix for which `discrete_form=form` and `Fields` were not working for tracers.

Also passing all the fields to the forcing function instead of only the prognostic ones


navidcy:
Any test? Only if it's easy... Otherwise forget it.

navidcy (approved):

simone-silvestri:
Added a test, I ll merge when tests pass

PR #2664, opened 2022-07-15
Fixes to Gent McWilliams parametrization

simone-silvestri:
with the current scheme tapering was not applied correctly and slopes would increase until blowing up (and there was a very devious bug in the `SmallSlopeIsopycnalTensor` constructor). 
This PR fixes it. 

A one degree lat-lon simulation with GM (k_skew=1000, k_symmetric=900) is incoming...


navidcy:
I'll have a look!

src/TurbulenceClosures/isopycnal_rotation_tensor_components.jl
```diff
@@ -52,10 +52,11 @@ struct SmallSlopeIsopycnalTensor{FT} <: AbstractIsopycnalTensor
     minimum_bz :: FT
 end
 
-SmallSlopeIsopycnalTensor(; minimum_bz = 0) = SmallSlopeIsopycnalTensor(minimum_bz)
+SmallSlopeIsopycnalTensor(FT::DataType=Float64; minimum_bz = FT(0)) = SmallSlopeIsopycnalTensor(minimum_bz)
 
-@inline function isopycnal_rotation_tensor_xz_fcc(i, j, k, grid::AbstractGrid{FT}, buoyancy, tracers, slope_model::SmallSlopeIsopycnalTensor) where FT
+@inline function isopycnal_rotation_tensor_xz_fcc(i, j, k, grid::AbstractGrid{FT}, buoyancy, tracers, slope_model::SmallSlopeIsopycnalTensor, slope_limiter) where FT
     bx = ∂x_b(i, j, k, grid, buoyancy, tracers)
+    by = ℑxyᶠᶜᵃ(i, j, k, grid, ∂yᶜᶠᶜ, buoyancy_perturbation, buoyancy.model, tracers)
```
navidcy, line 59: why we need this here?

navidcy:
Can you explain what was the issue before? Was the `taper_factor_ccc` function wrong?

simone-silvestri:
because we have to taper _inside_ the `R` terms, if we just taper outside it will not have the intended effect

This is because `bz` is very small (can go down to 1e-20) so calculating the slope is quite a delicate operation. In  a realistic simulation (as this one) it can well be (and it happens) that bz changes of 4-5 orders of magnitude between face and centers and this means that the taper factor (which is always calculated at centers) does not "see" the actual slope that you calculate in the rotation tensor, leaving you with R factors of the order of 1e10.

In this way, if you taper locally, you ensure that all elements of the tensor are tapered properly

src/TurbulenceClosures/isopycnal_rotation_tensor_components.jl
simone-silvestri: to calculate the tapering factor, we can probably do it more efficiently, I have't worked out the details completely yet

navidcy: but this function never uses it

navidcy:
Gotcha! Damn... we missed that when we implemented with @glwagner. I remember we were pondering whether it would matter where to taper...

Good catch.

simone-silvestri:
yeah, we could define `taper_factor_fcc`, `taper_factor_cfc` and `taper_factor_ccf` to have it more clean

src/TurbulenceClosures/isopycnal_rotation_tensor_components.jl
navidcy: I believe you missed the tapering in this function.

navidcy:
Can I push a commit?

navidcy:
> yeah, we could define `taper_factor_fcc`, `taper_factor_cfc` and `taper_factor_ccf` to have it more clean

I think If we do that and call them inside the `isopycnal_rotation_tensor_...` functions that would imply computing the slopes twice. What you did seems more efficient. I'll just add a remark in the docstring for how the tapering is done.

src/TurbulenceClosures/isopycnal_rotation_tensor_components.jl
navidcy: I believe you missed the tapering in this function...

simone-silvestri: Ah yeah! I was using #2603 😅, sure

navidcy:
@simone-silvestri feel free to revert back the changes in [943c782](https://github.com/CliMA/Oceananigans.jl/pull/2664/commits/943c782e1c9d4a62e91cb5d2d0a96723fbaf3b92). I think you were missing a tapering in `isopycnal_rotation_tensor_xz_fcc`.

navidcy (approved):

src/TurbulenceClosures/isopycnal_rotation_tensor_components.jl
```diff
@@ -47,54 +47,81 @@ rotation tensor are:
...
 """
 struct SmallSlopeIsopycnalTensor{FT} <: AbstractIsopycnalTensor
     minimum_bz :: FT
 end
 
-SmallSlopeIsopycnalTensor(; minimum_bz = 0) = SmallSlopeIsopycnalTensor(minimum_bz)
+SmallSlopeIsopycnalTensor(FT::DataType=Float64; minimum_bz = FT(0)) = SmallSlopeIsopycnalTensor(minimum_bz)
 
-@inline function isopycnal_rotation_tensor_xz_fcc(i, j, k, grid::AbstractGrid{FT}, buoyancy, tracers, slope_model::SmallSlopeIsopycnalTensor) where FT
+@inline function isopycnal_rotation_tensor_xz_fcc(i, j, k, grid::AbstractGrid, buoyancy, tracers, slope_model::SmallSlopeIsopycnalTensor, slope_limiter)
```
francispoulin, line 66: Do we want to split this into two lines, so that it's easier to read?

src/TurbulenceClosures/isopycnal_rotation_tensor_components.jl
francispoulin: This comment above is about `bx` I presume, since `by` and `bz` are both used?

src/TurbulenceClosures/isopycnal_rotation_tensor_components.jl
navidcy: Done

src/TurbulenceClosures/isopycnal_rotation_tensor_components.jl
navidcy: Now they are after [943c782](https://github.com/CliMA/Oceananigans.jl/pull/2664/commits/943c782e1c9d4a62e91cb5d2d0a96723fbaf3b92)

glwagner:
If we have different taper functions at different locations, this implies tthat GM is no longer stricly isopycnal, right?

simone-silvestri:
you can decide to take the minimum at all locations if you prefer

glwagner:
This was the original rationale for using a single tapering function. It's mentioned as important in some papers (but I'd have to dig that up). Of course, if it's not _possible_ to simultaneously have stability _and_ adiabaticity then we should choose stability. But if we can have both that is definitely preferred.

simone-silvestri:
then the minimum of the tapering on `ccf`, `cfc` and `fcc` is what you want, we can implement that

glwagner:
Perhaps this simply means that we should keep the name `taper_factor_ccc`, but redefine its implementation to be more stable along the lines you're suggesting?

simone-silvestri:
sure, or just `minimum_taper_factor`

glwagner:
If we only have one implementation of a taper factor (and it doesn't have a location) then I guess it's just `taper_factor`.

simone-silvestri:
Merge when tests pass?

simone-silvestri:
I ll go ahead and merge this, in the end this PR calculates the tapering factor as the minimum of the tapering at `fcc`, `cfc` and `ccf`. Haven't really tested for stability but I guess @sandreza will run 1 degree simulations which will give us an answer. If this method is less stable we can always revert to calculating the tapering _inside_ the tensor although that gives us a flux which is not formally adiabatic

src/TurbulenceClosures/turbulence_closure_implementations/isopycnal_skew_symmetric_diffusivity.jl
```diff
@@ -116,36 +116,66 @@ struct FluxTapering{FT}
...
-    by = ℑyᵃᶜᵃ(i, j, k, grid, ∂y_b, buoyancy, tracers)
-    bz = ℑzᵃᵃᶜ(i, j, k, grid, ∂z_b, buoyancy, tracers)
+@inline function tapering_factor(i, j, k, grid, closure, tracers, buoyancy)
 
-    slope_x = - bx / bz
-    slope_y = - by / bz
-    slope² = ifelse(bz <= 0, zero(grid), slope_x^2 + slope_y^2)
+    ϵᶠᶜᶜ = tapering_factorᶠᶜᶜ(i, j, k, grid, closure, tracers, buoyancy)
+    ϵᶜᶠᶜ = tapering_factorᶜᶠᶜ(i, j, k, grid, closure, tracers, buoyancy)
+    ϵᶜᶜᶠ = tapering_factorᶜᶜᶠ(i, j, k, grid, closure, tracers, buoyancy)
```
glwagner, line 134: What's the difference between "tapering factor" and "taper factor"? Is this just a typo?

src/TurbulenceClosures/turbulence_closure_implementations/isopycnal_skew_symmetric_diffusivity.jl
```diff
@@ -116,36 +116,66 @@ struct FluxTapering{FT}
...
 
 """
-    taper_factor_ccc(i, j, k, grid::AbstractGrid{FT}, buoyancy, tracers, tapering::FluxTapering) 
+    taper_factor(i, j, k, grid, closure, tracers, buoyancy) 
 
 Return the tapering factor `min(1, Sₘₐₓ² / slope²)`, where `slope² = slope_x² + slope_y²`
-that multiplies all components of the isopycnal slope tensor. All slopes involved in the
-tapering factor are computed at the cell centers.
+that multiplies all components of the isopycnal slope tensor. The tapering factor is calculated on all the
+faces involved in the isopycnal slope tensor calculation. The minimum value of tapering is selected.
```
glwagner, line 123: Should we select the minimum, or maximum? Wouldn't maximum be more conservative?

glwagner: @simone-silvestri

PR #2674, opened 2022-07-21
`ImplicitFreeSurface` default in `HydrostaticFreeSurfaceModel`

simone-silvestri:


glwagner (approved):

glwagner:
Should we set gravitational acceleration explicitly in the default?

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
```diff
@@ -44,8 +44,8 @@ end
 GridFittedBottom(bottom_height) = GridFittedBottom(bottom_height, CenterImmersedCondition())
 
 function Base.summary(ib::GridFittedBottom)
-    hmax = maximum(ib.bottom_height)
-    hmin = minimum(ib.bottom_height)
+    hmax = maximum(parent(ib.bottom_height))
+    hmin = minimum(parent(ib.bottom_height))
```
glwagner, line 48: What's the issue?

simone-silvestri: scalar indexing when `ib.bottom_height` is an `OffsetArray{CuArray}`

glwagner: Ah, I forgot that it wasn't a field.

PR #2679, opened 2022-07-23
Fix a few typos in README, docs and docstrings

pitmonticone:


tomchor (approved):

francispoulin (approved):

docs/src/appendix/benchmarks.md
```diff
@@ -143,7 +143,7 @@ Environment:
 └─────────────┴─────┴─────────┴─────────┴─────────┘
 ```
 
-Like the shallow water model, it can be seen at grid size `64^3` that the GPU is beginning to be saturated as speedups rapidly increase. At grid sizes `128^3` and `256^3` we see the speedup stablise to around 400 times.
+Like the shallow water model, it can be seen at grid size `64^3` that the GPU is beginning to be saturated as speedups rapidly increase. At grid sizes `128^3` and `256^3` we see the speedup stabilise to around 400 times.
```
glwagner, line 146: Should be "stabilize"!

glwagner:
> No description provided.

Thanks!

pitmonticone:
> Thanks!

Thank you all for the development and maintenance of this wonderful Julia package!

docs/src/appendix/benchmarks.md
pitmonticone: Ops I'm sorry. I've assumed you wanted to adopt the British version.

Issue #2681, opened 2022-07-25
Strange behaviour of Lagrangian Particles near top of domain


jagoosw:
Hi all,

I've come across a strange issue with Lagrangian particles where if they're near the top of the domain they always jump down by Δz after the first intermediate time step. I have tracked this down to the boundary condition enforcement where there is an off by one error choosing the maximum allowed position. The error would also have occurred in the x and y directions too.

I will pull request this fix: https://github.com/jagoosw/Oceananigans.jl/tree/particle_tracking

I realized @simone-silvestri is working on an overhaul of Lagrangian particle tracking but it seems to not be fixed in that branch either.

Hope this helps!

MWE:
```
using Oceananigans

grid = RectilinearGrid(size=(2,2,2), extent=(2,2,2))
particles =  LagrangianParticles(x=[0.5], y=[0.5], z=[-0.5])
model = NonhydrostaticModel(; grid, timestepper=:RungeKutta3, particles=particles)
set!(model, u=0, v=0, w=0)
sim = Simulation(model, Δt=1, stop_time=1)
printz(sim)=println(sim.model.particles.properties.z)
sim.callbacks[:z] = Callback(printz)
run!(sim)
```
Result:
```
[ Info: Initializing simulation...
[-0.5]
[ Info:     ... simulation initialization complete (429.036 ms)
[ Info: Executing initial time step...
[ Info:     ... initial time step complete (22.760 seconds).
[ Info: Simulation is stopping. Model time 1 second has hit or exceeded simulation stop time 1 second.
[-1.5]
```
Results with fix
```
[ Info: Initializing simulation...
[-0.5]
[ Info:     ... simulation initialization complete (10.887 ms)
[ Info: Executing initial time step...
[ Info:     ... initial time step complete (1.866 ms).
[ Info: Simulation is stopping. Model time 1 second has hit or exceeded simulation stop time 1 second.
[-0.5]
```

tomchor:
Fixed by https://github.com/CliMA/Oceananigans.jl/pull/2682

PR #2682, opened 2022-07-25
Fixed Lagrangian Particle boundary condition enforcement issue #2681

jagoosw:
Fixes Issue #2681


glwagner:
Nice!

Is there a test we could add that might catch this bug?

jagoosw:
Now I've looked harder I've realized that this issue is only for bounded domains so my original fix is incorrect for periodic domains since they are [xmin, xmax) not [xmin, xmax] like bounded domains. Fixed this now.

I've changed the position of the particles in the existing tests so that they should be affected by the change if that works? For the x I put it on the edge so it gets advected out (and will have been looped round even without the fix), y near the boundary but should not get looped round, and for the z I've put it near the boundary where previously it will have been moved but should not have been.

simone-silvestri:
Nice! Good catch

src/LagrangianParticleTracking/update_particle_properties.jl
```diff
@@ -22,12 +22,47 @@ along a `Periodic` dimension, put them on the other side.
     return x
 end
 
+# Fallback if we skip one cell
+@inline adjust_coord(x, args...) where N = x
+
+@inline adjust_coord(x, nodefunc, i, ::Val{0} , grid, rest) = x
+@inline adjust_coord(x, nodefunc, i, ::Val{-1}, grid, rest) = nodefunc(Face(), i+1, grid) - (x - nodefunc(Face(), i+1, grid)) * rest
+@inline adjust_coord(x, nodefunc, i, ::Val{1} , grid, rest) = nodefunc(Face(), i, grid)   + (nodefunc(Face(), i, grid)   - x) * rest
```
glwagner, line 30: `ifelse` might be a better design here

simone-silvestri:
@jagoosw do you mind if I commit to try to fix the tests?

jagoosw:
> @jagoosw do you mind if I commit to try to fix the tests?

Not at all! They were working before but something must have changed. Thanks!

simone-silvestri:
@jagoosw it seems like everything is ok! When the tests pass I ll merge, good job

simone-silvestri (approved):

jagoosw:
@simone-silvestri great, thank you!

src/LagrangianParticleTracking/update_particle_properties.jl
```diff
@@ -50,17 +50,19 @@ position based on the previous position (we bounce back a certain restitution fr
...
+            zₚ = adjust_coord(zₚ, znode, kₒ, kᵈ, grid, restitution)
         end
     end
 
     return xₚ, yₚ, zₚ
 end
 
-@inline function update_particle_position!(particles, p, restitution, grid::AbstractGrid{FT, TX, TY, TZ}, Δt, velocities) where {FT, TX, TY, TZ}
+@inline maxnode(::Bounded, nodes, N) = nodes[N+1]
+@inline maxnode(::Periodic, nodes, N) = nodes[N]
```
glwagner, line 63: Missing @inbounds

PR #2687, opened 2022-07-28
Small bugfix

simone-silvestri:
I ve introduced a small bug in the previous PR, this solves it

### DIFF
--- a/src/LagrangianParticleTracking/update_particle_properties.jl
+++ b/src/LagrangianParticleTracking/update_particle_properties.jl
@@ -43,7 +43,7 @@ position based on the previous position (we bounce back a certain restitution fr
     k = Base.unsafe_trunc(Int, k)
    
     if immersed_cell(i, j, k, grid)
-        iₒ, jₒ, kₒ   = fractional_indices(old_pos..., (Center(), Center(), Center()), grid.underlying_grid)
+        iₒ, jₒ, kₒ = old_indices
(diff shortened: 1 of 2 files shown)


simone-silvestri:
Probably in a next PR we will want a bit more tests for Particle tracking on Immersed Boundary grids

src/LagrangianParticleTracking/update_particle_properties.jl
```diff
@@ -91,7 +91,7 @@ end
 @kernel function _advect_particles!(particles, restitution, grid::ImmersedBoundaryGrid, Δt, velocities)
     p = @index(Global)
 
-    old_pos = (particles.x[p], particles.y[p], particles.y[p])
+    old_pos = deepcopy((particles.x[p], particles.y[p], particles.z[p]))
```
glwagner, line 94: Do you need `deepcopy`?

simone-silvestri: the positions change in `update_particle_positions!` just below. I am not sure if we need deepcopy though...

glwagner: One (unstated) requirement of the particle tracking algorithm is to support huge numbers of particles (hundreds of millions or billions) so we likely want to avoid the allocation a `deepcopy` would incur.

glwagner (approved):

tomchor:
Bump patch version before merging so that we can tag a release?

PR #2688, opened 2022-07-30
Add GPU functionality to Multigrid solver

### LINKED ISSUE #2657: Add GPU functionality to Multigrid solver
We should make sure that multigrid solver works on GPU.

cc @elise-palethorpe

elise-palethorpe:
Closes #2657 

cc @navidcy


navidcy:
@elise-palethorpe, we need to modify:

https://github.com/CliMA/Oceananigans.jl/blob/5644b347a540fa64897e3db587c39bc9a87de5ed/test/test_implicit_free_surface_solver.jl#L97

src/Solvers/multigrid_solver.jl
```diff
@@ -128,30 +163,84 @@ function fill_matrix_elements!(A, template_field, linear_operator!, args...)
...
+
+    for k = 1:Nz, j in 1:Ny, i in 1:Nx
+        eᵢⱼₖ .= 0
+        ∇²eᵢⱼₖ .= 0
+        CUDA.@allowscalar eᵢⱼₖ[i, j, k] = 1
+        fill_halo_regions!(eᵢⱼₖ)
+        linear_operator!(∇²eᵢⱼₖ, eᵢⱼₖ, args...)
+        count = 0
+        for n = 1:Nz, m in 1:Ny, l in 1:Nx
+            # CUDA.@allowscalar value = ∇²eᵢⱼₖ[l, m, n]
```
navidcy, line 188: delete?

navidcy, line 188:
```suggestion
```

src/Solvers/multigrid_solver.jl
```diff
@@ -128,30 +163,84 @@ function fill_matrix_elements!(A, template_field, linear_operator!, args...)
...
     interior(x) .= reshape(solver.x_array, Nx, Ny, Nz)
 end
 
+function finalize_solver(solver::MultigridSolver)
+    s = solver.amgx_solver_struct
+    if s != nothing
+        close(s.device_matrix); close(s.device_x); close(s.device_b); close(s.amgx_solver); close(s.resources); close(s.config); 
+        AMGX.finalize_plugins(); AMGX.finalize()
+    end
+end
```
navidcy, line 237: few comments here

1. `finalize_solver` seems to modify its argument so perhaps rename to `finalize_solver!`?
2. instead of having an `if` statement use multiple dispatch?
  ```Julia
  julia> f(x, y) = x+y

  julia> f(x, ::Nothing) = 2x
  f (generic function with 2 methods)

  julia> f(2, 3)
  5

  julia> f(2, nothing)
  4
  ```

So perhaps something like:

```Julia
finalize_solver!(solver::MultigridSolver) = finalize_solver!(solver.amgx_solver_struct)
finalize_solver!(::Nothing) = nothing
function finalize_solver!(s::SomeOtherTypeOfTheAMGXMultigridSolver)
    close(s.device_matrix)
    close(s.device_x)
    close(s.device_b)
    close(s.amgx_solver)
    close(s.resources)
    close(s.config)
    AMGX.finalize_plugins()
    AMGX.finalize()
 end
```

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -184,6 +188,19 @@ function solve!(η, implicit_free_surface_solver::MGImplicitFreeSurfaceSolver, r
...
+        s = solver.amgx_solver_struct
+        if s != nothing
+            # FIXME only allocate this once
+            csr_matrix = CuSparseMatrixCSR(transpose(solver.matrix))
+            @inline sub_one(x) = convert(Int32, x-1)
+            AMGX.upload!(s.device_matrix, 
+                map(sub_one, csr_matrix.rowPtr),
+                map(sub_one, csr_matrix.colVal),
+                csr_matrix.nzVal
+                )
```
navidcy, line 200:
```suggestion
            AMGX.upload!(s.device_matrix, 
                         map(sub_one, csr_matrix.rowPtr),
                         map(sub_one, csr_matrix.colVal),
                         csr_matrix.nzVal
                         )
```

src/Solvers/multigrid_solver.jl
```diff
@@ -75,7 +84,7 @@ function MultigridSolver(linear_operation!::Function,
 
     arch = architecture(template_field)
 
-    arch == GPU() && error("Multigrid solver is only supported on CPUs.")
+    # arch == GPU() && error("Multigrid solver is only supported on CPUs.")
```
navidcy, line 87:
```suggestion
```

src/Solvers/multigrid_solver.jl
navidcy: I think this sort of philosophy can be used in other parts of this PR?

validation/elliptic_solvers/doubly_bounded_poisson_gpu.jl
```diff
@@ -0,0 +1,124 @@
...
+@info "Solving the Poisson equation with a conjugate gradient preconditioned iterative solver WITH algebraic multigrid as preconditioner..."
+@time solve!(φ_cgmg, cgmg_solver, r, arch, grid)
+
+fill_halo_regions!(φ_cgmg)
+finalize_solver(mgs)
+
+@show φ_fft
+@show φ_cg
+@show φ_mg
+@show φ_cgmg
```
navidcy, line 124:
```suggestion
@show φ_cgmg

```

navidcy: Is the `doubly_bounded_poisson_gpu.jl` script very different from the cpu version? can we consolidate the two scripts into one easily? (not high priority)

navidcy:
I think it'll be cleaner if we define two subtypes for `MultigridSolvers, e.g.

```Julia
abstract type MultigridSolver end

mutable struct MultigridCPUSolver <: MultigridSolver
    X
    Y
end

mutable struct MultigridGPUSolver <: MultigridSolver
    X
    Y
end

mg_cpu = MultigridCPUSolver(1, 2)

mg_gpu = MultigridGPUSolver(2, 3)

typeof(mg_cpu) <: MultigridSolver

typeof(mg_gpu) <: MultigridSolver
```

navidcy:
@elise-palethorpe, I don't see the AMGX package in `Project.toml` nor in `Manifest.toml`.

Btw, when I tried to add it locally on my Mac I got this:

```Julia
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.6.7 (2022-07-19)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(Oceananigans) pkg> st
     Project Oceananigans v0.77.1
      Status `~/Research/OC.jl/Project.toml`
  [c963dde9] AMGX v0.1.2
  [79e6a3ab] Adapt v3.3.3
  [2169fc97] AlgebraicMultigrid v0.5.1
  [052768ef] CUDA v3.11.0
  [72cfdca4] CUDAKernels v0.3.3
  [a8cc5b0e] Crayons v4.1.1
  [7445602f] CubedSphere v0.2.0
  [ffbed154] DocStringExtensions v0.8.6
  [7a1cc6ca] FFTW v1.5.0
  [c27321d9] Glob v1.3.0
  [40713840] IncompleteLU v0.2.0
  [42fd0dbc] IterativeSolvers v0.9.2
  [033835bb] JLD2 v0.4.22
  [63c18a36] KernelAbstractions v0.7.2
  [da04e1cc] MPI v0.19.2
  [85f8d34a] NCDatasets v0.12.4
  [6fe1bfb0] OffsetArrays v1.12.6
  [bac558e1] OrderedCollections v1.4.1
  [0e08944d] PencilArrays v0.17.5
  [4a48f351] PencilFFTs v0.14.0
  [6038ab10] Rotations v1.3.1
  [d496a93d] SeawaterPolynomials v0.2.3
  [09ab397b] StructArrays v0.6.11
  [bc48ee85] Tullio v0.3.4
  [ade2ca70] Dates
  [b77e0a4c] InteractiveUtils
  [37e2e46d] LinearAlgebra
  [56ddb016] Logging
  [44cfe95a] Pkg
  [de0858da] Printf
  [9a3f8284] Random
  [2f01184e] SparseArrays
  [10745b16] Statistics

(Oceananigans) pkg> precompile
Precompiling project...
  ✗ Oceananigans
  0 dependencies successfully precompiled in 11 seconds (99 already precompiled)

ERROR: The following 1 direct dependency failed to precompile:

Oceananigans [9e8cae18-63c1-5223-a75c-80ca9d6e9a09]

Failed to precompile Oceananigans [9e8cae18-63c1-5223-a75c-80ca9d6e9a09] to /Users/navid/.julia/compiled/v1.6/Oceananigans/jl_Z5b4Xf.
ERROR: LoadError: LoadError: LoadError: InitError: UndefVarError: libamgxsh not defined
Stacktrace:
  [1] getproperty
    @ ./Base.jl:26 [inlined]
  [2] __init__()
    @ AMGX ~/.julia/packages/AMGX/GFHHN/src/AMGX.jl:30
  [3] _include_from_serialized(path::String, depmods::Vector{Any})
    @ Base ./loading.jl:696
  [4] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String)
    @ Base ./loading.jl:782
  [5] _require(pkg::Base.PkgId)
    @ Base ./loading.jl:1020
  [6] require(uuidkey::Base.PkgId)
    @ Base ./loading.jl:936
  [7] require(into::Module, mod::Symbol)
    @ Base ./loading.jl:923
  [8] include(mod::Module, _path::String)
    @ Base ./Base.jl:384
  [9] include(x::String)
    @ Oceananigans.Solvers ~/Research/OC.jl/src/Solvers/Solvers.jl:1
 [10] top-level scope
    @ ~/Research/OC.jl/src/Solvers/Solvers.jl:48
 [11] include(mod::Module, _path::String)
    @ Base ./Base.jl:384
 [12] include(x::String)
    @ Oceananigans ~/Research/OC.jl/src/Oceananigans.jl:5
 [13] top-level scope
    @ ~/Research/OC.jl/src/Oceananigans.jl:195
 [14] include
    @ ./Base.jl:384 [inlined]
 [15] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::Nothing)
    @ Base ./loading.jl:1235
 [16] top-level scope
    @ none:1
 [17] eval
    @ ./boot.jl:360 [inlined]
 [18] eval(x::Expr)
    @ Base.MainInclude ./client.jl:446
 [19] top-level scope
    @ none:1
during initialization of module AMGX
in expression starting at /Users/navid/Research/OC.jl/src/Solvers/multigrid_solver.jl:5
in expression starting at /Users/navid/Research/OC.jl/src/Solvers/Solvers.jl:1
in expression starting at /Users/navid/Research/OC.jl/src/Oceananigans.jl:1
```

Was it smooth for you?

src/Solvers/multigrid_solver.jl
```diff
@@ -128,30 +163,84 @@ function fill_matrix_elements!(A, template_field, linear_operator!, args...)
...
+function fill_matrix_elements2!(template_field, linear_operator!, args...)
+    Nx, Ny, Nz = size(template_field)
+    make_column(f) = reshape(interior(f), Nx*Ny*Nz)
+
+    eᵢⱼₖ = similar(template_field)
+    ∇²eᵢⱼₖ = similar(template_field)
+
+    arch = architecture(template_field)
+    colptr = array_type(arch){Int64}(undef, Nx*Ny*Nz+1)
+    rowval = array_type(arch){Int64}(undef, 0)
```
navidcy, line 175:
```suggestion
    colptr = array_type(arch){Int}(undef, Nx*Ny*Nz+1)
    rowval = array_type(arch){Int}(undef, 0)
```

`Int` is either `Int64` or `Int32` depending on the machine you are on. Probably better idea?

src/Solvers/multigrid_solver.jl
```diff
@@ -86,55 +106,120 @@ function MultigridSolver(linear_operation!::Function,
...
+    AMGX.initialize_plugins()
+    # FIXME! Also pass tolerance
+    config = AMGX.Config(Dict("monitor_residual" => 1, "max_iters" => maxiter, "store_res_history" => 1, "tolerance" => reltol));
+    resources = AMGX.Resources(config)
+    solver = AMGX.Solver(resources, AMGX.dDDI, config)
+    device_matrix = AMGX.AMGXMatrix(resources, AMGX.dDDI)
+    device_b = AMGX.AMGXVector(resources, AMGX.dDDI)
+    device_x = AMGX.AMGXVector(resources, AMGX.dDDI)
+    csr_matrix = CuSparseMatrixCSR(transpose(matrix))
+    @inline sub_one(x) = convert(Int32, x-1)
```
navidcy, line 132:
```suggestion
    @inline subtract_one(x) = convert(Int32, x-1)
```

navidcy: Let's make the name more verbose. What do you think?

Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Climate Modeling Alliance and contributors"]
 version = "0.77.1"
```
navidcy, line 4:
```suggestion
version = "0.77.2"
```

navidcy:
The MG solver tests fail on the curvilinear grid.
https://buildkite.com/clima/oceananigans/builds/8380#0182bd4a-b305-42af-9390-a05e5b039826

I'm wondering whether it's because the grid spacing is not uniform...

navidcy:
@elise-palethorpe as you suggested, I think we should construct the matrix similarly to how the Heptadiagonal free-surface solver does because it takes forever... We should keep the functionality for construction any matrix from the `linear_operation!` function though somewhere?

navidcy:
@elise-palethorpe, look at [this error](https://buildkite.com/clima/oceananigans/builds/8479#01830c59-efb2-4373-89f9-4679bbb38cfd/36-801). I think it's something related with AMGX not being finalized...?

src/Solvers/multigrid_solver.jl
```diff
@@ -86,55 +104,136 @@ function MultigridSolver(linear_operation!::Function,
...
+    arch == GPU() && begin
+        try
+            print("trying")
+            AMGX.initialize()
+            AMGX.initialize_plugins()
+        catch e
+            AMGX.finalize_plugins()
+            AMGX.finalize()        
+            AMGX.initialize()
+            AMGX.initialize_plugins()
```
navidcy, line 127:
```suggestion
            AMGX.initialize()
            AMGX.initialize_plugins()
        catch e
            AMGX.finalize_plugins()
            AMGX.finalize()
            
            AMGX.initialize()
            AMGX.initialize_plugins()
```

src/Solvers/multigrid_solver.jl
```diff
@@ -86,55 +104,136 @@ function MultigridSolver(linear_operation!::Function,
...
+    make_column(f) = reshape(interior(f), Nx*Ny*Nz)
 
-    fill_matrix_elements!(A, template_field, linear_operator!, args...)
+    eᵢⱼₖ = similar(template_field)
+    ∇²eᵢⱼₖ = similar(template_field)
     
-    return arch_sparse_matrix(architecture(template_field), A)
+    for k = 1:Nz, j in 1:Ny, i in 1:Nx
+        eᵢⱼₖ .= 0
+        ∇²eᵢⱼₖ .= 0
```
navidcy, line 183:
```suggestion
        parent(eᵢⱼₖ) .= 0
        parent(∇²eᵢⱼₖ) .= 0
```

src/Solvers/multigrid_solver.jl
```diff
@@ -86,55 +104,136 @@ function MultigridSolver(linear_operation!::Function,
...
+
+    colptr = CuArray{Int}(undef, Nx*Ny*Nz+1)  # Can we infer the type of int?
+    rowval = CuArray{Int}(undef, 0)
+    nzval  = CuArray{FT}(undef, 0)
+
+    CUDA.@allowscalar colptr[1] = 1
+
     for k = 1:Nz, j in 1:Ny, i in 1:Nx
         eᵢⱼₖ .= 0
         ∇²eᵢⱼₖ .= 0
```
navidcy, line 211:
```suggestion
        parent(eᵢⱼₖ) .= 0
        parent(∇²eᵢⱼₖ) .= 0
```

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -56,17 +60,31 @@ step `Δt`, gravitational acceleration `g`, and free surface at the `n`-th time-
 """
 function MGImplicitFreeSurfaceSolver(grid::AbstractGrid, 
                                      settings = nothing,
-                                     gravitational_acceleration = nothing, 
+                                     gravitational_acceleration = g_Earth,
+                                     reduced_dim = (false, false, false),
```
navidcy, line 64: @elise-palethorpe are you sure about this? shouldn't it be `(false, false, true)`?

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -78,72 +96,17 @@ function MGImplicitFreeSurfaceSolver(grid::AbstractGrid,
     settings[:maxiter] = get(settings, :maxiter, grid.Nx * grid.Ny)
     settings[:reltol] = get(settings, :reltol, min(1e-7, 10 * sqrt(eps(eltype(grid)))))
 
-    right_hand_side = Field{Center, Center, Nothing}(grid)
-
     # initialize solver with Δt = nothing so that linear matrix is not computed;
     # see `initialize_matrix` methods
     solver = MultigridSolver(Az_∇h²ᶜᶜᶜ_linear_operation!, ∫ᶻ_Axᶠᶜᶜ, ∫ᶻ_Ayᶜᶠᶜ;
                              template_field = right_hand_side,
+                             matrix = matrix,
```
navidcy, line 103:
```suggestion
                             matrix,
```

Cleaner.

After `;` just giving a variable implies that you are passing a keyword argument whose name is the variable's name...

```Julia
julia> a = 1; b = 2;

julia> (; a, b)
(a = 1, b = 2)

julia> (; a, b) == (a=a, b=b)
true
```

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -213,4 +206,4 @@ function compute_implicit_free_surface_right_hand_side!(rhs, implicit_solver::MG
     
     wait(device(arch), event)
     return nothing
-end
+end
```
navidcy, line 209:
```suggestion
end

```

navidcy:
@elise-palethorpe have you pushed the latest non-allocating code?

elise-palethorpe:
> @elise-palethorpe have you pushed the latest non-allocating code?

No sorry, I forgot to push! I have just now

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -56,17 +60,31 @@ step `Δt`, gravitational acceleration `g`, and free surface at the `n`-th time-
...
+
     # Initialize vertically integrated lateral face areas
     ∫ᶻ_Axᶠᶜᶜ = Field{Face, Center, Nothing}(with_halo((3, 3, 1), grid))
     ∫ᶻ_Ayᶜᶠᶜ = Field{Center, Face, Nothing}(with_halo((3, 3, 1), grid))
 
     vertically_integrated_lateral_areas = (xᶠᶜᶜ = ∫ᶻ_Axᶠᶜᶜ, yᶜᶠᶜ = ∫ᶻ_Ayᶜᶠᶜ)
 
     compute_vertically_integrated_lateral_areas!(vertically_integrated_lateral_areas)
+    coeffs = compute_matrix_coefficients(vertically_integrated_lateral_areas, grid, gravitational_acceleration)
+    matrix_constructors, diagonal, problem_size = matrix_from_coefficients(arch, right_hand_side, coeffs, reduced_dim)
```
navidcy, line 78: why second argument here is `right_hand_side`? where is `matrix_from_coefficients` defined?

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -56,17 +60,31 @@ step `Δt`, gravitational acceleration `g`, and free surface at the `n`-th time-
...
+
+    # Placeholder preconditioner and matrix are calculated using a "placeholder" timestep of -1.0
+    # They are then recalculated before the first time step of the simulation.
+
+    placeholder_constructors = deepcopy(matrix_constructors)
+    M = prod(problem_size)
+    update_diag!(placeholder_constructors, arch, M, M, diagonal, 1.0, 0)
+
+    matrix = arch_sparse_matrix(arch, placeholder_constructors)
     fill_halo_regions!(vertically_integrated_lateral_areas)
```
navidcy, line 88: should we fill the halos *before* we use `vertically_integrated_lateral_areas` in the matrix construction?

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -78,72 +96,20 @@ function MGImplicitFreeSurfaceSolver(grid::AbstractGrid,
...
-                      dependencies = device_event(arch))
-    wait(event_c)
-    return diag
-end
-
-@kernel function _compute_diag!(diag, grid, g)
-    i, j = @index(Global, NTuple)
-    @inbounds diag[i, j, 1]  = - Azᶜᶜᶜ(i, j, 1, grid) / g
-end
+finalize_solver!(::Any) = nothing
```
navidcy, line 112: actually this `(::Any)` should replace the `finalize_solver!(::MultigridCPUSolver) = nothing` in the `Solvers/multigrid_solver.jl`. Then nothing extra is needed here.

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -199,6 +165,35 @@ function solve!(η, implicit_free_surface_solver::MGImplicitFreeSurfaceSolver, r
...
+        unsafe_free!(constructors)
+
+        s = solver.amgx_solver
+        solver.amgx_solver.csr_matrix = CuSparseMatrixCSR(transpose(solver.matrix))
+        @inline subtract_one(x) = convert(Int32, x-1)
+        AMGX.upload!(s.device_matrix, 
+                        map(subtract_one, solver.amgx_solver.csr_matrix.rowPtr),
+                        map(subtract_one, solver.amgx_solver.csr_matrix.colVal),
+                        solver.amgx_solver.csr_matrix.nzVal
+                        )
```
navidcy, line 188:
```suggestion
        AMGX.upload!(s.device_matrix, 
                     map(subtract_one, solver.amgx_solver.csr_matrix.rowPtr),
                     map(subtract_one, solver.amgx_solver.csr_matrix.colVal),
                     solver.amgx_solver.csr_matrix.nzVal
                     )
```

alignment :)

src/Solvers/multigrid_solver.jl
```diff
@@ -59,25 +81,24 @@ Arguments
 * `reltol, abstol`: Relative and absolute tolerance for convergence of the algorithm.
                     The iteration stops when `norm(A * x - b) < max(reltol * norm(b), abstol)`.
 
-* `amg_algorithm`: Algebraic Multigrid algorithm defining mapping between different grid spacings
-
-!!! compat "Multigrid solver on GPUs"
-    Currently Multigrid solver is only supported on CPUs.
+* `amg_algorithm`: Algebraic Multigrid algorithm defining mapping between different grid spacings.
+                   Note: This keyword is relevant only on the CPU.
```
navidcy, line 85:
```suggestion
                   Options are `RugeStubenAMG()` (default) or `SmoothedAggregationAMG()`.
                   Note: This keyword is relevant *only* on the CPU.
```

src/Solvers/multigrid_solver.jl
```diff
@@ -86,55 +107,145 @@ function MultigridSolver(linear_operation!::Function,
...
+        resources = AMGX.Resources(config)
+        solver = AMGX.Solver(resources, AMGX.dDDI, config)
+        device_matrix = AMGX.AMGXMatrix(resources, AMGX.dDDI)
+        device_b = AMGX.AMGXVector(resources, AMGX.dDDI)
+        device_x = AMGX.AMGXVector(resources, AMGX.dDDI)
+        csr_matrix = CuSparseMatrixCSR(transpose(matrix))
+        
+        # # FIXME do on GPU
+        # loop! = zero_index!(Architectures.device(arch), 16, length(csr_matrix.rowPtr))
+        # event = loop!(csr_matrix.rowPtr; dependencies=device_event(arch))
```
navidcy, line 148: is there something to be done here?

src/Solvers/multigrid_solver.jl
```diff
@@ -86,55 +107,145 @@ function MultigridSolver(linear_operation!::Function,
...
+                                  b_array,
+                                  amgx_solver
+                                  )
+    end
 end
 
+# @kernel function zero_index!(cuarray :: CuArray)
+#     i = @index(Global, Linear)
+#     cuarray[i] = convert(Int32, cuarray[i] - 1)
+# end
```
navidcy, line 184: ?

src/Solvers/multigrid_solver.jl
```diff
@@ -86,55 +107,145 @@ function MultigridSolver(linear_operation!::Function,
...
     Nx, Ny, Nz = size(template_field)
+    FT = eltype(template_field.grid)
+
     make_column(f) = reshape(interior(f), Nx*Ny*Nz)
 
     eᵢⱼₖ = similar(template_field)
     ∇²eᵢⱼₖ = similar(template_field)
-    
+
+    colptr = CuArray{Int}(undef, Nx*Ny*Nz+1)  # Can we infer the type of int?
```
navidcy, line 216:
```suggestion
    colptr = CuArray{Int}(undef, Nx*Ny*Nz+1)
```

src/Solvers/multigrid_solver.jl
```diff
@@ -86,55 +107,145 @@ function MultigridSolver(linear_operation!::Function,
...
+            AMGX.initialize_plugins()
+        catch e
+            AMGX.finalize_plugins()
+            AMGX.finalize()        
+            AMGX.initialize()
+            AMGX.initialize_plugins()
+        end
+
+        if reltol == 0
+            @info "Multigrid solver will use abstol = ($abstol)"
```
navidcy, line 133:
```suggestion
            @info "Multigrid solver will use abstol = $abstol"
```

src/Solvers/multigrid_solver.jl
```diff
@@ -86,55 +107,145 @@ function MultigridSolver(linear_operation!::Function,
...
+            config = AMGX.Config(Dict("monitor_residual" => 1, "max_iters" => maxiter, "store_res_history" => 1, "tolerance" => abstol))
+        else
+            @info "Multigrid solver will use reltol = ($reltol)"
```
navidcy, line 136:
```suggestion
            @info "Multigrid solver with relative tolerance = $reltol"
```

navidcy:
@glwagner [AMGX.jl](https://github.com/JuliaGPU/AMGX.jl) only has prebuilt libraries for linux systems. That said, calling `using AMGX` will fail on, e.g., Mac OS X. What's the best way to optionally load AMGX.jl? Or is there another way around that?

src/Solvers/multigrid_solver.jl
elise-palethorpe: Yes, I was trying to do the `subtract_one` on a gpu which should be very straight forward but I'm getting 
```
Make sure your kernel function ends in `return`, `return nothing` or `nothing`.
If the returned value is of type `Union{}`, your Julia code probably throws an exception.
Inspect the code with `@device_code_warntype` for more details.
```

elise-palethorpe:
Some benchmarks from running `implicit_surface_solvers_benchmark.jl` 

On GPU: 
FFT: 0.902434 seconds (1.58 M allocations: 549.219 MiB, 13.25% gc time) 
PCG:  37.329798 seconds (17.07 M allocations: 2.563 GiB, 1.85% gc time)
Heptadiagonal:  4.323861 seconds (6.06 M allocations: 675.694 MiB, 4.72% gc time, 1.32% compilation time)
MG:   4.495123 seconds (1.50 M allocations: 492.900 MiB, 2.16% gc time)
PCG with FFT preconditioner:  22.673894 seconds (3.53 M allocations: 864.587 MiB, 0.77% gc time)

On CPU:
FFT:   6.548544 seconds (3.61 M allocations: 1.910 GiB, 14.07% gc time, 1.30% compilation time)
PCG:   32.354310 seconds (20.29 M allocations: 7.041 GiB, 5.80% gc time)
Heptadiagonal:   9.795938 seconds (3.51 M allocations: 1.595 GiB, 39.13% gc time)
MG:  11.868795 seconds (3.85 M allocations: 3.991 GiB, 43.10% gc time)
PCG with FFT preconditioner: 10.473279 seconds (5.38 M allocations: 2.654 GiB, 9.37% gc time)

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
elise-palethorpe: Yes! Thank you :)

navidcy:
@elise-palethorpe I merged #2654 so we can add the MGPCG in the benches?

navidcy:
We have variables `∇²eᵢⱼₖ` implying that `∇²` acts on `eᵢⱼₖ`. But that's not always the case, right?

Let's change all `∇²eᵢⱼₖ` -> `Aeᵢⱼₖ`? And add a small remark, e.g.,

```Julia
    eᵢⱼₖ = similar(template_field)   # the basis vector
    Aeᵢⱼₖ = similar(template_field)  # A * eᵢⱼₖ
```

navidcy:
@simone-silvestri can you review this?

navidcy:
@elise-palethorpe, can we combine the two `doubly_bounded_poisson.jl` scripts into one at some point?

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -20,7 +22,9 @@ The multigrid implicit free-surface solver.
 
 $(TYPEDFIELDS)
 """
-mutable struct MGImplicitFreeSurfaceSolver{S, V, F, R, C, D}
+mutable struct MGImplicitFreeSurfaceSolver{A, S, V, F, R, C, D}
+    "Architecture"
+    architecture :: A
     "The multigrid solver"
     multigrid_solver :: S
     "The vertically-integrated lateral areas"
```
simone-silvestri, line 30: You don't need this right?

src/Solvers/multigrid_solver.jl
```diff
@@ -86,54 +108,144 @@ function MultigridSolver(linear_operation!::Function,
...
-        A[:, Ny*Nx*(k-1) + Nx*(j-1) + i] .= make_column(∇²eᵢⱼₖ)
+        count = 0
+        for n = 1:Nz, m in 1:Ny, l in 1:Nx
+            CUDA.@allowscalar if ∇²eᵢⱼₖ[l, m, n] != 0
+                append!(rowval, Ny*Nx*(n-1) + Nx*(m-1) + l)
+                CUDA.@allowscalar append!(nzval, ∇²eᵢⱼₖ[l, m, n])
+                count += 1
+            end
+        end
+        CUDA.@allowscalar colptr[Ny*Nx*(k-1) + Nx*(j-1) + i + 1] = colptr[Ny*Nx*(k-1) + Nx*(j-1) + i] + count
```
simone-silvestri, line 237: can you do this on a GPU? If not, I would just do everything on a CPU and in the end copy it in the GPU with `CuArray`, @allowscalar is veeeery slow because you are continuously copying from GPU to CPU

simone-silvestri:
> Some benchmarks from running `implicit_surface_solvers_benchmark.jl`
> 
> On GPU: FFT: 0.902434 seconds (1.58 M allocations: 549.219 MiB, 13.25% gc time) PCG: 37.329798 seconds (17.07 M allocations: 2.563 GiB, 1.85% gc time) Heptadiagonal: 4.323861 seconds (6.06 M allocations: 675.694 MiB, 4.72% gc time, 1.32% compilation time) MG: 4.495123 seconds (1.50 M allocations: 492.900 MiB, 2.16% gc time) PCG with FFT preconditioner: 22.673894 seconds (3.53 M allocations: 864.587 MiB, 0.77% gc time)
> 
> On CPU: FFT: 6.548544 seconds (3.61 M allocations: 1.910 GiB, 14.07% gc time, 1.30% compilation time) PCG: 32.354310 seconds (20.29 M allocations: 7.041 GiB, 5.80% gc time) Heptadiagonal: 9.795938 seconds (3.51 M allocations: 1.595 GiB, 39.13% gc time) MG: 11.868795 seconds (3.85 M allocations: 3.991 GiB, 43.10% gc time) PCG with FFT preconditioner: 10.473279 seconds (5.38 M allocations: 2.654 GiB, 9.37% gc time)

nice

A problem with having an iterative matrix solver algorithm (like the pcg) is going to high resolutions (or conversely large time steps). 
The matrix becomes poorly conditioned because the diagonal (which depends on `Δt²`) decreases faster than the off-diagonals (which depend on the integrated areas) and the number of iterations required to converge spike up.
This means that on a GPU both eddy-permitting and resolving simulations are dominated by the cost of the implicit solver.
Our default preconditioner (for the Heptadiagonal solver) also loses efficiency when increasing `Δt²` (or increasing resolution) because it assumes that the matrix is diagonally dominant

I would be very interesting to know if with a multigrid algorithm this problem is less severe.

simone-silvestri:
can we put `validation/implicit_free_surface/implicit_surface_solvers_benchmark.jl` in the benchmark folder?

src/Solvers/multigrid_solver.jl
```diff
@@ -1,22 +1,45 @@
...
+                        matrix :: L
+                        abstol :: T
+                        reltol :: T
+                       maxiter :: Int
+                       x_array :: F
+                       b_array :: F
+                 amg_algorithm :: M
+end
+
+mutable struct MultigridGPUSolver{A, G, L, T, F, S} <: MultigridSolver{A, G, L, T, F}
```
simone-silvestri, line 23: it seems that `MultiGridCPUSolver` and `MultiGridGPUSolver` have the same fields. 

Can you just define 
  ```
mutable struct MultigridSolver{A, G, L, T, F, M} 
                  architecture :: A
                          grid :: G
                        matrix :: L
                        abstol :: T
                        reltol :: T
                       maxiter :: Int
                       x_array :: F
                       b_array :: F
                 amg_algorithm :: M
end

const MultigridCPUSolver = MultigridSolver{<:CPU}
const MultigridGPUSolver = MultigridSolver{<:GPU}
```

Comment (navidcy) [reply src/Solvers/multigrid_solver.jl:23]: Oh, they don't. But your suggestion might come handy indeed! I had forgotten that this `MultigridSolver{<:CPU}` would work!

File: src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl (hunk 25 above)
Comment (navidcy) [reply src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl:30]: Nice catch!

Comment (navidcy) [reply src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl:30]: Also @elise-palethorpe, for the fields below I'd suggest the somehow more explanatory docstrings:

```julia
    "The matrix constructors of the linear operator without the diagonal `Az / (g Δt²)` term"
    matrix_constructors :: C
    "The diagonal matrix term `Az / g`"
    diagonal :: D
```

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -56,10 +60,13 @@ step `Δt`, gravitational acceleration `g`, and free surface at the `n`-th time-
...
+                                     gravitational_acceleration = g_Earth,
+                                     reduced_dim = (false, false, true),
                                      placeholder_timestep = -1.0)
     arch = architecture(grid)
 
+    right_hand_side = Field{Center, Center, Nothing}(grid)
+
     # Initialize vertically integrated lateral face areas
     ∫ᶻ_Axᶠᶜᶜ = Field{Face, Center, Nothing}(with_halo((3, 3, 1), grid))
     ∫ᶻ_Ayᶜᶠᶜ = Field{Center, Face, Nothing}(with_halo((3, 3, 1), grid))
```
navidcy, line 72: @simone-silvestri is 3 halos enough for this here?

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -78,72 +97,19 @@ function MGImplicitFreeSurfaceSolver(grid::AbstractGrid,
     settings[:maxiter] = get(settings, :maxiter, grid.Nx * grid.Ny)
     settings[:reltol] = get(settings, :reltol, min(1e-7, 10 * sqrt(eps(eltype(grid)))))
```
navidcy, line 98: is this `10 * ` left over from debugging?

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
simone-silvestri: yep, actually one is enough for everything except for the `DiagonallyDominantInversePreconditioner` which does not work anyway at the moment

navidcy:
> 

@simone-silvestri any ideas for how to solve this?

src/Solvers/multigrid_solver.jl
```diff
@@ -86,54 +108,144 @@ function MultigridSolver(linear_operation!::Function,
...
+        solver = AMGX.Solver(resources, AMGX.dDDI, config)
+        device_matrix = AMGX.AMGXMatrix(resources, AMGX.dDDI)
+        device_b = AMGX.AMGXVector(resources, AMGX.dDDI)
+        device_x = AMGX.AMGXVector(resources, AMGX.dDDI)
+        csr_matrix = CuSparseMatrixCSR(transpose(matrix))
+        
+        # # FIXME do on GPU
+        # loop! = zero_index!(Architectures.device(arch), 16, length(csr_matrix.rowPtr))
+        # event = loop!(csr_matrix.rowPtr; dependencies=device_event(arch))
+        @inline subtract_one(x) = convert(Int32, x-1)
```
navidcy, line 150:
```suggestion
        @inline subtract_one(x) = x - oneunit(x)
```
simpler?

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
```diff
@@ -199,6 +165,35 @@ function solve!(η, implicit_free_surface_solver::MGImplicitFreeSurfaceSolver, r
...
+        constructors = deepcopy(implicit_free_surface_solver.matrix_constructors)
+        M = prod(size(η))
+        update_diag!(constructors, arch, M, M, implicit_free_surface_solver.diagonal, Δt, 0)
+        solver.matrix = arch_sparse_matrix(arch, constructors) 
+
+        unsafe_free!(constructors)
+
+        s = solver.amgx_solver
+        solver.amgx_solver.csr_matrix = CuSparseMatrixCSR(transpose(solver.matrix))
+        @inline subtract_one(x) = convert(Int32, x-1)
```
navidcy, line 183:
```suggestion
        @inline subtract_one(x) = x - oneunit(x)
```

validation/implicit_free_surface/implicit_surface_solvers_benchmark.jl
```diff
@@ -1,14 +1,17 @@
 using Oceananigans
 using Oceananigans.Units
 using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBottom
-using Oceananigans.Models.HydrostaticFreeSurfaceModels: FFTImplicitFreeSurfaceSolver, MGImplicitFreeSurfaceSolver
+using Oceananigans.Models.HydrostaticFreeSurfaceModels: FFTImplicitFreeSurfaceSolver, MGImplicitFreeSurfaceSolver, finalize_solver!
 using Printf
+using Oceananigans.Architectures: architecture
```
navidcy, line 6:
```suggestion
```

why we need this? do we?

src/Solvers/multigrid_solver.jl
elise-palethorpe: Yes very good point. My julia kernel programming isn't good enough to know how to do this because we have a global `count` variable and would need to concatenate the `rowval` and `nzval` arrays together at the end - is there a way to do this? I'm happy to just do so on a CPU

validation/implicit_free_surface/implicit_surface_solvers_benchmark.jl
elise-palethorpe: We don't :)

src/Solvers/multigrid_solver.jl
navidcy: Since it's a one-off thing, I think we can do it on the CPU and convert to CuArrays at the end?

src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl
elise-palethorpe: Yes I'll delete it

navidcy: if we just leave `:reltol, sqrt(eps(eltype(grid))))` will it error?

navidcy:
@simone-silvestri you can run the GPU profiler on the `validation/elliptic_solvers/doubly_bounded_poisson.jl` on this branch. You can set `arch = GPU()` and it should work.

src/Solvers/multigrid_solver.jl
navidcy:
```Julia

  oneunit(x::T)
  oneunit(T::Type)


  Returns T(one(x)), where T is either the type of the argument or (if a type is passed) the argument. This differs from one for dimensionful quantities:
  one is dimensionless (a multiplicative identity) while oneunit is dimensionful (of the same type as x, or of type T).

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> oneunit(3.7)
  1.0

  julia> import Dates; oneunit(Dates.Day)
  1 day
```

navidcy: It will reduce the chance of a type instability

navidcy:
I'm not sure if I'm benchmarking correctly on the GPU. I used `@benchmark CUDA.@sync function_to_bench()`. Running the `doubly_bounded_poisson.jl` script with `N = 64` on tartarus gave me:

### On CPU

solvers_benchmark_CPU

### On GPU

solvers_benchmark_GPU

navidcy:
Perhaps for higher N gpu wins? When @elise-palethorpe did some benches on the baroclinic instability problem with `HydrostaticFreeSurfaceModel` the GPU was definitely faster.

src/Solvers/multigrid_solver.jl
elise-palethorpe: @simone-silvestri I'm having some trouble with the copying the arrays to CPU implementation due to the call to the `linear_operator!`. For the laplacian we use the [compute_∇²!](https://github.com/CliMA/Oceananigans.jl/blob/13b7f2366726b3e52303dee67bab264941088554/validation/elliptic_solvers/doubly_bounded_poisson.jl#L37-L50) function which takes `arch` as an argument. This means when we convert arrays to the CPU the kernel call still tries to execute on a GPU (as `arch = GPU()`) which rightly throws an error. To keep things generic we parse the arguments to the kernel call with `args...` so are no in general able to edit the `arch` parameter.

For the `HydrostaticFreeSurfaceModel` we have used your code to hardcode the matricies so `@allowscalar` is not required in this case.

src/Solvers/multigrid_solver.jl
```diff
@@ -60,25 +83,25 @@ Arguments
...
+* `amg_algorithm`: Algebraic Multigrid algorithm defining mapping between different grid spacings.
+                   Options are `RugeStubenAMG()` (default) or `SmoothedAggregationAMG()`.
+                   Note: This keyword is relevant *only* on the CPU.
 """
 function MultigridSolver(linear_operation!::Function,
                          args...;
                          template_field::AbstractField,
                          maxiter = prod(size(template_field)),
-                         reltol = sqrt(eps(eltype(template_field.grid))),
+                         reltol = 0.1*sqrt(eps(eltype(template_field.grid))),
```
navidcy, line 94: remove the 0.1 from here?

src/Solvers/multigrid_solver.jl
```diff
@@ -60,25 +83,25 @@ Arguments
...
                          abstol = 0,
                          amg_algorithm = RugeStubenAMG(),
+                         matrix = nothing,
```
navidcy, line 97: update the docstring to include the `matrix` kwarg

src/Solvers/multigrid_solver.jl
```diff
@@ -60,25 +83,25 @@ Arguments
...
                          )
 
     arch = architecture(template_field)
-
-    arch == GPU() && error("Multigrid solver is only supported on CPUs.")
-
-    matrix = initialize_matrix(template_field, linear_operation!, args...)
+
```
navidcy, line 101:
```suggestion

```

navidcy:
>[AMGX.jl](https://github.com/JuliaGPU/AMGX.jl) only has prebuilt libraries for linux systems. That said, calling using AMGX will fail on, e.g., Mac OS X. What's the best way to optionally load AMGX.jl? Or is there another way around that?

We may be able to optionally call `using AMGX` + all the `AMGX`-related functionality by using:

[`@static` + `islinux()`]( https://docs.julialang.org/en/v1/manual/handling-operating-system-variation/) + [`CUDA.has_cuda_gpu()`](https://cuda.juliagpu.org/stable/api/essentials/#CUDA.has_cuda_gpu)

src/Solvers/multigrid_solver.jl
elise-palethorpe: Sounds good. Should we let the `linear_operator!` default to something if `matrix` is not nothing?

src/Solvers/multigrid_solver.jl
elise-palethorpe: Yes, good catch

src/Solvers/multigrid_solver.jl
navidcy: The way it's written now `linear_operation!` is an obligatory arg.
Hm... so how it's written now one could provide, e.g., a `linear_operation!` that corresponds to the Laplacian and a `matrix` kwarg that corresponds to something else. Then the `linear_operation!` is overwritten, right?

I think it could be confusing. How about something like:

```Julia
function MultigridSolver(linear_operation!::Function,
                         args...;
                         template_field::AbstractField,
                         maxiter = prod(size(template_field)),
                         reltol = sqrt(eps(eltype(template_field.grid))),
                         abstol = 0,
                         amg_algorithm = RugeStubenAMG()
                         )

    arch = architecture(template_field)

    matrix = initialize_matrix(arch, template_field, linear_operation!, args...)

    return MultigridSolver(matrix, args...; template_field, maxiter, reltol, abstol, amg_algorithm)
end

function MultigridSolver(matrix::TypeOfMatrix,
                         args...;
                         template_field::AbstractField,
                         maxiter = prod(size(template_field)),
                         reltol = sqrt(eps(eltype(template_field.grid))),
                         abstol = 0,
                         amg_algorithm = RugeStubenAMG()
                         )

    ...
    
end
```

?

src/Solvers/multigrid_solver.jl
```diff
@@ -87,78 +109,199 @@ function MultigridSolver(linear_operation!::Function,
...
+    
+    s = solver.amgx_solver
+    close(s.device_matrix)
+    close(s.device_x)
+    close(s.device_b)
+    close(s.solver)
+    close(s.resources)
+    close(s.config)
+    AMGX.finalize_plugins()
+    AMGX.finalize()
```
navidcy, line 296:
```suggestion
    AMGX.finalize()

    return nothing
```

src/Solvers/multigrid_solver.jl
navidcy: @simone-silvestri we'll leave the `allowscalar` here... it's only used for the demonstration of the Poisson equation. The implicit free surface solver gives the matrix so it doesn't use the `linear_operation!` to construct the matrix.

@elise-palethorpe let's open an issue after this PR is merged to keep a reminder that this can be further optimized.

navidcy (approved):

src/Oceananigans.jl
```diff
@@ -138,6 +141,12 @@ import Base:
     getindex, lastindex, setindex!,
     push!
 
+const hasamgx = @static Sys.islinux() ? true : false
+
+macro ifhasamgx(expr)
```
navidcy, line 146:
```suggestion
"Boolean denoting whether AMGX.jl can be loaded on machine."
const hasamgx = @static Sys.islinux() ? true : false

"""
    @ifhasamgx expr

Evaluate `expr` only if `hasamgx == true`.
"""
macro ifhasamgx(expr)

```

navidcy: I'll go ahead and commit this so tests start running

src/Solvers/multigrid_solver.jl
simone-silvestri: I would just do something like this
```
function initialize_matrix(::GPU, template_field, linear_operator!, args...)
  cpu_template = #copy field to CPU
  
  A = initialize_matrix(CPU(), cpu_template, linear_operator!, args...) # Here maybe if args... are on the GPU() copy them all to CPU()
  
  return CuSparseMatrixCSC(A)
end
```

navidcy: I don't think that would work because `GPU()` is hiding in `args...`. But I'll copy this in #2737 and we can chat there.

PR #2699, opened 2022-08-17
Change `'\n'` -> `"\n"`

navidcy:
Our style uses double quotes for strings everywhere except `'\n'`. This PR homogenizes the style.

### DIFF
--- a/src/AbstractOperations/conditional_operations.jl
+++ b/src/AbstractOperations/conditional_operations.jl
@@ -136,9 +136,9 @@ compute_at!(c::ConditionalOperation, time) = compute_at!(c.operand, time)
 
 Base.show(io::IO, operation::ConditionalOperation) =
     print(io,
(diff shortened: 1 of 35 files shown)


tomchor (approved):

francispoulin (approved):
Good idea and thanks for doing the work!

glwagner:
> Our style uses double quotes for strings everywhere except `'\n'`. This PR homogenizes the style.

Ah! This was intentional --- note that single quotes denote a character, not a string. So '\n' is the single newline character. It doesn't really matter, but `"\n"` isn't necessary, because we don't need a full-bore string if we only have one character. I thought using single quotes just clarified that it was one character (can get lost when reading, since there is two literal characters when we read it).

navidcy:
We can undo this. Didn't know.

glwagner:
Def not worth undoing!

PR #2703, opened 2022-08-19
Use Julia v1.6.7 for CI

navidcy:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM julia:1.6.6
+FROM julia:7
```
tomchor, line 1: Is this line correct?

tomchor (approved):

Dockerfile
navidcy: Definitely not! Thanks!

francispoulin (approved):

PR #2705, opened 2022-08-21
Increase TagBot lookback

navidcy:
(thanks @DilumAluthge!)


DilumAluthge:
21 won't be enough, since https://github.com/JuliaRegistries/General/pull/65252 was merged more than 21 days ago.

I'd set it to something like 25.

Also, once you get the tag for 0.77.1 created, I'd recommend setting the lookback back to 3. Otherwise, every TagBot run will take a really long time.

navidcy:
Great! Thanks!

PR #2721, opened 2022-08-31
Ignore halos and immersed cells in calculating shallow water advection time-scale

glwagner:
Also specifies that the methods only work for `RectilinearGrid`, and cleans up the formatting.

Some background on #2720 with @pjrusello

### DIFF
--- a/src/Models/ShallowWaterModels/shallow_water_cell_advection_timescale.jl
+++ b/src/Models/ShallowWaterModels/shallow_water_cell_advection_timescale.jl
@@ -3,7 +3,7 @@
 
 import Oceananigans.Utils: cell_advection_timescale
 
-function shallow_water_cell_advection_timescale(uh, vh, h, grid, formulation)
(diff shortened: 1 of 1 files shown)


glwagner:
@pjrusello see if this fixes the issue.

pjrusello:
I’ll be able to test next week, I’m on vacation the next few days. Thanks for the troubleshooting on this.

pjrusello:
I tested this with the simple case from discussion #2720 and with the Bickley Jet example adding a CFL calculation to a status message. It looks like CFL is calculated correctly and the `TimeStepWizard` is functioning properly now.

glwagner:
Good! I wonder if we can add a test here too to wrap this up.

glwagner:
@simone-silvestri @francispoulin I think you two were the last to look at `TimeStepWizard` for the shallow water model. Any thoughts?

tomchor (approved):

navidcy:
Shall we merge this?

PR #2724, opened 2022-09-04
Active particles with tracer interactions

jagoosw:
Hi all, 

I am currently working on a biogeochemistry modelling environment with @johnryantaylor using Oceananigans, and as part of this have have up with a scheme to have "active" particles that interact with tracers.

How I currently have this set up: when the particle dynamics are run, the particles can increase/decrease the concentration of tracers in the cells surrounding them, but this only allows for explicit Euler integration. A better way todo this seems to be to have the particles uptake/exudation of tracers contribute to their tendencies during the time stepping as I have implemented here.

I'm unsure if this is implemented in the best way throughout but I'm fairly satisfied that I have it working as desired. 

Hope everyone is happy with this?

--------

To summaries what I have changed:

- Particle setup: particles can get a parameter called `active_properties` which is a tuple of named tuples of particle properties and tracers (e.g. `((property=:t, tracer=:x),).`) The idea being that the particle dynamics function (as already implemented) changes the particle property to set a rate of uptake/exudation of a tracer, and the below function would integrate this change to the specified tracer.

- `calculate_particle_tendency_contributions!` added after each `calculate_tendencies!` call: function goes through each particle, finds its 8 nearest cells, and adds the relevant fraction of each particle property (divided by the cell volume) to the tracer tendencies so it can be integrated by the time stepper (like tracer forcing)

### DIFF
--- a/src/LagrangianParticleTracking/update_particle_properties.jl
+++ b/src/LagrangianParticleTracking/update_particle_properties.jl
@@ -128,10 +128,16 @@ function update_particle_properties!(lagrangian_particles, model, Δt)
     events = []
 
     for (field_name, tracked_field) in pairs(lagrangian_particles.tracked_fields)
-        compute!(tracked_field)
-        particle_property = getproperty(lagrangian_particles.properties, field_name)
+        if isnothing(tracked_field)
+            tracked_field = fields(model)[field_name]
+         else
+             compute!(tracked_field)
+        end
+        
         LX, LY, LZ = location(tracked_field)
 
+        particle_property = getproperty(lagrangian_particles.properties, field_name)
+    
         update_field_property_kernel! = update_field_property!(device(arch), workgroup, worksize)
 
         update_event = update_field_property_kernel!(particle_property, lagrangian_particles.properties, model.grid,
--- a/src/LagrangianParticleTracking/LagrangianParticleTracking.jl
+++ b/src/LagrangianParticleTracking/LagrangianParticleTracking.jl
@@ -15,6 +15,7 @@ using Oceananigans.ImmersedBoundaries: immersed_cell
 using Oceananigans.Architectures: device, architecture
 using Oceananigans.Fields: interpolate, datatuple, compute!, location, fractional_indices, fractional_y_index
 using Oceananigans.Utils: prettysummary, launch!
+using Oceananigans: fields
 
 import Base: size, length, show
 
@@ -109,5 +110,5 @@ function Base.show(io::IO, lagrangian_particles::LagrangianParticles)
 end
 
 include("update_particle_properties.jl")
-
+include("tracer_interaction_utilities.jl")
 end # module
--- a/src/LagrangianParticleTracking/tracer_interaction_utilities.jl
+++ b/src/LagrangianParticleTracking/tracer_interaction_utilities.jl
@@ -1,0 +1,60 @@
+using KernelAbstractions.Extras.LoopInfo: @unroll
+using Oceananigans.Operators: volume
+using Oceananigans.Grids: AbstractGrid
+
+@inline get_node(::Bounded, i, N) = min(max(i, 1), N)
+@inline get_node(::Periodic, i, N) = ifelse(i < 1, N - i, ifelse(i > N, 1 + i, i))
+
+@inline function get_nearest_nodes(x, y, z, grid, loc)
+    i, j, k = fractional_indices(x, y, z, loc, grid)
+
+    # Convert fractional indices to unit cell coordinates 0 <= (ξ, η, ζ) <=1
+    # and integer indices (with 0-based indexing).
+    ξ, i = modf(i)
+    η, j = modf(j)
+    ζ, k = modf(k)
+
+    if (ξ, η, ζ) == (0, 0, 0) #particle on grid point special case
+        return ((Int(i+1), Int(j+1), Int(k+1), 1), ), 1.0
+    else
+        nodes = repeat([(1, 1, 1, NaN)], 8)
+        _normfactor = 0.0
+        @unroll for n=1:8
+            # Move around cube corners getting node indices (0 or 1) and distances to them
+            # Distance is d when the index is 0, or 1-d when it is 1
+            a = 0^(1+(-1)^n)
+            di = 0^abs(1-a)+ξ*(-1)^a
+
+            b = 0^(1+(-1)^floor(n/2))
+            dj = 0^abs(1-b)+η*(-1)^b
+
+            c = 0^(1+(-1)^floor(n/4))
+            dk = 0^abs(1-c)+ζ*(-1)^c
+
+            nodes[n] = (Int(i+1)+a, 
+                            Int(j+1)+b, 
+                            Int(k+1)+c, 
+                            sqrt(di^2+dj^2+dk^2))
+            _normfactor += 1 ./sqrt(di^2+dj^2+dk^2)
+        end
+        return nodes, 1/_normfactor
+    end
+end
+
+@kernel function calculate_particle_tendency_kernel!(property, tendency, particles, grid::AbstractGrid{FT, TX, TY, TZ}) where {FT, TX, TY, TZ}
+    p = @index(Global)
+    #density = :density in keys(particles.parameters) ? particles.parameters.density : 1
+    density = 1
+
+    LX, LY, LZ = location(tendency)
+    nodes, normfactor = @inbounds get_nearest_nodes(particles.properties.x[p], particles.properties.y[p], particles.properties.z[p], grid, (LX(), LY(), LZ()))
+
+    @unroll for (i, j, k, d) in nodes 
+        # Reflect back on Bounded boundaries or wrap around for Periodic boundaries
+        i, j, k = (get_node(TX(), i, grid.Nx), get_node(TY(), j, grid.Ny), get_node(TZ(), k, grid.Nz))
+
+        node_volume = volume(i, j, k, grid, LX(), LY(), LZ())
+        value = density * @inbounds property[p] * normfactor / (d * node_volume)
+        @inbounds tendency[i, j, k] += value	
+    end
+end
--- a/validation/biogeochemistry/active_particles.jl
+++ b/validation/biogeochemistry/active_particles.jl
@@ -1,0 +1,74 @@
+using Oceananigans, StructArrays, Printf, JLD2, KernelAbstractions
+using Oceananigans.Architectures: device
+using Oceananigans.Fields: TracerFields
+using Oceananigans.LagrangianParticleTracking: calculate_particle_tendency_kernel!
+# This example will not neccesarily keep tracers above zero because two particles may/will take from the same cell in the same timestep
+
+grid = RectilinearGrid(;topology = (Bounded, Bounded, Periodic), size=(10, 10, 1), extent=(1, 1, 1))
+struct Particle{T}
+    x :: T
+    y :: T
+    z :: T
+    A :: T
+    A_sink :: T
+    B_source :: T
+end
+
+# particle randomly walks around and turns tracer a into tracer b
+function dynamics!(particles, model, Δt)
+    particles.properties.x .+= (-1).^rand(Bool, length(particles)).*rand(length(particles))*Δt
+    particles.properties.y .+= (-1).^rand(Bool, length(particles)).*rand(length(particles))*Δt
+
+    #particles go round and turns tracer a into b
+    particles.properties.A_sink .= -particles.properties.A ./ (50 .+ particles.properties.A)
+    particles.properties.B_source .= particles.properties.A ./ (50 .+ particles.properties.A)
+
+    workgroup = min(length(particles), 256)
+    worksize = length(particles)
+
+    arch = model.grid.architecture
+    
+    Gp_kernel! = calculate_particle_tendency_kernel!(device(arch), workgroup, worksize)
+
+    model.auxiliary_fields.Gₚ.A .= 0
+    model.auxiliary_fields.Gₚ.B .= 0
+
+    Gp_event_A = Gp_kernel!(particles.properties.A_sink, model.auxiliary_fields.Gₚ.A, particles, model.grid, dependencies = Event(device(arch)))
+    Gp_event_B = Gp_kernel!(particles.properties.B_source, model.auxiliary_fields.Gₚ.B, particles, model.grid, dependencies = Event(device(arch)))
+
+    events=[Gp_event_A, Gp_event_B]
+
+    wait(device(arch), MultiEvent(Tuple(events)))
+end
+
+P=2
+
+xs = 0.5*ones(P)
+ys = 0.5*ones(P)
+zs = -0.5*ones(P)
+as = zeros(P)
+a_sink_s = zeros(P)
+b_source_s = zeros(P)
+
+particles = StructArray{Particle}((xs, ys, zs, as, a_sink_s, b_source_s))
+
+Gₚ = TracerFields((:A, :B), grid)
+
+lagrangian_particles = LagrangianParticles(particles; tracked_fields=(A=nothing, ), dynamics=dynamics!)
+@info "Initialized Lagrangian particles"
+
+@inline a_sink(i, j, k, clock, grid, c) = c.Gₚ.A[i, j, k]
+@inline b_source(i, j, k, clock, grid, c) = c.Gₚ.B[i, j, k]
+
+A_forcing = Forcing(a_sink, discrete_form=true)
+B_forcing = Forcing(b_source, discrete_form=true)
+
+model = NonhydrostaticModel(grid=grid, tracers=(:A, :B), forcing = (A=A_forcing, B=B_forcing), particles=lagrangian_particles, auxiliary_fields = (; Gₚ))
+
+set!(model, A=1)
+
+@info "Constructed a model"
+
+sim = Simulation(model, Δt=1e-1, stop_time=20)
+
+#run!(sim)


jagoosw:
The test that’s failed says on buildkite that the client was lost and there’s no errors from Julia, I’m guessing if I can get them to rerun it might be fine since the other GPU tasks ran fine?

simone-silvestri:
I can try rerunning it!

jagoosw:
Thank you!

jagoosw:
It failed again so I ran it on our cluster which errored in the same way just saying `killed`. It's failing on my new test and I think I've probably not implemented the `calculate_particle_tendency_contributions!` in a GPU friendly way so will try and resolve this tomorrow.

jagoosw:
I've looked into it more now and the memory usage blows up on line 13 or 14 of calculate_particle_tendency_contributions.jl

To try and more closely copy other similar things I've tried changing 10 and 11 to be
```
property = particles.properties[getindex(field.property, particles.properties)]
tendency = model.timestepper.Gⁿ[getindex(field.tracer, model.timestepper.Gⁿ)]
```

with `@inline getindex(value::Symbol, fields) = findfirst(keys(fields) .== value)` but the same thing happens.

I also checked and the memory usage isn't high on CPU. I don't have much experience with GPUs so am not really sure how to debug this so would really appreciate any advice.

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,75 @@
...
+    worksize = length(particles)
+
+    events = []
+
+    for field in active_properties
+        property = getproperty(particles.properties, field.property)
+        tendency = getproperty(model.timestepper.Gⁿ, field.tracer)
+
+        Gp_kernal! = calculate_particle_tendency_kernel!(device(arch), workgroup, worksize)
+        Gp_event = Gp_kernal!(property, tendency, particles, model.tracers, model.grid)
```
glwagner, line 14:
```suggestion
        Gp_kernel! = calculate_particle_tendency_kernel!(device(arch), workgroup, worksize)
        Gp_event = Gp_kernel!(property, tendency, particles, model.tracers, model.grid)
```

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,75 @@
...
+    return nothing
+end
+
+calculate_particle_tendency_contributions!(model) = calculate_particle_tendency_contributions!(model.particles, model)
+calculate_particle_tendency_contributions!(particles::Nothing, model) = nothing
+calculate_particle_tendency_contributions!(particles, model) = !isnothing(particles.parameters) ? (:active_properties in keys(particles.parameters) ? calculate_particle_tendency_contributions!(particles, model, particles.parameters.active_properties) : nothing) : nothing
+calculate_particle_tendency_contributions!(particles, model, active_fields::Nothing) = nothing
+
+@inline function getmask(x, y, z, grid::AbstractGrid{FT, TX, TY, TZ}, loc) where {FT, TX, TY, TZ}
+    (fi, i::Int), (fj, j::Int), (fk, k::Int) = modf.(fractional_indices(x, y, z, loc, grid))
```
glwagner, line 30: Can you write this in a few lines? I'm not sure what it does. Also do we need the annotations `i::Int`? What does that do?

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,75 @@
...
+
+    nodesi = arch_array(grid.architecture, zeros(Int, 2, 2, 2))
+    nodesj = arch_array(grid.architecture, zeros(Int, 2, 2, 2))
+    nodesk = arch_array(grid.architecture, zeros(Int, 2, 2, 2))
+    d = arch_array(grid.architecture, zeros(2,2,2))
```
glwagner, line 35:
```suggestion
    d = arch_array(grid.architecture, zeros(2, 2, 2))
```

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,75 @@
...
+
+    return nothing
+end
+
+calculate_particle_tendency_contributions!(model) = calculate_particle_tendency_contributions!(model.particles, model)
+calculate_particle_tendency_contributions!(particles::Nothing, model) = nothing
+calculate_particle_tendency_contributions!(particles, model) = !isnothing(particles.parameters) ? (:active_properties in keys(particles.parameters) ? calculate_particle_tendency_contributions!(particles, model, particles.parameters.active_properties) : nothing) : nothing
+calculate_particle_tendency_contributions!(particles, model, active_fields::Nothing) = nothing
+
+@inline function getmask(x, y, z, grid::AbstractGrid{FT, TX, TY, TZ}, loc) where {FT, TX, TY, TZ}
```
glwagner, line 29:
```suggestion
@inline function get_mask(x, y, z, grid::AbstractGrid{FT, TX, TY, TZ}, loc) where {FT, TX, TY, TZ}
```

A more descriptive function name might help code-readers understand this more quickly!

src/LagrangianParticleTracking/update_particle_properties.jl
```diff
@@ -128,10 +128,24 @@ function update_particle_properties!(lagrangian_particles, model, Δt)
...
+            elseif field_name in keys(model.auxiliary_fields)
+                tracked_field = getproperty(model.auxiliary_fields, field_name)
+            elseif field_name in (:u, :v, :w)
+                tracked_field = getproperty(model.velocities, field_name)
+            else
+                throw(ArgumentError("$field_name is not a model field that can be tracked"))
+            end
+        else
+            compute!(tracked_field)
+        end
```
glwagner, line 143: I'm not sure I entirely follow this. But would it be easier to do something like:

```julia
tracked_field = fields(model)[field_name]
```

?

test/test_lagrangian_particle_tracking.jl
```diff
@@ -12,6 +12,7 @@ struct TestParticle{T}
     v :: T
     w :: T
     s :: T
+    t :: T
```
glwagner, line 15: `t` is easily confused with "time", can we use a different letter?

test/test_lagrangian_particle_tracking.jl
```diff
@@ -105,45 +106,45 @@ function run_simple_particle_tracking_tests(arch, timestepper; vertically_stretc
...
-    tracked_fields = merge(velocities, (; s=speed))
+    tracked_fields = merge(velocities, (; s=speed, t=nothing))
 
     # Test second constructor
     lagrangian_particles = LagrangianParticles(particles; tracked_fields)
     @test lagrangian_particles isa LagrangianParticles
 
     model = NonhydrostaticModel(grid=grid, timestepper=timestepper,
-                                velocities=velocities, particles=lagrangian_particles)
+                                velocities=velocities, particles=lagrangian_particles, tracers=(:t, ))
```
glwagner, line 124: This is valid and preferred:

```suggestion
                                velocities=velocities, particles=lagrangian_particles, tracers=:t)
```

glwagner, line 124: This is valid and preferred:

```julia
tracers=:t
```

Altogether:

```julia
model = NonhydrostaticModel(; grid, timestepper, velocities, particles=lagrangian_particles, tracers=:t)
```

test/test_lagrangian_particle_tracking.jl
```diff
@@ -255,6 +263,44 @@ function run_simple_particle_tracking_tests(arch, timestepper; vertically_stretc
...
+    u, v, w = velocities
+    speed = Field(√(u*u + v*v + w*w))
+    tracked_fields = merge(velocities, (; s=speed))
+    active_properties = ((property=:t, tracer=:t), )
+
+    lagrangian_particles = LagrangianParticles(particles; tracked_fields, parameters=(active_properties=active_properties, ))
+    @test lagrangian_particles isa LagrangianParticles
+
+    model = NonhydrostaticModel(grid=grid, timestepper=timestepper,
+                                velocities=velocities, particles=lagrangian_particles, tracers=(:t, ))
```
glwagner, line 280:
```suggestion
    model = NonhydrostaticModel(; grid, timestepper, velocities, particles=lagrangian_particles, tracers=:t)
```

test/test_lagrangian_particle_tracking.jl
```diff
@@ -255,6 +263,44 @@ function run_simple_particle_tracking_tests(arch, timestepper; vertically_stretc
...
+    ts = arch_array(arch, ones(P))
+    particles = StructArray{TestParticle}((xs, ys, zs, us, vs, ws, ss, ts))
+
+    velocities = VelocityFields(grid)
+    u, v, w = velocities
+    speed = Field(√(u*u + v*v + w*w))
+    tracked_fields = merge(velocities, (; s=speed))
+    active_properties = ((property=:t, tracer=:t), )
+
+    lagrangian_particles = LagrangianParticles(particles; tracked_fields, parameters=(active_properties=active_properties, ))
```
glwagner, line 276:
```suggestion
    lagrangian_particles = LagrangianParticles(particles; tracked_fields, parameters=(; active_properties))
```

test/test_lagrangian_particle_tracking.jl
```diff
@@ -255,6 +263,44 @@ function run_simple_particle_tracking_tests(arch, timestepper; vertically_stretc
...
+                                velocities=velocities, particles=lagrangian_particles, tracers=(:t, ))
+
+    set!(model, u=1, v=1)
+
+    sim = Simulation(model, Δt=1e-2, stop_iteration=1)
+
+    jld2_filepath = "test_particles.jld2"
+    sim.output_writers[:particles_jld2] =
+        JLD2OutputWriter(model, (; particles=model.particles),
+                        filename=jld2_filepath, schedule=IterationInterval(1))
```
glwagner, line 289:
```suggestion
        JLD2OutputWriter(model, (; particles=model.particles),
                         filename=jld2_filepath, schedule=IterationInterval(1))
```

test/test_lagrangian_particle_tracking.jl
```diff
@@ -255,6 +263,44 @@ function run_simple_particle_tracking_tests(arch, timestepper; vertically_stretc
...
+
+    nc_filepath = "test_particles.nc"
+    sim.output_writers[:particles_nc] =
+        NetCDFOutputWriter(model, model.particles, filename=nc_filepath, schedule=IterationInterval(1))
+
+    sim.output_writers[:checkpointer] = Checkpointer(model, schedule=IterationInterval(1),
+                                                    dir = ".", prefix = "particles_checkpoint")
```
glwagner, line 296:
```suggestion
    sim.output_writers[:checkpointer] = Checkpointer(model, schedule=IterationInterval(1),
                                                     dir = ".", prefix = "particles_checkpoint")
```

test/test_lagrangian_particle_tracking.jl
```diff
@@ -255,6 +263,44 @@ function run_simple_particle_tracking_tests(arch, timestepper; vertically_stretc
...
+
+    model = sim.model
+    run!(sim)
+
+    ni, nj, nk, d = Oceananigans.LagrangianParticleTracking.getmask(0.0, 0.0, 0.5, grid, (Center(), Center(), Center()))
+    vols = [Oceananigans.Operators.volume(i, nj[idx], nk[idx], grid, Center(), Center(), Center()) for (idx, i) in enumerate(ni)]
+    @test sum(model.tracers.t) ≈ (sum(1e-2*P ./vols) / 8)
```
glwagner, line 303: What's being tested here? I don't understand this, and the `1e-2` and `8` need to be explained.

glwagner:
> Particle setup: particles can get a parameter called active_properties which is a tuple of named tuples of particle properties and tracers (e.g. ((property=:t, tracer=:x),).) The idea being that the particle dynamics function (as already implemented) changes the particle property to set a rate of uptake/exudation of a tracer, and the below function would integrate this change to the specified tracer.

Can you please restate this? I think a code example would be helpful with an explanation on what's being accomplished. Also can you explain why this can't be implemented with a Callback? I.e., why do we desire this as a source code feature?

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
jagoosw: This gets the x, y, z position of the particle in terms of its position on the grid in i, j, k as a tuple `(i, j, k)` `modf` then splits each of i, j, k into its integer and fractional parts i.e. `i` and `fi`. Because the integer part is used to index an array later if I don't specify it as `Int` it doesn't work because `modf` returns them both as floats.

I'm not sure how best to clarify this but could expand it to:

```suggestion
    pos = fractional_indices(x, y, z, loc, grid)
    (fi, i::Int), (fj, j::Int), (fk, k::Int) = modf.(pos)
```

Although I'm not sure this actually clarifies very much

src/LagrangianParticleTracking/update_particle_properties.jl
jagoosw: Oh yeah that works great! That's exactly what I was looking for here.

```suggestion
        if isnothing(tracked_field)
           tracked_field = fields(model)[field_name]
        else
            compute!(tracked_field)
        end
```

test/test_lagrangian_particle_tracking.jl
jagoosw: Will change

test/test_lagrangian_particle_tracking.jl
jagoosw: Good to know, thank you!

test/test_lagrangian_particle_tracking.jl
jagoosw: This test checks that the correct total change is made to the tracer concentration, and each particle is adding 1 unit/s for `1e-2`s spreading spread over the volume of the 8 nearest cells.

I'll add another test I had in my other implementation that checks that the mask function is selecting the correct cells too.

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,75 @@
...
+
+@inline getnode(::Bounded, i, N) = min(max(i, 1), N)
+@inline getnode(::Periodic, i, N) = i < 1 ? N - i : (i > N ? 1 + i : i)
+
+@kernel function calculate_particle_tendency_kernel!(property, tendency, particles, model, grid)
+    p = @index(Global)
+    density = :density in keys(particles.parameters) ? particles.parameters.density : 1
+    
+    LX, LY, LZ = location(tendency)
+    nodesi, nodesj, nodesk, d = @inbounds getmask(particles.properties.x[p], particles.properties.y[p], particles.properties.z[p], grid, (LX(), LY(), LZ()))
```
jagoosw, line 64:
```suggestion
nodesi, nodesj, nodesk, d = @inbounds get_mask(particles.properties.x[p], particles.properties.y[p], particles.properties.z[p], grid, (LX(), LY(), LZ()))
```

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,75 @@
...
+                nodesk[a, b, c] = getnode(TZ(), k+c, grid.Nz)
+            end
+        end
+        
+        normfactor=1/sum(1 ./ d)
+        d ./= normfactor
+    end
+
+    return nodesi, nodesj, nodesk, d
+end
```
jagoosw, line 54: It may reduce the clarity of what this function does too much but I rewrote this function and its about 10x faster:

```suggestion
@inline function get_mask(x, y, z, grid, loc)
    (fi, i::Int), (fj, j::Int), (fk, k::Int) = modf.(fractional_indices(x, y, z, loc, grid))

    if (fi, fj, fk) == (0, 0, 0) #particle on grid point special case
        return ((i, j, k, 1), ), 1.0
    else
        nodes = repeat([(1, 1, 1, NaN)], 8)
        _normfactor = 0.0
        @unroll for n=1:8
            a, b, c = 0^(1+(-1)^n), 0^(1+(-1)^floor(n/2)), 0^(1+(-1)^floor(n/4))
            di, dj, dk = 0^abs(1-a)+fi*(-1)^a, 0^abs(1-b)+fj*(-1)^b, 0^abs(1-c)+fk*(-1)^c #when (e.g. a=0 this makes di, the x distance from the particle to the grid point di, and when a=1 it is 1-fi
            nodes[n] = (i+1+a, 
                            j+1+b, 
                            k+1+c, 
                            sqrt(di^2+dj^2+dk^2))
                            _normfactor += 1 ./sqrt(di^2+dj^2+dk^2)
        end
        return nodes, 1/_normfactor
    end
end
```

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,75 @@
...
+        
+    for (idx, i) in enumerate(nodesi) #tried a vector version of this and was much slower
+        @inbounds begin
+            j = nodesj[idx]; k = nodesk[idx]
+            vol = volume(i, j, k, grid, LX(), LY(), LZ())
+            value = density * property[p] / (d[idx] * vol)
+
+            tendency[i, j, k] += value
+        end
+    end
```
jagoosw, line 74: Corresponding to changes [here](https://github.com/CliMA/Oceananigans.jl/pull/2724/files#r964376612)

```suggestion
    nodes, normfactor = @inbounds get_mask(particles.properties.x[p], particles.properties.y[p], particles.properties.z[p], grid, (LX(), LY(), LZ()))
        
    @unroll for (i, j, k, d) in nodes #tried a vector version of this and was much slower
        i, j, k = (getnode(TX(), i, grid.Nx), getnode(TY(), j, grid.Ny), getnode(TZ(), k, grid.Nz))
        vol = volume(i, j, k, grid, LX(), LY(), LZ())
        value = density * @inbounds property[p] * normfactor/ (d * vol)
        @inbounds tendency[i, j, k] += value
    end
```

test/test_lagrangian_particle_tracking.jl
```diff
@@ -255,6 +263,43 @@ function run_simple_particle_tracking_tests(arch, timestepper; vertically_stretc
...
+        NetCDFOutputWriter(model, model.particles, filename=nc_filepath, schedule=IterationInterval(1))
+
+    sim.output_writers[:checkpointer] = Checkpointer(model, schedule=IterationInterval(1),
+                                                     dir = ".", prefix = "particles_checkpoint")
+
+    model = sim.model
+    run!(sim)
+
+    ni, nj, nk, d = Oceananigans.LagrangianParticleTracking.getmask(0.0, 0.0, 0.5, grid, (Center(), Center(), Center()))
+    vols = [Oceananigans.Operators.volume(i, nj[idx], nk[idx], grid, Center(), Center(), Center()) for (idx, i) in enumerate(ni)]
```
jagoosw, line 301: Corresponds to changes [here](https://github.com/CliMA/Oceananigans.jl/pull/2724/files#r964377399)

```suggestion
    nodes, normfactor = getmask(0.0, 0.0, 0.5, grid, (Center(), Center(), Center()))
    vols = [Oceananigans.Operators.volume(getnode(topo[1](), i, grid.Nx), getnode(topo[2](), j, grid.Ny), getnode(topo[3](), k, grid.Nz), grid, Center(), Center(), Center()) for (i, j, k, d) in nodes]
```

jagoosw:
> Can you please restate this? I think a code example would be helpful with an explanation on what's being accomplished. Also can you explain why this can't be implemented with a Callback (or forcing function)? I.e., why do we desire this as a source code feature? Just to be clear, I'm not saying it should be one way or another, but it's good to have explicit justification for source code features (which are expensive to maintain and require resources to test)!

No problem at all, [heres](https://gist.github.com/jagoosw/213d7f2ce3707768d4ddb07f8827cdd2) an example (not sure where best to put this/if you'd want it in the examples folder). This is about the simplest version I could think of where we have a particle randomly walking around converting tracer a into tracer b. The specific use case we've been using this for is modelling kelp fronds as particles which grow (variety of particle properties change) depending on how much nutrients the particles uptake from a biogeochemical model, and also release tracer back into the model.

test

Although I've had this working as custom dynamics of the particles the effect of the particles tendency can only be integrated with explicit Euler, i.e. at each substep it just does `tracer[i, j, k] += value*Δt`, where as this solution allows it to be properly integrated along with the other tendencies. I couldn't think of a way to implement this as a forcing function before, but perhaps I could use a callback to update an auxiliary field with the tendencies of the particles, and then add this as a forcing function, although I imagine that could use a lot more memory if there were a large grid and small amount of particles.

src/LagrangianParticleTracking/update_particle_properties.jl
glwagner: Why is there a branch for `nothing` `tracked_field`? I don't totally understand what's being supported here.

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
glwagner: Is `pos` short for "position"?

glwagner: > This gets the x, y, z position of the particle in terms of its position on the grid in i, j, k as a tuple (i, j, k) modf then splits each of i, j, k into its integer and fractional parts i.e. i and fi.

I'm a little hazy on particle advection, but I think we need this kind of algorithm also to advected particles (ie, to interpolate gridded velocities to the particle position). Can we reuse the code for that?

test/test_lagrangian_particle_tracking.jl
glwagner: Something just like

```julia
Ncells  = 8
tolerance = 1e-2
total_volume = [Oceananigans.Operators.volume(i, nj[idx], nk[idx], grid, Center(), Center(), Center()) for (idx, i) in enumerate(ni)]
average_P = sum(P ./ (Ncells * total_volume))
@test sum(model.tracers.t) ≈ tolerance * average_P
```

could help. But I think I got the names wrong above, and I'm not sure why we compare `t` to `P`.

test/test_lagrangian_particle_tracking.jl
glwagner: Same general comments as above; writing this on more lines will help with readability

glwagner:
> > Can you please restate this? I think a code example would be helpful with an explanation on what's being accomplished. Also can you explain why this can't be implemented with a Callback (or forcing function)? I.e., why do we desire this as a source code feature? Just to be clear, I'm not saying it should be one way or another, but it's good to have explicit justification for source code features (which are expensive to maintain and require resources to test)!
> 
> No problem at all, [heres](https://gist.github.com/jagoosw/213d7f2ce3707768d4ddb07f8827cdd2) an example (not sure where best to put this/if you'd want it in the examples folder). This is about the simplest version I could think of where we have a particle randomly walking around converting tracer a into tracer b. The specific use case we've been using this for is modelling kelp fronds as particles which grow (variety of particle properties change) depending on how much nutrients the particles uptake from a biogeochemical model, and also release tracer back into the model. I also envisage using this to exert drag on the flow at some point.
> 
> test [ test ](https://user-images.githubusercontent.com/26657828/188803448-86bc2fe2-dbbc-447a-93c6-ff238f6dde06.gif) [ ](https://user-images.githubusercontent.com/26657828/188803448-86bc2fe2-dbbc-447a-93c6-ff238f6dde06.gif)
> 
> Although I've had this working as custom dynamics of the particles the effect of the particles tendency can only be integrated with explicit Euler, i.e. at each substep it just does `tracer[i, j, k] += value*Δt`, where as this solution allows it to be properly integrated along with the other tendencies. I couldn't think of a way to implement this as a forcing function before, but perhaps I could use a callback to update an auxiliary field with the tendencies of the particles, and then add this as a forcing function, although I imagine that could use a lot more memory if there were a large grid and small amount of particles.
> 
> (I fix the tests that's failed until we've discussed the other changes since its only failed because I forgot to change `getmask` to `get_mask` in the test script)

Nice! Can you paste a more minimal example directly in this PR too so it's preserved here for the future?

glwagner:
> (not sure where best to put this/if you'd want it in the examples folder)

The right place would be `validation/`, as we only use `examples/` for features that are widely-used and well-tested (examples are expensive, because they have to run during CI).

test/test_lagrangian_particle_tracking.jl
jagoosw: Okay I'll write it out more explicitly like that to add some clarity. Here `P` is the number of particles and `1e-2`is the amount each particle should have reduced the amount of tracer by.

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
jagoosw: Yeah sorry `pos` was position. Thats a good idea, the advection interpolates the velocity fields and does have todo the same thing:
https://github.com/CliMA/Oceananigans.jl/blob/af21542ba01afe92e7fc4de6251dec55ed5182c7/src/Fields/interpolate.jl#L122-L129

I will copy the notation they use.

jagoosw:
Is there a way for me to cancel tests so they don't have to run every commit?

src/LagrangianParticleTracking/update_particle_properties.jl
jagoosw: There's this issue with tracked fields where (we don't think) you can specify a tracer as a tracked field without a weird work around because the field isn't defined until after the model is defined, but the model can't be defined until the particles have been defined, and this is a solution that directly references the tracer field in the model instead. I wrote perhaps a better explanation of the problem [here](https://github.com/CliMA/Oceananigans.jl/pull/2662#issue-1304970333).

This isn't a particularly nice solution but I'm not sure how else to implement it since `tracked_fields` is a named tuple, maybe instead of `nothing` I could define an empty function such as `Tracer()` (like grids does with `Perodic()` etc.), then do something like:
```
function compute!(comp::Tracer(), field_name, model)
    comp = fields(model)[field_name]
end
compute!(comp::ComputedField, field_name, model) = compute!(comp)
```
Then in place of the above just have `compute!(tracked_field, field_name, model)`

glwagner:
> Is there a way for me to cancel tests so they don't have to run every commit?

You can include `[skip ci]` in the commit message: https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

jagoosw:
I've spent some time trying to diagnose the GPU issue and it's quite strange, when I copy the test line by line into a REPL it doesn't cause the memory issue and runs fine. After doing that (so everything has been compiled etc.) I then ran the test and the memory usage is fine until the last test, then crashes again. I will investigate more...

jagoosw:
So I think that the tests generally are very close to the memory limit (which I think is around 1GB, not sure how that's being set). It looks like the tracer sinking isn't actually using much more memory but its enough to push it over. If I remove the output writing and speed field tracking (which don't get tested in this run anyway) from the final test it seems to keep the memory usage lower.

I can't test it myself on a GPU right now so am going to push and hopefully it'll work!

jagoosw:
> > So I think that the tests generally are very close to the memory limit (which I think is around 1GB, not sure how that's being set). It looks like the tracer sinking isn't actually using much more memory but its enough to push it over. If I remove the output writing and speed field tracking (which don't get tested in this run anyway) from the final test it seems to keep the memory usage lower.
> > I can't test it myself on a GPU right now so am going to push and hopefully it'll work!
> 
> Interesting. I don't believe we make any attempt to manage GPU memory. However, the tests are all quite lightweight (the largest are probably the regression tests, at 16^3?) The GPU we use for CI has 24 GB total memory. Up to 16 jobs can run simultaneously. I'm not sure this is consistent with a test being just over the limit, because this would mean the tests would intermittently pass, right? It's slightly stochastic how many CI jobs are running simultaneously.

I'm not really sure where this limit is coming from because I get the same behaviour on our GPU which is on a node with 40GB of memory, and it consistently fails when the memory usage of the process gets to fractionally over 1GB. I think the spike in memory usage is occurring on the line `Gp_kernel! = calculate_particle_tendency_kernel!(device(arch), workgroup, worksize)` which I wouldn't have thought would allocate much memory?

glwagner:
Why does the test allocate so much memory? Is it possible to design an inexpensive test? It's important that CI is absolutely as cheap as possible.

src/TimeSteppers/runge_kutta_3.jl
```diff
@@ -100,6 +100,7 @@ function time_step!(model::AbstractModel{<:RungeKutta3TimeStepper}, Δt)
     #
 
     calculate_tendencies!(model)
+    calculate_particle_tendency_contributions!(model)
```
glwagner, line 103: Why add these here rather than to `calculate_tendencies!`?

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,84 @@
...
+
+    wait(device(arch), MultiEvent(Tuple(events)))
+
+    return nothing
+end
+
+calculate_particle_tendency_contributions!(model) = calculate_particle_tendency_contributions!(model.particles, model)
+calculate_particle_tendency_contributions!(particles::Nothing, model) = nothing
+calculate_particle_tendency_contributions!(particles, model) = !isnothing(particles.parameters) ? (:active_properties in keys(particles.parameters) ? calculate_particle_tendency_contributions!(particles, model, particles.parameters.active_properties) : nothing) : nothing
```
glwagner, line 27: Trying to understand this: your proposed API asks users to add a property `active_properties` to the particle parameters? If they don't add it, then due to this `if`-statement, nothing happens?

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,84 @@
...
+calculate_particle_tendency_contributions!(particles, model, active_fields::Nothing) = nothing
+
+@inline function get_mask(x, y, z, grid, loc)
```
glwagner, line 30: What's the purpose of this function? Maybe add a docstring?

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,84 @@
...
+                            Int(k+1)+c, 
+                            sqrt(di^2+dj^2+dk^2))
+            _normfactor += 1 ./sqrt(di^2+dj^2+dk^2)
+        end
+        return nodes, 1/_normfactor
+    end
+end
+
+@inline get_node(::Bounded, i, N) = min(max(i, 1), N)
+@inline get_node(::Periodic, i, N) = i < 1 ? N - i : (i > N ? 1 + i : i)
```
glwagner, line 67: You may need to use `ifelse` here; otherwise it likely will not be performant (especially on GPU).

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
```diff
@@ -0,0 +1,84 @@
...
+    LX, LY, LZ = location(tendency)
+    nodes, normfactor = @inbounds get_mask(particles.properties.x[p], particles.properties.y[p], particles.properties.z[p], grid, (LX(), LY(), LZ()))
+        
+    @unroll for (i, j, k, d) in nodes 
+        # Reflect back on Bounded boundaries or wrap around for Periodic boundaries
+        i, j, k = (get_node(TX(), i, grid.Nx), get_node(TY(), j, grid.Ny), get_node(TZ(), k, grid.Nz))
+
+        node_volume = volume(i, j, k, grid, LX(), LY(), LZ())
+        value = density * @inbounds property[p] * normfactor / (d * node_volume)
+        @inbounds tendency[i, j, k] += value
```
glwagner, line 82: Does this mean when that this feature is only able to add or subtract to a tracer tendency?

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
glwagner: Would another solution add a new property to `LagrangianParticles`? This would allow us to use dispatch at a higher level rather than these nested if statements.

glwagner:
> I couldn't think of a way to implement this as a forcing function before, but perhaps I could use a callback to update an auxiliary field with the tendencies of the particles, and then add this as a forcing function, although I imagine that could use a lot more memory if there were a large grid and small amount of particles.

I'm still wondering about this solution, which may both be _more general_ (since users can effect arbitrary changes to tracer tendencies, not just source/sink terms) and is less heavy in the source code. Right now, this PR makes some significant source code changes for what I think is a rather narrow application, which is disproportional compared to most of the other features we have I think.

> although I imagine that could use a lot more memory if there were a large grid and small amount of particles.

It seems the memory requirement of the forcing function approach is proportional to the number of _tracers_, not the number of particles --- right? We would nee an "auxiliary tendency" for every tracer; the memory requirement doesn't change based on the number of particles.

If for some reason we cannot use a forcing function approach, we'll have to think carefully about how to design a flexible user API for "particle-induced tracer forcing". I think if we are going to add this then it should be more general than just a source/sink calculation --- ie, we should attempt to have a small source code addition that makes a wide range of new behavior possible, if we can.

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
jagoosw: Adding a new property probably would be the most straight forward

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
jagoosw: It works out which grid points are the nearest to the particle (since particles can be at any arbitrary position rather than on grid points), and returns a tuple of tuples of (i, j, k, d) where d is the distance from the particle to point ijk, along with a normalization factor which is 1/sum(1/d), so the contribution you add to ijk is tendency*d_ijk*normalization factor.

I'll add a docstring

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
jagoosw: Okay, will look into this

jagoosw:
> I'm still wondering about this solution, which may both be _more general_ (since users can effect arbitrary changes to tracer tendencies, not just source/sink terms) and is less heavy in the source code. Right now, this PR makes some significant source code changes for what I think is a rather narrow application, which is disproportional compared to most of the other features we have I think.

> If for some reason we cannot use a forcing function approach, we'll have to think carefully about how to design a flexible user API for "particle-induced tracer forcing". I think if we are going to add this then it should be more general than just a source/sink calculation --- ie, we should attempt to have a small source code addition that makes a wide range of new behavior possible, if we can.

I've reconsidered this now and think we may not even need any source code change to properly implement this behaviour and integrate some particle induced tracer forcing.

If we have a model with some tracers `(:A, :B)` which we want to force with some particles, we can define an auxiliary field `G\_p=TracerFields((:A, :B), grid)`. We can then have a particle dynamics function that modifies `model.auxiliary_fields.G\_p`, for example sets the points surrounding the particles to `-A[i, j, k]` for `A`, and `+A[i, j, k]` for `B` like in my example above. You can then define a discrete forcing function for each of the tracers like `a_forcing(i, j, k, grid, clock, model_fields) = model_fields.G\_p.A[i, j, k]` etc. (this would only work with my other PR #2733 added).

> It seems the memory requirement of the forcing function approach is proportional to the number of tracers, not the number of particles --- right? We would nee an "auxiliary tendency" for every tracer; the memory requirement doesn't change based on the number of particles.

My reservation with this was that if there's e.g. 1 particle effecting loads of tracers you have to store lots more fields, but I see now that this is much less of an issue.

Sorry about making this much more complicated than it should have been!

jagoosw:
I think we do need the change I made to the tracked fields so that tracers and auxiliary fields can be tracked properly although I don't like the `(field_name=nothing, )`setup as I currently have it. Would it be okay for me to change it to a tuple of either model field names, or computed fields, e.g. `(:A, :u, speed)` where `speed = Field(√(u*u + v*v + w*w))`? And should I make a new PR with this change in it?

src/LagrangianParticleTracking/calculate_particle_tendency_contributions.jl
glwagner: how about calling it `nearest_cell_center` ?

glwagner:
No worries @jagoosw ! This is a neat application!

It seems maybe that the function `get_mask` (or `nearest_cell_center`? If user facing we should come up with a good name) may nevertheless be a crucial part of any implementation, right?

I suspect we still may want some source code features to support this application, but we may not need to add a new routine to modify the tendencies?

jagoosw:
Thanks!
> It seems maybe that the function `get_mask` (or `nearest_cell_center`? If user facing we should come up with a good name) may nevertheless be a crucial part of any implementation, right?

> I suspect we still may want some source code features to support this application, but we may not need to add a new routine to modify the tendencies?

I suppose it might be useful to provide some functions like that to make it easier to implement. Possibly also the kernel function I put in the example gist and a wrapper like `force_nearest(particle_properties, tendency_field, particles, grid)` so there's an easy interface for people to add it to their particle dynamics function?

glwagner:
Can we start by refactoring this PR to implement this functionality in an example script? It could perhaps go in [`validation/biogeochemistry`](https://github.com/CliMA/Oceananigans.jl/tree/main/validation/biogeochemistry).

Once we have that cleaned up, we might be able to see what parts of it belong in the source code versus left to users?

jagoosw:
Sure, I'll sort that out later

jagoosw:
Sorry this took so long, I've put the example together and moved some functions to a different file in the source code. Do you think these are the appropriate functions to keep since they can fairly generally be used to make an infinitesimal Lagrangian particle change the tracers?

There is a small issue with how I've done it that you still need to externally define the forcing functions, auxiliary tendency fields, and reset the tendency field to zero in the particle dynamic function. This will be solved with the #2772 though.

Issue #2730, opened 2022-09-09
Auxiliary fields not directly available to tracer forcing functions


jagoosw:
Hi all,

I think it would be useful for auxiliary fields to be available to tracer forcing functions in the same way that other tracers and the velocity fields are. I'm currently working round this by putting auxiliary fields in the parameters but this is quite cumbersome.

It would be straightforward to implement by changing the `fields` functions e.g.:

`fields(model::NonhydrostaticModel) = merge(model.velocities, model.tracers, model.auxiliary_fields)`

I'm happy to try and do this later if there's no reason not to?

jagoosw:
Just looked and this is actually implemented in `HydrostaticFreeSurfaceModels` just not in the other too. I'll change them and modify the multiple tracer test to use an auxiliary field.

simone-silvestri:
sounds good!

PR #2733, opened 2022-09-11
Add direct access to auxiliary fields for forcing functions

jagoosw:
To implement #2730

### DIFF
--- a/docs/src/model_setup/forcing_functions.md
+++ b/docs/src/model_setup/forcing_functions.md
@@ -97,7 +97,7 @@ of GPU-compiliability.
 
 ### Forcing functions that depend on model fields
 
-Forcing functions may depend on model fields evaluated at the `x, y, z` where forcing is applied.
+Forcing functions may depend on model fields (velocity, tracers or auxiliary fields) evaluated at the `x, y, z` where forcing is applied.
 Here's a somewhat non-sensical example:
 
 ```jldoctest field_dependent_forcing
--- a/src/Models/NonhydrostaticModels/calculate_nonhydrostatic_tendencies.jl
+++ b/src/Models/NonhydrostaticModels/calculate_nonhydrostatic_tendencies.jl
@@ -49,7 +49,7 @@ function calculate_interior_tendency_contributions!(model)
     background_fields    = model.background_fields
     velocities           = model.velocities
     tracers              = model.tracers
-    auxiliary_fields = model.auxiliary_fields
+    auxiliary_fields     = model.auxiliary_fields
     hydrostatic_pressure = model.pressures.pHY′
     diffusivities        = model.diffusivity_fields
     forcings             = model.forcing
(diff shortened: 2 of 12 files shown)


src/Models/NonhydrostaticModels/calculate_nonhydrostatic_tendencies.jl
```diff
@@ -49,6 +49,7 @@ function calculate_interior_tendency_contributions!(model)
     background_fields    = model.background_fields
     velocities           = model.velocities
     tracers              = model.tracers
+    auxiliary_fields = model.auxiliary_fields
```
navidcy, line 52:
```suggestion
    auxiliary_fields     = model.auxiliary_fields
```

docs/src/model_setup/forcing_functions.md
```diff
@@ -97,7 +97,7 @@ of GPU-compiliability.
 
 ### Forcing functions that depend on model fields
 
-Forcing functions may depend on model fields evaluated at the `x, y, z` where forcing is applied.
+Forcing functions may depend on model fields (velocity, tracers or auxiliary fields) evaluated at the `x, y, z` where forcing is applied.
```
navidcy, line 100: That's a nice addition. Does `ShallowWaterModel` accept auxiliary fields? Could we add that functionality there? Should we? cc @francispoulin

jagoosw: Thanks. I've never used `ShallowWaterModels` myself so wanted to have a play with them before I tried, but it looks like it should be relatively straight forward like with the non hydrostatic model.

glwagner (approved):
I think this is great, and clean implementation!

One note is that this should probably be done uniformly across models as @navidcy says; basically nearly identical changes for `HydrostaticFreeSurfaceModel` in the constructor and definition of `fields`.

Possibly, the docs for `Forcing` should be updated as well.

Note that we have some explaining. Now `fields` is not _exactly_ prognostic or auxiliary fields, because some auxiliary fields are left out --- most notably, pressures and diffusivities. I think it's ok, we just need to say "field dependencies can be velocities, tracers, auxiliary fields, and free surface displacement".

docs/src/model_setup/forcing_functions.md
jagoosw: I've just started looking into implementing this and it looks like `auxiliary_fields` aren't implemented at all in `ShallowWaterModels`

jagoosw:
Not sure why this test failed since I only changed some formatting

navidcy:
Sometimes tests fail to load. Merely restarting them does the job. Did that. Let’s see.

jagoosw:
Thank you

navidcy:
@jagoosw I update the docstrings to mention auxiliary fields. I'll merge when tests pass. Thanks!

jagoosw:
@navidcy fantastic, thank you!

PR #2734, opened 2022-09-12
Correct computation for buoyancy dissipation rate, `χ_diff` + demonstrate `Integral`

### LINKED ISSUE #2735: Volume integral of buoyancy variance changed between v0.71.6 and v0.72.4
I just noticed that the Nusselt number calculation result in the horizontal convection example changed between v0.71.6 and v0.72.4. It almost doubled from around 10 at

https://clima.github.io/OceananigansDocumentation/v0.71.6/generated/horizontal_convection/#Load-saved-output,-process,-visualize

to around 20 in 

https://clima.github.io/OceananigansDocumentation/v0.72.4/generated/horizontal_convection/#Load-saved-output,-process,-visualize

The calculation is done via

```Julia
χ_diff = κ * b★^2 * π * tanh(2π * H / Lx)

for i = 1:length(t)
    b = b_timeseries[i]
    sum!(∫ⱽ_mod²_∇b, (∂x(b)^2 + ∂z(b)^2) * volume)
    Nu[i] = (κ *  ∫ⱽ_mod²_∇b[1, 1, 1]) / χ_diff
end
```

Did anything changed between these two version in the way `sum!` or `volume` works?

Here's the diff between v0.71.6 and v0.72.4: https://github.com/CliMA/Oceananigans.jl/compare/v0.71.6...v0.72.4

navidcy:
In the horizontal convection example `χ_diff` is defined as a volume average of the buoyancy dissipation but when computing it were not dividing by the total volume. Now fixed. To compute the buoyancy dissipation the halos were crucial so as to take into account correct boundary conditions for `b` -- the example now remarks on this and saves output using `with_halos = true`.

Closes #2735.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -158,6 +158,12 @@ makedocs(bib,
  checkdocs = :exports
 )
 
+@info "Cleaning up temporary .jld2 and .nc files created by doctests..."
+
+for file in vcat(glob("docs/*.jld2"), glob("docs/*.nc"))
+    rm(file)
+end
+
 deploydocs(
           repo = "github.com/CliMA/OceananigansDocumentation.git",
       versions = ["stable" => "v^", "v#.#.#", "dev" => "dev"],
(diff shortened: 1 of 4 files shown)


glwagner (approved):
Should we also use `Integral`?

https://github.com/CliMA/Oceananigans.jl/blob/13b7f2366726b3e52303dee67bab264941088554/src/AbstractOperations/metric_field_reductions.jl#L74

navidcy:
Done!

glwagner (approved):

PR #2723, opened 2022-09-03
Fix test for implicit free surface with `ImmersedBoundary` and indexed `ReducedField`s

navidcy:
There is a test for immersed boundary inside `test_implicit_free_surface.jl` but it seems to be *trivially passing* (see, e.g., the [log](https://buildkite.com/clima/oceananigans/builds/8449#0182eb0d-74d3-440e-92a7-eb0ffecfb042/25-734) from tests run on `main`).

One problem was that `set_simple_divergent_velocity!` function defined at
https://github.com/CliMA/Oceananigans.jl/blob/af21542ba01afe92e7fc4de6251dec55ed5182c7/test/test_implicit_free_surface_solver.jl#L17-L35
was setting a non-zero `u` velocity in the center of the domain at `k = 1`, but that was inside the immersed boundary. I changed the [`set_simple_divergent_velocity!`](https://github.com/CliMA/Oceananigans.jl/blob/f585ef8a3ccd80d1418d9fed5d573ed7d3ecb4d3/test/test_implicit_free_surface_solver.jl#L17-L46) function to address that. However, still there is an issue...

I managed to pinpoint it down to the step:

https://github.com/CliMA/Oceananigans.jl/blob/f585ef8a3ccd80d1418d9fed5d573ed7d3ecb4d3/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl#L140

Here's a demonstration... (not a MWE). I have a model with non-zero `u` velocity:

```
julia> model.velocities.u
129×1×5 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 128×1×5 ImmersedBoundaryGrid{Float64, Bounded, Periodic, Bounded} on CPU with 3×3×3 halo
├── boundary conditions: FieldBoundaryConditions
│   └── west: Open, east: Open, south: Periodic, north: Periodic, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 135×7×11 OffsetArray(::Array{Float64, 3}, -2:132, -2:4, -2:8) with eltype Float64 with indices -2:132×-2:4×-2:8
    └── max=0.1, min=0.0, mean=0.000159236

julia> model.velocities.u[imid, 1, 3]
0.1
```

```Julia
julia> ∫ᶻQ = model.free_surface.barotropic_volume_flux
[...]

Issue #1886, opened 2021-07-22
GPU compilation errors complex GPU AbstractOperations (


glwagner:
This is a summary of the current salient issues discussed on #1241. Much of that discussion is out of date; however one issue that remains is that complex GPU AbstractOperations can produce PTX code with function signatures that consume too much "parameter space".

To reproduce this issue:

```julia
using Oceananigans
model = NonhydrostaticModel(architecture=GPU(), grid=RegularRectilinearGrid(size=(1, 1, 1), extent=(1, 1, 1)))
u, v, w = model.velocities
```

and then

```julia
julia> compute!(ComputedField(∂x(u)^2 + ∂y(v)^2 + ∂z(w)^2 + ∂x(w)^2 + ∂y(w)^2))
ERROR: Failed to compile PTX code (ptxas exited with code 255)
[...]

PR #2740, opened 2022-09-15
BC and Immersed Reductions for Sliced Fields 

simone-silvestri:
This PR is the successor of #2723

The overall objective is to improve the support of `Sliced` and `Windowed` fields such that we can represent the Free surface as a Sliced field (a regular field which lives at indices `(:, :, grid.Nz+1)`)

To do so we must
- implement halo filling support
- implement conditional reductions for windowed field

bonus
- allow interaction of windowed fields and regular fields through propagation of indices in abstract operations 

cc @glwagner @navidcy

### DIFF
--- a/src/AbstractOperations/at.jl
+++ b/src/AbstractOperations/at.jl
@@ -54,12 +54,15 @@ end
 """
     utility to propagate field indices in abstract operations
 """
-# Numbers and functions do not have indices
-indices(f::Function) = (:, :, :)
-indices(f::Number)   = (:, :, :)
 
-# Fallback (used for KernelFunctionOperation)
-indices(f) = (:, :, :)
+using Oceananigans.Fields: default_indices
+
+# Numbers and functions do not have index restrictions
+indices(f::Function) = default_indices(3)
+indices(f::Number)   = default_indices(3)
+
+# Fallback (used by KernelFunctionOperation)
+indices(f) = default_indices(3)
 
 # easy index propagation 
 function interpolate_indices(args...; loc_operation = (Center, Center, Center))
@@ -76,26 +79,29 @@ end
 interpolate_index(::Colon, ::Colon, args...)       = Colon()
 interpolate_index(::Colon, b::UnitRange, args...)  = b
 
-# REMEMBER! Not supported abstract operations which require an interpolation of sliced fields!
 function interpolate_index(a::UnitRange, ::Colon, loc, new_loc)  
     a = corrected_index(a, loc, new_loc)
-    if first(a) > last(a)
-        throw(ArgumentError("Cannot interpolate from $loc to $new_loc a Sliced field!"))
-    end
+
+    # Abstract operations that require an interpolation of a sliced fields are not supported!
+    first(a) > last(a) && throw(ArgumentError("Cannot interpolate a Sliced field from $loc to $(new_loc)!"))
     return a
 end
 
-# REMEMBER: issue an error when the indices are not compatible (e.g. parallel fields on different planes)
 function interpolate_index(a::UnitRange, b::UnitRange, loc, new_loc)   
     a = corrected_index(a, loc, new_loc)
-    if first(a) > last(a)
-        throw(ArgumentError("Cannot interpolate from $loc to $new_loc a Sliced field!"))
-    end
-    return UnitRange(max(first(a), first(b)), min(first(a), last(b)))
+
+    # Abstract operations that require an interpolation of a sliced fields are not supported!
+    first(a) > last(a) && throw(ArgumentError("Cannot interpolate a Sliced field from $loc to $(new_loc)!"))
+    
+    indices = UnitRange(max(first(a), first(b)), min(first(a), last(b)))
+    
+    # Abstract operations between parallel non-intersecating windowed fields are not 
+    first(indices) > last(indices) && throw(ArgumentError("indices $(a) and $(b) are not compatible!"))
+    return indices
 end
 
-# Windowed Fields interpolate from a `Center` to a `Face` lose the first index,
-# viceverse, windowed fields interpolated from `Face`s to `Center`s lose the last index
+# Windowed Fields interpolated from `Center`s to `Face`s lose the first index.
+# Viceverse, windowed fields interpolated from `Face`s to `Center`s lose the last index
 corrected_index(a, ::Type{Face},   ::Type{Face})   = UnitRange(first(a),   last(a))
 corrected_index(a, ::Type{Center}, ::Type{Center}) = UnitRange(first(a),   last(a))
 corrected_index(a, ::Type{Face},   ::Type{Center}) = UnitRange(first(a),   last(a)-1)
--- a/src/AbstractOperations/at.jl
+++ b/src/AbstractOperations/at.jl
@@ -63,7 +63,6 @@ function interpolate_indices(args, loc_op)
     idxs = Any[:, :, :]
     for i in 1:3
         for arg in args
-            @show idxs
             idxs[i] = interpolate_index(indices(arg)[i], idxs[i], location(arg)[i], loc_op[i])
         end
     end
@@ -77,7 +76,6 @@ interpolate_index(::Colon, b::UnitRange, args...)  = b
 # REMEMBER! Not supported abstract operations which require an interpolation of sliced fields!
 function interpolate_index(a::UnitRange, ::Colon, loc, new_loc)  
     a = corrected_index(a, loc, new_loc)
-    @show a
     if first(a) > last(a)
         throw(ArgumentError("Cannot interpolate from $loc to $new_loc a Sliced field!"))
     end
@@ -97,5 +95,5 @@ end
 # viceverse, windowed fields interpolated from `Face`s to `Center`s lose the last index
 corrected_index(a, ::Type{Face},   ::Type{Face})   = UnitRange(first(a),   last(a))
 corrected_index(a, ::Type{Center}, ::Type{Center}) = UnitRange(first(a),   last(a))
-corrected_index(a, ::Type{Face},   ::Type{Center}) = UnitRange(first(a)+1, last(a))
-corrected_index(a, ::Type{Center}, ::Type{Face})   = UnitRange(first(a),   last(a)-1)
+corrected_index(a, ::Type{Face},   ::Type{Center}) = UnitRange(first(a),   last(a)-1)
+corrected_index(a, ::Type{Center}, ::Type{Face})   = UnitRange(first(a)+1, last(a))
--- a/src/AbstractOperations/at.jl
+++ b/src/AbstractOperations/at.jl
@@ -58,6 +58,9 @@ end
 indices(f::Function) = (:, :, :)
 indices(f::Number)   = (:, :, :)
 
+# Fallback (used for KernelFunctionOperation)
+indices(f) = (:, :, :)
+
 # easy index propagation 
 function interpolate_indices(args, loc_op)
     idxs = Any[:, :, :]
--- a/src/AbstractOperations/at.jl
+++ b/src/AbstractOperations/at.jl
@@ -51,10 +51,6 @@ macro at(location, abstract_operation)
     return wrapped_operation
 end
 
-"""
-    utility to propagate field indices in AbstractOperations
-"""
-
 using Oceananigans.Fields: default_indices
 
 # Numbers and functions do not have index restrictions
@@ -64,7 +60,13 @@ indices(f::Number)   = default_indices(3)
 # Fallback (used by KernelFunctionOperation)
 indices(f) = default_indices(3)
 
-# easy index propagation 
+
+"""
+    interpolate_indices(operands..., loc_operation = abstract_operation_location)
+
+Utility to propagate operands' indices in `AbstractOperations`s with multiple operands 
+(`BinaryOperation`s and `MultiaryOperation`s)
+"""
 function interpolate_indices(args...; loc_operation = (Center, Center, Center))
     idxs = Any[:, :, :]
     for i in 1:3
--- a/src/AbstractOperations/at.jl
+++ b/src/AbstractOperations/at.jl
@@ -77,32 +77,23 @@ interpolate_index(::Colon, b::UnitRange, args...)  = b
 # otherwise we lose the last index
 # REMEMBER! Not supported abstract operations which require an interpolation of sliced fields!
 function interpolate_index(a::UnitRange, ::Colon, loc, new_loc)  
-    if loc == new_loc
-        return a
-    else
-        if a[1] == a[2]
-            throw(ArgumentError("Cannot interpolate from $loc to $new_loc a Sliced field!"))
-        end
-        if new_loc == Face
-            return UnitRange(a[1]+1, a[2])
-        else
-            return UnitRange(a[1], a[2]-1)    
-        end
+    a = corrected_index(a[1], loc, new_loc)
+    if a[1] > a[2]
+        throw(ArgumentError("Cannot interpolate from $loc to $new_loc a Sliced field!"))
     end
+    return a
 end
 
 # REMEMBER: issue an error when the indices are not compatible (e.g. parallel fields on different planes)
 function interpolate_index(a::UnitRange, b::UnitRange, loc, new_loc)   
-    if loc == new_loc
-        return UnitRange(max(a[1], b[1]), min(a[2], b[2]))
-    else
-        if a[1] == a[2]
-            throw(ArgumentError("Cannot interpolate from $loc to $new_loc a Sliced field!"))
-        end
-        if new_loc == Face
-            return UnitRange(max(a[1]+1, b[1]),min(a[2], b[2]))
-        else
-            return UnitRange(max(a[1], b[1]),min(a[2]-1, b[2]))
-        end
+    a = corrected_index(a[1], loc, new_loc)
+    if a[1] > a[2]
+        throw(ArgumentError("Cannot interpolate from $loc to $new_loc a Sliced field!"))
     end
+    return UnitRange(max(a[1], b[1]), min(a[2], b[2]))
 end
+
+corrected_index(a, ::Face, ::Face)     = UnitRange(a[1], a[2])
+corrected_index(a, ::Center, ::Center) = UnitRange(a[1], a[2])
+corrected_index(a, ::Face, ::Center)   = UnitRange(a[1]+1, a[2])
+corrected_index(a, ::Center, ::Face)   = UnitRange(a[1], a[2]-1)
--- a/src/AbstractOperations/at.jl
+++ b/src/AbstractOperations/at.jl
@@ -59,21 +59,50 @@ indices(f::Function) = (:, :, :)
 indices(f::Number)   = (:, :, :)
 
 # easy index propagation 
-function interpolate_indices(args...)
+function interpolate_indices(args, loc_op)
     idxs = Any[:, :, :]
     for i in 1:3
         for arg in args
-            idxs[i] = interpolate_index(indices(arg)[i], idxs[i])
+            idxs[i] = interpolate_index(indices(arg)[i], idxs[i], location(arg)[i], loc_op[i])
         end
     end
 
     return Tuple(idxs)
 end
 
+interpolate_index(::Colon, ::Colon, args...)       = Colon()
+interpolate_index(::Colon, b::UnitRange, args...)  = b
+
+# If we interpolate from a `Center` to a `Face` we lose the first index,
+# otherwise we lose the last index
+# REMEMBER! Not supported abstract operations which require an interpolation of sliced fields!
+function interpolate_index(a::UnitRange, ::Colon, loc, new_loc)  
+    if loc == new_loc
+        return a
+    else
+        if a[1] == a[2]
+            throw(ArgumentError("Cannot interpolate from $loc to $new_loc a Sliced field!"))
+        end
+        if new_loc == Face
+            return UnitRange(a[1]+1, a[2])
+        else
+            return UnitRange(a[1], a[2]-1)    
+        end
+    end
+end
+
 # REMEMBER: issue an error when the indices are not compatible (e.g. parallel fields on different planes)
-# Non supported abstract operations which require an interpolation of sliced fields!
-# (maybe also issue an error in that case)
-interpolate_index(::Colon, ::Colon) = Colon()
-interpolate_index(a, ::Colon) = a
-interpolate_index(::Colon, b) = b
-interpolate_index(a, b)       = max(first(a), first(b)):min(last(a), last(b))
+function interpolate_index(a::UnitRange, b::UnitRange, loc, new_loc)   
+    if loc == new_loc
+        return UnitRange(max(a[1], b[1]), min(a[2], b[2]))
+    else
+        if a[1] == a[2]
+            throw(ArgumentError("Cannot interpolate from $loc to $new_loc a Sliced field!"))
+        end
+        if new_loc == Face
+            return UnitRange(max(a[1]+1, b[1]),min(a[2], b[2]))
+        else
+            return UnitRange(max(a[1], b[1]),min(a[2]-1, b[2]))
+        end
+    end
+end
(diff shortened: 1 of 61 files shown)


simone-silvestri:
@glwagner, @navidcy 
Poll to decide where to put `η` in `z`. 

The possibilities are:
- location `Face`, index `grid.Nz+1`. Pros: natural location of `η` field (same position of the `w` velocity).
- location `Center`, index `grid.Nz`. Pros: every calculation dealing with `η` is now performed at centers (ex `Az_∇h²ᶜᶜᶜ` and `linear_operation!`. 

In case of the second option should we change all the metrics to reflect the fact that we are at faces?

navidcy:
I think `η` belongs at the top-face since it's time-derivative is `w`.

glwagner:
Can't we define the topmost face metrics to be identical to the first cell?

src/Models/HydrostaticFreeSurfaceModels/fft_based_implicit_free_surface_solver.jl
```diff
@@ -110,8 +110,8 @@ end
 
 @kernel function fft_implicit_free_surface_right_hand_side!(rhs, grid, g, Lz, Δt, ∫ᶻQ, η)
     i, j = @index(Global, NTuple)
-    Az = Azᶜᶜᶜ(i, j, 1, grid)
+    Az = Azᶜᶜᶜ(i, j, grid.Nz, grid)
     δ_Q = flux_div_xyᶜᶜᶜ(i, j, 1, grid, ∫ᶻQ.u, ∫ᶻQ.v)
-    @inbounds rhs[i, j, 1] = (δ_Q - Az * η[i, j, 1] / Δt) / (g * Lz * Δt * Az)
+    @inbounds rhs[i, j, grid.Nz] = (δ_Q - Az * η[i, j, grid.Nz] / Δt) / (g * Lz * Δt * Az)
```
navidcy, line 115: `rhs` is a 2D array I believe so perhaps this should be
```suggestion
    @inbounds rhs[i, j, 1] = (δ_Q - Az * η[i, j, grid.Nz] / Δt) / (g * Lz * Δt * Az)
```
?

simone-silvestri:
Ok I can change it, at moment is the same field as `η`

navidcy:
@simone-silvestri, seems like the output writer is trying to save with halos despite `with_halos = false`.
https://buildkite.com/clima/oceananigans/builds/8658#018376f6-45e7-4a9b-a40f-c9d5c76d5517/27-398

Is this a clue?

navidcy:
> @simone-silvestri, seems like the output writer is trying to save with halos despite `with_halos = false`. https://buildkite.com/clima/oceananigans/builds/8658#018376f6-45e7-4a9b-a40f-c9d5c76d5517/27-398
> 
> Is this a clue?

I tried to trace it down. What's happening is that 

https://github.com/CliMA/Oceananigans.jl/blob/c08e3d8c1b2e6639a2726bc544885ebf8853e504/src/OutputWriters/netcdf_output_writer.jl#L423

returns a `data` as an array **with the halos** (wrongly probably) and then

https://github.com/CliMA/Oceananigans.jl/blob/c08e3d8c1b2e6639a2726bc544885ebf8853e504/src/OutputWriters/netcdf_output_writer.jl#L424

fallsback to 

https://github.com/CliMA/Oceananigans.jl/blob/c08e3d8c1b2e6639a2726bc544885ebf8853e504/src/OutputWriters/netcdf_output_writer.jl#L489

src/OutputWriters/output_construction.jl
```diff
@@ -43,10 +61,12 @@ end
 
 function construct_output(user_output::Union{AbstractField, Reduction}, grid, user_indices, with_halos)
     indices = output_indices(user_output, grid, user_indices, with_halos)
+    user_output = maybe_sliced_field(user_output, indices)
+
     return construct_output(user_output, indices)
 end
 
-construct_output(user_output::Field, indices) = view(user_output, indices...)
+construct_output(user_output::Field, indices) = user_output
```
navidcy, line 69: why is this fallback introduced?

test/test_implicit_free_surface_solver.jl
```diff
@@ -29,31 +30,15 @@ function set_simple_divergent_velocity!(model)
...
-    if grid isa RectilinearGrid
-        Δx, Δy = grid.Δxᶠᵃᵃ, grid.Δyᵃᶜᵃ
-    end
-
-    if grid isa ImmersedBoundaryGrid && grid.underlying_grid isa RectilinearGrid
-        Δx, Δy = grid.underlying_grid.Δxᶠᵃᵃ, grid.underlying_grid.Δyᵃᶜᵃ
-    end
+    # pick a surface cell at the middle of the domain
+    i, j, k = Int(floor(grid.Nx / 2)) + 1, Int(floor(grid.Ny / 2)) + 1, grid.Nz
+    inactive_cell(i, j, k, grid) && error("The nudged cell at ($i, $j, $k) is inactive.")
```
navidcy, line 35: why we revert these?

test/test_implicit_free_surface_solver.jl
```diff
@@ -29,31 +30,15 @@ function set_simple_divergent_velocity!(model)
...
 
-    if grid isa LatitudeLongitudeGrid
-        Δx, Δy = CUDA.@allowscalar grid.Δxᶠᶜᵃ[i], grid.Δyᶜᶠᵃ
-    end
-    if grid isa ImmersedBoundaryGrid && grid.underlying_grid isa LatitudeLongitudeGrid
-        Δx, Δy = CUDA.@allowscalar grid.underlying_grid.Δxᶠᶜᵃ[i], grid.underlying_grid.Δyᶜᶠᵃ
-    end
+    Δy = Δyᶜᶠᶜ(i, j, k, grid)
+    Δz = Δzᶜᶠᶜ(i, j, k, grid)
```
navidcy, line 38: and these?

navidcy: oh these might be better... I didn't know about these: `Δyᶜᶠᶜ`, `Δzᶜᶠᶜ`. Don't these require `CUDA.@allowscalar`?

test/test_implicit_free_surface_solver.jl
```diff
@@ -29,31 +30,15 @@ function set_simple_divergent_velocity!(model)
...
 
-    Δz = CUDA.@allowscalar grid.Δzᵃᵃᶜ[k]
-
-    # Instead of prescribing the velocity, we prescribe the value of the zonal transport
-    # in a cell, i.e., `u * Δy * Δz`. This way the norm(rhs) of the free-surface solver
-    # does not depend on the grid extend and resolution.
+    # We prescribe the value of the zonal transport in a cell, i.e., `u * Δy * Δz`. This
+    # way `norm(rhs)` of the free-surface solver does not depend on the grid extensd/resolution.
```
navidcy, line 41:
```suggestion
    # way `norm(rhs)` of the free-surface solver does not depend on the grid extent/resolution.
```

simone-silvestri:
You are right, I did something wrong with the output writers, I hadn't realized it is a problem of halos, I ll try to give a better look.

The only thing that I had to change for output writers is to validate BC when non trivial indices are given as an argument. 

That `view(field, indices...)` confused me though. I guess by changing that I had fields saved with halos?

test/test_implicit_free_surface_solver.jl
simone-silvestri: Hmmm they should... I am not sure why they don't error

test/test_implicit_free_surface_solver.jl
simone-silvestri: I think `Δx` was not used right?

src/OutputWriters/output_construction.jl
simone-silvestri: I am a bit confused about this `construct_output`. Why does a field output to a `view` while an `AbstractOperation` outputs to a field instead of a `view`? cc @glwagner

simone-silvestri:
Ok I see, `view(field, indices...)` returns a field with a `view` of the data at the indices. But then it's ok to just instead create a field with the correct data so that the computation of the `ComputedField` will be done on the correct indices instead of on all the grid. That was what I tried to do with the `maybe_sliced_field` that then supersedes the need of `view(field, indices...)` since the offset data is sliced correctly then. But this maybe needs a change in the `with_halos` handling?

navidcy (approved):

navidcy:
Is there a test or example that saves output of the free surface? Perhaps that will sort out your concern at 1. above @simone-silvestri?

simone-silvestri:
Well the issue is that if we want to save the output at the surface `(:, :, grid.Nz)` we cannot include the free surface if this is defined at `grid.Nz+1` (it would not exist at `grid.Nz`). I see it a little bit like an inconvenience, but if for you guys it's not a problem it's fine

navidcy:
> Well the issue is that if we want to save the output at the surface `(:, :, grid.Nz)` we cannot include the free surface if this is defined at `grid.Nz+1` (it would not exist at `grid.Nz`). I see it a little bit like an inconvenience, but if for you guys it's not a problem it's fine

It’s ok. We’ll make a not in the docstrings or in the example!

src/Models/HydrostaticFreeSurfaceModels/matrix_implicit_free_surface_solver.jl
```diff
@@ -104,7 +104,7 @@ end
     Az   = Azᶜᶜᶜ(i, j, 1, grid)
```
navidcy, line 104: @simone-silvestri, should this be `Az   = Azᶜᶜᶜ(i, j, Nz, grid)`

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -25,7 +25,7 @@ end
     # time-averaging
     @inbounds U̅[i, j, 1] +=  velocity_weight * U[i, j, 1]
     @inbounds V̅[i, j, 1] +=  velocity_weight * V[i, j, 1]
-    @inbounds η̅[i, j, 1] +=  free_surface_weight * η[i, j, 1]
+    @inbounds η̅[i, j, 1] +=  free_surface_weight * η[i, j, grid.Nz]
```
navidcy, line 28: or is it `η[i, j, grid.Nz + 1]`?

simone-silvestri:
Theoretically, it should be ok, but there is a problem with `conditional_fluxes`. The last cell is a `peripheral_node` since the grid is Bounded in the z direction, so that derivative will be 0. Probably I should change that behaviour to have derivatives go to zero only on `immersed_peripheral_nodes`, but I am not sure about the implication of that. It might take a while to fix if we want to change those derivatives. But I guess I ll have to do it...

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
simone-silvestri: it's at `Nz+1` now

src/Models/HydrostaticFreeSurfaceModels/matrix_implicit_free_surface_solver.jl
simone-silvestri: good catch

simone-silvestri:
I was wrong, `conditional_derivatives` can calculate the `x` and `y` derivative at a z `Face` at `grid.Nz+1`. Good!

simone-silvestri:
I think if the tests pass we are ready to merge

navidcy (approved):

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -14,18 +14,18 @@ using Oceananigans.Operators
...
-    @inbounds V[i, j, 1] +=  Δτ * (-g * Hᶜᶠ[i, j] * ∂yᶜᶠᶜ(i, j, 1, grid, η) + Gⱽ[i, j, 1])
+    @inbounds U[i, j, 1] +=  Δτ * (-g * Hᶠᶜ[i, j] * ∂xᶠᶜᶠ(i, j, grid.Nz+1, grid, η) + Gᵁ[i, j, 1])
+    @inbounds V[i, j, 1] +=  Δτ * (-g * Hᶜᶠ[i, j] * ∂yᶜᶠᶠ(i, j, grid.Nz+1, grid, η) + Gⱽ[i, j, 1])
 end
 
 @kernel function split_explicit_free_surface_substep_kernel_2!(grid, Δτ, η, U, V, η̅, U̅, V̅, velocity_weight, free_surface_weight)
     i, j = @index(Global, NTuple)
     # ∂τ(η) = - ∇⋅U
-    @inbounds η[i, j, 1] -=  Δτ * div_xyᶜᶜᶜ(i, j, 1, grid, U, V)
+    @inbounds η[i, j, grid.Nz+1] -=  Δτ * div_xyᶜᶜᶜ(i, j, 1, grid, U, V)
```
navidcy, line 24:
```suggestion
    @inbounds η[i, j, grid.Nz+1] -=  Δτ * div_xyᶜᶜᶜ(i, j, grid.Nz, grid, U, V)
```

??

simone-silvestri: correct! (technically I am not sure how to treat `ReducedFields` probably we should still go at `grid.Nz+1`?)

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -46,16 +46,14 @@ end
 function SplitExplicitFreeSurface(grid; gravitational_acceleration = g_Earth,
                                         settings = SplitExplicitSettings(200))
 
-    η = ZFaceField(grid, indices = (:, :, size(grid, 3)+1))
+    η = FreeSurfaceDisplacementField(velocities, free_surface, grid)
```
simone-silvestri, line 49:
```suggestion
η = ZFaceField(grid, indices = (:, :, size(grid, 3)+1))
```
velocities is not defined here

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
navidcy: but doesn't `div_xyᶜᶜᶜ` involve `Azᶜᶜᶜ(i, j, k)`? That should be at `k = Nz`, right?

simone-silvestri: you're right. then we should write a `div_xyᶜᶜᶠ` for fields at `Face`s

navidcy: Should we have had a test to pick this up?

glwagner: I don't think this is correct. `U` and `V` are _reduced_ fields with only one grid point. This differs from _sliced_ fields, which exist at a particular model level. `k=1` is better.

glwagner: The problem with `k=Nz` is that it suggests incorrectly that `U` and `V` depend on z.

glwagner: @navidcy didn't realize that `U` and `V` were barotropic; that's why he made his comment (that's why we need to use `k=1` and also include a comment.)

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -13,25 +13,29 @@ using Oceananigans.Operators
...
+    @inbounds V[i, j, 1] +=  Δτ * (-g * Hᶜᶠ[i, j] * ∂yᶜᶠᶠ(i, j, k_surface, grid, η) + Gⱽ[i, j, 1])
 end
 
 @kernel function split_explicit_free_surface_substep_kernel_2!(grid, Δτ, η, U, V, η̅, U̅, V̅, velocity_weight, free_surface_weight)
     i, j = @index(Global, NTuple)
+    k_surface = grid.Nz+1
+    
     # ∂τ(η) = - ∇⋅U
-    @inbounds η[i, j, 1] -=  Δτ * div_xyᶜᶜᶜ(i, j, 1, grid, U, V)
+    @inbounds η[i, j, k_surface] -=  Δτ * div_xyᶜᶜᶠ(i, j, k_surface, grid, U, V)
```
glwagner, line 28:
```suggestion
    @inbounds η[i, j, k_surface] -=  Δτ * div_xyᶜᶜᶠ(i, j, 1, grid, U, V)
```

glwagner (changes requested):

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -13,25 +13,29 @@ using Oceananigans.Operators
...
     # ∂τ(U) = - ∇η + G
-    @inbounds U[i, j, 1] +=  Δτ * (-g * Hᶠᶜ[i, j] * ∂xᶠᶜᶜ(i, j, 1, grid, η) + Gᵁ[i, j, 1])
-    @inbounds V[i, j, 1] +=  Δτ * (-g * Hᶜᶠ[i, j] * ∂yᶜᶠᶜ(i, j, 1, grid, η) + Gⱽ[i, j, 1])
+    @inbounds U[i, j, 1] +=  Δτ * (-g * Hᶠᶜ[i, j] * ∂xᶠᶜᶠ(i, j, k_surface, grid, η) + Gᵁ[i, j, 1])
+    @inbounds V[i, j, 1] +=  Δτ * (-g * Hᶜᶠ[i, j] * ∂yᶜᶠᶠ(i, j, k_surface, grid, η) + Gⱽ[i, j, 1])
 end
 
 @kernel function split_explicit_free_surface_substep_kernel_2!(grid, Δτ, η, U, V, η̅, U̅, V̅, velocity_weight, free_surface_weight)
     i, j = @index(Global, NTuple)
+    k_surface = grid.Nz+1
```
glwagner, line 25: If you like

```suggestion
    k_barotropic = 1
```

But actually, I think a comment is better. Reduced fields have _no index_ in the reduced direction. Barotropic velocities are reduced fields.

glwagner:
I have a question. Do `AbstractOperations` need indices? It seems like it should be possible to "compute" the indices of an operation by recursively evaluating `indices`.

Consider:

```julia
a = CenterField(grid, indices=(:, :, 1))
b = CenterField(grid)

ab = a * b
```

We then need to be able to evaluate `indices(ab)`. For this we can have something like

```julia
function indices(ab::BinaryOperation)
    ai = indices(a)
    bi = indices(b)
    return intersection(ai, bi)
end
```

It's helpful to keep the "size" of the operations as small as possible due to parameter space restrictions on the GPU. This might help.

simone-silvestri:
Yeah, it can be done like that!

glwagner:
> Yeah, it can be done like that!

I'm worried about parameter space issues like those encountered on https://github.com/CliMA/Oceananigans.jl/issues/1886...

src/AbstractOperations/binary_operations.jl
```diff
@@ -29,12 +29,15 @@ end
 # Recompute location of binary operation
 @inline at(loc, β::BinaryOperation) = β.op(loc, at(loc, β.a), at(loc, β.b))
 
+indices(β::BinaryOperation) = interpolate_indices(β.a, β.b)
```
glwagner, line 32: Doesn't `interpolate_indices` need to know about `▶a` and `▶b`? Because the indices of the operator depend on both the indices of the underlying field, and how that field is interpolated?

simone-silvestri: (if we interpolate in the direction of the "slice")

glwagner: Alternatively it needs to know the location of the fields (that might be easier since you can design some logic around multiple dispatch, potentially saving some code)

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
simone-silvestri: here we need `k_surface = grid.Nz+1` because `η` is at the surface. I ll add a comment

simone-silvestri:
Ready to go?

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -11,6 +11,9 @@ using Oceananigans.Operators
 ∂t(U) = - gH∇η + f
 =#
 
+# the free surface η and its average η̄ are located on `Face`s at the surface (grid.Nz +1). All other intermediate variables
+# (U, V, , Ū, V̄) are barotropic fields (`ReducedField`) for which a k index is not defined
+
 @kernel function split_explicit_free_surface_substep_kernel_1!(grid, Δτ, η, U, V, Gᵁ, Gⱽ, g, Hᶠᶜ, Hᶜᶠ)
     i, j = @index(Global, NTuple)
     k_surface = grid.Nz+1
```
glwagner, line 19: or `k_top` as we refer to this location elsewhere

glwagner:
We should have a new test or two for the "index computation", right?

glwagner:
Here's some ideas:

```julia
grid = RectilinearGrid(size=(4, 5, 6), extent=(1, 1, 1))
a = CenterField(grid)
b = CenterField(grid, indices = (2, :, :))
@test indices(a * b) == (2:2, :, :)

# repeat for y and z slices?

c = CenterField(grid)
d = XFaceField(grid, indices=(2:3, :, :))
@test indices(c * d) == (3:3, :, :) # right?

# repeat for y and z...
```

Those, plus maybe some tests for `UnaryOperation`, etc, cover the non-trivial cases I can think of.

glwagner (approved):

src/AbstractOperations/at.jl
```diff
@@ -93,7 +93,7 @@ function interpolate_index(a::UnitRange, b::UnitRange, loc, new_loc)
     # Abstract operations that require an interpolation of a sliced fields are not supported!
     first(a) > last(a) && throw(ArgumentError("Cannot interpolate a Sliced field from $loc to $(new_loc)!"))
     
-    indices = UnitRange(max(first(a), first(b)), min(first(a), last(b)))
+    indices = UnitRange(max(first(a), first(b)), min(last(a), last(b)))
     
     # Abstract operations between parallel non-intersecating windowed fields are not 
     first(indices) > last(indices) && throw(ArgumentError("indices $(a) and $(b) are not compatible!"))
```
glwagner, line 99:
```suggestion
    first(indices) > last(indices) && throw(ArgumentError("BinaryOperation operand indices $(a) and $(b) of do not intersect!"))
```

src/AbstractOperations/at.jl
```diff
@@ -93,7 +93,7 @@ function interpolate_index(a::UnitRange, b::UnitRange, loc, new_loc)
     # Abstract operations that require an interpolation of a sliced fields are not supported!
     first(a) > last(a) && throw(ArgumentError("Cannot interpolate a Sliced field from $loc to $(new_loc)!"))
```
glwagner, line 94:
```suggestion
    first(a) > last(a) && throw(ArgumentError("Cannot interpolate a SlicedField from $loc to $(new_loc)!"))
```

or just "sliced field" if we are not planning to define the type alias SlicedField (we reserve TitleCase for types + type aliases)

src/AbstractOperations/at.jl
```diff
@@ -52,7 +52,7 @@ macro at(location, abstract_operation)
 end
 
 """
-    utility to propagate field indices in abstract operations
+    utility to propagate field indices in AbstractOperations
```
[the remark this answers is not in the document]
simone-silvestri: I don't think so, I change it to a comment?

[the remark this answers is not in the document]
glwagner: A comment is probably better if its not a docstring.

PR #2746, opened 2022-09-22
Bug fix for general interpolation

simone-silvestri:
which was used for the `DiscreteDiffusionFunction`


navidcy (approved):
Looks good. Did it fix the issue?

simone-silvestri:
Well, I was looking into `DiscreteDiffusionFunction` because I have a weird behavior in my simulation which I believed could arise from the Leith viscosity that uses that object. Unfortunately, this bug impacts only fields defined on `Face`s while horizontal viscosity (for which I am using `DiscreteDiffusionFunction`) is defined on `Center`s, so no luck there... Nonetheless, it is still a bug to be solved

Issue #2751, opened 2022-09-25
`calc_κᶜᶜᶜ()` is only defined for AMD and questions about its default behavior


tomchor:
Sorry in advance if I'm packing too much stuff into one issue. Let me know if that's the case and I can separate this into other issues.

The function `calc_κᶜᶜᶜ()` is only defined for the AMD closure, so for every other closure it silently defaults to https://github.com/CliMA/Oceananigans.jl/blob/56c70fb07eeeb8171958dc62ed7948ff2617cf7a/src/TurbulenceClosures/turbulence_closure_utils.jl#L33

So one obvious thing is that we definitely should define `calc_κᶜᶜᶜ()` for other closures.

A bigger question regards the use of multiple dispatch to create this "default" behavior for when the desired behavior isn't implemented yet. In this case, silently defaulting to `zero()` made be lose some hours of work because I thought there was something wrong with my script, when in reality `calc_κᶜᶜᶜ()` just wasn't defined for the closure I was using. Sure, this is a bit my fault for not checking the code thoroughly, but from a user perspective my opinion is that it would have been better to throw a "method not found" error or a warning then have the code fail silently. So I propose we either get rid of that general default behavior and just let the code fail with an error for the unimplemented cases, or at least add a warning to the lines above. Thoughts?

A follow-up question/comment regards the organization of the `TurbulenceClosure` module. Part of the reason for my delay in catching the  `calc_κᶜᶜᶜ()` issue is that I often get confused about the organizational aspect in that module. For example, there are several functions that, based on name, appear to do the same thing (`νᶜᶜᶜ`, `calc_νᶜᶜᶜ`, `getdiffusivity`; there's also a `calc_κᶜᶜᶜ` but no `κᶜᶜᶜ` function) and often the naming of the variables also doesn't help much. For example in these lines:

https://github.com/CliMA/Oceananigans.jl/blob/56c70fb07eeeb8171958dc62ed7948ff2617cf7a/src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl#L138

https://github.com/CliMA/Oceananigans.jl/blob/56c70fb07eeeb8171958dc62ed7948ff2617cf7a/src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl#L160

[...]

PR #2752, opened 2022-09-25
Changes to naming in the  `TurbulenceClosures` module

tomchor:
This PR will rename some variables and funcitons in the `TurbulenceClosures` module to make things more clear.

~It will also remove the fallback dispatch of `calc_ κᶜᶜᶜ()` and define `calc_ κᶜᶜᶜ()` at least for the Smagorisnky-Lilly closure.~

It also changes the behavior of `viscosity()` to return a tuple when given a tuple, instead of `sum`ming all the viscosities in the tuple. This avoids misuse by unattentive users who have a tuple with different-formulation-viscosities (for example horizontal and vertical formulations). ([This comment](https://github.com/CliMA/Oceananigans.jl/pull/2752#issuecomment-1270199938) provides an example.)

Closes https://github.com/CliMA/Oceananigans.jl/issues/2751

CC @simone-silvestri


simone-silvestri:
We could also just change the signature of the `calc_` functions to pass all the fields instead of just velocities and tracers separately `fields(model)`

tomchor:
> We could also just change the signature of the `calc_` functions to pass all the fields instead of just velocities and tracers separately `fields(model)`

Yeah I like that idea. Also, is there anything that keeps us from passing `model` itself? It seems like it'd simplify the interfaces

simone-silvestri:
There is no adapt method for the model. I don't know how it would play to send the whole model to the GPU. We might have some parameter space issue

src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl
```diff
@@ -96,15 +96,25 @@ filter width `Δᶠ`, and strain tensor dot product `Σ²`.
...
+
+@inline function calc_nonlinear_κᶜᶜᶜ(i, j, k, grid, closure::SmagorinskyLilly, buoyancy, U, C, ::Val{tracer_index}) where {tracer_index}
+    νₑ = calc_nonlinear_νᶜᶜᶜ(i, j, k, grid, closure, buoyancy, U, C)
+
+    @inbounds Pr = closure.Pr[tracer_index]
+
+    return νₑ / Pr
+end
+
+
```
simone-silvestri, line 117: I am not sure this is needed, `κₑ` is given as a `BinaryOperation` so calculating `νₑ` should be enough

tomchor: My thinking here is that it'd be good to unify the interface across models. I was also planning on creating dispatches of this function for `ScalarDiffusivity()` and tuple closures. It'd make things like these easier (in Oceanostics): https://github.com/tomchor/Oceanostics.jl/blob/29a544d3decd833d9f86da05b66a7392197c3b93/src/Oceanostics.jl#L25-L39

For example, as a user, I don't know what function I should use to get the diffusivities and viscosities that works for different closures. As far as I know there isn't one. This is my attempt to get us closer.

Am I missing something?

simone-silvestri: Ok I agree, but I would also leave that `calc_nonlinear_κᶜᶜᶜ` is used to calculate the diffusivity while `κᶜᶜᶜ` is used to retrieve the diffusivity

glwagner: Unity may be a good goal, but we shouldn't add code until the moment we need it / without tests.

glwagner:
Why introduce a vector? What does it represent? (The diagonal elements of a hypothetical viscosity tensor, which is defined implicitly?) How would it be used?

tomchor:
> Why introduce a vector? What does it represent? (The diagonal elements of a hypothetical viscosity tensor, which is defined implicitly?) How would it be used?

In cases where the viscosity is a tuple, it would return the correct viscosities/diffusivities in each direction. For example in a case like the snippet below

```
tri = ScalarDiffusivity(ThreeDimensionalFormulation(), ν=1e-3, κ=1e-4)
hor = ScalarDiffusivity(HorizontalFormulation(), ν=1e-3, κ=(b=1e-4, c=1e-5))
ver = ScalarDiffusivity(VerticalFormulation(), ν=1e-4, κ=(b=1e-5, c=1e-6))
tup = (hor, ver, tri)
```

using `viscosity(model.closure, model.diffusivity_fields)` in this PR returns the expected viscosity vector with different viscosities in each direction. This would, imo, make calculating diagnostics where the viscosity is important (most notably dissipation terms) easier for the user.

Using the same code in the master branch returns one number (`1e-3+1e-3+1e-4`) as the viscosity, which isn't what one would expect, and which requires some extra attention and code from the user when calculating viscosity terms.

simone-silvestri:
I see. It would be nice to have this as a user API convenience, but then to be complete we must include all the other entries of the tensor. It might get a bit tedious when you want to include isopycnal diffusivities like `TwoDimensionalLeith` or GM (`IsopycnalSkewSymmetricDiffusivity`) for which we do not compute the tensor but directly the flux

tomchor:
> I see. It would be nice to have this as a user API convenience, but then to be complete we must include all the other entries of the tensor. It might get a bit tedious when you want to include isopycnal diffusivities like `TwoDimensionalLeith` or GM (`IsopycnalSkewSymmetricDiffusivity`) for which we do not compute the tensor but directly the flux

I think the other tensor entries are much less frequently needed, so I'd propose we start only with the simple stuff and add the rest when needed.

I was also thinking that for inner-working purposes it's best to keep the current `viscosity()` behavior (always returning a 'Number`) and create a second function that can be exported to the user which has the array behavior I described. Thoughts?

glwagner:
Why is a vector that represents the diagonal elements of a hypothetical viscosity tensor useful?

glwagner:
I think what I'm asking is what you envision using this vector for. I don't see the purpose of it. One could just as easily inspect each closure in the tuple individually to see what the viscosity for each closure type is. This method is unambiguous and far more general than developing a "vector abstraction" for the diagonal components of a viscosity tensor.

tomchor:
> I think what I'm asking is what you envision using this vector for. I don't see the purpose of it. One could just as easily inspect each closure in the tuple individually to see what the viscosity for each closure type is. This method is unambiguous and far more general than developing a "vector abstraction" for the diagonal components of a viscosity tensor.
> 
> That said, I agree that the output of `viscosity` for a closure tuple isn't useful. It's summing the viscosities together. This makes sense when the formulations are the same for each component, but not when they are different. So we should change that.
> 
> The main use case envisioned is when you have two `ScalarDiffusivity` with `ThreeDimensionalFormulation`. In that case, `viscosity` returns an object (a `BinaryOperation`) representing the sum of a the nonlinear diffusivity (a field) and the background molecular component (a number). This can then be used in subsequent `AbstractOperations` for calculations.
> 
> When the closure tuple involves multiple `ScalarDiffusivity` with heterogeneous formulations, we need a different abstraction.
> 
> We also need to deal with the case where the closure tuple contains non-scalar-diffusivity closures.
> 
> To design an abstraction, I think we should start with a use case, which can help us develop requirements for the abstraction. Once we have requirements, we can implement something minimal and simple that's easy to reason with and that will generalize to more complicated use cases we may want to consider in the future (hopefully).



I agree with this. I'll revert `viscosity()` to its original formulation for now. My only question/suggestion is: should we check in `viscosity(::Tuple)` that the all the elements have the same formulation? That way we can throw a warning (or maybe even an error) when trying to add different formulation closures.

glwagner:
I wonder if we should return a tuple instead, and let the user compute the sum if that's valid? Eg

```julia
viscosity(diffusivities, closure_tuple) # returns a tuple

# For users who know this is valid:
nu = sum(viscosity(diffusivities, closure_tuple))
```

tomchor:
> I wonder if we should return a tuple instead, and let the user compute the sum if that's valid? Eg
> 
> ```julia
> viscosity(diffusivities, closure_tuple) # returns a tuple
> 
> # For users who know this is valid:
> nu = sum(viscosity(diffusivities, closure_tuple))
> ```

I like this idea! It's implemented right now. 

@glwagner @simone-silvestri  Let me know if there's anything left to adjust in this PR

src/Models/ShallowWaterModels/shallow_water_diffusion_operators.jl
```diff
@@ -17,7 +17,7 @@ import Oceananigans.TurbulenceClosures:
                         calculate_nonlinear_viscosity!, 
                         viscosity,
                         with_tracers,
-                        calc_νᶜᶜᶜ,
+                        calc_nonlinear_νᶜᶜᶜ,
```
glwagner, line 20: Why this change?

src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
```diff
@@ -34,38 +34,38 @@ const AHBD = AbstractScalarBiharmonicDiffusivity{<:HorizontalFormulation}
...
+@inline diffusive_flux_x(i, j, k, grid, closure::AIBD, K, ::Val{id}, c, clk, fields, b) where id = κ_σᶠᶜᶜ(i, j, k, grid, closure, K, Val(id), clk, fields, biharmonic_mask_x, ∂xᶠᶜᶜ, ∇²ᶜᶜᶜ, c)
+@inline diffusive_flux_y(i, j, k, grid, closure::AIBD, K, ::Val{id}, c, clk, fields, b) where id = κ_σᶜᶠᶜ(i, j, k, grid, closure, K, Val(id), clk, fields, biharmonic_mask_y, ∂yᶜᶠᶜ, ∇²ᶜᶜᶜ, c)
+@inline diffusive_flux_z(i, j, k, grid, closure::AIBD, K, ::Val{id}, c, clk, fields, b) where id = κ_σᶜᶜᶠ(i, j, k, grid, closure, K, Val(id), clk, fields, biharmonic_mask_z, ∂zᶜᶜᶠ, ∇²ᶜᶜᶜ, c)
+@inline diffusive_flux_x(i, j, k, grid, closure::AHBD, K, ::Val{id}, c, clk, fields, b) where id = κ_σᶠᶜᶜ(i, j, k, grid, closure, K, Val(id), clk, fields, biharmonic_mask_x, ∂x_∇²h_cᶠᶜᶜ, c)
+@inline diffusive_flux_y(i, j, k, grid, closure::AHBD, K, ::Val{id}, c, clk, fields, b) where id = κ_σᶜᶠᶜ(i, j, k, grid, closure, K, Val(id), clk, fields, biharmonic_mask_y, ∂y_∇²h_cᶜᶠᶜ, c)
+@inline diffusive_flux_z(i, j, k, grid, closure::AVBD, K, ::Val{id}, c, clk, fields, b) where id = κ_σᶜᶜᶠ(i, j, k, grid, closure, K, Val(id), clk, fields, biharmonic_mask_z, ∂³zᶜᶜᶠ, c)
```
glwagner, line 68: The abbreviation is supposed to make this easier to read on a screen that isn't huge, why do we think this isn't necessary anymore?

src/TurbulenceClosures/turbulence_closure_implementations/leith_enstrophy_diffusivity.jl
```diff
@@ -135,8 +135,8 @@ end
 
     νₑⁱʲᵏ = ℑxᶠᵃᵃ(i, j, k, grid, νₑ)
 
-    ∂x_c = ∂xᶠᶜᶜ(i, j, k, grid, c)
-    ∂z_c = ℑxzᶠᵃᶜ(i, j, k, grid, ∂zᶜᶜᶠ, c)
+    ∂x_c = ∂xᶠᶜᶜ(i, j, k, grid, tracer)
+    ∂z_c = ℑxzᶠᵃᶜ(i, j, k, grid, ∂zᶜᶜᶠ, tracer)
```
glwagner, line 139: I prefer `c`. It's easy to confuse `tracer` and `tracers`.

glwagner:
I think some of the changes here are positive. However, I don't see the point of changing `calc_νᶜᶜᶜ` to `calc_nonlinear_νᶜᶜᶜ`. I also prefer `c` to `tracer`.

src/Models/ShallowWaterModels/shallow_water_diffusion_operators.jl
tomchor: The rationale was to differentiate from `calc_νᶜᶜᶜ()` from `νᶜᶜᶜ()`, which at first seem to do the same thing. That said, I'm okay with other names. What would you suggest?

src/TurbulenceClosures/turbulence_closure_implementations/leith_enstrophy_diffusivity.jl
tomchor: That's a good point. I was mainly going for consistency with `tracers`. I'm going to change this one back.

src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
tomchor: I see your point about keeping the code short (i.e., with few columns), but they also make the code (this module more than others imo) less readable.

That said, my only screen is 13' (hard to find someone that uses a smaller screen :grin:) and the whole line still fits in one line. I can always include some line breaks if you think keeping a small column count is important!

src/TurbulenceClosures/turbulence_closure_implementations/leith_enstrophy_diffusivity.jl
tomchor: Rolled back. I kept the `tracer` instances in the AMD code though because it appears there in instances of the "`for tracer in tracers`" type, so it's clearer and there's nothing called `∂z_c`.

src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
glwagner: I think we have to keep it all on one line (to catch bugs)

glwagner: I agree that it's less readable IF it fits in one line. It's all trade-offs, this one was made intentionally...

tomchor:
> I think some of the changes here are positive. However, I don't see the point of changing `calc_νᶜᶜᶜ` to `calc_nonlinear_νᶜᶜᶜ` (it's ambiguous what "nonlinear" refers to, so I don't think the name change improves the situation --- that's not to say there isn't room for improvement). I also prefer `c` to `tracer`.

Thanks for the review. I'm relatively agnostic about what we call things. Mostly I'd just like to make this module more readable.

src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
tomchor: Okay, rolled back

src/TurbulenceClosures/turbulence_closure_implementations/leith_enstrophy_diffusivity.jl
```diff
@@ -86,9 +86,9 @@ const ArrayOrField = Union{AbstractArray, AbstractField}
     return wxz² + wyz²
 end
 
-@inline calc_νᶜᶜᶜ(i, j, k, grid, closure::TwoDimensionalLeith{FT}, buoyancy, U, C) where FT =
-    (closure.C * Δᶠ(i, j, k, grid, closure))^3 * sqrt(  abs²_∇h_ζ(i, j, k, grid, U)
-                                              + abs²_∇h_wz(i, j, k, grid, U.w))
+@inline calc_nonlinear_νᶜᶜᶜ(i, j, k, grid, closure::TwoDimensionalLeith{FT}, buoyancy, velocities, tracers) where FT =
+    (closure.C * Δᶠ(i, j, k, grid, closure))^3 * sqrt(  abs²_∇h_ζ(i, j, k, grid, velocities)
+                                              + abs²_∇h_wz(i, j, k, grid, velocities.w))
```
glwagner, line 91: Some alignment has been broken here. I think it's supposed to be

```suggestion
@inline calc_nonlinear_νᶜᶜᶜ(i, j, k, grid, closure::TwoDimensionalLeith{FT}, buoyancy, velocities, tracers) where FT =
    (closure.C * Δᶠ(i, j, k, grid, closure))^3 * sqrt(   abs²_∇h_ζ(i, j, k, grid, velocities)
                                                      + abs²_∇h_wz(i, j, k, grid, velocities.w))
```

src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl
```diff
@@ -96,15 +96,25 @@ filter width `Δᶠ`, and strain tensor dot product `Σ²`.
...
 end
 
+
+@inline function calc_nonlinear_κᶜᶜᶜ(i, j, k, grid, closure::SmagorinskyLilly, buoyancy, velocities, tracers, ::Val{tracer_index}) where {tracer_index}
+    νₑ = calc_nonlinear_νᶜᶜᶜ(i, j, k, grid, closure, buoyancy, velocities, tracers)
+
+    @inbounds Pr = closure.Pr[tracer_index]
+
+    return νₑ / Pr
+end
```
glwagner, line 115: We don't need this, because we don't call `calculate_nonlinear_tracer_diffusivity!`.

An abstract type like "`AbstractDynamicScalarDiffusivity` or `AbstractNonlinearScalarDiffusivity` (I think I like "dynamic" more since it refers to function rather than form), with a complement of functions and a default `calculate_diffusivities!` interface, might clean up this code.

src/Models/ShallowWaterModels/shallow_water_diffusion_operators.jl
glwagner: Do we even need separate `calc_νᶜᶜᶜ()` and `νᶜᶜᶜ()` if they do the same thing?

src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl
tomchor: Agreed.

tomchor:
@glwagner this almost became stale but I think it's ready to review and possibly merge. It doesn't re-work the code like you suggest [here](https://github.com/CliMA/Oceananigans.jl/pull/2752/#issuecomment-1315578507), but it does make the functions more easily understandable on first pass by being more verbose. 

It also changes the behavior of `viscosity()` and `diffusivity()`, which no longer sums over all individual closures by default when the closure is a tuple, avoiding user error (as you suggested [here](https://github.com/CliMA/Oceananigans.jl/pull/2752/#issuecomment-1273620776)).

Finally it also changes the function `calc_κᶜᶜᶜ` ro `calc_nonlinear_κᶜᶜᶜ` in order to differentiate from `κᶜᶜᶜ`. Although I'm agnostic about the name and could change it to anything else.

This PR used to also remove a fallback that was a bit problematic because it silently returned zero diffusivity values for closures like Smagorinsky, but it seems some other PR got around to that first before merging this one :)

glwagner:
Ok, I will review! Except, tests are failing? Also should we merge main?

tomchor:
> Ok, I will review! Except, tests are failing? Also should we merge main?

Thanks!

And yeah there was a typo in my last commit but it should be fixed as tests should be passing. I also merged main in commit [479056a](https://github.com/CliMA/Oceananigans.jl/pull/2752/commits/479056a24db8e510ba057ff79df82f567f85c03a) (yesterday) so we should be good regarding that.

tomchor:
@glwagner  Tests are passing now!

glwagner (approved):
Looks good!

PR #2753, opened 2022-09-26
Add multigrid preconditioned for the heptadiagonal solver

elise-palethorpe:
Implement multigrid preconditioning for the heptadiagonal solver.

cc @navidcy


navidcy:
Is it working? Did you try to bench?

elise-palethorpe:
On GPU yes! On CPU no difference - not sure why yet.

mgcg_cpu
mgcg_gpu

Sorry about the unpleasantly coloured axes - they wouldn't show up in black.

navidcy:
> On GPU yes! On CPU no difference - not sure why yet.
> 
> mgcg_cpu mgcg_gpu
> 

And `mgcg` is wrong label I suppose? It's the mg-preconditioned hepta?

elise-palethorpe:
> And `mgcg` is wrong label I suppose? It's the mg-preconditioned hepta?

Yes though the heptadiagonal solver was using the cg algorithm so it is also mgcg just not using the `PreconditionedConjugateGradientSolver`.

simone-silvestri:
Nice results!

simone-silvestri:
Is `N` the resolution? So, indeed, using a multigrid preconditioner solves our high-resolution problem?

navidcy:
Yeap, N is horizontal resolution. These are benchmarks for the baroclinic adjustment problem with hydrostatic free-surface model. Elise wants to run now some 1/4-degree near-global benches.

What's the "high-resolution problem"?

simone-silvestri:
The preconditioner we used at high resolution is an asymptotic approximation of the inverse matrix that assumes that the original matrix is diagonally dominant. This is an increasingly worst approximation as you increase the resolution, deteriorating the performance when cells size is decreased

elise-palethorpe:
Thanks Simone! So I guess [this line](https://github.com/CliMA/Oceananigans.jl/blob/edef0fc49a1019bfba093ffa1c0bcce23535ac41/src/Solvers/sparse_preconditioners.jl#L70) explains why we only see a performance increase on GPU? I don't understand why ILU is hard to parallelise - would you mind explaining this?

elise-palethorpe:
Thanks Simone! So I guess [this line](https://github.com/CliMA/Oceananigans.jl/blob/edef0fc49a1019bfba093ffa1c0bcce23535ac41/src/Solvers/sparse_preconditioners.jl#L70) explains why we only see a performance increase on GPU? I don't understand why ILU is hard to parallelise - would you mind explaining this?

simone-silvestri:
the problem with `ILU` is that it requires a solution to a triangular matrix system at each iteration, which is very difficult to parallelize (it requires a gaussian elimination and backward substitution step which are inherently serial). If instead, we assume that the preconditioner is an approximation of the inverse matrix of the linear system, we can just multiply the preconditioning matrix to the residual to precondition the solver. Matrix multiplications are very fast on the GPU, triangular solves are very slow (you can try doing `A\b` on CPU and GPU). So despite needing a little bit more iteration it definitely improves the time to solution when compared to a full triangular solve done every iteration.

simone-silvestri:
We have a better preconditioner that the `:ApproximateInverse` which is the `:SparseInverse`. The problem with the latter is that (probably I implemented in a stuping way) it requires an iteration on a column-wise operation (a little bit like the `linear_operation!` applied to each column), so for high-resolution problems (where we would need it and the mesh is large), it is unfeasible to use.

elise-palethorpe:
Thank you! That is very clear. Could you also please point me in the direction of a resourse to understand the asymptotic diagonal approximation of the inverse?

src/Models/HydrostaticFreeSurfaceModels/matrix_implicit_free_surface_solver.jl
```diff
@@ -139,3 +139,5 @@ end
         diag[i, j, 1]  = - Azᶜᶜᶜ(i, j, 1, grid) / g
     end
 end
+
+@ifhasamgx finalize_solver!(s::MatrixImplicitFreeSurfaceSolver) = finalize_solver!(s.matrix_iterative_solver)
```
navidcy, line 143:
```suggestion
@ifhasamgx finalize_solver!(s::MatrixImplicitFreeSurfaceSolver) = finalize_solver!(s.matrix_iterative_solver)

```

benchmark/benchmark_implicit_surface_solvers.jl
```diff
@@ -3,124 +3,132 @@ using Oceananigans.Units
...
+        @info "Benchmark with $implicit_free_surface_solver free surface implicit solver:"
+        @timeit to "$implicit_free_surface_solver and N=$N" run!(simulation)
 
-    finalize_solver!(model.free_surface.implicit_step_solver)
+        finalize_solver!(model.free_surface.implicit_step_solver)
+    end
+    # show(to)
 end
+
+finalize_AMGX(arch)
```
navidcy, line 134:
```suggestion
finalize_AMGX(arch)

```

src/Solvers/multigrid_solver.jl
```diff
@@ -164,22 +164,30 @@ end
...
+                              amgx_solver
+                              )
+end
+
+@ifhasamgx function AMGXMultigridSolver(matrix::CuSparseMatrixCSC, maxiter = 1, reltol = sqrt(eps(eltype(matrix))), abstol = 0)
+    tolerance, convergence = reltol == 0 ? (abstol, "ABSOLUTE") : (reltol, "RELATIVE_INI_CORE")
     try
+        global config = AMGX.Config(Dict("monitor_residual" => 1, "max_iters" => maxiter, "store_res_history" => 1, "tolerance" => tolerance, "convergence" => convergence))
+    catch e 
+        @info "It appears you are using the multigrid solver on GPU. Have you called `initialize_AMGX`?"
```
navidcy, line 187:
```suggestion
        @info "It appears you are using the multigrid solver on GPU. Have you called `initialize_AMGX()`?"
```

src/Solvers/multigrid_solver.jl
```diff
@@ -307,29 +304,45 @@ end
...
     AMGX.copy!(solver.x_array, s.device_x)
 
     interior(x) .= reshape(solver.x_array, Nx, Ny, Nz)
 end
 
-@ifhasamgx function finalize_solver!(solver::MultigridGPUSolver)
-    @info "Finalizing the Multigrid solver on GPU"
-    
-    s = solver.amgx_solver
+initialize_AMGX(::CPU) = nothing
```
navidcy, line 313: Perhaps add docstrings for `initialize_AMGX` and `finalize_AMGX`? Also, are these exported?

src/Solvers/sparse_preconditioners.jl
```diff
@@ -174,3 +176,36 @@ function sparse_inverse_preconditioner(A::AbstractMatrix; ε, nzrel)
...
+    AMGX.copy!(x, p.amgx_solver.device_x)
+end
+
+@ifhasamgx mul!(b, p::MultigridGPUPreconditioner, x) = mul!(b, p.amgx_solver.csr_matrix, x)
+
+@ifhasamgx function \(p::MultigridGPUPreconditioner, b)
+    ldiv!(similar(b), p, b)
+end
+
+@ifhasamgx finalize_solver!(p::MultigridGPUPreconditioner) = finalize_solver!(p.amgx_solver)
```
navidcy, line 211:
```suggestion
@ifhasamgx finalize_solver!(p::MultigridGPUPreconditioner) = finalize_solver!(p.amgx_solver)

```

src/Solvers/sparse_preconditioners.jl
```diff
@@ -174,3 +176,36 @@ function sparse_inverse_preconditioner(A::AbstractMatrix; ε, nzrel)
    Minv = arch_sparse_matrix(architecture(A), Minv_cpu)
    return SparseInversePreconditioner(Minv)
 end
+
+function multigrid_preconditioner(A::AbstractMatrix;)
+    ml = create_multilevel(RugeStubenAMG(), A)
+    return aspreconditioner(ml)
+end
```
navidcy, line 183:
```suggestion
multigrid_preconditioner(A::AbstractMatrix) = aspreconditioner(create_multilevel(RugeStubenAMG(), A))
```

src/Solvers/sparse_preconditioners.jl
```diff
@@ -174,3 +176,36 @@ function sparse_inverse_preconditioner(A::AbstractMatrix; ε, nzrel)
...
+
+@ifhasamgx function multigrid_preconditioner(A::CuSparseMatrixCSC;)
+    return MultigridGPUPreconditioner(AMGXMultigridSolver(A))
+end
```
navidcy, line 187:
```suggestion
@ifhasamgx multigrid_preconditioner(A::CuSparseMatrixCSC) = MultigridGPUPreconditioner(AMGXMultigridSolver(A))
```

src/Solvers/sparse_preconditioners.jl
navidcy, line 211:
```suggestion
@ifhasamgx finalize_solver!(p::MultigridGPUPreconditioner) = finalize_solver!(p.amgx_solver)

```

navidcy:
Nice job!

I'm wondering why the mg preconditioner seems to have almost no effect on the heptadiagonal solver on CPU.

elise-palethorpe:
> I'm wondering why the mg preconditioner seems to have almost no effect on the heptadiagonal solver on CPU.

I think this just means the multigrid preconditioning is very similar to the ILU preconditioning that is the default for the heptadigagonal solver on CPU. I believe it would outperform a non preconditioned conjugate gradient solver, perhaps I should also run those tests?

navidcy:
Oh I see. The Heptadiagonal solver has a default preconditioner. So if you give it the identity preconditioner and compare with mg the mg should be better, right?

src/Solvers/multigrid_solver.jl
elise-palethorpe: Sounds good. I don't really understand exporting - why would I export something if it can already be called from the `benchmark` directory?

elise-palethorpe:
Indeed! I will confirm that now :)

simone-silvestri:
> Thank you! That is very clear. Could you also please point me in the direction of a resourse to understand the asymptotic diagonal approximation of the inverse?

It is not well documented actually. You can find some (albeit simple) explanation here https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/96JC02775. There is also an explanation in the docstrings

src/Solvers/multigrid_solver.jl
navidcy: Export means that the method is available in the main Oceananigans namespace. E.g., at the moment

```Julia
julia> using Oceananigans

julia> initialize_AMGX()
ERROR: UndefVarError: initialize_AMGX not defined
Stacktrace:
 [1] top-level scope
   @ REPL[2]:1
 [2] top-level scope
   @ ~/.julia/packages/CUDA/DfvRa/src/initialization.jl:52
```

But perhaps this is specialized enough so we can leave it _hidden_, and only available when 

```Julia
julia> using Oceananigans

julia> using Oceananigans.Solvers: initialize_AMGX

julia> initialize_AMGX(CPU())
```

navidcy (approved):
Looks good to me. @simone-silvestri can you also have a look?

src/Solvers/sparse_preconditioners.jl
```diff
@@ -174,3 +176,31 @@ function sparse_inverse_preconditioner(A::AbstractMatrix; ε, nzrel)
...
+    AMGX.copy!(x, p.amgx_solver.device_x)
+end
+
+@ifhasamgx mul!(b, p::MultigridGPUPreconditioner, x) = mul!(b, p.amgx_solver.csr_matrix, x)
+
+@ifhasamgx function \(p::MultigridGPUPreconditioner, b)
+    ldiv!(similar(b), p, b)
+end
+
+@ifhasamgx finalize_solver!(p::MultigridGPUPreconditioner) = finalize_solver!(p.amgx_solver)
```
navidcy, line 206:
```suggestion
@ifhasamgx begin
    ldiv!(p::MultigridGPUPreconditioner, b) = copyto!(b, p \ b)

    function ldiv!(x, p::MultigridGPUPreconditioner, b)
        x .= 0
        AMGX.upload!(p.amgx_solver.device_b, b)
        AMGX.upload!(p.amgx_solver.device_x, x)
        AMGX.solve!(p.amgx_solver.device_x, p.amgx_solver.solver, p.amgx_solver.device_b)
        AMGX.copy!(x, p.amgx_solver.device_x)
    end

    mul!(b, p::MultigridGPUPreconditioner, x) = mul!(b, p.amgx_solver.csr_matrix, x)

    \(p::MultigridGPUPreconditioner, b) = ldiv!(similar(b), p, b)
    
    finalize_solver!(p::MultigridGPUPreconditioner) = finalize_solver!(p.amgx_solver)
end
```

navidcy: Apparently you can include multiline expressions in a custom made macro. I tested it and it works. Make things a bit cleaner.

src/Solvers/sparse_preconditioners.jl
```diff
@@ -174,3 +176,31 @@ function sparse_inverse_preconditioner(A::AbstractMatrix; ε, nzrel)
    Minv = arch_sparse_matrix(architecture(A), Minv_cpu)
    return SparseInversePreconditioner(Minv)
 end
+
+multigrid_preconditioner(A::AbstractMatrix) = aspreconditioner(create_multilevel(RugeStubenAMG(), A))
```
navidcy, line 180: Would something like that work? Defaults to `RugeStubenAMG()` but allows for one to use `SmoothedAggregationAMG()` if they'd like?
 
```suggestion
multigrid_preconditioner(A::AbstractMatrix; method = RugeStubenAMG()) = aspreconditioner(create_multilevel(method, A))
```

src/Solvers/sparse_preconditioners.jl
```diff
@@ -74,6 +75,7 @@ end
 build_preconditioner(::Val{nothing},            A, settings)  = Identity()
 build_preconditioner(::Val{:SparseInverse},     A, settings)  = sparse_inverse_preconditioner(A, ε = settings.ε, nzrel = settings.nzrel)
 build_preconditioner(::Val{:AsymptoticInverse}, A, settings)  = asymptotic_diagonal_inverse_preconditioner(A, asymptotic_order = settings.order)
+build_preconditioner(::Val{:Multigrid},         A, settings)  = multigrid_preconditioner(A;)
```
navidcy, line 78:
```suggestion
build_preconditioner(::Val{:Multigrid},         A, settings)  = multigrid_preconditioner(A)
```

simone-silvestri:
If you have the benches for a global model post them here, that would be interesting to see

simone-silvestri (approved):

navidcy:
> If you have the benches for a global model post them here, that would be interesting to see

Soon! @elise-palethorpe is onto it

src/Solvers/multigrid_solver.jl
```diff
@@ -310,9 +310,11 @@ end
     interior(x) .= reshape(solver.x_array, Nx, Ny, Nz)
 end
 
-initialize_AMGX(::CPU) = nothing
-finalize_AMGX(::CPU) = nothing
-
+"""
+    initialize_AMGX(arch)
+Initializes the AMGX package required to use the multigrid solver on GPU. 
+This function should be called before creating a multigrid solver on GPU.
```
navidcy, line 316:
```suggestion
    initialize_AMGX(architecture)
    
Initialize the AMGX package required to use the multigrid solver on `architecture`. 
This function needs to be called before creating a multigrid solver on GPU.
```

src/Solvers/multigrid_solver.jl
```diff
@@ -325,10 +327,20 @@ function initialize_AMGX(::GPU)
     end
 end
 
+initialize_AMGX(::CPU) = nothing
+
+"""
+    finalize_AMGX(arch)
+Finalizes the AMGX package required to use the multigrid solver on GPU. 
+This should be called after `finalize_solver!`.
```
navidcy, line 335:
```suggestion
    finalize_AMGX(architecture)

Finalize the AMGX package required to use the multigrid solver on `architecture`. 
This should be called after `finalize_solver!`.
```

glwagner:
Ultimately we want to merge the heptadiagonal solver and pcg solver, right @simone-silvestri ?

simone-silvestri:
Yep! That's the idea.

glwagner:
Does this PR make progress towards that? I just don't want work to be wasted; eg work spent adding features to heptadiagonal solver (or the other way around?)

Issue #2709, opened 2022-08-23
Can't use `GradientBoundaryCondition` on the GPU


glwagner:
Using Julia 1.7.3 and Oceananigans 0.77.1, this code:

```julia
using Oceananigans

grid = RectilinearGrid(GPU(), size=(1, 1, 2), extent=(1, 1, 1))
buoyancy_bottom_bc = GradientBoundaryCondition(1e-5)
buoyancy_bcs = FieldBoundaryConditions(bottom=buoyancy_bottom_bc)
model = NonhydrostaticModel(; grid, tracers=:b, buoyancy=BuoyancyTracer(),
                            boundary_conditions=(; b=buoyancy_bcs))
simulation = Simulation(model; Δt=1, stop_iteration=2)
run!(simulation)
```

produces:

```julia
julia> include("simple_model.jl")
[...]

PR #2754, opened 2022-09-28
Remove tupled halo filling

simone-silvestri:
this PR removes tupled halo filling for the moment because of a problem with differently typed boundary conditions fed as a Tuple to the GPU.

See issue #2709


glwagner:
Does this reduce performance or is the effect negligible? (Just curious.)

simone-silvestri:
looking at the benchmarks in #2335 it seems that it does. But only for some configurations

simone-silvestri:
Actually, since I can reproduce neither #2709 nor #2709, I believe that maybe the problem lies in either the `CUDA` or the `KernelAbstraction` package. We might be able to solve it just by upgrading

glwagner:
It's very plausible that it depends on CUDA and also Julia version.

glwagner (approved):

simone-silvestri:
I think the issue is solved by upgrading CUDA and julia, should I close this?

glwagner:
Yeah, let's close this.

PR #2760, opened 2022-09-30
High resolution multi GPU global simulation

simone-silvestri:
I am currently running a 0.083-degree global ocean simulation on Tesla V100 GPUs. I am benchmarking a bit to understand where we can target optimization to improve the performance and I thought I would leave the benchmarks here to document them for later use

(we do not have to merge this PR)


navidcy:
> I am currently running a 0.083-degree global ocean simulation

I needed to invert that to understand that is 1/12-degree :)

simone-silvestri:
I will actually close this PR and write a section in the benchmarks documentation

PR #2771, opened 2022-10-04
Fix output of sliced fields

simone-silvestri:
This fixes issue #2770 but might introduce other problems.

@glwagner, do you know why we restrict to the interior the indices in case of `(:, :, :)` can we do it without it an call `interior` when outputting the data?

### DIFF
--- a/src/OutputWriters/output_construction.jl
+++ b/src/OutputWriters/output_construction.jl
@@ -2,7 +2,7 @@ using Oceananigans.Fields: validate_indices, Reduction
 using Oceananigans.AbstractOperations: AbstractOperation, ComputedField
 using Oceananigans.Grids: default_indices
 
-restrict_to_interior(::Colon, loc, topo, N) = interior_indices(loc, topo, N)
+restrict_to_interior(::Colon, loc, topo, N) = Colon()
 restrict_to_interior(::Colon, ::Type{Nothing}, topo, N) = UnitRange(1, 1)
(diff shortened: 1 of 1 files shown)


simone-silvestri:
There is a larger problem with `window_data` that looks at `parent(data)`. The latter does not preserve the offsetted indices of a sliced field...

glwagner:
> This fixes issue #2770 but might introduce other problems.
> 
> 
> 
> @glwagner, do you know why we restrict to the interior the indices in case of `(:, :, :)`? 
> 
> Can we do it without it and call `interior` when outputting the data?

I think the purpose of "restrict to interior" is to take any indices and restrict them to the interior of the domain (eg discard halos).

What do you propose we do differently?

I don't totally understand the output question. But, we can't only use interior for output because we support outputting data with halos. In fact, including halos in output should become default when we feel post processing via FieldTimeSeries is mature enough. It's better to preserve halo regions (most importantly for problems with Value/Gradient boundary conditions, but also for ensuring that post processing calculations, derivatives, and interpolations are fully featured).

simone-silvestri:
I see. I was thinking that we could solve the issue with the sliced fields' outputwriter in that simple way, but it turns out it is a deeper problem since `view(field, indices(field)...)` does not work for sliced fields.

glwagner:
There appears to be a user API issue. There could also be a bug but I'm not sure. I'll summarize on #2770.

glwagner:
should we close this?

glwagner:
closing but re-open if needed

Issue #2772, opened 2022-10-06
State callbacks


jagoosw:
Hi all,

Following discussion with @glwagner we think it would be useful to implement a callback like construction between each sub step. This would allow users to modify tendency fields externally to normal forcing functions.

My use case for this is having Lagrangian particles which can effect tracers, for example if the particle is some organism that takes up one tracer and excretes another. We discussed a different possible implementation of this in #2724 but came to the conclusion that its much better to do it a different way but the [example](https://gist.github.com/jagoosw/213d7f2ce3707768d4ddb07f8827cdd2) I came up with there possibly explains this case better.
![Example animation](https://user-images.githubusercontent.com/26657828/188803674-9dfa61a5-2de8-4612-b8db-9c8918a9a0fb.gif)

Callbacks are implemented in `run!`:
https://github.com/CliMA/Oceananigans.jl/blob/b288527ccf34fc17ec4b31a1e9da7bd8b8d25a10/src/Simulations/run.jl#L131

And this could probably be implemented fairly easily by:
- Adding a property to `Simulation` or, probably better to the model structs since the time steppers are only passed the `model` not `sim` like `run!` is
- In the time stepper functions we can then add `[callback(sim) for callback in values(model.state_callbacks)]` either just after `calculate_tendencies!` in each substep
https://github.com/CliMA/Oceananigans.jl/blob/b288527ccf34fc17ec4b31a1e9da7bd8b8d25a10/src/TimeSteppers/runge_kutta_3.jl#L98-L115
or at the end of each `calculate_tendencies!` function
https://github.com/CliMA/Oceananigans.jl/blob/b288527ccf34fc17ec4b31a1e9da7bd8b8d25a10/src/Models/NonhydrostaticModels/calculate_nonhydrostatic_tendencies.jl#L12-L36

navidcy:
Was this closed by #2773?

glwagner:
Yes

PR #2773, opened 2022-10-07
State callbacks

jagoosw:
Implements callbacks between sub-steps as described in #2772 

I chose to add them to the model since that is the only part of the simulation they have access to when they run, and went with using the same `Callback` constructor so that parameters can still be passed to the functions despite the fact that whatever the specified schedule they have to run at every sub-step.

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
@@ -27,7 +27,7 @@ validate_tracer_advection(tracer_advection::AbstractAdvectionScheme, grid) = tra
 PressureField(grid) = (; pHY′ = CenterField(grid))
 
 mutable struct HydrostaticFreeSurfaceModel{TS, E, A<:AbstractArchitecture, S,
-                                           G, T, V, B, R, F, P, U, C, Φ, K, AF} <: AbstractModel{TS}
+                                           G, T, V, B, R, F, P, U, C, Φ, K, AF, SC} <: AbstractModel{TS}
   
           architecture :: A        # Computer `Architecture` on which `Model` is run
                   grid :: G        # Grid of physical points on which `Model` is solved
@@ -45,6 +45,7 @@ mutable struct HydrostaticFreeSurfaceModel{TS, E, A<:AbstractArchitecture, S,
     diffusivity_fields :: K        # Container for turbulent diffusivities
            timestepper :: TS       # Object containing timestepper fields and parameters
       auxiliary_fields :: AF       # User-specified auxiliary fields for forcing functions and boundary conditions
+      state_callbacks :: SC     # Callbacks called between each substep
 end
 
 """
@@ -182,9 +183,13 @@ function HydrostaticFreeSurfaceModel(; grid,
 
     advection = merge((momentum=momentum_advection,), tracer_advection_tuple)
 
+    # State callbacks - function called on sim between each substep
+    state_callbacks = OrderedDict{Symbol, Callback}()
+
     model = HydrostaticFreeSurfaceModel(arch, grid, clock, advection, buoyancy, coriolis,
                                         free_surface, forcing, closure, particles, velocities, tracers,
-                                        pressure, diffusivity_fields, timestepper, auxiliary_fields)
+                                        pressure, diffusivity_fields, timestepper, auxiliary_fields,
+                                        state_callbacks)
 
     update_state!(model)
 
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
@@ -27,7 +27,7 @@ validate_tracer_advection(tracer_advection::AbstractAdvectionScheme, grid) = tra
 PressureField(grid) = (; pHY′ = CenterField(grid))
 
 mutable struct HydrostaticFreeSurfaceModel{TS, E, A<:AbstractArchitecture, S,
-                                           G, T, V, B, R, F, P, U, C, Φ, K, AF, SC} <: AbstractModel{TS}
+                                           G, T, V, B, R, F, P, U, C, Φ, K, AF} <: AbstractModel{TS}
   
           architecture :: A        # Computer `Architecture` on which `Model` is run
                   grid :: G        # Grid of physical points on which `Model` is solved
@@ -45,7 +45,6 @@ mutable struct HydrostaticFreeSurfaceModel{TS, E, A<:AbstractArchitecture, S,
     diffusivity_fields :: K        # Container for turbulent diffusivities
            timestepper :: TS       # Object containing timestepper fields and parameters
       auxiliary_fields :: AF       # User-specified auxiliary fields for forcing functions and boundary conditions
-       state_callbacks :: SC     # Callbacks called between each substep
 end
 
 """
@@ -183,13 +182,9 @@ function HydrostaticFreeSurfaceModel(; grid,
 
     advection = merge((momentum=momentum_advection,), tracer_advection_tuple)
 
-    # State callbacks - function called on sim between each substep
-    state_callbacks = OrderedDict{Symbol, Any}()
-
     model = HydrostaticFreeSurfaceModel(arch, grid, clock, advection, buoyancy, coriolis,
                                         free_surface, forcing, closure, particles, velocities, tracers,
-                                        pressure, diffusivity_fields, timestepper, auxiliary_fields,
-                                        state_callbacks)
+                                        pressure, diffusivity_fields, timestepper, auxiliary_fields)
 
     update_state!(model)
 
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
@@ -45,7 +45,7 @@ mutable struct HydrostaticFreeSurfaceModel{TS, E, A<:AbstractArchitecture, S,
     diffusivity_fields :: K        # Container for turbulent diffusivities
            timestepper :: TS       # Object containing timestepper fields and parameters
       auxiliary_fields :: AF       # User-specified auxiliary fields for forcing functions and boundary conditions
-      state_callbacks :: SC     # Callbacks called between each substep
+       state_callbacks :: SC     # Callbacks called between each substep
 end
 
 """
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
@@ -184,7 +184,7 @@ function HydrostaticFreeSurfaceModel(; grid,
     advection = merge((momentum=momentum_advection,), tracer_advection_tuple)
 
     # State callbacks - function called on sim between each substep
-    state_callbacks = OrderedDict{Symbol, Callback}()
+    state_callbacks = OrderedDict{Symbol, Any}()
 
     model = HydrostaticFreeSurfaceModel(arch, grid, clock, advection, buoyancy, coriolis,
                                         free_surface, forcing, closure, particles, velocities, tracers,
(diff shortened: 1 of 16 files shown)


jagoosw:
I can't see where/if callbacks are tested to add tests for these?

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -45,6 +45,7 @@ mutable struct HydrostaticFreeSurfaceModel{TS, E, A<:AbstractArchitecture, S,
     diffusivity_fields :: K        # Container for turbulent diffusivities
            timestepper :: TS       # Object containing timestepper fields and parameters
       auxiliary_fields :: AF       # User-specified auxiliary fields for forcing functions and boundary conditions
+      state_callbacks :: SC     # Callbacks called between each substep
```
glwagner, line 48:
```suggestion
       state_callbacks :: SC     # Callbacks called between each substep
```

src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -51,6 +52,7 @@ mutable struct ShallowWaterModel{G, A<:AbstractArchitecture, T, V, U, R, F, E, B
             diffusivity_fields :: K         # Container for turbulent diffusivities
                    timestepper :: TS        # Object containing timestepper fields and parameters
                    formulation :: FR        # Either conservative or vector-invariant
+                   state_callbacks :: SC     # Callbacks called between each substep
```
glwagner, line 55: alignment issue

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -46,6 +47,7 @@ mutable struct NonhydrostaticModel{TS, E, A<:AbstractArchitecture, G, T, B, R, S
       pressure_solver :: S        # Pressure/Poisson solver
     immersed_boundary :: I        # Models the physics of immersed boundaries within the grid
      auxiliary_fields :: AF       # User-specified auxiliary fields for forcing functions and boundary conditions
+     state_callbacks :: SC     # Callbacks called between each substep
```
glwagner, line 50:
```suggestion
      state_callbacks :: SC     # Callbacks called between each substep
```

glwagner:
What do we think about the pros and cons of an alternative design that adds `state_callbacks` to `Simulation` and extends `time_step(model, dt, state_callbacks)` to accept the state callbacks as a positional argument?

It might be slightly less code and we don't have to bloat the model structs anymore. It also keeps the callback implementations in `Simulation`.

jagoosw:
My concern with associating it with the simulation was that, unless we pass the simulation to `time_step!` as well, the callback can't access it which might be confusing since normal callbacks are `func(sim)`. So I thought it may be clearer if callbacks associated with the model were `func(model)` instead. Not sure what the best solution is though?

glwagner:
One aspect of this change is that the models will no longer be concretely typed (because `OrderedDict{Symbol, Callback}` is not a concrete type). I'm not sure this matters, but it's worth noting this.

Thinking about the API issues, I wonder if there's another advantage in putting the state callbacks in `Simulation`, that we have a system in which a "model" is not modified after its created (eg we are not adding callbacks to it) --- that's something that we do with Simulations. With this PR, we mix the concepts of models and simulations. Which may be ok...

@simone-silvestri @navidcy what do you think?

jagoosw:
> One aspect of this change is that the models will no longer be concretely typed (because `OrderedDict{Symbol, Callback}` is not a concrete type). I'm not sure this matters, but it's worth noting this.
> 
> Thinking about the API issues, I wonder if there's another advantage in putting the state callbacks in `Simulation`, that we have a system in which a "model" is not modified after its created (eg we are not adding callbacks to it) --- that's something that we do with Simulations. With this PR, we mix the concepts of models and simulations. Which may be ok...

Perhaps we could change it so you give the model a Tuple or NamedTuple of callbacks when you specify it so the API is more consistent and the model not modified after? I think it could be argued that state callbacks are part of the model rather than the simulation, like forcing functions are, since they could be an intrinsic part of how the model evolves, e.g. if it modified the tendencies. I suppose this argument could be made for some callbacks as well though so I don't know whats best!

glwagner:
@navidcy @simone-silvestri any thoughts?

I thought of a few more considerations:

It may be easier to document this feature if it's in `Simulation`, because it can be explained in conjunction with "normal" `Callback`. Ie, "Here are different flavors of callback users can employ to alter their simulation..."

An even more radical suggestion is to actually combine this feature with `Simulation.callbacks`. In other words, we don't nee d a separate _list_, but rather an additional identifier that says _where and how_ the callback will be employed. Right now a `Callback` has the properties `func`, `parameters`, and `schedule`. We could add a fourth property, something like `flavor`, which would be:

* `flavor=StepCallback()` when called every time-step / iteration
* `flavor=StateCallback()` when called inside `update_state()`
* and more.

navidcy:
I'd also intuitively prefer if this is part of simulation.

simone-silvestri:
I like the idea of having the call-time (before, after, or in between the time stepping) as a property of the `Callback`. 

With regards to where to position the callbacks, having the callback as a part of the simulation is probably the cleanest but that would mean passing the simulation to the time step, which is not ideal. 

Do we need to support that accurate callback (happening at each RK substep)? For example, for particle tracking we only support Euler stepping, if we don't have to support RK for callbacks we can leave them outside the time stepping function

jagoosw:
I wonder if this would be better proposed as a custom dynamics option like with particles? 

I'm just concerned that you could implement some dynamics like this that are a core part of the model dynamics rather than a simulation of that model. For example, if you had a computationally intensive source/sink term, instead of recalculating for the tracer it's going to and from you could add a custom dynamic that calculates it once, takes from one and adds to the other (this is how PISCES implements a lot of the growth to prevent repetitive calculation of the same values, I don't know if this would be faster in this setting without testing but just as an example).

glwagner:
> but that would mean passing the simulation to the time step, which is not ideal.

We can create a schema where the argument to `Callback` depends on "flavor" / "callsite", etc, similar to how we have different "forms" for boundary conditions and forcing functions.

> Do we need to support that accurate callback (happening at each RK substep)?

I guess as @jagoosw argues this is an important feature to support for advanced applications. There are a few different "intra time step" callsites we might envision (modifications to `calculate_tendencies!` differ from modifications to `update_state!`).

glwagner:
> (this is how PISCES implements a lot of the growth to prevent repetitive calculation of the same values, I don't know if this would be faster in this setting without testing but just as an example).

I agree that we need this feature, but from a software design perspective one might argue that such an invasive change to model dynamics belongs in the source code itself, rather than a callback (and this of course how it works for PISCES, or any model other than Oceananigans --- since Oceananigans is the only model where we can envision such a fine-grained callback schema).

I think having a callback infrastructure like this effectively allows people who implement advanced / custom applications to keep their code up to date with other changes to Oceananigans. The alternative model, which is simply to create a new version of the source code for your custom application, leads to much more code duplication. So you could say that the callback feature could be important for sharing advanced custom modifications to Oceananigans, without creating a terrible tangle in `main`.

jagoosw:
I've had a go at moving it to the `Simulation` and very happy to try different things until we find whatever is best.

I've added a property to callbacks called substep which is by default false but if true then the callback is passed to `time_step!(model, \Delta t; callbacks)`, I made this an optional argument in case it is called elsewhere (like in some of the tests).

jagoosw:
In case its of use this is the code I've been checking it with:
```
using Oceananigans
grid = RectilinearGrid(size=(1, 1, 1), extent = (1, 1, 1))
simulation = Simulation(model, Δt = 1, stop_iteration = 3)
test_sim(sim) = @info "Sim? $(isa(sim, Oceananigans.Simulation))"
test_model(model) = @info "Model? $(isa(model, Oceananigans.AbstractModel))"
simulation.callbacks[:inter] = Callback(test_sim)
simulation.callbacks[:intra] = Callback(test_model; callsite=TendencyCallback)
run!(simulation)
```
And everything should return `true`

glwagner:
Can the `callsite` be instantiated? Also I think the name of the types should correspond to the callsite, like this:

```julia
Callback(func, callsite=TendencyCallsite())
Callback(func, callsite=TimeStepCallsite()) # default
```

jagoosw:
> Can the `callsite` be instantiated? Also I think the name of the types should correspond to the callsite, like this:
> 
> ```julia
> Callback(func, callsite=TendencyCallsite())
> Callback(func, callsite=TimeStepCallsite()) # default
> ```

Okay set it up like this now and will add a test

glwagner:
@jagoosw tests pass --- ok if I merge?

glwagner (approved):

jagoosw:
> @jagoosw tests pass --- ok if I merge?

Yep, all good on my end!

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
```diff
@@ -17,7 +17,7 @@ hydrostatic pressure) to the current model state.
 """
 update_state!(model::HydrostaticFreeSurfaceModel) = update_state!(model, model.grid)
 
-function update_state!(model::HydrostaticFreeSurfaceModel, grid)
+function update_state!(model::HydrostaticFreeSurfaceModel, callbacks=[])
```
navidcy, line 20: why this didn't become 

```Julia
update_state!(model::HydrostaticFreeSurfaceModel, grid, callbacks=[])
```

jagoosw: I must have done this by mistake, looking now I don't think `grid` gets used anywhere in that function though?

navidcy: It’s used to differentiate a method for when the grid is single-column. See 

https://github.com/CliMA/Oceananigans.jl/blob/a86ef32581192d82a655e1b2ff4411e79917a379/src/Models/HydrostaticFreeSurfaceModels/single_column_model_mode.jl#L50

I believe I fixed it in #2786

glwagner: I believe we also need

```julia
update_state!(model::HydrostaticFreeSurfaceModel, callbacks=[]) = update_state!(model, model.grid, callbacks)
```

johnryantaylor:
@jagoosw, @glwagner: I just noticed that the documentation page on Callbacks refers to model.state_callbacks which I think is now depreciated, based on this discussion.

jagoosw:
The whole callback page is a bit out of date, one of the example snippets is also erroring.

glwagner:
Where is the documentation on callbacks? I can't find it.

navidcy:
> @jagoosw, @glwagner: I just noticed that the documentation page on Callbacks refers to model.state_callbacks which I think is now depreciated, based on this discussion.

@johnryantaylor is this from the Oceananigans Docs somewhere? Are you sure it's the latest or `dev` version of the docs you are looking at? (Sometimes if you google "Oceananigans docs" google returns to you an outdated version...)

glwagner:
I don't think we have any documentation about callbacks in Oceananigans. We definitely should though (there's quite a bit that is missing or stale right now). The docs need attention...

jagoosw:
Is it not from here: https://clima.github.io/OceananigansDocumentation/stable/model_setup/callbacks/

navidcy:
> Is it not from here: https://clima.github.io/OceananigansDocumentation/stable/model_setup/callbacks/

I couldn't find my way to here! I don't see this page appearing in the docs-table of contents on the left side of the page. How did you figure out this link?

glwagner:
Hmm, yes I think that is an orphan page in the docs with no link...

johnryantaylor:
Yes, that is the page that I saw.  I got there by searching the docs, not using the index.

navidcy:
> Yes, that is the page that I saw.  I got there by searching the docs, not using the index.
> […](#)

Good to know! So if we want some page to disappear we need to delete the `.md` file, not just remove it from the index tree...

I opened [an issue](#3275) so we can discuss about updating the page or deleting it completely or what not.

PR #2701, opened 2022-08-18
Bump Julia to v1.8

navidcy:

PR #2775, opened 2022-10-12
Update dependencies

navidcy:


simone-silvestri:
Can we update also cuda to the 3.12 version?

navidcy:
It needs julia 1.8 I think. 
#2701

PR #2778, opened 2022-10-14
Reverting changes done by #2740

simone-silvestri:
Changes in PR #2740 made the global simulation crash. These changes will be reverted for the moment. They will be reintroduced after more testing


glwagner (approved):

simone-silvestri:
The problems seen in the global simulation were actually fixed by #2774. The test was performed without updating the code to the latest version. I will close this PR

PR #2412, opened 2022-04-05
FFT-based preconditioner for immersed rectilinear grids

glwagner:
This PR implements an FFT-based preconditioner for `PCGImplicitFreeSurfaceSolver` on immersed boundary `RectilinearGrid`. Note that we have a direct FFT solver for _non-immersed_ `RectilinearGrid`.

# Premise

The premise of the FFT-based preconditioner is to use the non-immersed FFT-based algorithm for implicit free surface solves on rectilinear grids as a _preconditioner_ for a conjugate gradient iterative solver. In a domain with variable bathymetry, the conjugate gradient method is used to solve the "implicit step equation"

```
(∇ʰ⋅ H ∇ʰ - 1 / (g Δt²)) ηⁿ⁺¹ = 1 / (g Δt) ∇ʰ ⋅ Q★ - 1 / (g Δt²) ηⁿ
```

where `H(x, y)` is the bottom depth (there are some subtleties in a proper finite-volume discretization of this equation, which we'll ignore for this high-level overview). Above, `ηⁿ⁺¹` is the free surface at time-step `n+1` and `ηⁿ` is the free surface at time-step `n`, `Q★` is the _predictor_ barotropic mass flux that we are correcting, `g` is gravitational acceleration, and `Δt` is the time-step.

The FFT-based preconditioner instead solves

```
[∇² - 1 / (g H̄ Δt²)] ηⁿ⁺¹ = (∇ʰ ⋅ Q★ - ηⁿ / Δt) / (g H̄ Δt)
```

where `H̄` is an "effective constant depth" (here we use `grid.Lz`, but this could be changed in the future).

This PR adds `validation/immersed_boundaries/bumpy_baroclinic_adjustment.jl` to test the effectiveness of this solution. A summary of what we found using that validation experiment and some more careful benchmarking is

* The FFT-based preconditioner reduces the iterations required for convergence from O(100) (usually 300-400) to O(10) over no preconditioning for the bumpy baroclinic problem.
* The FFT-based preconditioner is roughly 2x faster on both CPU and GPU than the current fastest methods (matrix iterative solver with ILU and sparse approximate inverse preconditioner, respectively).
* Unlike the other fastest methods, the time-step can be freely modified with no cost when using the FFT-based preconditioner

Notes:

[...]

PR #2780, opened 2022-10-17
Solver for nonhydrostatic IBG

simone-silvestri:
cc @tomchor


glwagner (changes requested):
Let's use PreconditionedConjugateGradientSolver with an FFT-based preconditioner (ie the original FFT solver) for this.

tomchor:
> Let's use PreconditionedConjugateGradientSolver with an FFT-based preconditioner (ie the original FFT solver) for this.

Can you explain a bit the reason for this choice, just so that I can understand better? At first it seems to me that using a FFT-based preconditioner would still produce Gibbs phenomenon effects in situations where the original FFT-based solver also would, no?

glwagner:
Is it the pressure field that has oscillations, or the velocity field?

Either way, I don't think that's important. A preconditioner should be some kind of approximate inverse for A, where we are solving for x in Ax = b --- or in other words, an approximate solver. The preconditioner does not have to be "exact" or "error free", since it's really the CG iteration that solves the problem. The idea is that the FFT-based solver yields a pressure field that is _almost_ correct.

A similar strategy succeeded in this PR: https://github.com/CliMA/Oceananigans.jl/pull/2412

tomchor:
> Either way, I don't think that's important. A preconditioner should be some kind of approximate inverse for A, where we are solving for x in Ax = b --- or in other words, an approximate solver. The preconditioner does not have to be "exact" or "error free", since it's really the CG iteration that solves the problem. The idea is that the FFT-based solver yields a pressure field that is _almost_ correct.

Ah, I see, so the idea behind using the FFT-based solver specifically as a preconditioner is because it's probably faster than other combination (if I understand correctly).

Thanks!

glwagner:
> > Either way, I don't think that's important. A preconditioner should be some kind of approximate inverse for A, where we are solving for x in Ax = b --- or in other words, an approximate solver. The preconditioner does not have to be "exact" or "error free", since it's really the CG iteration that solves the problem. The idea is that the FFT-based solver yields a pressure field that is _almost_ correct.
> 
> Ah, I see, so the idea behind using the FFT-based solver specifically as a preconditioner is because it's probably faster than other combination (if I understand correctly).
> 
> Thanks!

For sure, and we could use other preconditioners. But I suspect the FFT will be _much_ faster than any other preconditioner, which is why I think it's so interesting (and somewhat novel --- though similar approaches have been used for other problems).

simone-silvestri:
yep, i ll close

PR #2782, opened 2022-10-18
An attempt to bump KernelAbstractions 0.8 + CUDAKernels 0.4

navidcy:
supersedes #2781


navidcy:
So a lot of things break when I bump KernelAbstractions 0.8.4 + CUDAKernels 0.4.3....

simone-silvestri:
Some `@inline` are missing in `conditional_derivatives.jl`

simone-silvestri:
Other `@inline`s missing in `Advection/topologically_conditional_interpolation.jl`

navidcy:
I found a few more -- let's see.

simone-silvestri:
Nice, maybe this will also increase the performance!

navidcy:
There are several issues with `fill_halos`, e.g.
https://buildkite.com/clima/oceananigans/builds/8920#0183f07f-61ab-4914-baf5-c227ebe4c3ef/16-332

Are these missing `@inline`s?

simone-silvestri:
Yep (in `fill_halo_regions_value_gradient.jl`)

vchuravy:
Yeah KA got rid of the force-inlining on the GPU. @lcw started the work of adding it back in a better way than before. The benefit is that we are no longer using Cassette so errors should be vastly improved.

navidcy:
OK, there are still:

- something with `regrid!` and `length`?
  https://buildkite.com/clima/oceananigans/builds/8922#0183f16b-a7d8-4da0-a8b8-25afbb402f81/16-433
- https://buildkite.com/clima/oceananigans/builds/8922#0183f16b-a7f9-49df-ad9b-62f98a92be8a/24-595
- something with `convert`?
  https://buildkite.com/clima/oceananigans/builds/8922#0183f16b-a81c-4a0a-a552-0edffc90518f/24-459

navidcy:
@simone-silvestri can you have a look here? Some help with this would be great... Just have a look at the error of the tests that fail and let me know if you can understand where the root of the manner is?

simone-silvestri:
Weird errors, it looks like the GPU calculations are inaccurate

navidcy:
@vchuravy feel free to merge if you feel it's for the best... tests won't pass :)

navidcy:
(I even cancelled the ci job to save resources)

tomchor:
@navidcy should we close this pr now that KA was updated?

navidcy:
definitely

Issue #2453, opened 2022-04-21
Instability of `IsopycnalSkewSymmetricDiffusivity` at high resolution


glwagner:
At high enough ~~vertical~~ vertical + horizontal resolution, it seems that the `zonally_averaged_baroclinic_adjustment.jl` validation test blows up with `IsopycnalSkewSymmetricDiffusivity`; specifically:

```julia
gerdes_koberle_willebrand_tapering = FluxTapering(1e-2)
gent_mcwilliams_diffusivity = IsopycnalSkewSymmetricDiffusivity(κ_skew = 1000,
                                                                κ_symmetric = 900,
                                                                slope_limiter = gerdes_koberle_willebrand_tapering)
```

For example, with 50m vertical resolution (Nz=20, Lz=1000m, 20km horizontal resolution) we find

https://user-images.githubusercontent.com/15271942/164360760-ef03fc9d-e07a-4b50-9e80-bd79997cfa1d.mp4

while at 25m resolution (same extent, horizontal resolution, time-step) we find

https://user-images.githubusercontent.com/15271942/164360772-e39a08e4-eb5e-48e4-bfc4-cde1e241ec0f.mp4

glwagner:
@sandreza perhaps relevant to you

glwagner:
@navidcy

glwagner:
Ok, so I noticed that decreasing the time-step to 1 minute (from 5-10 minutes) makes things stable:

https://user-images.githubusercontent.com/15271942/164360890-7940aeea-1128-42c1-975a-b86812abed6a.mp4

do we need to great the vertical diffusive component implicitly? @christophernhill do you know (or anyone else)?

glwagner:
Reducing the horizontal resolution (while maintaining the same high vertical resolution) also helps:

https://user-images.githubusercontent.com/15271942/164365464-8cc99073-67f1-4b5d-ba12-25956c95b0c0.mp4

I think this suggests that the issue is a sort of vanilla diffusive time-step restriction. We can't treat _horizontal_ diffusion implicitly without some effort (but this is probably not relevant to any real problems, which are likely limited by advective time-scale).

So perhaps this issue is just "too big a time-step for the given resolution"... ?

Issue #2450, opened 2022-04-20
Improvements to CATKEVerticalDiffusivity


glwagner:
I'm opening this issue to have a discussion and solicit feedback on a few proposed improvements to CATKEVerticalDiffusivity.

The most crucial improvement was to implement semi-implicit time-stepping for the dissipation term in the TKE equation, in #2426 . In `validation/windy_convection.jl`, the time-step had to be 30 seconds for CATKE stability before this PR. After that PR, a time-step of 10 minutes was stable. For models ~ 1 deg resolution we need to be able to take 20 minute time-steps, so this change was necessary if CATKE were to be used in a global simulation.

Another change being implemented in #2449 is to calculate the CATKE diffusivities at vertical cell interfaces, rather than cell centers. This is natural because the diffusivity is _used_ at vertical cell interfaces (for horizontal momentum components and tracers), and also because many of the quantities involved in calculating the CATKE diffusivity (buoyancy frequency, shear) are also located at vertical cell interfaces.

I think there's a few other developments that may be needed:

- Revision of the "convective adjustment" part of CATKE. We haven't yet succeeded in formulating a version of CATKE that's more accurate when "convective adjustment" is included (I use quotes, because its a dynamic convective adjustment rather than the typical static-diffusivity convective adjustment)
- Should we consider also calculating TKE at vertical cell interfaces? This makes sense if we think of TKE as "mostly" vertical velocity variance. If we do this, we may need to change the boundary conditions that we use for TKE.
- A cheaper / simpler stability function. We currently use a tanh-based stability function (the stability function is an Ri-dependent coefficient that decreases mixing coefficients when Ri increases). We might use a piecewise linear function (cheaper) or perhaps an exponential (an exponential stability function improved Ri-based results in #2423)
- CATKE's TKE equation needs to be generalized so that shear production and buoyancy flux include the _total_ contribution, not just the contribution from CATKE diffusion (@simone-silvestri this is especially important if we want to pair CATKE with a mass flux scheme)

Does anyone else have any suggestion or things on their wishlist?

cc @adelinehillier @sandreza @navidcy @simone-silvestri

PR #2786, opened 2022-10-27
CATKE and GM in a near-global 1 degree simulation + updates Ri-based vertical diffusivity

### LINKED ISSUE #2800: `update_state!` for `HydrostaticFreeSurfaceModel` seems that overwrites itself?
I think after #2773 we get this warning about method redefinition...

https://github.com/CliMA/Oceananigans.jl/tree/a86ef32581192d82a655e1b2ff4411e79917a379

```Julia
julia> using Oceananigans
WARNING: Method definition update_state!(Oceananigans.Models.HydrostaticFreeSurfaceModels.HydrostaticFreeSurfaceModel{TS, E, A, S, G, T, V, B, R, F, P, U, C, Φ, K, AF} where AF where K where Φ where C where U where P where F where R where B where V where T where G where S where A<:Oceananigans.Architectures.AbstractArchitecture where E where TS) in module HydrostaticFreeSurfaceModels at /Users/navid/Research/OC2.jl/src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl:18 overwritten at /Users/navid/Research/OC2.jl/src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl:20.
  ** incremental compilation may be fatally broken for this module **

```

glwagner:
This PR supports the implementation of a 1 degree simulation with CATKE and GM, and a calibration of GM coefficients in the near-global context against ECCO data. The simulation is implemented in ClimaOcean.jl:

https://github.com/CliMA/ClimaOcean.jl/blob/main/src/one_degree_global_simulation.jl

This PR supercedes #2597.

Closes #2800


src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -166,7 +166,8 @@ end
     ntuple(Val(length(bottom_bc))) do n
         Base.@_inline_meta
         @inbounds begin
-            _fill_bottom_halo!(i, j, grid, c[n], bottom_bc[n], loc[n], args...)
+            #_fill_bottom_halo!(i, j, grid, c[n], bottom_bc[n], loc[n], args...)
+            _fill_flux_bottom_halo!(i, j, 1, grid, c[n])
```
glwagner, line 170: @simone-silvestri whats this?

simone-silvestri: I don't know, you put it in. Maybe because of #2709?

navidcy: Regardless of who did it, should we revert back to `_fill_bottom_halo!`?

src/Oceananigans.jl
```diff
@@ -142,18 +141,31 @@ import Base:
     push!
 
 "Boolean denoting whether AMGX.jl can be loaded on machine."
-const hasamgx = @static Sys.islinux() ? true : false
+# TODO: find a way to check whether the libraries for AMGX and NETCDF 
+# (libamgxsh and libnetcdf, respectively) are installed on the machine
+const hasamgx   = false #@static Sys.islinux() ? true : false
```
navidcy, line 146: why this change?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
```diff
@@ -1,3 +1,33 @@
...
+
+Turbulent kinetic energy dissipation
+====================================
+
+Surface flux model
+==================
+
+```math
+Qᵉ = - Cᴰ * (Cᵂu★ * u★³ + CᵂwΔ * w★³)
+```
```
navidcy, line 15:
```suggestion
```
Qᵉ = - Cᴰ * (Cᵂu★ * u★³ + CᵂwΔ * w★³)
```
```

src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
```diff
@@ -43,39 +41,25 @@ a decreasing function of the Richardson number, ``Ri``.
...
-* `Ri₀κ` (`Float64` parameter): ``Ri`` threshold for decreasing viscosity. Default: -0.13
-* `Riᵟκ` (`Float64` parameter): Width over which ``Ri`` decreases to 0. Default: 0.6
-* `coefficient_z_location` (`Face()` or `Center()`): The vertical location of the diffusivity and viscosity.
-                                                     Default: `Face()`.
-* `Ri_dependent_tapering`: The ``Ri``-dependent tapering. Default: `ExponentialRiDependentTapering()`.
+* ν₀ (Float64 parameter): Non-convective viscosity.
+* κ₀ (Float64 parameter): Non-convective diffusivity for tracers.
+* κᶜ (Float64 parameter): Convective adjustment diffusivity for tracers.
+* Ri₀ (Float64 parameter): Ri threshold for decreasing viscosity and diffusivity.
+* Riᵟ (Float64 parameter): Ri-width over which viscosity and diffusivity decreases to 0.
```
navidcy, line 48:
```suggestion
* `ν₀` (Float64): Non-convective viscosity.
* `κ₀` (Float64): Non-convective diffusivity for tracers.
* `κᶜ` (Float64): Convective adjustment diffusivity for tracers.
* `Ri₀` (Float64): ``Ri`` threshold for decreasing viscosity and diffusivity.
* `Riᵟ` (Float64): ``Ri``-width over which viscosity and diffusivity decreases to 0.
```

validation/vertical_mixing_closures/windy_catke_convection.jl
```diff
@@ -0,0 +1,107 @@
...
+#####
+
+convective_adjustment = ConvectiveAdjustmentVerticalDiffusivity(convective_κz=0.1, convective_νz=0.01)
+
+grid = RectilinearGrid(size=32, z=(-256, 0), topology=(Flat, Flat, Bounded))
+coriolis = FPlane(f=1e-4)
+
+N² = 1e-5
+Qᵇ = +1.2e-7
+Qᵘ = -1e-4 #
```
navidcy, line 23:
```suggestion
Qᵘ = -1e-4
```

src/Oceananigans.jl
navidcy, line 146:
```suggestion
const hasamgx   = @static Sys.islinux() ? true : false

```

glwagner: We can't use this condition --- for example, MIT's Satori cluster is linux, but does not have AMGX (because it's a power 9 architecture). How do we correctly determine whether a user's architecture has AMGX?

glwagner: We are using a system that is linux but does not have AMGX.

src/BoundaryConditions/fill_halo_regions.jl
glwagner: Yeah, we have to.

glwagner:
Note to self that we will need to update the docs (and also notation I think) for the Ri based diffusivity, since the "diffuscion parameter" is actually a time-scale:

https://github.com/CliMA/Oceananigans.jl/blob/1143866d2d069075fe8fd5fa7d71b1ab18943f2d/src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl#L148-L149

src/Oceananigans.jl
navidcy: ok, I see...

@elise-palethorpe do you remember what else we had tried here?

glwagner: would a try catch work?

```julia
const hasamgx =
    try
        using AMGX
        true
    catch
        false
    end

Comment (navidcy) [reply src/Oceananigans.jl:146]: We tried that! It won't work.... because `using AMGX` errors **only** the first time it's called!

```Julia
(base) navid:OC.jl/ (glw/catke-plus-gm) $ julia --project                                                                                     [8:48:09]
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.2 (2022-09-29)
 _/ |\__'_|_|_|\__'_|  |
|__/                   |

julia> using AMGX
ERROR: InitError: UndefVarError: libamgxsh not defined
Stacktrace:
 [1] getproperty
   @ ./Base.jl:31 [inlined]
 [2] __init__()
   @ AMGX ~/.julia/packages/AMGX/GFHHN/src/AMGX.jl:30
 [3] _include_from_serialized(pkg::Base.PkgId, path::String, depmods::Vector{Any})
   @ Base ./loading.jl:831
 [4] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String, build_id::UInt64)
   @ Base ./loading.jl:1039
 [5] _require(pkg::Base.PkgId)
   @ Base ./loading.jl:1315
 [6] _require_prelocked(uuidkey::Base.PkgId)
   @ Base ./loading.jl:1200
 [7] macro expansion
   @ ./loading.jl:1180 [inlined]
 [8] macro expansion
   @ ./lock.jl:223 [inlined]
 [9] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1144
during initialization of module AMGX

julia> using AMGX

julia> a = 1
1
```

navidcy:
See https://github.com/JuliaGPU/AMGX.jl/issues/22

src/Oceananigans.jl
```diff
@@ -142,18 +141,31 @@ import Base:
     push!
 
 "Boolean denoting whether AMGX.jl can be loaded on machine."
-const hasamgx = @static Sys.islinux() ? true : false
+# TODO: find a way to check whether the libraries for AMGX and NETCDF 
+# (libamgxsh and libnetcdf, respectively) are installed on the machine
+const hasamgx   = false #@static Sys.islinux() ? true : false
+const hasnetcdf = false
```
navidcy, line 147: why is this line needed?

src/Oceananigans.jl
glwagner: But what's the issue? Doesn't this still solve the problem?

glwagner: Are you saying that the problem is you can spuriously get `hasamgx=true` if you wrote `using AMGX` before `using Oceananigans`? I do agree that in this case, its wrong that `hasamgx=true`. On the other hand, it would seem that the situation is still happy, because one can use Oceananigans. If they _actually_ tried to use AMGX, they'd get another error (presumably). We could change the name of the variable to `definitely_doesnt_have_amgx` (with the inverse value) if want to be clear, like

```julia
const definitely_doesnt_have_amgx =
    try
        using AMGX
        false
    catch
        true
    end
```

Or am I missing something?

src/Oceananigans.jl
glwagner: NetCDF_jll isn't available for power 9, so sometimes it's a pain to install. This let's us use Oceananigans even if we don't have working NetCDF

navidcy: What about all the output writer tests? We won’t have them?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/mixing_length.jl
```diff
@@ -101,92 +100,149 @@ end
...
 
 @inline function shear_mixing_lengthᶜᶜᶠ(i, j, k, grid, e, velocities, tracers, buoyancy)
     FT = eltype(grid)
     ∂z_u² = ℑxᶜᵃᵃ(i, j, k, grid, ϕ², ∂zᶠᶜᶠ, velocities.u)
     ∂z_v² = ℑyᵃᶜᵃ(i, j, k, grid, ϕ², ∂zᶜᶠᶠ, velocities.v)
-    S = sqrt(∂z_u² + ∂z_v²)
+    S² = ∂z_u² + ∂z_v²
     e⁺ = ℑzᵃᵃᶠ(i, j, k, grid, ψ⁺, e)
-    return ifelse(S == 0, FT(Inf), sqrt(e⁺) / S)
+    return ifelse(S == 0, FT(Inf), sqrt(e⁺ / S²))
```
navidcy, line 119:
```suggestion
    return ifelse(S² == 0, FT(Inf), sqrt(e⁺ / S²))
```

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/mixing_length.jl
```diff
@@ -280,9 +280,12 @@ end
...
 
-    #return ℓᵟ + ℓ★ + ℓʰ
-    Cʷℓ = closure.mixing_length.Cʷℓ
-    return smoothmax(Cʷℓ, ℓᵟ, ℓ★, ℓʰ)
+    #return ℓᵟ
+    return ℓᵟ + ℓ★
+    #return ℓᵟ #+ ℓ★ + ℓʰ
+
+    #Cʷℓ = closure.mixing_length.Cʷℓ
+    #return smoothmax(Cʷℓ, ℓᵟ, ℓ★, ℓʰ)
```
navidcy, line 288: what's all these commented-out lines?

validation/vertical_mixing_closures/windy_catke_convection.jl
```diff
@@ -0,0 +1,122 @@
...
+                             Cᴷe⁺ = 0.5)
+
+turbulent_kinetic_energy_equation =
+    TurbulentKineticEnergyEquation(CᴰRiʷ = 0.5,
+                                   Cᵂu★ = 0.5,
+                                   CᵂwΔ = 0.5,
+                                   Cᴰ⁻ = 0.5)
+
+closure = CATKEVerticalDiffusivity(; mixing_length, turbulent_kinetic_energy_equation)
+#closure = RiBasedVerticalDiffusivity()
```
navidcy, line 43:
```suggestion
```

from the filename alone I think this should be dropped

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
```diff
@@ -24,35 +73,61 @@ end
...
-    #                   = - Cᴰ e³² / ℓ
+    #   then      L e = - ϵ
+    #                 = - Cᴰ e³² / ℓ
     #
-    #   * and thus    L = - Cᴰ √e / ℓ
-    
-    return - Cᴰ * sqrt(e⁺) / ℓ
+    #   and thus    L = - Cᴰ √e / ℓ .
+
+    return - Cᴰ * sqrt(abs(eᵢ)) / ℓ
```
navidcy, line 123: why `abs`? `eᵢ` could be negative? is it just for small numerical errors?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
```diff
@@ -94,3 +169,159 @@ end
...
+
+    return - Cᵂu★ * u★^3 - CᵂwΔ * wΔ³
+end
+
+""" Computes the friction velocity u★ based on fluxes of u and v. """
+@inline function friction_velocity(i, j, grid, clock, fields, velocity_bcs)
+    FT = eltype(grid)
+    Qᵘ = getbc(velocity_bcs.u, i, j, grid, clock, fields) 
+    Qᵛ = getbc(velocity_bcs.v, i, j, grid, clock, fields) 
+    return sqrt(sqrt(Qᵘ^2 + Qᵛ^2))
```
navidcy, line 236:
```suggestion
    return (Qᵘ^2 + Qᵛ^2)^(FT(1/4))
```

```Julia
julia> A = rand(); B = rand();

julia> @time  sqrt(sqrt(A^2+B^2));
  0.000014 seconds (5 allocations: 80 bytes)

julia> @time  sqrt(sqrt(A^2+B^2));
  0.000014 seconds (5 allocations: 80 bytes)

julia> @time (A^2+B^2)^(Float64(1/4));
  0.000012 seconds (4 allocations: 64 bytes)

julia> @time (A^2+B^2)^(Float64(1/4));
  0.000010 seconds (4 allocations: 64 bytes)
```

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
glwagner: Yes, due to time discretization errors `e` can be negative (unfortunately).

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
glwagner: Hmm, your timings don't show a measurable difference. `btime` might though. Also we should do this on the GPU, that's where I expect a larger difference. Probably better to compute over an array too for the timing.

glwagner: As proof:

```julia
julia> A = rand(); B = rand();

julia> @time  sqrt(sqrt(A^2+B^2));
  0.000009 seconds (5 allocations: 80 bytes)

julia> @time  sqrt(sqrt(A^2+B^2));
  0.000009 seconds (5 allocations: 80 bytes)

julia> @time (A^2+B^2)^(Float64(1/4));
  0.000008 seconds (4 allocations: 64 bytes)

julia> @time (A^2+B^2)^(Float64(1/4));
  0.000010 seconds (4 allocations: 64 bytes)
```

glwagner:
```julia
julia> A = rand(1024, 1024); B = rand(1024, 1024);

julia> @time @. (A^2+B^2)^(Float64(1/4));
  0.059826 seconds (14 allocations: 8.000 MiB)

julia> @time @. sqrt(sqrt(A^2+B^2));
  0.002154 seconds (16 allocations: 8.001 MiB)
```

glwagner: `sqrt` looks to be significantly faster than `pow(a, Float64)` on the CPU. Not sure about GPU though.

src/Oceananigans.jl
glwagner: We can't merge with `hasnetcdf=false`. Instead we need to figure out how to determine whether the user has a working version of `netcdf`, ie

```julia
const hasnetcdf =
    try
        using NCDatasets
        true
    catch
        false
    end
```

If `NetCDF_jll` isn't available for a user's platform, they have to install it manually somehow (either by finding a binary via miniconda or another service, or by using the instructions provided by NetCDF). They shouldn't have to modify Oceananigans source code (but they may have to re-build it, if they have already compiled it and at that time `hasnetcdf=false`).

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
navidcy: OK!

src/Oceananigans.jl
navidcy: would

```Julia
const hasnetcdf =
    try
        using NCDatasets
        true
    catch
        false
    end
```

be OK for what you are running on Satori?  Shall we do this for now?

navidcy:
`test_ensemble_hydrostatic_free_surface_models.jl` fails on this PR and I can't figure out why. In particular, the case with many different Coriolises...

```julia
julia> include("test_ensemble_hydrostatic_free_surface_models.jl")
```
fails...

This is affecting the tests on https://github.com/CliMA/ParameterEstimocean.jl/pull/295

navidcy:
> `test_ensemble_hydrostatic_free_surface_models.jl` fails on this PR and I can't figure out why. In particular, the case with many different Coriolises...
> 
> ```julia
> julia> include("test_ensemble_hydrostatic_free_surface_models.jl")
> ```
> 
> fails...
> 
> This is affecting the tests on [CliMA/ParameterEstimocean.jl#295](https://github.com/CliMA/ParameterEstimocean.jl/pull/295)

fixed it

src/Fields/set!.jl
```diff
@@ -39,26 +39,14 @@ function set!(u::Field, f::Union{Array, CuArray, OffsetArray})
...
-            # SubArray.
-            u_parent isa SubArray && (u_parent = Array(u_parent))
-            v_parent isa SubArray && (v_parent = Array(v_parent))
-            copyto!(u_parent, v_parent)
-        catch # just copy interior points
-            v_data = arch_array(architecture(u), v.data)
-            interior(u) .= interior(v_data, location(v), v.grid, v.indices)
-        end
+        v_data = arch_array(architecture(u), v.data)
+        interior(u) .= interior(v_data, location(v), v.grid, v.indices)
```
navidcy, line 49: so only interior points are copied, right?
then 

https://github.com/CliMA/Oceananigans.jl/blob/edb4268d34e86ec9e202649307e53444ef92c5e7/test/test_nonhydrostatic_models.jl#L150-L154

will fail!

Do we want `set!` to call `fill_halos!`?

navidcy: Why not simply:

```julia
set!(u::Field, v::Field) = set!(u::Field, v.data)
```

glwagner: what `set!` does that call? Maybe that will work!

navidcy: I guess this:
https://github.com/CliMA/Oceananigans.jl/blob/87a046fdd7da4358e6da2b1774ddc7c258f87948/src/Fields/set!.jl#L35-L39

simone-silvestri: That should work

src/Oceananigans.jl
```diff
@@ -143,7 +143,7 @@ import Base:
 "Boolean denoting whether AMGX.jl can be loaded on machine."
 # TODO: find a way to check whether the libraries for AMGX and NETCDF 
 # (libamgxsh and libnetcdf, respectively) are installed on the machine
-const hasamgx   = false #@static Sys.islinux() ? true : false
+const hasamgx   = @static (Sys.islinux() && Sys.ARCH == :x86_64) ? true : false
```
simone-silvestri, line 146: Is it possible to have a way to detect if the library is installed?

src/Fields/set!.jl
glwagner: https://github.com/CliMA/Oceananigans.jl/blob/3557db3692ff8e1978b126369230c9bacae42ae8/src/Fields/broadcasting_abstract_fields.jl#L26-L39

glwagner: I don't think `Base.Broadcast.materialize!(dest.data, bc)` does what we want; it includes halo regions. So I think we need to keep broadcasting over `interior` when setting fields.

glwagner: Side note I think we need to clean up broadcasting...

navidcy: Yes I had troubles with `set!(u::Field, v::Field) = set!(u::Field, v.data)` over in ParameterEstimocean.jl. Let's leave it as is?

navidcy: But what about the halos? They are not copied like this. Should we change the tests?

https://github.com/CliMA/Oceananigans.jl/blob/edb4268d34e86ec9e202649307e53444ef92c5e7/test/test_nonhydrostatic_models.jl#L150-L154

src/Oceananigans.jl
navidcy: If [this](https://github.com/JuliaGPU/AMGX.jl/pull/24) merges then we can just call `using AMGX` and we won't get error.

navidcy:
@glwagner, regarding the error with the simulations with multiple Coriolises:
seems like adding a `@show "hi"` *before* the `@apply_regionally correct_velocities_and_store_tendecies!(model, Δt)` in

https://github.com/CliMA/Oceananigans.jl/blob/b1997713b35f41b42a1f5cc5660d6ceb2ec3e305/src/TimeSteppers/quasi_adams_bashforth_2.jl#L95-L97

makes the error go away. Does this give a hint for what might be causing it?

glwagner:
Probably race condition with apply regionally then. Cc @simone-silvestri

simone-silvestri:
Unlikely since that will just execute `pressure_correct_velocities!(model, Δt)` and `store_tendencies!(model)`. It might be an upstream synchronization problem in the `calculate_pressure_correction!`

glwagner:
In this case `pressure_correct_velocities!(model, dt) = nothing` because we are using `HydrostaticFreeSurfaceModel`

src/Oceananigans.jl
```diff
@@ -142,18 +138,36 @@ import Base:
     push!
 
 "Boolean denoting whether AMGX.jl can be loaded on machine."
-const hasamgx = @static Sys.islinux() ? true : false
+# TODO: find a way to check whether the libraries for AMGX and NETCDF 
+# (libamgxsh and libnetcdf, respectively) are installed on the machine
```
navidcy, line 142: this looks useful...

```Julia
julia> using Libdl

help?> find_library
search: find_library

  find_library(names, locations)


  Searches for the first library in names in the paths in the locations list, DL_LOAD_PATH, or system library paths (in that order) which can successfully be dlopen'd. On success, the return value will be one of
  the names (potentially prefixed by one of the paths in locations). This string can be assigned to a global const and used as the library name in future ccall's. On failure, it returns the empty string.
```

navidcy:
> Unlikely since that will just execute `pressure_correct_velocities!(model, Δt)` and `store_tendencies!(model)`. It might be an upstream synchronization problem in the `calculate_pressure_correction!`

Actually, even putting `@show "hi"` **just before** `calculate_pressure_correction!` also does the job.

src/Oceananigans.jl
glwagner: nice

glwagner:
> > Unlikely since that will just execute `pressure_correct_velocities!(model, Δt)` and `store_tendencies!(model)`. It might be an upstream synchronization problem in the `calculate_pressure_correction!`
> 
> Actually, even putting `@show "hi"` **just before** `calculate_pressure_correction!` also does the job.

Right, `calculate_pressure_correction!` does nothing for the free surface model so it shouldn't matter where it's placed relative to that.

Is that the only place within `time_step!` that fixes it?

src/Oceananigans.jl
navidcy: I don't think it's useful since `find_library("libnetcdf")` gives empty for systems that NCDatasets works or not...

navidcy: there is some discussion that seems relevant at https://github.com/Alexander-Barth/NCDatasets.jl/issues/150

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
```diff
@@ -19,7 +19,9 @@ update_state!(model::HydrostaticFreeSurfaceModel) = update_state!(model, model.g
 Update peripheral aspects of the model (auxiliary fields, halo regions, diffusivities,
 hydrostatic pressure) to the current model state.
 """
-function update_state!(model::HydrostaticFreeSurfaceModel, grid)
+update_state!(model::HydrostaticFreeSurfaceModel) = update_state!(model, model.grid)
+
+function update_state!(model::HydrostaticFreeSurfaceModel, callbacks=[])
```
glwagner, line 24: This doesn't look right. Shouldn't we have a 3-argument version of `update_state!`? Separating out `model.grid` is how we do some performance optimizations for "single column model mode".

navidcy: damn! sorry... I may have messed up during merging latest main

navidcy: or is it OK?
https://github.com/CliMA/Oceananigans.jl/blob/7f17c1373c8f3ac1d0dbde31f799cda29ce146f0/src/Models/HydrostaticFreeSurfaceModels/single_column_model_mode.jl#L43

navidcy: We need to add the `callbacks=[]` positional argument in `update_state!(model::HydrostaticFreeSurfaceModel, grid::SingleColumnGrid)`, right?

navidcy: and also replace the line you are referring to with `update_state!(model::HydrostaticFreeSurfaceModel, grid::AbstractGrid, callbacks=[])

navidcy: I'm a bit confused because in `main` at the moment I see:

navidcy: I think I fixed it.

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
```diff
@@ -21,7 +21,7 @@ hydrostatic pressure) to the current model state.
 """
 update_state!(model::HydrostaticFreeSurfaceModel) = update_state!(model, model.grid)
```
glwagner: Don't we need to pass callbacks through this function?

navidcy: it's like this for all other models... so I was confused.

navidcy: Yes, it's changed! :) see [8a678d8](https://github.com/CliMA/Oceananigans.jl/pull/2786/commits/8a678d8c2c2c53844f773f529a5398aee12f032a)

navidcy:
@glwagner have a look at the PR when you can.

navidcy:
We still need to think a better way to conditionally import NCDatasets

src/Oceananigans.jl
```diff
@@ -143,18 +139,36 @@ import Base:
...
+const hasamgx   = @static (Sys.islinux() && Sys.ARCH == :x86_64) ? true : false
+
+const hasnetcdf =
+    try
+        using NCDatasets
+        true
+    catch
+        false
+    end
```
navidcy, line 152: Since `try-catch` with `using NCDatasets` doesn't work, should we do this for now?

```suggestion
const hasnetcdf = @static (Sys.islinux() && Sys.ARCH == :x86_64) ? true : false
```

navidcy (approved):
I think this is ready.

I run the `one_degree_near_global_simulation.jl` from ClimaOcean.jl to confirm that we didn't break anything.

```Julia
julia> include("one_degree_near_global_simulation.jl")
┌ Warning: RiBasedVerticalDiffusivity is an experimental turbulence closure that
│ is unvalidated and whose default parameters are not calibrated for
│ realistic ocean conditions or for use in a three-dimensional
│ simulation. Use with caution and report bugs and problems with physics
│ to https://github.com/CliMA/Oceananigans.jl/issues.
└ @ Oceananigans.TurbulenceClosures ~/.julia/packages/Oceananigans/X9Lwx/src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl:68
boundary_layer_turbulence_closure = RiBasedVerticalDiffusivity{VerticallyImplicitTimeDiscretization}
[ Info: Reading initial conditions...
[ Info: ... read initial conditions (38.936 ms)
[ Info: Reading boundary conditions...
[ Info: ... read boundary conditions (603.298 ms)
┌ Info: Created 360×150×48 ImmersedBoundaryGrid{Float64, Periodic, Bounded, Bounded} on GPU with 5×5×5 halo:
│ ├── immersed_boundary: GridFittedBottom(min(h)=-9.02e+03, max(h)=1.00e+02)
│ ├── underlying_grid: 360×150×48 LatitudeLongitudeGrid{Float64, Periodic, Bounded, Bounded} on GPU with 5×5×5 halo and with precomputed metrics
│ ├── longitude: Periodic λ ∈ [-180.0, 180.0) regularly spaced with Δλ=1.0
│ ├── latitude:  Bounded  φ ∈ [-75.0, 75.0]   regularly spaced with Δφ=1.0
└ └── z:         Bounded  z ∈ [-5244.5, 0.0]  variably spaced with min(Δz)=10.0, max(Δz)=410.5
[ Info: Building a model...
┌ Info: ... built HydrostaticFreeSurfaceModel{GPU, ImmersedBoundaryGrid}(time = 0 seconds, iteration = 0)
│ ├── grid: 360×150×48 ImmersedBoundaryGrid{Float64, Periodic, Bounded, Bounded} on GPU with 5×5×5 halo
│ ├── timestepper: QuasiAdamsBashforth2TimeStepper
│ ├── tracers: (T, S)
│ ├── closure: Tuple with 4 closures:
│ │   ├── HorizontalScalarDiffusivity{ExplicitTimeDiscretization}(ν=50000.0, κ=(T=0.0, S=0.0, e=50000.0))
│ │   ├── RiBasedVerticalDiffusivity{VerticallyImplicitTimeDiscretization}
│ │   ├── VerticalScalarDiffusivity{VerticallyImplicitTimeDiscretization}(ν=PiecewiseConstantVerticalDiffusivity(-49.0, 0.0001, 0.0001), κ=(T=1.0e-5, S=1.0e-5))
│ │   └── IsopycnalSkewSymmetricDiffusivity(κ_skew=900.0, κ_symmetric=900.0)
│ ├── buoyancy: SeawaterBuoyancy with g=9.80665 and BoussinesqEquationOfState{Float64} with -ĝ = ZDirection
│ ├── free surface: ImplicitFreeSurface with gravitational acceleration 9.80665 m s⁻²
│ │   └── solver: MatrixImplicitFreeSurfaceSolver
└ └── coriolis: HydrostaticSphericalCoriolis{Oceananigans.Coriolis.WetCellEnstrophyConservingScheme, Float64}.
[ Info: Model building time: 2.058 minutes
┌ Warning: model.closure contains a function somewhere in its hierarchy and will not be checkpointed.
└ @ Oceananigans.OutputWriters ~/.julia/packages/Oceananigans/X9Lwx/src/OutputWriters/checkpointer.jl:88
[ Info: Running a simulation with Δt = 20 minutes
[ Info: Time:     345 days, iteration: 0, max(|u|): 0.00e+00 ms⁻¹, wmax: 0.00e+00, loc: (1, 1, 1), wall time: 42.331 seconds
[ Info:     ... simulation initialization complete (44.549 seconds)
[ Info:     ... initial time step complete (39.366 minutes).
[ Info: Time: 345.007 days, iteration: 10, max(|u|): 1.63e-02 ms⁻¹, wmax: 4.19e-04, loc: (345, 20, 16), wall time: 40.263 minutes
[ Info: Time: 345.014 days, iteration: 20, max(|u|): 3.53e-02 ms⁻¹, wmax: 6.63e-04, loc: (345, 20, 14), wall time: 579.856 ms
[ Info: Time: 345.021 days, iteration: 30, max(|u|): 7.10e-02 ms⁻¹, wmax: 7.41e-04, loc: (345, 20, 14), wall time: 550.208 ms
[ Info: Time: 345.028 days, iteration: 40, max(|u|): 1.08e-01 ms⁻¹, wmax: 7.51e-04, loc: (352, 21, 19), wall time: 531.924 ms
[ Info: Time: 345.035 days, iteration: 50, max(|u|): 1.38e-01 ms⁻¹, wmax: 9.09e-04, loc: (352, 21, 19), wall time: 542.333 ms
[ Info: Time: 345.042 days, iteration: 60, max(|u|): 1.70e-01 ms⁻¹, wmax: 1.04e-03, loc: (352, 21, 19), wall time: 588.031 ms
[ Info: Time: 345.049 days, iteration: 70, max(|u|): 1.92e-01 ms⁻¹, wmax: 1.09e-03, loc: (352, 21, 19), wall time: 573.768 ms
[ Info: Time: 345.056 days, iteration: 80, max(|u|): 2.07e-01 ms⁻¹, wmax: 1.09e-03, loc: (352, 21, 19), wall time: 564.119 ms
[ Info: Time: 345.062 days, iteration: 90, max(|u|): 2.62e-01 ms⁻¹, wmax: 1.12e-03, loc: (353, 21, 15), wall time: 580.637 ms
[ Info: Time: 345.069 days, iteration: 100, max(|u|): 3.13e-01 ms⁻¹, wmax: 1.22e-03, loc: (353, 21, 15), wall time: 567.374 ms
[ Info: Time: 345.076 days, iteration: 110, max(|u|): 3.54e-01 ms⁻¹, wmax: 1.30e-03, loc: (353, 21, 15), wall time: 548.773 ms
[ Info: Time: 345.083 days, iteration: 120, max(|u|): 3.80e-01 ms⁻¹, wmax: 1.37e-03, loc: (352, 21, 19), wall time: 538.128 ms
[ Info: Time: 345.090 days, iteration: 130, max(|u|): 4.06e-01 ms⁻¹, wmax: 1.44e-03, loc: (352, 21, 19), wall time: 555.213 ms
[ Info: Time: 345.097 days, iteration: 140, max(|u|): 4.95e-01 ms⁻¹, wmax: 1.53e-03, loc: (352, 21, 19), wall time: 626.512 ms
[ Info: Time: 345.104 days, iteration: 150, max(|u|): 5.74e-01 ms⁻¹, wmax: 1.68e-03, loc: (352, 21, 19), wall time: 1.675 seconds
[ Info: Time: 345.111 days, iteration: 160, max(|u|): 6.34e-01 ms⁻¹, wmax: 1.84e-03, loc: (352, 21, 19), wall time: 570.976 ms
[ Info: Time: 345.118 days, iteration: 170, max(|u|): 6.65e-01 ms⁻¹, wmax: 1.95e-03, loc: (352, 21, 19), wall time: 578.960 ms
[ Info: Time: 345.125 days, iteration: 180, max(|u|): 6.63e-01 ms⁻¹, wmax: 2.01e-03, loc: (352, 21, 19), wall time: 526.001 ms
[ Info: Time: 345.132 days, iteration: 190, max(|u|): 6.29e-01 ms⁻¹, wmax: 2.05e-03, loc: (352, 21, 19), wall time: 511.413 ms
[ Info: Time: 345.139 days, iteration: 200, max(|u|): 5.69e-01 ms⁻¹, wmax: 2.10e-03, loc: (352, 21, 19), wall time: 522.023 ms
[ Info: Time: 345.146 days, iteration: 210, max(|u|): 4.99e-01 ms⁻¹, wmax: 2.17e-03, loc: (352, 21, 19), wall time: 518.441 ms
[ Info: Time: 345.153 days, iteration: 220, max(|u|): 5.66e-01 ms⁻¹, wmax: 2.25e-03, loc: (352, 21, 19), wall time: 554.285 ms
[ Info: Time: 345.160 days, iteration: 230, max(|u|): 6.29e-01 ms⁻¹, wmax: 2.33e-03, loc: (352, 21, 19), wall time: 530.179 ms
[ Info: Time: 345.167 days, iteration: 240, max(|u|): 6.70e-01 ms⁻¹, wmax: 2.40e-03, loc: (352, 21, 19), wall time: 553.379 ms
[ Info: Time: 345.174 days, iteration: 250, max(|u|): 6.91e-01 ms⁻¹, wmax: 2.47e-03, loc: (352, 21, 19), wall time: 567.494 ms
[ Info: Time: 345.181 days, iteration: 260, max(|u|): 6.95e-01 ms⁻¹, wmax: 2.54e-03, loc: (352, 21, 19), wall time: 552.757 ms
[ Info: Time: 345.188 days, iteration: 270, max(|u|): 7.09e-01 ms⁻¹, wmax: 2.58e-03, loc: (352, 21, 19), wall time: 590.691 ms
...
```

src/Models/HydrostaticFreeSurfaceModels/calculate_hydrostatic_free_surface_tendencies.jl
```diff
@@ -7,7 +7,7 @@ using Oceananigans.Utils: work_layout
 using Oceananigans.Fields: immersed_boundary_condition
 
 """
-    calculate_tendencies!(model::NonhydrostaticModel)
+    calculate_tendencies!(model::HydrostaticFreeSurfaceModel)
```
navidcy, line 10:
```suggestion
    calculate_tendencies!(model::HydrostaticFreeSurfaceModel, callbacks)
```

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
```diff
@@ -24,35 +73,61 @@ end
...
 
 const VITD = VerticallyImplicitTimeDiscretization
 
-@inline dissipation(i, j, k, grid, closure::FlavorOfCATKE{<:VITD}, args...) = zero(eltype(grid))
+@inline function buoyancy_flux(i, j, k, grid, closure::FlavorOfCATKE{<:VITD}, velocities, tracers, buoyancy, diffusivities)
+    wb = ℑzᵃᵃᶜ(i, j, k, grid, buoyancy_fluxᶜᶜᶠ, tracers, buoyancy, diffusivities)
+    
+    #κᶻ = ℑzᵃᵃᶜ(i, j, k, grid, diffusivities.Kᶜ)
+    #N² = ℑzᵃᵃᶜ(i, j, k, grid, ∂z_b, buoyancy, tracers)
+    #wb = - κᶻ * N²
```
navidcy, line 85: what's these commented lines?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
```diff
@@ -94,3 +169,159 @@ end
...
+
+@inline inner_top_tke_flux(i, j, grid, clock, fields, parameters, closure_tuple::Tuple{<:Any, <:Any}, buoyancy) =
+    top_tke_flux(i, j, grid, clock, fields, parameters, closure_tuple[1], buoyancy) + 
+    top_tke_flux(i, j, grid, clock, fields, parameters, closure_tuple[2], buoyancy)
+
+@inline top_tke_flux(i, j, grid, clock, fields, parameters, closure_tuple::Tuple, buoyancy) =
+    top_tke_flux(i, j, grid, clock, fields, parameters, closure_tuple[1], buoyancy) + 
+    inner_top_tke_flux(i, j, grid, clock, fields, parameters, closure_tuple[2:end], buoyancy)
+
+
```
navidcy, line 213:
```suggestion
```

src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
```diff
@@ -1,81 +1,81 @@
...
-* `ν₀` (`Float64` parameter): Convective adjustment viscosity. Default: 0.92
-* `Ri₀ν` (`Float64` parameter): ``Ri`` threshold for decreasing viscosity. Default: -1.34
-* `Riᵟν` (`Float64` parameter): Width over which ``Ri`` decreases to 0. Default: 0.61
-* `κ₀` (`Float64` parameter): Convective adjustment diffusivity for tracers. Default: 0.18
-* `Ri₀κ` (`Float64` parameter): ``Ri`` threshold for decreasing viscosity. Default: -0.13
-* `Riᵟκ` (`Float64` parameter): Width over which ``Ri`` decreases to 0. Default: 0.6
-* `coefficient_z_location` (`Face()` or `Center()`): The vertical location of the diffusivity and viscosity.
-                                                     Default: `Face()`.
-* `Ri_dependent_tapering`: The ``Ri``-dependent tapering. Default: `ExponentialRiDependentTapering()`.
+* `ν₀` (Float64): Non-convective viscosity.
```
navidcy, line 50:
```suggestion
* `Ri_dependent_tapering`: The ``Ri``-dependent tapering. Default: `ExponentialRiDependentTapering()`.
* `ν₀` (Float64): Non-convective viscosity.
```

src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
```diff
@@ -1,81 +1,81 @@
...
+              "realistic ocean conditions or for use in a three-dimensional \n" *
+              "simulation. Use with caution and report bugs and problems with physics \n" *
+              "to https://github.com/CliMA/Oceananigans.jl/issues."
+    end
 
     TD = typeof(time_discretization)
-    LZ = typeof(coefficient_z_location)
-    N = typeof(ν₀)
-    K = typeof(κ₀)
     R = typeof(Ri_dependent_tapering)
```
navidcy, line 76:
```suggestion
```

navidcy:
@glwagner, are any of the things discussed in #2450 being done here? Definitely doing the computation on cell's faces.

navidcy:
Wondering if problems discussed in https://github.com/CliMA/Oceananigans.jl/issues/2453 are also gone with this PR (or they were perhaps gone earlier?).

glwagner:
> @glwagner, are any of the things discussed in #2450 being done here? Definitely doing the computation on cell's faces.

A few of those points are addressed. I don't think we should move TKE to faces now (we seem to get good results with centered TKE). The last point, which is to ensure that "shear production" and "buoyancy flux" account for fluxes from other closures, is important and not addressed. I can fix that fairly easily here.

navidcy:
> > @glwagner, are any of the things discussed in #2450 being done here? Definitely doing the computation on cell's faces.
> 
> A few of those points are addressed. I don't think we should move TKE to faces now (we seem to get good results with centered TKE). The last point, which is to ensure that "shear production" and "buoyancy flux" account for fluxes from other closures, is important and not addressed. I can fix that fairly easily here.

No need to do more in this PR. It's been a big one! I was just checking.

glwagner:
Hmm yeah I agree, let's merge this now and continue work in a new PR.

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
glwagner: Yes this is wrong.

glwagner: It should be

```julia
 update_state!(model::HydrostaticFreeSurfaceModel, callbacks=[]) = update_state!(model, model.grid, callbacks) 
```

Then

```julia
 function update_state!(model::HydrostaticFreeSurfaceModel, grid, callbacks) 
```

and for the single column grid

```julia
 function update_state!(model::HydrostaticFreeSurfaceModel, grid::SingleColumnGrid, callbacks) 
```

navidcy: It's fixed.

PR #2797, opened 2022-11-02
Remove extra memory allocations of `output_construction`

### LINKED ISSUE #2794: Increasing memory use on GPUs with recent versions
I'm not sure if this is known (or even expected) or not, but with recent versions of Oceananigans I've noticed that we're allocating more and more memory on GPUs. Basically the scenario is that I've running simulations with a given script for the past 6 months or so. The simulation and model itself has stayed the same, I've been running on the same GPUs, and the only changes that I made were those forced by changes in Oceanigans. 

When I started this setup, I could run up to ~105 million points. Then as the versions increased I had to decrease to size of simulations to 100 million and then to 95 million. Now with the change to Julia 1.8 (I had been using Julia 1.6 until version 0.77.5) there was a steep increase in memory use and my max size went from 95 to 80 million. Compared to my original size of ~105 million, that's a decrease of about 25% in the grid size.

Has anyone else noticed this? Is this expected since we added more features/flexibility to the code lately?

A few notes:

- My simulation is pretty complex (it's got IBM, forcings for every variable, a tracer, LES closures and drag BCs), so I think for most other simulations the max size would be smaller
- I understand that part of the memory allocation depends on the package dependencies (and the Julia version itself), not being directly controlled by Oceananigans code
- It's also worth noting that since the cluster I use (Casper) hasn't installed Julia 1.8, I started trying the newest version from the pre-compiled binaries for Julia 1.8. I don't know if that can somehow affect the memory allocation since the binaries aren't optimized for that specific machine (I haven't yet tried compiling Julia 1.8 from source)

CC @wenegrat @whitleyv

simone-silvestri:
https://github.com/CliMA/Oceananigans.jl/pull/2740 Introduced extra memory allocation in output construction of ComputedFields to satisfy sliced boundary conditions constraints.

This PR changes the definition of `view(f::Field, indices...)` in order to automatically satisfy sliced boundary condition constraints and avoid unnecessary memory allocation.

closes #2794 ?

### DIFF
--- a/src/Fields/field.jl
+++ b/src/Fields/field.jl
@@ -309,10 +309,12 @@ function Base.view(f::Field, i, j, k)
     # OffsetArray around a view of parent with appropriate indices:
     windowed_data = offset_windowed_data(f.data, loc, grid, window_indices)  
 
+    boundary_conditions = FieldBoundaryConditions(window_indices, f.boundary_conditions)
+
     return Field(loc,
                  grid,
                  windowed_data,
-                 f.boundary_conditions, # keep original boundary conditions
+                 boundary_conditions, # keep original boundary conditions
                  window_indices,
                  f.operand,
                  f.status)
--- a/src/Fields/field.jl
+++ b/src/Fields/field.jl
@@ -314,7 +314,7 @@ function Base.view(f::Field, i, j, k)
     return Field(loc,
                  grid,
                  windowed_data,
-                 boundary_conditions, # keep original boundary conditions
+                 boundary_conditions,
                  window_indices,
                  f.operand,
                  f.status)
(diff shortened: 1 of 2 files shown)


src/Fields/field.jl
```diff
@@ -309,10 +309,12 @@ function Base.view(f::Field, i, j, k)
     # OffsetArray around a view of parent with appropriate indices:
     windowed_data = offset_windowed_data(f.data, loc, grid, window_indices)  
 
+    boundary_conditions = FieldBoundaryConditions(window_indices, f.boundary_conditions)
+
     return Field(loc,
                  grid,
                  windowed_data,
-                 f.boundary_conditions, # keep original boundary conditions
+                 boundary_conditions, # keep original boundary conditions
```
glwagner, line 317:
```suggestion
                 boundary_conditions,
```

glwagner:
This looks like the right change, but I can't figure out the reasoning or pros/cons between this and what we did previously. Can you explain?

simone-silvestri:
The crux of the problem is having `nothing` boundary conditions in the sliced direction for computed fields, because `compute!` calls `fill_halo_regions!`. Before this PR, we were allocating a new field with new boundary conditions in the output construction. this PR reconstructs the field as a field `view` with the correct boundary conditions without having to call `new_data`

(This was not a problem if an `AbstractOperations` was passed to the `OutputWriter` instead of a `Field(operation)`, because in that case the slicing was handled by the output construction directly).

tomchor (approved):

tomchor:
Thanks, @simone-silvestri ! This appears to be working really well!

Issue #2512, opened 2022-05-03
Biogeoceananigans.jl


glwagner:
There's been a bunch of interest in using Oceananigans for physical-biogeochemical interaction studies --- problems where systems of reacting tracers that represent either oceanic biological systems, chemical reactions and cycles, or both interact with turbulence and fluid dynamics simulated by Oceananigans.

In order to organize the community's work (and also to support some work at Clima on biogeochemistry in Oceananigans), I propose that we create a new package that interfaces with `Oceananigans` --- could it be... `Biogeoceananigans.jl`... ? --- to:

1. Facilitate sharing code, and collaboration on implementation and testing of biogeochemistry models to be used in Oceananigans simulations
2. Develop documentation and a suite of examples to illustrate usage, setup, and analysis of numerical experiments with biogeochemistry 

To achieve either of these there's no question we need a _particular place_ to collaborate on re-useable code (rather than working independently). But also, I think the Oceananigans.jl repository is not the best repo to use to achieve the above goals, because it's big and complex, which might make it harder for potential developers to contribute and see their place. I also think it would slow development down, because, for example, we'll have to make sure all unit tests for differential operators pass before we can add a new biogeochemistry model implementation. I think development might be faster and more accessible if we do it in a different repo.

There are also some details to discuss regarding implementation. Oceananigans' design already supports reacting systems via `Forcing`. Oceananigans `Forcing` are arbitrary functions of spatial coordinates, time, prognostic model fields, and forcing function parameters --- or alternatively, indices `i, j, k`, `grid`, `clock`, and a NamedTuple of model fields that can be indexed into arbitrarily. @iuryt and @syou83syou83 (and perhaps others) have experimented in this direction.

[...]

PR #2802, opened 2022-11-04
Adds an interface for inserting biogeochemistry models into Oceananigans models

glwagner:
Plus a simple test with a plankton growth/death model similar to `examples/convecting_plankton.jl`.

TODO:
* [x] Add the interface to `HydrostaticFreeSurfaceModel`
* [x] Better interface for defining the biogeochemical source function
* [x] Better interface for automagically adding biogeochemical tracers?

With @jagoosw


jagoosw:
I've had a play implementing an NPD model in this framework now and think that the other way to define the source functions is preferential:
https://github.com/CliMA/Oceananigans.jl/blob/45971e4b113182958aac43357e2a4633510617db/test/test_biogeochemistry.jl#L33-L37

We could even go one step further and define a model like this:
```
struct NPD
    Kₙ :: Float64
    m :: Float64
    nitrif :: Float64
end

validate_biogeochemistry(::NPD, tracernames) = all([T ∈ tracernames for T in [:N, :P, :D]])

@inline function (model::NPD)(i, j, k, grid, ::Val{:N}, clock, fields)
    P = @inbounds fields.P[i, j, k]
    N = @inbounds fields.N[i, j, k]
    D = @inbounds fields.D[i, j, k]

    return model.nitrif*D - P*N/(N+model.Kₙ) 
end

@inline function (model::NPD)(i, j, k, grid, ::Val{:P}, clock, fields)
    P = @inbounds fields.P[i, j, k]
    N = @inbounds fields.N[i, j, k]
    return P*N/(N+model.Kₙ) - model.m*P
end

@inline function (model::NPD)(i, j, k, grid, ::Val{:D}, clock, fields)
    P = @inbounds fields.P[i, j, k]
    D = @inbounds fields.D[i, j, k]

    return model.m*P - model.nitrif*D
end

@inline (model::NPD)(args...) = 0.0
```
This negates the need to define a `get_biogeochemial_forcing` function, but you do have to define the zero function (last line) and I'm not sure how clear and usable this API is to most users?

navidcy:
related to #2512

jagoosw:
@johnryantaylor do you have any preferences for this API?

glwagner:
We can use abstract types to obviate the need for "zero functions":

```julia
abstract type AbstractBiogeochemistry end

struct NutrientsPlanktonDetritus{FT} <: AbstractBiogeochemistry
    background_nutrients :: FT
    mortality_rate :: FT
    nitrification :: FT
end
```

then with

```julia
(ab::AbstractBiogeochemistry)(i, j, k, grid, val_tracer_name, clock, fields) = zero(grid)
```

users don't need to define the "netural biogeochemical forcing" themselves.

Very Important: always use verbose names! 

I like this interface. Let's figure out if `Val{symbol}` is GPU-friendly.

glwagner:
Instead of `validate_biogeochemistry`, maybe

```julia
required_biogeochemical_tracers(biogeochemistry)
```

is a better syntax. Then users don't have to write their own error messages (we'll handle that in Oceananigans). It also gives us flexibility regarding the choice between 1) automagically adding the biogeochemical tracers or 2) requiring the user to add tracers themselves.

We starting to use tracers for a few important things (TKE, biogeochemistry, buoyancy models). We may also need to come up with a system for handling "name clashes" gracefully. For example, someone might introduce a biogeochemistry model with tracer `e`, not realizing that this would prevent them from using the biogeochemistry model with CATKE. There's a tension between simplicity and readability (which we get with short, intuitive tracer names like `e`, `T`, `S`), and catastrophic "name clashes" that will inevitably occur as model complexity increases.

jagoosw:
Ah I thought we could do something with abstract types for that but couldn't get it to work before. I'll try this on a GPU now if I can get one.

And that makes sense for the validation.

jagoosw:
`test/test_biogeochemistry.jl` as it currently is in this branch runs on GPU so it looks like `Val{Symbol}` does work!

jagoosw:
I've had a go at implementing a proper NPZD model (rather than one I just made up on the fly) and have some thoughts on how we should modify the API:
- I think we need an `required_biogeochemical_auxiliary_fields` like `required_biogeochemical_tracers` because for most models we're going to want the user to at least specify a PAR field (I suppose we may want this to also check the shape of the field because some models may have a pre defined depth dependence of PAR so we might want the user to specify a 2D PAR field rather than doing it properly by integrating a 3D field)
- Given what you said the other day about callbacks only being used for features that should be built into Oceananigans we might want to have a think about how a BGC model can specify the attenuation of PAR. You've mentioned that we could define some kind of integrated field?

Also you can see my implementation [here](https://github.com/OceanBioME/OceanBioME.jl/blob/Oceananigans-Update/src/Models/AdvectedPopulations/NPZD.jl) and a script using it [here](https://github.com/OceanBioME/OceanBioME.jl/blob/Oceananigans-Update/examples/NPZD.jl) since I thought it was probably too complicated for the test (and will change that back to a 1 variable model later). Not finished making it work but yet but will be done soon.

glwagner:
> I think we need an required_biogeochemical_auxiliary_fields like required_biogeochemical_tracers because for most models we're going to want the user to at least specify a PAR field (I suppose we may want this to also check the shape of the field because some models may have a pre defined depth dependence of PAR so we might want the user to specify a 2D PAR field rather than doing it properly by integrating a 3D field)

Should the user specify this, or should the biogeochemical model add it its struct (taking in `grid` for this purpose) and evaluate it during `update_biogeochemical_state`?

glwagner:
> It might be helpful to have a simpler interface for advection in biogeochemical models. Although a user could just add another forcing, I think the only way for a model to automatically add an advective forcing is how I've implimented it in the below example. I Think this works quite well since a lot of BGC models write the sinking terms with the other forcing terms, but its a little cumbersome to write e.g. sinking = div_Uc(i, j, k, grid, bgc.adv_scheme, bgc.u⃗ᵖ, fields.P), and model makers will need to do the setup stuff I've done to make the advective velocity fields.


Can't we add advective terms via the biogeochemical forcing term?

glwagner:
I think we can impose a little more structure that eases biogeochemical model development.

One route is to build out a layer on top of `AbstractBiogeochemistry` for models with a common form, something like

```julia
struct TracerBiogeochemicalModel{BGC, C, A}
    biogeochemical_tracers :: BGC
    settling_advection_schemes :: C
    biogeochemical_model :: A
end
```

and perhaps more properties. This is similar to how `SeawaterBuoyancy` works, in that it provides a concrete structure with a slot `equation_of_state` whereby "external" packages like `SeawaterPolynomials` can insert custom behavior.

Another possibility is to build out a new abstract type below `AbstractBiogeochemistry` with a function-based interface (somehow).

I was also thinking it would be nice to supply a "continuous form" interface, so that model developers can implement functions that look something like

```julia
biogeochemical_transition(x, y, z, t, N, P, Z, D, parameters)
```

rather than having to use the "discrete form".

jagoosw:
> > I think we need an required_biogeochemical_auxiliary_fields like required_biogeochemical_tracers because for most models we're going to want the user to at least specify a PAR field (I suppose we may want this to also check the shape of the field because some models may have a pre defined depth dependence of PAR so we might want the user to specify a 2D PAR field rather than doing it properly by integrating a 3D field)
> 
> Should the user specify this, or should the biogeochemical model add it its struct (taking in `grid` for this purpose) and evaluate it during `update_biogeochemical_state`?

This is a good point, I think doing it as part of the model in update state is a much better solution

jagoosw:
> Can't we add advective terms via the biogeochemical forcing term?
> 
> edit: I see you mentioned that, so I might be missing something... I'll take a look at your examples.
> 
> PS we should add those scripts to `validation/biogeochemistry/`



> I think we can impose a little more structure that eases biogeochemical model development.
> 
> One route is to build out a layer on top of `AbstractBiogeochemistry` for models with a common form, something like
> 
> ```julia
> struct TracerBasedBiogeochemistry
>     biogeochemical_tracers
>     drift_advection_schemes
>     drift_velocities
>     transitions
>     auxiliary_fields
> end
> ```
> 
> and perhaps more properties. This is similar to how `SeawaterBuoyancy` works, in that it provides a concrete structure with a slot `equation_of_state` whereby "external" packages like `SeawaterPolynomials` can insert custom behavior.
> 
> Another possibility is to build out a new abstract type below `AbstractBiogeochemistry` with a function-based interface (somehow).
> 
> I was also thinking it would be nice to supply a "continuous form" interface, so that model developers can implement functions that look something like
> 
> ```julia
> biogeochemical_transition(x, y, z, t, N, P, Z, D, parameters)
> ```
> 
> rather than having to use the "discrete form".
> 
> This sort of structure could also be provided by an external package. The advantage of including it here is that we get tighter coupling with Oceananigans development.

This seems like a good idea, I will have a think and try and come up with something

jagoosw:
Sorry meant to put [skip ci]

jagoosw:
I've added `TracerBasedBiogeochemistry` and an example both using and not using it and they both seem to work now:
npd_example
I am slightly concerned with how I've setup the forcing for `TracerBasedBiogeochemistry` because I assume the weird stuff going on with the interpolation stuff in `ContinuousForcing` is there for a reason (and we can't use the normal continuous forcing stuff because it relies on the index of the tracer which we don't have as the model isn't finished when we setup the biogeochemical model).

I'll have a go at adding an example with PAR integration to check its straightforward to add to a model later.

jagoosw:
Quickly benchmarked stepping the simulation and as setup in `test/test_biogeochemistry_npzd.jl` (just realized this is incorrectly named) got:
```
BenchmarkTools.Trial: 3359 samples with 1 evaluation.
 Range (min … max):  1.162 ms …  11.420 ms  ┊ GC (min … max): 0.00% … 62.52%
 Time  (median):     1.324 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.479 ms ± 842.222 μs  ┊ GC (mean ± σ):  5.25% ±  7.80%

  ▇█▆▄▃▂▂▁                                                    ▁
  ████████▇▆▆▅▄▁▃▁▁▃▃▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▆▆ █
  1.16 ms      Histogram: log(frequency) by time      7.82 ms <

 Memory estimate: 1.52 MiB, allocs estimate: 2522.

```
And in `test/test_biogeochemistry_tracer_based.jl`got:
```
BenchmarkTools.Trial: 978 samples with 1 evaluation.
 Range (min … max):  2.071 ms … 77.292 ms  ┊ GC (min … max): 0.00% …  0.00%
 Time  (median):     3.359 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   5.091 ms ±  4.405 ms  ┊ GC (mean ± σ):  5.49% ± 10.75%

  ▄▂▃▃█▃                                                      
  ███████▄▃▂▂▂▂▁▁▁▂▂▂▂▂▂▂▁▁▁▂▂▃▄▅▄▅▃▃▃▃▃▃▂▂▂▂▂▁▂▂▂▁▁▁▁▁▁▁▁▁▂ ▃
  2.07 ms        Histogram: frequency by time        18.3 ms <

 Memory estimate: 3.84 MiB, allocs estimate: 6731.
```
So there seems to be a lot of room for optimization in how I've written `TracerBasedBiogeochemistry`. The difference when I just benchmarked `time_step!` was slightly less but still ∼2x slower.

(Time to increment stop_iteration by 1 and run! after doing that for an initial step)

Project.toml
```diff
@@ -31,6 +36,7 @@ PencilFFTs = "4a48f351-57a6-4416-9ec4-c37015456aae"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
glwagner, line 39: @jagoosw we don't want these packages in Oceananigans Project.toml. Can you remove them?

glwagner: same for Measures, GLMakie, InteractiveErrors, CairoMakie, BenchmarkTools

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,134 @@
...
+
+    return NamedTuple{keys(sinking_speeds)}(sinking_velocities)
+end
+
+# we can't use the standard `ContinuousForcing` regularisation here because it requires all the tracers to be inplace to have the correct indices
+struct ContinuousBiogeochemicalForcing
+    func
+    params
+    field_dependencies
+end
```
glwagner, line 80: This probably has to be concretely typed

glwagner: Also need `parameters`, not `params`

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,134 @@
...
+
+ContinuousBiogeochemicalForcing(func; parameters=nothing, field_dependencies=()) = ContinuousBiogeochemicalForcing(func, parameters, field_dependencies)
+
+function BiogeochemicalForcing(func; parameters=nothing, discrete_form=false, field_dependencies=())
+    if discrete_form
+        return DiscreteBiogeochemicalForcing(func, parameters)
+    else
+        return ContinuousBiogeochemicalForcing(func, parameters, field_dependencies)
+    end
+end
```
glwagner, line 92: Do we need to re-invent `DiscreteForcing` and `ContinuousForcing`, or can we use them as is?

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,134 @@
...
+
+function regularize_biogeochemical_forcing(forcing::Function)
+    return ContinuousBiogeochemicalForcing(forcing)
+end
+
+regularize_biogeochemical_forcing(forcing) = forcing
+
+@inline getargs(fields, field_dependencies, i, j, k, params::Nothing) =@inbounds map(field_name -> fields[field_name][i, j, k], field_dependencies)
+@inline getargs(fields, field_dependencies, i, j, k, params) = @inbounds tuple(map(field_name -> fields[field_name][i, j, k], field_dependencies)..., params)
```
glwagner, line 101: Doesn't this function exist elsewhere?

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,134 @@
...
+required_biogeochemical_auxiliary_fields(bgc::AbstractBiogeochemistry) = ()
+
+"""Sets up a tracer based biogeochemical model in a similar way to SeawaterBouyancy"""
+struct TracerBasedBiogeochemistry <: AbstractBiogeochemistry
+    biogeochemical_tracers
+    reactions
+    advection_scheme
+    sinking_velocities
+    auxiliary_fields
+end
```
glwagner, line 63: This has to be concretely typed

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,134 @@
...
+        error("$(req_auxiliary_fields) must be among the list of auxiliary fields to use $(typeof(bgc).name.wrapper)")
+    
+    return nothing
+end
+
+required_biogeochemical_tracers(::NoBiogeochemistry) = ()
+required_biogeochemical_auxiliary_fields(bgc::AbstractBiogeochemistry) = ()
+
+"""Sets up a tracer based biogeochemical model in a similar way to SeawaterBouyancy"""
+struct TracerBasedBiogeochemistry <: AbstractBiogeochemistry
```
glwagner, line 57: Is this the right name?

Project.toml
jagoosw: Apologies didn't mean to add these

src/Biogeochemistry.jl
jagoosw: `DiscreteBiogeochemicalForcing` is just an alias for `DiscreteForcing` but we definitly can't use `ContnuousForcing` because it makes interpolation operators dependent on the index of tracers so the full model needs to be setup before you can create the forcing which doesn't work here.

src/Biogeochemistry.jl
jagoosw: Ah I should probably use these like continous forcing does:
https://github.com/CliMA/Oceananigans.jl/blob/0163628e3dc810eed76440cc6ceef9870cc50f7f/src/Operators/interpolation_utils.jl#L34-L42
But I don't really understand why these are defined 6 times?

src/Biogeochemistry.jl
jagoosw: Maybe something just like `BiogeochemicalModel` would be better?

src/Biogeochemistry.jl
glwagner: We can use the same procedure / code that we use for ordinary `forcing`, I think

test/test_biogeochemistry.jl
```diff
@@ -12,6 +12,26 @@ import Oceananigans.Biogeochemistry:
...
+    PAR[i, j, grid.Nz] =  surface_PAR*exp(bgc.water_light_attenuation_coefficient * z - bgc.phytoplankton_light_attenuation_coefficient * ∫chl)
+    for k=grid.Nz-1:-1:1
+        z = grid.zᵃᵃᶜ[k]
+        dz = grid.zᵃᵃᶜ[k+1] - z 
+    
+        ∫chl += P[i, j, grid.Nz]^bgc.phytoplankton_light_attenuation_exponent*dz
+    
+        PAR[i, j, k] =  PAR[i, j, k+1]*exp(- bgc.water_light_attenuation_coefficient * dz - bgc.phytoplankton_light_attenuation_coefficient * ∫chl)
+    end
+end
```
glwagner, line 34: Take care with style issues / inconsistent whitespace above

test/test_biogeochemistry.jl
```diff
@@ -12,6 +12,26 @@ import Oceananigans.Biogeochemistry:
     biogeochemical_advection_scheme, 
     update_biogeochemical_state!
 
+@kernel function update_PAR!(PAR, grid, P, t, bgc) 
+    i, j = @index(Global, NTuple) 
+    
+    surface_PAR = bgc.surface_PAR(t)
+    
+    z = grid.zᵃᵃᶜ[grid.Nz]
```
glwagner, line 22: This only works on rectilinear grids. Use

```suggestion
    z = znode(Center(), Center(), Center(), i, j, grid.Nz, grid)
```

for generality on all grids.

test/test_biogeochemistry.jl
```diff
@@ -12,6 +12,26 @@ import Oceananigans.Biogeochemistry:
...
+    
+    ∫chl = - P[i, j, grid.Nz]^bgc.phytoplankton_light_attenuation_exponent*z
+    PAR[i, j, grid.Nz] =  surface_PAR*exp(bgc.water_light_attenuation_coefficient * z - bgc.phytoplankton_light_attenuation_coefficient * ∫chl)
+    for k=grid.Nz-1:-1:1
+        z = grid.zᵃᵃᶜ[k]
+        dz = grid.zᵃᵃᶜ[k+1] - z
```
glwagner, line 28: Same as above. In general, one should _never_ reference the property `zᵃᵃᶜ`. Instead, use functions to extract nodes and metrics.

test/test_biogeochemistry.jl
glwagner: It also may help to unpack some of the parameters, eg

```julia
n = bgc.phytoplankton_light_attenuation_exponent
```

or whatever concise mathematical notation is used.

src/Biogeochemistry.jl
```diff
@@ -127,22 +127,25 @@ Example
...
 """
-struct SomethingBiogeochemistry{T, S, U, A, P} <: AbstractContinuousFormBiogeochemistry
+struct SomethingBiogeochemistry{T, S, U, A, P, SU} <: AbstractContinuousFormBiogeochemistry
     biogeochemical_tracers :: NTuple{N, Symbol} where N
     transitions :: T
     advection_schemes :: S
     drift_velocities :: U
     auxiliary_fields :: A
     parameters :: P
+    state_updates :: SU
```
glwagner, line 137: Let's see if we can come up with a pattern that avoids the need for this

test/test_biogeochemistry.jl
```diff
@@ -150,9 +153,92 @@ outputs = (w = model.velocities.w,
...
+    mortality_rate = 0.1/day,
+
+    water_light_attenuation_coefficient = 0.12,
+    phytoplankton_light_attenuation_coefficient = 0.06,
+    phytoplankton_light_attenuation_exponent = 0.6,
+    surface_PAR = t -> 100*max(0.0, sin(t*π/(12hours)))
+)
+
+biogeochemistry = SomethingBiogeochemistry(tracers = :P, 
+                                           auxiliary_fields = :PAR,
```
glwagner, line 191: I don't think we want to construct auxiliary fields automagically, because this is restrictive (eg, we have to assume the fields have a certain location?)

I think we should demonstrate "proper" usage of auxiliary fields with non-trivial compute methods; ie we define `compute!` for them. This is vastly preferred especially because it will allow the abstraction to be used both for output and during post processing without boiler plate.

jagoosw: Okay I've had a crack at implimenting integral fields. In case you don't think we want an abstract version of it and it should just be a part of a biogeochemical model (although I'm not sure how you could do that without a lot of what I've written for it), I've put it in a different branch [here](https://github.com/CliMA/Oceananigans.jl/tree/jsw/integral-fields). I'm also not exactly sure how we can define indexing an integral since the whole thing needs to be computed (or at least the columns being indexed).

I've also only done it for $\int^0_{-z_k}F\left[<:\text{Any}, <:\text{Any}, <:\text{Center}\right]dz$ for now because I can't see a way todo it generally so will need to repeat for the different axes/directions/field locations like diriviatives does.

<details>
  <summary>Some rambelling about why I think we have todo it like this (unlike how the Integral field reduction is done)</summary>
  I also just realised that `Integral` is a field reduction already provided giving the total integral along an axis. That is implimented as a sum of `field*volume` so I tried rethinking how I've done it to be as simple as that but don't think its correct todo just `field*volume` in this case because the integral between points at the same location as the field being integrated is an integral over half of two different cells. 
</details>

jagoosw: As I've written it we could then write (for the simple light attenuaiton I implimented in the test)
```julia
PAR_fraction = Field(exp(-∫⁻dz(χ*P^e + k)))
```
Although this might mean we have to add right at the end of the model defimnition after the `tracers = TracerFields...` line?

glwagner: IntegralField is cool! I suggest we merge this PR without that though because I think it will be a fairly extensive effort to make IntegralField work properly and benchmarked and everything (it may require different methods since accumulation is fairly different on CPU vs GPU). Instead we should use this method:

```julia
struct PAROperand
    # parameters here, remember to concretely type
end

# In principle, we can restrict to Center, Center, Center, but we may not want to:
const PARField = Field{<:Any, <:Any, <:Any, <:PAROperand}

import Oceananigans.Fields: compute!

function compute!(par::PARField)
    # define here
end
```

glwagner: PS what does "PAR" stand for?

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,181 @@
...
+
+    return bgc(val_tracer_name, x, y, z, clock.time, fields_ijk...)
+end
+
+@inline (bgc::AbstractContinuousFormBiogeochemistry)(val_tracer_name, x, y, z, t, fields...) = zero(x)
+
+struct NoBiogeochemistry <: AbstractBiogeochemistry end
+
+tracernames(tracers) = keys(tracers)
+tracernames(tracers::Tuple) = tracers
```
glwagner, line 89: Is this function defined in Utils?

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,181 @@
...
+biogeochemistry = Biogeochemistry(tracers = :P, transitions = (; P=growth))
+"""
+struct SomethingBiogeochemistry{T, S, U, A, P, SU} <: AbstractContinuousFormBiogeochemistry
+    biogeochemical_tracers :: NTuple{N, Symbol} where N
+    transitions :: T
+    advection_schemes :: S
+    drift_velocities :: U
+    auxiliary_fields :: A
+    parameters :: P
+    state_updates :: SU
```
glwagner, line 137:
```suggestion
```

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,181 @@
...
+Sets up a tracer based biogeochemical model in a similar way to SeawaterBuoyancy.
+
+Example
+=======
+
+@inline growth(x, y, z, t, P, μ₀, λ, m) = (μ₀ * exp(z / λ) - m) * P 
+
+biogeochemistry = Biogeochemistry(tracers = :P, transitions = (; P=growth))
+"""
+struct SomethingBiogeochemistry{T, S, U, A, P, SU} <: AbstractContinuousFormBiogeochemistry
```
glwagner, line 130:
```suggestion
struct SomethingBiogeochemistry{T, S, U, A, P} <: AbstractContinuousFormBiogeochemistry
```

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,181 @@
...
+@inline required_biogeochemical_auxiliary_fields(bgc::SomethingBiogeochemistry) = bgc.auxiliary_fields
+
+@inline biogeochemical_drift_velocity(bgc::SomethingBiogeochemistry, ::Val{tracer_name}) where tracer_name = tracer_name in keys(bgc.drift_velocities) ? bgc.drift_velocities[tracer_name] : 0
+@inline biogeochemical_advection_scheme(bgc::SomethingBiogeochemistry, ::Val{tracer_name}) where tracer_name = tracer_name in keys(bgc.advection_schemes) ? bgc.advection_schemes[tracer_name] : nothing
+
+@inline update_biogeochemical_state!(bgc::SomethingBiogeochemistry, model) = bgc.state_updates(bgc, model)
+
+@inline (bgc::SomethingBiogeochemistry)(::Val{tracer_name}, x, y, z, t, fields_ijk...) where tracer_name = tracer_name in bgc.biogeochemical_tracers ? bgc.transitions[tracer_name](x, y, z, t, fields_ijk..., bgc.parameters...) : 0.0
+
+function maybe_velocity_fields(drift_speeds)
```
glwagner, line 150: Needs docstring

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,181 @@
...
+@inline nothingfunction(args...) = nothing
+
+function SomethingBiogeochemistry(;tracers,
+                          transitions, 
+                          drift_velocities = NamedTuple(), 
+                          adveciton_schemes = 
+                                NamedTuple{keys(drift_velocities)}(repeat([CenteredSecondOrder()], length(drift_velocities))),  
+                          auxiliary_fields=(), 
+                          parameters=NamedTuple(), 
+                          state_updates = nothingfunction)
```
glwagner, line 173: Style problems, code is hard to read

test/test_biogeochemistry.jl
```diff
@@ -0,0 +1,308 @@
...
```
glwagner, line 191: Should be this eventually:

```suggestion
                                           auxiliary_fields = (; PAR=PARField(grid)), 
```

test/test_biogeochemistry.jl
```diff
@@ -0,0 +1,308 @@
...
+    water_light_attenuation_coefficient = 0.12,
+    phytoplankton_light_attenuation_coefficient = 0.06,
+    phytoplankton_light_attenuation_exponent = 0.6,
+    surface_PAR = t -> 100*max(0.0, sin(t*π/(12hours)))
+)
+
+biogeochemistry = SomethingBiogeochemistry(tracers = :P, 
+                                           auxiliary_fields = :PAR, 
+                                           transitions = (; P=growth),
+                                           state_updates = integrate_light!,
```
glwagner, line 193:
```suggestion
```

test/test_biogeochemistry.jl
glwagner: Also this won't work on the GPU.

test/test_biogeochemistry.jl
jagoosw: Makses sense, after I started I realised it was a bigger challenge than I thought it would be!

I'll try and make these changes later.

PAR stands for photosynthetically available radiation, perhaps it would be more clear to call it light here.

src/Biogeochemistry.jl
jagoosw: Just checked and its in `field_tuples`

johnryantaylor:
I would suggest sticking with PAR rather than light.  PAR is very common in biogeochemistry and it is made up of specific bands of light and hence “light” and PAR aren’t exactly equivalent.  Calling it “light” could cause confusion if someone calculates PAR from the incoming solar radiation (light).

src/Biogeochemistry.jl
glwagner: I guess its simple enough to keep here...

test/test_biogeochemistry.jl
glwagner: or how about `PhotosytheticAvailableRadiation`?

glwagner: `Light` seems pretty simple too though. Or simply `AvailableRadiation`.

glwagner:
> I would suggest sticking with PAR rather than light.  PAR is very common in biogeochemistry and it is made up of specific bands of light and hence “light” and PAR aren’t exactly equivalent.  Calling it “light” could cause confusion if someone calculates PAR from the incoming solar radiation (light).
> […](#)

I agree its good to be precise! Perhaps `PhotosyntheticAvailableRadiation` is best then.

test/test_biogeochemistry.jl
jagoosw: Just trying to impliment this now and the issue is that because `compute`! doesn't get the model the `P` field needs to be defined before the `PAR` field can be. Therefore we either need to manually define the tracers before, or perhaps in `validate_biogeochemistry!` make the PAR field?

glwagner: Okay, in that case we should probably just compute `P` in `update_biogeochemical_state!`.

jagoosw:
In the example with the PAR field I'm having to call `TracerFields` from inside `validate_biogeochemistry` which gives this warning:
```julia
 Warning:     P was assigned an immersed FluxBoundaryCondition: Nothing, but this is not supported on
│     1×1×2 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
│     The immersed boundary condition on P will have no effect.
└ @ Oceananigans.BoundaryConditions ~/.julia/packages/Oceananigans/jMMeT/src/BoundaryConditions/field_boundary_conditions.jl:160
```

I don't think that when theres no boundary conditions this is an issue, but a) the warning is concerning to users, and b)  I'm not sure how this will behave when boundary conditions are set the normal way?

glwagner:
> In the example with the PAR field I'm having to call `TracerFields` from inside `validate_biogeochemistry` which gives this warning:
> 
> ```julia
>  Warning:     P was assigned an immersed FluxBoundaryCondition: Nothing, but this is not supported on
> │     1×1×2 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
> │     The immersed boundary condition on P will have no effect.
> └ @ Oceananigans.BoundaryConditions ~/.julia/packages/Oceananigans/jMMeT/src/BoundaryConditions/field_boundary_conditions.jl:160
> ```
> 
> I don't think that when theres no boundary conditions this is an issue, but a) the warning is concerning to users, and b) I'm not sure how this will behave when boundary conditions are set the normal way?

Can you paste the code snippet that produces the warning here?

jagoosw:
> Can you paste the code snippet that produces the warning here?

I'm not exactly sure which part of this code is producing it but as an MWE, this produces the same error:
```julia
grid = RectilinearGrid(size=(10, 10, 20), extent=(1, 1, 2))
model = NonhydrostaticModel(;grid, tracers=TracerFields((:A, :B), grid))
```

glwagner:
What if we write 

```julia
model = NonhydrostaticModel(; grid, tracers=(:A, :B))
```

?

I think the problem you're seeing is an issue with `TracerFields`. But we should tackle that separately, not in this PR.

jagoosw:
That works fine, but I have to use `TracerFields` in `validate_biogeochemistry` in order to setup the `PAR` field because we need the `P` tracer field to already be defined. 

I'm less concerned now that I realise it works fine when we otherwise specify boundary conditions.

glwagner:
Can we compute PAR in update_biogeochemistry_state! So that we don't have to predefine TracerFields?

jagoosw:
Yeah, I can change it back to be similar to how it was before?

glwagner:
> Yeah, I can change it back to be similar to how it was before?

I'm not sure exactly how it was before --- but if that works, it works! I can also help (next week, it's Thanksgiving right now over here in the US)

jagoosw:
> > Yeah, I can change it back to be similar to how it was before?
> 
> I'm not sure exactly how it was before --- but if that works, it works! I can also help (next week, it's Thanksgiving right now over here in the US)

Done this now, to make it work with `SomethingBiogeochemistry` I had to add the light attenuation model as a property of the `SomethingBiogeochemistry` so it can integrate, I think it works okay?

glwagner:
A few more comments:

* Perhaps we can name `SomethingBiogeochemistry` "BasicBiogeochemistry". I think this is appropriate communication about the purpose of this object; large, sophisticated packages should implement their own models from the ground up.
* I think `light_attenuation_model` is too specific for `BasicBiogeochemistry`. For example, it only pertains to biology; however some models are chemistry-only.

@jagoosw do you mind if I address these, and also adapt `test_biogeochemistry.jl` into a form that's close to merge-ready?

jagoosw:
I think that name makes sense!

For the light attenuaiton model, I'm not sure how it would work without making it a property of the model? If it is a parameter then the update state would need to check if there is a parameter before which might be more ambiguous to set up? 

I've rewritten it a little bit to make the integration a nothing operation when the user doesn't specify a model but not sure what the best solution might be here. Perhaps we could make it more generic so users can call any kind of update state operation?

glwagner:
I'm still sort of worried that we shouldn't have `BasicBiogeochemistry` at all because I'm worried that will tempt child packages into using it... (the double-edged sword of convenience)

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,199 @@
...
+
+"""Ensure that `tracers` contains biogeochemical tracers and `auxiliary_fields` contains biogeochemical auxiliary fields (e.g. PAR)."""
+@inline function validate_biogeochemistry(tracers, auxiliary_fields, bgc, grid, clock)
+    req_tracers = required_biogeochemical_tracers(bgc)
+    tracers = all_fields_present(tracers, req_tracers, grid)
+
+    req_auxiliary_fields = required_biogeochemical_auxiliary_fields(bgc)
+
+    all(field ∈ tracernames(auxiliary_fields) for field in req_auxiliary_fields) ||
+        error("$(req_auxiliary_fields) must be among the list of auxiliary fields to use $(typeof(bgc).name.wrapper)")
```
glwagner, line 115: Do we want required auxiliary fields? I think this makes models harder to implement. It's not necessary, because any biogeochemistry-speciifc auxiliary fields can be computed in `update_biogeochemical_state!`.

jagoosw:
> > If it is a parameter then the update state would need to check if there is a parameter before which might be more ambiguous to set up?
> 
> The user has to define `update_biogeochemical_state!` for the particular `BasicBiogeochemistry` they would like to use; if they want to compute light attenuation, they'll have to write that.
> 
> However, I can envision child packages to Oceananigans providing more structured functionality for users. I think we just want to keep the Oceananigans interface as simple and general as possible since we expected it to be used by many different types of chemistry and biology models.

Ah I didn't realise you could specalise methods for e.g. a particular `BasicBiogeochemistry`

jagoosw:
> I'm still sort of worried that we shouldn't have `BasicBiogeochemistry` at all because I'm worried that will tempt child packages into using it... (the double-edged sword of convenience)

It is also going to be an extra thing to maintain that probably isn't that useful?

glwagner:
> > I'm still sort of worried that we shouldn't have `BasicBiogeochemistry` at all because I'm worried that will tempt child packages into using it... (the double-edged sword of convenience)
> 
> It is also going to be an extra thing to maintain that probably isn't that useful?

That's my main concern. It's basically a convenience thing for model developers who are just getting started. So it is slightly friendly, but it's not a good idea if it slows everyone down...

src/Biogeochemistry.jl
jagoosw: I started rewriting one of the models we've been using using the new interface and think we need required auxiliary fields to check the user is setting up PAR fields for the model.

[Heres](https://github.com/OceanBioME/OceanBioME.jl/blob/Oceananigans-Update/examples/column.jl) an example of a script using it, and [the model code](https://github.com/OceanBioME/OceanBioME.jl/tree/Oceananigans-Update/src/Models/AdvectedPopulations/LOBSTER).

test/test_biogeochemistry.jl
```diff
@@ -0,0 +1,326 @@
...
+=#
+
+#####
+##### Setting up the integration of the Photosynthetically Available Radiation
+#####
+
+struct SimplePhyotosyntheticallyActiveRadiation{FT, SP, P, A}
+    water_light_attenuation_coefficient :: FT
+    phytoplankton_light_attenuation_coefficient :: FT
+    phytoplankton_light_attenuation_exponent :: FT
```
glwagner, line 81: Love the verbosity, very clear!

src/Biogeochemistry.jl
glwagner: I think I understand --- is it because we currently only have access to `model.auxiliary_fields` within a biogeochemical transition function?

We can perhaps generalize this by using an extractor function to get the auxiliary fields for continuous form biogeochemistry, and merging the model auxiliary fields with the biogeochemical auxiliary fields there... what do you think?

glwagner: Those links go to a private repo (but I can see them)

jagoosw: You can get auxiliary fields in normal forcing functions I think, my issue is just that if we can't check if a users put an auxiliary field in then it would probably give some weird errors when we try to integrate the PAR etc.

Do you mean to have a separate property for the models with biogeochemical auxiliary fields?

jagoosw: > Those links go to a private repo (but I can see them)

Oh yeah I forgot, they will be public soon

glwagner: > Do you mean to have a separate property for the models with biogeochemical auxiliary fields?

We can implement an interface function `biogeochemical_auxiliary_fieilds(bgc)` that graps any "additional" auxiliary fields that are not part of `model.auxiliary_fields`, and merges them into the list of fields that we subsequently sift through for the purpose of passing them to the transition functions.

This would then relieve the user from needing to explicitly pass things like `PAR` into `auxiliary_fields` in the model constructor.

jagoosw: Ah I see, so the auxiliary field is part of the bgc instead? I think this make sense

glwagner: Yes, basically `model.biogeochemistry` could then potentially have its own list of auxiliary fields, leaving `model.auxiliary_fields` untouched. (Perhaps `model.auxiliary_fields` would be better called `model.user_auxiliary_fields`.)

jagoosw:
I've had a go at separating biogeochemical auxiliary fields now, does this look like what we want?

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,197 @@
...
+tracernames(tracers) = keys(tracers)
+tracernames(tracers::Tuple) = tracers
+
+@inline function has_biogeochemical_tracers(fields::NamedTuple, required_fields, grid)
+
+    for field_name in required_fields
+        if !(field_name in keys(fields))
+            fields = merge(fields, NamedTuple{(field_name, )}((CenterField(grid), )))
+        end
+    end
```
glwagner, line 101: @jagoosw this is hard to parse. Does this add a `CenterField` when one of the required BGC fields isn't found in a named tuple of tracers (eg because the tracers have been predefined by the user?)

I feel this might lead to subtle bugs in user code, ie when they mispell a name when building a named tuple, the user won't get an error. Instead the model will run but will silently be wrong / incorrect in a suble way that might be hard to figure out.

I suggest for the case of `NamedTuple` we throw a helpful error if all the fields are not provided ("The names x, y, z were not found but need to be included with <biogeochemical model>..." or something.

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,197 @@
...
+
+    # Return tracers and aux fields so that users may overload and
+    # define their own special auxiliary fields (e.g. PAR in test)
+    return tracers, auxiliary_fields 
+end
+
+required_biogeochemical_tracers(::NoBiogeochemistry) = ()
+required_biogeochemical_auxiliary_fields(bgc::AbstractBiogeochemistry) = ()
+
+#=
```
glwagner, line 131: I commented this out. Should we get rid of BasicBiogeochemistry?

validation/biogeochemistry/simple_plankton_continuous_form_biogeochemistry.jl
```diff
@@ -0,0 +1,182 @@
...
+                                   phytoplankton_light_attenuation_coefficient = 0.3,
+                                   phytoplankton_light_attenuation_exponent = 0.6,
+                                   surface_PAR = default_surface_radiation,
+                                   advection_scheme = nothing)
+
+    if sinking_velocity != 0
+        advection_scheme = CenteredSecondOrder()
+    end
+
+    u, v, w = maybe_constant_field.((0.0, 0.0, - sinking_velocity))
```
glwagner, line 49: In general we should try to avoid writing `0.0` because it assumes / promotes to double precision. `0` would work fine here I think.

src/Biogeochemistry.jl
jagoosw: So that a user can add their own tracers, and we automatically add tracers, we probably need to check if all of the bgc tracers are present - or if none of the bgc tracers are present (in which case we add them) - otherwise we error?

src/Biogeochemistry.jl
jagoosw: I don't think its that valuble to have and maintain

src/Biogeochemistry.jl
jagoosw: So:
```
user_set_all_tracers = all(tracer ∈ tracers for tracer in req_tracers)
user_set_some_tracers = any(tracer ∈ tracers for tracer in req_tracers)

if user_set_some_tracers && !user_set_all_tracers
    error("useful error saying that the model uses those tracer names, or that they need to define all of the names")
elseif !user_set_all_tracers
    for ... # as above end
end
```
?

glwagner:
```julia
if !all(name in req_tracers for name in keys(tracers))
    throw(ArgumentError("Helpful error about required tracer fields in tracers NamedTuple"))
end

return nothing
```

src/Biogeochemistry.jl
glwagner: ok!

jagoosw:
@glwagner happy with the solution for tracer adding?

jagoosw:
I don't really know why the tests failed

src/Biogeochemistry.jl
```diff
@@ -92,20 +92,27 @@ struct NoBiogeochemistry <: AbstractBiogeochemistry end
 tracernames(tracers) = keys(tracers)
 tracernames(tracers::Tuple) = tracers
 
-@inline function has_biogeochemical_tracers(fields::NamedTuple, required_fields, grid)
-
-    for field_name in required_fields
-        if !(field_name in keys(fields))
-            fields = merge(fields, NamedTuple{(field_name, )}((CenterField(grid), )))
+add_tracer(tracers::Tuple, name, grid) = tuple(tracers..., name)
+add_tracer(tracers::NamedTuple, name, grid) = merge(tracers, NamedTuple{(name)}((CenterField(grid), )))
```
glwagner, line 96:
```suggestion
add_tracer(tracers::NamedTuple, name, grid) = merge(tracers, NamedTuple(name => CenterField(grid)))
```

glwagner: Maybe `add_biogeochemical_tracer`? It's helpful if the names are more specific when searching the code for functions / names, etc.

jagoosw:
@glwagner I think the tests are failing because the `hydrostatic_turbulent_kinetic_energy_tendency` tracer function is being given `val_tracer_name` when its not expecting it, but when I modify it to take it after `val_tracer_index` it still fails, do you have any ideas?

glwagner:
I'll take a look

jagoosw:
A different error!

glwagner:
@simone-silvestri @navidcy can you review this PR?

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,138 @@
...
+"""
+Update tracer tendencies.
+
+Called at the end of calculate_tendencies!
+"""
+update_biogeochemical_state!(bgc, model) = nothing
+
+@inline biogeochemical_drift_velocity(bgc, val_tracer_name) = nothing
+@inline biogeochemical_advection_scheme(bgc, val_tracer_name) = nothing
+@inline biogeochemical_auxiliary_fieilds(bgc) = NamedTuple()
```
simone-silvestri, line 34: typo in `fields`

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,138 @@
...
+
+"""
+Update tracer tendencies.
+
+Called at the end of calculate_tendencies!
+"""
+update_biogeochemical_state!(bgc, model) = nothing
+
+@inline biogeochemical_drift_velocity(bgc, val_tracer_name) = nothing
+@inline biogeochemical_advection_scheme(bgc, val_tracer_name) = nothing
```
simone-silvestri, line 33: why do we need this and not just a scheme for the tracer?

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,138 @@
...
+#####
+##### Default (discrete form) biogeochemical source
+#####
+
+abstract type AbstractBiogeochemistry end
+
+@inline function biogeochemistry_rhs(i, j, k, grid, bgc, val_tracer_name::Val{tracer_name}, clock, fields) where tracer_name
+    U_drift = biogeochemical_drift_velocity(bgc, val_tracer_name)
+    scheme = biogeochemical_advection_scheme(bgc, val_tracer_name)
+    src = biogeochemical_transition(i, j, k, grid, bgc, val_tracer_name, clock, fields)
```
simone-silvestri, line 45: what is a `biogeochemical_transition`? Shall we put a comment?

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,138 @@
...
+    x = xnode(Center(), Center(), Center(), i, j, k, grid)
+    y = ynode(Center(), Center(), Center(), i, j, k, grid)
+    z = znode(Center(), Center(), Center(), i, j, k, grid)
+
+    return bgc(val_tracer_name, x, y, z, clock.time, fields_ijk...)
+end
+
+@inline (bgc::AbstractContinuousFormBiogeochemistry)(val_tracer_name, x, y, z, t, fields...) = zero(x)
+
+struct NoBiogeochemistry <: AbstractBiogeochemistry end
```
simone-silvestri, line 90: why can't this not be a `Nothing`? (then in the model constructor you just pass a nothing)

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,138 @@
...
+update_biogeochemical_state!(bgc, model) = nothing
+
+@inline biogeochemical_drift_velocity(bgc, val_tracer_name) = nothing
+@inline biogeochemical_advection_scheme(bgc, val_tracer_name) = nothing
+@inline biogeochemical_auxiliary_fieilds(bgc) = NamedTuple()
+
+#####
+##### Default (discrete form) biogeochemical source
+#####
+
```
simone-silvestri, line 39: probably it is good to describe in a docstring what `AbstractBioGeochemistry` is and what is required of a type that subtypes it (the same of `AbstractContinuousFormBiogechimstry`)

src/Biogeochemistry.jl
jagoosw: I think we probably don't need it since the only place a function is dispatched for it is:
```julia
required_biogeochemical_tracers(::NoBiogeochemistry) = ()
```
And I don't think there's a reason we cant just change that to be `bgc` or `::Nothing` like all of the other functions

src/Biogeochemistry.jl
jagoosw: `biogeochemical_transition` is the user defined biogeochemical forcing, I can add a docstring or we could just change to line to:
```julia
src = bgc(i, j, k, grid, val_tracer_name, clock, fields)
```
Since that is what it does anyway?

src/Biogeochemistry.jl
jagoosw: I don't know if there was a reason for this @glwagner, but it would simplify making a bgc model for a user if we just used `model.advection`?

simone-silvestri: I guess the trade-off here is between simplicity and flexibility. Do you envision the biological tracer having any particular advection scheme different than the tracer that needs to be stored in the bcg model?

simone-silvestri: Hmm, ok I see what you mean. If stability is the reasoning I would argue in favor of removing the additional advection scheme also in `AdvectiveForcing`. 

There should be no reason for a velocity to be treated in a `special` way, as the reconstruction (especially if using upwinding) applies to the tracer, not to the velocity.

If there is instability due to dispersion and your tracer is advected in two different ways it might be difficult (and a headache) to understand which of the "different" velocities is causing it.

simone-silvestri: It might actually be even better to just sum the velocities so that the whole upwinding is consistent if different velocities have different signs.

jagoosw: I see what you mean. I think it would take quite a lot of restructuring of the tendency kernels to impliment advective forcing in this way? Based on this I would also guess it would be better to change how backgound velocities are added since this will also be upwinded incorrectly:
```julia
 - div_Uc(i, j, k, grid, advection, velocities, c)
 - div_Uc(i, j, k, grid, advection, background_fields.velocities, c)
```

I'm happy to try and do this but it might make sense as a different PR as it would be quite a lot of changes?

src/Biogeochemistry.jl
jagoosw: Not sure if its exactly what you wanted but had a go at this now

src/Biogeochemistry.jl
simone-silvestri: yeah... I guess advecting with like this something might suffice
```
julia> struct Utot{U, Ub, Ubgc}
       u           :: U
       uBackground :: Ub
       uBiogeochem :: Ubgc
       end

julia> Base.getindex(u::Utot, i, j, k) = u.U[i, j, k] + u.ubackgound[i, j, k] + u.uBiogechem[i, j, k]
```
but you don't have to do it here.

simone-silvestri: I ll open an issue for that

src/Biogeochemistry.jl
jagoosw: We did need this, for continuous form it does something else. I'll add back and add comment.

jagoosw: As below, it turns out we do need this for the continous form:

https://github.com/CliMA/Oceananigans.jl/blob/85920168d1e0214ba9586f9bfa492a8b8aaefd32/src/Biogeochemistry.jl#L113-L126

So I've added a commend explaining what it does

src/Biogeochemistry.jl
simone-silvestri: looks good

src/Biogeochemistry.jl
jagoosw: Found the reason for this, since we dispatch on the the model i.e. `(bgc::AbstractBiogeochemistry)(i, j, ...` we have to have some way to deal with there being no bgc.

The way we had this was to define `NoBiogeochemistry` so that it fell back to `@inline (bgc::AbstractBiogeochemistry)(i, j, k, grid, val_tracer_name, clock, fields) = zero(grid)`.

For consistency with how other optional features are handeled I could instread just add:
`@inline biogeochemistry_rhs(i, j, k, grid, ::Nothing, args...) = zero(grid)` which seems to solve the problem?

jagoosw: This must also be more efficient than the origional `biogeochemsitry_rhs` for `NoBiogeochemistry`

jagoosw: Just checked and this is how no particles are dealt with `update_particle_properties!(::Nothing, model, Δt) = nothing`

jagoosw:
Any other changes needed now?

src/Biogeochemistry.jl
glwagner: I prefer the design we have

glwagner: I agree it'd be better to combine the velocities. The reason we haven't tried that yet is because I regard these as experimental features. So I'd like to let the design mature a bit before adding that complication. I suggest we keep the velocities separate for now.

We'll also have to discuss how to name these velocities.

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,177 @@
...
+ - `biogeochemical_auxiliary_fields(bgc::BiogeochemicalModel)` which returns a `NamedTuple`
+    of the models auxiliary fields (e.g. `(PAR = bgc.light_attenuation.PAR_field, )`)
+ - `biogeochemical_drift_velocity(bgc::BiogeochemicalModel, ::Val{:TRACER_NAME})` which 
+    returns a velocity fields (i.e. a `NamedTuple` of fields with keys `u`, `v` & `w`)
+    for each tracer
+ - `biogeochemical_advection_scheme(bgc::BiogeochemicalModel, ::Val{:TRACER_NAME})` which
+    returns an advection scheme for each tracer.
+ - `update_biogeochemical_state!(bgc::BiogeochemicalModel, model)` (optional) to update the
+    model state
+"""
```
glwagner, line 52: The formatting needs to be fixed

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,177 @@
...
```
glwagner, line 98: This isn't formatted as a docstring

src/Biogeochemistry.jl
simone-silvestri: ok, then we should add a warning to not use upwinding. From what I understood false extrema and non-bounds-preserving algorithms are very detrimental in biogeochemistry models

glwagner: Can we leave it for now and develop a solution after this PR is merged?

src/Biogeochemistry.jl
jagoosw: How should it be for an absract type like this?

glwagner: Same indentation as for a docstring

jagoosw: Tried to fix this now

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -124,8 +124,8 @@ function NonhydrostaticModel(;    grid,
                     diffusivity_fields = nothing,
                        pressure_solver = nothing,
                      immersed_boundary = nothing,
-                      auxiliary_fields = NamedTuple(),
-calculate_only_active_cells_tendencies = false)
+                      auxiliary_fields = NamedTuple()
+    )
```
glwagner, line 128:
```suggestion
                      auxiliary_fields = NamedTuple())
```

src/Biogeochemistry.jl
glwagner: I think we should leave this here... but note that I don't think many other abstract types have detailed docstrings. The pitfall is that the docstring has to be maintained if the interface changes.

src/Biogeochemistry.jl
```diff
@@ -34,25 +34,32 @@ update_biogeochemical_state!(bgc, model) = nothing
...
+
+    - `biogeochemical_advection_scheme(bgc::BiogeochemicalModel, ::Val{:tracer_name})` which
+       returns an advection scheme for each tracer.
+
+    - `update_biogeochemical_state!(bgc::BiogeochemicalModel, model)` (optional) to update the
+        model state.
 """
 abstract type AbstractBiogeochemistry end
 
-@inline biogeochemistry_rhs(i, j, k, grid, ::Nothing, val_tracer_name::Val{tracer_name}, clock, fields) where tracer_name = zero(grid)
```
jagoosw: I think I had to specify it like this to stop the tests failing like they did here

src/Biogeochemistry.jl
```diff
@@ -0,0 +1,195 @@
...
+    AbstractBiogeochemistry
+
+Abstract type for biogeochemical models. To define a biogeochemcial relaionship
+the following functions must have methods defined where `BiogeochemicalModel`
+is a subtype of `AbstractBioeochemistry`:
+
+    - `(bgc::BiogeochemicalModel)(i, j, k, grid, ::Val{:tracer_name}, clock, fields)` which 
+       returns the biogeochemical reaction for for each tracer.
+
+    - `required_biogeochemical_tracers(::BiogeochemicalModel)` which returns a tuple of
```
navidcy, line 41: 2 spaces for markdown

navidcy (approved):
This looks good to me! The docstring for `AbstractBiochemistry` should evolve into a documentation page in the Docs soon! That can be a new PR. :)

glwagner:
@seamanticscience docs for `Biogeochemistry` could be something to work on once this is merged

jagoosw:
Is this ready to merge now?

navidcy:
I think so. @glwagner?

glwagner:
yay!

PR #2808, opened 2022-11-08
Better docstrings for sparse utility functions in `Solvers`

navidcy:


src/Solvers/sparse_approximate_inverse.jl
```diff
@@ -13,53 +13,64 @@ mutable struct SpaiIterator{VF<:AbstractVector, SV<:SparseVector, VI<:AbstractVe
...
 """
-the sparse_approximate_inverse function calculates a SParse Approximate Inverse M ≈ A⁻¹ to be used as a preconditioner
-Since it can be applied to the residual with just a matrix multiplication instead of the solution
-of a triangular linear problem it makes it very appealing to GPU use
+    sparse_approximate_inverse(A::AbstractMatrix; ε::Float64, nzrel)
 
-The algorithm implemeted here calculates M following the specifications found in
+Compute a Sparse Approximate Inverse `M ≈ A⁻¹` to be used as a preconditioner.
+Since it can be applied to the residual with just a matrix multiplication instead
+of the solution of a triangular linear problem it makes it very appealing to use
```
tomchor, line 20:
```suggestion
of the solution of a triangular linear problem, it makes it very appealing to use
```

src/Solvers/sparse_approximate_inverse.jl
```diff
@@ -13,53 +13,64 @@ mutable struct SpaiIterator{VF<:AbstractVector, SV<:SparseVector, VI<:AbstractVe
...
 In particular, the algorithm tries to minimize the Frobenius norm of
 
-‖ A mⱼ - eⱼ ‖ where mⱼ and eⱼ are the jₜₕ column of matrix M and the identity matrix I, respectively
+```
+‖ A mⱼ - eⱼ ‖
+```
 
-Since we are solving for an "sparse approximate" inverse (i.e. a sparse version of the actually non-sparse A⁻¹),
-we start assuming that mⱼ has a sparsity pattern J, which means that
+where `mⱼ` and `eⱼ` denote the `jₜₕ` column of matrix `M` and the identity
```
tomchor, line 33:
```suggestion
where `mⱼ` and `eⱼ` denote the j-th column of matrix `M` and the identity
```

src/Solvers/sparse_approximate_inverse.jl
```diff
@@ -13,53 +13,64 @@ mutable struct SpaiIterator{VF<:AbstractVector, SV<:SparseVector, VI<:AbstractVe
...
+matrix `I`, respectively. Since we are solving for an "sparse approximate" inverse
+(i.e., a sparse version of the actually non-sparse `A⁻¹`), we start assuming that
```
tomchor, line 35: Should this clarification come at the top?

src/Solvers/sparse_preconditioners.jl
```diff
@@ -105,21 +105,24 @@ end
...
+- the 1st order expansion corresponds to `M = D⁻¹(I - (A - D)D⁻¹) ≈ A⁻¹`
 
-all preconditioners are calculated on CPU and then moved to the GPU. 
-Additionally the first order expansion has a method to calculate the preconditioner directly on the GPU
-`asymptotic_diagonal_inverse_preconditioner_first_order(A)` in case of variable time step where the preconditioner
-has to be recalculated often
-"""
+- the 2nd order expansion corresponds to `M = D⁻¹(I - (A - D)D⁻¹ + (A - D)D⁻¹(A - D)D⁻¹) ≈ A⁻¹`
+
+All preconditioners are calculated on CPU and then moved to the GPU.
```
tomchor, line 120:
```suggestion
All preconditioners are calculated on CPU and, if the model is based on a GPU architecture, then moved to the GPU.
```

tomchor (approved):
LGTM

Made some very minor comments that you can feel free to discard if you think they don't improve things

src/Solvers/sparse_approximate_inverse.jl
navidcy: Good point. Moved it and rephrased a bit.

PR #2816, opened 2022-11-09
Fix wrong looping over indices of OffsetArray

### LINKED ISSUE #2815: Dangerous behavior with `@inbounds` within for loop over `1:lenght(ξ::OffSetArray)`
This is a tricky point -- We should be very careful here. It can easily may be a bug!?

https://github.com/CliMA/Oceananigans.jl/blob/91341616c313bc48cf122945a2e138c9442e21de/src/Grids/conformal_cubed_sphere_face_grid.jl#L127-L129

`ξ` and `η` are offset arrays...!

Perhaps it's better to replace `1:length(ξ)` with `eachindex(ξ)`; see https://docs.julialang.org/en/v1/devdocs/boundscheck/#Caution!

navidcy:
~see [https://docs.julialang.org/en/v1/devdocs/boundscheck/\#Caution](https://docs.julialang.org/en/v1/devdocs/boundscheck/#Caution)~

Closes #2815

### DIFF
--- a/src/Grids/conformal_cubed_sphere_face_grid.jl
+++ b/src/Grids/conformal_cubed_sphere_face_grid.jl
@@ -115,7 +115,7 @@ function ConformalCubedSphereFaceGrid(arch::AbstractArchitecture = CPU(),
     ZS = (Zᶜᶜᵃ, Zᶠᶜᵃ, Zᶜᶠᵃ, Zᶠᶠᵃ)
 
     for (ξ, η, X, Y, Z) in zip(ξS, ηS, XS, YS, ZS)
-        for i in 1:length(ξ), j in 1:length(η)
+        for i in 1:Nξ, j in 1:Nη
             @inbounds X[i, j], Y[i, j], Z[i, j] = conformal_cubed_sphere_mapping(ξ[i], η[j])
         end
     end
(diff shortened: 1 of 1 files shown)


glwagner:
The docs link is broken. What does it say?

src/Grids/conformal_cubed_sphere_face_grid.jl
```diff
@@ -115,7 +115,7 @@ function ConformalCubedSphereFaceGrid(arch::AbstractArchitecture = CPU(),
     ZS = (Zᶜᶜᵃ, Zᶠᶜᵃ, Zᶜᶠᵃ, Zᶠᶠᵃ)
 
     for (ξ, η, X, Y, Z) in zip(ξS, ηS, XS, YS, ZS)
-        for i in 1:length(ξ), j in 1:length(η)
+        for i in 1:Nξ, j in 1:Nη
```
glwagner, line 118: What does this have to do with julia docs? I agree that its correct now however, and that previously it was wrong.

navidcy:
> The docs link is broken. What does it say?

Fixed

src/Grids/conformal_cubed_sphere_face_grid.jl
navidcy: It was just a note on how inbounds can be used wrong and give no error.

glwagner:
Ah I see. That advice is about writing generic code, but we don't need that here.

src/Grids/conformal_cubed_sphere_face_grid.jl
glwagner: It's not a problem with inbounds though right? It's a problem with the loop bounds being incorrect.

navidcy:
Right.

src/Grids/conformal_cubed_sphere_face_grid.jl
navidcy: Right

glwagner (approved):

PR #2817, opened 2022-11-10
Improve immersed boundary performance

simone-silvestri:
When using the `GPU`, launching a grid-sized kernel with an immersed boundary implies computing the kernel for all the i, j, and k indices since we use non-branching if conditions.

This is quite detrimental in case of large-compute bound kernels such as tendency calculations.

This PR removes the tendency calculations in the immersed regions by passing an `active_cells_map` that allows to iterate only on wet cells


simone-silvestri:
For large complex domains, we can obtain an excellent speedup. In the case of the quarter-degree ocean (where the roughly 42% of the domain is immersed), the speedup is about 1.47X!

To document the speedup, here is the profile for the `validation/near_global_lat_lon/near_global_quarter_degree.jl` 

On main:
Screen Shot 2022-11-16 at 10 14 53 AM

On this branch:
Screen Shot 2022-11-16 at 10 14 48 AM

The tendency calculations (the four large kernels in the profile) speedup of more than a factor 2!

src/ImmersedBoundaries/ImmersedBoundaries.jl
```diff
@@ -100,17 +100,19 @@ abstract type AbstractImmersedBoundary end
 ##### ImmersedBoundaryGrid
 #####
 
-struct ImmersedBoundaryGrid{FT, TX, TY, TZ, G, I, Arch} <: AbstractGrid{FT, TX, TY, TZ, Arch}
+struct ImmersedBoundaryGrid{FT, TX, TY, TZ, G, I, M, Arch} <: AbstractGrid{FT, TX, TY, TZ, Arch}
     architecture :: Arch
     underlying_grid :: G
     immersed_boundary :: I
+    wet_cells_map :: M
```
glwagner, line 107: What's the different between an "active cell" and a "wet cell"? What is the type of the "map"?

simone-silvestri: The map is a `arch_array` of tuples of unsigned integers, where
```
const MAXUInt8  = 2^8  - 1
const MAXUInt16 = 2^16 - 1
const MAXUInt32 = 2^32 - 1

N = maximum(size(grid))
Type = N > MAXUInt8 ? (N > MAXUInt16 ? (N > MAXUInt32 ? UInt64 : UInt32) : UInt16) : UInt8
```

`wet_cell` is the opposite of `immersed_cell`.  We could rename it to `inactive_cell`. Then we have to figure out what this means for partial cells, but I guess it won't be too difficult

simone-silvestri:
> Awesome!
> 
> How big is `wet_cells_map`? Should we extend `ImmersedBoundaryGrid` so that users can opt in/opt out of this feature?

It is not that large, I figured that with `UInt16` we can cover almost all practical cases. `wet_cell_map` will be a vector of _active_ locations defined by a tuple of indices. Depending on what % of the domain is active, the size of the map will be

`size(grid) * %active * 3 * sizeof(Type)`

where `Type` is the `UInt` with the minimum size required to represent `maximum(size(grid)))`

I guess adding the opt in/opt out can be nice and is very easy because of 

https://github.com/CliMA/Oceananigans.jl/blob/b7835edbac0a7318d49462addb1c1046014e39c4/src/ImmersedBoundaries/wet_cells_map.jl#L10-L21

which is used now to calculate the map only when we construct the model.

simone-silvestri:
@glwagner @navidcy, should be ready to merge (if tests pass)

src/ImmersedBoundaries/ImmersedBoundaries.jl
simone-silvestri: I have renamed all instances of "wet" to "active"

navidcy:
Do it!

navidcy (approved):

simone-silvestri:
@navidcy do you know why documentation keeps failing?

navidcy:
> @navidcy do you know why documentation keeps failing?

Yes. The size repo of OceananigansDocumentation has grown a lot. Will try to fix that.

PR #2818, opened 2022-11-11
Encourage specialization on functions in conditional derivatives

glwagner:
Long ago we used this technique to encourage the compile to inline more. Not sure if it helps these days.

If this does help @whitleyv, we should put it back (many previous instances where we used this pattern have since been changed).


whitleyv:
I don't think you need to change it. I'm running on an older version of Julia, so I didn't update to 0.78.1. But the MEW I added in https://github.com/tomchor/Oceanostics.jl/issues/91 seems to work with 0.78.1 on main without the fix.

glwagner:
Ok! Compiler heuristics change across different versions of Julia too, so it's possible that this pattern (which used to encourage the compiler to specialize on specific function types `f`) may be irrelevant on julia 1.8.

PR #2819, opened 2022-11-11
Simplify the baroclinic adjustment example

navidcy:
w @glwagner

### DIFF
--- a/examples/baroclinic_adjustment.jl
+++ b/examples/baroclinic_adjustment.jl
@@ -16,7 +16,7 @@ using Oceananigans
 using Oceananigans.Units
 
 # ## Grid
-#
+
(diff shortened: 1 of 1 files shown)


glwagner (approved):

simone-silvestri:
It works with no explicit dissipation? Nice

glwagner:
We should check the docs example directly, but it was just fine (at higher resolution) in some tests I did.

francispoulin (approved):

navidcy:
I've run with `Nx, Ny, Nz = 256, 256, 60`:

https://user-images.githubusercontent.com/7112768/201448473-4ca01d6f-bcfc-4f33-b697-7113c6687016.mp4

Seems OK. I'll merge.

PR #2820, opened 2022-11-11
Refactor vector invariant advection

### LINKED ISSUE #2271: Improvements to stretched `WENO5` API and new method `validate_advection`
A few ideas:

1. Allow `grid` as a positional argument so we can write `WENO5(grid)` rather than `WENO5(grid=grid)`
2. Emit a warning about "stretched WENO" in the model constructor rather than `WENO5`. It's misleading to throw a warning when using WENO5 on a uniform grid; we should only throw a warning if users specify `WENO5` with a stretched grid.
3. Alternatively to 2, we can re-build WENO in model constructors with a `validate_advection` method. We already have `validate_momentum_advection` for hydrostatic models:

https://github.com/CliMA/Oceananigans.jl/blob/b3ddbc84c8f35aaf5b93fbbfdb4cffcada5c6533/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl#L127

The downside to 3 is that it's bad for testing, since it might _prevent_ us from using the "uniform" flavor of WENO on stretched grids. (@simone-silvestri is that true?) Also something that's not clear to me --- @simone-silvestri do coefficients like

https://github.com/CliMA/Oceananigans.jl/blob/b3ddbc84c8f35aaf5b93fbbfdb4cffcada5c6533/src/Advection/weno_fifth_order.jl#L30

become `Nothing` in regular directions, even with `WENO5(grid=grid)`? Or are they only `Nothing` for `WENO5()`?

simone-silvestri:
swap orders of WENO reconstruction and VectorInvariant advection so instead of 
```
WENO(grid, vector_invariant = VelocityStencil(), order = 5)
```
we write
```
reconstruction_scheme = WENO(grid, order = 5)
VectorInvariant(scheme = reconstruction_scheme, stencil = VelocityStencil())
```

This means that now `VectorInvariant` works with any reconstruction scheme. I.e., it is possible to write also
```
reconstruction_scheme = UpwindFifthOrder()
VectorInvariant(scheme = reconstruction_scheme)
```

Closes #2271

### DIFF
--- a/src/Advection/vector_invariant_advection.jl
+++ b/src/Advection/vector_invariant_advection.jl
@@ -22,7 +22,6 @@ VectorInvariant(grid; scheme::S = EnstrophyConservingScheme(), stencil = nothing
 const VectorInvariantEnergyConserving    = VectorInvariant{<:EnergyConservingScheme}
 const VectorInvariantEnstrophyConserving = VectorInvariant{<:EnstrophyConservingScheme}
 
-
 const WENOVectorInvariantVel{N, FT, XT, YT, ZT, VI, WF, PP}  = 
       VectorInvariant{WENO{N, FT, XT, YT, ZT, WF, PP}, VI} where {N, FT, XT, YT, ZT, VI<:VelocityStencil, WF, PP}
 const WENOVectorInvariantVort{N, FT, XT, YT, ZT, VI, WF, PP} = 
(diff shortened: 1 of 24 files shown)


glwagner (approved):
excellent

make sure to bump the version before merging

glwagner:
This makes progress towards

https://github.com/CliMA/Oceananigans.jl/issues/2454

It will also allow us to more easily implement new reconstruction schemes to use with vector invariant advection... should we come up with any.

glwagner:
Should we have both `scheme` and `reconstruction`, or are the two identical?

simone-silvestri:
I guess they are identical.

src/Advection/vector_invariant_advection.jl
```diff
@@ -22,14 +22,9 @@ VectorInvariant(; scheme::S = EnstrophyConservingScheme(), stencil = nothing) wh
...
-const WENOVectorInvariantVel{N, FT, XT, YT, ZT, VI, WF, PP}  = 
-      VectorInvariant{<:WENO{N, FT, XT, YT, ZT, WF, PP}, VI} where {N, FT, XT, YT, ZT, VI<:VelocityStencil, WF, PP}
-const WENOVectorInvariantVort{N, FT, XT, YT, ZT, VI, WF, PP} = 
-      VectorInvariant{<:WENO{N, FT, XT, YT, ZT, WF, PP}, VI} where {N, FT, XT, YT, ZT, VI<:VorticityStencil, WF, PP}
-const WENOVectorInvariant{N, FT, XT, YT, ZT, VI, WF, PP} =      
-      VectorInvariant{<:WENO{N, FT, XT, YT, ZT, WF, PP}, VI} where {N, FT, XT, YT, ZT, VI<:SmoothnessStencil, WF, PP}
+const WENOVectorInvariant{VI} = VectorInvariant{<:WENO, VI} where VI
 
-required_halo_size(scheme::WENOVectorInvariant{N}) where N = N + 1
+required_halo_size(scheme::WENOVectorInvariant) = required_halo_size(scheme.advection) + 1
```
glwagner, line 27: 💪

glwagner: isn't it `scheme.scheme` or something now

glwagner: maybe should be `scheme.vorticity_reconstruction` or something

navidcy:
Closing in favor of #2852

PR #2827, opened 2022-11-17
Throw error if halo size is inappropriate for NonhydrostaticModel

### LINKED ISSUE #2107: Should we replace grid if halos are wrong, or throw an error?
Right now in the `NonhydrostaticModel`, we remake a user-provided `grid` if the halos are not big enough:

https://github.com/CliMA/Oceananigans.jl/blob/04ca8e2f143afd53fd60bdaeb885a4dc1ed5825c/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl#L127

This is a nice convenience for users experimenting with different advection schemes, closures, etc.

One major caveat, however, is that models that incorporate user-created fields, when those fields are provided at the time of model construction, may have been created on the "wrong" grid. This problem is most acute for `PrescribedVelocityFields` with the `HydrostaticFreeSurfaceModel`; however the issue is generic.

It doesn't appear to be very common for users to create their own fields, which is perhaps why no great problems have been detected. Nevertheless, this is an insidious "gotcha" that could lead to hard to define or undetectable bugs.

One solution is to throw an error if the grid's halos are not big enough, rather than recreating it. The error will instruct the user to specify larger halos. This damages usability (since users need to know and care about their halos) but makes user experiments more resilient against bugs.

In the future, if we also use large halos by default (eg `(3, 3, 3)` as proposed by #1245) then such an error would be encountered only rarely.

glwagner:
Closes #2107


src/Models/Models.jl
```diff
@@ -16,6 +17,39 @@ architecture(model::AbstractModel) = model.architecture
...
+#####
+##### extract_boundary_conditions:
+#####
+##### Recursive util for building NamedTuples of boundary conditions from NamedTuples of fields.
+#####
+##### Note: ignores tuples, including tuples of Symbols (tracer names) and
+##### tuples of DiffusivityFields (which occur for tupled closures)
+#####
+
+extract_boundary_conditions(::Nothing) = NamedTuple()
```
simone-silvestri, line 42: doesn't this belong in Fields?

glwagner: good idea

glwagner: done

simone-silvestri (approved):

navidcy (approved):

glwagner:
We probably want to revisit once we have #3732

Issue #2058, opened 2021-11-13
Simplify way users have access to grid metrics


navidcy:
We should figure out an interface to help users access grid metrics. We shouldn't have to write things like `grid.Δxᵃᵃᶜ` in the examples.

_Originally posted by @glwagner in https://github.com/CliMA/Oceananigans.jl/issues/2050#issuecomment-967266791_

navidcy:
We need a function like `dx = grid_spacing(grid, 1)` or something which returns a number (if constant) or a field (if not).

_Originally posted by @glwagner in https://github.com/CliMA/Oceananigans.jl/issues/2050#issuecomment-967748805_

tomchor:
> We need a function like `dx = grid_spacing(grid, 1)` or something which returns a number (if constant) or a field (if not).
> 
> _Originally posted by @glwagner in [#2050 (comment)](https://github.com/CliMA/Oceananigans.jl/pull/2050#issuecomment-967748805)_

Wouldn't it be better to have something like `Δx = grid_spacing(model.velocities.v, 1)`? Because I'm assuming the `Δx` has to change depending if they want to use it on a quantity that lies on a `Face` or `Center` node in that direction. Maybe just `Δx = grid_spacing(Face, grid, 1)`, or something akin to the already-existing `nodes` function, that returns all 3 at the same time?

glwagner:
That's correct that this function needs to specify location.

navidcy:
also `, 1)` means nothing... better to be `, :x)`a

glwagner:
It's mirroring Base julia, eg

```julia
julia> a = rand(2, 3, 4)
2×3×4 Array{Float64, 3}:
[:, :, 1] =
 0.264469  0.977549  0.937297
 0.491955  0.710279  0.315507

[:, :, 2] =
 0.243796  0.783644  0.820073
 0.682626  0.351825  0.503396

[:, :, 3] =
 0.726912  0.287171  0.671998
 0.234873  0.993365  0.913062

[:, :, 4] =
 0.507489  0.568977  0.327032
 0.59475   0.547478  0.459589

julia> size(a, 3)
4julia> a = rand(2, 3, 4)
2×3×4 Array{Float64, 3}:
[:, :, 1] =
 0.264469  0.977549  0.937297
 0.491955  0.710279  0.315507

[:, :, 2] =
 0.243796  0.783644  0.820073
 0.682626  0.351825  0.503396

[:, :, 3] =
[...]

PR #2842, opened 2022-11-27
Creates functions for grid spacings

tomchor:
The main goal for this PR is to create a user-facing function to obtain grid spacings.

EDIT:

 The current behavior is that for every grid type there are functions `xspacings()`, `xspacing()`, `xnodes()` and `xnode()` (and similarly for y and z) that follow the pattern illustrated below:

```julia
julia> grid = RectilinearGrid(size=(4, 4, 1), extent=(1, 1, 1))
4×4×1 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.25
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=0.25
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=1.0

julia> xnodes(grid, Center(), Center(), Face())
4-element view(OffsetArray(::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, -2:7), 1:4) with eltype Float64:
 0.125
 0.375
 0.625
 0.875

julia> xnode(2, 2, 2, grid, Center(), Center(), Face())
0.375

julia> xspacings(grid, Center(), Center(), Face())
0.25

julia> xspacing(1, 2, 3, grid, Center(), Center(), Face())
0.25
```

These functions also appropriately handle stretched coordinates. In addition, both `xspacings()` and `xnodes()` have a `with_halos` kwarg that controls whether of not halos points are included.

Convenience functions with fewer arguments are also defined for the cases where that's possible. For example:

```julia
julia> xnode(2,2,2, grid, Center(), Center(), Face())
0.375

julia> xnode(2, grid, Center())
0.375
```

I added tests for these functions for every grid in a way that doesn't create new grid instantiations.


glwagner:
It might make sense to discuss the API on an issue.

src/Grids/rectilinear_grid.jl
```diff
@@ -423,37 +423,21 @@ return_metrics(::RectilinearGrid) = (:xᶠᵃᵃ, :xᶜᵃᵃ, :yᵃᶠᵃ, :y
...
 
-function min_Δx(grid::RectilinearGrid)
-    topo = topology(grid)
-    if topo[1] == Flat
-        return Inf
-    else
-        return min_number_or_array(grid.Δxᶜᵃᵃ)
-    end
-end
+Δx(::Type{Center}, grid::RectilinearGrid) = topology(grid)[1] == Flat ? Inf : grid.Δxᶜᵃᵃ
```
glwagner, line 429: I don't think we want to use these names, and we also don't want to define "one-location" functions. This was a mistake with the xnodes, ynodes, etc functions.

src/ImmersedBoundaries/immersed_grid_metrics.jl
```diff
@@ -35,6 +35,9 @@ end
 
 
 return_metrics(grid::ImmersedBoundaryGrid) = return_metrics(grid.underlying_grid)
+Δx(grid::ImmersedBoundaryGrid) = Δx(grid.underlying_grid)
+Δy(grid::ImmersedBoundaryGrid) = Δy(grid.underlying_grid)
+Δz(grid::ImmersedBoundaryGrid) = Δz(grid.underlying_grid)
```
glwagner, line 40: I don't think we should define no-location versions

glwagner:
Actually there already is an issue! Here:

https://github.com/CliMA/Oceananigans.jl/issues/2058

Put some comments, I propose we come to consensus and then clean up this PR to close that issue.

glwagner:
The issue with `Δx` is that these are already defined as "GridMetricOperation"s, eg objects that can be used within abstract operations. A simple solution is to follow the "nodes" semantics (might be better anyways)

tomchor:
> It might make sense to discuss the API on an issue.

Yeah, sorry, I created this to try some things out so but it should have been marked as a WIP (which I just did now). I agree we need to discuss first before implementing this.

src/ImmersedBoundaries/immersed_grid_metrics.jl
tomchor: Agreed. Fixed.

navidcy:
Let's add tests also?

navidcy:
And use these functions in examples in place of things like $\Delta x^{aaf}$?

tomchor:
> Let's add tests also?

> And use these functions in examples in place of things like Δxaaf?

For sure. I'm just waiting for us to converge on usage in https://github.com/CliMA/Oceananigans.jl/issues/2058 before I include these changes.

test/test_grids.jl
```diff
@@ -388,8 +388,8 @@ function test_correct_tanh_grid_spacings(FT, Nz)
     @test all(isapprox.(  grid.zᵃᵃᶜ[1:Nz],    zᵃᵃᶜ.(1:Nz)   ))
     @test all(isapprox.( grid.Δzᵃᵃᶜ[1:Nz],   Δzᵃᵃᶠ.(1:Nz)   ))
 
-    @test all(isapprox.(zspacing(Face, grid), grid.Δzᵃᵃᶠ[1:Nz]))
-    @test all(isapprox.(zspacing(Center, grid), grid.Δzᵃᵃᶜ[1:Nz]))
+    @test all(isapprox.(zspacing(Face, grid), grid.Δzᵃᵃᶠ))
+    @test all(isapprox.(zspacing(Center, grid), grid.Δzᵃᵃᶜ))
```
glwagner, line 392:
```suggestion
    @test all(isapprox.(zspacing(grid, Face()), grid.Δzᵃᵃᶠ))
    @test all(isapprox.(zspacing(grid, Center()), grid.Δzᵃᵃᶜ))
```

examples/ocean_wind_mixing_and_convection.jl
```diff
@@ -63,8 +63,8 @@ grid = RectilinearGrid(CPU();
 
 fig = Figure(resolution=(1200, 800))
 ax = Axis(fig[1, 1], ylabel = "Depth (m)", xlabel = "Vertical spacing (m)")
-lines!(ax, grid.Δzᵃᵃᶜ[1:grid.Nz], grid.zᵃᵃᶜ[1:grid.Nz])
-scatter!(ax, grid.Δzᵃᵃᶜ[1:Nz], grid.zᵃᵃᶜ[1:Nz])
+lines!(ax, zspacings(grid, Center())[1:grid.Nz], znodes(grid, Center())[1:grid.Nz])
+scatter!(ax, zspacings(grid, Center())[1:Nz], znodes(grid, Center())[1:Nz])
```
glwagner, line 67: Why do we need the `1:Nz`?

navidcy: to avoid showing the spacings for halo points... but can't we do this somehow else?

src/Oceananigans.jl
```diff
@@ -23,6 +23,7 @@ export
     LatitudeLongitudeGrid,
     ConformalCubedSphereFaceGrid,
     xnodes, ynodes, znodes, nodes,
+    xspacings, yspacings, zspacings,
```
navidcy, line 26: we should have a docstring for the exported methods

examples/ocean_wind_mixing_and_convection.jl
tomchor: Like with an optional flag for `zspacing()`?

glwagner: Should we simply define these functions so that they don't include halo points?

navidcy: I think an optional flag would be OK. These are not function that are used in time-stepping so it should be alright (perfomance-wise). E.g., `zspacings(grid, Center(); include_halos = true)` if we want to include the halos; otherwise `include_halos = false` by default?

`interior(zspacings(grid, Center()))` would also be intuitive but this requires that `zspacings` returns a field... and we probably don't wanna go down that road.

glwagner: I think we should return fields eventually but only once we have plotting recipes for Makie.

src/BuoyancyModels/nonlinear_equation_of_state.jl
```diff
@@ -7,15 +7,15 @@ const f = Face()
 
 """ Return the geopotential height at `i, j, k` at cell centers. """
 @inline Zᶜᶜᶜ(i, j, k, grid) =
-    ifelse(k < 1,       znode(i, j,       1, grid, c, c, c) + (1 - k) * Δzᶜᶜᶠ(i, j, 1, grid),
-    ifelse(k > grid.Nz, znode(i, j, grid.Nz, grid, c, c, c) - (k - grid.Nz) * Δzᶜᶜᶠ(i, j, grid.Nz, grid),
-                        znode(i, j,       k, grid, c, c, c)))
+    ifelse(k < 1,       znode(i, j,       1, grid, c, c, c, with_halos=true) + (1 - k) * Δzᶜᶜᶠ(i, j, 1, grid),
+    ifelse(k > grid.Nz, znode(i, j, grid.Nz, grid, c, c, c, with_halos=true) - (k - grid.Nz) * Δzᶜᶜᶠ(i, j, grid.Nz, grid),
+                        znode(i, j,       k, grid, c, c, c, with_halos=true)))
```
glwagner, line 12: I don't think we want a `with_halos` option for the singleton versions (ie functions of `i, j, k`). What would this mean anyways?

src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl
```diff
@@ -324,14 +324,14 @@ const Lᶜᶜᶠ = Tuple{Center, Center, Face}
...
-@inline κᶜᶜᶠ(i, j, k, grid, loc, κ::F, clock, args...) where F<:Function = κ(node(i, j, k, grid, c, c, f)..., clock.time)
+@inline νᶜᶜᶜ(i, j, k, grid, loc, ν::F, clock, args...) where F<:Function = ν(node(i, j, k, grid, c, c, c, with_halos=true)..., clock.time)
+@inline νᶠᶜᶠ(i, j, k, grid, loc, ν::F, clock, args...) where F<:Function = ν(node(i, j, k, grid, f, c, f, with_halos=true)..., clock.time)
+@inline νᶜᶠᶠ(i, j, k, grid, loc, ν::F, clock, args...) where F<:Function = ν(node(i, j, k, grid, c, f, f, with_halos=true)..., clock.time)
+@inline νᶠᶠᶜ(i, j, k, grid, loc, ν::F, clock, args...) where F<:Function = ν(node(i, j, k, grid, f, f, c, with_halos=true)..., clock.time)
+
+@inline κᶠᶜᶜ(i, j, k, grid, loc, κ::F, clock, args...) where F<:Function = κ(node(i, j, k, grid, f, c, c, with_halos=true)..., clock.time)
+@inline κᶜᶠᶜ(i, j, k, grid, loc, κ::F, clock, args...) where F<:Function = κ(node(i, j, k, grid, c, f, c, with_halos=true)..., clock.time)
+@inline κᶜᶜᶠ(i, j, k, grid, loc, κ::F, clock, args...) where F<:Function = κ(node(i, j, k, grid, c, c, f, with_halos=true)..., clock.time)
```
glwagner, line 335: We don't want kwargs in low-level functions that are meant to inline on the GPU

glwagner: This applies to these "coefficient getter" functions, as well as `xnode`, `ynode`, `znode`

src/CubedSpheres/conformal_cubed_sphere_grid.jl
```diff
@@ -186,42 +186,45 @@ end
...
+const OSSG = OrthogonalSphericalShellGrid
 
-@inline φnode(i, j, k, grid::OrthogonalSphericalShellGrid, LX::Face,   LY::Face,   LZ) = @inbounds grid.φᶠᶠᵃ[i, j]
-@inline φnode(i, j, k, grid::OrthogonalSphericalShellGrid, LX::Face,   LY::Center, LZ) = @inbounds grid.φᶠᶜᵃ[i, j]
-@inline φnode(i, j, k, grid::OrthogonalSphericalShellGrid, LX::Center, LY::Face,   LZ) = @inbounds grid.φᶜᶠᵃ[i, j]
-@inline φnode(i, j, k, grid::OrthogonalSphericalShellGrid, LX::Center, LY::Center, LZ) = @inbounds grid.φᶜᶜᵃ[i, j]
+@inline λnode(i, j, k, grid::OSSG, LX::Face,   LY::Face,   LZ; kwargs...) = @inbounds grid.λᶠᶠᵃ[i, j]
+@inline λnode(i, j, k, grid::OSSG, LX::Face,   LY::Center, LZ; kwargs...) = @inbounds grid.λᶠᶜᵃ[i, j]
+@inline λnode(i, j, k, grid::OSSG, LX::Center, LY::Face,   LZ; kwargs...) = @inbounds grid.λᶜᶠᵃ[i, j]
+@inline λnode(i, j, k, grid::OSSG, LX::Center, LY::Center, LZ; kwargs...) = @inbounds grid.λᶜᶜᵃ[i, j]
```
glwagner, line 194: We don't want `kwargs` on low-level functions meant to be in-lined on the GPU (are kwargs supported in GPU kernels?)

src/Grids/latitude_longitude_grid.jl
```diff
@@ -323,13 +323,13 @@ end
 @inline znodes(grid::LatitudeLongitudeGrid, ZL::Face  ; with_halos=false) = with_halos ? grid.zᵃᵃᶠ : view(grid.zᵃᵃᶠ, interior_indices(typeof(ZL), topology(grid, 3), grid.Nz))
 @inline znodes(grid::LatitudeLongitudeGrid, ZL::Center; with_halos=false) = with_halos ? grid.zᵃᵃᶜ : view(grid.zᵃᵃᶜ, interior_indices(typeof(ZL), topology(grid, 3), grid.Nz))
 
-@inline xnode(i, grid::LatitudeLongitudeGrid, XL::CoF) = xnodes(grid, XL)[i]
-@inline ynode(j, grid::LatitudeLongitudeGrid, YL::CoF) = ynodes(grid, YL)[j]
-@inline znode(k, grid::LatitudeLongitudeGrid, ZL::CoF) = znodes(grid, ZL)[k]
+@inline xnode(i, grid::LatitudeLongitudeGrid, XL::CoF; kwargs...) = xnodes(grid, XL; kwargs...)[i]
+@inline ynode(j, grid::LatitudeLongitudeGrid, YL::CoF; kwargs...) = ynodes(grid, YL; kwargs...)[j]
+@inline znode(k, grid::LatitudeLongitudeGrid, ZL::CoF; kwargs...) = znodes(grid, ZL; kwargs...)[k]
```
glwagner, line 328: We don't want to use this pattern for these low-level functions that are supposed to be inlined on the GPU

src/CubedSpheres/conformal_cubed_sphere_grid.jl
tomchor: > We don't want `kwargs` on low-level functions meant to be in-lined on the GPU (are kwargs supported in GPU kernels?)

The GPU tests are passing (with the exception of two for now) so I think they're supported.

Can you explain why `kwargs` should be avoided? I haven't noticed any slowdowns because of that on my tests so I don't know what the downside is (although admittedly I haven't done a rigorous benchmarking routine yet).

src/BuoyancyModels/nonlinear_equation_of_state.jl
tomchor: I introduced those kwargs based on our discussion [here](https://github.com/CliMA/Oceananigans.jl/pull/2842/#discussion_r1065130759). I think that was a nice idea and it makes for fewer (and imho clearer) lines of code. I didn't know there was a downside to using them for GPUs.

src/CubedSpheres/conformal_cubed_sphere_grid.jl
glwagner: It may not matter here, but the compiler cannot specialize on the types of keyword arguments. The "Performance Tips" section of the julia docs used to recommend avoiding them, but I don't see a reference to that now, so the situation may be different...

Still it does not work with our function composition scheme. For example, we cannot take the difference of a node if kwargs are required.

I also don't understand the logic of the API --- why would users have to specify `with_halo=true` or `false` when asking for the node at index `i, j, k`? Does it just control whether a bounds error is thrown or not for the particular index?

Finally, I'm worried that if we do eventually generalize the code so that `xnodes` returns a field, this pattern will have to be refactored, again.

tomchor: > It may not matter here, but the compiler cannot specialize on the types of keyword arguments. 

I took that into consideration and, as it is now, things work as intended and specialize appropriately.

 
> Still it does not work with our function composition scheme. For example, we cannot take the difference of a node if kwargs are required. Composition of operators is the purpose of the `i, j, k, grid, args...` syntax pattern.
> 
> I also don't understand the logic of the API --- why would users have to specify `with_halo=true` or `false` when asking for the node at index `i, j, k`? Does it just control whether a bounds error is thrown or not for the particular index?

Yeah that was my intention. Although I see your point that this might not make much sense.

I'm focusing on get all tests passing for now and then I'll change the behavior. For now I'll change `xnode()` etc to not have any kwargs and alwats return the full array. Although maybe we can keep kwargs for `xnodes` for now and see how it looks before merging? Since this is more of a user-facing function I can see that being useful (e.g. per [the example that motivated this](https://github.com/CliMA/Oceananigans.jl/pull/2842/#discussion_r1065130759)).

src/Fields/abstract_field.jl
```diff
@@ -79,9 +79,9 @@ interior(f::AbstractField) = f
 @propagate_inbounds ynode(i, j, k, ψ::AbstractField) = ynode(i, j, k, ψ.grid, instantiated_location(ψ)...)
 @propagate_inbounds znode(i, j, k, ψ::AbstractField) = znode(i, j, k, ψ.grid, instantiated_location(ψ)...)
 
-xnodes(ψ::AbstractField) = xnodes(ψ.grid, instantiated_location(ψ, 1))
-ynodes(ψ::AbstractField) = ynodes(ψ.grid, instantiated_location(ψ, 2))
-znodes(ψ::AbstractField) = znodes(ψ.grid, instantiated_location(ψ, 3))
+xnodes(ψ::AbstractField) = xnodes(ψ.grid, instantiated_location(ψ)[1])
+ynodes(ψ::AbstractField) = ynodes(ψ.grid, instantiated_location(ψ)[2])
+znodes(ψ::AbstractField) = znodes(ψ.grid, instantiated_location(ψ)[3])
```
glwagner, line 84: We need the three location versions here to be valid on all grids right?

glwagner: Even RectilinearGrid + ImmersedBoundaryGrid will need the three location versions to accomodate all possible immersed boundaries (even partial cells)

tomchor: Right. Good point!

glwagner: Also I think we want to propagate kwargs through this function so we can use `with_halo`

tomchor:
Some of the GPU runs are receiving a cancellation signal before they're done. For example https://buildkite.com/clima/oceananigans/builds/9600#0185e9fa-b03e-4dec-ab71-7ac309d243a5

Does anyone know why?

glwagner:
No idea. Is some nefarious agent cancelling the GPU builds?

navidcy:
Had the same issue the other day on a different PR.

docs/src/model_setup/output_writers.md
```diff
@@ -106,6 +106,7 @@ provided that their `dimensions` are provided:
 
 ```jldoctest
 using Oceananigans
+using Oceananigans.Grids: xnodes_reshaped, ynodes_reshaped
```
glwagner, line 109: should these functions be called "reshaped_xnodes", etc (eg as one would say in english)? Why introduce new names rather than using `xnodes(args... ; reshaped=true)`?

tomchor: I can change the names, no problem. I chose these names because a user can type x, press tab on REPL, and "discover" these functions. But I don't have strong feeling about it.

I created these separate functions because the compiler can't dispatch on keyword arguments. And because of that I was running into problems specializing the `xnodes` functions properly for different grids (given that now we also have the `with_halos` flag).

It's possible to do it, but it would make the `xnodes` code significantly more complex. Since the reshaped version of these functions are only used in some tests and in one `NetCDFOutputWriter`'s example throughout the whole code, it didn't feel like it was worth complicating the main `xnode()` function for it.

In fact, I feel we could nuke these reshaped functions and just reshape arrays manually in the few places in the code we use them.

Thoughts?

glwagner: Note that reshaping is a user feature, not something that's needed internally as far as I know. Reshaping is needed for some plotting libraries (I think Plots and PyPlot), but not for Makie.

glwagner: As for the names, we've attempted to use the convention "write it as you would say it" to make it more natural to speak and communicate about the code. But we could prioritize REPL discovery instead. I think though its important to have consistent style, so we should make that style change across the code base, not just with one set of functions?

glwagner: Can you explain the difficulty in dispatching and point us to the place in the code this manifests?

glwagner: Right, I think we need the function definitions

```julia
function xnodes(grid; with_halos=false, reshape=false)
```

etc

glwagner: One can introduce a design f the form

```julia
# Below is the inner function, defined by specific grids
_xnodes(grid::ParticularGrid) = ...

# This is the outer user-facing function
function xnodes(grid; reshape=false, with_halos=false)
    x = _xnodes(grid)

    if with_halos
        # remove halos
    end

    if reshape
      # reshape x
    end

    return x
end
```

tomchor:
I'm not sure what do to for spacings for `LatLonGrid`s. It has spacings in `x`, `y`, `λ`, `φ`:

```julia
Δxᶜᶜᵃ  Δxᶜᶠᵃ  Δxᶠᶜᵃ  Δxᶠᶠᵃ  Δyᶜᶠᵃ  Δyᶠᶜᵃ  Δzᵃᵃᶜ  Δzᵃᵃᶠ  Δλᶜᵃᵃ  Δλᶠᵃᵃ  Δφᵃᶜᵃ  Δφᵃᶠᵃ
```

Do we want to create separate `xspacings()` and `λspacings()` functions here?

docs/src/model_setup/output_writers.md
tomchor: > Note that reshaping is a user feature, not something that's needed internally as far as I know. Reshaping is needed for some plotting libraries (I think Plots and PyPlot), but not for Makie.

I just tested both of those and was able to plot heatmaps and lines without reshaping. Maybe it was an old version of those packages that needed it? I think since this is a user-feature for plotting, if none of the major plotting libraries require reshaping I vote we simply get rid of it. Are there any other major plotting packages that we could check?

glwagner:
> I'm not sure what do to for spacings for `LatLonGrid`s. It has spacings in `x`, `y`, `λ`, `φ`:
> 
> ```julia
> Δxᶜᶜᵃ  Δxᶜᶠᵃ  Δxᶠᶜᵃ  Δxᶠᶠᵃ  Δyᶜᶠᵃ  Δyᶠᶜᵃ  Δzᵃᵃᶜ  Δzᵃᵃᶠ  Δλᶜᵃᵃ  Δλᶠᵃᵃ  Δφᵃᶜᵃ  Δφᵃᶠᵃ
> ```
> 
> Do we want to create separate `xspacings()` and `λspacings()` functions here?

Yes I think we need both --- one for linear distances (in meters) and one for angular distances (in degrees)

tomchor:
I think this is ready for review. The current behavior is that for every grid type there are functions `xspacings()`, `xspacing()`, `xnodes()` and `xnode()` (and similarly for y and z) that follow the pattern illustrated below:

```julia
julia> grid = RectilinearGrid(size=(4, 4, 1), extent=(1,1,1))
4×4×1 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.25
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=0.25
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=1.0

julia> xnodes(grid, Center(), Center(), Face())
4-element view(OffsetArray(::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, -2:7), 1:4) with eltype Float64:
 0.125
 0.375
 0.625
 0.875

julia> xnode(2,2,2, grid, Center(), Center(), Face())
0.375

julia> xspacings(grid, Center(), Center(), Face())
0.25

julia> xspacing(1, 2, 3, grid, Center(), Center(), Face())
0.25
```

These functions also appropriately handle stretched coordinates. In addition, both `xspacings()` and `xnodes()` have a `with_halos` kwarg that controls whether of not halos points are included.

Convenience functions with fewer arguments are also defined for the cases where that's possible. For example:

```julia
julia> xnode(2,2,2, grid, Center(), Center(), Face())
0.375

julia> xnode(2, grid, Center())
0.375
```

I added tests for these functions for every grid in a way that doesn't create new grid instantiations.

@glwagner the only issue left on my end is what to do about `xnodes_reshaped()`. For now I've renamed it to `reshaped_xnodes()` like you suggested, but I'm reluctant to merge that functionality into `xnodes()` since it would complicate the code and I couldn't find much use for the reshaped array. I've tried plotting different types of plots with different packages and none of them required reshaping the nodes array, so maybe that behavior comes from old versions of some plotting packages.

I'd vote for us keeping things as it is (a separate function called `reshaped_xnodes()` (and same for y, z)). Or to just nuke that function, since reshaping the array isn't hard for the user to do anyway.

tomchor:
@glwagner @navidcy @simone-silvestri can I get a review whenever you have the chance? I'd like to merge this soon so that I can start working on https://github.com/CliMA/Oceananigans.jl/pull/2652. Plus this is already starting to get merge conflicts.

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -801,15 +801,77 @@ function Base.show(io::IO, grid::OrthogonalSphericalShellGrid, withsummary=true)
...
+zspacings(grid::OrthogonalSphericalShellGrid, LZ::CellLocation; with_halos=false) = zspacings(grid)
+@inline zspacing(k, grid::OrthogonalSphericalShellGrid, LZ::CellLocation) = zspacings(grid)
+
+@inline xspacings(grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = xspacings(grid, LX, LY; kwargs...)
+@inline yspacings(grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = yspacings(grid, LX, LY; kwargs...)
+@inline zspacings(grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = zspacings(grid, LZ; kwargs...)
+
+@inline xspacing(i, j, grid::OSSG, LX::CellLocation, LY::CellLocation) = xspacings(grid, LX, LY, with_halos=true)[i, j]
+@inline yspacing(i, j, grid::OSSG, LX::CellLocation, LY::CellLocation) = yspacings(grid, LX, LY, with_halos=true)[i, j]
+@inline zspacing(k, grid::OSSG, LZ::CellLocation)                      = zspacings(grid, LZ)
```
navidcy, line 877: what's the difference of 877 and 869?

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -801,15 +801,77 @@ function Base.show(io::IO, grid::OrthogonalSphericalShellGrid, withsummary=true)
...
+@inline yspacings(grid::OSSG, LX::Center, LY::Face;   with_halos=false) = with_halos ? grid.Δyᶜᶠᵃ :
+    view(grid.Δyᶜᶠᵃ, interior_indices(typeof(LX), topology(grid, 1), grid.Nx), interior_indices(typeof(LY), topology(grid, 2), grid.Ny))
+@inline yspacings(grid::OSSG, LX::Face,   LY::Center; with_halos=false) = with_halos ? grid.Δyᶠᶜᵃ :
+    view(grid.Δyᶠᶜᵃ, interior_indices(typeof(LX), topology(grid, 1), grid.Nx), interior_indices(typeof(LY), topology(grid, 2), grid.Ny))
+@inline yspacings(grid::OSSG, LX::Face,   LY::Face;   with_halos=false) = with_halos ? grid.Δyᶠᶠᵃ :
+    view(grid.Δyᶠᶠᵃ, interior_indices(typeof(LX), topology(grid, 1), grid.Nx), interior_indices(typeof(LY), topology(grid, 2), grid.Ny))
+
+zspacings(grid::OrthogonalSphericalShellGrid) = grid.Δz
+zspacings(grid::OrthogonalSphericalShellGrid, LZ::CellLocation; with_halos=false) = zspacings(grid)
+@inline zspacing(k, grid::OrthogonalSphericalShellGrid, LZ::CellLocation) = zspacings(grid)
```
navidcy, line 869:
```suggestion
zspacings(grid::OSSG) = grid.Δz
zspacings(grid:: OSSG, LZ::CellLocation; with_halos=false) = zspacings(grid)
@inline zspacing(k, grid:: OSSG, LZ::CellLocation) = zspacings(grid)
```
suggest we use the shortcut for homogeneity

src/Grids/orthogonal_spherical_shell_grid.jl
navidcy:
```Julia
julia> using Oceananigans
WARNING: Method definition zspacing(Any, Oceananigans.Grids.OrthogonalSphericalShellGrid{FT, TX, TY, TZ, A, R, Arch} where Arch where R where A where TZ where TY where TX where FT, Union{Oceananigans.Grids.Center, Oceananigans.Grids.Face}) in module Grids at /Users/navid/Research/OC2.jl/src/Grids/orthogonal_spherical_shell_grid.jl:869 overwritten at /Users/navid/Research/OC2.jl/src/Grids/orthogonal_spherical_shell_grid.jl:877.
  ** incremental compilation may be fatally broken for this module **
```

tomchor: Good catch. 877 is redundant.

src/Grids/orthogonal_spherical_shell_grid.jl
tomchor: Done

navidcy:
I'll try to have a look at this today/tomorrow.

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -801,15 +801,76 @@ function Base.show(io::IO, grid::OrthogonalSphericalShellGrid, withsummary=true)
...
+@inline zspacings(grid::OSSG) = grid.Δz
+@inline zspacings(grid::OSSG, LZ::CellLocation; with_halos=false) = zspacings(grid)
+
+@inline xspacings(grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = xspacings(grid, LX, LY; kwargs...)
+@inline yspacings(grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = yspacings(grid, LX, LY; kwargs...)
+@inline zspacings(grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = zspacings(grid, LZ; kwargs...)
+
+@inline xspacing(i, j, grid::OSSG, LX::CellLocation, LY::CellLocation) = xspacings(grid, LX, LY, with_halos=true)[i, j]
+@inline yspacing(i, j, grid::OSSG, LX::CellLocation, LY::CellLocation) = yspacings(grid, LX, LY, with_halos=true)[i, j]
+@inline zspacing(k,    grid::OSSG, LZ::CellLocation)                   = zspacings(grid)
```
navidcy, line 876: does this assume that zspacing for OSSG is regular?

```suggestion
@inline zspacing(k,    grid::OSSG, LZ::CellLocation)                   = zspacings(grid)[k]

```

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -801,15 +801,76 @@ function Base.show(io::IO, grid::OrthogonalSphericalShellGrid, withsummary=true)
...
+@inline znodes(grid::OSSG, LZ::Center; with_halos=false) = with_halos ? grid.zᵃᵃᶜ : view(grid.zᵃᵃᶜ, interior_indices(typeof(LZ), topology(grid, 3), grid.Nz))
+
+
+@inline xnode(i, j, grid::OSSG, LX::CellLocation, LY::CellLocation) = xnodes(grid, LX, LY, with_halos=true)[i, j]
+@inline ynode(i, j, grid::OSSG, LX::CellLocation, LY::CellLocation) = ynodes(grid, LX, LY, with_halos=true)[i, j]
+@inline znode(k, grid::OSSG, LZ::CellLocation)                      = znodes(grid, LZ, with_halos=true)[k]
+
+@inline xnode(i, j, k, grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation) = xnode(i, j, grid, LX, LY)
+@inline ynode(i, j, k, grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation) = ynode(i, j, grid, LY, LY)
+@inline znode(i, j, k, grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation) = znode(k, grid, LZ)
```
navidcy, line 842: I'm a bit confused, e.g., with:

```
@inline znode(k, grid::OSSG, LZ::CellLocation)                      = znodes(grid, LZ, with_halos=true)[k]
@inline znode(i, j, k, grid::OSSG, LX::CellLocation, LY::CellLocation, LZ::CellLocation) = znode(k, grid, LZ)
```

So the second one calls the first and that calls `znodes()` which computes all the nodes only to select then the `k`-th one? Isn't this a bit of a waste?

src/Grids/latitude_longitude_grid.jl
```diff
@@ -315,24 +315,27 @@ function Base.show(io::IO, grid::LatitudeLongitudeGrid, withsummary=true)
...
+@inline xnodes(grid::LatLonGrid, LX::Center; with_halos=false) = with_halos ? grid.λᶜᵃᵃ : view(grid.λᶜᵃᵃ, interior_indices(typeof(LX), topology(grid, 1), grid.Nx))
+@inline ynodes(grid::LatLonGrid, LY::Face  ; with_halos=false) = with_halos ? grid.φᵃᶠᵃ : view(grid.φᵃᶠᵃ, interior_indices(typeof(LY), topology(grid, 2), grid.Ny))
+@inline ynodes(grid::LatLonGrid, LY::Center; with_halos=false) = with_halos ? grid.φᵃᶜᵃ : view(grid.φᵃᶜᵃ, interior_indices(typeof(LY), topology(grid, 2), grid.Ny))
+@inline znodes(grid::LatLonGrid, LZ::Face  ; with_halos=false) = with_halos ? grid.zᵃᵃᶠ : view(grid.zᵃᵃᶠ, interior_indices(typeof(LZ), topology(grid, 3), grid.Nz))
+@inline znodes(grid::LatLonGrid, LZ::Center; with_halos=false) = with_halos ? grid.zᵃᵃᶜ : view(grid.zᵃᵃᶜ, interior_indices(typeof(LZ), topology(grid, 3), grid.Nz))
+
+@inline xnode(i, grid::LatLonGrid, LX::CellLocation) = xnodes(grid, LX; with_halos=true)[i]
+@inline ynode(j, grid::LatLonGrid, LY::CellLocation) = ynodes(grid, LY; with_halos=true)[j]
+@inline znode(k, grid::LatLonGrid, LZ::CellLocation) = znodes(grid, LZ; with_halos=true)[k]
```
navidcy, line 330: same comment as before...  seems a bit cumbersome to compute all nodes and then just select one.

src/Grids/grid_utils.jl
```diff
@@ -225,121 +226,98 @@ index_range_offset(::Colon, loc, topo, halo)          = - interior_parent_offset
...
-
-Examples
-========
+    znodes(grid, LX, LY, LZ, with_halos=false)
 
 ```jldoctest znodes
 julia> using Oceananigans
 
 julia> horz_periodic_grid = RectilinearGrid(size=(3, 3, 3), extent=(2π, 2π, 1), halo=(1, 1, 1),
                                                  topology=(Periodic, Periodic, Bounded));
```
navidcy, line 274:
```suggestion
julia> horz_periodic_grid = RectilinearGrid(size=(3, 3, 3), extent=(2π, 2π, 1), halo=(1, 1, 1),
                                            topology=(Periodic, Periodic, Bounded));
```

navidcy (commented):
If I ask for the nodes on a Flat dimension I get this.

Should we just return nothing?

```Julia
julia> using Oceananigans

julia> grid = RectilinearGrid(size=(10, 12), extent=(1, 1), topology=(Bounded, Bounded, Flat))
10×12×1 RectilinearGrid{Float64, Bounded, Bounded, Flat} on CPU with 3×3×0 halo
├── Bounded  x ∈ [0.0, 1.0]       regularly spaced with Δx=0.1
├── Bounded  y ∈ [0.0, 1.0]       regularly spaced with Δy=0.0833333
└── Flat z

julia> c = CenterField
CenterField (generic function with 2 methods)

julia> c = CenterField(grid)
10×12×1 Field{Center, Center, Center} on RectilinearGrid on CPU
├── grid: 10×12×1 RectilinearGrid{Float64, Bounded, Bounded, Flat} on CPU with 3×3×0 halo
│   └── west: ZeroFlux, east: ZeroFlux, south: ZeroFlux, north: ZeroFlux, bottom: Nothing, top: Nothing, immersed: ZeroFlux
└── data: 16×18×1 OffsetArray(::Array{Float64, 3}, -2:13, -2:15, 1:1) with eltype Float64 with indices -2:13×-2:15×1:1

julia> znodes(c)
StepRangeLen(1.0, 0.0, 1)
```

src/Grids/grid_utils.jl
```diff
@@ -225,121 +226,98 @@ index_range_offset(::Colon, loc, topo, halo)          = - interior_parent_offset
...
+    reshaped_xnodes(loc, grid; kwargs...)
+"""
+function reshaped_xnodes(grid, loc::CellLocation; kwargs...)
+    x = xnodes(grid, loc; kwargs...)
+    return Base.reshape(x, length(x), 1, 1)
+end
+
+"""
+    reshaped_ynodes(loc, grid; kwargs...)
+"""
```
navidcy, line 305: these docstrings are placeholders... we need to fill them up and explain on what these functions actually do, right?

src/Grids/grid_utils.jl
```diff
@@ -225,121 +226,98 @@ index_range_offset(::Colon, loc, topo, halo)          = - interior_parent_offset
...
-`Face` nodes include the boundary points.
-
-Keyword argument
-================
-- `reshape`: With `reshape=false` (default) the output is a 1D array while with 
-  `reshape=true` the output is a 3D array with size `1×1×Nz`.
-
-Examples
-========
+    znodes(grid, LX, LY, LZ, with_halos=false)
```
navidcy, line 268: This docstring seems incomplete. Needs explanation of what this function does, right?

src/Grids/grid_utils.jl
```diff
@@ -225,121 +226,98 @@ index_range_offset(::Colon, loc, topo, halo)          = - interior_parent_offset
...
-`Face` nodes include the boundary points.
+    ynodes(grid, LX, LY, LZ, with_halos=false)
 
-Keyword argument
-================
-- `reshape`: With `reshape=false` (default) the output is a 1D array while with 
-  `reshape=true` the output is a 3D array with size `1×Ny×1`.
-
-See [`znodes`](@ref) for examples.
+See `znodes` for examples.
```
navidcy, line 263: Why did we remove the doc @ref?

```suggestion
See [`znodes`](@ref) for examples.
```
?

src/Grids/grid_utils.jl
```diff
@@ -225,121 +226,98 @@ index_range_offset(::Colon, loc, topo, halo)          = - interior_parent_offset
...
-end
+@inline xnodes(grid, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = xnodes(grid, LX; kwargs...)
 
 """
-    ynodes(loc, grid, reshape=false)
-
-Return a view over the interior `loc=Center` or `loc=Face` nodes
-on `grid` in the ``y``-direction. For `Bounded` directions,
-`Face` nodes include the boundary points.
+    ynodes(grid, LX, LY, LZ, with_halos=false)
```
navidcy, line 261: This docstring seems incomplete. Needs explanation of what this function does, right?

src/Grids/grid_utils.jl
```diff
@@ -225,121 +226,98 @@ index_range_offset(::Colon, loc, topo, halo)          = - interior_parent_offset
...
-
-Return a view over the interior `loc=Center` or `loc=Face` nodes
-on `grid` in the ``x``-direction. For `Bounded` directions,
-`Face` nodes include the boundary points.
-
-Keyword argument
-================
-- `reshape`: With `reshape=false` (default) the output is a 1D array while with 
-  `reshape=true` the output is a 3D array with size `Nx×1×1`.
+    xnodes(grid, LX, LY, LZ, with_halos=false)
```
navidcy, line 254: This docstring seems incomplete. Needs explanation of what this function does, right?

src/Grids/grid_utils.jl
```diff
@@ -225,121 +226,98 @@ index_range_offset(::Colon, loc, topo, halo)          = - interior_parent_offset
...
 
 See `znodes` for examples.
```
navidcy, line 256:
```suggestion
See [`znodes`](@ref) for examples.
```

src/Grids/grid_utils.jl
```diff
@@ -393,6 +385,14 @@ function pop_flat_elements(tup, topo)
     return Tuple(new_tup)
 end
 
+"""
+    min_number_or_array(var)
+
+Return the minimum value of `var`.
+"""
+@inline min_number_or_array(var) = var
+@inline min_number_or_array(var::AbstractVector) = minimum(parent(var))
```
navidcy, line 394: I don't understand why this is needed. `minimum` seems to do the same?

```Julia
julia> minimum([1, 2, 3])
1

julia> minimum(1)
1
```

navidcy: Is the `parent()` important here?

test/test_grids.jl
```diff
@@ -161,18 +163,42 @@ function test_regular_rectilinear_grid_properties_are_same_type(FT)
     return nothing
 end
 
-function test_xnode_ynode_znode_are_correct(FT)
+function test_regular_rectilinear_xnode_ynode_znode_and_spacings_are_correct(FT)
```
navidcy, line 166:
```suggestion
function test_regular_rectilinear_xnode_ynode_znode_and_spacings_are_correct(arch, FT)
```

test/test_grids.jl
```diff
@@ -161,18 +163,42 @@ function test_regular_rectilinear_grid_properties_are_same_type(FT)
...
     N = 3
     grid = RectilinearGrid(CPU(), FT, size=(N, N, N), x=(0, π), y=(0, π), z=(0, π),
```
navidcy, line 168:
```suggestion
    grid = RectilinearGrid(arch, FT, size=(N, N, N), x=(0, π), y=(0, π), z=(0, π),
```

navidcy (commented):
left a few more comments

src/Grids/latitude_longitude_grid.jl
navidcy: This appears many times actually.

src/Grids/grid_utils.jl
tomchor: @glwagner had comments about this function (thread starts [here](https://github.com/CliMA/Oceananigans.jl/pull/2842#discussion_r1087425320)) and it's not clear to be that it'll remain there. So yes, for now that's a placeholder. Sorry for not making that clear

src/Grids/orthogonal_spherical_shell_grid.jl
tomchor: Yes, that's what it is. It's by design since if we change the name of `Δzᵃᵃᶠ` we just need to do it once in `znodes()`. So my idea was to help code maintainability.

Plus, if I understand it correctly, the spacings `grid.Δzᵃᵃᶠ` etc. for all grids are calculated once and stored in `grid`, so we're not actually computing them each time we call `zspacing()`.

navidcy: I'm sorry but I don't understand.

What was the problem with the previous way `znode()` was implemented?

Perhaps I got into this discussion late and that's why I don't understand...?

test/test_grids.jl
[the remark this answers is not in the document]
tomchor: That's a good point. I'll figure out how to add a GPU test.

src/Grids/grid_utils.jl
tomchor: Right! Thanks for catching that. I just included an explanation

src/Grids/grid_utils.jl
tomchor: Done

src/Grids/grid_utils.jl
tomchor: Added explanation

src/Grids/grid_utils.jl
tomchor: I didn't write that code (I just moved it), so I don't know why `parent()` is important. It also seems to me like we can use just `minimum`. I can change that line and we see if the tests pass!

src/Grids/orthogonal_spherical_shell_grid.jl
tomchor: It may also just be a bad decision on my part. But just to explain, I'll take `z`-nodes on a `Center` for a `RectilinearGrid` as an example.  Currently the nodes functions are defined here:

https://github.com/CliMA/Oceananigans.jl/blob/09ff3a9865fefde7921c63321073c61406f64bb5/src/Grids/rectilinear_grid.jl#L364-L376

The old `all_z_nodes(Center, grid)` is what's now `znodes(grid, Center())`.

At some point (I think on https://github.com/CliMA/Oceananigans.jl/issues/2058) @glwagner suggested that we should have one method for `znode()` and `znodes()` that take all three cell locations (i.e. `LX`, `LY`, `LZ`; to maintain some uniformity) and whenever possible that we should have another method that takes only the necessary location for the `z` position to be retrieved (`LZ` for a rect grid). So to implement that we need the methods

```julia
znodes(grid, LZ)
znodes(grid, LX, LY, LZ)
znode(k, grid, LZ)
znode(i, j, k, grid, LX, LY, LZ)
```

We could implement them as something like

```julia
znodes(grid, LZ)  = grid.zᵃᵃᶠ
znodes(grid, LX, LY, LZ)  = grid.zᵃᵃᶠ
znode(k, grid, LZ) =  grid.zᵃᵃᶠ[k]
znode(i, j, k, grid, LX, LY, LZ) = grid.zᵃᵃᶠ[k]
```

But my thinking was that, apart from being repetitive, if we decide to change the name of  `zᵃᵃᶠ` to something else in the future, we need to change that in at least 4 places. Which is why I chose to implement this as

```julia
znodes(grid, LZ)  = grid.zᵃᵃᶠ
znodes(grid, LX, LY, LZ)  = znodes(grid, LZ)
znode(k, grid, LZ) =  znodes(grid, LZ)[k]
znode(i, j, k, grid, LX, LY, LZ) = znode(k, grid, LZ)
```

That way the broader functions build on top of the more specialized ones. It's a bit more complicated for the spacing functions (since they can be a `Number` or an `Array` depending if the direction is stretched or not), but the same reasoning goes. 

We can definitely change how these are implemented in this PR, but that was my rationale for doing it the way I did. @navidcy I hope this helps you understand better! Let me know what needs more clarification.

navidcy: Why not something like:

```Julia
znodes(grid, ::Face)  = grid.zᵃᵃᶠ
znodes(grid, ::Center)  = grid.zᵃᵃᶜ
znodes(grid, LX, LY, LZ)  = znodes(grid, LZ)

znode(k, grid, LZ) =  grid.zᵃᵃᶠ[k]
znode(i, j, k, grid, LX, LY, LZ) = znode(k, grid, LZ) 
```

tomchor: Yeah I can do that too, np. Does this have advantages over the way it's currently written?

navidcy: it reads more natural in my head... 

I feel bad making suggestions if things have been discussed already.

@glwagner you have an opinion on this? it's just that e.g.
```Julia
znode(k, grid, LZ) = znodes(grid, LZ, with_halos=true)[k]
```
feels weird --- is it just me?

tomchor: > I feel bad making suggestions if things have been discussed already.

These minutiae weren't discussed. Just the methods. So you're in a good spot for making suggestions!

navidcy: @tomchor mind if I push a few things?

I started having to add some of this functionality in #2867 so it'd be better to finish this otherwise it'd be double work. ;)

tomchor: Go for it!

navidcy:
ok, let's see how many things I broke now...

(I'll see what happens with the tests and try to fix them...!)

src/Grids/grid_utils.jl
```diff
@@ -363,12 +337,26 @@ function nodes(loc, grid::AbstractGrid; reshape=false)
...
+##### << Spacings >>
+#####
+
+@inline xspacing(i, j, k, grid, LX::CellLocation, LY::CellLocation, LZ::CellLocation) = xspacing(i, grid, LX)
+@inline yspacing(i, j, k, grid, LX::CellLocation, LY::CellLocation, LZ::CellLocation) = yspacing(j, grid, LY)
+@inline zspacing(i, j, k, grid, LX::CellLocation, LY::CellLocation, LZ::CellLocation) = zspacing(k, grid, LZ)
+
+@inline xspacings(grid, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = xspacings(grid, LX; kwargs...)
+@inline yspacings(grid, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = yspacings(grid, LY; kwargs...)
+@inline zspacings(grid, LX::CellLocation, LY::CellLocation, LZ::CellLocation; kwargs...) = zspacings(grid, LZ; kwargs...)
```
navidcy, line 357:
```suggestion
```

As written these methods assume that `x/y/zspacings` return 1D arrays...
Was the idea to have them as general fallbacks and only define new methods where needed?

src/Grids/grid_utils.jl
```diff
@@ -393,6 +381,14 @@ function pop_flat_elements(tup, topo)
     return Tuple(new_tup)
 end
 
+"""
+    min_number_or_array(var)
+
+Return the minimum value of `var`.
+"""
+@inline min_number_or_array(var) = var
+@inline min_number_or_array(var::AbstractVector) = minimum(parent(var))
```
navidcy, line 390:
```suggestion
```

src/Grids/grid_utils.jl
navidcy:
```Julia
julia> using CUDA, OffsetArrays

julia> a = 2.0
 2.0

julia> b = CUDA.randn(4, 4)
4×4 CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}:
 -0.0972989  -0.0783534   0.167717   1.1295
  0.0279738   0.682576    0.238343  -0.158799
  1.0389     -0.49114     1.74628   -0.892839
  0.682562   -1.47588    -0.7601     1.26731

julia> c = OffsetArray(b, -3, -2)
4×4 OffsetArray(::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, -2:1, -1:2) with eltype Float32 with indices -2:1×-1:2:
┌ Warning: Performing scalar indexing on task Task (runnable) @0x00007fa4c226d5a0.
│ Invocation of getindex resulted in scalar indexing of a GPU array.
│ This is typically caused by calling an iterating implementation of a method.
│ Such implementations *do not* execute on the GPU, but very slowly on the CPU,
│ and therefore are only permitted from the REPL for prototyping purposes.
│ If you did intend to index this array, annotate the caller with @allowscalar.
└ @ GPUArraysCore ~/.julia/packages/GPUArraysCore/B3xv7/src/GPUArraysCore.jl:103
 -0.0972989  -0.0783534   0.167717   1.1295
  0.0279738   0.682576    0.238343  -0.158799
  1.0389     -0.49114     1.74628   -0.892839
  0.682562   -1.47588    -0.7601     1.26731

julia> minimum(a)
2.0

julia> minimum(b)
-1.4758823f0

julia> minimum(c)
┌ Warning: Performing scalar indexing on task Task (runnable) @0x00007fa674cec010.
│ Invocation of getindex resulted in scalar indexing of a GPU array.
│ This is typically caused by calling an iterating implementation of a method.
│ Such implementations *do not* execute on the GPU, but very slowly on the CPU,
│ and therefore are only permitted from the REPL for prototyping purposes.
│ If you did intend to index this array, annotate the caller with @allowscalar.
└ @ GPUArraysCore ~/.julia/packages/GPUArraysCore/B3xv7/src/GPUArraysCore.jl:103
-1.4758823f0
```

navidcy: I believe `min_number_or_array` is redundant. I'm nuking it.

src/Coriolis/non_traditional_beta_plane.jl
```diff
@@ -71,17 +71,17 @@ using Oceananigans.Grids: ynode, znode
...
-      - two_Ωᶻ(coriolis, ynode(Center(), j, grid), znode(Center(), k, grid)) * ℑyᵃᶜᵃ(i, j, k, grid, U.v))
+     (  two_Ωʸ(coriolis, ynode(j, grid, Center()), znode(k, grid, Center())) * ℑzᵃᵃᶜ(i, j, k, grid, U.w)
+      - two_Ωᶻ(coriolis, ynode(j, grid, Center()), znode(k, grid, Center())) * ℑyᵃᶜᵃ(i, j, k, grid, U.v))
 
 @inline x_f_cross_U(i, j, k, grid, coriolis::NonTraditionalBetaPlane, U) =
     ℑxᶠᵃᵃ(i, j, k, grid, two_Ωʸw_minus_two_Ωᶻv, coriolis, U)
 
 @inline y_f_cross_U(i, j, k, grid, coriolis::NonTraditionalBetaPlane, U) =
-    two_Ωᶻ(coriolis,ynode(Face(), j, grid), znode(Center(), k, grid)) * ℑxyᶜᶠᵃ(i, j, k, grid, U.u)
+    two_Ωᶻ(coriolis,ynode(j, grid, Face()), znode(k, grid, Center())) * ℑxyᶜᶠᵃ(i, j, k, grid, U.u)
```
navidcy, line 81: this assumes a specific grid; better use the generic `ynode(i, j, k, grid,...) ?

src/Coriolis/non_traditional_beta_plane.jl
```diff
@@ -71,17 +71,17 @@ using Oceananigans.Grids: ynode, znode
...
 
 @inline z_f_cross_U(i, j, k, grid, coriolis::NonTraditionalBetaPlane, U) =
-   -two_Ωʸ(coriolis, ynode(Center(), j, grid), znode(Face(), k, grid)) * ℑxzᶜᵃᶠ(i, j, k, grid, U.u)
+   -two_Ωʸ(coriolis, ynode(j, grid, Center()), znode(k, grid, Face())) * ℑxzᶜᵃᶠ(i, j, k, grid, U.u)
```
navidcy, line 84: this assumes a specific grid; better use the generic `ynode(i, j, k, grid,...)` ?

glwagner: Yes. All internal should use the generic. I'm even skpetical we should have the convenience versions because it lets users write non-generic code.

src/Coriolis/non_traditional_beta_plane.jl
glwagner: Yes

src/Grids/grid_utils.jl
glwagner: Why do we need the type annotations?

glwagner: like, what's the point of writing `::CellLocation`?

src/Grids/grid_utils.jl
```diff
@@ -224,122 +217,103 @@ index_range_offset(::Colon, loc, topo, halo)          = - interior_parent_offset
...
+    return Base.reshape(y, 1, length(y), 1)
+end
+
+"""
+    reshaped_znodes(loc, grid; kwargs...)
+"""
+function reshaped_znodes(grid, loc::CellLocation; kwargs...)
+    z = znodes(grid, loc; kwargs...)
+    return Base.reshape(z, 1, 1, length(z))
+end
```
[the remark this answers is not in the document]
glwagner: yeah we need generic code

src/Grids/grid_utils.jl
navidcy: true, will clean

src/Coriolis/non_traditional_beta_plane.jl
navidcy: fixed with 646a0fc4e

src/Coriolis/non_traditional_beta_plane.jl
navidcy: fixed with 646a0fc4e

src/Fields/abstract_field.jl
```diff
@@ -71,19 +71,19 @@ interior(f::AbstractField) = f
 ##### Coordinates of fields
 #####
 
-@propagate_inbounds xnode(i, ψ::AbstractField{LX, LY, LZ}) where {LX, LY, LZ} = xnode(LX(), i, ψ.grid)
-@propagate_inbounds ynode(j, ψ::AbstractField{LX, LY, LZ}) where {LX, LY, LZ} = ynode(LY(), j, ψ.grid)
-@propagate_inbounds znode(k, ψ::AbstractField{LX, LY, LZ}) where {LX, LY, LZ} = znode(LZ(), k, ψ.grid)
+@propagate_inbounds xnode(i, ψ::AbstractField{LX, LY, LZ}) where {LX, LY, LZ} = xnode(i, ψ.grid, LX())
+@propagate_inbounds ynode(j, ψ::AbstractField{LX, LY, LZ}) where {LX, LY, LZ} = ynode(j, ψ.grid, LY())
+@propagate_inbounds znode(k, ψ::AbstractField{LX, LY, LZ}) where {LX, LY, LZ} = znode(k, ψ.grid, LZ())
```
navidcy, line 76: this is not grid-agnostic

navidcy: delete?

src/Grids/grid_utils.jl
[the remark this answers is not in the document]
tomchor: @glwagner we were discussing the need for these functions before, no? I couldn't find the thread exactly but basically I couldn't find a use for these functions and I think we should delete them.

You mentioned they were there because `Plots.jl` or some other plotting library needs arguments of a particular shape. I tried a few of these plotting libraries and found that they work just fine without these reshaped versions of `nodes`.

src/Grids/grid_utils.jl
glwagner: It's bad style in general to overrestrict method generality. If we add another location type, code that uses that new type will fail unless we update `CellLocation`. So overrestriction of methods makes the code harder to develop and generally requires more code to be written for new features. That's why it's against YASGuide, which we loosely follow in Oceananigans. Basically, you are making an assumption that you always know what valid locations are. But you don't.

glwagner: Another way to think about this convention is that you should try to avoid relying `MethodError` to catch bugs. We do violate this convention in a few places (mostly typing keyword arguments) because the alternative is just too onerous and its very important.

glwagner: In this case, the only reason `xnodes(grid, 1)` appears is because you are making the specific changes in this PR. So you are using a very temporary syntax error to justify a permanent code change. Is that a good justificaiton?

src/Grids/grid_utils.jl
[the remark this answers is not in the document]
navidcy: I'm deleting them. If a user want to reshape they can do that as they wish :)

src/Fields/abstract_field.jl
glwagner: We can probably even delete these, I don't think they are useful.

src/Grids/grid_utils.jl
glwagner: See the first bullet point here:

https://github.com/jrevels/YASGuide#programming-guidelines

tomchor: > In this case, the only reason `xnodes(grid, 1)` appears is because you are making the specific changes in this PR. So you are using a very temporary syntax error to justify a permanent code change. Is that a good justificaiton?

I only _noticed_ it when changing this PR. It is, however, still on `main`. I just can't remember where, but I'll link it here if I find it.

src/Grids/grid_utils.jl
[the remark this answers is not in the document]
glwagner: great

test/test_grids.jl
```diff
@@ -198,13 +198,13 @@ function test_regular_rectilinear_xnode_ynode_znode_and_spacings(arch, FT)
...
+        @test all(xspacings(grid, Center()) .≈ FT(π/N))
+        @test all(yspacings(grid, Center()) .≈ FT(π/N))
+        @test all(zspacings(grid, Center()) .≈ FT(π/N))
 
-        @test xspacings(grid, Face()) ≈ FT(π/N)
-        @test yspacings(grid, Face()) ≈ FT(π/N)
-        @test zspacings(grid, Face()) ≈ FT(π/N)
+        @test all(xspacings(grid, Face()) .≈ FT(π/N))
+        @test all(yspacings(grid, Face()) .≈ FT(π/N))
+        @test all(zspacings(grid, Face()) .≈ FT(π/N))
```
glwagner, line 207: Could also write this (saving some memory)

```suggestion
        @test all(x ≈ FT(π/N) for x in xspacings(grid, Face()))
        @test all(y ≈ FT(π/N) for y in yspacings(grid, Face()))
        @test all(z ≈ FT(π/N) for z in zspacings(grid, Face()))
```

src/Grids/grid_utils.jl
glwagner: I guess ideally we want to move towards an API where such a pattern doesn't emerge. Maybe if the API encourages using `xnodes(field)` rather than `xnodes(grid, location(field)...)`.

glwagner: Because restricting unused arguments to certain types can catch some silly errors like using `1` instead of a location, but won't catch all errors. So it's not a general solution in addition to the other problems noted here (making the code harder to develop, etc).

navidcy:
There are so many combinations e.g. in lat-lon grid (say when lat is regular and z is but lon is not..., etc) that I'm tempted to remove all convenience functions and only leave the ones that you give all 3 locations....

glwagner:
Personally I think it makes sense to start simple from an API perspective, and add convenience when its warranted

navidcy:
OK, I vote we drop all these conveniences and just keep:

```julia
xspacings(grid, LX, LY, LZ; with_halos=false)
yspacings(grid, LX, LY, LZ; with_halos=false)
...
```
and

```julia
xspacing(i, j, k, grid, LX, LY, LZ; with_halos=false)
yspacing(i, j, k, grid, LX, LY, LZ; with_halos=false)
...
```

What do you think?

glwagner:
> OK, I vote we drop all these conveniences and just keep:
> 
> ```julia
> xspacings(grid, LX, LY, LZ; with_halos=false)
> yspacings(grid, LX, LY, LZ; with_halos=false)
> ...
> ```
> 
> and
> 
> ```julia
> xspacing(i, j, k, grid, LX, LY, LZ; with_halos=false)
> yspacing(i, j, k, grid, LX, LY, LZ; with_halos=false)
> ...
> ```
> 
> What do you think?

I don't think the kwargs mean anything for these extractors so we should eliminate

tomchor:
> > OK, I vote we drop all these conveniences and just keep:
> > ```julia
> > xspacings(grid, LX, LY, LZ; with_halos=false)
> > yspacings(grid, LX, LY, LZ; with_halos=false)
> > ...
> > ```
> >       
> >     
> > 
> >     
> >   
> > and
> > ```julia
> > xspacing(i, j, k, grid, LX, LY, LZ; with_halos=false)
> > yspacing(i, j, k, grid, LX, LY, LZ; with_halos=false)
> > ...
> > ```
> >   
> > What do you think?
> 
> I don't think the kwargs mean anything for these extractors so we should eliminate

`xspacing()`, `yspacing` and `zspacing` already don't have kwargs. I think that was a typo by @navidcy. He's asking about removing specialized methods such as `xspacing(i, grid, LX)`.

@navidcy you can go ahead with that. If you want I can do that today also. Jut let me know since I don't wanna create conflicts

navidcy:
> I don't think the kwargs mean anything for these extractors so we should eliminate

true, my mistake when I wrote the GitHub comment
it's not in the code ;)

navidcy:
@tomchor shall I merge if tests pass?

docs/src/model_setup/output_writers.md
```diff
@@ -107,16 +107,24 @@ provided that their `dimensions` are provided:
...
 model = NonhydrostaticModel(grid=grid)
 
 simulation = Simulation(model, Δt=1.25, stop_iteration=3)
 
 f(model) = model.clock.time^2; # scalar output
-g(model) = model.clock.time .* exp.(znodes(Center, grid)); # vector/profile output
-h(model) = model.clock.time .* (   sin.(xnodes(Center, grid, reshape=true)[:, :, 1])
-                            .*     cos.(ynodes(Face, grid, reshape=true)[:, :, 1])) # xy slice output
+
+g(model) = model.clock.time .* exp.(znodes(Center, grid)) # vector/profile output
```
navidcy, line 120:
```suggestion
g(model) = model.clock.time .* exp.(znodes(Center, grid)) # single-column profile output (vector)
```

navidcy:
How do I get the spacings using `Δx`?

navidcy:
That's a very good point then! I'm not sure.

tomchor:
It was my impression that `Δx` worked differently. I even asked a similar question [here](https://github.com/CliMA/Oceananigans.jl/issues/2058#issuecomment-1368277630) about it. But if it returns the same thing, maybe indeed `xspacing` is redundant. I believe the functions `xspacings()` and such (which return an array) are novel and necessary though, right @simone-silvestri ?

navidcy:
So we may just wanna clean it up a bit -- will do that ;)

glwagner:
That's true that `xspacing` etc are redundant. We could also change `xspacings` to `Δx(grid, loc...)`...

simone-silvestri:
ah ok, I see. `Operators.Δx` is different from `AbstractOperations.Δx`. The first is just a generalization of the spacing operators that accepts locations as inputs and returns the correct `Δxᵃᵃᵃ`, the latter is a `GridMetricOperation` to be used in combination with other fields in an `AbstractOperation`.

Maybe we can remove the local `xspacing` from grids and rename `Operators.Δx` to `Operators.xspacing`, to remove both code duplication and confusion with `AbstractOperation`.

@tomchor, sure the `xspacings` function are needed to retrieve the whole spacings

simone-silvestri:
Ok, make sense! But I would suggest renaming the `Operators` ones instead or reimplementing a new function

glwagner:
Or we can rename the grid metric operation to `dx`?

simone-silvestri:
We could also do that (I would prefer it since in general I kind of prefer the `Δ` naming over `spacing` but that is just a personal preference)

tomchor:
How about we keep `AbstractOperations.Δx` as a grid metric operation with the same name and rename `Operators.Δx` to `xspacing()`? That way there's no confusion over what `Δx` means (since there will only be the grid metric with that name) and the names for the `Operators` have a more uniform nomenclature:

```julia
xnode(i, j, k, grid, ...)
xnodes(grid, ...)
xspacing(i, j, k, grid...)
xspacings(grid...)
```

navidcy:
I like @glwagner suggestion for exporting aliases, `const xspacing = Δx`, etc. Then we would have the homogeneous:
```julia
xnode(i, j, k, grid, ...)
xnodes(grid, ...)
xspacing(i, j, k, grid...)
xspacings(grid...)
```
as @tomchor also votes for.

For the grid metric, potentially `dx` is nice name. But we can discuss further on that outside of this PR.

So as is now, the Operators module comes after Grids so the way I can see this working is that `Operators` export these aliases. `Grids` could export specialized `λspacing` and `φspacing` methods for particular grids?

navidcy:
code could use `@inbounds` here, right?

https://github.com/CliMA/Oceananigans.jl/blob/fe5e3941ba1c2fcf5fa4d46264dc53d2dbf6851c/src/Operators/spacings_and_areas_and_volumes.jl#L138-L148

https://github.com/CliMA/Oceananigans.jl/blob/fe5e3941ba1c2fcf5fa4d46264dc53d2dbf6851c/src/Operators/spacings_and_areas_and_volumes.jl#L202-L209

navidcy:
Can/should we add the equivalent of `λspacing` and `φspacing` in `Oceananigans.Operators`, e.g., `Δλ`, `Δφ`?

navidcy:
Would it make sense if we define `x/y/zspacing` inside `Grids` as a fallback placeholder and then add a method inside `Operators`? This way we can import both `x/y/znode` and `x/y/zspacing` with, e.g.,

```Julia
using Oceananigans.Grids: xnode, spacing
```

glwagner:
@navidcy also comments that the `Operators` module should probably actually be inside `Grids`... which I agree with

navidcy:
omg, tests seems to be going to pass...

navidcy:
> @navidcy also comments that the `Operators` module should probably actually be inside `Grids`... which I agree with

Yeah... But in a different PR :)

navidcy:
> Would it make sense if we define `x/y/zspacing` inside `Grids` as a fallback placeholder and then add a method inside `Operators`? This way we can import both `x/y/znode` and `x/y/zspacing` with, e.g.,
> 
> ```julia
> using Oceananigans.Grids: xnode, spacing
> ```

In a different PR; perhaps one that brings Operators as part of or as a submodule of Grids.

glwagner (approved):

PR #2843, opened 2022-11-30
Bugfix in heuristic workgroup

glwagner:
Could be (left, right) or (right, left)... hm!


simone-silvestri:
does it change the performance?

glwagner:
I think it will for slices (I'd have to go digging but I recall mysterious benchmarks showing 10x slowdown for yz models). Probably worth doing the benchmarks before merging.

simone-silvestri (approved):

navidcy:
Merge this? Bench it?

glwagner:
I'll merge this now since benchmarking / major changes are planned anyways...

PR #2847, opened 2022-12-07
Fix bug loading `netcdf`

navidcy:


src/Oceananigans.jl
```diff
@@ -143,7 +143,7 @@ import Base:
 # (libnetcdf) are installed on the machine
 
 "Boolean denoting whether NCDatasets.jl can be loaded on machine."
-const hasnetcdf = @static (Sys.islinux() && Sys.ARCH == :x86_64) ? true : false
+const hasnetcdf = @static (Sys.islinux() && Sys.ARCH == :x86_64) ? false : true
```
glwagner, line 146: would

```julia
const hasnetcdf = try
    using NCDatasets
    true
catch
    false
end
```

work?

glwagner: I think we need a comment why linux + x86_64 means we don't have ncdatasets? Does it depend on the status of NetCDF_jll? There might be a relevant issue there to link to

navidcy:
@aramirezreyes can you confirm all is good with this PR?

src/Oceananigans.jl
navidcy: Actually, at https://github.com/JuliaBinaryWrappers/NetCDF_jll.jl they claim that they support linux + x86_64. But `libc=glibc`? What does a that mean? Probably it requires a specific C compiler?

navidcy:
@glwagner should we try on Satori (where the issue with netcdf was if I remember correctly)?

navidcy:
@aramirezreyes, I reverted back from try-catch. Can you try again to confirm that it's OK for you?

src/Oceananigans.jl
navidcy: Shall we add a comment pointing to https://github.com/Alexander-Barth/NCDatasets.jl/issues/150 ?

aramirezreyes:
It seems to continue working in this version.

src/Oceananigans.jl
```diff
@@ -143,7 +143,7 @@ import Base:
 # (libnetcdf) are installed on the machine
 
 "Boolean denoting whether NCDatasets.jl can be loaded on machine."
-const hasnetcdf = @static (Sys.islinux() && Sys.ARCH == :x86_64) ? true : false
+const hasnetcdf = @static (Sys.islinux() && Sys.ARCH == :powerpc64le) ? false : true
```
glwagner, line 146: I still don't grasp the problem with try-catch. Does it not work? It's better.

navidcy: as I said,
> Actually, try-catch won't cut the mustard for satori (=won't do the job)

I don't know why though :)

glwagner: Note that people who manage to install NetCDF on powerpc will _not_ be able to use NetCDF with Oceananigans despite all their hardwork! They'll have to change this source code. That shows you this is actually wrong, in principle. The problem is not at all that NetCDF doesn't work on powerpc. The problem is that binaries are not _provided automagically by Julia_ because NetCDF_jll doesn't work for powerpc.

navidcy: It doesn't work. I added it and then called `using Oceananigans` and I was getting errors related with `libnetcdf` not existing, etc.

navidcy: OK, I agree!
I just can't figure out a way to implement this so that it spits no errors on a machine that does not have the NetCDF_jll.

`try-catch` is a great suggestion but unfortunately doesn't work :(

glwagner: why doesn't it work? What's the error?

navidcy: perhaps we need to include the `try-catch` inside the `__init__()`?

glwagner: Maybe a test with NetCDF_jll directly rather than NCDatasets?

navidcy: OK, let me do it again

navidcy:
```Julia
navidcon@satori-login-001:~/Oceananigans.jl$ julia --project
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.2 (2022-09-29)
 _/ |\__'_|_|_|\__'_|  |
|__/                   |

julia> using Oceananigans
ERROR: InitError: UndefVarError: libnetcdf not defined
Stacktrace:
  [1] nc_inq_libvers
    @ ~/.julia/packages/NCDatasets/ipGBH/src/netcdf_c.jl:242 [inlined]
  [2] netcdf_version()
    @ NCDatasets ~/.julia/packages/NCDatasets/ipGBH/src/netcdf_c.jl:2157
  [3] init_certificate_authority()
    @ NCDatasets ~/.julia/packages/NCDatasets/ipGBH/src/netcdf_c.jl:2170
  [4] __init__()
    @ NCDatasets ~/.julia/packages/NCDatasets/ipGBH/src/NCDatasets.jl:33
  [5] _include_from_serialized(pkg::Base.PkgId, path::String, depmods::Vector{Any})
    @ Base ./loading.jl:831
  [6] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String, build_id::UInt64)
    @ Base ./loading.jl:1039
  [7] _tryrequire_from_serialized(modkey::Base.PkgId, build_id::UInt64)
    @ Base ./loading.jl:898
  [8] _tryrequire_from_serialized(pkg::Base.PkgId, path::String)
    @ Base ./loading.jl:971
  [9] _require(pkg::Base.PkgId)
    @ Base ./loading.jl:1347
 [10] _require_prelocked(uuidkey::Base.PkgId)
    @ Base ./loading.jl:1200
 [11] macro expansion
    @ ./loading.jl:1180 [inlined]
 [12] macro expansion
    @ ./lock.jl:223 [inlined]
 [13] require(into::Module, mod::Symbol)
    @ Base ./loading.jl:1144
during initialization of module NCDatasets

Julia>
```

navidcy:
OK, after https://github.com/Alexander-Barth/NCDatasets.jl/pull/193 we just have to wait for NCDatatsets v0.12.10 to be released and then we can merge this. And we'll be then free from conditional `using NCDatasets`.

navidcy:
works on satori; works on my M1 MacBook Pro... I'll merge when tests pass

navidcy:
@glwagner I think you need to approve?

aramirezreyes:
Continues to work for me on M2. Thanks!

tomchor (approved):

PR #2859, opened 2022-12-22
Increased `interpolate` speed

jagoosw:
Hi all,

I noticed that some of the array indexing in `interpolate` weren't `@inbounds`'ed so added them, and then also noticed that `issorted` tooke a good fraction of the time when interpolating fields on irregular grids so I removed that since the funciton is only used internally and is always going to be given sorted vectors.

This results in a ~5x speedup so seems worth it.

### DIFF
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -19,25 +19,21 @@ using a binary search. The input array `vec` has to be monotonically increasing.
 Code credit: https://computersciencehub.io/julia/code-for-binary-search-algorithm-julia
 """
 @inline function index_binary_search(vec, val, array_size)
-    if issorted(vec) 
-        low = 0
-        high = array_size - 1
+    low = 0
+    high = array_size - 1
 
-        while low + 1 < high 
-            mid = middle_point(low, high)
-            if @inbounds vec[mid + 1] == val 
-                return (mid + 1, mid + 1)
-            elseif @inbounds vec[mid + 1] < val
-                low = mid
-            else
-                high = mid
-            end
+    while low + 1 < high 
+        mid = middle_point(low, high)
+        if @inbounds vec[mid + 1] == val 
+            return (mid + 1, mid + 1)
+        elseif @inbounds vec[mid + 1] < val
+            low = mid
+        else
+            high = mid
         end
-
-        return (low + 1, high + 1)
-    else
-        throw(error("Vector not sorted, unable to search value"))
     end
+
+    return (low + 1, high + 1)
 end
 
 @inline function fractional_index(array_size::Int, val::FT, vec) where {FT}
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -16,7 +16,7 @@ vec[low] <= val && vec[high] >= val
 
 using a binary search. The input array `vec` has to be monotonically increasing.
 
-Code credit: https://computersciencehub.io/julia/code-for-binary-search-algorithm-julia
+Code credit: https://gist.github.com/cuongld2/8e4fed9ba44ea2b4598f90e7d5b6c612/155f9cb595314c8db3a266c3316889443b068017
 """
 @inline function index_binary_search(vec, val, array_size)
     low = 0
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -25,9 +25,9 @@ Code credit: https://computersciencehub.io/julia/code-for-binary-search-algorith
 
         while low + 1 < high 
             mid = middle_point(low, high)
-            if vec[mid + 1] == val 
+            if @inbounds vec[mid + 1] == val 
                 return (mid + 1, mid + 1)
-            elseif vec[mid + 1] < val
+            elseif @inbounds vec[mid + 1] < val
                 low = mid
             else
                 high = mid


navidcy:
Sounds great! Can you post here the code with which you benchmarked it?

jagoosw:
Yep! 
``` julia
using Oceanananigans, BenchmarkTools

using Oceananigans.Fields: interpolate

# Regularly spaced grids are interpolated differently
grid = RectilinearGrid(size=(100, 100, 100), x = [i^1.1 for i in 1:101], y = [i^1.2 for i in 1:101], z = [i^2 for i in 1:101])

field = CenterField(grid)

@benchmark interpolate(field, rand(), rand(), -rand())
```
```bash
BenchmarkTools.Trial: 10000 samples with 181 evaluations.
 Range (min … max):  584.481 ns …  26.258 μs  ┊ GC (min … max): 0.00% … 97.60%
 Time  (median):     612.343 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   634.835 ns ± 300.484 ns  ┊ GC (mean ± σ):  0.40% ±  0.98%

  ▆██▄▂▁                                                        ▂
  ██████▇▇▆▆▅▅▅▅▆▅▄▄▅▃▁▄▃▁▁▃▁▃▁▃▁▃▁▄▃▁▁▃▃▁▁▃▃▁▁▃▁▁▁▁▁▁▁▃▄▄▁▁▃▃█ █
  584 ns        Histogram: log(frequency) by time       1.49 μs <

 Memory estimate: 64 bytes, allocs estimate: 4.
```

Adding the inbounding improves this to:
```bash
BenchmarkTools.Trial: 10000 samples with 182 evaluations.
 Range (min … max):  577.841 ns …  28.209 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     605.769 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   625.685 ns ± 444.392 ns  ┊ GC (mean ± σ):  0.42% ± 0.98%

        █▃                                                       
  ▄▄▄▄▄▅██▆▅▄▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▂
  578 ns           Histogram: frequency by time          839 ns <

 Memory estimate: 64 bytes, allocs estimate: 4.
```

And removing the `issorted` check improves a lot more to:
```bash
BenchmarkTools.Trial: 10000 samples with 911 evaluations.
 Range (min … max):  118.231 ns …  11.848 μs  ┊ GC (min … max): 0.00% … 98.73%
 Time  (median):     125.412 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   135.838 ns ± 198.933 ns  ┊ GC (mean ± σ):  2.87% ±  2.40%

   █▅                                                            
  ▆██▆▄▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▁▂▂▂▁▁▂▂▂▁▁▂▂▁▂▁▂▂▁▂▂▁▂▁▂▁▂▁▁▂▂ ▂
  118 ns           Histogram: frequency by time          324 ns <

 Memory estimate: 64 bytes, allocs estimate: 4.
```

jagoosw:
For reference, with the same sized grid but regularly spaced the same benchmark gives this:
```bash
BenchmarkTools.Trial: 10000 samples with 916 evaluations.
 Range (min … max):  116.721 ns …   6.288 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     123.181 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   131.844 ns ± 171.767 ns  ┊ GC (mean ± σ):  2.51% ± 2.39%

        █                                                        
  ▄▃▃▃▃▃█▅▄▅▇▇▄▃▃▃▃▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▁▁▂▂▂▁▂▂▂▂ ▂
  117 ns           Histogram: frequency by time          165 ns <

 Memory estimate: 64 bytes, allocs estimate: 4.
```

jagoosw:
Also the reference link is dead and not archived anywhere, but I'm pretty sure its [this code](https://gist.github.com/cuongld2/8e4fed9ba44ea2b4598f90e7d5b6c612/155f9cb595314c8db3a266c3316889443b068017), from [this article](https://ledinhcuong99.medium.com/binary-search-in-julia-language-cdcef9d4364)

navidcy (approved):
It's actually more than what it claims!
Not only it's faster, but now it also works on GPUs! (It wasn't working before).

navidcy:
@jagoosw let's update the link? @glwagner was it you that wrote this? Is the links that @jagoosw points to above correct?

navidcy:
@glwagner was there any thinking behind having `issorted` there?

navidcy:
> It's actually more than what it claims!
> Not only it's faster, but now it also works on GPUs! (It wasn't working before).

I must have been hallucinating or did something wrong because still `interpolate` requires scalar operations... :)

navidcy:
@jagoosw I think we should update the link and merge

glwagner:
I didn't write it.

https://github.com/CliMA/Oceananigans.jl/blob/ae3a6dfa59d7308f7e36024e80f1491534caa113/src/Fields/interpolate.jl#L116

this is a high-level function; it isn't for use on the GPU. I don't know it's purpose.

jagoosw:
> I didn't write it.
> 
> https://github.com/CliMA/Oceananigans.jl/blob/ae3a6dfa59d7308f7e36024e80f1491534caa113/src/Fields/interpolate.jl#L116
> 
> this is a high-level function; it isn't for use on the GPU. I don't know it's purpose.

I think the only place it is used is in Lagrangian Particle tracking, for all particles it is used to get the velocity:
```
@inline function update_particle_position!(particles, p, restitution, grid::AbstractGrid{FT, TX, TY, TZ}, Δt, velocities) where {FT, TX, TY, TZ}
    # Advect particles using forward Euler.
    @inbounds u = interpolate(velocities.u, Face(), Center(), Center(), grid, particles.x[p], particles.y[p], particles.z[p]) 
    @inbounds v = interpolate(velocities.v, Center(), Face(), Center(), grid, particles.x[p], particles.y[p], particles.z[p]) 
    @inbounds w = interpolate(velocities.w, Center(), Center(), Face(), grid, particles.x[p], particles.y[p], particles.z[p]) 
```
And is also used to interpolate tracked fields. So it would be useful if this worked on GPUs.

jagoosw:
Should I merge?

navidcy:
Go for it

glwagner:
This function:

https://github.com/CliMA/Oceananigans.jl/blob/ae3a6dfa59d7308f7e36024e80f1491534caa113/src/Fields/interpolate.jl#L116

isn't used for Lagrangian particle tracking.

That's this function:

https://github.com/CliMA/Oceananigans.jl/blob/ae3a6dfa59d7308f7e36024e80f1491534caa113/src/Fields/interpolate.jl#L143

which is supposed to be GPU friendly.

jagoosw:
I checked and `interpolate(field, x, y, z)` isn't used anywhere in the source code but in a few validation experiment. It basically does the same thing, but extracts the fields location and grid. I've tried changing it to just be a wrapper for `interpolate(field, LX, LY, LZ, grid, x, y, z)` like:
```julia
@inline interpolate(field::AbstractField{LX, LY, LZ, G, T, N}, x, y, z) where {LX, LY, LZ, G, T, N} = interpolate(field, LX(), LY(), LZ(), G, x, y, z)
```
but this fails as a dynamic funciton invocation. I also tried changing it to:
```
@inline function interpolate(field, x, y, z)
    LX, LY, LZ = location(field)
    grid = field.grid
    return interpolate(field, LX(), LY(), LZ(), grid, x, y, z)
end
```

but this errors with `Reason: unsupported call to an unknown function (call to jl_f_getfield)`, so I'm not sure its going to be straight forward or worthwhile trying to make the high level version work on GPU.

Also, if we want to test interpolation, it always fails on GPU because of scalar indexing if called directly, but if wrapped in a kernel function is fine:
```julia
@kernel function test!(field, grid, res, x, y, z)
    n = @index(Global)
    LX, LY, LZ = location(field)
    interpolate(field, Center(), Center(), Center(), grid, x[n], y[n], z[n])
end
```
(If I put `grid = field.grid ` in the kernel function it also fails like above).

glwagner:
> but this errors with Reason: unsupported call to an unknown function (call to jl_f_getfield), so I'm not sure its going to be straight forward or worthwhile trying to make the high level version work on GPU.

I think that's because on the GPU, `field` is an offset array and has no property `grid`.

glwagner:
The function signature 

```julia
@inline function interpolate(field, LX, LY, LZ, grid, x, y, z)
```

is confusing: `field` can be either an array (eg `data(field::Field)`) or `field::Field` itself. Perhaps this should be documented or even the name changed `field` -> `data` or `field_data`.

navidcy:
Interesting. I see that error a lot in #2782

navidcy:
If it’s an array how will field.grid work?

glwagner:
It wont. We can't write `field.grid` on the GPU.

navidcy:
I’m confused though. Is the field struct different when it lives on GPU? No, right?

glwagner:
Why do we want `interpolate(field, x, y, z)` to work on the GPU?

navidcy:
We might not. Where is interpolate used?

glwagner:
Yes, most containers are "adapted" on the GPU. We adapt `field` to `field.data`:

https://github.com/CliMA/Oceananigans.jl/blob/7debded91d24520d72857c120303bcd22b805104/src/Fields/field.jl#L394

And note that `field.data` will be further adapted. For example, `CuArray` becomes `CuDeviceArray`.

navidcy:
> Yes, most containers are "adapted" on the GPU. We adapt `field` to `field.data`:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/7debded91d24520d72857c120303bcd22b805104/src/Fields/field.jl#L394
> 
> And note that `field.data` will be further adapted. For example, `CuArray` becomes `CuDeviceArray`.

GOTHCA!

navidcy:
OK, may I suggest the following:

If `interpolate(field, x, y, z)` would only work for CPU fields then let's type it, e.g. `interpolate(field::Field{CPU}, x, y, z)` and also make a note in the docstring!

glwagner:
Another option is to delete it.

I don't know about the type restriction. `GPU` is fine with `@allowscalar`.

jagoosw:
> We might not. Where is interpolate used?

Its not used anywhere in the source code but in quite a few validation scripts

jagoosw:
> > but this errors with Reason: unsupported call to an unknown function (call to jl_f_getfield), so I'm not sure its going to be straight forward or worthwhile trying to make the high level version work on GPU.
> 
> I think that's because within a kernal on the GPU, `field` in
> 
> ```julia
> @inline function interpolate(field, x, y, z)
>     LX, LY, LZ = location(field)
>     grid = field.grid
>     return interpolate(field, LX(), LY(), LZ(), grid, x, y, z)
> end
> ```
> 
> is an OffsetArray and has no property `grid` (or a location).

Ah this makes sense, thank you

glwagner:
> > We might not. Where is interpolate used?
> 
> Its not used anywhere in the source code but in quite a few validation scripts

Can you point us to a few?

jagoosw:
I think these are the only two instances

https://github.com/CliMA/Oceananigans.jl/blob/7debded91d24520d72857c120303bcd22b805104/validation/immersed_boundaries/cylinder_flow_with_tracer.jl#L243-L250

https://github.com/CliMA/Oceananigans.jl/blob/7debded91d24520d72857c120303bcd22b805104/validation/near_global_lat_lon/cyclic_interpolate_utils.jl#L18-L30

Issue #2857, opened 2022-12-13
Error when writing `Field`s reduced over 3 dimensions to NetCDF


tomchor:
When I try to write a `Field` that's reduced over 3 dimensions to a NetCDF file I get an error. For the example the MWE below

```julia
using Oceananigans

grid = RectilinearGrid(size=(8,8,8,), extent=(1,1,1))
model = NonhydrostaticModel(grid=grid)


u, v, w = model.velocities
u_mean = Field(Average(u))
outputs = (; u_mean,)

filename = "mwe.nc"
writer_nc = NetCDFOutputWriter(model, outputs;
                               filename = "avgnc.nc",
                               schedule = IterationInterval(1),
                               overwrite_existing = true,)

simulation = Simulation(model, Δt=1, stop_iteration=5)
simulation.output_writers[:fields] = writer_nc

run!(simulation)
```

produces this error:

```
ERROR: LoadError: MethodError: no method matching Float32(::Array{Float32, 0})
Closest candidates are:
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number} at char.jl:50
  (::Type{T})(::Base.TwicePrecision) where T<:Number at twiceprecision.jl:266
  (::Type{T})(::Complex) where T<:Real at complex.jl:44
  ...
Stacktrace:
  [1] setindex!(v::NCDatasets.Variable{Float32, 1, NCDatasets.NCDataset{Nothing}}, data::Array{Float32, 0}, indexes::Int64)
    @ NCDatasets ~/.julia/packages/NCDatasets/dXXHC/src/variable.jl:308
  [2] setindex!(v::NCDatasets.CFVariable{Float32, 1, NCDatasets.Variable{Float32, 1, NCDatasets.NCDataset{Nothing}}, NCDatasets.Attributes{NCDatasets.NCDataset{Nothing}}, NamedTuple{(:fillvalue, :missing_values, :scale_factor, :add_offset, :calendar, :time_origin, :time_factor), Tuple{Nothing, Tuple{}, Nothing, Nothing, Nothing, Nothing, Nothing}}}, data::Array{Float32, 0}, indexes::Int64)
    @ NCDatasets ~/.julia/packages/NCDatasets/dXXHC/src/cfvariable.jl:765
[...]

PR #2865, opened 2023-01-14
Fixes writing of 3d-reduced `Field`s to NetCDF

tomchor:
This fixes the error we were getting when writing `Field`s reduced over 3 dimensions to disk with `NetCDFOutputWriter` according to the upstream provided in a PR at NCDatasets: https://github.com/Alexander-Barth/NCDatasets.jl/issues/197.

This PR also adds a test to catch this in the future.

For now this is only working on the `master` branch of  `NCDatasets` so tests should fail for now, but once a new version of NCDatasets is released I'll update the packages.

Closes https://github.com/CliMA/Oceananigans.jl/issues/2857


glwagner:
Can you explicitly link the PR / issue in NCDatasets?

glwagner:
Here's an issue:

https://github.com/Alexander-Barth/NCDatasets.jl/issues/197

tomchor:
We're getting errors when running the tests on buildkite that I'm not getting when running on a GPU locally.

For example this:

```julia
Field boundary conditions [GPU, RectilinearGrid]: Test Failed at /net/ocean/home/data44/data5/glwagner/.buildkite-agent/builds/sverdrup-7/clima/oceananigans/test/test_computed_field.jl:475
--
  | Expression: #= /net/ocean/home/data44/data5/glwagner/.buildkite-agent/builds/sverdrup-7/clima/oceananigans/test/test_computed_field.jl:475 =# CUDA.@allowscalar all(ST.data[1:Nx, 1:Ny, 0] .== ST.data[1:Nx, 1:Ny, 1])
```

The above line works for me on GPU when doing `include("test_computed_field.jl")`, but for some reason fails on buildkite.

@glwagner @simone-silvestri any ideas as to why? The fact that I can't reproduce these errors locally is making it hard for me to solve them

test/test_computed_field.jl
```diff
@@ -212,9 +212,11 @@ function computation_including_boundaries(arch)
     model = NonhydrostaticModel(grid=grid)
 
     u, v, w = model.velocities
-    @. u.data = 1 + rand()
-    @. v.data = 2 + rand()
-    @. w.data = 3 + rand()
+    CUDA.@allowscalar begin
```
glwagner, line 215: Is there any way to do this without `@allowscalar`? We should be reducing where allowscalar appears in tests, not adding new tests with this.

test/test_computed_field.jl
```diff
@@ -301,7 +303,7 @@ function computations_with_averaged_field_derivative(model)
 
     set!(model, T = (x, y, z) -> 3 * z)
 
-    return all(interior(shear)[2:3, 2:3, 2:3] .== interior(T)[2:3, 2:3, 2:3])
+    return CUDA.@allowscalar all(interior(shear)[2:3, 2:3, 2:3] .== interior(T)[2:3, 2:3, 2:3])
```
glwagner, line 306: Same as above. We should minimize usage of `@allowscalar` in tests. This is one of the largest sources of technical debt in our tests and has incurred a lot of pain in the past

test/test_computed_field.jl
```diff
@@ -320,7 +322,7 @@ function computations_with_computed_fields(model)
     tke = Field(tke_op)
     compute!(tke)
 
-    return all(interior(tke)[2:3, 2:3, 2:3] .== 9/2)
+    return CUDA.@allowscalar all(interior(tke)[2:3, 2:3, 2:3] .== 9/2)
```
glwagner, line 325: Same as above

glwagner:
There are a lot of new instances of `@allowscalar`, but rather than adding new instances we should be refactoring the tests so they don't appear.

When we find that we have to use `@allowscalar`, it often indicates that our `Field` infrastructure is somehow deficient / doesn't support necessary operations, which causes us to resort to indexing and other syntax that requires `@allowscalar`.

@navidcy @simone-silvestri

tomchor:
> There are a lot of new instances of `@allowscalar`, but rather than adding new instances we should be refactoring the tests so they don't appear.

I added these because it was the only way to make tests pass locally. However, I can't fully reproduce tests results locally anyway, like I mentioned in my previous comment, so these may well be unnecessary (since these lines might be passing on buildkite).

test/test_computed_field.jl
glwagner, line 325:
```suggestion
    return all(interior(tke, 2:3, 2:3, 2:3) .== 9/2)
```

glwagner:
I suggested a syntax change that could help

tomchor:
> I suggested a syntax change that could help

You're right it does help. I'll try replacing these one by one and see if that helps with the error on buildkite. Although if would be useful to figure out why I'm not getting the same errors locally.

glwagner:
Awesome!

tomchor:
Apparently the new syntax does help avoid `@allowscalar` instances, and things do compile locally for me, but the errors on buildkite are still [there](https://buildkite.com/clima/oceananigans/builds/9727#01860b25-0d9a-419e-bd01-3ec11bcb8c6b/38-603):

```
Computations with Averaged Fields [GPU, RectilinearGrid]: Test Failed at /net/ocean/home/data44/data5/glwagner/.buildkite-agent/builds/sverdrup-13/clima/oceananigans/test/test_computed_field.jl:583
--
  | Expression: all(interior(tke_yz) .== 9 / 2)
```

Any ideas on what might be the cause of the differences between builkite and my local server? If someone could also run one of the failing tests on a GPU locally and see if they get the same errors that buildkite is throwing, that would be helpful.

tomchor:
Finally got the tests passing! It was something having to do with GPUCompiler.jl.

This is ready to merge/review.

tomchor:
@navidcy @simone-silvestri @glwagner with https://github.com/CliMA/Oceananigans.jl/pull/2899 being merged, this bugfix PR is pretty trivial. Can I get a review whenever any of you have the time?

simone-silvestri (approved):

test/test_netcdf_output_writer.jl
```diff
@@ -504,10 +504,53 @@ function test_netcdf_function_output(arch)
...
+    model = NonhydrostaticModel(grid = grid,
+                                timestepper = :RungeKutta3,
+                                tracers = (:c,), 
+                                coriolis = nothing, 
+                                buoyancy = nothing, 
+                                closure = nothing)
+    set!(model, c=1)
+
+    Δt = 1/64 # Nice floating-point number
+    simulation = Simulation(model, Δt=Δt, stop_time=50Δt)
```
glwagner, line 521: Can we do fewe stepsr? Our tests currently stretch the limits of our resources so we need to be as parsimonious as possible when adding new tests. Note also that compilation cost is the main thing. If this test can be combined with another test, that'd be ideal.

glwagner: Also suggest using `stop_iteration`

tomchor: Sorry @glwagner I ended up merging before you commented. I copied the test template for others in the same file so there are more tests that we apply this change to. Would you like me to open another PR for this?

glwagner: Right, I'm pointing out that we don't want to copy/paste test code now without care, since a lot of our test code is poorly written / wasteful and our CI is straining under the pressure... :-/

glwagner: Any PRs that reduce test cost will be greatly appreciated! I can't tell if all the changes will make a big difference, you are better placed to analyze that.

tomchor: After running these a few time for this PR, I'd say that this one won't make much of a difference. But I have identified several tests that instantiate its own model each that could be merged together. That, I think, will have a more significant impact. I'll open a PR about it once some of my other PRs are merged!

glwagner: sounds good

PR #2870, opened 2023-01-20
Don't push docs preview

navidcy:
Unfortunately the previews clog up the [OceananigansDocumentation](https://github.com/CliMA/OceananigansDocumentation) repository and we need to perform manual cleanup often...

I suggest we drop the doc previews for the moment. We could build the docs manually _or_ temporarily enable them for a particular PR that includes Doc changes and then disable them again.


simone-silvestri:
good idea!

simone-silvestri (approved):

PR #2867, opened 2023-01-18
Multiregions + CubedSphere

navidcy:

PR #2871, opened 2023-01-21
CubedSphere grid construction + tests

navidcy:
The cubed sphere grid from the MIT as read from DataDeps has the faces ordered:
  1. `+ x` direction
  2. `+ y` direction
  3. `+ z` direction (Arctic)
  4. `- x` direction
  5. `- y` direction
  6. `- z` direction (Antarctica)
 
So to avoid confusion for (users and developers!) I thought I make sure that the ConformalCubedSphere follows the same convention.

This PR does that + it adds tests to check whether the ConformalCubedSphereGrid metrics from the grid read from file and the grid constructed in Oceananigans.jl are the same.

After this is merged we'll move on to #2867 to constructing `ConformalCubedSphereGrid` using MultiRegion (an possibly nuke the `CubedSpheres` module completely). We will have a `ConformalCubedSphereGrid` (which is the same as the `cs32` grid read from file) to bench our MultiRegion implementation with!

(This PR also moves some of the spherical trigonometry utility function to `grid_utils.jl`.)

### DIFF
--- a/src/CubedSpheres/conformal_cubed_sphere_grid.jl
+++ b/src/CubedSpheres/conformal_cubed_sphere_grid.jl
@@ -128,22 +128,22 @@ function ConformalCubedSphereGrid(arch = CPU(), FT=Float64; face_size, z, face_h
     size, halo = face_size, face_halo
 
     # +x face (face 1)
-    x⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotX(π/2))
+    x⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotX(π/2)*RotY(π/2))
 
     # +y face (face 2)
-    y⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotY(π/2))
+    y⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotY(π)*RotX(-π/2))
 
     # +z face (face 3)
-    z⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=nothing)
+    z⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotZ(π))
 
     # -x face (face 4)
-    x⁻_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotX(-π/2))
+    x⁻_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotX(π)*RotY(-π/2))
 
     # -y face (face 5)
-    y⁻_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotY(-π/2))
+    y⁻_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotY(π/2)*RotX(π/2))
 
     # -z face (face 6)
-    z⁻_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotX(π))
+    z⁻_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotZ(π/2)*RotX(π))
 
     faces = (
         x⁺_face_grid,
--- a/src/CubedSpheres/conformal_cubed_sphere_grid.jl
+++ b/src/CubedSpheres/conformal_cubed_sphere_grid.jl
@@ -127,14 +127,14 @@ function ConformalCubedSphereGrid(arch = CPU(), FT=Float64; face_size, z, face_h
 
     size, halo = face_size, face_halo
 
-    # +z face (face 1)
-    z⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=nothing)
-
-    # +x face (face 2)
+    # +x face (face 1)
     x⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotX(π/2))
 
-    # +y face (face 3)
+    # +y face (face 2)
     y⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotY(π/2))
+
+    # +z face (face 3)
+    z⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=nothing)
 
     # -x face (face 4)
     x⁻_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotX(-π/2))
@@ -146,9 +146,9 @@ function ConformalCubedSphereGrid(arch = CPU(), FT=Float64; face_size, z, face_h
     z⁻_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=RotX(π))
 
     faces = (
-        z⁺_face_grid,
         x⁺_face_grid,
         y⁺_face_grid,
+        z⁺_face_grid,
         x⁻_face_grid,
         y⁻_face_grid,
         z⁻_face_grid
--- a/src/Grids/grid_utils.jl
+++ b/src/Grids/grid_utils.jl
@@ -447,6 +447,11 @@ are the triangle's inner angles.
 
 It has been known since Euler and Lagrange that ``\\tan(E/2) = P / (1 + \\cos a + \\cos b + \\cos c)``, where
 ``P = (1 - \\cos²a - \\cos²b - \\cos²c + 2 \\cos a \\cos b \\cos c)^{1/2}``.
+
+References
+==========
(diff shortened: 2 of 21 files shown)


navidcy:
this is ready for review :)

src/Grids/grid_utils.jl
```diff
@@ -431,3 +432,101 @@ coordinate_summary(Δ::Union{AbstractVector, AbstractMatrix}, name) =
...
+Return the area of a spherical triangle on the unit sphere with sides `a`, `b`, and `c`.
+
+The area of a spherical triangle on the unit sphere is ``E = A + B + C - π``, where ``A``, ``B``, and ``C``
+are the triangle's inner angles.
+
+It has been known since Euler and Lagrange that ``\\tan(E/2) = P / (1 + \\cos a + \\cos b + \\cos c)``, where
+``P = (1 - \\cos²a - \\cos²b - \\cos²c + 2 \\cos a \\cos b \\cos c)^{1/2}``.
+"""
+function spherical_area_triangle(a::Number, b::Number, c::Number)
+    cosa, cosb, cosc = cos.((a, b, c))
```
glwagner, line 452:
```suggestion
    cosa = cos(a)
    cosb = cos(b)
    cosc = cos(c)
```

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -127,7 +133,7 @@ function OrthogonalSphericalShellGrid(architecture::AbstractArchitecture = CPU()
                                       rotation = nothing)
 
     @warn "OrthogonalSphericalShellGrid is still under development. Use with caution!"
-    @warn "Azᶜᶜᵃ are correct for interior points, but Azᶠᶜᵃ, Azᶜᶠᵃ, Azᶠᶠᵃ are wrong close to boundaries"
+    @warn "Metrics close to boundaries might be wrong!"
```
glwagner, line 136: why?

src/CubedSpheres/conformal_cubed_sphere_grid.jl
```diff
@@ -127,28 +127,28 @@ function ConformalCubedSphereGrid(arch = CPU(), FT=Float64; face_size, z, face_h
 
     size, halo = face_size, face_halo
 
-    # +z face (face 1)
-    z⁺_face_grid = OrthogonalSphericalShellGrid(arch, FT; size, z, halo, radius, rotation=nothing)
+    # +x face (face 1)
```
glwagner, line 130: We need a sketch with lat-lon lines overlaying the cube so we understand the "+z", etc terminology!

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -193,90 +203,112 @@ function OrthogonalSphericalShellGrid(architecture::AbstractArchitecture = CPU()
...
 
-    ΔS = [Δxᶜᶜᵃ,  Δxᶠᶜᵃ,  Δxᶜᶠᵃ,  Δxᶠᶠᵃ,  Δyᶜᶜᵃ,  Δyᶜᶠᵃ,  Δyᶠᶜᵃ,  Δyᶠᶠᵃ]
+    ΔS = [Δxᶜᶜᵃ,  Δxᶠᶜᵃ,  Δxᶜᶠᵃ,  Δxᶠᶠᵃ,  Δyᶜᶜᵃ,  Δyᶜᶠᵃ,  Δyᶠᶜᵃ,  Δyᶠᶠᵃ ]
     ΔΣ = [Δσxᶜᶜᵃ, Δσxᶠᶜᵃ, Δσxᶜᶠᵃ, Δσxᶠᶠᵃ, Δσyᶜᶜᵃ, Δσyᶜᶠᵃ, Δσyᶠᶜᵃ, Δσyᶠᶠᵃ]
 
-    # convert degrees to meters
     for (Δs, Δσ) in zip(ΔS, ΔΣ)
         @. Δσ = deg2rad(Δσ)
         @. Δs = radius * Δσ
     end
```
glwagner, line 294: It's a bit hard to read like this, would it be better to simply write explicitly things like

```julia
@. Δxᶜᶜᵃ = radius * Δσxᶜᶜᵃ
```

etc?

glwagner: I think I prefer not using the loop

glwagner (approved):

src/CubedSpheres/conformal_cubed_sphere_grid.jl
navidcy: Yeap, I plan to add some figures in the docs. For now, is the figures at https://clima.github.io/CubedSphere.jl/stable/conformal_cubed_sphere/ helpful? +z is the face that looks in the positive z direction.

src/Grids/grid_utils.jl
```diff
@@ -431,3 +432,101 @@ coordinate_summary(Δ::Union{AbstractVector, AbstractMatrix}, name) =
...
+"""
+    spherical_area_triangle(a::Number, b::Number, c::Number)
+
+Return the area of a spherical triangle on the unit sphere with sides `a`, `b`, and `c`.
+
+The area of a spherical triangle on the unit sphere is ``E = A + B + C - π``, where ``A``, ``B``, and ``C``
+are the triangle's inner angles.
+
+It has been known since Euler and Lagrange that ``\\tan(E/2) = P / (1 + \\cos a + \\cos b + \\cos c)``, where
+``P = (1 - \\cos²a - \\cos²b - \\cos²c + 2 \\cos a \\cos b \\cos c)^{1/2}``.
```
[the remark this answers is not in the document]
navidcy: I added citations to the works by Euler and Lagrange. ;)

navidcy:
@glwagner asks why we have the warning saying that the metrics near the edge of the `OrthogonalSphericalShellGrid` might be wrong. I should have clarified there that I "The _area_ metrics might be wrong near the boundaries".

I explain here:

To compute the areas near the edge we actually need to know the node locations of the halos, and to do that we need to implement proper halo filling. Why is that? Take for For example, consider a face near the edge, then:

```julia
Δxᶠᶠᵃ[1, j, k] = xᶜᶠᵃ[1, j, k] - xᶜᶠᵃ[0, j, k]
```

but we can't compute the lat-lon position of `xᶜᶠᵃ[0, j, k]` from the conformal map because it belongs to a different panel. However, from symmetry argument, one can say that this would be just twice the distance from the first cell center to the edge of the panel:

```julia
Δxᶠᶠᵃ[1, j, k] = 2 * (xᶜᶠᵃ[1, j, k] - xᶠᶠᵃ[1, j, k])
```

So that's that. The distance metrics are actually correct, _even near the edges_.

How do we compute now areas? An approximation is 

```Julia
    @. Azᶜᶜᵃ = Δxᶜᶜᵃ * Δyᶜᶜᵃ
    @. Azᶠᶜᵃ = Δxᶠᶜᵃ * Δyᶠᶜᵃ
    @. Azᶜᶠᵃ = Δxᶜᶠᵃ * Δyᶜᶠᵃ
    @. Azᶠᶠᵃ = Δxᶠᶠᵃ * Δyᶠᶠᵃ
```

But that's only an approximation because these quadrilaterals are not symmetrical. Anywhere in the interior of the grid we can improve on this by computing the exact area of the spherical quadrilateral given that we know the location of each of its vertices. But for a quadrilateral that touches the edge some of its vertices would belong to the next panels, i.e., to get their location we'd need proper halo filling. E.g., `Azᶠᶜᵃ[i, j]` has vertices\

```julia
   (φᶜᶠᵃ[i-1,  j ], λᶠᶠᵃ[i-1,  j ])
   (φᶜᶠᵃ[ i ,  j ], λᶠᶠᵃ[ i ,  j ])
   (φᶜᶠᵃ[ i , j+1], λᶠᶠᵃ[ i , j+1])
   (φᶜᶠᵃ[i-1, j+1], λᶠᶠᵃ[i-1, j+1])
```

thus `Azᶠᶜᵃ[1, j]` would require the location of `(φᶜᶠᵃ[0, j+1], λᶠᶠᵃ[0, j+1])` and `(φᶜᶠᵃ[0,  j ], λᶠᶠᵃ[0,  j ])`.

But now that I'm thinking about it, again from symmetry we would expect that the area of quadrilateral with vertices at

```julia
   (φᶜᶠᵃ[ 0 ,  j ], λᶜᶠᵃ[ 0 ,  j ])
   (φᶜᶠᵃ[ 1 ,  j ], λᶜᶠᵃ[ 1 ,  j ])
   (φᶜᶠᵃ[ 1 , j+1], λᶜᶠᵃ[ 1 , j+1])
   (φᶜᶠᵃ[ 0 , j+1], λᶜᶠᵃ[ 0 , j+1])
```

should be double the area of the quadrilateral with vertices

```julia
   (φᶠᶠᵃ[ 1 ,  j ], λᶠᶠᵃ[ 1 ,  j ])
   (φᶜᶠᵃ[ 1 ,  j ], λᶜᶠᵃ[ 1 ,  j ])
   (φᶜᶠᵃ[ 1 , j+1], λᶜᶠᵃ[ 1 , j+1])
   (φᶠᶠᵃ[ 1 , j+1], λᶠᶠᵃ[ 1 , j+1])
```

I'll fix this and remove the warning!

src/Grids/orthogonal_spherical_shell_grid.jl
navidcy: done

src/Grids/grid_utils.jl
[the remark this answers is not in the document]
simone-silvestri: Ahah nice 18th century references.

PR #2874, opened 2023-01-24
Improve the performance of the HydrostaticFreeSurfaceModel

simone-silvestri:
In large computations, advection tendency and implicit free surface are the bottleneck.

This can be improved (probably) with an improved `SplitExplicitFreeSurface` that scales better at high resolutions and by introducing shared memory for the calculation of advection. (Another way to improve advection performance is to split the kernels based on the distance to a solid boundary such that we do not have to perform the calculations of all orders starting from 1, but it is probably a slightly more complex solution)

This is an explorative PR to try to address both problems.


src/Advection/positivity_preserving_tracer_advection_operators.jl
```diff
@@ -29,8 +29,8 @@ end
 
     cᵢⱼ  = c[i, j, k]
 
-    c₊ᴸ =  _left_biased_interpolate_xᶠᵃᵃ(i+1, j, k, grid, advection, c)
-    c₊ᴿ = _right_biased_interpolate_xᶠᵃᵃ(i+1, j, k, grid, advection, c)
+    c₊ᴸ =  _left_biased_interpolate_xᶠᵃᵃ(i+0x1, j, k, grid, advection, c)
+    c₊ᴿ = _right_biased_interpolate_xᶠᵃᵃ(i+0x1, j, k, grid, advection, c)
```
glwagner, line 33: Is there a macro we could use to convert all integers to `Uint8` or something?

navidcy:
Is this PR relevant or stale after #2888

simone-silvestri:
It is kind of stale. On the other hand, I wouldn't delete the branch, because it has some shared memory usage that we maybe want to try later in the future.

glwagner:
You can close the PR but keep the branch

PR #2879, opened 2023-01-25
Automatically-generated function signatures with DocStringExtensions

tomchor:
This PR implements automatically-generated function signatures with DocStringExtensions. I think this will avoid human errors when creating docstrings and also making documenting stuff easier, which hopefully will ultimately help users and novice developers.

Starting this as a draft since I wanna see what effects this will have on the docs.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.3"
+julia_version = "1.8.5"
 manifest_format = "2.0"
 project_hash = "47b13a64760654458c294e116e59e2bbbf21ea8c"
 
@@ -128,7 +128,7 @@ version = "4.3.0"
 [[deps.CompilerSupportLibraries_jll]]
 deps = ["Artifacts", "Libdl"]
 uuid = "e66e0078-7015-5450-92f7-15fbd957f2ae"
-version = "0.5.2+0"
+version = "1.0.1+0"
 
 [[deps.Crayons]]
 git-tree-sha1 = "249fe38abf76d48563e2f4556bebd215aa317e15"
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -168,6 +168,6 @@ deploydocs(
           repo = "github.com/CliMA/OceananigansDocumentation.git",
       versions = ["stable" => "v^", "v#.#.#", "dev" => "dev"],
      forcepush = true,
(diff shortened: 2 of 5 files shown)


tomchor:
Docs preview should be at https://clima.github.io/OceananigansDocumentation/preview/PR2879

glwagner:
Nice! We probably want to replace docstrings judiciously, since there are some we may want to reduce (for the purpose of readability / understability) relative to the automatically generated versions?

tomchor:
> Nice! We probably want to replace docstrings judiciously, since there are some we may want to reduce (for the purpose of readability / understability) relative to the automatically generated versions?

I think the way DocStringExtensions includes things is pretty minimally invasive so I don't think too-verbose docstring will be an issue (if I understand your point correctly). But I agree with your point.

Just as a point of comparison, here's what the start of the docstrings for `RectilinearGrid` looks like on main:

```julia
help?> RectilinearGrid
search: RectilinearGrid

  RectilinearGrid([architecture = CPU(), FT = Float64];
                  size,
                  x = nothing,
                  y = nothing,
                  z = nothing,
                  halo = nothing,
                  extent = nothing,

  Creates a RectilinearGrid with size = (Nx, Ny, Nz) grid points.
```

and here's what it looks like o this branch after using DocStringExtensions:

```julia
help?> RectilinearGrid
search: RectilinearGrid

  RectilinearGrid()
  RectilinearGrid(architecture)
  RectilinearGrid(
      architecture,
      FT;
      size,
      x,
      y,
      z,
      halo,
      extent,
      topology
  )
  

  Creates a RectilinearGrid with size = (Nx, Ny, Nz) grid points.
```

For now the automatic signature isn't showing the default for each argument, which isn't great, but there might be a way to include that with some options.

tomchor:
It looks like one would expect in the docs as well:

image

glwagner:
I prefer the version on `main`. How about you?

tomchor:
> I prefer the version on `main`. How about you?

As it is I also prefer main. But I'll try to tweak DocStringExtensions to try and improve the output.

tomchor:
It seems like DocStringExtensions doesn't yet have the ability to specify default arguments, but this feature might come in a [not-so-distant future](https://github.com/JuliaDocs/DocStringExtensions.jl/issues/19#issuecomment-1418284136).

I'll close this PR for now and open it again when this feature is implemented upstream.

PR #2883, opened 2023-01-29
Defines some binary operations with `ZeroField`

tomchor:
This PR defines some binary operations involving `ZeroField`s that might be useful, and adds tests for those operations. The issue came up at https://github.com/tomchor/Oceanostics.jl/pull/104

### DIFF
--- a/src/AbstractOperations/binary_operations.jl
+++ b/src/AbstractOperations/binary_operations.jl
@@ -125,6 +125,12 @@ function define_binary_operator(op)
 
         $op(a::AF, b::Number) = $op(location(a), a, b)
         $op(a::Number, b::AF) = $op(location(b), a, b)
+
+        $op(a::AF, b::ConstantField) = $op(location(a), a, b.constant)
+        $op(a::ConstantField, b::AF) = $op(location(b), a.constant, b)
+
+        $op(a::Number, b::ConstantField) = ConstantField($op(a, b.constant))
+        $op(a::ConstantField, b::Number) = ConstantField($op(a.constant, b))
     end
 end
 
@@ -195,10 +201,10 @@ import Base: +, -, *, /, ==
 using Oceananigans.Fields: ZeroField, ConstantField
 
 ==(::ZeroField, ::ZeroField) = true
-==(c1::ConstantField, c2::ConstantField) = c1.constant == c2.constant
 
 ==(zf::ZeroField, cf::ConstantField) = 0 == cf.constant
 ==(cf::ConstantField, zf::ZeroField) = ==(zf, cf)
+==(c1::ConstantField, c2::ConstantField) = c1.constant == c2.constant
 
 +(a::ZeroField, b::AbstractField) = b
 +(a::AbstractField, b::ZeroField) = +(b, a)
--- a/src/AbstractOperations/binary_operations.jl
+++ b/src/AbstractOperations/binary_operations.jl
@@ -188,6 +188,38 @@ macro binary(ops...)
 end
 
 #####
+##### BinaryOperations with ZeroField
+#####
+
+import Base: +, -, *, /, ==
+using Oceananigans.Fields: ZeroField
+
+==(::ZeroField, ::ZeroField) = true
+
++(a::ZeroField, b::AbstractField) = b
++(a::AbstractField, b::ZeroField) = +(b, a)
++(a::ZeroField, b::Number) = b
++(a::Number, b::ZeroField) = +(b, a)
+
+-(a::ZeroField, b::AbstractField) = -b
+-(b::AbstractField, a::ZeroField) = b
+-(a::ZeroField, b::Number) = -b
+-(b::Number, a::ZeroField) = b
+
+*(a::ZeroField, b::AbstractField) = a
+*(b::AbstractField, a::ZeroField) = *(a, b)
+*(a::ZeroField, b::Number) = a
+*(b::Number, a::ZeroField) = *(a, b)
+
+/(a::ZeroField, b::AbstractField) = a
+/(b::AbstractField, a::ZeroField) = Inf
+/(a::ZeroField, b::Number) = a
+/(b::Number, a::ZeroField) = Inf
+
+^(a::ZeroField, b::Number) = a
+
+
+#####
 ##### Nested computations
 #####
 
--- a/src/AbstractOperations/binary_operations.jl
+++ b/src/AbstractOperations/binary_operations.jl
@@ -225,7 +225,7 @@ using Oceananigans.Fields: ZeroField, ConstantField
 /(a::ZeroField, b::AbstractField) = a
 /(a::AbstractField, b::ZeroField) = ConstantField(convert(eltype(a), Inf))
 /(a::ZeroField, b::Number) = a
-/(a::Number, b::ZeroField) = ConstantField(a/convert(eltype(a), 0))
+/(a::Number, b::ZeroField) = ConstantField(a / convert(eltype(a), 0))
 
 
 #####
--- a/src/AbstractOperations/binary_operations.jl
+++ b/src/AbstractOperations/binary_operations.jl
@@ -207,24 +207,24 @@ using Oceananigans.Fields: ZeroField, ConstantField
 ==(c1::ConstantField, c2::ConstantField) = c1.constant == c2.constant
 
 +(a::ZeroField, b::AbstractField) = b
-+(a::AbstractField, b::ZeroField) = +(b, a)
++(a::AbstractField, b::ZeroField) = a
 +(a::ZeroField, b::Number) = ConstantField(b)
-+(a::Number, b::ZeroField) = +(b, a)
++(a::Number, b::ZeroField) = ConstantField(a)
 
 -(a::ZeroField, b::AbstractField) = -b
--(b::AbstractField, a::ZeroField) = b
+-(a::AbstractField, b::ZeroField) = a
 -(a::ZeroField, b::Number) = ConstantField(-b)
--(b::Number, a::ZeroField) = ConstantField(b)
+-(a::Number, b::ZeroField) = ConstantField(a)
 
 *(a::ZeroField, b::AbstractField) = a
-*(b::AbstractField, a::ZeroField) = *(a, b)
+*(a::AbstractField, b::ZeroField) = b
 *(a::ZeroField, b::Number) = a
-*(b::Number, a::ZeroField) = *(a, b)
+*(a::Number, b::ZeroField) = b
 
 /(a::ZeroField, b::AbstractField) = a
-/(b::AbstractField, a::ZeroField) = ConstantField(Inf)
+/(a::AbstractField, b::ZeroField) = ConstantField(Inf)
 /(a::ZeroField, b::Number) = a
-/(b::Number, a::ZeroField) = ConstantField(Inf)
+/(a::Number, b::ZeroField) = ConstantField(Inf)
 
 
 #####
--- a/src/AbstractOperations/binary_operations.jl
+++ b/src/AbstractOperations/binary_operations.jl
@@ -216,8 +216,6 @@ using Oceananigans.Fields: ZeroField
 /(a::ZeroField, b::Number) = a
 /(b::Number, a::ZeroField) = Inf
 
-^(a::ZeroField, b::Number) = a
-
 
 #####
 ##### Nested computations
--- a/src/AbstractOperations/binary_operations.jl
+++ b/src/AbstractOperations/binary_operations.jl
@@ -204,7 +204,7 @@ using Oceananigans.Fields: ZeroField, ConstantField
 ==(::ZeroField, ::ZeroField) = true
 
 ==(zf::ZeroField, cf::ConstantField) = 0 == cf.constant
-==(cf::ConstantField, zf::ZeroField) = ==(zf, cf)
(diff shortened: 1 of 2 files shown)


src/AbstractOperations/binary_operations.jl
```diff
@@ -187,6 +187,36 @@ macro binary(ops...)
...
+
+*(a::ZeroField, b::AbstractField) = a
+*(b::AbstractField, a::ZeroField) = *(a, b)
+*(a::ZeroField, b::Number) = a
+*(b::Number, a::ZeroField) = *(a, b)
+
+/(a::ZeroField, b::AbstractField) = a
+/(b::AbstractField, a::ZeroField) = Inf
+/(a::ZeroField, b::Number) = a
+/(b::Number, a::ZeroField) = Inf
```
simone-silvestri, line 217: Shouldn't this be `ConstantField(Inf)`?

src/AbstractOperations/binary_operations.jl
```diff
@@ -187,6 +187,36 @@ macro binary(ops...)
...
+==(::ZeroField, ::ZeroField) = true
+
++(a::ZeroField, b::AbstractField) = b
++(a::AbstractField, b::ZeroField) = +(b, a)
++(a::ZeroField, b::Number) = b
++(a::Number, b::ZeroField) = +(b, a)
+
+-(a::ZeroField, b::AbstractField) = -b
+-(b::AbstractField, a::ZeroField) = b
+-(a::ZeroField, b::Number) = -b
```
simone-silvestri, line 206: And maybe this `ConstantField(-b)` and so on?

src/AbstractOperations/binary_operations.jl
tomchor: Also, I just realized that it'd probably be good if we made stuff like the operation below work too, no?

```julia
julia> 1 + ConstantField(1)
ERROR: type ConstantField has no field grid
```

src/AbstractOperations/binary_operations.jl
```diff
@@ -187,6 +193,40 @@ macro binary(ops...)
...
+using Oceananigans.Fields: ZeroField, ConstantField
+
+==(::ZeroField, ::ZeroField) = true
+
+==(zf::ZeroField, cf::ConstantField) = 0 == cf.constant
+==(cf::ConstantField, zf::ZeroField) = ==(zf, cf)
+==(c1::ConstantField, c2::ConstantField) = c1.constant == c2.constant
+
++(a::ZeroField, b::AbstractField) = b
++(a::AbstractField, b::ZeroField) = +(b, a)
```
glwagner, line 210: It might be clearer to be a bit more explicit by writing

```julia
+(a::AbstractField, b::ZeroField) = a
```

and same for the others (I was confused when I read this code initially --- though I do understand the motivation for the other forms)

src/AbstractOperations/binary_operations.jl
```diff
@@ -187,6 +193,40 @@ macro binary(ops...)
...
++(a::Number, b::ZeroField) = +(b, a)
+
+-(a::ZeroField, b::AbstractField) = -b
+-(b::AbstractField, a::ZeroField) = b
+-(a::ZeroField, b::Number) = ConstantField(-b)
+-(b::Number, a::ZeroField) = ConstantField(b)
+
+*(a::ZeroField, b::AbstractField) = a
+*(b::AbstractField, a::ZeroField) = *(a, b)
+*(a::ZeroField, b::Number) = a
```
glwagner, line 221:
```suggestion
*(a::ZeroField, b::Number) = zero(b)
```

src/AbstractOperations/binary_operations.jl
```diff
@@ -187,6 +193,40 @@ macro binary(ops...)
...
+
++(a::ZeroField, b::AbstractField) = b
++(a::AbstractField, b::ZeroField) = +(b, a)
++(a::ZeroField, b::Number) = ConstantField(b)
++(a::Number, b::ZeroField) = +(b, a)
+
+-(a::ZeroField, b::AbstractField) = -b
+-(b::AbstractField, a::ZeroField) = b
+-(a::ZeroField, b::Number) = ConstantField(-b)
+-(b::Number, a::ZeroField) = ConstantField(b)
```
glwagner, line 217:
```suggestion
-(a::ZeroField, b::Number) = -b
-(b::Number, a::ZeroField) = b
```

is this better? It avoids the wrapper which could simplify operations (eg they can be reduced by julia lowering)

src/AbstractOperations/binary_operations.jl
```diff
@@ -187,6 +193,40 @@ macro binary(ops...)
...
+
+*(a::ZeroField, b::AbstractField) = a
+*(b::AbstractField, a::ZeroField) = *(a, b)
+*(a::ZeroField, b::Number) = a
+*(b::Number, a::ZeroField) = *(a, b)
+
+/(a::ZeroField, b::AbstractField) = a
+/(b::AbstractField, a::ZeroField) = ConstantField(Inf)
```
glwagner, line 225:
```suggestion
/(b::AbstractField, a::ZeroField) = convert(eltype(b), Inf)
```

src/AbstractOperations/binary_operations.jl
```diff
@@ -187,6 +193,40 @@ macro binary(ops...)
...
+-(b::AbstractField, a::ZeroField) = b
+-(a::ZeroField, b::Number) = ConstantField(-b)
+-(b::Number, a::ZeroField) = ConstantField(b)
+
+*(a::ZeroField, b::AbstractField) = a
+*(b::AbstractField, a::ZeroField) = *(a, b)
+*(a::ZeroField, b::Number) = a
+*(b::Number, a::ZeroField) = *(a, b)
+
+/(a::ZeroField, b::AbstractField) = a
```
glwagner, line 224:
```suggestion
/(a::ZeroField, b::AbstractField) = zero(eltype(b))
```

src/AbstractOperations/binary_operations.jl
```diff
@@ -187,6 +193,40 @@ macro binary(ops...)
...
++(a::ZeroField, b::Number) = ConstantField(b)
++(a::Number, b::ZeroField) = +(b, a)
+
+-(a::ZeroField, b::AbstractField) = -b
+-(b::AbstractField, a::ZeroField) = b
+-(a::ZeroField, b::Number) = ConstantField(-b)
+-(b::Number, a::ZeroField) = ConstantField(b)
+
+*(a::ZeroField, b::AbstractField) = a
+*(b::AbstractField, a::ZeroField) = *(a, b)
```
glwagner, line 220:
```suggestion
*(b::AbstractField, a::ZeroField) = zero(eltype(b))
```

src/AbstractOperations/binary_operations.jl
```diff
@@ -187,6 +193,40 @@ macro binary(ops...)
...
+*(a::ZeroField, b::Number) = a
+*(b::Number, a::ZeroField) = *(a, b)
+
+/(a::ZeroField, b::AbstractField) = a
+/(b::AbstractField, a::ZeroField) = ConstantField(Inf)
+/(a::ZeroField, b::Number) = a
+/(b::Number, a::ZeroField) = ConstantField(Inf)
```
glwagner, line 227:
```suggestion
/(b::Number, a::ZeroField) = convert(eltype(b), Inf)
```

src/AbstractOperations/binary_operations.jl
glwagner: Sorry I missed this. I think it's better not to use the `ConstantField` wrapper when it can be avoided, since this will yield simpler expression trees?

src/AbstractOperations/binary_operations.jl
glwagner: Ok, sorry I missed this comment. I gave the opposite advice below so we should discuss.

src/AbstractOperations/binary_operations.jl
tomchor: Agreed. Fixed

src/AbstractOperations/binary_operations.jl
simone-silvestri: but in this case, wouldn't we lose the information? (the result is not an abstract field anymore). I guess it depends on the end goal.

tomchor: In general I think it's likely that operations between a `ZeroField` and a `Number` are going to be used with other `AbstractFields`. So my guess is that outputting `ConstantField`s in those cases is probably advantageous more often than not.

glwagner: Okay, I think it makes sense to keep the `ConstantField` wrapper so that things remain in Field-world. Because, in fact, operations are simplified by these rules eg

```julia
julia> two = ConstantField(2)
ConstantField(2)

julia> c = CenterField(grid)
4×4×4 Field{Center, Center, Center} on RectilinearGrid on CPU

julia> 3 * two * c
BinaryOperation at (Center, Center, Center)
└── tree:
    * at (Center, Center, Center)
    ├── 6
    └── 4×4×4 Field{Center, Center, Center} on RectilinearGrid on CPU
```

is the same as

```julia
julia> 3 * 2 * c
BinaryOperation at (Center, Center, Center)
└── tree:
    * at (Center, Center, Center)
    ├── 6
    └── 4×4×4 Field{Center, Center, Center} on RectilinearGrid on CPU
```

glwagner: (I was worried that `3 * two * c` would be more complicated than `3 * 2 * c`)

src/AbstractOperations/binary_operations.jl
glwagner: Note that `Inf` is type `Float64`.

glwagner: So we don't want to use `Inf` because this can cause unintended type promotion for `Float32` grids, etc.

src/AbstractOperations/binary_operations.jl
glwagner:
```suggestion
*(a::AbstractField, b::ZeroField) = b
```

src/AbstractOperations/binary_operations.jl
tomchor: Good point. However, with that code a user can't use integers since there's no integer Inf:

```julia
julia> 2/ZeroField()
ERROR: InexactError: Int64(Inf)
Stacktrace:
 [1] Int64
   @ ./float.jl:788 [inlined]
 [2] convert(#unused#::Type{Int64}, x::Float64)
   @ Base ./number.jl:7
 [3] /(a::Int64, b::ZeroField{Int64, 3})
   @ Oceananigans.AbstractOperations ~/repos/Oceananigans.jl2/src/AbstractOperations/binary_operations.jl:228
 [4] top-level scope
   @ REPL[19]:1
 [5] top-level scope
   @ ~/.julia/packages/CUDA/DfvRa/src/initialization.jl:52

julia> convert(eltype(2), Inf)
ERROR: InexactError: Int64(Inf)
Stacktrace:
 [1] Int64
   @ ./float.jl:788 [inlined]
 [2] convert(#unused#::Type{Int64}, x::Float64)
   @ Base ./number.jl:7
 [3] top-level scope
   @ REPL[20]:1
 [4] top-level scope
   @ ~/.julia/packages/CUDA/DfvRa/src/initialization.jl:52
```

Are we okay with that?

src/AbstractOperations/binary_operations.jl
tomchor: Fixed

src/AbstractOperations/binary_operations.jl
glwagner: I'm not sure what else would be correct

tomchor: I guess an alternative way would be to define functions like

```julia
/(a::Number, b::ZeroField) = ConstantField(a/convert(eltype(a), 0))
```

and let Julia handle it. That way the output would follow the usual Julia conventions which rely on `propagate`:

```julia
julia> 2/0.0
Inf
```

tomchor: After thinking about it, I do think the above way is preferred for `::Number/::ZeroField`. So I made that change and changed the `AbstractField` division according your suggestion.

If you're okay with this, I think this is ready for merging.

src/AbstractOperations/binary_operations.jl
```diff
@@ -223,9 +223,9 @@ using Oceananigans.Fields: ZeroField, ConstantField
 *(a::Number, b::ZeroField) = b
 
 /(a::ZeroField, b::AbstractField) = a
-/(a::AbstractField, b::ZeroField) = ConstantField(Inf)
+/(a::AbstractField, b::ZeroField) = ConstantField(convert(eltype(a), Inf))
 /(a::ZeroField, b::Number) = a
-/(a::Number, b::ZeroField) = ConstantField(Inf)
+/(a::Number, b::ZeroField) = ConstantField(a/convert(eltype(a), 0))
```
glwagner, line 228:
```suggestion
/(a::Number, b::ZeroField) = ConstantField(a / convert(eltype(a), 0))
```

glwagner (approved):

src/AbstractOperations/binary_operations.jl
glwagner: Looks good

PR #2885, opened 2023-01-30
Clean up methods for computing matrix that corresponds to `linear_operation!` used by MultigridSolver

### LINKED ISSUE #2882: `Solvers.initialize_matrix` does not infer boundary conditions from `template_field`
I noticed that `similar(::Field)` does not transfer boundary conditions

https://github.com/CliMA/Oceananigans.jl/blob/a86ef32581192d82a655e1b2ff4411e79917a379/src/Fields/field.jl#L217-L227

so we might need to update 

https://github.com/CliMA/Oceananigans.jl/blob/c8a65a4fdeff25722104ced3a0c74d2d921ae1cf/src/Solvers/multigrid_solver.jl#L225-L244

and in particular 

https://github.com/CliMA/Oceananigans.jl/blob/c8a65a4fdeff25722104ced3a0c74d2d921ae1cf/src/Solvers/multigrid_solver.jl#L230

navidcy:
As discussed in #2882, `similar(::Field)` does not copy boundary conditions...

~This PR gives us the option to pass boundary conditions to `initialize_matrix` methods.~

This PR:
- renames `initialize_matrix` ->  `compute_matrix_for_linear_operation`
- moves into `matrix_solver_utils.jl`
- adds docstring

Resolves #2882


src/Solvers/multigrid_solver.jl
```diff
@@ -222,13 +222,20 @@ end
 @inline create_multilevel(::RugeStubenAMG, A) = ruge_stuben(A)
 @inline create_multilevel(::SmoothedAggregationAMG, A) = smoothed_aggregation(A)
 
-function initialize_matrix(::CPU, template_field, linear_operator!, args...)
+function initialize_matrix(::CPU, template_field, linear_operator!, args...; boundary_conditions=nothing)
```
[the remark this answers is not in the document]
navidcy: It's a generic utility that we have so that one can create the matrix that corresponds to `linear_operator!`. The matrix can be the one used by multigrid solver or the other matrix-based solvers.

At the moment it's not used in the implicit free surface solver. The 4 nested loops were rendering this very very slow so we ended up hard-coding the matrix that corresponds to the free surface eq instead of using this utility that would work for _any_ `linear_operator!`

https://github.com/CliMA/Oceananigans.jl/blob/c21d5d6ac95e762cac85154c8f3c4de63631fca7/src/Models/HydrostaticFreeSurfaceModels/mg_implicit_free_surface_solver.jl#L87 

But I was trying to use this to get the matrix that corresponds to some other `linear_operator!` with boundary conditions other from the default for a `CenterField` and I wasn't getting what I was expecting. I realized that boundary conditions were not communicated. Not sure if a kwarg is the best solution or modifying `similar(::Field)`...

navidcy:
btw the function belongs to a generic `matrix_utils.jl` file or something and instead of `initialize_matrix` it should be called `build_matrix`

src/Solvers/multigrid_solver.jl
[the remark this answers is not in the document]
glwagner: Got it! But this PR only affects one solver?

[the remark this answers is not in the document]
navidcy: yes

glwagner:
> btw the function belongs to a generic `matrix_utils.jl` file or something and instead of `initialize_matrix` it should be called `build_matrix`

"operator" might be more clear? Or some other semantics. We want to say something like "Solve A*x=b, where A is..." I don't think we want to say "A is The Matrix". Something more specific?

navidcy:
Yes. In the meantime I found more problems with this utility function. There are lots of assumptions that it operates on CenterFields for example.

glwagner:
Got it.

Also whatever changes we make re: boundary conditions I think should also be made to the PreconditionedConjugateGradientSolver

src/Solvers/matrix_solver_utils.jl
```diff
@@ -95,3 +95,124 @@ end
 @inline validate_laplacian_size(N, dim) = dim == true ? N : 1
   
 @inline ensure_diagonal_elements_are_present!(A) = fkeep!(A, (i, j, x) -> (i == j || !iszero(x)))
+
+"""
+    make_column(f::Field)
+
+Return the column vector with the interior of a field `f`.
+"""
+function make_column(f::Field)
```
glwagner, line 104: Should we extend `vec` for this?

Docs for `vec`:

```
help?> vec
search: vec Vector VecOrMat VecElement BitVector DenseVector DenseVecOrMat StridedVector StridedVecOrMat AbstractVector AbstractVecOrMat InvalidStateException

  vec(a::AbstractArray) -> AbstractVector

  Reshape the array a as a one-dimensional column vector. Return a if it is already an AbstractVector. The resulting array shares the same underlying data as a, so it
  will only be mutable if a is mutable, in which case modifying one will also modify the other.
```

navidcy: sounds like we should!

navidcy:
```
julia> grid = LatitudeLongitudeGrid(arch, size = (3, 3, 3),
                                    longitude = (0, 60),
                                    latitude = (15, 75),
                                    z = (-1, 0), halo=(3, 3, 4))
3×3×3 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×4 halo and with precomputed metrics
├── longitude: Bounded  λ ∈ [0.0, 60.0]  regularly spaced with Δλ=20.0
├── latitude:  Bounded  φ ∈ [15.0, 75.0] regularly spaced with Δφ=20.0
└── z:         Bounded  z ∈ [-1.0, 0.0]  regularly spaced with Δz=0.333333

julia> c = CenterField(grid)
3×3×3 Field{Center, Center, Center} on LatitudeLongitudeGrid on CPU
├── grid: 3×3×3 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×4 halo and with precomputed metrics
│   └── west: ZeroFlux, east: ZeroFlux, south: ZeroFlux, north: ZeroFlux, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 9×9×11 OffsetArray(::Array{Float64, 3}, -2:6, -2:6, -3:7) with eltype Float64 with indices -2:6×-2:6×-3:7

julia> vec(c) == reshape(interior(c), 3*3*3)
true
```

src/Solvers/matrix_solver_utils.jl
```diff
@@ -95,3 +95,114 @@ end
...
+
+        CUDA.@allowscalar eᵢⱼₖ[i, j, k] = 1
+
+        fill_halo_regions!(eᵢⱼₖ)
+
+        linear_operation!(Aeᵢⱼₖ, eᵢⱼₖ, args...)
+
+        count = 0
+        for n in 1:Nz, m in 1:Ny, l in 1:Nx
+            Aeᵢⱼₖₗₘₙ = CUDA.@allowscalar Aeᵢⱼₖ[l, m, n]
```
glwagner, line 196: should we make a kernel?

navidcy: it's only for CPU
is there still advantage? I guess it can be speedup by the number of threads?

navidcy: Yes! Good point.

navidcy: I think I'd like to merge this so that I can move the AMGX stuff out of Oceananigans. So creation of a kernel can wait...

simone-silvestri (approved):

PR #2888, opened 2023-01-31
Parallel and Serial Split Explicit Free Surface

simone-silvestri:
This PR improves the currently implemented split explicit surface solver in serial mode
(partially using the algorithm in Shchepetkin & McWilliams, Ocean Modelling 9, 2005 albeit with a linear free surface) and implements single-node parallel (`MultiRegion`), and multi-node parallel (`Distributed`) split explicit free surface

In addition, quite some improvements have been made to the distributed module to allow a distributed `LatitudeLongituteGrid` and `IBG` and various comments in the `MultiRegion` module

Edit: MPI does not exploit CUDA-aware message passing for CuArray views, so this PR also implements buffered Halo communication for distributed models, unifying a bit of the code for Distributed and MultiRegion. This is a fundamental step to achieve the goals of the next PR which will deal with heterogenous distributed - shared models (i.e. a MultiRegionGrid of a DistributedGrid) 

Edit Edit: apparently CUDA-aware MPI allows passing views of CuArrays. Buffers are still implemented for those architectures where CUDA-aware MPI is not available

Edit Edit Edit: apparently for Oceananigans that sends strided memory, buffers are crucial because
> Note that derived datatypes, both contiguous and non-contiguous, are supported. However, the non-contiguous
> datatypes currently have high overhead because of the many calls to cuMemcpy to copy all the pieces of the
> buffer into the intermediate buffer.

(from [openmpi-link](https://www.open-mpi.org/faq/?category=runcuda#mpi-cuda-support))

Edit: This PR adds the possibility of distributing along a `Bounded` direction, and the possibility of having correct 2D parallelization with Coriolis. Previously it was not possible because corners were not communicated correctly

Doesn't have to be merged now.

(Validation cases are not necessary and will be deleted prior to merging)


simone-silvestri:
Difference between fill halo on 4 GPUs between main and this PR

This is the script
```
using Oceananigans
using Oceananigans.Distributed

using MPI
using CUDA

MPI.Init()
comm     = MPI.COMM_WORLD
rank        = MPI.Comm_rank(comm)
Nranks   = MPI.Comm_size(comm)
local_comm = MPI.Comm_split_type(comm, MPI.COMM_TYPE_SHARED, rank)
local_rank    = MPI.Comm_rank(local_comm)

CUDA.device!(local_rank)

child_arch = GPU()
arch = MultiArch(child_arch; ranks = (Nranks, 1, 1))

grid  = RectilinearGrid(arch, size = (1440, 600, 100), extent = (1, 1, 1), halo = (5, 5, 5))
field = CenterField(grid)

for i in 1:4
   fill_halo_regions!(field)
end

@show "finished compiling"

start_time = time_ns()
for i in 1:100
    fill_halo_regions!(field)
end

@show rank, Float64.(time_ns() - start_time) * 1e-6
```

This is on main
```
"finished compiling" = "finished compiling"
"finished compiling" = "finished compiling"
"finished compiling" = "finished compiling"
"finished compiling" = "finished compiling"
(rank, Float64.(time_ns() - start_time) * 1.0e-6) = (0, 32910.190702)
(rank, Float64.(time_ns() - start_time) * 1.0e-6) = (1, 32906.484328)
(rank, Float64.(time_ns() - start_time) * 1.0e-6) = (2, 32908.786154)
(rank, Float64.(time_ns() - start_time) * 1.0e-6) = (3, 32908.788149)
```

This PR

```
"finished compiling" = "finished compiling"
"finished compiling" = "finished compiling"
"finished compiling" = "finished compiling"
"finished compiling" = "finished compiling"
(rank, Float64.(time_ns() - start_time) * 1.0e-6) = (1, 283.194961)
(rank, Float64.(time_ns() - start_time) * 1.0e-6) = (2, 283.23857899999996)
(rank, Float64.(time_ns() - start_time) * 1.0e-6) = (3, 283.226133)
(rank, Float64.(time_ns() - start_time) * 1.0e-6) = (0, 283.19493)
```

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -3,8 +3,8 @@ module BoundaryConditions
 export
     BCType, Flux, Gradient, Value, Open,
     BoundaryCondition, getbc, setbc!,
-    PeriodicBoundaryCondition, OpenBoundaryCondition, NoFluxBoundaryCondition, CommunicationBoundaryCondition,
-    FluxBoundaryCondition, ValueBoundaryCondition, GradientBoundaryCondition,
+    PeriodicBoundaryCondition, OpenBoundaryCondition, NoFluxBoundaryCondition, CommunicationBoundaryCondition, 
+    FluxBoundaryCondition, ValueBoundaryCondition, GradientBoundaryCondition, HaloCommunicationBoundaryCondition,
```
glwagner, line 7: What's the difference between "CommunicationBoundaryCondition" and "HaloCommunicationBoundaryCondition"

simone-silvestri: One is for shared communication (MultiRegion) the other for distributed communication (DistributedGrid). Probably we can rename them to reflect this difference

glwagner: So they both communicate halos?

glwagner: I think if they both communicate the same thing then one way to distinguish them would be `MultiRegionCommunicationBoundaryCondition` and `DistributedCommunicationBoundaryCondition`. That names them after the modules they refer to, I think

glwagner:
"MultiArch" and "MultiRegion" is also confusing

src/Advection/vector_invariant_advection.jl
```diff
@@ -124,7 +124,7 @@ Adapt.adapt_structure(to, scheme::VectorInvariant{N, FT}) where {N, FT} =
 const VectorInvariantEnergyConserving    = VectorInvariant{<:Any, <:Any, <:EnergyConservingScheme}
 const VectorInvariantEnstrophyConserving = VectorInvariant{<:Any, <:Any, <:EnstrophyConservingScheme}
 
-const VectorInvariantConserving = Union{VectorInvariantEnergyConserving, VectorInvariantEnstrophyConserving}
+const VectorInvariantVerticallyConserving  = VectorInvariant{<:Any, <:Any, <:Any, Nothing, <:Any, <:Any, <:EnergyConservingScheme}
```
glwagner, line 127: what does VerticallyConserving mean?

src/Distributed/distributed_fields.jl
```diff
@@ -1,12 +1,22 @@
...
+using Oceananigans.Fields: validate_field_data, validate_boundary_conditions, validate_indices
+
 function Field((LX, LY, LZ)::Tuple, grid::DistributedGrid, data, old_bcs, indices::Tuple, op, status)
     arch = architecture(grid)
+    validate_field_data((LX, LY, LZ), data, grid, indices)
+    validate_boundary_conditions((LX, LY, LZ), grid, old_bcs)
+    indices = validate_indices(indices, (LX, LY, LZ), grid)
     new_bcs = inject_halo_communication_boundary_conditions(old_bcs, arch.local_rank, arch.connectivity)
-    buffers = FieldBoundaryBuffers(nothing, nothing, nothing, nothing)
+    buffers = FieldBoundaryBuffers(grid, data, new_bcs)
```
glwagner, line 14: Do we still need this or are the buffers now stored in the boundary conditions?

src/Distributed/distributed_fields.jl
```diff
@@ -1,12 +1,22 @@
...
     return Field{LX, LY, LZ}(grid, data, new_bcs, indices, op, status, buffers)
 end
 
 const DistributedField      = Field{<:Any, <:Any, <:Any, <:Any, <:DistributedGrid}
 const DistributedFieldTuple = NamedTuple{S, <:NTuple{N, DistributedField}} where {S, N}
+
+# TODO: make sure the definition of architecture is consistent
```
glwagner, line 21: consistent with what?

src/Distributed/distributed_grids.jl
```diff
@@ -1,19 +1,30 @@
...
 using Oceananigans.Grids: cpu_face_constructor_x, cpu_face_constructor_y, cpu_face_constructor_z
+using Oceananigans.Grids: R_Earth, metrics_precomputed
+
+using Oceananigans.ImmersedBoundaries
 
 import Oceananigans.Grids: RectilinearGrid, LatitudeLongitudeGrid, with_halo
 
 const DistributedGrid{FT, TX, TY, TZ} = AbstractGrid{FT, TX, TY, TZ, <:MultiArch}
+const DistributedRectilinearGrid{FT, TX, TY, TZ, FX, FY, FZ, VX, VY, VZ} =
+    RectilinearGrid{FT, TX, TY, TZ, FX, FY, FZ, VX, VY, VZ, <:MultiArch} where {FT, TX, TY, TZ, FX, FY, FZ, VX, VY, VZ}
```
glwagner, line 15: Should the type of the architecture be moved forward among the type parameters to make this more readable?

src/Distributed/distributed_grids.jl
```diff
@@ -59,22 +74,32 @@ function RectilinearGrid(arch::MultiArch, FT = Float64;
...
+"""
+    LatitudeLongitudeGrid(arch::MultiArch, FT=Float64; kw...)
+
+Return the rank-local portion of `LatitudeLongitudeGrid` on `arch`itecture.
+"""
 function LatitudeLongitudeGrid(arch::MultiArch,
-                               FT = Float64; 
-                               precompute_metrics = false,
+                               FT::DataType = Float64; 
+                               precompute_metrics = true,
```
glwagner, line 84: Does this change performance?

src/Distributed/distributed_grids.jl
```diff
@@ -121,13 +151,18 @@ function reconstruct_global_grid(grid::RectilinearGrid)
...
     y = cpu_face_constructor_y(grid)
     z = cpu_face_constructor_z(grid)
 
-    xG = assemble(x, nx, Rx, i, arch)
-    yG = assemble(y, ny, Ry, j, arch)
-    zG = assemble(z, nz, Rz, k, arch)
+    ## This will not work with 2D parallelizations!!
+    xG = Rx == 1 ? x : assemble(x, nx, Rx, ri, arch)
+    yG = Ry == 1 ? y : assemble(y, ny, Ry, rj, arch)
+    zG = Rz == 1 ? z : assemble(z, nz, Rz, rk, arch)
```
glwagner, line 165: the names `xG` etc are potentially confusing to people who know MITgcm because they have a meaning there (totally unrelated to this one).

src/Distributed/distributed_grids.jl
```diff
@@ -166,10 +270,61 @@ end
...
 
 function scatter_local_grids(arch::MultiArch, global_grid::LatitudeLongitudeGrid)
     x, y, z, topo, sz, halo = scatter_grid_properties(global_grid)
-    return LatitudeLongitudeGrid(arch, eltype(global_grid); size=sz, longitude=x, latitude=y, z=z, halo=halo)
+    return LatitudeLongitudeGrid(arch, eltype(global_grid); size=sz, longitude=x, latitude=y, z=z, halo=halo, topology=topo)
+end
+
+function scatter_local_grids(arch::MultiArch, global_grid::ImmersedBoundaryGrid)
+    ib   = global_grid.immersed_boundary
+    ug   = global_grid.underlying_grid
```
glwagner, line 283:
```suggestion
    ib = global_grid.immersed_boundary
    ug = global_grid.underlying_grid
```

src/Distributed/distributed_grids.jl
```diff
@@ -166,10 +270,61 @@ end
...
+insert_connected_topology(::Type{Periodic}, R, r) = ifelse(R == 1, Periodic, FullyConnected)
+
+""" 
+    reconstruct_global_topology(T, R, r, comm)
+
+reconstructs the global topology associated with the local topologies `T`, the amount of ranks 
+in `T` direction (`R`) and the local rank index `r`. If all ranks hold a `FullyConnected` topology,
+the global topology is `Periodic`, otherwise it is `Bounded`
+"""
+function reconstruct_global_topology(T, R, r, comm)
```
glwagner, line 313: Is it "reconstruct", or just "construct"?

src/Distributed/halo_communication.jl
```diff
@@ -65,75 +70,83 @@ function tupled_fill_halo_regions!(full_fields, grid::DistributedGrid, args...;
...
 
-#####
-##### fill_west_and_east_halos!   }
-##### fill_south_and_north_halos! } for non-communicating boundary conditions (fallback)
-##### fill_bottom_and_top_halos!  }
-#####
+function fill_halo_event!(task, halo_tuple, c, indices, loc, arch::MultiArch, barrier, grid::DistributedGrid, args...; kwargs...)
+    fill_halo!  = halo_tuple[1][task]
+    bc_left     = halo_tuple[2][task]
+    bc_right    = halo_tuple[3][task]
```
glwagner, line 108:
```suggestion
    fill_halo! = halo_tuple[1][task]
    bc_left    = halo_tuple[2][task]
    bc_right   = halo_tuple[3][task]
```

src/Distributed/partition_assemble.jl
```diff
@@ -21,11 +23,92 @@ and `arch`itecture.
...
+
+        return arch_array(child_architecture(arch), c_local)
+    end
+end
+
+"""
+    reconstruct_global_array(arch, c_global, (nx, ny, nz))
+
+reconstruct local arrays (2D of size nx, ny or 3D of size nx, ny, nz) in local arrays.
+Usefull for boundary arrays, forcings and initial conditions
```
glwagner, line 77:
```suggestion
Returns a local arrays (2D of size nx, ny or 3D of size nx, ny, nz).
Useful for boundary arrays, forcings and initial conditions.
```

I don't completely understand how this function is used. We are constructing a global array, but the input is `c_global`?

glwagner: Also is this just "construct" not "reconstruct"?

src/Distributed/partition_assemble.jl
```diff
@@ -21,11 +23,92 @@ and `arch`itecture.
...
+"""
+reconstruct_global_array(arch, c_local::Function, N) = c_local
+
+function reconstruct_global_array(arch, c_local::AbstractArray, n)
```
glwagner, line 81: "construct" or "reconstruct"? (Is there a sense that `c_local` existed as a global array at some prior time?)

src/ImmersedBoundaries/ImmersedBoundaries.jl
```diff
@@ -214,6 +221,10 @@ As well as
 @inline immersed_peripheral_node(i, j, k, ibg::IBG, LX, LY, LZ) =  peripheral_node(i, j, k, ibg, LX, LY, LZ) &
                                                                   !peripheral_node(i, j, k, ibg.underlying_grid, LX, LY, LZ)
 
+@inline immersed_inactive_node(i, j, k, ibg::IBG, LX, LY, LZ) =  inactive_node(i, j, k, ibg, LX, LY, LZ) &
+                                                                  !inactive_node(i, j, k, ibg.underlying_grid, LX, LY, LZ)
```
glwagner, line 225:
```suggestion
@inline immersed_inactive_node(i, j, k, ibg::IBG, LX, LY, LZ) =  inactive_node(i, j, k, ibg, LX, LY, LZ) &
                                                                !inactive_node(i, j, k, ibg.underlying_grid, LX, LY, LZ)
```

glwagner (commented):
I tried my best. There's too much to fully review.

src/BoundaryConditions/BoundaryConditions.jl
simone-silvestri: Ok, I'll wait to have a commit with all tests passing, then I will do all the refactoring. So I know I have a commit to revert to if I mess up

src/Advection/vector_invariant_advection.jl
simone-silvestri: A conservative scheme in the vertical direction

src/Distributed/distributed_fields.jl
simone-silvestri: nono, still in the Fields. I would have to say that having them in the field was very convenient for this PR and it probably will be if we have a common buffer for multiple fields.

src/Distributed/distributed_grids.jl
simone-silvestri: I remember on GPU it did.

src/Distributed/distributed_fft_based_poisson_solver.jl
```diff
@@ -4,6 +4,8 @@ using PencilArrays: Permutation
 import Oceananigans.Solvers: poisson_eigenvalues, solve!
 import Oceananigans.Architectures: architecture
 
+using FFTW
```
glwagner, line 7: I think this used to be

```julia
import FFTW
```

which was lost somehow

simone-silvestri:
a 20-year integration of the quarter degree simulation (in `validation/near_global_lat_lon/near_global_quarter_degree.jl`) using the following as a free surface:

```
Δt = 10minutes  # probably we can go to 10min or 15min?
CFL = 0.7

wave_speed =  sqrt(g_Earth * grid.Lz)
Δg         = 1 / sqrt(1 / min_Δx(grid)^2 + 1 / min_Δy(grid)^2)

@show substeps = Int(ceil(2 * Δt / (CFL / wave_speed * Δg)))

free_surface = SplitExplicitFreeSurface(; substeps)
```

https://user-images.githubusercontent.com/33547697/217328677-9c737e8d-140b-4a9f-92b9-d42f5d364a3d.mp4

src/Distributed/distributed_fields.jl
simone-silvestri: `architecture(grid)` and `architecture(field)` would give two different results, so I am a bit hesitant about this. I think in the future I want to leave the architecture just as `CPU` or `GPU` and move the `Distributed` property to the grid.

src/MultiRegion/multi_region_boundary_conditions.jl
```diff
@@ -68,11 +68,11 @@ function fill_halo_regions!(c::MultiRegionObject, bcs, indices, loc, mrg::MultiR
     neighbors = Reference(c.regional_objects)
     buff      = Reference(buffers.regional_objects)
     
-    @apply_regionally event1 = fill_halo_event!(1, halo_tuple, c, indices, loc, arch, device_event(arch), mrg, neighbors, buff, args...; kwargs...)
-    @apply_regionally event2 = fill_halo_event!(2, halo_tuple, c, indices, loc, arch, event2,             mrg, neighbors, buff, args...; kwargs...)
-    @apply_regionally event3 = fill_halo_event!(3, halo_tuple, c, indices, loc, arch, event3,             mrg, neighbors, buff, args...; kwargs...)
+    event1 = construct_regionally(fill_halo_event!, 1, halo_tuple, c, indices, loc, arch, device_event(arch), mrg, neighbors, buff, args...; kwargs...)
+    event2 = construct_regionally(fill_halo_event!, 2, halo_tuple, c, indices, loc, arch, event2,             mrg, neighbors, buff, args...; kwargs...)
+    event3 = construct_regionally(fill_halo_event!, 3, halo_tuple, c, indices, loc, arch, event3,             mrg, neighbors, buff, args...; kwargs...)
```
glwagner: What do 1, 2, 3 stand for?

simone-silvestri: Fill halo in direction 1, 2 and 3. We do not know which direction comes first a priori so the fill halo are anonymous. I could put it in a loop like we do in `fill_halo_regions.jl`.

glwagner: got it

simone-silvestri:
This is ready to merge (I'll figure out later on issue #2922)

glwagner:
> I share here a double drake experiment that makes full use of this PR
> 
> #### Grid setup
> * latitude longitude grid from 75 S to 75 N
> * 1/3 of a degree in the horizontal (1080 points in longitude, 450 in latitude)
> * 150 exponentially stretched vertical levels for a 3km deep ocean
> * double drake bathymetry (https://doi.org/10.1175/2009JCLI3197.1)
> 
> #### Top BC:
> * temperature: restoring to reference profile (cosine shape)
> * salinity: prescribed latitudinally dependent surface flux
> * zonal velocity: prescribed latitudinally dependent wind stress
> 
> #### Bottom BC:
> * velocities: linear bottom drag with a drag coefficient of 0.003 ms⁻¹
> 
> #### Initial conditions
> * zero velocities
> * exponentially stratified temperature with SST equal to the reference temperature
> * constant salinity
> 
> #### Model setup
> * linear equation of state
> * Richardson-based diffusivity for BL mixing
> * vertical background viscosity and diffusivity of 5e-4 and 3e-5, respectively
> * vector invariant momentum advection with WENO for vorticity and divergence flux as well as vertical transport (no horizontal viscosity)
> * WENO for tracer advection (no horizontal diffusivity)
> * Split explicit free surface using an averaging shape function and a CFL of 0.7 (23 substeps per time step)
> 
> #### Simulation setup
> * time step of 10 minutes
> * ran on 2 MPI processes with CUDA-aware MPI
> * performs about 10 simulated years per day
> 
> On the left, there is the free surface evolution, on the right the surface vertical vorticity (evolved for 9 years)

This would be a good setup to add to `ClimaOcean.jl`.

simone-silvestri:
at the moment is implemented in [OceanScalingTests.jl](https://github.com/simone-silvestri/OceanScalingTests.jl). I can implement it also in ClimaOcean.jl

.buildkite/pipeline.yml
```diff
@@ -22,6 +22,9 @@ steps:
       - "$SVERDRUP_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.precompile()'"
       - "$SVERDRUP_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.status()'"
 
+      # Set up the mpiexecjl command
+      - "$SVERDRUP_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using MPI; MPI.install_mpiexecjl()'"
```
glwagner, line 26: Can we remove this since we aren't doing distributed tests on sverdrup?

benchmark/distributed_shallow_water_model_mpi.jl
```diff
@@ -30,7 +30,7 @@ Ry = parse(Int, ARGS[4])
 @info "Setting up distributed shallow water model with N=($Nx, $Ny) grid points and ranks=($Rx, $Ry) on rank $local_rank..."
 
 topo = (Periodic, Periodic, Flat)
-arch = MultiArch(CPU(),topology=topo, ranks=(Rx, Ry, 1), communicator=MPI.COMM_WORLD)
+arch = DistributedArch(CPU(),topology=topo, ranks=(Rx, Ry, 1), communicator=MPI.COMM_WORLD)
```
glwagner, line 33:
```suggestion
arch = DistributedArch(CPU(), topology=topo, ranks=(Rx, Ry, 1), communicator=MPI.COMM_WORLD)
```

benchmark/distributed_shallow_water_model_mpi.jl
```diff
@@ -30,7 +30,7 @@ Ry = parse(Int, ARGS[4])
...
 distributed_grid = RectilinearGrid(arch, topology=topo, size=(Nx, Ny), extent=(1, 1))
 model = ShallowWaterModel(grid=distributed_grid, gravitational_acceleration=1.0)
 set!(model, h=1.0)
```
glwagner, line 36:
```suggestion
set!(model, h=1)
```

examples/shallow_water_Bickley_jet.jl
```diff
@@ -103,7 +103,7 @@ v = vh / h
 compute!(ω)
 
 ## Copy mean vorticity to a new field
-ωⁱ = Field{Face, Face, Nothing}(model.grid)
+ωⁱ = Field((Face, Face, Nothing), model.grid)
```
glwagner, line 106: Why?

src/Advection/vector_invariant_advection.jl
glwagner: The other form also conserves momentum right? The difference is not whether they conserve momentum but whether we are using the "conservative form" (eg with this semantics, the "other" scheme is also "conserving"). How about `VerticallyConservativeVectorInvariant`?

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -27,18 +27,32 @@ for dir in (:west, :east, :south, :north, :bottom, :top)
...
-
+# For inhomogeneous BC we extract the _last_ one 
+# example 
+# `bc.west <: DCBC`
+# `bc.east <: PBC`
+# `extract_west_or_east_bc(bc) == bc.west`
+    
+  extract_west_or_east_bc(bc) = sort([bc.west,   bc.east],  lt=fill_first)[2]
+extract_south_or_north_bc(bc) = sort([bc.south,  bc.north], lt=fill_first)[2]
+ extract_bottom_or_top_bc(bc) = sort([bc.bottom, bc.top],   lt=fill_first)[2]
```
glwagner, line 38: Does this allocate? Using a tuple instead of array might not, eg `(bc.bottom, bc.top)`

src/Distributed/distributed_fields.jl
glwagner: Do you mean it is more convenient to have buffers in `Field.boundary_buffers` versus something like `Field.boundary_conditions.top.boundary_buffer`? Why is the first more convenient?

src/Distributed/distributed_fields.jl
glwagner: why don't we use 

```julia
architecture(field) = architecture(field.grid)
```
?

src/Fields/constant_field.jl
```diff
@@ -21,6 +21,6 @@ ConstantField(constant) = ConstantField{3}(constant)
 
 const CF = Union{ConstantField, ZeroField, OneField}
 
-@inline Base.summary(f::CF) = string("ConstantField(", prettysummary(f.constant), ")")
+@inline Base.summary(f::CF) = string("ConstantField(", f[1, 1, 1], ")")
```
glwagner, line 24: Why don't we use `prettysummary` here?

src/Fields/field_boundary_buffers.jl
```diff
@@ -25,16 +28,119 @@ function FieldBoundaryBuffers(grid, data, boundary_conditions)
...
+
+fills `buffers.send` from OffsetArray `c` preparing for message passing. If we are on CPU
+we do not need to fill the buffers as the transfer can happen through views
+"""
+function fill_west_and_east_send_buffers!(c::OffsetArray, buffers::FieldBoundaryBuffers, grid)
+    Hx, Hy, _ = halo_size(grid)
+    Nx, Ny, _ = size(grid)
+
+     _fill_west_send_buffer!(parent(c), buffers.west, Hx, Nx)
+     _fill_east_send_buffer!(parent(c), buffers.east, Hx, Nx)
```
glwagner, line 71:
```suggestion
    _fill_west_send_buffer!(parent(c), buffers.west, Hx, Nx)
    _fill_east_send_buffer!(parent(c), buffers.east, Hx, Nx)
```

src/Fields/field_boundary_buffers.jl
```diff
@@ -25,16 +28,119 @@ function FieldBoundaryBuffers(grid, data, boundary_conditions)
...
+end
+
+function fill_south_and_north_send_buffers!(c::OffsetArray, buffers::FieldBoundaryBuffers, grid)
+    Hx, Hy, _ = halo_size(grid)
+    Nx, Ny, _ = size(grid)
+
+     _fill_south_send_buffer!(parent(c), buffers.south, Hy, Ny)
+     _fill_north_send_buffer!(parent(c), buffers.north, Hy, Ny)
```
glwagner, line 79:
```suggestion
    _fill_south_send_buffer!(parent(c), buffers.south, Hy, Ny)
    _fill_north_send_buffer!(parent(c), buffers.north, Hy, Ny)
```

src/Fields/field_boundary_buffers.jl
```diff
@@ -25,16 +28,119 @@ function FieldBoundaryBuffers(grid, data, boundary_conditions)
...
+    _recv_from_south_buffer!(parent(c), buffers.south, Hy, Ny)
+    _recv_from_north_buffer!(parent(c), buffers.north, Hy, Ny)
+end
+
+function recv_from_buffers!(c::OffsetArray, buffers::FieldBoundaryBuffers, grid, ::Val{:west_and_east})
+    Hx, Hy, _ = halo_size(grid)
+    Nx, Ny, _ = size(grid)
+
+     _recv_from_west_buffer!(parent(c), buffers.west, Hx, Nx)
+     _recv_from_east_buffer!(parent(c), buffers.east, Hx, Nx)
```
glwagner, line 115:
```suggestion
    _recv_from_west_buffer!(parent(c), buffers.west, Hx, Nx)
    _recv_from_east_buffer!(parent(c), buffers.east, Hx, Nx)
```

src/Fields/field_boundary_buffers.jl
```diff
@@ -25,16 +28,119 @@ function FieldBoundaryBuffers(grid, data, boundary_conditions)
...
+end
+
+function recv_from_buffers!(c::OffsetArray, buffers::FieldBoundaryBuffers, grid, ::Val{:south_and_north})
+    Hx, Hy, _ = halo_size(grid)
+    Nx, Ny, _ = size(grid)
+
+    _recv_from_south_buffer!(parent(c), buffers.south, Hy, Ny)
+    _recv_from_north_buffer!(parent(c), buffers.north, Hy, Ny)
```
glwagner, line 123:
```suggestion
   _recv_from_south_buffer!(parent(c), buffers.south, Hy, Ny)
   _recv_from_north_buffer!(parent(c), buffers.north, Hy, Ny)
```

src/Forcings/advective_forcing.jl
```diff
@@ -45,7 +45,7 @@ settling = AdvectiveForcing(UpwindBiasedFifthOrder(), w=w_Stokes)
 AdvectiveForcing with the UpwindBiased scheme:
 ├── u: ZeroField{Int64}
 ├── v: ZeroField{Int64}
-└── w: ConstantField(-1.97096)
+└── w: ConstantField(-1.970964448157431)
```
glwagner, line 48: Is this because we aren't using `prettysummary` anymore? I think we should use it --- it could also help make things look cleaner if running with different precisions becomes common

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
```diff
@@ -67,21 +68,37 @@ Return a grid with `GridFittedBottom` immersed boundary.
 Computes ib.bottom_height and wraps in an array.
 """
 function ImmersedBoundaryGrid(grid, ib::AbstractGridFittedBottom)
-    bottom_field = Field{Center, Center, Nothing}(grid)
+    bottom_field = Field((Center, Center, Nothing), grid)
```
glwagner, line 71: Why change syntax? (We could eliminate the first constructor if we don't like it --- it's only purpose is readability)

src/Utils/Utils.jl
```diff
@@ -23,6 +23,8 @@ import CUDA  # To avoid name conflicts
 instantiate(x) = x
 instantiate(X::DataType) = X()
 
+getname(type) = typeof(type).name.wrapper
```
glwagner, line 26: Perhaps this should be called `getnamewrapper` (not sure what the difference between `name` and `name.wrapper` is)

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
```diff
@@ -67,21 +68,37 @@ Return a grid with `GridFittedBottom` immersed boundary.
...
     set!(bottom_field, ib.bottom_height)
     fill_halo_regions!(bottom_field)
     offset_bottom_array = dropdims(bottom_field.data, dims=3)
 
-    # TODO: maybe clean this up
-    IB = typeof(ib).name.wrapper
-    new_ib = IB(offset_bottom_array)
+    new_ib = getname(ib)(offset_bottom_array)
```
glwagner, line 76: I think we should keep the `TODO` comment because I'm not sure this is a robust pattern

glwagner: There may be a more julian way to write this

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
```diff
@@ -67,21 +68,37 @@ Return a grid with `GridFittedBottom` immersed boundary.
...
+        cpu_array    = arch_array(CPU(), ib.bottom_height)[1:N[1], 1:N[2]] 
+        set!(bottom_field, cpu_array)
+        fill_halo_regions!(bottom_field)
+        offset_bottom_array = dropdims(bottom_field.data, dims=3)
+    
+        new_ib = getname(ib)(offset_bottom_array)
+    
+        return ImmersedBoundaryGrid(grid, new_ib)
+    end
+
```
glwagner, line 101: Should we write a new function for a `grid` on `DistributedArch` instead?

glwagner: I'm not sure why that `if` statement will always only be triggered for distributed arch

glwagner: So I think it's better to be explicit. We could also write

```julia
if architecture(grid) isa DistributedArch
```

for example

src/Models/HydrostaticFreeSurfaceModels/distributed_split_explicit_free_surface.jl
```diff
@@ -0,0 +1,87 @@
...
+    Gⱽ = Field((Center, Face,   Nothing), grid)
+    
+    Hᶠᶜ = Field((Face,   Center, Nothing), grid)
+    Hᶜᶠ = Field((Center, Face,   Nothing), grid)
+    Hᶜᶜ = Field((Center, Center, Nothing), grid)
+    
+    vertical_height!(Hᶠᶜ, (Face, Center, Center))
+    vertical_height!(Hᶜᶠ, (Center, Face, Center))
+
+    vertical_height!(Hᶜᶜ, (Center, Center, Center))
```
glwagner, line 19: Are there other heights? Horizontal height?

src/Models/HydrostaticFreeSurfaceModels/distributed_split_explicit_free_surface.jl
```diff
@@ -0,0 +1,87 @@
...
+    kernel_size    = augmented_kernel_size(grid)
+    kernel_offsets = augmented_kernel_offsets(grid)
+    
+    return SplitExplicitAuxiliary(Gᵁ, Gⱽ, Hᶠᶜ, Hᶜᶠ, Hᶜᶜ, kernel_size, kernel_offsets)
+end
+
+@inline function vertical_height!(height, location)
+    dz = GridMetricOperation(location, Δz, height.grid)
+    sum!(height, dz)
+end
```
glwagner, line 33:
```suggestion
    return sum!(height, dz)
end
```

src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
```diff
@@ -89,8 +89,8 @@ function FreeSurface(free_surface::ImplicitFreeSurface{Nothing}, velocities, gri
     gravitational_acceleration = convert(eltype(grid), free_surface.gravitational_acceleration)
 
     # Initialize barotropic volume fluxes
-    barotropic_x_volume_flux = Field{Face, Center, Nothing}(grid)
-    barotropic_y_volume_flux = Field{Center, Face, Nothing}(grid)
+    barotropic_x_volume_flux = Field((Face, Center, Nothing), grid)
+    barotropic_y_volume_flux = Field((Center, Face, Nothing), grid)
```
glwagner, line 93: This style change is omnipresent -- what's the reasoning?

src/Models/HydrostaticFreeSurfaceModels/pcg_implicit_free_surface_solver.jl
```diff
@@ -43,8 +43,8 @@ step `Δt`, gravitational acceleration `g`, and free surface at time-step `n` `
 """
 function PCGImplicitFreeSurfaceSolver(grid::AbstractGrid, settings, gravitational_acceleration=nothing)
     # Initialize vertically integrated lateral face areas
-    ∫ᶻ_Axᶠᶜᶜ = Field{Face, Center, Nothing}(with_halo((3, 3, 1), grid))
-    ∫ᶻ_Ayᶜᶠᶜ = Field{Center, Face, Nothing}(with_halo((3, 3, 1), grid))
+    ∫ᶻ_Axᶠᶜᶜ = Field((Face, Center, Nothing), with_halo((3, 3, 1), grid))
+    ∫ᶻ_Ayᶜᶠᶜ = Field((Center, Face, Nothing), with_halo((3, 3, 1), grid))
```
glwagner, line 47: Why do we manually set the halos here? Maybe comment?

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
```diff
@@ -40,20 +40,17 @@ function update_state!(model::HydrostaticFreeSurfaceModel, grid, callbacks)
 end
 
 # Mask immersed fields
-function masking_actions!(model)
+function masking_actions!(model, grid)
```
glwagner, line 43:
```suggestion
function mask_immersed_model_fields!(model, grid)
```

src/MultiRegion/multi_region_abstract_operations.jl
```diff
@@ -1,12 +1,17 @@
...
-const MultiRegionAbstractOperation = Union{MultiRegionBinaryOperation, MultiRegionUnaryOperation, MultiRegionMultiaryOperation}
-
+const MultiRegionDerivative{LX, LY, LZ, D, A, IN, AD, G, T} = Derivative{LX, LY, LZ, D, A, IN, AD, <:MultiRegionGrid, T} where {LX, LY, LZ, D, A, IN, AD, T}
+const MultiRegionKernelFunctionOperation{LX, LY, LZ, P, G, T, K, D} = KernelFunctionOperation{LX, LY, LZ, P, <:MultiRegionGrid, T, K, D} where {LX, LY, LZ, P, T, K, D}
+
+const MultiRegionAbstractOperation = Union{MultiRegionBinaryOperation, 
+                                            MultiRegionUnaryOperation,
+                                         MultiRegionMultiaryOperation,
+                                                MultiRegionDerivative,
+                                   MultiRegionKernelFunctionOperation}
```
glwagner, line 14: Should we left justify ?

src/MultiRegion/multi_region_boundary_conditions.jl
```diff
@@ -59,17 +60,18 @@ fill_halo_regions!(c::MultiRegionObject, ::Nothing, args...; kwargs...) = nothin
...
 
-    halo_tuple = construct_regionally(permute_boundary_conditions, bcs)
+    neighbors = Reference(c.regional_objects)
+    buff      = Reference(buffers.regional_objects)
 
     for task = 1:3
         barrier = device_event(arch)
         apply_regionally!(fill_halo_event!, task, halo_tuple, 
-                          c, indices, loc, arch, barrier, mrg, Reference(c.regional_objects), Reference(buffers.regional_objects), 
+                          c, indices, loc, arch, barrier, mrg, neighbors, buff,
```
glwagner, line 74: Irrelevant to this PR but note that `Reference` is easily confused with `Base.Ref`

src/MultiRegion/multi_region_field.jl
```diff
@@ -110,6 +116,8 @@ end
 set!(mrf::MultiRegionField, v)  = apply_regionally!(set!,  mrf, v)
 fill!(mrf::MultiRegionField, v) = apply_regionally!(fill!, mrf, v)
 
+set!(mrf::MultiRegionField, f::Function)  = apply_regionally!(set!,  mrf, f)
```
glwagner, line 119:
```suggestion
set!(mrf::MultiRegionField, f::Function)  = apply_regionally!(set!, mrf, f)
```

src/MultiRegion/multi_region_split_explicit_free_surface.jl
```diff
@@ -0,0 +1,60 @@
+using Oceananigans.AbstractOperations: GridMetricOperation, Δz
+using Oceananigans.Models.HydrostaticFreeSurfaceModels: SplitExplicitState, SplitExplicitFreeSurface
+
+import Oceananigans.Models.HydrostaticFreeSurfaceModels: FreeSurface, SplitExplicitAuxiliary
+
+function SplitExplicitAuxiliary(grid::MultiRegionGrid)
```
glwagner, line 6: AuxiliaryData? AuxiliaryFields?

src/Oceananigans.jl
```diff
@@ -208,6 +207,7 @@ include("TurbulenceClosures/TurbulenceClosures.jl")
 include("Forcings/Forcings.jl")
 
 include("ImmersedBoundaries/ImmersedBoundaries.jl")
```
glwagner, line 209: Note that Coriolis introduces immersed boundary concepts so we are a little out of order here

src/Oceananigans.jl
```diff
@@ -208,6 +207,7 @@ include("TurbulenceClosures/TurbulenceClosures.jl")
 include("Forcings/Forcings.jl")
 
 include("ImmersedBoundaries/ImmersedBoundaries.jl")
+include("Distributed/Distributed.jl")
```
glwagner, line 210: This seems out of order

examples/shallow_water_Bickley_jet.jl
simone-silvestri: I thought there was a problem with the constructor but in the end, there wasn't. Why do we have this construction anyways?

src/Fields/constant_field.jl
simone-silvestri: It was not the actual constant, it was approximated

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
simone-silvestri: Aaaah, that is the purpose? It confused me a lot. I vote for keeping just one

src/Models/HydrostaticFreeSurfaceModels/distributed_split_explicit_free_surface.jl
simone-silvestri: no just that

src/Models/HydrostaticFreeSurfaceModels/distributed_split_explicit_free_surface.jl
simone-silvestri: why?

src/Oceananigans.jl
simone-silvestri: I would argue that they are grid concepts (`peripheral_node` exists also for a non immersed grid)

src/Forcings/advective_forcing.jl
simone-silvestri: ok

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
simone-silvestri: this might be helpful also as a safety net for non distributed architectures when the OffsetArray passed does not have correct halos

src/Models/HydrostaticFreeSurfaceModels/distributed_split_explicit_free_surface.jl
glwagner: Should we use a different name? `calculate_column_height` or something?

glwagner: @navidcy suggests:

```julia
calculate_column_height!(Hᶜᶠ, (Center, Face, Center), ...)
```

?

We are also confused: is `Hᶜᶠ` a field (so this uses `Hᶜᶠ.grid`?) Why do we also need to pass the location?

Finally I'll note that `Hᶜᶠ` is ambiguous (unless you understand the code in detail) because it is missing the third location. So someone glancing at the code cannot interpret this without the third letter to disambiguate (or reading the code in detail to understand). Here would be `a` for the vertical location.

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
glwagner: We need the second one because we need it to write things like `field1 = Field(location(field2), grid)`. 

The other constructor `field = Field{LX, LY, LZ}(grid)` is intended to be used when we aren't using `location()` to build the field.

glwagner: So just to summarize: the constructor `Field(loc, grid)` is a _convenience_ for interfacing with the function `location(f)`, which returns a tuple. However, the intent is not to use it generally because it is not literate: the tuple (LX, LY, LZ)` only appears transiently in the code, and is not encoded in the type anywhere.

The field location is embedded in the field type as `Field{LX, LY, LZ, ...}`. So writing `Field{LX, LY, LZ}` is "literate": new users reading the construct `f = Field{Center, Face, Center}(grid)` can infer that the type of that object is `f isa Field{Center, Face, Center}`.

We are also working towards a system in which types that are passed to functions are "instantiated". However, we can still pass types directly to type parameter specification (eg `{Center, Face, Center}`) uninstantiated. Some could be confused whether to use `f = Field((Center(), Face(), Center()), grid)` or the other...

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
glwagner: I think we want to _avoid_ that "silent safety" because `ImmersedBoundaryGrid(grid, ib::AbstractGridFittedBottom{<:OffsetArray})` is an internal function, not a user facing function. Such a safety net creates ambiguity about the user interface; users can make a mistake but not receive an error. We want to avoid that.

glwagner: In other words I think it is preferable here to receive an error if there is some mix-up about the grid halos... every experiment should use just one grid.

glwagner: Using `if architecture(grid) isa DistributedArch` will help preserve that error.

src/Fields/constant_field.jl
glwagner: I still think its better to use `prettysummary` for `summary`, which is consistent with julia Base (ie we use `Base.Ryu.writeshortest` which is also used by julia `Base` for summaries of floats).

glwagner: I propose we just keep using `prettysummary` everywhere, and then change the behavior of `prettysummary` if we want to do that. That way we don't have to make tiny changes all throughout the code.

glwagner: In a different PR. Let's change less in this PR.

src/Forcings/advective_forcing.jl
glwagner: Looks like we still aren't using `prettysummary`

docs/src/model_setup/forcing_functions.md
```diff
@@ -306,7 +306,7 @@ sinking = AdvectiveForcing(UpwindBiasedFifthOrder(), w=w_sediment)
 AdvectiveForcing with the UpwindBiased scheme:
 ├── u: ZeroField{Int64}
 ├── v: ZeroField{Int64}
-└── w: ConstantField(-0.00352102)
+└── w: ConstantField(-0.003521024784182679)
```
glwagner: Are we using `prettysummary`?

src/MultiRegion/multi_region_field.jl
```diff
@@ -59,8 +59,11 @@ Base.size(f::GriddedMultiRegionField) = size(getregion(f.grid, 1))
                       getregion(f.status, r),
                       getregion(f.boundary_buffers, r))
 
-@inline reconstruct_global_field(f::AbstractField) = f
-
+"""
+    reconstruct_global_field(mrf)
+reconstructs the global field from a `MultiRegionField`. The global field 
+is always reconstructed on the `CPU`
```
glwagner, line 65: Not complete sentence, not correctly formatted

src/MultiRegion/multi_region_grid.jl
```diff
@@ -29,7 +30,27 @@ end
 
 const ImmersedMultiRegionGrid = MultiRegionGrid{FT, TX, TY, TZ, P, <:MultiRegionObject{<:Tuple{Vararg{<:ImmersedBoundaryGrid}}}} where {FT, TX, TY, TZ, P}
 
-function MultiRegionGrid(global_grid; partition = XPartition(2), devices = nothing)
+"""
+    MultiRegionGrid(global_grid; partition = XPartition(2), devices = nothing)
+
+Splits a `global_grid` into different regions handled by `devices`
```
glwagner, line 36:
```suggestion
Split a `global_grid` into different regions handled by `devices`.
```

src/MultiRegion/multi_region_grid.jl
```diff
@@ -29,7 +30,27 @@ end
...
+
+Keyword Arguments
+=================
+
+- `partition`: the partitioning required. The implemented partitioning are `XPartition` 
+               (division along the x direction) and `YPartition` (division along the y direction)
+- `devices`: the devices to allocate memory on. `nothing` will allocate memory on the `CPU`, for 
+             `GPU` computation it is possible to specify the total number of `GPU`s or the specific
+             `GPU`s to allocate memory on. The number of devices does not have to match the number of
+             regions
```
glwagner, line 51: Missing periods?

src/TurbulenceClosures/turbulence_closure_implementations/isopycnal_skew_symmetric_diffusivity.jl
```diff
@@ -74,7 +74,7 @@ end
 function DiffusivityFields(grid, tracer_names, bcs, closure::FlavorOfISSD{TD}) where TD
     if TD() isa VerticallyImplicitTimeDiscretization
         # Precompute the _tapered_ 33 component of the isopycnal rotation tensor
-        return (; ϵ_R₃₃ = Field{Center, Center, Face}(grid))
+        return (; ϵ_R₃₃ = Field((Center, Center, Face), grid))
```
glwagner, line 77: Why?

src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
```diff
@@ -106,8 +106,8 @@ with_tracers(tracers, closure::FlavorOfRBVD) = closure
 
 # Note: computing diffusivities at cell centers for now.
 function DiffusivityFields(grid, tracer_names, bcs, closure::FlavorOfRBVD)
-    κ = Field{Center, Center, Face}(grid)
-    ν = Field{Center, Center, Face}(grid)
+    κ = Field((Center, Center, Face), grid)
+    ν = Field((Center, Center, Face), grid)
```
glwagner, line 110: Why?

src/Models/HydrostaticFreeSurfaceModels/distributed_split_explicit_free_surface.jl
simone-silvestri: This function integrates `z` into `H`. `Hᶜᶠ` and so on are defined in `split_explicit_free_surface.jl`. It is a field because it has a location. The location is passed because you want to integrate the correct `Δz` corresponding to the depth location. It is a 2D field (reduced in the vertical) so `a` might be misleading.

test/test_distributed_poisson_solvers.jl
```diff
@@ -67,7 +67,7 @@ end
 
 function divergence_free_poisson_solution_triply_periodic(grid_points, ranks)
     topo = (Periodic, Periodic, Periodic)
-    arch = MultiArch(CPU(), ranks=ranks, topology = topo)
+    arch = DistributedArch(CPU(), ranks=ranks, topology = topo)
```
glwagner, line 70:
```suggestion
    arch = DistributedArch(CPU(), ranks=ranks, topology=topo)
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -29,9 +30,15 @@ struct SplitExplicitFreeSurface{𝒩, 𝒮, ℱ, 𝒫 ,ℰ} <: AbstractFreeSurfa
 end
 
 # use as a trait for dispatch purposes
-SplitExplicitFreeSurface(; gravitational_acceleration = g_Earth, substeps = 200) =
+"""
+    SplitExplicitFreeSurface(; gravitational_acceleration = g_Earth, kwargs...) 
+
+Constructor for the `SplitExplicitFreeSurface`, for more information on the available `kwargs...`
+see `SplitExplicitSettings`
```
glwagner, line 37: Since this is the primary user interface we should probably put all the information about using `SplitExplicitFreeSurface` here.

src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
simone-silvestri: it is always the same motivation. We can revert everything if we agree on it, but it would be nice to have a once consistent way to define a field.

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +167,128 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
+    Possible barotropic time-stepping scheme. 
 
-    return SplitExplicitSettings(substeps,
-        velocity_weights,
-        free_surface_weights)
+- `AdamsBashforth3Scheme`: η = f(U, Uᵐ⁻¹, Uᵐ⁻²) then U = f(η, ηᵐ, ηᵐ⁻¹, ηᵐ⁻²)
+- `ForwardBackwardScheme`: η = f(U)             then U = f(η)
+"""
+struct AdamsBashforth3Scheme end
+struct ForwardBackwardScheme end
```
glwagner, line 205: Maybe should make these specific to split explicit so we don't confuse with "other" time stepping schemes -- helps when using `grep` to search the code base (or reviewing PRs)

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +167,128 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
+    τ₀ = (p + 2) * (p + q + 2) / (p + 1) / (p + q + 1) 
+    return (τ / τ₀)^p * (1 - (τ / τ₀)^q) - r * (τ / τ₀)
 end
 
-function SplitExplicitSettings(substeps)
-    velocity_weights = Tuple(ones(substeps) ./ substeps)
-    free_surface_weights = Tuple(ones(substeps) ./ substeps)
+@inline averaging_cosine_function(τ) = τ >= 0.5 && τ <= 1.5 ? 1 + cos(2π * (τ - 1)) : 0.0
+
+@inline averaging_fixed_function(τ) = 1.0
```
glwagner, line 215:
```suggestion
@inline cosine_averaging_kernel(τ::FT) where FT = τ >= 0.5 && τ <= 1.5 ? FT(1 + cos(2π * (τ - 1))) : zero(FT)

@inline constant_averaging_kernel(τ) = 1
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +167,128 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
+- `substeps`: The number of substeps that divide the range `(t, t + 2Δt)`. NOTE: not all averaging functions
+              require to substep till `2Δt`. The number of substeps will be reduced automatically to the last index
+              of `averaging_weights` for which `averaging_weights > 0`
+- `averaging_function`: function of `τ` used to average `U` and `η` within the barotropic advancement.
+                        `τ` is the fractional substep going from 0 to 2 with the baroclinic time step `t + Δt`
+                        located at `τ = 1`. This function should be centered at `τ = 1` (i.e. ∑(aₘ⋅m/M) = 1)
+- `timestepper`: Time stepping scheme used, either `ForwardBackwardScheme` or `AdamsBashforth3Scheme`
+"""
+function SplitExplicitSettings(; substeps = 200, 
+                                 averaging_function = averaging_shape_function,
```
glwagner, line 237:
```suggestion
                                 barotropic_averaging_kernel = averaging_shape_function,
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +167,128 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
+                                 timestepper = ForwardBackwardScheme())
+
+    τᶠ = range(0.0, 2.0, length = substeps+1)
```
glwagner, line 240:
```suggestion
    τᶠ = range(0, 2, length = substeps+1)
```

It seems this part of the code will not respect `Float32`. Probably better to build that in now than have to revisit in the future.

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +167,128 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
+
+Base.show(io::IO, sefs::SplitExplicitFreeSurface) = print(io, "$(summary(sefs))\n")
+
+function reset!(sefs::SplitExplicitFreeSurface)
+    for name in propertynames(sefs.state)
+        var = getproperty(sefs.state, name)
+        fill!(var, 0.0)
+    end
+    fill!(sefs.auxiliary.Gᵁ, 0.0)
+    fill!(sefs.auxiliary.Gⱽ, 0.0)
```
glwagner, line 289:
```suggestion
    fill!(sefs.auxiliary.Gᵁ, 0)
    fill!(sefs.auxiliary.Gⱽ, 0)
```

this will promote `0` to the type of `sefs.auxiliary.Gᵁ`. Since type promotion has almost no cost, this is efficient.

Using lower types like integer, which will be promoted to the given floating point type, is ok. But we want to avoid the opposite --- using _higher_ floating point types than the eltype of the `grid`.

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +167,128 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
-    return SplitExplicitSettings(substeps,
-        velocity_weights,
-        free_surface_weights)
+- `AdamsBashforth3Scheme`: η = f(U, Uᵐ⁻¹, Uᵐ⁻²) then U = f(η, ηᵐ, ηᵐ⁻¹, ηᵐ⁻²)
+- `ForwardBackwardScheme`: η = f(U)             then U = f(η)
+"""
+struct AdamsBashforth3Scheme end
+struct ForwardBackwardScheme end
+
+# Weights that minimize dispersion error from http://falk.ucsd.edu/roms_class/shchepetkin04.pdf (p = 2, q = 4, r = 0.18927)
```
glwagner, line 218:
```suggestion
# (p = 2, q = 4, r = 0.18927) minimize dispersion error from Shchepetkin and McWilliams (2005): https://doi.org/10.1016/j.ocemod.2004.08.002 
```

Note the above link is to the personal website of Falk Fedderson

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +167,128 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
+@inline function averaging_shape_function(τ; p = 2, q = 4, r = 0.18927)
```
glwagner, line 219:
```suggestion
@inline function optimal_barotropic_averging_kernel(τ; p = 2, q = 4, r = 0.18927) 
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +167,128 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
-    return SplitExplicitSettings(substeps = substeps,
-        velocity_weights = velocity_weights,
-        free_surface_weights = free_surface_weights)
+"""
+    SplitExplicitSettings(; substeps = 200, 
+                            averaging_function = averaging_shape_function,
+                            timestepper = ForwardBackwardScheme())
+
+constructor for `SpliExplicitSettings`. The keyword arguments can be provided
+directly to the `SplitExplicitFreeSurface` constructor
```
glwagner: fragment with no punctuation

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -29,9 +30,26 @@ struct SplitExplicitFreeSurface{𝒩, 𝒮, ℱ, 𝒫 ,ℰ} <: AbstractFreeSurfa
 end
 
 # use as a trait for dispatch purposes
-SplitExplicitFreeSurface(; gravitational_acceleration = g_Earth, substeps = 200) =
+"""
+    SplitExplicitFreeSurface(; gravitational_acceleration = g_Earth, kwargs...) 
+
+Constructor for the `SplitExplicitFreeSurface`, for more information on the available `kwargs...`
+see `SplitExplicitSettings`
```
glwagner, line 37:
```suggestion
Construct a `SplitExplicitFreeSurface`.
```

src/MultiRegion/multi_region_field.jl
```diff
@@ -59,8 +59,12 @@ Base.size(f::GriddedMultiRegionField) = size(getregion(f.grid, 1))
                       getregion(f.status, r),
                       getregion(f.boundary_buffers, r))
 
-@inline reconstruct_global_field(f::AbstractField) = f
+"""
+    reconstruct_global_field(mrf)
 
+this function reconstructs the global field from a `MultiRegionField`. The global field 
+is always reconstructed on the `CPU`
```
glwagner, line 66:
```suggestion
Reconstruct a global field from `mrf::MultiRegionField` on the `CPU`.
```

This might help: https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation

test/test_split_explicit_free_surface_solver.jl
```diff
@@ -13,7 +15,8 @@ import Oceananigans.Models.HydrostaticFreeSurfaceModels: SplitExplicitState, Spl
 
         grid = RectilinearGrid(arch, topology = topology, size = (Nx, Ny, Nz), x = (0, Lx), y = (0, Ly), z = (-Lz, 0), halo=(1, 1, 1))
 
-        sefs = SplitExplicitFreeSurface(grid)
+        settings = SplitExplicitSettings(; barotropic_averaging_kernel = averaging_fixed_function)
+        sefs = SplitExplicitFreeSurface(grid; settings)
 
         sefs.η .= 0.0
```
glwagner, line 21:
```suggestion
        sefs.η .= 0
```

test/test_split_explicit_free_surface_solver.jl
```diff
@@ -84,10 +89,10 @@ import Oceananigans.Models.HydrostaticFreeSurfaceModels: SplitExplicitState, Spl
             Gⱽ .= 0.0
```
glwagner, line 89:
```suggestion
            Gⱽ .= 0
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +178,117 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
+    Δτ = τᶠ[2] - τᶠ[1]
+
+    averaging_weights = barotropic_averaging_kernel.(τᶠ[2:end]) 
+    idx = searchsortedlast(averaging_weights, 0.0, rev=true)
+    substeps = idx
+
+    averaging_weights = averaging_weights[1:idx]
+    mass_flux_weights = similar(averaging_weights)
+    
+    M = searchsortedfirst(τᶠ, 1.0) - 1
```
glwagner, line 250:
```suggestion
    M = searchsortedfirst(τᶠ, 1) - 1
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -132,55 +178,117 @@ function SplitExplicitAuxiliary(grid::AbstractGrid)
...
+"""
+function SplitExplicitSettings(; substeps = 200, 
+                                 barotropic_averaging_kernel = averaging_shape_function,
+                                 timestepper = ForwardBackwardScheme())
+
+    τᶠ = range(0, 2, length = substeps+1)
+    Δτ = τᶠ[2] - τᶠ[1]
+
+    averaging_weights = barotropic_averaging_kernel.(τᶠ[2:end]) 
+    idx = searchsortedlast(averaging_weights, 0.0, rev=true)
```
glwagner, line 244:
```suggestion
    idx = searchsortedlast(averaging_weights, 0, rev=true)
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
glwagner: These changes are just to make it easier to diagnose type instability bugs in the future

glwagner (changes requested):
Amazing PR, mostly need to clean up names and documentation.

src/Models/HydrostaticFreeSurfaceModels/distributed_split_explicit_free_surface.jl
navidcy: Is there anywhere else in the code we have a reduced field with `cf`, etc subscripts?

src/Advection/vector_invariant_advection.jl
simone-silvestri: Sorry, i meant conservation of energy not momentum. `VerticallyConserving` conserves energy, while all the others do not because, to conserve energy, the discretization of `wdu` has to be consistent with the discretization of `dK`. I can change it to `VerticallyEnergyConserving`

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -27,18 +27,32 @@ for dir in (:west, :east, :south, :north, :bottom, :top)
...
+
+  extract_west_or_east_bc(bc::Tuple) =   extract_west_or_east_bc.(bc)
+extract_south_or_north_bc(bc::Tuple) = extract_south_or_north_bc.(bc)
+ extract_bottom_or_top_bc(bc::Tuple) =  extract_bottom_or_top_bc.(bc)
+ 
 # Finally, the true fill_halo!
 const MaybeTupledData = Union{OffsetArray, NTuple{<:Any, OffsetArray}}
 
-"Fill halo regions in ``x``, ``y``, and ``z`` for a given field's data."
+"Fill halo regions in ``x``, ``y``, and ``z``  for a given field's data."
```
navidcy, line 47:
```suggestion
"Fill halo regions in ``x``, ``y``, and ``z`` for a given field's data."
```

navidcy (approved):

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -27,7 +27,20 @@ for dir in (:west, :east, :south, :north, :bottom, :top)
...
+# `bc.east <: PBC`
+# `extract_west_or_east_bc(bc) == bc.west`
+    
+  extract_west_or_east_bc(bc) = sort([bc.west,   bc.east],  lt=fill_first)[2]
+extract_south_or_north_bc(bc) = sort([bc.south,  bc.north], lt=fill_first)[2]
+ extract_bottom_or_top_bc(bc) = sort([bc.bottom, bc.top],   lt=fill_first)[2]
+
+  extract_west_or_east_bc(bc::Tuple) =   extract_west_or_east_bc.(bc)
+extract_south_or_north_bc(bc::Tuple) = extract_south_or_north_bc.(bc)
+ extract_bottom_or_top_bc(bc::Tuple) =  extract_bottom_or_top_bc.(bc)
```
glwagner, line 42: Remember that often for small things it can be better to use `map` than broadcasting (which allocates more). But it depends on what you're doing.

glwagner (approved):

src/Advection/vector_invariant_advection.jl
navidcy: Yes do that!

src/BoundaryConditions/fill_halo_regions.jl
simone-silvestri: I' ve changed it using `max`

glwagner: good on ya

src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl
simone-silvestri: So I think we should leave the check but just spit an error, I moved the resizing in the Distributed module

simone-silvestri:
I should have addressed all the changes except the `Field{loc...}(grid)` to `Field(loc, grid)`. I will do another PR that will refactor all the instances of `Field(loc, grid)` to `Field{loc...}(grid)`

glwagner:
> I should have addressed all the changes except the `Field{loc...}(grid)` to `Field(loc, grid)`. I will do another PR that will refactor all the instances of `Field(loc, grid)` to `Field{loc...}(grid)`

sounds good

src/Distributed/distributed_grids.jl
```diff
@@ -271,15 +271,41 @@ function with_halo(new_halo, grid::DistributedImmersedBoundaryGrid)
...
+function resize_immersed_boundary(ib::GridFittedBoundary{<:OffsetArray}, grid)
+
+    Nx, Ny, Nz = size(grid)
+    Hx, Hy, Nz = halo_size(grid)
+
+    mask_size = (Nx, Ny, Nz) .+ 2 .* (Hx, Hy, Hz)
+
+    # Check that the size of a bottom field are 
+    # consistent with the size of the field
+    if any(size(ib.mask) .!= mask_size)
```
glwagner, line 301: why `any`?

glwagner: This needs a comment at the least

src/Distributed/distributed_grids.jl
```diff
@@ -271,15 +271,41 @@ function with_halo(new_halo, grid::DistributedImmersedBoundaryGrid)
...
+"""
+resize_immersed_boundary(ib::AbstractGridFittedBottom, grid) = ib
+resize_immersed_boundary(ib::GridFittedBoundary, grid)       = ib
+
+function resize_immersed_boundary(ib::GridFittedBoundary{<:OffsetArray}, grid)
+
+    Nx, Ny, Nz = size(grid)
+    Hx, Hy, Nz = halo_size(grid)
+
+    mask_size = (Nx, Ny, Nz) .+ 2 .* (Hx, Hy, Hz)
```
glwagner, line 297: Is this the same as `total_size(grid)`?

glwagner: I think it's important to use the functions here. Remember you want to change the size of a `Center, Bounded` field. When you do that, the above code is wrong.

glwagner: I think you want `total_size((Center, Center, Nothing), grid)` via

https://github.com/CliMA/Oceananigans.jl/blob/e394bf7ef7e4341ffddd449f52f58cee84698813/src/Grids/grid_utils.jl#L84

PR #2890, opened 2023-02-01
Change default output array type for `JLD2OutputWriter` to `Array{Float64}`

glwagner:
While `Array{Float32}` is neat because user output is smaller, it's a bad default because it often leads to confusing results, like the fact that outputing `Field(Integral(c))` is different (possibly by `eps(Float32)`) than computing `Integral(c_output)` in post processing when `Field(Integral(c))` is close to 0.

Thus @xkykai and I propose to make the default `Array{Float64}`. We could regard outputting with `Array{Float32}` as an "optimization" which is premature as a default.

### DIFF
--- a/test/test_netcdf_output_writer.jl
+++ b/test/test_netcdf_output_writer.jl
@@ -268,6 +268,7 @@ function test_thermal_bubble_netcdf_output_with_halos(arch)
     view(model.tracers.T, i1:i2, j1:j2, k1:k2) .+= 0.01
 
     nc_filepath = "test_dump_with_halos_$(typeof(arch)).nc"
+    
     nc_writer = NetCDFOutputWriter(model, merge(model.velocities, model.tracers),
                                    filename = nc_filepath,
                                    schedule = IterationInterval(10),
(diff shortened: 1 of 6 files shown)


glwagner:
cc @navidcy we also ran into this problem with ParameterEstimocean once

navidcy:
> cc @navidcy we also ran into this problem with ParameterEstimocean once

I remember… took days to figure out what was happening

navidcy (approved):

navidcy:
Only for JLD2OutputeWriter?

glwagner:
> Only for JLD2OutputeWriter?

Good point.

navidcy (approved):

test/test_netcdf_output_writer.jl
```diff
@@ -196,12 +196,12 @@ function test_thermal_bubble_netcdf_output(arch)
...
-    @test eltype(ds2["yC"]) == Float32
-    @test eltype(ds2["yF"]) == Float32
-    @test eltype(ds2["zC"]) == Float32
-    @test eltype(ds2["zF"]) == Float32
+    @test eltype(ds2["xC"]) == Float64
+    @test eltype(ds2["xF"]) == Float64
+    @test eltype(ds2["yC"]) == Float64
+    @test eltype(ds2["yF"]) == Float64
+    @test eltype(ds2["zC"]) == Float64
+    @test eltype(ds2["zF"]) == Float64
```
navidcy, line 204: `ds2` is `nc_sliced_writer` which was constructed with explicit `array_type = Array{Float32}`

navidcy, line 204:
```suggestion
    @test eltype(ds2["xC"]) == Float32
    @test eltype(ds2["xF"]) == Float32
    @test eltype(ds2["yC"]) == Float32
    @test eltype(ds2["yF"]) == Float32
    @test eltype(ds2["zC"]) == Float32
    @test eltype(ds2["zF"]) == Float32
```

navidcy:

navidcy:
merge?
we didn't bump any patch release, is this OK?

glwagner:
Let's bump

navidcy:
why 0.79.3? we are only up to 0.79.1... am I missing something?

glwagner:
> why 0.79.3? we are only up to 0.79.1... am I missing something?

you;re right

PR #2892, opened 2023-02-01
Update `pressure_correct_velocities` so that `NonhydrostaticModel.velocities` are divergence-free on `ImmersedBoundaryGrid`

### LINKED ISSUE #2891: Is the velocity field for `NonhydrostaticModel` divergence free when using `ImmersedBoundaryGrid`?
The pressure correction is done here:

https://github.com/CliMA/Oceananigans.jl/blob/c21d5d6ac95e762cac85154c8f3c4de63631fca7/src/Models/NonhydrostaticModels/pressure_correction.jl#L37-L39

which uses the "immersed boundary aware" derivatives `∂xᶠᶜᶜ`, `∂yᶜᶠᶜ`, and `∂zᶜᶜᶠ`. Specifically, these derivatives _always_ return zero across an immersed boundary.

(This can be seen from this file: https://github.com/CliMA/Oceananigans.jl/blob/main/src/ImmersedBoundaries/conditional_derivatives.jl, though it is wrapped up in metaprogramming and may be hard to figure out at first glance.)

However, the pressure correction for `NonhydrostaticModel` is done in such away that the pressure field does not have zero derivative normal to immersed boundaries.

The consequence is that the velocity field in `NonhydrostaticModel` likely has divergence.

With an "immersed boundary unaware" pressure correction step, we have to make a choice. Either 

1. The velocity field is divergence free, but does not satisfy impenetrability; or
2. The velocity field satisfies impenetrability, but is not divergence free.

We used to make choice 1, but I think we are currently making choice 2. (Note that both are "wrong" in that the correct dynamics are _both_ divergence free and satisfy impenetrability.)

glwagner:
Previously, the pressure correction was done in a way that satisfied impenetrability, but produced a velocity field with a divergent layer along the immersed boundary.

This PR changes the pressure correction so that the velocity field is divergence free everywhere, but does not satisfy impenetrability.

Closes #2891 

cc @xkykai @tomchor @simone-silvestri


glwagner:
@tomchor this may affect your results. Suprisingly, with this change the velocity field appears to _both_ divergence free _and_ satisfies impenetrability. Prior to this PR, the velocity was not divergence free. Thus I think this is important for correctness. Do you want to evaluate it?

tomchor:
> @tomchor this may affect your results. Suprisingly, with this change the velocity field appears to _both_ divergence free _and_ satisfies impenetrability. Prior to this PR, the velocity was not divergence free. Thus I think this is important for correctness. Do you want to evaluate it?

Wow that's interesting. I don't know if I have the time to evaluate this right now, so feel free to move forward without me. But if you observe that the velocity retains impenetrability and is now divergence-free then there are no downsides to this PR!

tomchor:
CC @whitleyv @wenegrat

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -33,10 +33,10 @@ Update the predictor velocities u, v, and w with the non-hydrostatic pressure vi
...
 @kernel function _pressure_correct_velocities!(U, grid, Δt, pNHS)
     i, j, k = @index(Global, NTuple)
-
-    @inbounds U.u[i, j, k] -= ∂xᶠᶜᶜ(i, j, k, grid, pNHS) * Δt
-    @inbounds U.v[i, j, k] -= ∂yᶜᶠᶜ(i, j, k, grid, pNHS) * Δt
-    @inbounds U.w[i, j, k] -= ∂zᶜᶜᶠ(i, j, k, grid, pNHS) * Δt
+    
+    @inbounds U.u[i, j, k] -= δxᶠᵃᵃ(i, j, k, grid, pNHS) / Δxᶠᶜᶜ(i, j, k, grid) * Δt
+    @inbounds U.v[i, j, k] -= δyᵃᶠᵃ(i, j, k, grid, pNHS) / Δyᶜᶠᶜ(i, j, k, grid) * Δt
+    @inbounds U.w[i, j, k] -= δzᵃᵃᶠ(i, j, k, grid, pNHS) / Δzᶜᶜᶠ(i, j, k, grid) * Δt
```
tomchor, line 39: So the idea is that while `∂x` operators dispatch on IBGrids (and are thus zero across immersed boundaries), `δx` operators do not, right?

If so, I can then see how this change leads to a divergence-free velocity field, but I don't see how impenetrability can be satisfied everywhere...

Do you understand this behavior?

tomchor:
It might be good to add divergence and tracer budget calculations to either a validation or a test, no? What do you think?

glwagner:
Let's close this. More to come from @xkykai.

PR #2897, opened 2023-02-03
Fix in lat-lon grid `Δy` metric

navidcy:
I guess it doesn't make any difference since both Δφ spacings between faces or centers are the same. But for the sake of sanity it should be changed. Thus, I removed the "bug" label.


navidcy:
should I bump a patch release?

navidcy:
I'll add some tests here

simone-silvestri:
It makes the difference if (as suggested by @christophernhill) we are to use a grid that is stretched at the poles to maintain horizontal isotropy! good catch

simone-silvestri (approved):

glwagner (approved):

navidcy:

PR #2898, opened 2023-02-04
Fixes writing of 3d-reduced `Field`s to NetCDF while pinning `GPUArrays`

tomchor:
Attempt and making tests on https://github.com/CliMA/Oceananigans.jl/pull/2865 pass


navidcy:

PR #2902, opened 2023-02-05
Update CUDA to v4

vchuravy:


vchuravy:
@maleadt 

CUBLAS error (for `cublasDnrm2_v2`) in https://buildkite.com/clima/oceananigans/builds/9877#018622f8-7cb1-4d4a-b58b-9e12aec6c441

navidcy:
we can use AMGX v0.1.4 now ;)

maleadt:
```
[e9e359dc] CUDA_jll v10.0.130+3
```

That JLL is deprecated and shouldn't be used. For one, it `dlopen`s plenty of libraries eagerly, as can be seen at the start of the test:

```
     Testing Running tests...
[NVBLAS] NVBLAS_CONFIG_FILE environment variable is NOT set : relying on default config filename 'nvblas.conf'
[NVBLAS] Cannot open default config file 'nvblas.conf'
[NVBLAS] Config parsed
[NVBLAS] CPU Blas library need to be provided
```

Opening many copies of CUDA libraries easily breaks the whole system, and might explain the CUBLAS failure here.

vchuravy:
> That JLL is deprecated and shouldn't be used.

Okay that means we need to rebuild AMGX_jll

navidcy:
is AMGX creating troubles for moving to CUDA 4?

Issue #2869, opened 2023-01-19
`device kernel image is invalid` error when running relatively complex simulation on GPUs


tomchor:
When running a relatively complex, but small simulation, I'm getting errors on the GPU that I can't make sense of. It's hard to come up with a minimum working example that's truly small because the error seems to pop up only when there's some level of complexity, but here's what I have so far:

```julia
using Oceananigans
arch = GPU()

z_faces = collect(0:1:6)
grid = RectilinearGrid(arch, size=(8, 8, 6),
                       x=(0, 1), y=(0, 1), z=z_faces)

@inline b_bg(x, y, z, t) = x
B_field = BackgroundField(b_bg)

@inline sponge_u(x, y, z, t, u, p) = -p.σ * (u - p.α*p.u₀)
@inline sponge_v(x, y, z, t, v, p) = -p.σ * (v - p.α*p.u₀)
@inline sponge_w(x, y, z, t, w, p) = -p.σ * (w - p.α*p.u₀)
@inline sponge_b(x, y, z, t, b, p) = -p.σ * (b - p.α*p.u₀)

Fᵤ = Forcing(sponge_u, field_dependencies = :u, parameters = (; σ=1, u₀=1, α=4e-5))
Fᵥ = Forcing(sponge_v, field_dependencies = :v, parameters = (; σ=1, u₀=1, α=4e-5))
Fw = Forcing(sponge_w, field_dependencies = :w, parameters = (; σ=1, u₀=1, α=4e-5))
Fb = Forcing(sponge_b, field_dependencies = :b, parameters = (; σ=1, u₀=1, α=4e-5))

model = NonhydrostaticModel(; grid,
                            advection = WENO(grid=grid, order=5),
                            tracers = (:b, :τ1, :τ2),
                            closure = SmagorinskyLilly(C=0.1),
                            background_fields = (b=B_field,),
                            forcing = (u=Fᵤ, v=Fᵥ, w=Fw, b=Fb),
                            )
@info model

simulation = Simulation(model, Δt=1, stop_iteration=10)

run!(simulation)
```

This (and way more complex examples) runs fine on the CPU but when I run that on the GPU I get:

```
[ Info: Executing initial time step...
ERROR: LoadError: CUDA error: device kernel image is invalid (code 200, ERROR_INVALID_IMAGE)

Stacktrace:
  [1] CUDA.CuModule(data::Vector{UInt8}, options::Dict{CUDA.CUjit_option_enum, Any})
[...]

PR #2795, opened 2022-11-02
`MultiRegion` adaptation of the `NonhydrostaticModel`

simone-silvestri:
At the moment, the pressure solve is performed on one GPU/CPU, but this PR builds the infrastructure to allow multi-process global solves

PR #2904, opened 2023-02-06
Removes binary operation from Smagorinsky diffusivity calculation

tomchor:
This PR changes the calculation of the diffusivities in the `SmagorinskyLilly` so that binary operations are avoided. At the moment it does so by creating a `calc_nonlinear_κᶜᶜᶜ()` kernel, similar to what is done for the AMD closure.

Closes https://github.com/CliMA/Oceananigans.jl/issues/2869

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl
@@ -106,6 +106,11 @@ filter width `Δᶠ`, and strain tensor dot product `Σ²`.
 end
 
 
+@inline function calc_nonlinear_κᶜᶜᶜ(i, j, k, grid, closure::SmagorinskyLilly, tracer, ::Val{tracer_index}, νₑ) where {tracer_index}
+    return νₑ[i,j,k] / closure.Pr[tracer_index]
+end
+
+
 function calculate_diffusivities!(diffusivity_fields, closure::SmagorinskyLilly, model)
 
     arch = model.architecture
@@ -114,12 +119,22 @@ function calculate_diffusivities!(diffusivity_fields, closure::SmagorinskyLilly,
     velocities = model.velocities
     tracers = model.tracers
 
-    event = launch!(arch, grid, :xyz,
-                    calculate_nonlinear_viscosity!,
-                    diffusivity_fields.νₑ, grid, closure, buoyancy, velocities, tracers,
-                    dependencies = device_event(arch))
+    workgroup, worksize = work_layout(grid, :xyz)
+    viscosity_kernel! = calculate_nonlinear_viscosity!(device(arch), workgroup, worksize)
+    diffusivity_kernel! = calculate_nonlinear_tracer_diffusivity!(device(arch), workgroup, worksize)
 
-    wait(device(arch), event)
+    barrier = device_event(arch)
+    viscosity_event = viscosity_kernel!(diffusivity_fields.νₑ, grid, closure, buoyancy, velocities, tracers, dependencies=barrier)
+
+    events = [viscosity_event]
+
+    for (tracer_index, κₑ) in enumerate(diffusivity_fields.κₑ)
+        @inbounds tracer = tracers[tracer_index]
+        event = diffusivity_kernel!(κₑ, grid, closure, tracer, Val(tracer_index), diffusivity_fields.νₑ, dependencies=barrier)
+        push!(events, event)
+    end
+
+    wait(device(arch), MultiEvent(Tuple(events)))
 
     return nothing
 end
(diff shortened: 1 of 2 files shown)


glwagner:
This PR should just a few new lines:

```julia
@inline κᶠᶜᶜ(i, j, k, grid, closure::SmagorinskyLilly, K, ::Val{id}, args...) = ℑxᶠᵃᵃ(i, j, k, grid, K.νₑ) / closure.Pr[id]
```

and the same for `κᶜᶠᶜ` abd `κᶜᶜᶠ`.

tomchor:
> There's no need for a new kernel because the diffusivities are `νₑ / Pr` so they can be calculated on the fly.
> 
> We need to extend `κᶠᶜᶜ` and `κᶜᶠᶜ` and `κᶜᶜᶠ`
> 
> ```julia
> @inline κᶠᶜᶜ(i, j, k, grid, closure::SmagorinskyLilly, K, ::Val{id}, args...) = ℑxᶠᵃᵃ(i, j, k, grid, K.νₑ) / closure.Pr[id]
> @inline κᶜᶠᶜ(i, j, k, grid, closure::SmagorinskyLilly, K, ::Val{id}, args...) = ℑyᵃᶠᵃ(i, j, k, grid, K.νₑ) / closure.Pr[id]
> @inline κᶜᶜᶠ(i, j, k, grid, closure::SmagorinskyLilly, K, ::Val{id}, args...) = ℑzᵃᵃᶠ(i, j, k, grid, K.νₑ) / closure.Pr[id]
> ```

I may be missing something, but when I make those changes I still get the errors in #2869. The changes here are the only ones I've tried that seem to solve that.

I've implemented and push your solution to the [tc/smag-binary-op2 branch](https://github.com/CliMA/Oceananigans.jl/blob/tc/smag-binary-op2/src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl):

https://github.com/CliMA/Oceananigans.jl/blob/56a76ae24d7199a3efe822d06969acc7b6ce6b16/src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl#L127-L129

I'm getting a `device kernel image is invalid (code 200, ERROR_INVALID_IMAGE)` error with only those changes.

glwagner:
On that branch you are still forming the `BinaryOperation` which is the problem. You also need to remove the `BinaryOperation`. If you open a PR we can work on it?

glwagner:
Change 

https://github.com/CliMA/Oceananigans.jl/blob/56a76ae24d7199a3efe822d06969acc7b6ce6b16/src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl#L212-L224

to

```julia
return (; νₑ)
```

tomchor:
> On that branch you are still forming the `BinaryOperation` which is the problem. You also need to remove the `BinaryOperation`. If you open a PR we can work on it?

Sure, I'll open a new PR soon then!

But just to be clear, I get that the problem is that I'm still passing the binary operations, but if I don't pass them then `model.diffusivity_fields` won't get populated, right? Or is your point that we don't have to populate `model.diffusivity_fields` for Smagorinsky?

glwagner:
The change I suggested will still put the eddy viscosity into diffusivity_fields, and removes the eddy diffusivities (calculating them on the fly from Pr and eddy viscosity)

glwagner:
You can also redefine the user-facing function diffusivity to return the original BinaryOperation (which hopefully will not enter into kernels because of the extension of those three “getter” methods)

tomchor:
> The change I suggested will still put the eddy viscosity into diffusivity_fields, and removes the eddy diffusivities (calculating them on the fly from Pr and eddy viscosity)

Yes! Sorry I wasn't clear enough. I got that. I meant populate it with the diffuvisities specifically, so we're on the same page.

tomchor:
Sorry for going slow on this, but if I understand correctly what you're proposing is:

- Remove the calculation of diffusivities from `DiffusivityFields()` (and thus remove tracer diffusivities from `model.diffusivity_fields`)
- Specialize `κᶠᶜᶜ()` functions for `SmagoriknskyLilly` so that diffusivities are calculated on the fly

As opposed to what I'm doing here which is just to change the calculation of diffusivities in `DiffusivityFields()` from using a `BinaryOperation` to a kernel (keeping them in `model.diffusivity_fields`).

If I understand correctly both methods do the same number of operations (one calculation of $\nu_e$, a division by `Pr` and one interpolation for each face of each grid cell), no?

So I guess the advantage of what you're proposing is that it saves memory (since diffusivities are calculated on the fly), at the cost of a bit more code complexity (i.e., one more specialization).

Conversely, the direction this PR is going atm uses more memory (for the diffusivities) but in my opinion the code is a bit clearer, since there's one fewer specialization (i.e. `κᶠᶜᶜ()` remains the same) and the code in `smagorinsky_lilly.jl` looks more like the code in `anisotropic_minimum_dissipation.jl`, which makes things more standardized.

I'll defer to you either way, but I vote that we take the approach that this PR is currently doing since, as we discussed before in a few PRs, the code in `TurbulenceClosures` is already a bit on the complex side and not super easy to understand. So I think the standardization of having `smagorinsky_lilly.jl` have the same structure as `anisotropic_minimum_dissipation.jl` I'd argue is a benefit.

If we follow with this PR, the increase in memory should be around 15% for one tracer and less for more tracers, so relatively small, plus the memory limitation is about to become less severe since hopefully https://github.com/CliMA/Oceananigans.jl/pull/2795 will be merged soon?

@glwagner like I said I'll defer to you either way. So if you prefer your original suggestion I'll close this PR and open another with your suggested modifications :)

glwagner:
I think extending `κᶠᶜᶜ` is simpler --- less code and less memory usage.

glwagner:
I also don't like `calc_nonlinear_κᶜᶜᶜ`. I hope we can come up with a better interface for defining LES closures in the future.

glwagner:
The memory savings is a major advantage of this closure over `AnisotropicMinimumDissipation` for problems with very large numbers of tracers (eg biogeochemistry problems with 10+ tracers)

tomchor:
> The memory savings is a major advantage of this closure over `AnisotropicMinimumDissipation` for problems with very large numbers of tracers (eg biogeochemistry problems with 10+ tracers).
> 
> Note also that there is overhead to launching a kernel which cannot be ignored --- we can't estimate computational cost just by adding the number of operations. Typically (though not always) our goal is to reduce the number of kernel launches as much as possible.

Good points. I also wasn't aware that the kernel launch time was an important issue. I'll close this PR and open another one extending `κᶠᶜᶜ()`.

PR #2909, opened 2023-02-07
Make `construct_regionally` handle multiple return values

glwagner:
This is important for writing robust and readable code in our time-stepping algorithm.


glwagner:
@simone-silvestri

simone-silvestri (approved):

PR #2910, opened 2023-02-07
Non-blocking fill_halo

simone-silvestri:
`fill_halo_event!` is now a blocking function, i.e. we fill halos one direction at a time waiting in between. With this PR we try to use dependencies to achieve the same goal. This is an initial step to make `fill_halo_regions!` non-blocking


.buildkite/pipeline.yml
```diff
@@ -498,16 +498,17 @@ steps:
...
+  # - label: "🦝 gpu shallow water regression tests"
+  #   env:
+  #     JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
+  #     TEST_GROUP: "shallowwater_regression"
+  #   commands:
+  #     - "$SVERDRUP_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
+  #   agents:
+  #     queue: Oceananigans
+  #     architecture: GPU
+  #   depends_on: "init_gpu"
```
glwagner, line 511: This PR will affect this regression test, correct?

simone-silvestri:
Stale after the removal of events in KA

PR #2913, opened 2023-02-08
Fixes `diffusivity()` for Smagorinsky-Lilly closure

tomchor:
A recent PR broke the user-facing `diffusivity()` function for `SmagorinskyLilly` and this fixes it. It also adds a test to catch this in the future and exports `diffusivity()` at the `TurbulenceClosure` level, since it's a user-facing function.

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl
@@ -16,7 +16,7 @@ struct SmagorinskyLilly{TD, FT, P} <: AbstractScalarDiffusivity{TD, ThreeDimensi
 end
 
 @inline viscosity(::SmagorinskyLilly, K) = K.νₑ
-@inline diffusivity(::SmagorinskyLilly, K, ::Val{id}) where id = K.νₑ / closure.Pr[id]
+@inline diffusivity(closure::SmagorinskyLilly, K, ::Val{id}) where id = K.νₑ / closure.Pr[id]
 
 """
     SmagorinskyLilly(time_discretization = ExplicitTimeDiscretization, [FT=Float64;] C=0.16, Pr=1)
--- a/test/test_turbulence_closures.jl
+++ b/test/test_turbulence_closures.jl
@@ -8,11 +8,6 @@ for closure in closures
     end
 end
 
-function closure_instantiation(closurename)
-    closure = getproperty(TurbulenceClosures, closurename)()
-    return true
-end
-
 function constant_isotropic_diffusivity_basic(T=Float64; ν=T(0.3), κ=T(0.7))
     closure = ScalarDiffusivity(T; κ=(T=κ, S=κ), ν=ν)
     return closure.ν == ν && closure.κ.T == κ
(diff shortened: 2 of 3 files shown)


test/test_turbulence_closures.jl
```diff
@@ -193,8 +188,14 @@ end
...
-        for closure in closures
-            @test closure_instantiation(closure)
+        for closurename in closures
+            closure = getproperty(TurbulenceClosures, closurename)()
+            @test closure isa TurbulenceClosures.AbstractTurbulenceClosure
+
+            grid = RectilinearGrid(CPU(), size=(1, 1, 1), extent=(1, 2, 3))
+            model = NonhydrostaticModel(grid=grid, closure=closure, tracers=:c)
+            @test diffusivity(model.closure, model.diffusivity_fields, Val(:c)) isa Union{Number, Field, AbstractOperations.AbstractOperation}
+            @test viscosity(model.closure, model.diffusivity_fields) isa Union{Number, Field}
```
glwagner, line 198: This could require changes if any closure wants to output a different type... any way we can test functionality rather than type?

glwagner: How about testing that the object it returns can be used within an `AbstractOperation`? like

```julia
op = diffusivity() * ∂x(c); op[1, 1, 1] isa Number
```

glwagner: or even just `op[1, 1, 1] == 0.0` (its a weak test, but better than nothing)

glwagner, line 198:
```suggestion
            c = model.tracers.c
            u = model.velocities.u
            κ = diffusivity(model.closure, model.diffusivity_fields, Val(:c)) 
            κ_dx_c = κ * ∂x(c)
            ν = viscosity(model.closure, model.diffusivity_fields)
            ν_dx_u = ν * ∂x(c)
            @test ν_dx_u[1, 1, 1] == 0.0
            @test κ_dx_c[1, 1, 1] == 0.0
```

A crucial problem with our tests is that they are hard to maintain, which slows down development. So I'm trying to help with that. First we have to make sure that when we add new tests, they are easy to maintain. We should also refactor as much as possible existing tests to make them easier to maintain.

src/TurbulenceClosures/turbulence_closure_implementations/smagorinsky_lilly.jl
```diff
@@ -16,7 +16,7 @@ struct SmagorinskyLilly{TD, FT, P} <: AbstractScalarDiffusivity{TD, ThreeDimensi
 end
 
 @inline viscosity(::SmagorinskyLilly, K) = K.νₑ
-@inline diffusivity(::SmagorinskyLilly, K, ::Val{id}) where id = K.νₑ / closure.Pr[id]
+@inline diffusivity(closure::SmagorinskyLilly, K, ::Val{id}) where id = K.νₑ / closure.Pr[id]
```
glwagner, line 19: This failure would be caught just by calling `diffusivity(model.closure, model.diffusivity_fields, Val(:c))` right?

tomchor: Right!

test/test_turbulence_closures.jl
tomchor: Good points. I guess your suggestion already addresses these concerns, right? I just accepted it. Let's see if tests pass

glwagner (approved):

PR #2918, opened 2023-02-12
Changes wet/dry to active/inactive

### LINKED ISSUE #2833: Are cells "wet", "dry", "active", or "inactive"?
This abstraction:

https://github.com/CliMA/Oceananigans.jl/blob/4b4bef76d82fb573778c3321b9b4a49503ed0d9c/src/Coriolis/hydrostatic_spherical_coriolis.jl#L58

introduces the synonyms "wet/dry" for the existing semantics "active/inactive".

I think we should pick one and use it.

@simone-silvestri

### LINKED ISSUE #2880: WetCell -> ActiveCell ?
This type uses the terms "wet" and "dry":

https://github.com/CliMA/Oceananigans.jl/blob/c8a65a4fdeff25722104ced3a0c74d2d921ae1cf/src/Coriolis/hydrostatic_spherical_coriolis.jl#L12

whereas I believe the rest of the code uses the terms "active" and "inactive".

@simone-silvestri should this be changed?

navidcy:
Closes #2880, closes #2833

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -498,16 +498,16 @@ steps:
 ##### Regression
 #####
 
-  - label: "🦝 gpu shallow water regression tests"
-    env:
-      JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
-      TEST_GROUP: "shallowwater_regression"
-    commands:
-      - "$SVERDRUP_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
-    agents:
-      queue: Oceananigans
-      architecture: GPU
-    depends_on: "init_gpu"
+  # - label: "🦝 gpu shallow water regression tests"
+  #   env:
+  #     JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
+  #     TEST_GROUP: "shallowwater_regression"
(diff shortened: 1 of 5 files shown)


glwagner (approved):

.buildkite/pipeline.yml
```diff
@@ -498,16 +498,16 @@ steps:
...
+  # - label: "🦝 gpu shallow water regression tests"
+  #   env:
+  #     JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
+  #     TEST_GROUP: "shallowwater_regression"
+  #   commands:
+  #     - "$SVERDRUP_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
+  #   agents:
+  #     queue: Oceananigans
+  #     architecture: GPU
+  #   depends_on: "init_gpu"
```
glwagner, line 510: Did this code affect anything to do with the regression test?

simone-silvestri: No it doesn't

navidcy: A test that fails probably means something!

simone-silvestri: If you look at PR #2924 the tests pass there, so it is strong evidence that it is a synchronization issue. We should readd it in lowering the number of time steps although that will only mask the problem in my opinion. Since a lot of PRs have been merged since, I suggest merging #2888 and then halting merging until we regenerate the data

PR #2919, opened 2023-02-12
Update deps

navidcy:


tomchor (approved):

simone-silvestri:
the tests were not passing here

tomchor:
> the tests were not passing here

Yeah, I should have been more clear that I approved to merge after tests were passing, I didn't know we could even merge without tests passing. I thought that behavior was blocked

PR #2926, opened 2023-02-13
Avoid type instability in HydrostaticSphericalCoriolis

### LINKED ISSUE #2925: Type instability in `hydrostatic_spherical_coriolis.jl`?
Should this

https://github.com/CliMA/Oceananigans.jl/blob/e2841ca6dda4414fcd9177512faec557d6f12adc/src/Coriolis/hydrostatic_spherical_coriolis.jl#L60

be 

```Julia
1 - f(i, j, k, grid, args...)
```
to avoid type instability?

navidcy:
Closes #2925

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -498,16 +498,16 @@ steps:
 ##### Regression
 #####
 
-  - label: "🦝 gpu shallow water regression tests"
-    env:
-      JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
-      TEST_GROUP: "shallowwater_regression"
-    commands:
-      - "$SVERDRUP_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
-    agents:
-      queue: Oceananigans
-      architecture: GPU
-    depends_on: "init_gpu"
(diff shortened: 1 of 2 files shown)


glwagner (approved):
What does this function do?

simone-silvestri:
yeah you can change it to `1.0 -  f(i, j, k, grid, args...)` to `FT(!f(i, j, k, grid, args...))` I guess. 
I am not sure using `!peripheral_node` will work because it is a statement, not a function, and we need a function to compose in 
```
ℑxyᶠᶜᵃ(i, j, k, grid, revert_peripheral_node, peripheral_node, Center(), Face(), Center())
```

glwagner:
> yeah you can change it to `1.0 - f(i, j, k, grid, args...)` to `FT(!f(i, j, k, grid, args...))` I guess. I am not sure using `!peripheral_node` will work because it is a statement, not a function, and we need a function to compose in
> 
> https://github.com/CliMA/Oceananigans.jl/blob/e2841ca6dda4414fcd9177512faec557d6f12adc/src/Coriolis/hydrostatic_spherical_coriolis.jl#L69

I think @navidcy is suggesting to define a function

```julia
not_peripheral_node(args...) = !peripheral_node(args...)
```

which could then be used and doesn't need a docstring.

navidcy:
Yeah. Wouldn't that work?

simone-silvestri:
ok sure that would work!

navidcy:
You mean in the regression test or in general?

Issue #2922, opened 2023-02-13
Shallow water regression tests often fail on GPU


tomchor:
This isn't a huge issue but I've noticed recently that often the shallow water regression tests on GPU fail with the following error:

```
Shallow Water Bickley jet simulation [GPU, ConservativeFormulation]: Test Failed at /net/ocean/home/data44/data5/glwagner/.buildkite-agent/builds/sverdrup-6/clima/oceananigans/test/regression_tests/shallow_water_bickley_jet_regression.jl:93
  | Expression: all(test_fields.v .≈ truth_fields.v)
  | Stacktrace:
  | [1] macro expansion
  | @ /net/ocean/home/data44/data5/glwagner/julia-1.8.5/share/julia/stdlib/v1.8/Test/src/Test.jl:464 [inlined]
  | [2] run_shallow_water_regression(arch::GPU, formulation::ConservativeFormulation; regenerate_data::Bool)
  | @ Main /net/ocean/home/data44/data5/glwagner/.buildkite-agent/builds/sverdrup-6/clima/oceananigans/test/regression_tests/shallow_water_bickley_jet_regression.jl:93
```

Sometimes I need to restart the test twice for it to pass. Like I said, it's not a huge issue, but do we understand why that's happening? Is there anything we can do to prevent that behavior?

vchuravy:
@simone-silvestri and I think there is a missing synchronization for one of the boundary updates

simone-silvestri:
I am also doubting our regression data. For sure it is a synchronization problem, but it is difficult to understand wether it's on the data (generated long time ago) or in the current code

simone-silvestri:
You are also right, the data was generated on the CPU so it should be synchronized correctly.

simone-silvestri:
It might be also another problem not related to BC. 
When the test passes we get
```
[2023/02/12 17:00:57.335] INFO  Δv: min=-2.709402e-11, max=+2.834735e-11, mean=-2.606411e-14, absmean=+1.007853e-12, std=+2.676095e-12 (16512/16512 matching grid points)
```
On test failing
```
[2023/02/14 10:02:14.671] INFO  Δv: min=-3.212770e-11, max=+2.834735e-11, mean=-2.751160e-13, absmean=+1.256905e-12, std=+3.884381e-12 (16384/16512 matching grid points)
```
(same maximum difference larger mean). I will check which points are actually not matching, it might help us

simone-silvestri:
[...]

PR #2927, opened 2023-02-14
Removes GPU Regression test

simone-silvestri:
This PR temporarily removes GPU shallow water test (see issue #2922)


navidcy (approved):

glwagner:
Since it's a regression test, we will lose continuity of the code if any PRs are merged. So I suggest regenerating the data ASAP and merging the regression test back in.

PR #2933, opened 2023-02-21
Don't conflate the status of a sliced field with its parent field

### LINKED ISSUE #2931: Different results from identical tracers when using NetCDFOutputWriter
The example below creates a nonhydrostatic model with 6 identical tracers, calculates the vertical flux of those tracers in an identical way for all 6, and then writes it all to NetCDF. In the same file, it writes a vertical (x-z) slice of the fluxes using the `indices` flag, and an y-average of them:

```julia
using Oceananigans
using Oceananigans.Units

grid = RectilinearGrid(size=(16, 16, 16), extent = (500, 500, 120))

n_tracers = 6
tracer_symbols = [ Symbol(:τ,i) for i in 1:n_tracers ]
model = NonhydrostaticModel(; grid, tracers = (tracer_symbols...,))
@info model

uᵢ(x, y, z) = 1e-2 * randn()
set!(model, w=uᵢ)

tracer_IC_odd(x, y, z) = sin(2π*z/grid.Lz)
for i in 1:n_tracers
    @info "Setting tracer $i"
    expression = Meta.parse("set!(model, τ$i=tracer_IC_odd)")
    eval(expression)
end

simulation = Simulation(model, Δt=30, stop_time=0.15hours)

u, v, w = model.velocities

wτ = NamedTuple(Symbol(:w, key) => Field(w*τ) for (key,τ) in pairs(model.tracers))

outputs_full = (; wτ...)

outputs_yavg = NamedTuple( Symbol(key, :_yavg)=>Average(val, dims=(2,)) for (key, val) in zip(keys(outputs_full), outputs_full) )

outputs_xz1 = merge(outputs_full, outputs_yavg)
simulation.output_writers[:xz1_writer] = NetCDFOutputWriter(model, outputs_xz1;
                       filename = "data/TEST.nc",
                       schedule = TimeInterval(simulation.stop_time),
                       verbose=true,
                       indices = (:, 1, :),
                       overwrite_existing = true,
                       )
run!(simulation)
```

While the outputs should be the same (since the tracers and their advection are identical), I get different results for the y-averaged fluxes for different tracers:

image

Note that, while similar, tracers α=1,3,4 are different from α=2,5,6. The difference isn't large in this example, but can be made larger with more complexity in the calculations.

A couple of notes:

- Changing the line `wτ = NamedTuple(Symbol(:w, key) => Field(w*τ) for (key,τ) in pairs(model.tracers))` to 
```julia
wτ = NamedTuple(Symbol(:w, key) => w*τ for (key,τ) in pairs(model.tracers))
```
gets rid of the issue. Although doing the above prevents a user from using `Field(op, data=scratch_data.data)` to save memory, which in some cases (my case for example) is very important.

- Not passing the `indices` flag (i.e. writing 3D fields instead slices) apparently gets rid of this issue.

Here, even though the obvious easy solution is to separate averages from slices when writing, a user wouldn't know that since this fails silently and the wrong results can be pretty subtle (as the example above hopefully illustrates). For example, it popped up in one of my simulations and it took me a while to even realize what was happening, let alone figure out the solution.

My main question is: is this expected behavior? If so, should we somehow warn users (or even throw an error) do avoid mistakes?

glwagner:
Closes #2931 

We could add a test...

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.79.5"
+version = "0.79.4"
 
 [deps]
 AMGX = "c963dde9-0319-47f5-bf0c-b07d3c80ffa6"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.79.4"
+version = "0.79.5"
 
 [deps]
 AMGX = "c963dde9-0319-47f5-bf0c-b07d3c80ffa6"
--- a/src/AbstractOperations/computed_field.jl
+++ b/src/AbstractOperations/computed_field.jl
@@ -64,17 +64,20 @@ Compute `comp.operand` and store the result in `comp.data`.
 function compute!(comp::ComputedField, time=nothing)
     # First compute `dependencies`:
     @apply_regionally compute_at!(comp.operand, time)
-    @apply_regionally compute_field!(comp)
+
+    # Now perform the primary computation
+    @apply_regionally compute_computed_field!(comp)
 
     fill_halo_regions!(comp)
 
     return comp
 end
 
-function compute_field!(comp)
+function compute_computed_field!(comp)
     arch = architecture(comp)
     event = launch!(arch, comp.grid, size(comp), _compute!, comp.data, comp.operand, comp.indices)
     wait(device(arch), event)
+    return comp
 end
 
 """Compute an `operand` and store in `data`."""
--- a/src/Fields/field.jl
+++ b/src/Fields/field.jl
@@ -312,13 +312,18 @@ function Base.view(f::Field, i, j, k)
 
     boundary_conditions = FieldBoundaryConditions(window_indices, f.boundary_conditions)
 
+    # "Sliced" Fields created here share data with their parent.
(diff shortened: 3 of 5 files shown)


tomchor (approved):
Tested and indeed this PR fixes the issue

tomchor:
> Closes #2931
> 
> We could add a test...

A test would be good, but like you said it yourself the tests are already pretty resource-intensive so feel free to merge without a test if you think it wouldn't be worth it.

navidcy:
@tomchor can you confirm that this PR doesn't show that issue?

tomchor:
> @tomchor can you confirm that this PR doesn't show that issue?

Yes. I tested this with a couple different number of tracers and all was good

navidcy:
OK, I'll try to resolve #2922 before we merge this.

@glwagner bump patch release?

navidcy:
That commit was REALLY quick!

navidcy (approved):

navidcy:
I suggest we wait for https://github.com/CliMA/Oceananigans.jl/issues/2922. I'm on it.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.79.4"
+version = "0.79.5"
```
navidcy, line 4:
```suggestion
version = "0.79.4"
```

navidcy:
Since we didn't release v0.79.4 yet, I suggest we wait for this to merge + some other bug fixes in PRs that are ready to be merged after tests pass and release after.

cc @tomchor, @glwagner, @simone-silvestri

tomchor:
@glwagner @navidcy merge and register a new version?

PR #2934, opened 2023-02-21
Small changes to tilted bottom boundary layer example

tomchor:
This PR changes ~two things~ about the BBL example:

- The variable `κ` was used for both the von Karman constant and for the eddy diffusivity. Now it's just used for the von Karman constant
- ~When calling `NonhydrostaticModel` we were using ordered arguments, which could make a new user interpret that they should know the order of the arguments when calling `NonhydrostaticModel`. Now this call is made in a more usual way with keyword arguments, where the order doesn't matter.~

### DIFF
--- a/examples/tilted_bottom_boundary_layer.jl
+++ b/examples/tilted_bottom_boundary_layer.jl
@@ -128,10 +128,12 @@ v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
 # and a constant viscosity and diffusivity.
 
 ν = 1e-4 # m² s⁻¹, small-ish
-κ = ν
-closure = ScalarDiffusivity(; ν, κ)
+closure = ScalarDiffusivity(ν=ν, κ=ν)
 
-model = NonhydrostaticModel(; grid, buoyancy, coriolis, closure,
+model = NonhydrostaticModel(; grid,
+                            buoyancy = buoyancy,
+                            closure = closure,
+                            coriolis = coriolis,
                             timestepper = :RungeKutta3,
                             advection = UpwindBiasedFifthOrder(),
                             tracers = :b,
--- a/examples/tilted_bottom_boundary_layer.jl
+++ b/examples/tilted_bottom_boundary_layer.jl
@@ -127,7 +127,6 @@ v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
 # `UpwindBiasedFifthOrder` advection scheme, a `RungeKutta3` timestepper,
 # and a constant viscosity and diffusivity. Here we use a smallish value of ``10^{-4} m² s⁻¹``.
 
-ν = 1e-4 # m² s⁻¹, small-ish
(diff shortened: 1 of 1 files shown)


navidcy:
I don't understand what you mean in the second point with "we were using ordered arguments"?

Syntax
```Julia
f(; x, y, z)
```
is the same as
```Julia
f(x=x, y=y, z=z)
```
and the order does not matter, e.g.
```Julia
julia> x=1; y=2; z=3;

julia> function f(; x=0, y=0, z=0)
           @show x
           @show y
           @show z
           return x+y+z
           end

julia> f(; x, y, z)
x = 1
y = 2
z = 3
6

julia> f(; y, z, x)
x = 1
y = 2
z = 3
6

```

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -128,10 +128,12 @@ v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
 # and a constant viscosity and diffusivity.
```
navidcy, line 128:
```suggestion
# and a constant viscosity and diffusivity. Here we use a smallish value of ``10^{-4} m² s⁻¹``.
```

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -128,10 +128,12 @@ v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
...
 
 ν = 1e-4 # m² s⁻¹, small-ish
```
navidcy, line 130:
```suggestion
```

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -128,10 +128,12 @@ v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
...
-κ = ν
-closure = ScalarDiffusivity(; ν, κ)
+closure = ScalarDiffusivity(ν=ν, κ=ν)
```
navidcy, line 131:
```suggestion
closure = ScalarDiffusivity(ν=1e-4, κ=1e-4)
```

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -128,10 +128,12 @@ v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
...
 
-model = NonhydrostaticModel(; grid, buoyancy, coriolis, closure,
+model = NonhydrostaticModel(; grid,
+                            buoyancy = buoyancy,
+                            closure = closure,
+                            coriolis = coriolis,
```
navidcy, line 136: I suggest we revert this...
I don't see what's the issue.
Anything after `;` is a `kwarg` and the order is not important.

tomchor: Yes, sorry I misread what was there and thought the ordered arguments version was being used. Thanks for catching that

navidcy (approved):
let's wait for #2922

PR #2938, opened 2023-02-22
Make it easier to restart a simulation from a checkpoint with additional passive tracers

tomchor:
This feature adds support for using a checkpoint file to initialize a model that contains additional passive tracers that weren't present in the original simulation. The use case in mind is whenever a user wants to start passive tracers only after simulation spin-up, for example.

At the moment, on main, this isn't possible since if we try to pickup a simulation but a given variable can't be found in the checkpointer file, the code throws a warning when trying to set the data for that variable, and an error when trying to set its tendencies.

This PR changes the code so that it throws an error for both cases. After this PR a user can then do:

```julia
using Oceananigans

grid = RectilinearGrid(size = (4, 4, 4), extent = (1,1,1))

model_spinup = NonhydrostaticModel(; grid, tracers = :b)
set!(model_spinup, b=1)

simulation = Simulation(model_spinup, Δt = 1, stop_time = 10)
simulation.output_writers[:checkpointer] = checkpointer = Checkpointer(model_spinup,
                                                                       schedule=TimeInterval(5),
                                                                       prefix="checkpoint")

run!(simulation)

using Oceananigans.OutputWriters: write_output!
write_output!(checkpointer, model_spinup)

                            tracers = (keys(model_spinup.tracers)..., :t1, :t2))

@info "Restarting model with more tracers"
checkpoint_file_path = Oceananigans.OutputWriters.checkpoint_path(true, simulation.output_writers)
set!(model, checkpoint_file_path)

simulation = Simulation(model, Δt = 1, stop_time = 20)

run!(simulation)
```

On main this throws a `KeyError`. On this branch this produces:

```julia
[ Info:     ... simulation initialization complete (1.614 seconds)
[ Info:     ... initial time step complete (20.304 seconds).
[ Info: Simulation is stopping after running for 22.039 seconds.
[ Info: Simulation time 10 seconds equals or exceeds stop time 10 seconds.
[ Info: Restarting model with more tracers
┌ Warning: Could not restore t1 from checkpoint.
└ @ Oceananigans.OutputWriters ~/repos/Oceananigans.jl/src/OutputWriters/checkpointer.jl:218
┌ Warning: Could not restore t2 from checkpoint.
└ @ Oceananigans.OutputWriters ~/repos/Oceananigans.jl/src/OutputWriters/checkpointer.jl:218
┌ Warning: Could not restore tendencies for t1 from checkpoint.
└ @ Oceananigans.OutputWriters ~/repos/Oceananigans.jl/src/OutputWriters/checkpointer.jl:257
┌ Warning: Could not restore tendencies for t2 from checkpoint.
└ @ Oceananigans.OutputWriters ~/repos/Oceananigans.jl/src/OutputWriters/checkpointer.jl:257
[ Info:     ... simulation initialization complete (514.400 μs)
[ Info:     ... initial time step complete (41.614 seconds).
[ Info: Simulation is stopping after running for 41.750 seconds.
[ Info: Simulation time 20 seconds equals or exceeds stop time 20 seconds.

julia> interior(model.tracers.b)
4×4×4 view(::Array{Float64, 3}, 4:7, 4:7, 4:7) with eltype Float64:
[:, :, 1] =
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0

[:, :, 2] =
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0

[:, :, 3] =
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0

[:, :, 4] =
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0

julia> interior(model.tracers.t1)
4×4×4 view(::Array{Float64, 3}, 4:7, 4:7, 4:7) with eltype Float64:
[:, :, 1] =
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0

[:, :, 2] =
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0

[:, :, 3] =
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0

[:, :, 4] =
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0
```

One option to add safety is to modify `set!(model, filepath::AbstractString)` to take an option argument `skip_missing_variables = false` (or something), which would change the behavior from throwing a warning to throwing an error.

@whitleyv

### DIFF
--- a/src/OutputWriters/checkpointer.jl
+++ b/src/OutputWriters/checkpointer.jl
@@ -209,16 +209,12 @@ function set!(model, filepath::AbstractString)
         model_fields = prognostic_fields(model)
 
         for name in propertynames(model_fields)
-            try
+            if string(name) ∈ keys(file) # Test if variable exist in checkpoint
                 parent_data = file["$name/data"]
                 model_field = model_fields[name]
                 copyto!(model_field.data.parent, parent_data)
-            catch err
-                if err isa KeyError
-                    @warn "Could not restore $name from checkpoint."
-                else
-                    rethrow(err)
-                end
+            else
+                @warn "Field $name does not exist in checkpoint and could not be restored."
             end
         end
 
@@ -240,7 +236,7 @@ end
 
 function set_time_stepper_tendencies!(timestepper, file, model_fields)
     for name in propertynames(model_fields)
-        try
+        if string(name) ∈ keys(file["timestepper/Gⁿ"]) # Test if variable tendencies exist in checkpoint
             # Tendency "n"
             parent_data = file["timestepper/Gⁿ/$name/data"]
 
@@ -252,12 +248,8 @@ function set_time_stepper_tendencies!(timestepper, file, model_fields)
 
             tendency⁻_field = timestepper.G⁻[name]
             copyto!(tendency⁻_field.data.parent, parent_data)
-        catch err
-            if err isa KeyError
-                @warn "Could not restore tendencies for $name from checkpoint."
-            else
-                rethrow(err)
-            end
+        else
+            @warn "Tendencies for $name do not exist in checkpoint and could not be restored."
         end
     end
 
--- a/src/OutputWriters/checkpointer.jl
+++ b/src/OutputWriters/checkpointer.jl
@@ -213,8 +213,12 @@ function set!(model, filepath::AbstractString)
                 parent_data = file["$name/data"]
                 model_field = model_fields[name]
                 copyto!(model_field.data.parent, parent_data)
-            catch
-                @warn "Could not retore $name from checkpoint."
+            catch err
+                if err isa KeyError
+                    @warn "Could not restore $name from checkpoint."
+                else
+                    rethrow(err)
+                end
             end
         end
 
@@ -236,17 +240,25 @@ end
 
 function set_time_stepper_tendencies!(timestepper, file, model_fields)
     for name in propertynames(model_fields)
-        # Tendency "n"
-        parent_data = file["timestepper/Gⁿ/$name/data"]
+        try
+            # Tendency "n"
+            parent_data = file["timestepper/Gⁿ/$name/data"]
 
-        tendencyⁿ_field = timestepper.Gⁿ[name]
-        copyto!(tendencyⁿ_field.data.parent, parent_data)
+            tendencyⁿ_field = timestepper.Gⁿ[name]
+            copyto!(tendencyⁿ_field.data.parent, parent_data)
 
-        # Tendency "n-1"
-        parent_data = file["timestepper/G⁻/$name/data"]
+            # Tendency "n-1"
+            parent_data = file["timestepper/G⁻/$name/data"]
 
-        tendency⁻_field = timestepper.G⁻[name]
-        copyto!(tendency⁻_field.data.parent, parent_data)
+            tendency⁻_field = timestepper.G⁻[name]
+            copyto!(tendency⁻_field.data.parent, parent_data)
+        catch err
+            if err isa KeyError
+                @warn "Could not restore tendencies for $name from checkpoint."
+            else
+                rethrow(err)
+            end
+        end
     end
 
     return nothing


glwagner:
This feature would be used to restart a simulation from a checkpoint with additional passive tracers? Is that correct?

tomchor:
> This feature would be used to restart a simulation from a checkpoint with additional passive tracers? Is that correct?

That is correct. That's what I'm currently trying to do (and eventually @whitleyv too) and it's much easier if with we make this change with the Checkpointer.

glwagner:
Can we update the PR description to state this goal?

tomchor:
> Can we update the PR description to state this goal? "This feature adds support for using a checkpoint file to initialize a model that contains additional passive tracers..."

Done!

src/OutputWriters/checkpointer.jl
```diff
@@ -213,8 +213,12 @@ function set!(model, filepath::AbstractString)
                 parent_data = file["$name/data"]
                 model_field = model_fields[name]
                 copyto!(model_field.data.parent, parent_data)
-            catch
-                @warn "Could not retore $name from checkpoint."
+            catch err
+                if err isa KeyError
```
glwagner, line 217: Why KeyError? Maybe add a comment?

src/OutputWriters/checkpointer.jl
```diff
@@ -236,17 +240,25 @@ end
...
+            tendencyⁿ_field = timestepper.Gⁿ[name]
+            copyto!(tendencyⁿ_field.data.parent, parent_data)
+
+            # Tendency "n-1"
+            parent_data = file["timestepper/G⁻/$name/data"]
+
+            tendency⁻_field = timestepper.G⁻[name]
+            copyto!(tendency⁻_field.data.parent, parent_data)
+        catch err
+            if err isa KeyError
```
glwagner, line 256: Why KeyError?

src/OutputWriters/checkpointer.jl
tomchor: That's the expected error when a variable isn't present in the checkpoint. It's possible to have other errors though, so I think should `rethrow()` the error if something else goes wrong

tomchor: In any case your comment made me realize the way it was written probably isn't best practice, so I re-wrote the code using an if-else statement instead of a try-catch. The new code is shorter and (I think) clearer. What do you think?

src/OutputWriters/checkpointer.jl
```diff
@@ -209,16 +209,12 @@ function set!(model, filepath::AbstractString)
         model_fields = prognostic_fields(model)
 
         for name in propertynames(model_fields)
-            try
+            if string(name) ∈ keys(file) # Test if variable exist in checkpoint
```
glwagner, line 212: I like this! Easier to understand when reading it

glwagner (approved):

src/OutputWriters/checkpointer.jl
glwagner: I think it's easier to understand.

There's the concept of "Easier to Ask for Forgiveness than Permission" which claims that try/catch is better style esp for dynamic languages.

I think our project and code is unique. So while there are places that so-called "best practices" apply, there are also places where we should do what works best for us.

glwagner: The downside of "look before you leap" is when we have many error conditions which causes the if-statements to pile up. Here we have just one condition so I think explicit writing the condition is better and more readable.

PR #2944, opened 2023-02-24
`\textit{...}` -> `*...*` for correct rendering in `boundary_conditions.md`

jbisits:


glwagner (approved):
Thank you!

tomchor:
It seems like we're still getting the `all(test_fields.v .≈ truth_fields.v)` error in the [shallow water regression tests](https://buildkite.com/clima/oceananigans/builds/10225#01868315-b904-4152-b885-9b64cbab7111).

I'll restart it and see what happens. But we should probably keep an eye on that.

navidcy:
I restarted it an error was insisting... I reopened #2922. I'll make the timestep even smaller.

PR #2951, opened 2023-02-28
Return zero for `closure == nothing`

tomchor:
This is useful for writing general code from the user perspective. Ideally we'd want to return `zero(FT)`, but there's nowhere obvious to get the float type, so I think this is okay for now.

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_implementations/nothing_closure.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/nothing_closure.jl
@@ -5,3 +5,6 @@
 
 calculate_diffusivities!(diffusivities, ::Nothing, args...) = nothing
 calculate_diffusivities!(::Nothing, ::Nothing, args...) = nothing
+
+@inline viscosity(::Nothing, ::Nothing) = 0.0
+@inline diffusivity(::Nothing, ::Nothing, ::Val{id}) where id = 0.0
--- a/src/Oceananigans.jl
+++ b/src/Oceananigans.jl
@@ -71,6 +71,7 @@ export
     IsopycnalSkewSymmetricDiffusivity,
     FluxTapering,
     VerticallyImplicitTimeDiscretization,
+    viscosity, diffusivity,
 
     # Lagrangian particle tracking
     LagrangianParticles,
--- a/test/test_turbulence_closures.jl
+++ b/test/test_turbulence_closures.jl
@@ -200,7 +200,7 @@ end
             κ = diffusivity(model.closure, model.diffusivity_fields, Val(:c)) 
             κ_dx_c = κ * ∂x(c)
             ν = viscosity(model.closure, model.diffusivity_fields)
-            ν_dx_u = ν * ∂x(c)
+            ν_dx_u = ν * ∂x(u)
             @test ν_dx_u[1, 1, 1] == 0.0
             @test κ_dx_c[1, 1, 1] == 0.0
         end


tomchor:
I also noticed we're not testing `closure = nothing` in `test_turbulence_closures.jl` (although there's some implicit testing since that is the default option). So I'd also like to add `nothing` to the list of tested closures https://github.com/CliMA/Oceananigans.jl/blob/6cbe8cf339b58170c51846160d832a431e42715c/test/dependencies_for_runtests.jl#L57-L64

src/TurbulenceClosures/turbulence_closure_implementations/nothing_closure.jl
```diff
@@ -5,3 +5,6 @@
 
 calculate_diffusivities!(diffusivities, ::Nothing, args...) = nothing
 calculate_diffusivities!(::Nothing, ::Nothing, args...) = nothing
+
+@inline viscosity(::Nothing, ::Nothing) = 0.0
+@inline diffusivity(::Nothing, ::Nothing, ::Val{id}) where id = 0.0
```
navidcy, line 10:
```suggestion
@inline viscosity(::Nothing, ::Nothing) = 0
@inline diffusivity(::Nothing, ::Nothing, ::Val{id}) where id = 0
```

tomchor: Change applied. Thanks for the suggestion. Can I ask why `0` is preferred here though?

navidcy: Multiplying with `0` does not impose type conversion. See, e.g.,

```Julia
julia> a = Float32(10)
10.0f0

julia> b = Float64(10)
10.0

julia> typeof(0.0*a)
Float64

julia> typeof(0.0*b)
Float64

julia> typeof(0*a)
Float32

julia> typeof(0*b)
Float64
```

tomchor: Ah, I didn't know. Thanks for the tip/info!

navidcy: It doesn't work always...

```Julia
julia> typeof(0 * Int16(3))
Int64
```

navidcy: I guess 0 is `Int64` and thus multiplying `Int64` with `Int16` imposes conversion to `Int64`. :)

tomchor: > It doesn't work always...
> 
> ```julia
> julia> typeof(0 * Int16(3))
> Int64
> ```

Hmm, interesting. Although our models are going to be float regardless so if that only happens for Ints then I guess it doesn't matter for us

navidcy:
I don't really get why returning 0 is useful though. You say "for writing general code from the user perspective" but I don't quite understand that.

tomchor:
> I don't really get why returning 0 is useful though. You say "for writing general code from the user perspective" but I don't quite understand that.

What I mean is that it's makes it easier for a user to write something like:

```julia
u_sgs_flux(model) =  - viscosity(model.closure, model.diffusivity_fields) * ∂z(model.velocities.u)
```

and have it (correctly) return 0 when `closure=nothing`.

I've been writing utilities of this kind for the past couple of months for a project which has simulations both without and without closure, and I've been using if-else statements to deal with that. It's not the worst thing, but having these extra two lines in Oceananigans would help me (and possibly other users) have cleaner code in these instances.

Although admittedly I don't expect a big chunk of users to need that. The main reasons for me to propose this is that the behavior of `viscosity()` and `diffusivity()` will now be more consistent (i.e. they would work for all (?) closures we have so far), and it's only a simple two-line addition.

navidcy:
OK, I see.

@glwagner, @simone-silvestri, does returning `0` instead of `nothing` implies any allocations?

glwagner (approved):

glwagner:
> > I don't really get why returning 0 is useful though. You say "for writing general code from the user perspective" but I don't quite understand that.
> 
> What I mean is that it's makes it easier for a user to write something like:
> 
> ```julia
> u_sgs_flux(model) =  - viscosity(model.closure, model.diffusivity_fields) * ∂z(model.velocities.u)
> ```
> 
> and have it (correctly) return 0 when `closure=nothing`.
> 
> I've been writing utilities of this kind for the past couple of months for a project which has simulations both without and without closure, and I've been using if-else statements to deal with that. It's not the worst thing, but having these extra two lines in Oceananigans would help me (and possibly other users) have cleaner code in these instances.
> 
> Although admittedly I don't expect a big chunk of users to need that. The main reasons for me to propose this is that the behavior of `viscosity()` and `diffusivity()` will now be more consistent (i.e. they would work for all (?) closures we have so far), and it's only a simple two-line addition.

In the future instead of using an if-statement, you can extend `viscosity` and `diffusivity` in your own code (temporarily, until a PR like this one is merged)

navidcy (approved):

PR #2952, opened 2023-03-01
Convert docstring example to doctest

navidcy:


tomchor (approved):
LGTM

let's make sure tests pass though

PR #2924, opened 2023-02-13
Remove event system from KA and update to CUDA v4.1

vchuravy:
This likely removes some of the single device parallelism,
but decreases the risk for user mistake (everything is stream-ordered)

This will cause severe merge conflicts with open PRs and depends on currently unreleased
versions of CUDA & KernelAbstractions. So I might regret doing it now, but I was curious
to see what the damage would be.

PR #2953, opened 2023-03-02
MPI communication and computation overlap in the `HydrostaticFreeSurfaceModel`

### LINKED ISSUE #615: Overlapping computation and MPI halo communication
In PR #590 [WIP] I've prototyped how I've thought about adding support for distributed parallelism by adding a non-invasive `Distributed` MPI layer on top of Oceananigans to keep the core code MPI-free.

At last week's CliMA software meeting @lcw and @jkozdon have pointed out a potential limitation of this approach: when running on many nodes and communication starts to eat up a lot of compute time it becomes beneficial to overlap computation and communication. Abstractions such as `CLIMA.MPIStateArray` help a lot with this but require MPI to be "baked in".

Obviously this issue won't be tackled for a while until we have a working distributed model and need more performance, so I'm just documenting the issue here for future discussion.

I think we can achieve this by splitting a kernel like `calculate_interior_source_terms!` into two kernels, one that computes source terms "near" the boundary (1-2? grid points from any boundary as needed), then halo communication can happen while a second more compute-intensive kernel computes the source terms in the rest of the interior.

But that only helps with one particular instance of halo communication. There will be other halo communications needed that may be impossible to overlap with compute-intensive kernels. Pursuing overlapping in this manner to the extreme and applying it to as many kernels as possible may be detrimental to code clarity.

Once we want more distributed performance we should go through the algorithm and minimize the number of halo communications (i.e. calls to `fill_halo_regions!`).

cc @leios @jm-c
```[tasklist]
### Tasks
```

### LINKED ISSUE #1882: Multi GPU scaling is very poor
I recently ran the weak scaling shallow water model benchmark with the MultiGPU architecture on Satori, thanks to @christophernhill.
Here are the results:
<html>
<body>

size | ranks | min | median | mean | max | memory | allocs | samples
-- | -- | -- | -- | -- | -- | -- | -- | --
(4096, 256) | (1, 1) | 2.765 ms | 2.786 ms | 2.849 ms | 3.374 ms | 2.03 MiB | 5535 | 10
(4096, 512) | (1, 2) | 6.932 ms | 7.081 ms | 8.037 ms | 26.174 ms | 2.03 MiB | 5859 | 20
(4096, 1024) | (1, 4) | 12.592 ms | 14.603 ms | 16.417 ms | 31.468 ms | 2.03 MiB | 5859 | 40

</body>
</html>

<html>
<body>

size | ranks | slowdown | efficiency | memory | allocs
-- | -- | -- | -- | -- | --
(4096, 256) | (1, 1) | 1.0 | 1.0 | 1.0 | 1.0
(4096, 512) | (1, 2) | 2.54127 | 0.393505 | 1.00271 | 1.05854
(4096, 1024) | (1, 4) | 5.24053 | 0.19082 | 1.00271 | 1.05854

</body>
</html>

The results are not good but at least we can benchmark multi-GPU performance now.

simone-silvestri:
Overlapping MPI communication and computation in the `HydrostaticFreeSurfaceModel`.

In particular, this PR introduces the `async` keyword argument to `fill_halo_regions!` that allows launching MPI operations without waiting for the communication to complete. 
This allows hiding MPI passing of barotropic auxiliary variables behind the implicit vertical solver and prognostic variables behind the tendency calculations. The latter is done by splitting the tendency kernels into an _interior_ kernel that calculates  tendencies between `H` and `N-H`, and a _boundary_ kernel, executed once communication is complete, that calculates tendencies adjacent to boundaries.

This PR is contingent on #2924 

Closes #615
Closes #1882


glwagner:
There are changes in `NonhydrostaticModel` as well --- are we overlapping comms and compute there?

glwagner:
> In particular, this PR introduces the async keyword argument to fill_halo_regions! that allows launching MPI operations without waiting for the communication to complete.

Would `blocking=true/false` be a better word to describe how we are controlling `fill_halo_regions!`?

I think "async" isn't quite the right word. "async" describes the overall algorithm (eg performing some computation while the communication is taking place), but does not describe what is happening specifically within `fill_halo_regions`. It's possible to write `async=true` without overlapping communication and computation.

glwagner:
I'll wait till it's ready to review in detail, but my main high-level comment is that we need to ensure that, at the end of a time-step, both the prognostic state (horizontal velocities and tracers) and auxiliary state (pressure, vertical velocity, eddy diffusivities) are all consistent and available for output at the current model time.

If we intertwine communication with the computation of the auxiliary state and tendencies, then we should _define_ the tendencies as part of the auxiliary state. This will change the semantics and logic of the time stepping loop. But I think it at least as rational as our previous organization of events.

The main change is that tendencies will now be computed one "extra" time in a simulation (at the very last time-step, the tendencies are not needed if no further time-steps will be taken). In the vast majority of cases this extra cost is negligible because simulations run for hundreds or hundreds of thousands of time-steps.

There is the slightly possibility of pessimizing the edge case of a simulation that takes one time step, which may be useful for parameter estimation. For that purpose we may want to avoid computing the "extra" tendency. I'm thinking though that we should save that additional optimization until we need it.

glwagner:
PS a minor tweak that would help with the issues above would be to add a kwarg to `update_state!` like

```julia
function update_state!(model; compute_tendencies=true)
```

Then under some circumstances (ie in `set!` or for the last time-step potentially) we can avoid computing the tendencies if we want.

PS PS should we use "compute" rather than "calculate" ? We use both words to mean the same thing which is silly...

src/TimeSteppers/runge_kutta_3.jl
```diff
@@ -110,14 +107,13 @@ function time_step!(model::AbstractModel{<:RungeKutta3TimeStepper}, Δt; callbac
...
     store_tendencies!(model)
-    update_state!(model, callbacks)
     update_particle_properties!(model, first_stage_Δt)
 
     #
     # Second stage
     #
 
-    calculate_tendencies!(model, callbacks)
+    update_state!(model, callbacks)
```
glwagner: We want to leave `update_state!` where it is right?

src/TimeSteppers/runge_kutta_3.jl
```diff
@@ -145,7 +140,6 @@ function time_step!(model::AbstractModel{<:RungeKutta3TimeStepper}, Δt; callbac
     pressure_correct_velocities!(model, third_stage_Δt)
 
     tick!(model.clock, third_stage_Δt)
-    update_state!(model, callbacks)
```
glwagner, line 148: We need to `update_state!` at the end of a time-step not before

src/TimeSteppers/runge_kutta_3.jl
```diff
@@ -81,9 +81,6 @@ stage.
 function time_step!(model::AbstractModel{<:RungeKutta3TimeStepper}, Δt; callbacks=[])
     Δt == 0 && @warn "Δt == 0 may cause model blowup!"
 
-    # Be paranoid and update state at iteration 0, in case run! is not used:
-    model.clock.iteration == 0 && update_state!(model, callbacks)
```
glwagner, line 85: We want to keep this

src/TimeSteppers/quasi_adams_bashforth_2.jl
```diff
@@ -85,18 +85,14 @@ function time_step!(model::AbstractModel{<:QuasiAdamsBashforth2TimeStepper}, Δt
...
 
-    @apply_regionally calculate_tendencies!(model, callbacks)
-    
     ab2_step!(model, Δt, χ) # full step for tracers, fractional step for velocities.
     calculate_pressure_correction!(model, Δt)
 
     @apply_regionally correct_velocities_and_store_tendecies!(model, Δt)
 
     tick!(model.clock, Δt)
-    update_state!(model, callbacks)
```
glwagner, line 100: We need to keep this

glwagner:
Change

https://github.com/CliMA/Oceananigans.jl/blob/6009b2aafaa1517e8a08855fce5b7ea5a5033b5b/src/Models/HydrostaticFreeSurfaceModels/set_hydrostatic_free_surface_model.jl#L49

to 

```julia
update_state!(model; compute_tendencies=false)
```

src/TimeSteppers/runge_kutta_3.jl
simone-silvestri: it is there, just changed for a hot second 😅

src/Distributed/halo_communication.jl
```diff
@@ -274,9 +299,13 @@ for side in sides
             send_tag = $side_send_tag(arch, location, local_rank, rank_to_send_to)
 
             @debug "Sending " * $side_str * " halo: local_rank=$local_rank, rank_to_send_to=$rank_to_send_to, send_tag=$send_tag"
-            send_req = MPI.Isend(send_buffer, rank_to_send_to, send_tag, arch.communicator)
+            
+            send = @async begin
```
vchuravy: use `Threads.@spawn`. `@async` is eventually going to be deprrecated

src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
```diff
@@ -198,7 +194,7 @@ end
     κᵉ = ifelse(entraining, Cᵉ, zero(grid))
 
     # Shear mixing diffusivity and viscosity (diffused in the horizontal to add non-locality)
-    Ri = ℑxyᶜᶜᵃ(i, j, k, grid, ℑxyᶠᶠᵃ, Riᶜᶜᶠ, velocities, tracers, buoyancy)
+    Ri = Riᶜᶜᶠ(i, j, k, grid, velocities, N²)
```
glwagner, line 197: it's elsewhere defined with `tracers, buoyancy` so might be better to keep here. Should all be inlined anyways so no difference in compute.

glwagner:
Is this still active? Is there another PR that will replace it?

simone-silvestri:
I have another branch for the full Gordon Bell work, but it's probably better that I split it up in different PRs, so the splitting of communication and computation will be done here. (This PR is a bit outdated so I will update it)

navidcy:
closing in favor of #https://github.com/CliMA/Oceananigans.jl/pull/3125

PR #2968, opened 2023-03-12
Fixes typo in `abstract_scalar_biharmonic_diffusivity_closure.jl`

tomchor:


src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
```diff
@@ -1,7 +1,7 @@
 using Oceananigans.Grids: peripheral_node
 
 """
-    abstract type AbstractScalarDiffusivity <: AbstractTurbulenceClosure end
+    abstract type AbstractScalarBiharmonicDiffusivity <: AbstractTurbulenceClosure end
 
 Abstract type for closures with *isotropic* diffusivities.
```
navidcy, line 6:
```suggestion
Abstract type for closures with *isotropic* biharmonic diffusivities.
```

tomchor:
@navidcy's comment made me realize that `AbstractScalarDiffusivity` has a similar emphasis on isotropic: https://github.com/CliMA/Oceananigans.jl/blob/9e44b4407ce869787facb22c5fde5857197f0536/src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl#L3-L8

And while that is correct, it may be a bit misleading to users since we can use `ScalarDiffusiity` to set-up anisotropic diffusivities using Tuples. Should we change to wording there too?

src/TurbulenceClosures/abstract_scalar_biharmonic_diffusivity_closure.jl
```diff
@@ -1,9 +1,9 @@
 using Oceananigans.Grids: peripheral_node
 
 """
-    abstract type AbstractScalarDiffusivity <: AbstractTurbulenceClosure end
+    abstract type AbstractScalarBiharmonicDiffusivity <: AbstractTurbulenceClosure end
 
-Abstract type for closures with *isotropic* diffusivities.
+Abstract type for closures with *isotropic* biharmonic diffusivities.
```
glwagner, line 6:
```suggestion
Abstract type for closures with scalar biharmonic diffusivities.
```

glwagner (approved):

tomchor:
> > @navidcy's comment made me realize that `AbstractScalarDiffusivity` has a similar emphasis on isotropic:
> > https://github.com/CliMA/Oceananigans.jl/blob/9e44b4407ce869787facb22c5fde5857197f0536/src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl#L3-L8
> > 
> > And while that is correct, it may be a bit misleading to users since we can use `ScalarDiffusiity` to set-up anisotropic diffusivities using Tuples. Should we change to wording there too?
> 
> Hm no I don't think it's correct. `AbstractScalarDiffusivity` is for representing turbulence closures that have a scalar diffusivity, not isotropic diffusion (vertical or horizontal formulations are anisotropic).

I changed the text according to your comment. But just to note, I said that is technically correct because the naming given there uses the work isotropic when the ν doesn't change in the directions of the formulation:

https://github.com/CliMA/Oceananigans.jl/blob/75820d647dcd8c00e81151b66a3f585bb0d052a9/src/TurbulenceClosures/abstract_scalar_diffusivity_closure.jl#L16-L28

Which, again, I think it's a bit misleading, but afaik it's technically correct.

glwagner:
Sure but "horizontally isotropic" is not the same thing as "isotropic".

tomchor:
> It's just that "horizontally isotropic" is not the same thing as "isotropic". What is misleading?

sorry, nvmd, I'm explaning myself poorly. Nothing is misleading anymore since I removed the word "isotropic" in the last commit. We're good to merge whenever tests pass.

Issue #2198, opened 2022-01-27
`HydrostaticFreeSurface` model on a grid with `Flat z`


navidcy:
Does this make sense (physically)?

```Julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.6.5 (2021-12-19)
 _/ |\__'_|_|_|\__'_|  |
|__/                   |

julia> using Oceananigans
[ Info: Oceananigans will use 12 threads

julia> grid = RectilinearGrid(topology=(Periodic, Bounded, Flat), size=(1, 1), extent=(1, 2))
1×1×1 RectilinearGrid{Float64, Periodic, Bounded, Flat} on CPU with 1×1×0 halo
├── Periodic x ∈ [0.0, 1.0) regularly spaced with Δx=1.0
├── Bounded  y ∈ [0.0, 2.0] regularly spaced with Δy=2.0
└── Flat z

julia> model = HydrostaticFreeSurfaceModel(grid=grid)
HydrostaticFreeSurfaceModel{CPU, Float64}(time = 0 seconds, iteration = 0)
├── grid: 1×1×1 RectilinearGrid{Float64, Periodic, Bounded, Flat} on CPU with 1×1×0 halo
├── tracers: (:T, :S)
├── closure: Nothing
├── buoyancy: Buoyancy{SeawaterBuoyancy{Float64, LinearEquationOfState{Float64}, Nothing, Nothing}, Oceananigans.Grids.ZDirection}
└── coriolis: Nothing

julia> keys(fields(model))
(:u, :v, :η, :T, :S)
```

I guess not. Perhaps the `HydrostaticFreeSurfaceModel` should print a warning to users that try to do that?

francispoulin:
Good idea. Maybe even suggest the user try `ShallowWaterModel`, as that is a better way to try this configuration? ;)

glwagner:
Right, we can't support `Flat` z right now because we require two grid points in the vertical velocity (one grid point at the "bottom", one at the "top" coincident with the location of the free surface).

We might be able to refactor our algorithm to permit this case.

However, a `NonhydrostaticModel` with `Flat` z has equivalent physics (it's also hydrostatic...). Maybe the error message can point that out.

[...]

PR #2975, opened 2023-03-14
Multi-layer shallow water system

### LINKED ISSUE #2507: Multi-layer shallow water model

dhruvbhagtani:
Closes #2507. 

This PR attempts to extend shallow water system to multiple layers.


dhruvbhagtani:
The equations for conservative and vector-invariant multi-layer shallow water system are outlined in this doc.

[Multi_layer_shallow_water_model_formulation.pdf](https://github.com/CliMA/Oceananigans.jl/files/10974444/Multi_layer_shallow_water_model_formulation.pdf)

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
```diff
@@ -7,8 +7,43 @@ using Oceananigans.TurbulenceClosures: ∇_dot_qᶜ, ∂ⱼ_τ₁ⱼ, ∂ⱼ_τ
 @inline half_g_h²(i, j, k, grid, h, g)  = @inbounds 1/2 * g * h[i, j, k]^2
 @inline h_plus_hB(i, j, k, grid, h, hB) = @inbounds h[i, j, k] + hB[i, j, k]
 
-@inline x_pressure_gradient(i, j, k, grid, g, h, hB, formulation) = ∂xᶠᶜᶜ(i, j, k, grid, half_g_h², h, g)
-@inline y_pressure_gradient(i, j, k, grid, g, h, hB, formulation) = ∂yᶜᶠᶜ(i, j, k, grid, half_g_h², h, g)
+#@inline x_pressure_gradient(i, j, k, grid, g, h, hB, formulation) = ∂xᶠᶜᶜ(i, j, k, grid, half_g_h², h, g)
+#@inline y_pressure_gradient(i, j, k, grid, g, h, hB, formulation) = ∂yᶜᶠᶜ(i, j, k, grid, half_g_h², h, g)
+
```
navidcy, line 12:
```suggestion
```

generally commented-out code is not good practice... but OK to keep it there temporarily if it helps you

GitHub retains the full history of all commits so we can always go back and find anything we want!

src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -129,10 +129,14 @@ function ShallowWaterModel(;
...
-                            "Use `topology = ($(topology(grid, 1)), $(topology(grid, 2)), Flat)` " *
-                            "when constructing `grid`."))
+    #topology(grid, 3) === Flat ||
+    #    throw(ArgumentError("ShallowWaterModel requires `topology(grid, 3) === Flat`. " *
+    #                        "Use `topology = ($(topology(grid, 1)), $(topology(grid, 2)), Flat)` " *
+    #                        "when constructing `grid`."))
+
+
+    topology(grid, 3) === Flat ? println("Proceeding with single-layer shallow water model") : println(
+        "Proceeding with multi-layer shallow water model")
```
glwagner, line 139:
```suggestion
    grid.Nz == 1 ? println("Proceeding with single-layer shallow water model") : println(
        "Proceeding with multi-layer shallow water model")
```

navidcy:
Btw, the little bit of code you wrote for computing the pressure would _only_ work on CPU.

We need to discuss if we want to be launching 2D kernels (one for each layer) and then sum them _or_ launch 3D kernels.

navidcy:
Shall we extend the `ShallowWaterModel` type to incorporate single- and multi-layer configs? What other properties does a multilayer shallow water type need? The reduced gravities that correspond to each interface is an obvious one... any other?

The number of layers could be a parameter of the type so we can dispatch on it in various methods. E.g.,

```julia
mutable struct ShallowWaterModel{N, G, A<:AbstractArchitecture, T, GR, V, U, R, F, E, B, Q, C, K, TS, FR} <: AbstractModel{TS}
              number_of_layers :: Int       # Number of fluid layers
                          grid :: G         # Grid of physical points on which `Model` is solved
                  architecture :: A         # Computer `Architecture` on which `Model` is run
                         clock :: Clock{T}  # Tracks iteration number and simulation time of `Model`
    gravitational_acceleration :: GR        # Gravitational acceleration, full, or reduced
                     advection :: V         # Advection scheme for velocities, mass and tracers
                    velocities :: U         # Velocities in the shallow water model
                      coriolis :: R         # Set of parameters for the background rotation rate of `Model`
                       forcing :: F         # Container for forcing functions defined by the user
                       closure :: E         # Diffusive 'turbulence closure' for all model fields
                    bathymetry :: B         # Bathymetry/Topography for the model
                      solution :: Q         # Container for transports `uh`, `vh`, and height `h`
                       tracers :: C         # Container for tracer fields
            diffusivity_fields :: K         # Container for turbulent diffusivities
                   timestepper :: TS        # Object containing timestepper fields and parameters
                   formulation :: FR        # Either conservative or vector-invariant

    function ShallowWaterModel(number_of_layers, grid, all_other_things...)
        arch = architecture(grid)
        return new{number_of_layers}(number_of_layers, grid, arch, all_other_things...)
    end
end
```

dhruvbhagtani:
I'm not familiar with the Oceananigans code writing practice, but it looks like the parameter `number_of_layers` is redundant as it is already present in `grid.Nz`. Is the parameter added in a case where the user inputs the number of layers explicitly?

navidcy:
It's just a suggestion... Perhaps it's not needed. But having it as a parameter of the type could be useful, even if the parameter is the same value as `grid.Nz` because if it's part of type you can write methods that do different things based on the number of layers your model has. E.g.,

```Julia
function compute_this_and_that(model::ShallowWaterModel{1})
    a = 1 # simple calculation for single-layer models
    return a
end

function compute_this_and_that(model::ShallowWaterModel)
    number_of_layers = model.number_of_layers
    a = sum(rand(number_of_layers)) # more complicated stuff only needed for multi-layer models
    return a
end
```

This way we don't use `if number_of_layers > 1; do this; else; do that` which reduced performance.

glwagner:
We can dispatch on `grid.Nz`, but we could dispatch on `ShallowWaterModel{1}`, `ShallowWaterModel{2}`, etc

dhruvbhagtani:
Ah, I see your point now!

dhruvbhagtani:
Instead of having `number_of_layers` as a parameter, I've rewritten the code to have `vertical_model` as a parameter, which can either be `SingleLayerModel()` or `MultiLayerModel()`.

dhruvbhagtani:
Before proceeding further, I guess we should discuss a bit on whether we want a single parameter for gravitational acceleration (`effective_gravitational_acceleration`, which can encapsulate both gravitational acceleration `g` and reduced gravity `g'` terms) or two parameters (one for `g` and `g'` each), with Navid's code above an example of the latter. I am inclined towards the former as I feel it's cleaner (scientifically and for coding purposes), but happy to go either ways.

src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -186,9 +195,10 @@ function ShallowWaterModel(;
     closure = with_tracers(tracernames(tracers), closure)
 
     model = ShallowWaterModel(grid,
+                              vertical_model,
                               arch,
                               clock,
-                              eltype(grid)(gravitational_acceleration),
+                              gravitational_acceleration,
```
navidcy, line 201: if `gravitational_acceleration` is a number then we better ensure that it has the same type as the grid to avoid type instability

```suggestion
                              eltype(grid)(gravitational_acceleration),
```

src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -35,8 +35,9 @@ function ShallowWaterSolutionFields(grid, bcs, prognostic_names)
     return NamedTuple{prognostic_names[1:3]}((u, v, h))
 end
 
-mutable struct ShallowWaterModel{G, A<:AbstractArchitecture, T, GR, V, U, R, F, E, B, Q, C, K, TS, FR} <: AbstractModel{TS}
+mutable struct ShallowWaterModel{G, A<:AbstractArchitecture, NL, T, GR, V, U, R, F, E, B, Q, C, K, TS, FR} <: AbstractModel{TS}
                           grid :: G         # Grid of physical points on which `Model` is solved
+                vertical_model :: NL        # Either single-layer or multi-layer
```
glwagner: maybe `layer_formulation`? (we use the word "formulation" to refer to different ways of writing the equations, ie `VectorInvariantFormulation` and `ConservativeFormulation`... maybe we can reuse that word here.)

glwagner: I do like the economy and literalness of simply writing the number of layers though

navidcy: I also like the simplicity of a type field or constructor kwarg called `number_of_layers` instead of the tad more esoteric `vertical_model`.

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
```diff
@@ -7,8 +7,36 @@ using Oceananigans.TurbulenceClosures: ∇_dot_qᶜ, ∂ⱼ_τ₁ⱼ, ∂ⱼ_τ
...
-@inline y_pressure_gradient(i, j, k, grid, g, h, hB, formulation) = ∂yᶜᶠᶜ(i, j, k, grid, half_g_h², h, g)
+@inline x_pressure_gradient(i, j, k, grid, g, h, hB, ::SingleLayerModel, formulation) = ∂xᶠᶜᶜ(i, j, k, grid, half_g_h², h, g)
+@inline y_pressure_gradient(i, j, k, grid, g, h, hB, ::SingleLayerModel, formulation) = ∂yᶜᶠᶜ(i, j, k, grid, half_g_h², h, g)
+    
+@inline function x_pressure_gradient(i, j, k, grid, g, h, hB, ::MultiLayerModel, formulation)
+    Nz = grid.Nz
+    ## Should be set to grid dimensions
+    η[i, j, k] = hB[i, j, k] + cumsum(h, dims = 3)
+    ## Should be set to grid dimensions
+    pr = 0
```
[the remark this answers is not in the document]
dhruvbhagtani: A result of clumsy coding. It is the pressure, but I haven't tested this piece of code and don't trust it much. For now it is present as a placeholder on what operations to perform when tackled with a multi-layer system.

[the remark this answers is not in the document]
dhruvbhagtani: A result of clumsy coding. It is the pressure, but I haven't tested this piece of code and don't trust it much. For now it is present to symbolise what operations to perform when tackled with a multi-layer system.

src/Models/ShallowWaterModels/shallow_water_model.jl
dhruvbhagtani: Yes, I agree with you. I changed it because as of now, I wrote the code assuming a single `gravitational_acceleration` variable instead of two.

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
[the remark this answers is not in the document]
navidcy: let's call it `pressure` instead of `pr`?

dhruvbhagtani:
> Were you suggesting to combine these in a single `k`-long tuple, e.g.,
> 
> ```julia
> gravitational_acceleration .* (1, (ρ[k] - ρ[k-1]) / ρ[k-1], (ρ[k-1] - ρ[k-2]) / ρ[k-2], ..., (ρ[2] - ρ[1]) / ρ[1] )
> ```
> 
> which will become a number for single layers?

Yes, that's exactly what I was suggesting, since gravitational_acceleration could be understood as the value of g that each layer feels, so not sure why we need different parameters for both.

src/Models/ShallowWaterModels/shallow_water_model.jl
dhruvbhagtani: No inclinations here, so happy to revert back to `number_of_layers`.

navidcy:
> > Were you suggesting to combine these in a single `k`-long tuple, e.g.,
> > ```julia
> > gravitational_acceleration .* (1, (ρ[k] - ρ[k-1]) / ρ[k-1], (ρ[k-1] - ρ[k-2]) / ρ[k-2], ..., (ρ[2] - ρ[1]) / ρ[1] )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which will become a number for single layers?
> 
> Yes, that's exactly what I was suggesting, since gravitational_acceleration could be understood as the value of g that each layer feels, so not sure why we need different parameters for both.

Ok! This type field will be pf different type for single and multilayer models then.

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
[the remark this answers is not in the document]
dhruvbhagtani: Yes, I'll change it

francispoulin:
> Might make sense to resolve #2198 too

I see that it's now closed?

francispoulin:
I think this is a a big step forward and happy to share my two cents worth.

I agree that we should keep `gravitational_acceleration` the same as before.

We will need to have a bunch of `g'`.  Since we have a free surface, we assume that we have air above, then `g'` of the top layer is really just `g`.  That's why I would suggest having n reduced gravities for n layers.

We will need n interfaces and we could also store the n heights.  Sometimes it's convenient to have one or the other.  For efficiency we only need one, but for convenience we might want to have both.  What are people's preferences?

The more layers we have, the thinner each layer tends to be.  This means that it's more likely that layer depths can vanish. If this ever happens, then a numerical instability can happen.  I remember discussing with @simone-silvestri using the positive preserving property of WENO to deal with this but I also remember that a preliminary test showed that it wasn't actually ensure to be positive. Sadly, I didn't follow up.  Any idea if that was fixed?

simone-silvestri:
I think the WENO positivity preserving scheme we have is not actually positivity preserving because preserves 1D positivity. To have positivity preserving we need a 2D scheme (or at least to do proper tests/improvement on the scheme we have).

navidcy:
How about this @dhruvbhagtani?

```Julia
struct SingleLayer end
struct MultipleLayers end

struct ShallowWaterModel{N, T}
    number_of_layers :: Int
    parameter :: T

    function ShallowWaterModel(number_of_layers, parameter)
        if number_of_layers == 1
            return new{SingleLayer, typeof(parameter)}(number_of_layers, parameter)
        else number_of_layers > 1
            return new{MultipleLayers, typeof(parameter)}(number_of_layers, parameter)
        end
    end
end

const SingleLayerShallowWaterModel{T} = ShallowWaterModel{SingleLayer, T}

model1 = ShallowWaterModel(1, 2.0)

model2 = ShallowWaterModel(2, 2.3)

@show typeof(model1) <: ShallowWaterModel
true
@show typeof(model1) <: SingleLayerShallowWaterModel
true
@show typeof(model2) <: ShallowWaterModel
true
@show typeof(model2) <: SingleLayerShallowWaterModel
false
```

glwagner:
Or `SingleLayer` and `MultiLayer`

navidcy:
What I did above @dhruvbhagtani allow us both to have `number_of_layers` as a field of the type but also to have a parameter of the type which is either `SingleLayer` or `MultiLayer` on which we can dispatch on.

francispoulin:
> I think the WENO positivity preserving scheme we have is not actually positivity preserving because preserves 1D positivity. To have positivity preserving we need a 2D scheme (or at least to do proper tests/improvement on the scheme we have).

Thanks for the reply.  

If there is a paper that shows how to do this I might want to give it a try in april, after I finish teaching.  But if someone else wants to do it, you are welcome.

I will point out that the first order upwinding is positive preserving.  Not very accurate, but for testing it is fast and won't give rise to negative heights, which all other schemes can do.

src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -119,7 +126,8 @@ Keyword arguments
 function ShallowWaterModel(;
                            grid,
                            gravitational_acceleration,
-                      vertical_model = SingleLayerModel(),
+                vertical_formulation = SingleLayerModel(),
```
simone-silvestri, line 132: can we imply this from the number of layers?

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
```diff
@@ -7,34 +7,34 @@ using Oceananigans.TurbulenceClosures: ∇_dot_qᶜ, ∂ⱼ_τ₁ⱼ, ∂ⱼ_τ
...
-@inline y_pressure_gradient(i, j, k, grid, g, h, hB, ::SingleLayerModel, formulation) = ∂yᶜᶠᶜ(i, j, k, grid, half_g_h², h, g)
+@inline x_pressure_gradient(i, j, k, grid, g, h, hB, ::SingleLayerModel, ::ConservativeFormulation) = ∂xᶠᶜᶜ(i, j, k, grid, half_g_h², h, g)
+@inline y_pressure_gradient(i, j, k, grid, g, h, hB, ::SingleLayerModel, ::ConservativeFormulation) = ∂yᶜᶠᶜ(i, j, k, grid, half_g_h², h, g)
     
-@inline function x_pressure_gradient(i, j, k, grid, g, h, hB, ::MultiLayerModel, formulation)
+@inline function x_pressure_gradient(i, j, k, grid, g, h, hB, ::MultiLayerModel, ::ConservativeFormulation)
     Nz = grid.Nz
     ## Should be set to grid dimensions
-    η[i, j, k] = hB[i, j, k] + cumsum(h, dims = 3)
+    η[i, j, k] = hB[i, j, k] + cumsum(h, dims = 3)[i, j, k]
```
simone-silvestri, line 16: this `cumsum` might be very slow for a large number of layers. A way around it would be to add another field `∫h` to the model and precompute the sum

src/Models/ShallowWaterModels/shallow_water_model.jl
dhruvbhagtani: Not entirely sure what you are suggesting - we do use `number_of_layers` in the function to imply the `vertical_formulation`. Do you suggest we can do it within the function declaration itself?

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
dhruvbhagtani: True. I can work on it!

src/Models/ShallowWaterModels/shallow_water_model.jl
navidcy: @simone-silvestri is suggesting that the user should provide with the constructor `number_of_layers` and the constructor should put the right `vertical_model` formulation when constructing a model. Right?

src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -35,11 +35,14 @@ function ShallowWaterSolutionFields(grid, bcs, prognostic_names)
     return NamedTuple{prognostic_names[1:3]}((u, v, h))
 end
 
-mutable struct ShallowWaterModel{G, A<:AbstractArchitecture, T, GR, V, U, R, F, E, B, Q, C, K, TS, FR} <: AbstractModel{TS}
+mutable struct ShallowWaterModel{G, A<:AbstractArchitecture, VM, T, GR, RG, N, V, U, R, F, E, B, Q, C, K, TS, FR} <: AbstractModel{TS}
                           grid :: G         # Grid of physical points on which `Model` is solved
+          vertical_formulation :: VM        # Either single-layer or multi-layer
```
francispoulin, line 40: If we are using `vertical_formulation`, should we use `VF` instead of `VM`, to make it more intuitive?

src/Models/ShallowWaterModels/shallow_water_model.jl
dhruvbhagtani: Okay, but the user implicitly mentions the number of vertical layers when creating a grid, so why ask them for the `number_of_layers` for the second time when we ourselves can get it from the grid creation?

src/Models/ShallowWaterModels/shallow_water_model.jl
dhruvbhagtani: Agreed! I'll change it

src/Models/ShallowWaterModels/shallow_water_model.jl
navidcy: True. 

I think these are secondary details to deal later. Let’s first have a working model and we can finesse the user interface at the end.

simone-silvestri: I agree, with all of that, there are `number_of_layers` and `vertical_formulation` in the constructor of the model. Both of them are redundant

dhruvbhagtani: Thanks for the suggestion @simone-silvestri. It was just me being dense and not being able to follow - I've changed it and will push it.

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
```diff
@@ -7,8 +7,35 @@ using Oceananigans.TurbulenceClosures: ∇_dot_qᶜ, ∂ⱼ_τ₁ⱼ, ∂ⱼ_τ
...
+    ## Should be set to grid dimensions
+    pressure[i, j, k] = 0
+    for iter in range(k, Nz, step = 1)
+        pressure[i, j, k] = pressure[i, j, k] + gprime[iter] * η[i, j, iter]
+    end
+    pressure[i, j, k] = pressure[i, j, k] + g * η[i, j, Nz]
+    return ∂xᶠᶜᶜ(i, j, k, grid, h[i, j, k] * pressure[i, j, k], h, g)
+end
 
 @inline x_pressure_gradient(i, j, k, grid, g, h, hB, ::VectorInvariantFormulation) = g * ∂xᶠᶜᶜ(i, j, k, grid, h_plus_hB, h, hB)
```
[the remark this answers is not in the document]
dhruvbhagtani: I am starting to do it now!

navidcy:
@dhruvbhagtani are there plans to push forward on this at the moment?

dhruvbhagtani:
I feel like there is still a lot of work, and since I am nearing the end of my PhD, it is unfortunate I can't spend a lot of hours on it as of now. I'm sorry, but I think it's best to close this PR at the moment, although I'm interested in continuing it sometime in the future if time permits.

navidcy:
@dhruvbhagtani I'm closing this as it seems stale you (or anyone!) should feel free to re-open and resume whenever.

PR #2881, opened 2023-01-26
Horizontal regridding and stabilizing CATKE features

glwagner:
This PR adds regridding in x (and there was previously untested regridding in y). Regridding is only supported for tracer fields (though it's not a huge amount of work to extend the schemes for any field).

I believe the regridding is conservative for `RectilinearGrid`, and also conservative for `LatitudeLongitudeGrid` when the grids are aligned (ie when all grid lines of the coarser of the two grids is shared with the finer grid).

In the case of partial overlaps on curvilinear grids, the volumes of the overlaps are approximated and thus the regridding is not perfectly conservative. It could make sense to emit a warning in this case.

In principle, this should allow regridding in 3D by forming intermediate regridding in x, y, and z. We leave multi-dimensional regridding to the user for now.

This PR also adds a `minimum_turbulent_kinetic_energy` and `negative_turbulent_kinetic_energy_damping_timescale` to `CATKEVerticalDiffusivity`, either of which can help stabilize 2D or 3D simulations with CATKE (where due to advection TKE may go negative).

PR #2984, opened 2023-03-17
More informative message when user provides smaller halo size than ImmersedBoundraryGrid requires

### LINKED ISSUE #2983: Ambiguous halo size error in HydrostaticFreeSurfaceModel
I recently received the error

```
ERROR: LoadError: ArgumentError: The grid halo (3, 3, 3) must be at least equal to (4, 4, 4). Note that an ImmersedBoundaryGrid requires an extra halo point.
```

It's not clear what "extra halo point" means. I think this error needs to be a little bit more specific like, "Note that an ImmersedBoundaryGrid requires an extra halo point in all directions compared to a non-immersed grid."

navidcy:
Closes #2983

The PR also took the opportunity to remove the warning from the `regrid!` docstring since the warning does not hold after merging #2881

### DIFF
--- a/src/Fields/regridding_fields.jl
+++ b/src/Fields/regridding_fields.jl
@@ -15,10 +15,6 @@ const c = Center()
 
 Regrid field `b` onto the grid of field `a`. 
 
-!!! warning "Functionality limitation"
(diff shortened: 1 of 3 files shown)


glwagner (approved):

glwagner:
yahhhh

PR #2995, opened 2023-03-22
Time-averaged eddy diffusivities for `RiBasedVerticalDiffusivity`

glwagner:
This PR changes the behavior `RiBasedVerticalDiffusivity` to compute eddy diffusivities with a moving average in time, rather than using instantaneous values.

This has the effect of eliminating vertical "noise" that is present without time-averaging:

# Without time-averaging

https://user-images.githubusercontent.com/15271942/226771108-213d709c-a95c-4e1e-a469-8dd7826635c1.mp4

# With time-averaging

https://user-images.githubusercontent.com/15271942/226771068-cfe964a5-033f-46e2-989f-39623a4e2912.mp4

This PR also unifies the notation for eddy diffusivities between `CATKEVerticalDiffusivity`, `RiBasedVerticalDiffusivity`, and `ConvectiveAdjustmentVerticalDiffusivity`.

~~Before merging I plan to re-calibrate the parameters for `RiBasedVerticalDiffusivity`.~~


src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
```diff
@@ -189,21 +189,30 @@ end
...
+    κᶜ = diffusivities.κᶜ
+    κᵘ = diffusivities.κᵘ
+
+    # New diffusivities
+    κᶜ⁺ = κᶜᵃ + κᵉⁿ + κᶜ★
+    κᵘ⁺ = κᵘ★
+
+    # Update by averaging in time
+    @inbounds κᶜ[i, j, k] = 1/2 * (κᶜ[i, j, k] + κᶜ⁺)
+    @inbounds κᵘ[i, j, k] = 1/2 * (κᵘ[i, j, k] + κᵘ⁺)
```
navidcy, line 215: this is the mean with the previous value... so it's not a "time average", right? there is not time-scale involved here... or to phrase it better it's a time average over a timescale that is proportional with the tilmestep. The results will become time-step dependent, right?

glwagner: Yes, it's a time-average proportional to the time step. It's like forward-Euler-stepping this equation

```julia
dt κᶜ = (κᶜ⁺ - κᶜ) / 2Δt
```

you might be right that results are sensitive to time-step... i'll test that

navidcy: A counter-argument to my argument is that not doing anything at all still produces things that depend on the time-step.

navidcy (approved):

src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
glwagner: So we could fix the time-scale by writing

```julia
d/dt κᶜ = (κᶜ⁺ - κᶜ) / τ
```

Then we update with

```julia
κᶜ = (1 - Δt/τ) κᶜ + Δt / τ * κᶜ⁺
```

glwagner: But actually backward Euler is better:

```julia
κᶜ = (ϵ * κᶜ⁺ + κᶜ) / (1 + ϵ)
```

navidcy: Great!!

simone-silvestri:
merge when tests pass?

glwagner:
Def. Trying to recalibrate but it might prove harder than I thought...

glwagner:
It's not as solid as CATKE but still pretty good

https://user-images.githubusercontent.com/15271942/227428121-bfbe766f-23fe-4afb-afbc-87d68f617f74.mp4

Issue #2928, opened 2023-02-16
Inconsistent advection from "different" velocities


simone-silvestri:
As of now, our tracers are advected by three "different" velocities 
- Velocity
- Background velocity
- Advective forcing

to these, another additional velocity is being added in PR #2802 
All these velocities are applied differently and might have different advection schemes.

This might be a headache in terms of understanding the cause of stability problems. Imagine a situation in which a model is characterized by 3 different velocities that all advect a centered second-order reconstruction of the tracer. If we run into stability problems it might not be immediately clear what the course of action is: is the instability caused by the interaction of the tracer with velocity 1, 2, or 3?

This is not the only issue with this implementation. The implementation is _wrong_ if we use an upwinding reconstruction of the tracer for any of the velocities e.g.:

Imagine we have a Background velocity $U_b = 1$ and an advective forcing $U_a = -1$ and we upwind separately.
The tracer advective flux should be zero everywhere, but, if we upwind:
$$(U_b c) + (U_a c) = c^L - c^R$$

If we want to be able to advect with different velocities consistently we have to ensure that
- the same scheme is used
- we _sum_ the velocities and _not_ apply reconstruction separately

A way to do it could by defining
```
struct AdvectiveU{A, B, C, D}
       u        :: A
       ubkg  :: B
       ufrc   :: C
       ubgc :: D
end

Base.getindex(u::AdvectiveU, i, j, k) = u.u[i, j, k] + u.ubkg[i, j, k] + u.ubcg[i, j, k] + u.ufrc[i, j, k]
```
and passing this velocity to the tracer tendency kernel.

simone-silvestri:
This implies that the added fields must be divergence free

glwagner:
I wonder if we can lean on the compiler to avoid having to store objects directly, eg

```julia
struct SumOfFields{N, F}
    fields :: F
    function SumOfFields{N}(fields...) = new{N, typeof(fields)}(fields)
end

@inline getindex(s::SumOfFields{1}, i, j, k) = @inbounds s.fields[1][i, j, k]
@inline getindex(s::SumOfFields{2}, i, j, k) = @inbounds s.fields[1][i, j, k] + s.fields[2][i, j, k]
@inline getindex(s::SumOfFields{3}, i, j, k) = @inbounds s.fields[1][i, j, k] + s.fields[2][i, j, k] + s.fields[3][i, j, k]
```

then inside kernels maybe it'll be enough to write

```julia
[...]

PR #3027, opened 2023-03-24
Fixes inconsistent advection from "different" velocities

jagoosw:
I've had a go at fixing #2928 which doesn't seem to be as hard as I thought it would be using @glwagner's idea for `SumOfFields. So far I've only done the nonhydrostatic model kernels but it wasn't much work and looks like its worked. Using the same code as https://github.com/CliMA/Oceananigans.jl/issues/2928#issuecomment-1433398895 I now get:

https://user-images.githubusercontent.com/26657828/227528544-2f786e88-4243-40b6-850e-1992b1e2d8d9.mp4

Which I believe is what we were expecting.

I wasn't sure where to put the `SumOfFields` struct so currently have it in the `Fields.jl` file. These changes also negate the need for `biogeochemistry_rhs` since we don't need the advection to be with the biogeochemical tendencies (I think we origionally did this so we didn't need to change all of the kernels, but not we have todo that anyway), it does prevent us from using different advection schemes for the `biogeochemical_drift_velocity`.

To do:

- [x] Hydrostatic free surface kernels
- [x] Shallow water kernels
- [x] Tests
- [x] Benchmark
- [x] Fix particle advection


jagoosw:
I'm also not sure I've correctly interpreted how we combine background fields. From the docs the $\nabla\cdot(UC)$ term is neglected, but with this change I don't know if thats possible (or preferable) since now we just have `- div_𝐯u(i, j, k, grid, advection, total_velocities, total_velocities.u)` etc.

glwagner:
Awesome and nice work! Very helpful!

Thoughts:

* This is maybe more general than `Field` even... it's more like `SumOfArrays` (I know I suggested `SumOfFields`!) It can generalize to any array dimension by writing `getindex(soa::SumOfArrays, i...) = ...`. Maybe we should do that cause it could be handy in the future? Then it woul belong in `Utils` I think
* I don't think we should change formulation of the equations; we should only combine terms that can be combined. So `div_Uc(U1, c1) + div_Uc(U2, c1) + div_Uc(U1, c2)` should become `div_Uc(U1_U2, c1) + div_Uc(U1, c2)`. This gets upwinding right, even though we use different reconstructions for `c1` and `c2`.

src/Fields/Fields.jl
```diff
@@ -51,4 +53,13 @@ end
 
 include("set!.jl")
 
+struct SumOfFields{N, F}
+    fields :: F
+    SumOfFields{N}(fields...) where N = new{N, typeof(fields)}(fields)
+end
+
+@inline getindex(s::SumOfFields{1}, i, j, k) = @inbounds s.fields[1][i, j, k]
```
glwagner, line 61:
```suggestion
@propagate_inbounds getindex(s::SumOfFields{1}, i....) = getindex(s.fields[1], i...)
```

I believe that's what we want. Check out the docs for `Base.@propagate_inbounds`. (Note that we _really_ should use `propagate_inbounds` everywhere...)

src/Fields/Fields.jl
```diff
@@ -51,4 +53,13 @@ end
...
+@inline getindex(s::SumOfFields{2}, i, j, k) = @inbounds s.fields[1][i, j, k] + s.fields[2][i, j, k]
+@inline getindex(s::SumOfFields{3}, i, j, k) = @inbounds s.fields[1][i, j, k] + s.fields[2][i, j, k] + s.fields[3][i, j, k]
```
glwagner, line 63: You can also try to add

```julia
@propagate_inbounds function getindex(s::SumOfFields{N}, i...) where N
    first = getindex(SumOfFields(s.fields[1], s.fields[2], s.fields[3]), i...)
    last = getindex(SumOfFields(s.fields[4:N]...), i...)
    return first + last
end
```

To make this easier we might want the constructor

```julia
SumOfFields{N}(fields::Tuple) where N = new{N, typeof(fields)}(fields)
```

glwagner:
Big q is whether this works on the GPU 🤔

simone-silvestri:
Nice job. This should work on the GPU.
@glwagner, we can definitely separate `c1` and `c2` advection, just not `U1` and `U2` for the same `c`

glwagner:
> Nice job. This should work on the GPU. @glwagner, we can definitely separate `c1` and `c2` advection, just not `U1` and `U2` for the same `c`

Right, so the key is to get the upwinding direction right as I understand.

It is interesting that if you have two noisy fields which happen to sum to a smooth field, you could get more dissipation for "separate reconstruction" than "combined reconstruction". But that's just a thought experiment not something we think will happen... ?

jagoosw:
Tried making these changes now and looks like it works

jagoosw:
Another place I think we need to change is particle advection:
https://github.com/CliMA/Oceananigans.jl/blob/f3113c492f309846f026e0ac150dc8480f96e0f2/src/LagrangianParticleTracking/update_particle_properties.jl#L157-L161

Because currently background velocity fields don't get integrated which I think is wrong? At least in the case I was using them for I had to add the background velocities manually.

glwagner:
> Another place I think we need to change is particle advection:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/f3113c492f309846f026e0ac150dc8480f96e0f2/src/LagrangianParticleTracking/update_particle_properties.jl#L157-L161
> 
> Because currently background velocity fields don't get integrated which I think is wrong? At least in the case I was using them for I had to add the background velocities manually.

Totally agree particles should be advected with prognostic + background

jagoosw:
The example above worked on GPU

src/Models/NonhydrostaticModels/nonhydrostatic_tendency_kernel_functions.jl
```diff
@@ -119,8 +123,11 @@ pressure anomaly.
 
     model_fields = merge(velocities, tracers, auxiliary_fields)
 
-    return ( - div_𝐯v(i, j, k, grid, advection, velocities, velocities.v)
-             - div_𝐯v(i, j, k, grid, advection, background_fields.velocities, velocities.v)
+    total_velocities = (u = SumOfArrays{2}(velocities.u, background_fields.velocities.u),
```
simone-silvestri, line 126: wouldn't this give you a problem when `background_fields = NamedTuple()` (as in the default constructor)?

jagoosw: Thats how it is currently done, I think the defaults get expanded out to `(velocities = (u = ZeroField(grid), ...`

jagoosw:
Been on a flight so couldn't push but have worked out what was wrong (it was a typo).

Also, similarly to the particles, should both velocity components be being used in the Coriolis, Stokes drift, and immersed boundary shear (?) parts of the tendency kernels?

simone-silvestri:
ah nono, the additional velocities are just advective

jagoosw:
I've added a test for this in `test_dynamics` but not sure if that's the right place for it, or if we even want a test.

I haven't tried the test on GPU either so suspect it may fail.

jagoosw:
Is there a way around the order of `includ`ing files so that this works? If not, it probably isn't necessary to force that the `particles` slot in models is taken by a `Nothing` or `LagrangianParticles` as it is now? So we could get rid of the particles dependency in models and change the order to `include` particles after models?

I think I found a use case for removing the typing of the particles slot and overloading the update functions too but I can't remember what it was now.

jagoosw:
Ah this might be harder than I thought because timesteppers need the particle functions to be defined, I guess I could overcome that by

jagoosw:
Oops didn't mean todo that

jagoosw:
Oops wrong branch

test/test_field.jl
```diff
@@ -113,7 +113,7 @@ function run_field_interpolation_tests(FT, arch)
 
     # Maximum expected rounding error is the unit in last place of the maximum value
     # of f over the domain of the grid.
-    ε_max = f.(nodes(grid, (Face(), Face(), Face()), reshape=true)...) |> maximum |> eps
+    ε_max = (f.(nodes(grid, (Face(), Face(), Face()), reshape=true)...) |> maximum |> eps) * 10
```
glwagner, line 116:
```suggestion
    xf, yf, zf = nodes(grid, (Face(), Face(), Face()), reshape=true)
    f_max = maximum(f.(x, y, z))
    ε_max = eps(f_max)
    tolerance = 10 * ε_max
```

Let's make an effort...

jagoosw: I accidentally made this change on the wrong branch but will change it for #2912

glwagner:
> I've added a test for this in `test_dynamics` but not sure if that's the right place for it, or if we even want a test.
> 
> I haven't tried the test on GPU either so suspect it may fail.

I think a validation case would be appropriate. Ideally we would run some kind of convergence test but we don't have capacity for that now I think. The validation case could go here: https://github.com/CliMA/Oceananigans.jl/tree/main/validation/advection

glwagner:
> ~Is there a way around the order of `includ`ing files so that this works? If not, it probably isn't necessary to force that the `particles` slot in models is taken by a `Nothing` or `LagrangianParticles` as it is now? So we could get rid of the particles dependency in models and change the order to `include` particles after models?~
> 
> ~The only thing I can think of is that if a user passes something other than particles it might give a confusing error.~
> 
> ~I think I found a use case for removing the typing of the particles slot and overloading the update functions too but I can't remember what it was now.~
> 
> This won't work because particles need to already be defined for the timesteppers. My other solution would be to define a function `total_velocities` that is defined somewhere early on, and then define new methods for each model?

What is the problem you are trying to solve?

glwagner:
> Been on a flight so couldn't push but have worked out what was wrong (it was a typo).
> 
> Also, similarly to the particles, should both velocity components be being used in the Coriolis, Stokes drift, and immersed boundary shear (?) parts of the tendency kernels?

No. Don't change the RHS --- follow the docs equations.

src/LagrangianParticleTracking/update_particle_properties.jl
```diff
@@ -157,7 +157,7 @@ function update_particle_properties!(lagrangian_particles, model, Δt)
     advect_particles_kernel! = _advect_particles!(device(arch), workgroup, worksize)
 
     advect_particles_event = advect_particles_kernel!(lagrangian_particles.properties, lagrangian_particles.restitution, model.grid, Δt,
-                                                      datatuple(model.velocities),
+                                                      total_velocities(model),
```
glwagner, line 160: Why not pass in all the velocity fields explicitly?

test/test_dynamics.jl
```diff
@@ -392,6 +396,77 @@ function inertial_oscillations_work_with_rotation_in_different_axis(arch, FT)
...
+        A_hist[step, :] = model.tracers.A[1:grid.Nx, 1, 1]
+        time_step!(model, Δt/2)
+    end
+
+    A_truth = [A₀(x, 0, 0) for x in nodes(grid, Center(), Center(), Center())[1]]
+
+    @test all([all(A_hist[it, :] .≈ A_truth) for it in 1:1000])
+
+    return nothing
+end
```
glwagner, line 468: I think a validation case is enough

test/test_dynamics.jl
```diff
@@ -392,6 +396,77 @@ function inertial_oscillations_work_with_rotation_in_different_axis(arch, FT)
...
+    model = NonhydrostaticModel(; grid, 
+                                  tracers=:A, 
+                                  advection = UpwindBiased(),
+                                  background_fields = (; u))
+    
+    A₀(x, y, z) = x > 0.4 && x < 0.6 ? 1.0 : 0.0
+
+    set!(model, u = 1, A = A₀)
+
+    Δt = 0.2 / 1.0 * grid.Δxᶜᵃᵃ[1]
```
glwagner, line 417: Use functions to access grid metrics, don't write code that explicitly references grid properties (it doesn't generalize and will break if the names change)

test/test_dynamics.jl
```diff
@@ -392,6 +396,77 @@ function inertial_oscillations_work_with_rotation_in_different_axis(arch, FT)
...
+
+    A_hist = zeros(1000, grid.Nx)
```
glwagner, line 419: Use `Nx, Ny, Nz = size(grid)` to get `Nx`

test/test_dynamics.jl
```diff
@@ -392,6 +396,77 @@ function inertial_oscillations_work_with_rotation_in_different_axis(arch, FT)
...
+                                  tracers=:A, 
+                                  advection = UpwindBiased(),
+                                  biogeochemistry = SinkingParticles(),
+                                  background_fields = (; u))
+
+    set!(model, u = 2, A = A₀)
+
+    A_hist = zeros(1000, grid.Nx)
+
+    for step in 1:1000
```
glwagner, line 458: Why 1000? 10 might be better.

test/test_dynamics.jl
glwagner: The best is a little expensive and needs to be cleaned up. But we could also just write a validation case. We do test that background fields, etc work.

src/LagrangianParticleTracking/update_particle_properties.jl
jagoosw: As in pass velocities and background fields separately, or u/v/w separately? 

This was basically just the minimum change from what is currently implemented.

test/test_dynamics.jl
jagoosw: Okay will move

src/LagrangianParticleTracking/update_particle_properties.jl
glwagner: The answer might be that different models have different "total velocities", therefore it's simpler to dispatch at this level than farther out?

Generally when we make a change I think it's worth discussing whether the best choice is (1) extending the existing design or (2) implementing a new design (since the original code may not have been written with this concept in mind). Here `total_velocities` increase the burden for implementing new `AbstractModel` --- eg do we want to encode an expectation that all `AbstractModel` have the notion of "velocities"? This _could_ make sense as a design choice that all models wishing to participate in Lagrangian particle advection need to define `total_velocities` (or perhaps just "velocities"?) It's worth discussing.

jagoosw: Yeah that's why I had to implement it like this because hydrostatic models don't have background velocities like nonhydrostatic models. 

I see what you mean. While it's only being used here having the expectation that a model will have velocities if it's going to have particles seems reasonable? Up until now it's just been assumed that at `model.velocities` it would find a `u`, `v`, and `w` velocity components too (which I assume could be why shallow water models don't support particles).

Another thing to consider if we're coming up with a broader notion of models having velocities is that this doesn't include biogeochemical drift velocities so it more a specific momentum.

glwagner: one challenge is that I think as a user I would expect `total_velocities` to return an `AbstractField` or something that can be used in operations. But with `total_velocities` sometimes we get a `NamedTuple` of fields, sometimes we get a `NamedTuple` of `SumOfArrays`...

glwagner: Another solution would be to define an `advect_particles!(lagrangian_particles, model, Δt)` function that dispatches on the type of `model` ?

glwagner: Then each model decides how to advect particles (without having to make broader decisions on what constitutes "total velocities"

glwagner: Noticing also that this module is strange in that it defines a function of `model` before models even exist.

@jagoosw I think we should refactor this particle advection code to make more sense and avoid `total_velocities`. We can do that in this PR or another. If you want to tackle it here, can I commit to this PR?

glwagner: A consistent source of paranoia is how difficult it is going to be to split up this repo... we will very likely need to do that at some point, so I'm worried about circular dependencies and such (basically, all the functions that have to be defined at top-level --- thats bad / worrisome), which could make that difficult.

jagoosw: Okay yeah this makes sense and I like that idea. Happy to tackle it here and for you to commit to this PR

jagoosw: I believe that particles are defined before the models so that the type of particles can be forced to be `Union{Nothing, LagrangianParticles}` which probably isn't needed?

glwagner: I opened an issue to make this concern more clear: https://github.com/CliMA/Oceananigans.jl/issues/3048

jagoosw: Yeah this seems like a good idea. Having a generic type means users could create their own things that go in the particles slot instead (which I think could make the interface in OceanBioME nicer).

jagoosw:
Is there currently a problem with the GPU tests?

src/LagrangianParticleTracking/update_particle_properties.jl
jagoosw: I've set this up now, but not sure how to deal with the total velocities still? I guess I could define the function `total_velocities` in the `LagrangianParticles` dispatched on different model types now that it is after the models? And then its just a local funciton for the particles, but does mean more modifcaiton for usign particles in a different mode?

navidcy:
> Is there currently a problem with the GPU tests?

I think yes.

src/LagrangianParticleTracking/update_particle_properties.jl
glwagner: I have some ideas for how this should be implemented

glwagner: Been in a few meetings today, but I can have a chance at coding in a bit

src/Oceananigans.jl
```diff
@@ -153,6 +153,18 @@ Abstract supertype for models.
...
+    AbstractLagrangianParticles
+Abstract supertype for lagranigan particles.
+"""
+abstract type AbstractLagrangianParticles end
+
+update_particle_properties!(::Nothing, model, Δt) = nothing
+
+update_particle_properties!(model, Δt) = update_particle_properties!(model.particles, model, Δt)
+
+update_particle_properties!(particles::AbstractLagrangianParticles, model, Δt) = error("You need to define an `update_particle_properties!` method for $(typeof(particles))")
```
glwagner, line 166: We want to avoid new definitions as the top level, so we should remove this

glwagner: see https://github.com/CliMA/Oceananigans.jl/issues/3048

jagoosw: Should I move the fallbacks into the top of the particles file? I guess maybe in the model file maybe? I think we need fallbacks for the abstract type before the model definition.

jagoosw: Should I move the fallbacks into the top of the particles file? I guess maybe in the model file maybe? I think we need fallbacks for the abstract type before the model definition.

glwagner: I'm thinking we should move lagrangian particle stuff into `Models`. I don't want our work to conflict though, so let me know if you're ok with me working on this

jagoosw: Okay yeah makes sense, you go for it!

validation/advection/combined_velocity_fields.jl
```diff
@@ -0,0 +1,67 @@
...
+set!(model, u = 2, A = A₀)
+
+A_hist = zeros(1000, Nx)
+
+#=for step in 1:1000
+    A_hist[step, :] = model.tracers.A[1:Nx, 1, 1]
+    time_step!(model, Δt/2)
+end=#
+
+#@test all([all(A_hist[it, :] .≈ A_truth) for it in 1:1000])
```
glwagner, line 67: This needs to be cleaned up and I'd advocate for removing comments

glwagner:
Welp, I put a bit of effort into refacting the lagrangian particles implementation. It's now contained entirely with the module `Models`. @simone-silvestri you may want to take a look.

PS I think we should also move `TimeSteppers` into `Models` eventually.

src/Oceananigans.jl
glwagner: done, though we will see if tests pass.

jagoosw:
Should I have a go at fixing the tests @glwagner?

tomchor:
> Should I have a go at fixing the tests @glwagner?

I say go for it!

jagoosw:
Is there a way to make errors in kernels get thrown properly given the changes in #2924? Right now if something fails within a kernel no error is thrown (at  least how it is being used in the particle tracking in main).

simone-silvestri:
`KernelAbstraction.synchronize()` would be the same as the previous `wait(device(arch), event)` where the error would be thrown

jagoosw:
Another thing with the new kernels. I believe I've now got the particle advection kernels working, but unless some event causes a delay, for example printing some informaiton out, the `advect_lagrangian_particles!` event does not finish before the test tries to test that its been correctly done which causes it to fail. 

If I change it so that `advect_lagrangian_particles!` returns `advect_particles_kernel!(...`, and wrap `advect_particles_kernel!` in wait (i.e. `wait(advect_lagrangian_particles!(particles, model, Δt))`) then it does run correctly. 

This seems like a bigger issue with how kernel events are working? If other events aren't being finished in order this might cause other issues?

jagoosw:
> `KernelAbstraction.synchronize(device(arch))` would be the same as the previous `wait(device(arch), event)` where the error would be thrown

Thank you. I was about to comment this:
> Another thing with the new kernels. I believe I've now got the particle advection kernels working, but unless some event causes a delay, for example printing some informaiton out, the `advect_lagrangian_particles!` event does not finish before the test tries to test that its been correctly done which causes it to fail. 

>If I change it so that `advect_lagrangian_particles!` returns `advect_particles_kernel!(...`, and wrap `advect_particles_kernel!` in wait (i.e. `wait(advect_lagrangian_particles!(particles, model, Δt))`) then it does run correctly. 

>This seems like a bigger issue with how kernel events are working? If other events aren't being finished in order this might cause other issues?

But the new alternative should be to call `KernelAbstraction.synchronize(device(arch))` instead then?

simone-silvestri:
I thought everything should be synchronized correctly actually. 

If you are running on the GPU, the kernels run on a single stream at the moment (we still have to implement asynchronicity), so if you want to synchronize with the CPU you have to call `synchronize(device(arch))`, but that should happen automatically if you have a memory copy from device to host or vice-versa. 

On the CPU, the launch is synchronized 
see KernelAbstraction.jl:
(this is the kernel launch wrapped in `@sync`
```       
 @sync for tid in 1:Nthreads
     Threads.@spawn __thread_run(tid, len, rem, obj, ndrange, iterspace, args, dynamic)
 end
```
and, therefore `synchronize` does not do much
```
synchronize(::CPU) = nothing
```

where do you see this problem? Are you trying to look at GPU quantities from the CPU?

simone-silvestri:
Ah ok I see, you still have to merge main. Then synchronization problems will be fixed

jagoosw:
This was after merging main. If I don't include:
https://github.com/CliMA/Oceananigans.jl/blob/1db753ebeee2691205b7c0e6a03bba1ef9db4554/src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl#L131

The test for boundary restitution fails here:
https://github.com/CliMA/Oceananigans.jl/blob/1db753ebeee2691205b7c0e6a03bba1ef9db4554/test/test_lagrangian_particle_tracking.jl#L90-L95

simone-silvestri:
Ok, let me take a look. That synchronize should not be necessary

simone-silvestri:
I cannot reproduce the error on Tartarus. if I remove

https://github.com/CliMA/Oceananigans.jl/blob/1db753ebeee2691205b7c0e6a03bba1ef9db4554/src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl#L131

`inlcude("test/test_lagrangian_particle_tracinking.jl")` passes. Can you try removing that line and we look at the buildkite error? Maybe it has resolved itself

src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl
```diff
@@ -120,6 +120,7 @@ include("lagrangian_particle_advection.jl")
 
 update_lagrangian_particles!(model, Δt) = update_lagrangian_particles!(model.particles, model, Δt)
 update_lagrangian_particles!(::Nothing, model, Δt) = nothing
+update_lagrangian_particle_properties!(::ShallowWaterModel, args...) = nothing
```
glwagner, line 123:
```suggestion
update_lagrangian_particles!(::ShallowWaterModel, args...) = nothing
```

something is amiss here

src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl
```diff
@@ -120,7 +121,7 @@ include("lagrangian_particle_advection.jl")
 
 update_lagrangian_particles!(model, Δt) = update_lagrangian_particles!(model.particles, model, Δt)
 update_lagrangian_particles!(::Nothing, model, Δt) = nothing
-update_lagrangian_particle_properties!(::ShallowWaterModel, args...) = nothing
+update_lagrangian_particle_properties!(::ShallowWaterModel, Δt) = nothing
```
glwagner, line 124:
```suggestion
update_lagrangian_particle_properties!(model, Δt) = nothing
```

glwagner: Extend `update_lagrangian_particle_properties!` for the models that support particles

glwagner: Where does this get called too? I'm confused why its needed

jagoosw:
Not sure why the poisson solver test failed again but only on CPU

jagoosw:
The test passes locally so hopefully will work on this go, otherwise I think this is ready?

jagoosw:
Is there a way I can preview the docs without running all the examples locally?

navidcy:
> Is there a way I can preview the docs without running all the examples locally?

Yes. Comment them out both from the list of files to be literated and also from the docs pages below.

navidcy:
I see your issue. I'll fix it ;)

navidcy:
I believe your problem was that this line
https://github.com/CliMA/Oceananigans.jl/blob/79ed805201e28157856dd151abf100b002f38d24/docs/make.jl#L160
requires that all methods/types that are exported _have_ to be included in the docs.

navidcy:
I see that this PR moved LangangianParticles module inside Models. This is breaking change and we should bump minor release.

navidcy (commented):
bump minor release

validation/advection/combined_velocity_fields.jl
tomchor: @jagoosw clean this before merging as suggested by @glwagner ?

tomchor:
I don't mean to delay this PR, but I was wondering if we wanna also use `SumOfArrays` for background tracer fields. Id est, in `nonhydrostatic_tendency_kernel_functions.jl`, instead of having

```julia
    return ( - div_Uc(i, j, k, grid, advection, total_velocities, c)
             - div_Uc(i, j, k, grid, advection, velocities, background_fields_c)
             ...
```

we'd have

```julia
    return ( - div_Uc(i, j, k, grid, advection, total_velocities, total_c)
             ...
```

If we do that we'd also be including the advection of background tracers by background velocity fields in the tendency calculations, which [currently isn't included](https://clima.github.io/OceananigansDocumentation/stable/model_setup/background_fields/#Background-fields). Maybe even using `SumOfArrays` when calling other terms?

Are there reasons why we wouldn't wanna do that?

src/Models/NonhydrostaticModels/nonhydrostatic_tendency_kernel_functions.jl
```diff
@@ -231,12 +240,17 @@ velocity components, tracer fields, and precalculated diffusivities where applic
...
 
-    return ( - div_Uc(i, j, k, grid, advection, velocities, c)
-             - div_Uc(i, j, k, grid, advection, background_fields.velocities, c)
+    biogeochemical_velocities = biogeochemical_drift_velocity(biogeochemistry, val_tracer_name)
+
+    total_velocities = (u = SumOfArrays{3}(velocities.u, background_fields.velocities.u, biogeochemical_velocities.u),
+                        v = SumOfArrays{3}(velocities.v, background_fields.velocities.v, biogeochemical_velocities.v),
+                        w = SumOfArrays{3}(velocities.w, background_fields.velocities.w, biogeochemical_velocities.w))
+
+    return ( - div_Uc(i, j, k, grid, advection, total_velocities, c)
```
glwagner, line 249: @tomchor the total velocities are used to advect tracers here

glwagner:
> I don't mean to delay this PR, but I was wondering if we wanna also use `SumOfArrays` for background tracer fields. Id est, in `nonhydrostatic_tendency_kernel_functions.jl`, instead of having


> I don't mean to delay this PR, but I was wondering if we wanna also use `SumOfArrays` for background tracer fields. Id est, in `nonhydrostatic_tendency_kernel_functions.jl`, instead of having
> 
> ```julia
>     return ( - div_Uc(i, j, k, grid, advection, total_velocities, c)
>              - div_Uc(i, j, k, grid, advection, velocities, background_fields_c)
>              ...
> ```
> 
> we'd have
> 
> ```julia
>     return ( - div_Uc(i, j, k, grid, advection, total_velocities, total_c)
>              ...
> ```
> Are there reasons why we wouldn't wanna do that?

The main reason is that the purpose of background fields is to enable us to model a "perturbation" without having to model the background, as well.

The suggested change includes the background-background term, ie

```
U . grad C
```

but if we are interested in modeling the perturbation, then we would have subtracted the mean:

```
perturbation = u . grad c - U .grad C = (u' + U) . grad (c' + C) - U .grad C = (u' + U) .grad c' + u' .grad C.
```

Are there cases where you want to include the mean term `U . grad C`?

jagoosw:
> I see that this PR moved LangangianParticles module inside Models. This is breaking change and we should bump minor release.

Do I just need to change the version number in `Project.toml`?

navidcy:
Yeap. Crank up the number after the zero.

tomchor:
> For example, including mean terms in the equation set can prevent you from modeling uniform flow in a stratified tilted bottom boundary layer in a periodic domain, because U⋅∇C would not be periodic.

That's a good point. This is a good reason why we don't wanna include the mean-mean terms.

validation/advection/combined_velocity_fields.jl
```diff
@@ -0,0 +1,67 @@
...
+
+A_hist = zeros(1000, Nx)
+
+#=for step in 1:1000
+    A_hist[step, :] = model.tracers.A[1:Nx, 1, 1]
+    time_step!(model, Δt)
+end=#
+
+#@test all([all(A_hist[it, :] .≈ A_truth) for it in 1:1000])
+
```
tomchor, line 51:
```suggestion
```

validation/advection/combined_velocity_fields.jl
```diff
@@ -0,0 +1,67 @@
...
+Δt = 0.2 / 1.0 * Δx
+
+Nx, _, _ = size(grid)
+A_hist = zeros(1000, Nx)
+
+#=for step in 1:10
+    A_hist[step, :] = model.tracers.A[1:Nx, 1, 1]
+    time_step!(model, Δt)
+end=#
+
```
tomchor, line 31:
```suggestion
```

validation/advection/combined_velocity_fields.jl
```diff
@@ -0,0 +1,67 @@
+using Oceananigans
+using Oceananigans.Grids: xspacing
+
+# for `test_velocity_combination`
```
tomchor, line 4:
```suggestion
```

validation/advection/combined_velocity_fields.jl
```diff
@@ -0,0 +1,67 @@
...
+                              tracers=:A, 
+                              advection = UpwindBiased(),
+                              background_fields = (; u))
+
+A₀(x, y, z) = x > 0.4 && x < 0.6 ? 1.0 : 0.0
+
+set!(model, u = 1, A = A₀)
+
+Δx = xspacing(1, 1, 1, grid, Center(), Center(), Center())
+Δt = 0.2 / 1.0 * Δx
```
tomchor, line 22: Are these being used for anything atm?

navidcy:
Is there any syntax change need to be made in 

https://github.com/CliMA/Oceananigans.jl/blob/jsw/different-velocities/validation/lagrangian_particles/particles_in_convection.jl

I'm actually thinking now that if the name of the module didn't change, moving the module from `LagrangianParticleTracking` to `Models.LagrangianParticleTracking` might not imply any syntax change _given_ that the module name is exported.

@glwagner?

validation/advection/combined_velocity_fields.jl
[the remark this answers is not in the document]
glwagner: agree about cleaning this up.

glwagner:
> Is there any syntax change need to be made in
> 
> https://github.com/CliMA/Oceananigans.jl/blob/jsw/different-velocities/validation/lagrangian_particles/particles_in_convection.jl
> 
> I'm actually thinking now that if the name of the module didn't change, moving the module from `LagrangianParticleTracking` to `Models.LagrangianParticleTracking` might not imply any syntax change _given_ that the module name is exported.
> 
> @glwagner?

Does it run?

navidcy:
> > Is there any syntax change need to be made in
> > https://github.com/CliMA/Oceananigans.jl/blob/jsw/different-velocities/validation/lagrangian_particles/particles_in_convection.jl
> > I'm actually thinking now that if the name of the module didn't change, moving the module from `LagrangianParticleTracking` to `Models.LagrangianParticleTracking` might not imply any syntax change _given_ that the module name is exported.
> > @glwagner?
> 
> Does it run?

I see. The brute force approach. I'll try!

navidcy:
OK, it runs! So there was no breaking change. Sorry -- my bad. Fixed it. ;)

validation/advection/combined_velocity_fields.jl
```diff
@@ -24,11 +23,6 @@ set!(model, u = 1, A = A₀)
 Nx, _, _ = size(grid)
```
glwagner, line 20: `Nx = size(grid, 1)` would do...

validation/advection/combined_velocity_fields.jl
```diff
@@ -24,11 +23,6 @@ set!(model, u = 1, A = A₀)
...
 A_hist = zeros(1000, Nx)
```
glwagner, line 21: Why 1000? A comment might help future developers

validation/advection/combined_velocity_fields.jl
```diff
@@ -24,11 +23,6 @@ set!(model, u = 1, A = A₀)
...
 
-#=for step in 1:10
-    A_hist[step, :] = model.tracers.A[1:Nx, 1, 1]
-    time_step!(model, Δt)
-end=#
-
 A_truth = [A₀(x, 0, 0) for x in nodes(grid, Center(), Center(), Center())[1]]
```
glwagner, line 23:
```suggestion
A_truth = [A₀(x, 0, 0) for x in xnodes(grid, Center(), Center(), Center())]
```

validation/advection/combined_velocity_fields.jl
```diff
@@ -58,10 +45,3 @@ model = NonhydrostaticModel(; grid,
 set!(model, u = 2, A = A₀)
 
 A_hist = zeros(1000, Nx)
```
glwagner, line 42: Where is `A_hist` used?

src/Fields/Fields.jl
```diff
@@ -50,5 +50,4 @@ function field(loc, f::Field, grid)
 end
 
 include("set!.jl")
-
```
glwagner, line 53: Can you add this back

src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
```diff
@@ -99,6 +100,8 @@ Return a flattened `NamedTuple` of the prognostic fields associated with `Hydros
 displacement(free_surface) = free_surface.η
 displacement(::Nothing) = nothing
 
+update_lagrangian_particles!(model::HydrostaticFreeSurfaceModel, Δt) = update_lagrangian_particles!(model.particles, model, Δt)
```
glwagner, line 103:
```suggestion
# Unpack model.particles to update particle properties. See Models/LagrangianParticleTracking/LagrangianParticleTracking.jl
update_lagrangian_particles!(model::HydrostaticFreeSurfaceModel, Δt) = update_lagrangian_particles!(model.particles, model, Δt)
```

src/Models/LagrangianParticleTracking/update_lagrangian_particle_properties.jl
```diff
@@ -0,0 +1,36 @@
...
+    p = @index(Global)
+    @inbounds begin
+        x = particles.x[p]
+        y = particles.y[p]
+        z = particles.z[p]
+        particle_property[p] = interpolate(field, ℓx, ℓy, ℓz, grid, x, y, z)
+    end
+end
+
+function update_lagrangian_particle_properties!(particles, model, Δt)
```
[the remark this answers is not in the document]
glwagner: Should this be 

```julia
update_lagrangian_particle_properties!(particles, model::HydrostaticFreeSurfaceModel, Δt)
```

?

Also what is the difference between "update Lagrangian particles" and "update Lagrangian particle properties" ? (Maybe it's a sign we can do better given that I can't even remember...) Maybe clearer naming scheme could help?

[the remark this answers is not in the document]
glwagner: As for extending this method specifically... I'm not sure what the proper pattern is here. We assume that `model.grid` exists. Perhaps we can write `model::AbstractModel`, thus introducing the assumption that all `AbstractModel`s have `.grid`. It's ok if we leave it for this PR but we should clarify the interface eventually...

src/TimeSteppers/TimeSteppers.jl
```diff
@@ -48,6 +47,10 @@ function calculate_tendencies! end
 calculate_pressure_correction!(model, Δt) = nothing
 pressure_correct_velocities!(model, Δt) = nothing
 
+# Interface for time-stepping Lagrangian particles
+abstract type AbstractLagrangianParticles end
+update_lagrangian_particles!(model, Δt) = nothing
```
glwagner, line 52: Probably `TimeSteppers` also belongs within `Models` (not for this PR...)

test/test_field.jl
glwagner, line 116: Can we put in a comment about why "`10`"?

test/test_lagrangian_particle_tracking.jl
```diff
@@ -128,14 +128,16 @@ function run_simple_particle_tracking_tests(arch, timestepper; vertically_stretc
...
     sim.output_writers[:particles_jld2] =
         JLD2OutputWriter(model, (; particles=model.particles),
-                         filename=jld2_filepath, schedule=IterationInterval(1))
+                         filename=jld2_filepath, schedule=IterationInterval(1),
+                         overwrite_existing = true)
 
     nc_filepath = "test_particles.nc"
     sim.output_writers[:particles_nc] =
-        NetCDFOutputWriter(model, model.particles, filename=nc_filepath, schedule=IterationInterval(1))
+        NetCDFOutputWriter(model, model.particles, filename=nc_filepath, schedule=IterationInterval(1), overwrite_existing = true)
```
glwagner, line 136:
```suggestion
        NetCDFOutputWriter(model, model.particles, filename=nc_filepath, schedule=IterationInterval(1))
```

test/test_lagrangian_particle_tracking.jl
```diff
@@ -260,10 +262,10 @@ end
...
 @testset "Lagrangian particle tracking" begin
     for arch in archs, timestepper in (:QuasiAdamsBashforth2, :RungeKutta3)
-        @info "  Testing uniform grid Lagrangian particle tacking [$(typeof(arch)), $timestepper]..."
+        @info "  Testing uniform grid Lagrangian particle tracking [$(typeof(arch)), $timestepper]..."
         run_simple_particle_tracking_tests(arch, timestepper; vertically_stretched=false)
 
-        @info "  Testing stretched grid Lagrangian particle tacking [$(typeof(arch)), $timestepper]..."
-        run_simple_particle_tracking_tests(arch, timestepper; vertically_stretched=true)
+        #@info "  Testing stretched grid Lagrangian particle tracking [$(typeof(arch)), $timestepper]..."
+        #run_simple_particle_tracking_tests(arch, timestepper; vertically_stretched=true)
```
glwagner, line 269:
```suggestion
        @info "  Testing stretched grid Lagrangian particle tracking [$(typeof(arch)), $timestepper]..."
        run_simple_particle_tracking_tests(arch, timestepper; vertically_stretched=true)
```

?

validation/advection/combined_velocity_fields.jl
glwagner, line 1: This file doesn't do much so I'm not sure we should keep it. We're going to have to refactor (and probably delete) a lot of the validation tests sooner or later anyways...

glwagner (approved):

glwagner:
> Maybe it did need some of the overwrite existing because it writes different configurations to the same file
> 
> Update: passes locally if I add them back

hmm ok that's ok. Could also just manually delete the existing files.

tomchor (approved):
Nice work!

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -0,0 +1,172 @@
...
+        particles.y[p] = y⁺ 
+        particles.z[p] = z⁺ 
+    end
+end
+
+total_velocities(model::NonhydrostaticModel) = (u = SumOfArrays{2}(model.velocities.u, model.background_fields.velocities.u),
+                                                v = SumOfArrays{2}(model.velocities.v, model.background_fields.velocities.v),
+                                                w = SumOfArrays{2}(model.velocities.w, model.background_fields.velocities.w))
+
+total_velocities(model::HydrostaticFreeSurfaceModel) = model.velocities
```
simone-silvestri, line 159: can we remove these from here?

jagoosw:
I'm struggling to fix this test failing and am not sure why its not passing the combined velocities correctly

src/Utils/sum_of_arrays.jl
```diff
@@ -0,0 +1,26 @@
...
+    first = getindex(SumOfArrays{3}(s.arrays[1], s.arrays[2], s.arrays[3]), i...)
+    last = getindex(SumOfArrays{N - 3}(s.arrays[4:N]...), i...)
+    return first + last
+end
+
+@propagate_inbounds getindex(s::SumOfArrays{1}, i...) = getindex(s.arrays[1], i...)
+@propagate_inbounds getindex(s::SumOfArrays{2}, i...) = getindex(s.arrays[1], i...) + getindex(s.arrays[2], i...)
+@propagate_inbounds getindex(s::SumOfArrays{3}, i...) = getindex(s.arrays[1], i...) + getindex(s.arrays[2], i...) + getindex(s.arrays[3], i...)
+
+@propagate_inbounds adapt_structure(sum::SumOfArrays{N}, i...) where N = SumOfArrays{N}((adapt_structure(to, array) for array in sum.arrays)...)
```
simone-silvestri, line 26:
```suggestion
Adapt.adapt_structure(to, sum::SumOfArrays{N}) where N = SumOfArrays{N}((Adapt.adapt(to, array) for array in sum.arrays)...)
```

simone-silvestri: this should solve the tests

tomchor:
> Please could someone try rerunning the failed test since it shouldn't have changed?

It seems it passed already

jagoosw:
> > Please could someone try rerunning the failed test since it shouldn't have changed?
> 
> It seems it passed already

I think someone else must have restarted it, thanks!

jagoosw:
Ready to merge now?

tomchor:
> Ready to merge now?

Fine by me. But if this is a breaking change (I think it is, right?) then please bump the minor version beforehand?

navidcy:
woohooo

ankitsb:
Great job guys. @jagoosw have you been able to run this (https://github.com/CliMA/Oceananigans.jl/blob/19dac0b8f6b621057e250ecf0510b3c4c91915e9/validation/lagrangian_particles/particles_in_convection.jl#L30) on GPU?

jagoosw:
> Great job guys. @jagoosw have you been able to run this (https://github.com/CliMA/Oceananigans.jl/blob/19dac0b8f6b621057e250ecf0510b3c4c91915e9/validation/lagrangian_particles/particles_in_convection.jl#L30) on GPU?
> 
> 

I haven't run that case on GPU but would have thought it will run if you change the grid architecture and adapt the initial particle positions (which is described in the simulation tips part of the docs). I can have a look later to see if I can run it.

ankitsb:
> 

Thanks for your response.
It does work on `GPU()` architecture. Cheers!!

PR #3028, opened 2023-03-27
Brings back SWE regression

navidcy:
I was under the impression that it was brought back long ago...


francispoulin (approved):
Great that it passes.  Any idea what has changed since when it failed?

PR #3035, opened 2023-03-29
Updates `Docs/Buoyancy models and equation of state` + update SeawaterPolynomials to v0.3.2

jbisits:
This PR fixes the rendering of a LaTeX string and changes `eos = RoquetSeawaterPolynomial(:Freezing)` -> `eos = RoquetEquationOfState(:Freezing)` in the documentation for `buoyancy models and equation of state.md`. The reason for this change is that with the `eos` set as a `RoquetSeawaterPolynomial`,

```julia
julia> using SeawaterPolynomials.SecondOrderSeawaterPolynomials

julia> eos = RoquetSeawaterPolynomial(:Freezing)
0.7718 Sᴬ - 0.0491 Θ + 0.0 Θ² - 2.5681e-5 Θ Z + 0.0 Sᴬ² - 0.005027 Sᴬ Z + 0.0 Sᴬ Θ

julia> buoyancy = SeawaterBuoyancy(equation_of_state=eos)
SeawaterBuoyancy{Float64}:
├── gravitational_acceleration: 9.80665
└── equation of state: SecondOrderSeawaterPolynomial{Float64}
```

you get an error (or at least I did, attached at the bottom of this PR) when setting `model(;  ... buoyancy)`. Instead if we have the `eos` set as a `RoquetEquationOfState`,

```julia
julia> eos = RoquetEquationOfState(:Freezing)
BoussinesqEquationOfState{Float64}:
    ├── seawater_polynomial: SecondOrderSeawaterPolynomial{Float64}
    └── reference_density: 1024.6
    
julia> eos.seawater_polynomial # the density anomaly
ρ' = 0.7718 Sᴬ - 0.0491 Θ - 0.005027 Θ² - 2.5681e-5 Θ Z + 0.0 Sᴬ² + 0.0 Sᴬ Z + 0.0 Sᴬ Θ

julia> buoyancy = SeawaterBuoyancy(equation_of_state=eos)
SeawaterBuoyancy{Float64}:
├── gravitational_acceleration: 9.80665
└── equation of state: BoussinesqEquationOfState{Float64}
```

we can then set this as the `buoyancy` in the model, `model(; buoyancy)`, without an error.

As of SeawaterPolynomials.jl v0.3.2 another idealised nonlinear eos has been added so the number available has been updated in `buoyancy models and equation of state.md`to reflect this and the compat entry has been updated.

## Error from setting `buoyancy = SeawaterBuoyancy(equation_of_state= RoquetSeawaterPolynomial(:Freezing))`

```julia
        nested task error: MethodError: no method matching ρ′(::Float64, ::Float64, ::Float64, ::SecondOrderSeawaterPolynomial{Float64})
        Closest candidates are:
          ρ′(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any) at ~/.julia/packages/Oceananigans/Zg9Bd/src/BuoyancyModels/nonlinear_equation_of_state.jl:27
          ρ′(::Any, ::Any, ::Any, ::SeawaterPolynomials.BoussinesqEquationOfState{<:SecondOrderSeawaterPolynomial}) at ~/.julia/packages/SeawaterPolynomials/ig7CR/src/SecondOrderSeawaterPolynomials.jl:65
          ρ′(::Any, ::Any, ::Any, ::SeawaterPolynomials.BoussinesqEquationOfState{<:SeawaterPolynomials.TEOS10.TEOS10SeawaterPolynomial}) at ~/.julia/packages/SeawaterPolynomials/ig7CR/src/TEOS10.jl:186
```


Project.toml
```diff
@@ -60,7 +60,7 @@ OrderedCollections = "1.1"
 PencilArrays = "0.16, 0.17"
 PencilFFTs = "0.13.5, 0.14"
 Rotations = "1.0"
-SeawaterPolynomials = "0.3"
+SeawaterPolynomials = "0.3.2"
```
navidcy, line 63: we need to change the Manifest also.

navidcy: Only changing the Project compat entry won't force the package to update. So doctests would probably fail until you also actually update the package and push the new Manifest...

jbisits: Ah, I see! I have not done this before but I assume you can just edit the `Manifest.toml` entry 

```
[[deps.SeawaterPolynomials]]
git-tree-sha1 = "3e4e6c809e96ddcc0077bdb6944a1abb53fc382b"
uuid = "d496a93d-167e-4197-9f49-d3af4ff8fe40"
version = "0.3.0"
```

where version = "0.3.0" -> version = "0.3.2"?

jbisits: > Only changing the Project compat entry won't force the package to update. So doctests would probably fail until you also actually update the package and push the new Manifest...

Ok, I did think Manifest was usually off limits. Will do as you suggest and update the package, then push.

navidcy: No don't manually edit the manifest! Just update the SeawaterPolynomials locally. This would alter the manifest. ;)

navidcy: It may be that several other dependencies need to be updated.

jbisits: > No don't manually edit the manifest! Just update the SeawaterPolynomials locally. This would alter the manifest. ;)

Don't worry, the only thing in know about the Manifest is not touch the Manifest so would only have done this if I was explicitly told to do so.

jbisits:
> It may be that several other dependencies need to be updated.

Yep, a few others updated when I ran 

```julia
(Oceananigans) pkg> update SeawaterPolynomials
```

navidcy:
So some of the doctests fail. You can see, e.g., here
https://buildkite.com/clima/oceananigans/builds/10843#01872c04-2d2f-4dbb-94b3-925c0252a815
Seems like some of the updates introduced some more methods for `cos` :)
So if you change that 26 --> 30 it should work.

navidcy:
> Hopefully not too many more of those...

It could be!
you can run the docs locally if you like to check, following instructions [from the Docs](https://clima.github.io/OceananigansDocumentation/stable/contributing/#Documentation). If you do that then comment out the literate examples, e.g.,

https://github.com/jbisits/Oceananigans.jl/blob/601970be23ff53ce49d31857039b55c36d4e318e/docs/make.jl#L32-L42

and

https://github.com/jbisits/Oceananigans.jl/blob/601970be23ff53ce49d31857039b55c36d4e318e/docs/make.jl#L55-65

otherwise building the docs would take a while!

jbisits:
> > Hopefully not too many more of those...
> 
> It could be! you can run the docs locally if you like to check, following instructions [from the Docs](https://clima.github.io/OceananigansDocumentation/stable/contributing/#Documentation). If you do that then comment out the literate examples, e.g.,
> 
> https://github.com/jbisits/Oceananigans.jl/blob/601970be23ff53ce49d31857039b55c36d4e318e/docs/make.jl#L32-L42
> 
> and
> 
> https://github.com/jbisits/Oceananigans.jl/blob/601970be23ff53ce49d31857039b55c36d4e318e/docs/make.jl#L55-L65
> 
> otherwise building the docs would take a while!

True!!! If this one fails I will build locally and correct.

navidcy (approved):

navidcy:
(@jbisits, we have some issues with the CI. But as soon as tests pass I'll merge this.)

src/AbstractOperations/conditional_operations.jl
```diff
@@ -61,7 +61,7 @@ julia> f(i, j, k, grid, c) = i < 2; d = condition_operand(cos, c, f, 10)
 ConditionalOperation at (Center, Center, Center)
 ├── operand: 2×1×1 Field{Center, Center, Center} on RectilinearGrid on CPU
 ├── grid: 2×1×1 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
-├── func: cos (generic function with 30 methods)
+├── func: cos (generic function with 26 methods)
```
glwagner: should probably use `showmethods=false` for `prettysummary` to avoid this pain

glwagner: but we can fix it later

navidcy: Yes, not in this PR

PR #3037, opened 2023-03-29
Remove Tullio dependency; take #2

simone-silvestri:
Take two on PR #2252
i.e. use AbstractOperations to calculate the CFL instead of Tullio.jl


glwagner:
@navidcy and I implemented this CFL operation a bit differently but I can't remember where ... maybe @navidcy can point you to it

navidcy:
https://github.com/CliMA/Oceananigans.jl/blob/bcc34f07b3f949ea6fb34c7814f4b856d24924c2/src/Advection/cell_advection_timescale.jl#L1-L18

simone-silvestri:
Ok, it's doing the same thing, I'll just remove Tullio and use this for the CFL

simone-silvestri:
I am going to change a bit the definition to be in line with what we had before, so instead of 
```julia
@inline function cell_advection_timescaleᶜᶜᶜ(i, j, k, grid, u, v, w)
    Δx = Δxᶠᶜᶜ(i, j, k, grid)
    Δy = Δyᶜᶠᶜ(i, j, k, grid)
    Δz = Δzᶜᶜᶠ(i, j, k, grid)

    return @inbounds min(Δx / abs(u[i, j, k]),
                         Δy / abs(v[i, j, k]),
                         Δz / abs(w[i, j, k])) 
end
```
take all 3 dimensions
```
@inline function cell_advection_timescaleᶜᶜᶜ(i, j, k, grid, u, v, w)
    Δx = Δxᶠᶜᶜ(i, j, k, grid)
    Δy = Δyᶜᶠᶜ(i, j, k, grid)
    Δz = Δzᶜᶜᶠ(i, j, k, grid)

    return @inbounds 1 / (abs(u[i, j, k]) / Δx +
                                          abs(v[i, j, k]) / Δy +
                                         abs(w[i, j, k]) / Δz) 
end
```
it is a small difference but it is a more conservative definition (and more used in n-dimensional cases)

navidcy:
I guess there is diffusive cfl...?

src/Advection/cell_advection_timescale.jl
```diff
@@ -12,8 +12,7 @@ end
     Δy = Δyᶜᶠᶜ(i, j, k, grid)
     Δz = Δzᶜᶜᶠ(i, j, k, grid)
 
-    return @inbounds min(Δx / abs(u[i, j, k]),
-                         Δy / abs(v[i, j, k]),
-                         Δz / abs(w[i, j, k])) 
+    return @inbounds 1 / (abs(u[i, j, k]) / Δx +
+                          abs(v[i, j, k]) / Δy +
+                          abs(w[i, j, k]) / Δz)
```
navidcy, line 17: is this the proper way to compute CFL? I'll trust you on that... That's like the geometric mean of the timescales in each direction?

simone-silvestri: Imagine a 2D situation, we start from
$$\frac{\partial c}{\partial t} = u\frac{\partial c}{\partial x} + v\frac{\partial c}{\partial y}$$
We have advection in x and y, then we can approximate the CFL condition as
$$|u|\frac{\Delta t}{\Delta x} + |v|\frac{\Delta t}{\Delta y} < CFL (\text{ maximum 1 })$$
This means that 
$$\Delta t / \tau < CFL$$
where 
$$\tau = \left(\frac{|u|}{\Delta x} + \frac{|v|}{\Delta y} \right)^{-1}$$
The extension to 3D includes also the advection in z

navidcy: Sounds good! Why don't we include all this in the docstring of the `cell_advection`?

simone-silvestri:
Ok, I was a bit unsure which CFL definition is the best, so I checked the 2D stability and I got a slightly better proof. 
Let's start with the discretized version of the 2D advection equation (forward euler and 2nd order centered finite volume on a staggered regular cartesian grid)
$$\frac{c^{n+1} - c^n}{\Delta t} = u \left( \frac{c^n_i - c^n_{i-1}}{\Delta x}\right) + v \left( \frac{c^n_j - c^n_{j-1}}{\Delta y}\right)$$ 
We can assume that c is a two-dimensional wave depending on an x wavenumber $\kappa$ and a y wavenumber $\mathcal{l}$ and a time-dependent amplification factor $\xi(t)$, then
$c^n_{ij} = \xi^n \exp{(- \Im \kappa \cdot i \Delta x - \Im \mathcal{l}  \cdot j \Delta y)}$. (because of overlap with the x-index $i$, I defined $\Im$ as the imaginary number $\Im = \sqrt{-1}$)
Substituting this definition of $c$ and dividing through by $c^n_{ij}$ we get
$$\frac{\xi^{n+1} / \xi^n - 1}{\Delta t} = u \left(\frac{1 - \exp{(\Im \kappa \Delta x)}}{\Delta x}\right) +  v \left(\frac{1 -  \exp{(\Im  \mathcal{l} \Delta y)}}{\Im \Delta y}\right)$$
we can make use of $\exp{\Im \theta} = \cos{\theta} - \Im \sin{\theta}$ and rewrite a bit:
$$\frac{\xi^{n+1}}{\xi^n} = 1 + \Delta t \cdot \left[  \frac{u}{\Delta x} \left( 1 - \cos{\kappa \Delta x} + \Im \sin{\kappa\Delta x}\right) +  \frac{v}{\Delta y} \left( 1 - \cos{\mathcal{l} \Delta y} + \Im \sin{\mathcal{l} \Delta y} \right) \right]$$
Now, to ensure stability, the real part of $\xi^{n+1} / \xi^n$ should be bounded, so we have to ensure that
$$\Re \left| \frac{\xi^{n+1}}{\xi^n} \right| < 1$$
This yields
$$-2 < \Delta t \cdot \left[  \frac{u}{\Delta x} \left( 1 - \cos{\kappa \Delta x} \right) +  \frac{v}{\Delta y} \left( 1 - \cos{\mathcal{l} \Delta y} \right) \right] < 0$$
The right inequality does not limit $\Delta t$, but the left does:
$$\Delta t \cdot \left[  \frac{u}{\Delta x} \left( 1 - \cos{\kappa \Delta x} \right) +  \frac{v}{\Delta y} \left( 1 - \cos{\mathcal{l} \Delta y} \right) \right] < 2$$
The worst-case scenario occurs when both cosines evaluate to -1 for which we have
$$\Delta t \cdot \left(  \frac{u}{\Delta x}  +  \frac{v}{\Delta y} \right)< 1$$
Since the direction is arbitrary you can substitute $u$ and $v$ with their absolute values and you get
$$\Delta t < 1 / \left(  \frac{|u|}{\Delta x}  +  \frac{|v|}{\Delta y} \right)$$

src/Advection/cell_advection_timescale.jl
glwagner, line 17:
```suggestion

    ω = @inbounds abs(u[i, j, k]) / Δx + abs(v[i, j, k]) / Δy + abs(w[i, j, k]) / Δz
    return 1 / ω
```

or something like that

navidcy (approved):

src/Diagnostics/cfl.jl
```diff
@@ -83,31 +83,6 @@ DiffusiveCFL(Δt) = CFL(Δt, cell_diffusion_timescale)
...
-    min_timescale = minimum(
-        @tullio (min) timescale[k] := 1 / (  abs(u[i, j, k]) / Δxᶠᶜᶜ(i, j, k, grid)
-                                           + abs(v[i, j, k]) / Δyᶜᶠᶜ(i, j, k, grid)
-                                           + abs(w[i, j, k]) / Δzᶜᶜᶠ(i, j, k, grid))
-    )
+using Oceananigans.Advection: cell_advection_timescale
 
-    return min_timescale
-end
+accurate_cell_advection_timescale(model) = cell_advection_timescale(model.grid, model.velocities)
```
glwagner, line 88: Let's just delete "accurate" cell advection timescale. It's meaningless.

glwagner, line 88:
```suggestion
```

navidcy: Agree!

src/Diagnostics/cfl.jl
```diff
@@ -85,4 +85,3 @@ DiffusiveCFL(Δt) = CFL(Δt, cell_diffusion_timescale)
 
 using Oceananigans.Advection: cell_advection_timescale
```
navidcy, line 86:
```suggestion
```

navidcy: @glwagner won't we need a method for `cell_advection_timescale(model)`?

navidcy: There are a bunch of `accurate_cell_advection_timescale` calls in tests and validations scripts btw we should cleanup.

glwagner: delete them all? a great chance to delete lines.

navidcy:
merge?

PR #3042, opened 2023-03-30
Rework CI to use buildkite plugin

vchuravy:


navidcy (approved):
This looks good! Only downside that we loose the emojis for each step -- we can live without them.

@vchuravy any particular advantage to this over the previous version?

navidcy:
perhaps let's wait to wait from @glwagner and @simone-silvestri before we merge?

simone-silvestri:
This seems like a way forward, but I would wait for #2924 (that seems to be our bottleneck) to pass before changing CI

vchuravy:
@glwagner is this something you would want? Otherwise I will close this PR.

navidcy:
I think this might be helpful.

navidcy:
OK, I brought the emojis back!

navidcy (approved):

navidcy:
@vchuravy if/when the tests pass merge!

navidcy:
Something's wrong; all jobs seem to have the same error (those which failed and those which passed)?

navidcy:
@vchuravy could you have a look here? this seems to run, but e.g., were there any tests actually tested in https://buildkite.com/clima/oceananigans/builds/11987#018912fd-4e55-4160-87ce-0e7195b271aa ?

vchuravy:
Yeah adding the Unicode symbols broke the test, since the tests group detection doesn't recognize them

navidcy:
Oh damn, I’ll remove them then…

navidcy:
@vchuravy how about now? still there is some weird TOML parser error...

glwagner:
😢

PR #2979, opened 2023-03-16
`node(s)` return native coordinates for grids + `x/y/znode(s)` -> distances & `λ/φnode(s)` -> degrees

navidcy:
This PR makes `node` and `nodes` return the native coordinates of the grid. It changes the behavior of `x/ynode(s)`. Now `x/ynode(s)` _always_  returns `x/y` distances. The PR also introduces `λ/φnode(s)` for curvilinear grids.

This way, now `node(s)` behaviour didn't change. But now they return:

- `(xnode(s), ynode(s), znode(s))` for RectilinearGrid
- `(λnode(s), φnode(s), znode(s))` for CurvilinearGrids

Note that still there are `x/y/znode(s)` methods _for all grids_ and those methods actually return distances (in meters). Before this PR when `ynode` called on, e.g., a LatitudeLongitude grid would return latitude in degrees and this was tad confusing.

Closes #2978

PR #3045, opened 2023-03-31
Gather constants in `Constants` module

### LINKED ISSUE #2981: Constants are sometimes defined twice
The grids define their own `R_Earth`.
https://github.com/CliMA/Oceananigans.jl/blob/e881faee3de5f17cb382cd61e2b43c2069cadde7/src/Grids/latitude_longitude_grid.jl#L3

The `Coriolis` module defines another one...

Perhaps we should have a module `Constants`, add everything in there and load it first? This potentially would come handy when we try to couple?

navidcy:
Prior of these PR constants (like Earth's radius) were multiply defined. Now all constants are gathered in a module loaded just after units.

Closes #2981

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.80.0"
+version = "0.81.0"
 
 [deps]
 AMGX = "c963dde9-0319-47f5-bf0c-b07d3c80ffa6"
--- a/benchmark/benchmark_implicit_surface_solvers.jl
+++ b/benchmark/benchmark_implicit_surface_solvers.jl
@@ -1,7 +1,8 @@
 using Oceananigans
+using Oceananigans.Constants: R_Earth
 using Oceananigans.Units
 using Oceananigans.Advection: VelocityStencil
-using Oceananigans.Coriolis: HydrostaticSphericalCoriolis, R_Earth
+using Oceananigans.Coriolis: HydrostaticSphericalCoriolis
 using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBottom
 using Oceananigans.Models.HydrostaticFreeSurfaceModels: FFTImplicitFreeSurfaceSolver, MGImplicitFreeSurfaceSolver, finalize_solver!
 using Oceananigans.Solvers: initialize_AMGX, finalize_AMGX
--- a/docs/src/appendix/library.md
+++ b/docs/src/appendix/library.md
@@ -44,6 +44,13 @@ Modules = [Oceananigans.BuoyancyModels]
 Private = false
 ```
 
+## Constants
+
+```@autodocs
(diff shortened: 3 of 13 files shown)


navidcy:
This is a breaking change for some scripts that load, e.g., Earth's gravity from the `BuoyancyModels` module.

Let's aim to release a new version when both this PR and #2979 are merged.

glwagner:
Is this a good idea?

navidcy:
Apparently I think it's a good idea in order to make a PR :)

navidcy:
But I'm not sure which part are you referring as "good idea or not". The fact that they belong to a different module? Perhaps we could have these constants loaded/exported with Oceananigans main module. Is this what you are thinking?

glwagner:
We used to have the philosophy that constants should only be defined inside constructors, ie there should be no "global physical constants" in Oceananigans. Users should set them if they need that.

glwagner:
> Also I found it quite cumbersome in scripts when I wanted to load these constants I had to load one from `Grids` and one from `Coriolis` and one from `BuoyancyModels`.... That's a bit counterintuitive from a user's perspective.

Shouldn't you be defining and setting constants in the scripts if you are doing that?

navidcy:
Hm... OK... I wasn't aware of this philosophy.

You are saying that since everywhere users can construct types with prescribed constants there is no need for them to use the "globally-defined Oceananigans constants". OK right!

Somehow, though, we can't have every module redefining `R_Earth` and using it as default, right?

glwagner:
I guess there are two questions here; one regards user interface and the other regards how we determine defaults internally.

It's worth pointing out from the outset that we may have too many defaults right now; we should try to tackle that.

On user interfaces, we haven't converged on a way to set parameters like "radius of the Earth" and "gravitational acceleration". One philosophy could be to never use defaults. Unfortunately that is not very friendly / inconvenient, but on the upside it would help users ensure that these constants are always consistent.

I guess we have to decide on an approach for the user interface first; then the internal code will follow.

glwagner:
> Somehow, though, we can't have every module redefining `R_Earth` and using it as default, right?

I think you're right that if we are going to have a default value for this, then it should only be defined once.

But if we are going to go ahead with a "master module" approach, then we need to implement tests. 

What we should avoid is people trying to change constants in the master module as a way of setting parameters.

navidcy:
OK, I'll close this for now then and leave the issue open to remind us to heal the duplicate `R_Earth` that appear in Oceananigans...

I like the approach that users should set their constants depending on whatever choices they like. E.g., simulating an ocean in Europe or a fish tank on the top of Everest.

glwagner:
> OK, I'll close this for now then and leave the issue open to remind us to heal the duplicate `R_Earth` that appear in Oceananigans...
> 
> I like the approach that users should set their constants depending on whatever choices they like. E.g., simulating an ocean in Europe or a fish tank on the top of Everest.

Can you just get rid of the constant called "R_Earth"?

navidcy:
You mean remove the duplicate re-definition of `R_Earth`? I can do that.

glwagner:
> You mean remove the duplicate re-definition of `R_Earth`? I can do that.

Remove the default in offending cases?

PR #3046, opened 2023-03-31
Avoid redefining `R_Earth` in the `Coriolis` module

### LINKED ISSUE #2981: Constants are sometimes defined twice

navidcy:
This PR comes after the discussion in #3045.

Closes #2981.


navidcy:
(it was much easier to open a new PR rather than reverting changes...)

glwagner (approved):

Issue #3049, opened 2023-04-04
Move `ShallowWaterModel` to a different package?


navidcy:
Should we move `ShallowWaterModel` to its own package which would depend on Oceananigans?

There is an outstanding issue https://github.com/CliMA/Oceananigans.jl/issues/2922
We'd either need to sort it out. I believe to figure it out we need to plot/print out the tendencies from the first iteration for a version that worked and the first version that it didn't and trace back the issue. 

Perhaps it would make sense if this was a different model?

glwagner:
I think we should delete the regression test and put a warning in the model constructor "The ShallowWaterModel has not been validated".

It's a shame we've put so much time into the regression test as it is. I think the model likely has a few bugs to be ironed out. I'm also not convinced that we have all the numerical methods (ie vanishing layers, positive preserving advection) that would be needed to advertise the model as "production-ready"). One could argue that the regression test is misleading in that it gives the false impression that the model is in a state that shouldn't change.

glwagner:
For example, these lines multiply `h` (a field at ccc) with objects at `fcc` and `cfc` respectively

https://github.com/CliMA/Oceananigans.jl/blob/bcc34f07b3f949ea6fb34c7814f4b856d24924c2/src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl#L16-L17

PR #3050, opened 2023-04-04
Remove swe regression + add warning in `ShallowWaterModel`

navidcy:
Related to #3049


src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -127,6 +127,9 @@ function ShallowWaterModel(;
                  timestepper::Symbol = :RungeKutta3,
                          formulation = ConservativeFormulation())
 
+    @warn "The ShallowWaterModel is not being tested. Please be aware and exercise caution if you decide to use this modelfor science purposes."
+    @warn "In particular, the model is not validated against known solution. Use at your own risk."
```
glwagner:
```suggestion
    @warn "The ShallowWaterModel is currently unvalidated and should not be used for scientific research." 
```

glwagner (approved):

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.80.0"
+version = "0.81.0"
```
glwagner, line 4:
```suggestion
version = "0.80.1"
```

francispoulin:
I am happy to review this but a question first.

I thought the regression tests were passing, as of a week or so again. I presume something happened so that they don't run anymore?  Could you point me to where the problem is so I can better follow the conversation?

navidcy:
I also thought they were passing... But it ended up that they were commented out all this time that I thought they were passing.

The pass in erratic manner. Sometimes....

I believe that at some point there were changes in the tendencies. We need to understand whether the changes were intentional or introduced a bug.

francispoulin (approved):
This all looks good to me.

I did the regression tests both on my local computer and a server and they worked, but maybe they wouldn't always work.  I am happy to do further testings with this but I'm not sure what to try at the moment.

francispoulin:
There seem to be two errors.  The one with the cpu needs to instantiate.  

```
/bin/bash: line 2: 1539202 Bus error               (core dumped) /storage5/buildkite-agent/julia-1.8.5/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.instantiate(; verbose=true)'
--
  | 🚨 Error: The command exited with status 135
  | user command error: exit status 135

```

glwagner:
It's ready @navidcy 

@francispoulin right, that's why it's so hard to understand the problem. It only fails intermittently.

We have some validation of the shallow water model in the example. I propose that in order to move forward, we should generate a few additional validation tests including some with bathymetry. Once we have those and have analyzed them to ensure we are satisfied with the dynamics, we should then work on designing a better regression test --- one that won't "pass" when it should not. Then we can add that better regression test to our test suite, and remove the warning from the shallow water model constructor.

Probably best to also resolve the issue with the pressure gradient in presence of bathymetry. Maybe a topographic Rossby wave or some such would help build confidence that we are treating non-uniform bathymetry correctly with both of the formulations that we have.

francispoulin:
> It's ready @navidcy
> 
> @francispoulin right, that's why it's so hard to understand the problem. It only fails intermittently.
> 
> We have some validation of the shallow water model in the example. I propose that in order to move forward, we should generate a few additional validation tests including some with bathymetry. Once we have those and have analyzed them to ensure we are satisfied with the dynamics, we should then work on designing a better regression test --- one that won't "pass" when it should not. Then we can add that better regression test to our test suite, and remove the warning from the shallow water model constructor.
> 
> Probably best to also resolve the issue with the pressure gradient in presence of bathymetry. Maybe a topographic Rossby wave or some such would help build confidence that we are treating non-uniform bathymetry correctly with both of the formulations that we have.

Thanks for the explanation.

I agree that looking at a topographic Rossby wave is a good test to have.  In the case of a small linear slope and small amplitude wave, it should be close to an exact solution so that will be easy.

If we want more challenging tests, there are no shortage of those.  Lots of tests have shocks and people like to test those but I would suggest avoiding those, as shocks mean the aspect ratio is order one or larger, and that violates the underyling hydrostatic assumption in the model.

We could take the Bicklet jet example we have and put it over topography and compute the growth rates.  There is a Poulin and Flierl (2005) in JPO that studies that and it would be easy enough to put to get linear stability code to compute the growth rates.  That way we also have some theory to support the code.

Also, we could look at a wind-driven gyre problem with and without topography. Not a simple test but would help to validate the global simulations that @simone-silvestri did a while ago with the shallow water model.  Lots we can do!

I don't remember the issue about the pressure gradient term with bathymetry but happy to get into it sometime.

navidcy:
Btw, as a side-note, I wanted to revisit those global SWE simulations. I had the feeling that the bathymetry was flipped (and possibly that's why some currents were moving the other way!)

francispoulin:
> Btw, as a side-note, I wanted to revisit those global SWE simulations. I had the feeling that the bathymetry was flipped (and possibly that's why some currents were moving the other way!)

Glad you mentioned it as we never figured out why.   Definitely something to revisit after some simple tests.

If we start off with a topographic Rossby wave it will be very easy to know whether things are flipped.

glwagner:
> If we start off with a topographic Rossby wave it will be very easy to know whether things are flipped.

I think he means that things were flipped in the input / setup (not in the source code).

francispoulin:
This week I plan to revisit this and hopefully figure out how to make the `ShallowWaterModel` stable and not experimental.  Lots of nice work was started on the multi-layer version and I hope that we can continue to work towards that after we get the fluxes and the regression tests.

I just wanted to confirm that people are still supportive of this before I dive in too deep.

If yes, @simone-silvestri , would you have time to meet for 1 hour next week to fine tune things?

simone-silvestri:
Sure, it's a nice idea to get the multi-layer version going again.

francispoulin:
Thanks @simone-silvestri for the quick reply!  I'll work on it this work and message you on slack so we can find a time that is convenient for you for us to chat, and hopefully get this moving again.

Issue #3040, opened 2023-03-30
`minimum_zspacing` isn't working on the GPU


tomchor:
I haven't had the the time to investigate this for now, but `min_Δz()` (and supposedly the x and y version as well) complains about scalar indexing on the GPU when using irregular z spacing:

```
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore are only permitted from the REPL for prototyping purposes.
If you did intend to index this array, annotate the caller with @allowscalar.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] assertscalar(op::String)
    @ GPUArraysCore /glade/work/tomasc/.julia/packages/GPUArraysCore/HaQcr/src/GPUArraysCore.jl:103
  [3] getindex
    @ /glade/work/tomasc/.julia/packages/GPUArrays/7TiO1/src/host/indexing.jl:9 [inlined]
  [4] getindex
    @ /glade/work/tomasc/.julia/packages/OffsetArrays/TcCEq/src/OffsetArrays.jl:436 [inlined]
  [5] getindex
    @ ./subarray.jl:315 [inlined]
  [6] mapreduce_impl(f::typeof(identity), op::typeof(min), A::SubArray{Float64, 1, OffsetArrays.OffsetVector{Float64, CuArray{Float64, 1, CUDA.Mem.DeviceBuffer}}, Tuple{UnitRange{Int64}}, true}, first::Int64, last::Int64)
    @ Base ./reduce.jl:638
  [7] _mapreduce(f::typeof(identity), op::typeof(min), #unused#::IndexLinear, A::SubArray{Float64, 1, OffsetArrays.OffsetVector{Float64, CuArray{Float64, 1, CUDA.Mem.DeviceBuffer}}, Tuple{UnitRange{Int64}}, true})
    @ Base ./reduce.jl:442
  [8] _mapreduce_dim
    @ ./reducedim.jl:365 [inlined]
  [9] #mapreduce#765
    @ ./reducedim.jl:357 [inlined]
 [10] mapreduce
    @ ./reducedim.jl:357 [inlined]
 [11] #_minimum#787
    @ ./reducedim.jl:999 [inlined]
 [12] _minimum
    @ ./reducedim.jl:999 [inlined]
[...]

PR #3054, opened 2023-04-06
Perform some of the grid tests on GPU as well

navidcy:
In response to #3040 I make sure that some of the grid tests also happen on GPU.


tomchor:
Good idea!

@navidcy did you manually cancel the tests?

glwagner:
I don't think we can test right now since we ran out of buildkite minutes. We have to wait until April 10.

navidcy:
> Good idea!
> 
> @navidcy did you manually cancel the tests?

I did, because it was futile as @glwagner mentioned.

navidcy:
@glwagner can you review? should we make more grid tests run on GPU?

test/test_grids.jl
```diff
@@ -188,33 +188,35 @@ function test_regular_rectilinear_xnode_ynode_znode_and_spacings(arch, FT)
     for (grid_type, grid) in zip(grids_types, grids)
         @info "        Testing grid utils on $grid_type grid...."
 
-        @test xnode(2, grid, Center()) ≈ FT(π/2)
-        @test ynode(2, grid, Center()) ≈ FT(π/2)
-        @test znode(2, grid, Center()) ≈ FT(π/2)
+        CUDA.@allowscalar begin
+            @test xnode(2, grid, Center()) ≈ FT(π/2)
+            @test ynode(2, grid, Center()) ≈ FT(π/2)
+            @test znode(2, grid, Center()) ≈ FT(π/2)
```
glwagner, line 194: I don't think these use the GPU since the `grid` above is `CPU`?

glwagner: Maybe we don't need to test `xnode` separately from `xnodes` for GPU, what do you think?

navidcy: it's not a CPU grid -- I changed it

navidcy:
> Are there any tests for `xnodes` with stretched grids?

There seem to be no tests for `x/y/z/λ/φnodes` whatsoever. Shall we add?

tomchor (approved):

tomchor:
> > Are there any tests for `xnodes` with stretched grids?
> 
> There seem to be no tests for `x/y/z/λ/φnodes` whatsoever. Shall we add?

I think at this point there are tests for x/y/znodes? If not, it's a good idea to add a couple.

In any case this PR is already an improvement, so I approved it. @navidcy I'll leave it up to you if you wanna merge as is.

navidcy:
I added some more but they definitely don't cover _all_ cases.

Either I forgot or I am missing the point of `x/y/z/λ/φspacing` methods (the ones that give you a particular spacing). Like where would those be useful?

Btw, a periodic grid with faces at [0, 1, 11, 101] implying centers at [0.5, 6, 56] has the following spacings. It took me a while to confirm that the first spacing for Faces is actually correct.

```julia
julia> grid = RectilinearGrid(size=3, topology=(Periodic, Flat, Flat), x=[0, 1, 11, 101])
F = vcat(F₋, interiorF, F₊) = [-101.0, -100.0, -90.0, 0.0, 1.0, 11.0, 101.0, 102.0, 112.0, 202.0]
C = [(F[i + 1] + F[i]) / 2 for i = 1:TC] = [-100.5, -95.0, -45.0, 0.5, 6.0, 56.0, 101.5, 107.0, 157.0]
3×1×1 RectilinearGrid{Float64, Periodic, Flat, Flat} on CPU with 3×0×0 halo
├── Periodic x ∈ [0.0, 101.0)     variably spaced with min(Δx)=1.0, max(Δx)=90.0
├── Flat y
└── Flat z

julia> xspacings(grid, Center())
3-element view(OffsetArray(::Vector{Float64}, -2:5), 1:3) with eltype Float64:
  1.0
 10.0
 90.0

julia> xspacings(grid, Face())
3-element view(OffsetArray(::Vector{Float64}, -3:6), 1:3) with eltype Float64:
 45.5
  5.5
 50.0
```

navidcy:
I'm leaning towards merging as is or closing. @glwagner?

navidcy:
OK, I added some tests for variably spaced grids as well. @glwagner, what do you think?

navidcy:
@glwagner I'm thinking to close this. Shall I?

PR #3066, opened 2023-04-18
Start moving CUDA support into a package extension

vchuravy:
Main motivation here is to make it possible to move CUDA with Julia 1.9 to a optional dependency,
making loading faster for non GPU workloads as well as making it easier for other GPU backends to be added.

Package extensions is a backwards compatible 1.9 feature, in 1.8 and prior we still have to load CUDA.jl by default
but on 1.9 this is no longer required.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -2,6 +2,7 @@
 
 julia_version = "1.8.5"
 manifest_format = "2.0"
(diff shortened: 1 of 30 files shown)


glwagner:
Ok, we will continue this elsewhere!

vchuravy:
Thanks! I was doing spring cleaning of my stale PRs :)

PR #3075, opened 2023-04-19
Try to speed up documentation build

glwagner:
First thing is simply to use more threads.

But if we can uses parallelism at a little higher level we might do better.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,6 +1,7 @@
 [deps]
 CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+DocumenterCitations = "daee34ce-89f3-4625-b898-19384cb65244"
 Glob = "c27321d9-0574-5035-807b-f59d2c89b15c"
 JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,7 +1,6 @@
 [deps]
 CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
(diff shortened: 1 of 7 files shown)


glwagner:
The tests have sped up a lot (presumably because compile time is so much better with KA 0.9). This means that the docs build is a much bigger bottleneck. I think we're at 2h48m:

https://buildkite.com/clima/oceananigans/builds/11123#01879a9c-fde8-4691-a3f5-29ef6060e288

so we will see what we get down to.

navidcy:
It had some speedup effect…

navidcy (commented):
mutlithreading the docs built has no effect...

glwagner:
Yea weird.

glwagner:
Trying this now:

https://github.com/glwagner/Documenter.jl/blob/225b7cdadd451ed43ecdf5078439d72fad5b9994/src/Expanders.jl#L37

Though, it may be better to use `asyncmap` or something. But we'll see what happens this time around.

docs/Project.toml
```diff
@@ -1,7 +1,6 @@
 [deps]
 CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-DocumenterCitations = "daee34ce-89f3-4625-b898-19384cb65244"
```
navidcy, line 4: probably removing DocumenterCitations will make docs built to fail

navidcy: I see the problem probably is that you "upgraded" Documenter to v0.28 in your fork and then DocumenterCitations can't handle it due to compat restriction here --> https://github.com/ali-ramadhan/DocumenterCitations.jl/blob/5ef4fbf4b71cc04941ad5258adea666c046b73da/Project.toml

So perhaps just add back the `-DEV` in the Documenter version for these experiments?

glwagner: Should I have changed the version on my fork to 0.27? I'm confused what the right process is here

navidcy: That would also worked!! :)

navidcy: Do that. Then no changes in DocumenterCitation are needed

navidcy:
@glwagner did you try building the docs locally? If they don’t build locally then it’s futile.

navidcy:
Shall we close this PR in favor of #3135?

glwagner:
ah yea for sure

PR #3079, opened 2023-04-21
Fix bug in buoyancy implementation

### LINKED ISSUE #3036: Simulation with tilted gravity and background buoyancy increases in total energy
The example below creates a simulation with a rotated gravity vector, a background buoyancy field that's just a stratified background with linear stratification and it starts everything off with some resolved noise:

```julia
using Oceananigans
using Oceananigans.Units
using Oceananigans.Grids: min_Δx, min_Δz

θ_rad = 0.15
N²∞ = 5e-6
grid = RectilinearGrid(CPU(), topology=(Periodic, Flat, Bounded),
                       size=(50, 30), extent = (200, 80))

@inline b_bg(x, y, z, t, p) = p.N²∞ * (x * p.g1 +  z * p.g3)
B_field = BackgroundField(b_bg, parameters=(; N²∞, g1 = sin(θ_rad), g3 = cos(θ_rad)))

model = NonhydrostaticModel(grid = grid, timestepper = :RungeKutta3,
                            advection = UpwindBiasedFifthOrder(),
                            buoyancy = Buoyancy(model=BuoyancyTracer(), gravity_unit_vector=-[sin(θ_rad), 0, cos(θ_rad)]),
                            tracers = :b,
                            background_fields = (; b=B_field),
                            )

amp = 1e-3
xₚ = grid.Lx*rand(5); zₚ = -grid.Lz*rand(5) # random locations for 5 gaussians
σx = 4min_Δx(grid);   σz = 4min_Δz(grid) # spread of Gaussians
resolved_noise(x, y, z) = sum(@. amp * exp(-(x-xₚ)^2/σx^2 -(z-zₚ)^2/σz^2))
set!(model, u=resolved_noise, w=resolved_noise)

u, v, w = model.velocities
b = model.tracers.b

using Statistics: mean
u.data.parent .-= mean(u)
w.data.parent .-= mean(w)

simulation = Simulation(model, Δt = 0.5 * min_Δz(grid) / maximum(abs, v),
                        stop_time = 3days)

wizard = TimeStepWizard(max_change=1.01, cfl=0.2, diffusive_cfl=0.1, max_Δt=0.2/√N²∞)
simulation.callbacks[:wizard] = Callback(wizard, IterationInterval(2))

KE = Average((u^2 + w^2) / 2)
PE = Average(b^2/(2*N²∞))
outputs = (; u, w, b, KE, PE)

simulation.output_writers[:writer] = NetCDFOutputWriter(model, outputs;
                                                        filename = "mwe.nc",
                                                        schedule = TimeInterval(1hour),
                                                        with_halos = false,
                                                        overwrite_existing = true)
run!(simulation)

using Rasters, Plots
fpath_xz1 = simulation.output_writers[:writer].filepath
ds_xz1 = RasterStack(fpath_xz1)
u_lims = (-amp, +amp)

times = dims(ds_xz1, :Ti)
anim = @animate for frame in 1:length(times)
    p1 = heatmap(ds_xz1[:u][Ti=frame, yC=1]', clim=u_lims, c=:balance, title="u")
    p2 = heatmap(ds_xz1[:w][Ti=frame, yC=1]', clim=u_lims, c=:balance, title="w")
    p3 = heatmap(ds_xz1[:b][Ti=frame, yC=1]', title="b")
    p4 = plot(ds_xz1[:KE])
    plot!(p4, ds_xz1[:PE])
    vline!(p4, [times[frame]], label="time", linestyle=:dash)

    fullplot = plot(p1, p2, p3, p4, layout = Plots.grid(2, 2), size=(1000, 600))
end

mp4(anim, "mwe.mp4", fps=16)

```

I expected this simulation to slowly lose or approximately conserve total energy. Instead I get these grid-scale noises popping up and a build up of kinetic and potential energy seemingly out of nowhere:

https://user-images.githubusercontent.com/13205162/228424859-905f3dd6-b167-440b-b073-af09008f4f04.mp4

Notes: 

- Even if I add a background diffusivity with a reasonable value, the energy buildup still happens
- The energy buildup does not happen without a background buoyancy
- When `gravity_unit_vector == NegativeZDirection()` the KE stays approximately the same, but the PE increases. Although that might be due to the boundary conditions, I'm not sure at this point.
- When I start this with a larger velocity amplitude the small scale noise becomes less evident, but there's still an increase in the total energy (KE and PE) of the flow

Any guesses as to what might be causing this?

glwagner:
Buoyancy wasn't being reconstructed at the velocity locations.

Closes #3036

### DIFF
--- a/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
+++ b/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
@@ -3,17 +3,17 @@ using Oceananigans.ImmersedBoundaries: PartialCellBottom, ImmersedBoundaryGrid
 
 """
 Update the hydrostatic pressure perturbation pHY′. This is done by integrating
-the `buoyancy_perturbation` downwards:
+the `buoyancy_perturbationᶜᶜᶜ` downwards:
 
-    `pHY′ = ∫ buoyancy_perturbation dz` from `z=0` down to `z=-Lz`
+    `pHY′ = ∫ buoyancy_perturbationᶜᶜᶜ dz` from `z=0` down to `z=-Lz`
 """
 @kernel function _update_hydrostatic_pressure!(pHY′, grid, buoyancy, C)
     i, j = @index(Global, NTuple)
 
-    @inbounds pHY′[i, j, grid.Nz] = - z_dot_g_b(i, j, grid.Nz+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, grid.Nz+1, grid)
+    @inbounds pHY′[i, j, grid.Nz] = - z_dot_g_bᶜᶜᶠ(i, j, grid.Nz+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, grid.Nz+1, grid)
 
     @unroll for k in grid.Nz-1 : -1 : 1
-        @inbounds pHY′[i, j, k] = pHY′[i, j, k+1] - z_dot_g_b(i, j, k+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, k+1, grid)
+        @inbounds pHY′[i, j, k] = pHY′[i, j, k+1] - z_dot_g_bᶜᶜᶠ(i, j, k+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, k+1, grid)
     end
 end
 
--- a/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
+++ b/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
@@ -10,10 +10,10 @@ the `buoyancy_perturbation` downwards:
 @kernel function _update_hydrostatic_pressure!(pHY′, grid, buoyancy, C)
     i, j = @index(Global, NTuple)
 
-    @inbounds pHY′[i, j, grid.Nz] = - ℑzᵃᵃᶠ(i, j, grid.Nz+1, grid, z_dot_g_b, buoyancy, C) * Δzᶜᶜᶠ(i, j, grid.Nz+1, grid)
+    @inbounds pHY′[i, j, grid.Nz] = - z_dot_g_b(i, j, grid.Nz+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, grid.Nz+1, grid)
 
     @unroll for k in grid.Nz-1 : -1 : 1
-        @inbounds pHY′[i, j, k] = pHY′[i, j, k+1] - ℑzᵃᵃᶠ(i, j, k+1, grid, z_dot_g_b, buoyancy, C) * Δzᶜᶜᶠ(i, j, k+1, grid)
+        @inbounds pHY′[i, j, k] = pHY′[i, j, k+1] - z_dot_g_b(i, j, k+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, k+1, grid)
     end
 end
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.81.0"
+version = "0.81.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/BuoyancyModels/BuoyancyModels.jl
+++ b/src/BuoyancyModels/BuoyancyModels.jl
@@ -1,9 +1,9 @@
 module BuoyancyModels
 
 export
-    Buoyancy, BuoyancyTracer, SeawaterBuoyancy, buoyancy_perturbation,
+    Buoyancy, BuoyancyTracer, SeawaterBuoyancy, buoyancy_perturbationᶜᶜᶜ,
     LinearEquationOfState, RoquetIdealizedNonlinearEquationOfState, TEOS10,
-    ∂x_b, ∂y_b, ∂z_b, buoyancy_perturbation, x_dot_g_b, y_dot_g_b, z_dot_g_b,
+    ∂x_b, ∂y_b, ∂z_b, buoyancy_perturbationᶜᶜᶜ, x_dot_g_bᶠᶜᶜ, y_dot_g_bᶜᶠᶜ, z_dot_g_bᶜᶜᶠ,
     top_buoyancy_flux,
     buoyancy_frequency_squared,
     BuoyancyField
--- a/src/BuoyancyModels/buoyancy_field.jl
+++ b/src/BuoyancyModels/buoyancy_field.jl
@@ -4,6 +4,7 @@ using Oceananigans.Fields: Field, ZeroField
 buoyancy(::Nothing, args...) = ZeroField()
 buoyancy(::BuoyancyTracer, grid, tracers) = tracers.b
 
+# TODO: move to Models
 buoyancy(model) = buoyancy(model.buoyancy, model.grid, model.tracers)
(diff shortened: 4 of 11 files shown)


glwagner:
We should annotate `x_dot_g_b` with the staggered location (`fcc`, etc).

I think the `BuoyancyModels` module could also be cleaned up.

@tomchor let us know if this fixes things

tomchor:
Apparently something breaks in this bugfix. Forgetting about the tilted buoyancy stuff for a while, even when the buoyancy isn't tilted this results in accumulating energy. In the simulation below I'm just using `buoyancy = BuoyancyTracer()` and aligning the stratification accordingly:

https://user-images.githubusercontent.com/13205162/233681167-b3d21670-ddd4-47fe-92e6-5227fb2cf344.mp4

tomchor:
But applying this bugfix without separating the pressure into hydrostatic-nonhydrostatic works:

https://user-images.githubusercontent.com/13205162/233682034-393b5303-6aa1-494c-ac28-a28347b8bbbc.mp4

glwagner:
Nice, thanks for testing that. I found the second bug (which was created by the bugfix). I think it should work now.

This could allow us to eliminate the hydrostatic pressure from the nonhydrostatic model and thereby save a field. @simone-silvestri . Apparently all our prior tests were flawed because of this bug.

tomchor:
So it appears that there are two competing bug here, which kinda cancel out in the case where gravity aligned with the vertical direction. I also think that back in https://github.com/CliMA/Oceananigans.jl/pull/1910 when we saw some weird effects when not separating the pressure, this might have been it.

tomchor:
> Nice, thanks for testing that. I found the second bug (which was created by the bugfix). I think it should work now.
> 
> This could allow us to eliminate the hydrostatic pressure from the nonhydrostatic model and thereby save allocating one field. @simone-silvestri . Apparently all our prior tests were flawed because of this bug. We still need the hydrostatic pressure integral in the hydrostatic model.

Ha, you beat me to the punchline. But yes, agreed!

glwagner:
> So it appears that there are two competing bug here, which kinda cancel out in the case where gravity aligned with the vertical direction. I also think that back in #1910 when we saw some weird effects when not separating the pressure, this might have been it.

I fixed it. It was trivial (check out the diff)

glwagner:
As the diff shows, this PR reconstructs buoyancy at the location it's needed. For example, `x_dot_g_b` must reconstruct the buoyancy perturbation, which is located at tracer locations `ccc`, at the location `fcc`. 

Similarly, `z_dot_g_b` should be reconstructed at `ccf` (the location of vertical velocity). Note that we _also_ do this reconstructed when computing the hydrostatic pressure integral, therefore:

$$ p_h = \int^0_z \bar b_z \text{ } dz $$

where $\overline{()}^z$ indicates a reconstruction in $$z$$.

Previously, we were assuming within the hydrsotatic pressure integral that `z_dot_g_b` returned something at `ccc`.

The first commit on this PR fixed the reconstruction for `z_dot_g_b`. However, making this change also requires "undoing" the reconstruction in the hyrostatic pressure integral.

With both changes, the regression tests should pass _and_ tilted gravity should be fixed.

Hope that clears things up.

simone-silvestri (approved):

simone-silvestri:
Nice find!

tomchor (approved):
@glwagner this is a great catch! Thanks so much for your help with this issue.

tomchor:
> As the diff shows, this PR reconstructs buoyancy at the location it's needed. For example, `x_dot_g_b` must reconstruct the buoyancy perturbation, which is located at tracer locations `ccc`, at the location `fcc`.
> 
> Similarly, `z_dot_g_b` should be reconstructed at `ccf` (the location of vertical velocity). Note that we _also_ do this reconstructed when computing the hydrostatic pressure integral, therefore:
> 
> ph=∫z0b¯z dz
> 
> where ()―z indicates a reconstruction in z.
> 
> Previously, we were assuming within the hydrsotatic pressure integral that `z_dot_g_b` returned at `ccc`.

I think this is the only part that I'm confused about. Why was not interpolating `z_dot_g_b` okay?

I ran some tests to investigate this and found some interesting things.

As a background: when we originally tried to nuke the pressure separation, I noticed that some wave noise was being created. Unfortunately the videos of that PR were lost in time so I can't show them to you, but here is the relevant comment: https://github.com/CliMA/Oceananigans.jl/pull/1910#issuecomment-995840271

The worst offender was the internal wave example, where the wave packet seemed to break down pretty quickly without a hydrostatic pressure separation. So I just ran the same example using the `main` branch and branch https://github.com/CliMA/Oceananigans.jl/tree/tc/nhpressure2 (which now has this bugfix in and it doesn't separate the pressure). I ran it for 10 times longer and here are the videos of both (respectively):

https://user-images.githubusercontent.com/13205162/233692840-edd9780b-dbc4-4958-843d-9895ac85c32f.mp4

https://user-images.githubusercontent.com/13205162/233692959-841c2a2c-2a8a-40e2-9448-fa2943a67502.mp4

They look exactly the same to me. So whatever was causing the added wave noise in https://github.com/CliMA/Oceananigans.jl/pull/1910 when we removed the hydrostatic pressure (whether it was the things you changed in this PR or not), it's not causing the noise anymore.

I haven't tested the other docs examples but this was the worse offender, so I'm confident they'll look good too.

With this is mind, I'll open a draft PR with https://github.com/CliMA/Oceananigans.jl/tree/tc/nhpressure2 to see what happens to the tests. I think if the dynamics look the same we could reopen the discussion about removing the hydrostatic pressure.

glwagner:
Another benefit of removing the hydrostatic pressure is that we can improve the documentation and user interface too. For example, the hydrostatic pressure in the hydrostatic model is stored in something like

```
model.pressure.pHY
```

but we can just have a single `model.pressure` that has the pressure field for both models with this change. It's simpler. 

It's a significant refactor to remove hydrostatic pressure --- I don't think we want to do the hacky thing that we did to set up that PR that just keeps the same structures but sets pressure to `nothing`, we want to do it properly and update the docs.

tomchor:
> It's a significant refactor to remove hydrostatic pressure --- I don't think we want to do the hacky thing that we did to set up that PR that just keeps the same structures but sets pressure to `nothing`, we want to do it properly and update the docs.

Agreed. https://github.com/CliMA/Oceananigans.jl/pull/3080 is for now just a start to see if anything significant breaks. So far everything looks OK there

tomchor:
@glwagner can we bump the patch version and register a new release when you think this is ready?

navidcy:
Bump patch release?

glwagner:
yes

navidcy (approved):

src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
```diff
@@ -3,17 +3,17 @@ using Oceananigans.ImmersedBoundaries: PartialCellBottom, ImmersedBoundaryGrid
 
 """
 Update the hydrostatic pressure perturbation pHY′. This is done by integrating
-the `buoyancy_perturbation` downwards:
+the `buoyancy_perturbationᶜᶜᶜ` downwards:
 
-    `pHY′ = ∫ buoyancy_perturbation dz` from `z=0` down to `z=-Lz`
+    `pHY′ = ∫ buoyancy_perturbationᶜᶜᶜ dz` from `z=0` down to `z=-Lz`
```
navidcy, line 8: ok, is this really what's happening?

$$ \int_0^{-L_z} p' \mathrm{d}z $$

Seems a bit odd... why don't we flip the limits and add a minus sign?

glwagner: no, its

$$ \int^0_z p' \text{ d}z $$

glwagner: "down to" is supposed to mean the summation occurs in the -z direction

glwagner: You could instead say "we sum from z to 0". But that doesn't explain why the code says `for k = grid.Nz-1:-1:1`

navidcy:
merge?

PR #3092, opened 2023-04-30
Allow users to mask output

### LINKED ISSUE #3061: User interface for masking output on `ImmersedBoundaryGrid`
I have found it inconvenient that output is masked to `0`. For most purposes, I'd prefer `NaN`. I'm opening this issue to discuss a user interface that would implement such a feature.

My first idea is to add a property / kwarg to output writers called `immersed_values`.

Then we might write something like

```julia
output_writer = JLD2OutputWriter(model, output, immersed_values=NaN, other_kwargs...)
```

The default will be `immersed_values=nothing`. Then `fetch_output`:

https://github.com/CliMA/Oceananigans.jl/blob/72e2197e640ddf308fd40aa4658e7686e596bccd/src/OutputWriters/fetch_output.jl#L12-L15

would become something like

```julia

 function fetch_output(field::AbstractField, model, immersed_value) 
    !isnothing(immersed_value) && mask_immersed_field!(field, immersed_value)
    return parent(field) 
 end 
```

navidcy:
Resolves #3061

Questions:

Does it make sense to mask when`with_halos = true`? With `with_halos = true` implies that the users can use the output to compute derivative fields etc; will masking obscure this?


navidcy:
`mask_immersed_field!` was renamed to `mask_immersed!`; this is a breaking change, right?

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -336,8 +336,8 @@ function setup_split_explicit!(auxiliary, state, η, grid, Gu, Gv, Guⁿ, Gvⁿ,
     initialize_free_surface_state!(state, η)
 
     # Wait for predictor velocity update step to complete and mask it if immersed boundary.
```
navidcy, line 338: This wait comment is a leftover??

src/OutputWriters/fetch_output.jl
```diff
@@ -9,8 +10,9 @@ time(::Nothing) = nothing
 
 fetch_output(output, model) = output(model)
```
navidcy, line 11: What is this line here for? Is this a fallback of sorts? Do I need to replace it with `fetch_output(output, model, mask_value) = output(model)`?

navidcy: @glwagner, @simone-silvestri ?

navidcy:
I was trying to run `validation/immersed_boundaries/flow_over_hills.jl` and all simulations except the ref one are NaNing pretty quickly. On `main` it works!

glwagner:
> Resolves #3061
> 
> Questions:
> 
> Does it make sense to mask when`with_halos = true`? With `with_halos = true` implies that the users can use the output to compute derivative fields etc; will masking obscure this?

No, because derivatives are "boundary aware" and thus don't touch the regions that are masked.

src/OutputWriters/fetch_output.jl
glwagner: Yes it's a fallback, so that users can specify output which is a callable object of `model`.

glwagner: Technically we should be able to mask that output if `output(model)` returns a `Field` though.

glwagner: could change it to

```julia
function fetch_output(output, model, mask_value)
    fetched = output(model)
    if fetched isa Field
        !isnothing(mask_value) && mask_immersed(fetched, mask_value)
        return parent(fetched)
    end
    return fetched
end
```

Slightly friendlier behavior i think

src/OutputWriters/jld2_output_writer.jl
```diff
@@ -190,7 +190,7 @@ function JLD2OutputWriter(model, outputs; filename, schedule,
 
     initialize_jld2_file!(filepath, init, jld2_kw, including, outputs, model)
     
-    mask_value = mask_immersed is Number ? eltype(model.grid)(mask_immersed) : mask_immersed
+    mask_value = mask_immersed isa Number ? eltype(model.grid)(mask_immersed) : mask_immersed
```
glwagner, line 193: I don't think this matters. It would be converted inside `mask_immersed!`.

Also I think it's better to keep the same name as the kwarg because its simpler and easier to understand. Then there's a property `output_writer.mask_immersed` which should be the input value.

navidcy:
I'm trying to finish up this.

@glwagner, is there a reason you here put a function to return the kernel? Usually the kernel is launched and then the function returns nothing.

https://github.com/CliMA/Oceananigans.jl/blame/d7d146bc59f966d65eb66a875c16679bef365130/src/ImmersedBoundaries/mask_immersed_field.jl#L44

(cc @simone-silvestri)

src/OutputWriters/fetch_output.jl
```diff
@@ -1,16 +1,28 @@
...
+    if fetched isa Field
+        !isnothing(mask_immersed) && mask_immersed!(fetched, mask_immersed)
+        return parent(fetched)
+    end
+    return fetched
+end
+
+function fetch_output(field::AbstractField, model, mask_immersed)
     compute_at!(field, time(model))
+    !isnothing(mask_immersed) && mask_immersed!(field, mask_immersed)
```
navidcy, line 25: @glwagner I think this won't work because what happens is that, e.g., if `u` is the output to be saved then `model.velocities.u` is masked and then it includes NaNs at the next step so the whole state NaN...

At least that's what I found happening and I boil it down to this line here masking and then returning the parent of the field. Am I right?

glwagner: Ok that makes sense. But the real problem is how we define masking. Consider: the mask we apply to a field for output _should not affect time-stepping_. If it does, there's a problem. So let's fix that before hacking around it.

The issue is velocity components I think. The "mask" is defined to cover the points that are _on_ the boundary, not just inside it. Thus masking the velocity to 0 is how we ensure impenetrability.

The confusion is therefore: are velocity nodes _on_ the boundary immersed or not?

I think we should probably define this points as _not_ immersed. Therefore to fix the problem we can:

1. Implement a new function that satisfies impenetrability. We can start by doing this just for immersed grids.
2. Define "mask_immersed" as something that only affects cells that are _fully_ immersed. This has no effect on fully centered fields, but for staggered fields this implies a change: cells are only masked if they are not adjacent to active cells (I think the distinction between `inactive_node` and `peripheral_node` can be used for this).

navidcy: Yes, I was thinking of point 2 you make and modified `masked_immersed!` to use `inactive_node` (just to test the hypothesis) but that didn't work either.

simone-silvestri: There might also be a problem with the implicit solver to fix. Our strategy right now is to zero out elements of the matrix that correspond to immersed cells. Unfortunately `0 * NaN = NaN` so any `NaN` in the immersed boundary will be propagated immediately in the whole column

navidcy: Good point.

Can't we just apply the mask to the saved output?

simone-silvestri: Note that in contrast, `false * NaN = 0.0`

glwagner: > Good point.
> 
> Can't we just apply the mask to the saved output?

Yes but it's more complicated, since output can be windowed / sliced. It's possible though.

glwagner: > 

What do you mean by this exactly?

glwagner: > Can't we just apply the mask to the saved output?

What do you mean?

glwagner: > There might also be a problem with the implicit solver to fix. Our strategy right now is to zero out elements of the matrix that correspond to immersed cells. Unfortunately 0 * NaN = NaN so any NaN in the immersed boundary will be propagated immediately in the whole column

The immersed solver can then apply `mask_immersed!(velocities, 0)`.

simone-silvestri: I am referring to the vertically implicit solver. We do not use preconditioners, we use the Thomas algorithm and solve the tridiagonal system directly

glwagner: Oh _that_ "implicit solver".

I guess that is the code here:

https://github.com/CliMA/Oceananigans.jl/blob/a1f4f4e75e51a83cf9dec9e671866cff67de1211/src/ImmersedBoundaries/grid_fitted_immersed_boundaries.jl#L126-L148

Seems like we would need

https://github.com/CliMA/Oceananigans.jl/blob/a1f4f4e75e51a83cf9dec9e671866cff67de1211/src/Solvers/batched_tridiagonal_solver.jl#L179

to return 0 for `immersed_cell`.

To accomplish this without explicit masking we might have to wrap `field` in a `ConditionalOperation`:

https://github.com/CliMA/Oceananigans.jl/blob/a1f4f4e75e51a83cf9dec9e671866cff67de1211/src/TurbulenceClosures/vertically_implicit_diffusion_solver.jl#L197-L199

hmm...

Another possibility is to move `mask_immersed!` from `update_state!`; for example if we masked fields before `ab2_step!` then I think we'd be ok. Not sure if that would affect non-masked output. Hopefully not.

navidcy: > > Can't we just apply the mask to the saved output?
> 
> What do you mean?

I mean only mask the data written in the jld2/nc file, not the actual data in the u, v, w, fields, etc.

glwagner: That would require allocating a temporary array (in the GPU case we already do this on the CPU; however we would then have to mask on the CPU which could slow things down for large fields).

glwagner: But maybe what you're suggesting could be achieved by digging into JLD2's array-saving code. I'm not sure about NetCDF.

Issue #2960, opened 2023-03-06
`UniformStokesDrift` would benefit from a `parameters` keyword


tomchor:
~At the moment there's no way to run the MWE below on a GPU besides setting `Us` and `wave_vertical_scale` to be `const`ants~:

The example below fails on GPU due to `Us` and `wave_vertical_scale`:

```julia
using Oceananigans

grid = RectilinearGrid(GPU(), size=(4, 4, 4), extent = (1e3, 1e3, 120))

Us = 0.05; wave_vertical_scale = 5
∂z_uˢ(z, t) = Us * exp(z / wave_vertical_scale) / wave_vertical_scale

model = NonhydrostaticModel(; grid, stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ))
time_step!(model, 10)
```

As much as using `const`ants is a valid solution, it would be nice to have the option to pass them as parameters, such as

```julia
UniformStokesDrift(∂z_uˢ=∂z_uˢ, parameters=Us=0.05, wave_vertical_scale=5)
```

which currently isn't an option.

glwagner:
I'm up to add such a feature but to combat disinformation, we must note that it is possible to run without `const`ants:

```julia
using Oceananigans

grid = RectilinearGrid(GPU(), size=(4, 4, 4), extent = (1e3, 1e3, 120))

Base.@kwdef struct MonochromaticStokesShear
    Us :: Float64 = 0.05
    wave_vertical_scale :: Float64 = 5.0
end

@inline (mss::MonochromaticStokesShear)(z, t) = mss.Us * exp(z / mss.wave_vertical_scale) / mss.wave_vertical_scale

model = NonhydrostaticModel(; grid, stokes_drift = UniformStokesDrift(∂z_uˢ=MonochromaticStokesShear()))

time_step!(model, 10)
```

tomchor:
> I'm up to add such a feature but to combat disinformation, we must note that it is possible to run without `const`ants:
> 
> ```julia
> using Oceananigans
> 
> grid = RectilinearGrid(GPU(), size=(4, 4, 4), extent=(1, 1, 1))
> 
> Base.@kwdef struct MonochromaticStokesShear
>     Us :: Float64 = 0.05
>     wave_vertical_scale :: Float64 = 5.0
> end
> 
[...]

PR #3096, opened 2023-05-04
Adds `parameters` keyword to `UniformStokesDrift`

tomchor:
This PR adds a `parameters` keyword to `UniformStokesDrift`.

At the moment this only gets activated if `parameters isa NamedTuple`. Do we want to be more general than that?

Closes https://github.com/CliMA/Oceananigans.jl/issues/2960

### DIFF
--- a/examples/langmuir_turbulence.jl
+++ b/examples/langmuir_turbulence.jl
@@ -53,18 +53,19 @@ wavenumber = 2π / wavelength # m⁻¹
 
 ## The vertical scale over which the Stokes drift of a monochromatic surface wave
 ## decays away from the surface is `1/2wavenumber`, or
-vertical_scale = wavelength / 4π
+const vertical_scale = wavelength / 4π
 
 ## Stokes drift velocity at the surface
-Uˢ = amplitude^2 * wavenumber * frequency # m s⁻¹
+const Uˢ = amplitude^2 * wavenumber * frequency # m s⁻¹
 
+# The `const` declarations ensure that Stokes drift functions compile on the GPU.
+# To run this example on the GPU, one only needs to include `architecture = GPU()` in the
+# constructor for `RectilinearGrid` above.
+#
 # We only need the vertical derivative of the Stokes drift, which is
 
-∂z_uˢ(z, t, p) = 1 / p.vertical_scale * p.Uˢ * exp(z / p.vertical_scale)
-stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ, parameters = (; vertical_scale, Uˢ))
+∂z_uˢ(z, t) = 1 / vertical_scale * Uˢ * exp(z / vertical_scale)
 
-# Note that passing `vertical_scale` and `Uˢ` as parameters ensures that this simulation
-# runs on GPUs.
 #
 # !!! info "The Craik-Leibovich equations in Oceananigans"
 #     Oceananigans implements the Craik-Leibovich approximation for surface wave effects
@@ -126,7 +127,7 @@ model = NonhydrostaticModel(; grid, coriolis,
                             tracers = :b,
                             buoyancy = BuoyancyTracer(),
                             closure = AnisotropicMinimumDissipation(),
-                            stokes_drift = stokes_drift,
+                            stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ),
                             boundary_conditions = (u=u_boundary_conditions, b=b_boundary_conditions))
 
 # ## Initial conditions
--- a/examples/langmuir_turbulence.jl
+++ b/examples/langmuir_turbulence.jl
@@ -62,7 +62,11 @@ const Uˢ = amplitude^2 * wavenumber * frequency # m s⁻¹
 # To run this example on the GPU, one only needs to include `architecture = GPU()` in the
 # constructor for `RectilinearGrid` above.
 #
-# We only need the vertical derivative of the Stokes drift, which is
+# The Stokes drift profile is
+
+uˢ(z) = Uˢ * exp(z / vertical_scale)
+
+# and its `z`-derivative is
 
 ∂z_uˢ(z, t) = 1 / vertical_scale * Uˢ * exp(z / vertical_scale)
 
--- a/examples/langmuir_turbulence.jl
+++ b/examples/langmuir_turbulence.jl
@@ -59,7 +59,7 @@ const vertical_scale = wavelength / 4π
 const Uˢ = amplitude^2 * wavenumber * frequency # m s⁻¹
 
 # The `const` declarations ensure that Stokes drift functions compile on the GPU.
-# To run this example on the GPU, one only needs to include `architecture = GPU()` in the
+# To run this example on the GPU, include `GPU()` in the
 # constructor for `RectilinearGrid` above.
 #
 # The Stokes drift profile is
--- a/examples/langmuir_turbulence.jl
+++ b/examples/langmuir_turbulence.jl
@@ -81,7 +81,7 @@ uˢ(z) = Uˢ * exp(z / vertical_scale)
 #
 # The vertical derivative of the Stokes drift is
 
-∂z_uˢ(z, t) = 1 / vertical_scale * Uˢ * exp(z / vertical_scale)
+∂z_uˢ(z, t, p) = 1 / p.vertical_scale * Uˢ * exp(z / p.vertical_scale)
 
 # Finally, we note that the time-derivative of the Stokes drift must be provided
 # if the Stokes drift and surface wave field undergoes _forced_ changes in time.
@@ -132,7 +132,7 @@ model = NonhydrostaticModel(; grid, coriolis,
                             tracers = :b,
                             buoyancy = BuoyancyTracer(),
                             closure = AnisotropicMinimumDissipation(),
-                            stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ),
+                            stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ, parameters = (; vertical_scale)),
                             boundary_conditions = (u=u_boundary_conditions, b=b_boundary_conditions))
 
 # ## Initial conditions
--- a/examples/langmuir_turbulence.jl
+++ b/examples/langmuir_turbulence.jl
@@ -53,20 +53,18 @@ wavenumber = 2π / wavelength # m⁻¹
 
 ## The vertical scale over which the Stokes drift of a monochromatic surface wave
 ## decays away from the surface is `1/2wavenumber`, or
-const vertical_scale = wavelength / 4π
+vertical_scale = wavelength / 4π
 
 ## Stokes drift velocity at the surface
-const Uˢ = amplitude^2 * wavenumber * frequency # m s⁻¹
+Uˢ = amplitude^2 * wavenumber * frequency # m s⁻¹
 
-# The `const` declarations ensure that Stokes drift functions compile on the GPU.
-# To run this example on the GPU, write `architecture = GPU()` in the constructor
-# for `NonhydrostaticModel` below.
-#
-# The Stokes drift profile is
+# We only need the vertical derivative of the Stokes drift, which is
 
-uˢ(z) = Uˢ * exp(z / vertical_scale)
+∂z_uˢ(z, t, p) = 1 / p.vertical_scale * p.Uˢ * exp(z / p.vertical_scale)
+stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ, parameters = (; vertical_scale, Uˢ))
 
-# which we'll need for the initial condition.
+# Note that passing `vertical_scale` and `Uˢ` as parameters ensures that this simulation
+# runs on GPUs.
 #
 # !!! info "The Craik-Leibovich equations in Oceananigans"
 #     Oceananigans implements the Craik-Leibovich approximation for surface wave effects
@@ -79,10 +77,6 @@ uˢ(z) = Uˢ * exp(z / vertical_scale)
 #     [physics documentation](https://clima.github.io/OceananigansDocumentation/stable/physics/surface_gravity_waves/)
 #     for more information.
 #
-# The vertical derivative of the Stokes drift is
-
-∂z_uˢ(z, t, p) = 1 / p.vertical_scale * Uˢ * exp(z / p.vertical_scale)
-
 # Finally, we note that the time-derivative of the Stokes drift must be provided
 # if the Stokes drift and surface wave field undergoes _forced_ changes in time.
 # In this example, the Stokes drift is constant
@@ -132,7 +126,7 @@ model = NonhydrostaticModel(; grid, coriolis,
                             tracers = :b,
                             buoyancy = BuoyancyTracer(),
                             closure = AnisotropicMinimumDissipation(),
-                            stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ, parameters = (; vertical_scale)),
+                            stokes_drift = stokes_drift,
                             boundary_conditions = (u=u_boundary_conditions, b=b_boundary_conditions))
 
 # ## Initial conditions
--- a/src/StokesDrift.jl
+++ b/src/StokesDrift.jl
@@ -68,11 +68,10 @@ const USD = UniformStokesDrift
 @inline ∂t_vˢ(i, j, k, grid, sw::USD, args...) = sw.∂t_vˢ(znode(k, grid, Center()), args...)
 @inline ∂t_wˢ(i, j, k, grid::AbstractGrid{FT}, sw::USD, args...) where FT = zero(FT)
 
-@inline x_curl_Uˢ_cross_U(i, j, k, grid, sw::USD, U, args...) = @inbounds          ℑxzᶠᵃᶜ(i, j, k, grid, U.w) * sw.∂z_uˢ(znode(k, grid, Center()), args...)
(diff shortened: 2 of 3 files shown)


glwagner:
> This PR adds a `parameters` keyword to `UniformStokesDrift`.
> 
> At the moment this only gets activated if `parameters isa NamedTuple`. Do we want to be more general than that?
> 
> Closes #2960

It's better to fallback to the signature `(z, t)` when `parameters::Nothing`, and then otherwise include.

examples/langmuir_turbulence.jl
```diff
@@ -132,7 +132,7 @@ model = NonhydrostaticModel(; grid, coriolis,
                             tracers = :b,
                             buoyancy = BuoyancyTracer(),
                             closure = AnisotropicMinimumDissipation(),
-                            stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ),
+                            stokes_drift = UniformStokesDrift(∂z_uˢ=∂z_uˢ, parameters = (; vertical_scale)),
```
glwagner, line 135: Maybe keep this example simple?

src/StokesDrift.jl
```diff
@@ -26,24 +26,25 @@ abstract type AbstractStokesDrift end
 ##### Functions for "no surface waves"
 #####
 
-@inline ∂t_uˢ(i, j, k, grid::AbstractGrid{FT}, ::Nothing, time) where FT = zero(FT)
-@inline ∂t_vˢ(i, j, k, grid::AbstractGrid{FT}, ::Nothing, time) where FT = zero(FT)
-@inline ∂t_wˢ(i, j, k, grid::AbstractGrid{FT}, ::Nothing, time) where FT = zero(FT)
+@inline ∂t_uˢ(i, j, k, grid::AbstractGrid{FT}, ::Nothing, args...) where FT = zero(FT)
+@inline ∂t_vˢ(i, j, k, grid::AbstractGrid{FT}, ::Nothing, args...) where FT = zero(FT)
+@inline ∂t_wˢ(i, j, k, grid::AbstractGrid{FT}, ::Nothing, args...) where FT = zero(FT)
```
glwagner, line 31: You shouldn't need this right? I'm confused about the design here

src/StokesDrift.jl
```diff
@@ -26,24 +26,25 @@ abstract type AbstractStokesDrift end
...
 
 """
-    UniformStokesDrift{UZ, VZ, UT, VT} <: AbstractStokesDrift
+    UniformStokesDrift{P, UZ, VZ, UT, VT} <: AbstractStokesDrift
 
 Parameter struct for Stokes drift fields associated with surface waves.
 """
-struct UniformStokesDrift{UZ, VZ, UT, VT} <: AbstractStokesDrift
+struct UniformStokesDrift{P, UZ, VZ, UT, VT} <: AbstractStokesDrift
+    parameters :: P
```
glwagner, line 47: put this as the last property as elsewhere

src/StokesDrift.jl
```diff
@@ -53,30 +54,34 @@ end
...
+@inline ∂t_uˢ(i, j, k, grid, sw::USD, args...) = sw.∂t_uˢ(znode(k, grid, Center()), args...)
+@inline ∂t_vˢ(i, j, k, grid, sw::USD, args...) = sw.∂t_vˢ(znode(k, grid, Center()), args...)
+@inline ∂t_wˢ(i, j, k, grid::AbstractGrid{FT}, sw::USD, args...) where FT = zero(FT)
 
-@inline ∂t_wˢ(i, j, k, grid::AbstractGrid{FT}, sw::USD, time) where FT = zero(FT)
+@inline x_curl_Uˢ_cross_U(i, j, k, grid, sw::USD, U, args...) = @inbounds          ℑxzᶠᵃᶜ(i, j, k, grid, U.w) * sw.∂z_uˢ(znode(k, grid, Center()), args...)
+@inline y_curl_Uˢ_cross_U(i, j, k, grid, sw::USD, U, args...) = @inbounds          ℑyzᵃᶠᶜ(i, j, k, grid, U.w) * sw.∂z_vˢ(znode(k, grid, Center()), args...)
+@inline z_curl_Uˢ_cross_U(i, j, k, grid, sw::USD, U, args...) = @inbounds begin (- ℑxzᶜᵃᶠ(i, j, k, grid, U.u) * sw.∂z_uˢ(znode(k, grid, Face()), args...)
+                                                                                 - ℑyzᵃᶜᶠ(i, j, k, grid, U.v) * sw.∂z_vˢ(znode(k, grid, Face()), args...) )
+                                                                          end
```
glwagner, line 75: begin and end are not necessary

examples/langmuir_turbulence.jl
tomchor: I changed this because (iirc) at some point we agreed (on a issue somewhere) that the examples should be able to run on a GPU without any modifications whenever possible. Without this change this example will fail on GPUs, no?

src/StokesDrift.jl
tomchor: Removed

examples/langmuir_turbulence.jl
navidcy: We can run and see.

src/StokesDrift.jl
tomchor: Done

src/StokesDrift.jl
tomchor: You're right. I accidentally went too far with the search and replace. It's fixed now.

src/StokesDrift.jl
```diff
@@ -39,44 +39,48 @@ abstract type AbstractStokesDrift end
 #####
 
 """
-    UniformStokesDrift{UZ, VZ, UT, VT} <: AbstractStokesDrift
+    UniformStokesDrift{P, UZ, VZ, UT, VT} <: AbstractStokesDrift
```
navidcy, line 42: fix docstring?

navidcy, line 42:
```suggestion
    UniformStokesDrift{UZ, VZ, UT, VT, P} <: AbstractStokesDrift
```

src/StokesDrift.jl
```diff
@@ -39,44 +39,48 @@ abstract type AbstractStokesDrift end
...
 end
 
 addzero(args...) = 0
 
 """
-    UniformStokesDrift(; ∂z_uˢ=addzero, ∂z_vˢ=addzero, ∂t_uˢ=addzero, ∂t_vˢ=addzero)
+    UniformStokesDrift(; ∂z_uˢ=addzero, ∂z_vˢ=addzero, ∂t_uˢ=addzero, ∂t_vˢ=addzero, parameters=nothing)
 
 Construct a set of functions that describes the Stokes drift field beneath
 a uniform surface gravity wave field.
```
navidcy, line 60:
```suggestion
a horizontally-uniform surface gravity wave field.
```

?

examples/langmuir_turbulence.jl
navidcy: I tried running On `#main` and just adding `GPU()` in the grid seems to be enough for the example to run on GPU.

navidcy: Is this what you were implying here? Or did you mean that with the changes in this PR adding the kwarg in the model constructor is necessary?

tomchor:
> > This PR adds a `parameters` keyword to `UniformStokesDrift`.
> > At the moment this only gets activated if `parameters isa NamedTuple`. Do we want to be more general than that?
> > Closes #2960
> 
> It's better to fallback to the signature `(z, t)` when `parameters::Nothing`, and then otherwise include.

@glwagner Can you check now? I think it's simpler now and it does what you suggested.

examples/langmuir_turbulence.jl
tomchor: No, you're right. This does run on GPUs since a `vertical_scale` and `Us` and defined and `const`s, which I had missed before.

However, I do think using `parameters` is better since using `const`s prevents a user from playing with this example in the same session by changing values.

I suggested new changes which uses `parameters` to run on GPUs and still simplifies things. lmk what you think.

src/StokesDrift.jl
glwagner: It's good for `P` to be first so we can dispatch on

```julia
const UnparameterizedUSD = UniformStokesDrift{<:Nothing}
```

src/StokesDrift.jl
glwagner: yes

examples/langmuir_turbulence.jl
glwagner: no, `const` is fine for GPU. This is a convenience but not a necessity

glwagner: I'm only hesitant because this design is different than forcing or boundary conditions. In those, `parameters` applies to a single function. Here the design is subtly different because `parameters` applies to every component of the stokes shear. It's not a huge deal but I wonder if it's example-worthy if the design is not final

src/StokesDrift.jl
tomchor: > It's good for `P` to be first so we can dispatch on
> 
> ```julia
> const UnparameterizedUSD = UniformStokesDrift{<:Nothing}
> ```

@glwagner I'm confused about this comment. I had put `P` first originally for this reason exactly (easier to dispatch) but changed because of your comment [here](https://github.com/CliMA/Oceananigans.jl/pull/3096#discussion_r1185557203). Are you suggesting I change it back?

examples/langmuir_turbulence.jl
tomchor: Okay, I'll revert it.

examples/langmuir_turbulence.jl
```diff
@@ -59,14 +59,13 @@ const vertical_scale = wavelength / 4π
...
+# To run this example on the GPU, one only needs to include `architecture = GPU()` in the
+# constructor for `RectilinearGrid` above.
 #
-# The Stokes drift profile is
+# We only need the vertical derivative of the Stokes drift, which is
 
-uˢ(z) = Uˢ * exp(z / vertical_scale)
+∂z_uˢ(z, t) = 1 / vertical_scale * Uˢ * exp(z / vertical_scale)
 
-# which we'll need for the initial condition.
```
tomchor, line 69: Note that I changed a couple of things in this example. For example `uˢ(z)` wasn't used anywhere and didn't need to be defined. Also the message above incorrectly said that the `GPU()` option went into the `NonhydrostaticModel` constructor, which isn't the case anymore.

src/StokesDrift.jl
glwagner: The order of type parameters and properties are independent. My comment was about properties, not type parameters.

examples/langmuir_turbulence.jl
glwagner: We have to define the Stokes drift. We could do it with a comment rather than code, but code is better and no less clear. Can you put it back?

src/StokesDrift.jl
tomchor: Ah, I see. Thanks for clarifying. It's done.

examples/langmuir_turbulence.jl
```diff
@@ -59,14 +59,17 @@ const vertical_scale = wavelength / 4π
 const Uˢ = amplitude^2 * wavenumber * frequency # m s⁻¹
 
 # The `const` declarations ensure that Stokes drift functions compile on the GPU.
-# To run this example on the GPU, write `architecture = GPU()` in the constructor
-# for `NonhydrostaticModel` below.
+# To run this example on the GPU, one only needs to include `architecture = GPU()` in the
```
glwagner, line 62:
```suggestion
# To run this example on the GPU, include `architecture = GPU()` in the
```

src/StokesDrift.jl
```diff
@@ -39,44 +39,48 @@ abstract type AbstractStokesDrift end
...
+@inline ∂t_vˢ(i, j, k, grid, sw::USDnoP, time) = sw.∂t_vˢ(znode(k, grid, Center()), time)
 
-@inline z_curl_Uˢ_cross_U(i, j, k, grid, sw::USD, U, time) = @inbounds begin (
-    - ℑxzᶜᵃᶠ(i, j, k, grid, U.u) * sw.∂z_uˢ(znode(k, grid, Face()), time)
-    - ℑyzᵃᶜᶠ(i, j, k, grid, U.v) * sw.∂z_vˢ(znode(k, grid, Face()), time) )
-end
+@inline x_curl_Uˢ_cross_U(i, j, k, grid, sw::USDnoP, U, time) = @inbounds    ℑxzᶠᵃᶜ(i, j, k, grid, U.w) * sw.∂z_uˢ(znode(k, grid, Center()), time)
+@inline y_curl_Uˢ_cross_U(i, j, k, grid, sw::USDnoP, U, time) = @inbounds    ℑyzᵃᶠᶜ(i, j, k, grid, U.w) * sw.∂z_vˢ(znode(k, grid, Center()), time)
+@inline z_curl_Uˢ_cross_U(i, j, k, grid, sw::USDnoP, U, time) = @inbounds (- ℑxzᶜᵃᶠ(i, j, k, grid, U.u) * sw.∂z_uˢ(znode(k, grid, Face()), time)
+                                                                           - ℑyzᵃᶜᶠ(i, j, k, grid, U.v) * sw.∂z_vˢ(znode(k, grid, Face()), time) )
```
glwagner, line 84:
```suggestion
                                                                           - ℑyzᵃᶜᶠ(i, j, k, grid, U.v) * sw.∂z_vˢ(znode(k, grid, Face()), time))
```

src/StokesDrift.jl
```diff
@@ -39,44 +39,48 @@ abstract type AbstractStokesDrift end
...
+@inline ∂t_uˢ(i, j, k, grid, sw::USDnoP, time) = sw.∂t_uˢ(znode(k, grid, Center()), time)
+@inline ∂t_vˢ(i, j, k, grid, sw::USDnoP, time) = sw.∂t_vˢ(znode(k, grid, Center()), time)
 
-@inline z_curl_Uˢ_cross_U(i, j, k, grid, sw::USD, U, time) = @inbounds begin (
-    - ℑxzᶜᵃᶠ(i, j, k, grid, U.u) * sw.∂z_uˢ(znode(k, grid, Face()), time)
-    - ℑyzᵃᶜᶠ(i, j, k, grid, U.v) * sw.∂z_vˢ(znode(k, grid, Face()), time) )
-end
+@inline x_curl_Uˢ_cross_U(i, j, k, grid, sw::USDnoP, U, time) = @inbounds    ℑxzᶠᵃᶜ(i, j, k, grid, U.w) * sw.∂z_uˢ(znode(k, grid, Center()), time)
+@inline y_curl_Uˢ_cross_U(i, j, k, grid, sw::USDnoP, U, time) = @inbounds    ℑyzᵃᶠᶜ(i, j, k, grid, U.w) * sw.∂z_vˢ(znode(k, grid, Center()), time)
+@inline z_curl_Uˢ_cross_U(i, j, k, grid, sw::USDnoP, U, time) = @inbounds (- ℑxzᶜᵃᶠ(i, j, k, grid, U.u) * sw.∂z_uˢ(znode(k, grid, Face()), time)
```
glwagner, line 83:
```suggestion
@inline z_curl_Uˢ_cross_U(i, j, k, grid, sw::USDnoP, U, time) = @inbounds (- ℑxzᶜᵃᶠ(i, j, k, grid, U.u) * sw.∂z_uˢ(znode(i, j, k, grid, Center(), Center(), Face()), time)
```

src/StokesDrift.jl
glwagner, line 84:
```suggestion
                                                                           - ℑyzᵃᶜᶠ(i, j, k, grid, U.v) * sw.∂z_vˢ(znode(k, grid, Face()), time))
```

test/test_stokes_drift.jl
```diff
@@ -1,10 +1,12 @@
...
+
 function instantiate_stokes_drift()
-    ∂t_uˢ(z, t) = exp(z/20) * cos(t)
-    ∂t_vˢ(z, t) = exp(z/20) * cos(t)
-    ∂z_uˢ(z, t) = exp(z/20) * cos(t)
-    ∂z_vˢ(z, t) = exp(z/20) * cos(t)
+    ∂t_uˢ(z, t, p) = exp(z/p.vertical_scale) * cos(t)
+    ∂t_vˢ(z, t, p) = exp(z/p.vertical_scale) * cos(t)
+    ∂z_uˢ(z, t, p) = exp(z/p.vertical_scale) * cos(t)
+    ∂z_vˢ(z, t, p) = exp(z/p.vertical_scale) * cos(t)
```
glwagner, line 7:
```suggestion
    ∂t_uˢ(z, t, h) = exp(z/h) * cos(t)
    ∂t_vˢ(z, t, h) = exp(z/h) * cos(t)
    ∂z_uˢ(z, t, h) = exp(z/h) * cos(t)
    ∂z_vˢ(z, t, h) = exp(z/h) * cos(t)
```

test/test_stokes_drift.jl
```diff
@@ -1,10 +1,12 @@
...
     stokes_drift = UniformStokesDrift(∂t_uˢ=∂t_uˢ, ∂t_vˢ=∂t_vˢ,
-                                      ∂z_uˢ=∂z_uˢ, ∂z_vˢ=∂z_vˢ)
+                                      ∂z_uˢ=∂z_uˢ, ∂z_vˢ=∂z_vˢ, parameters=(; vertical_scale = 20))
```
glwagner, line 9:
```suggestion
                                      ∂z_uˢ=∂z_uˢ, ∂z_vˢ=∂z_vˢ, parameters=20)
```

examples/langmuir_turbulence.jl
glwagner: I guess we don't use the kwarg `architecture` anymore

PR #3102, opened 2023-05-08
Simplify implementation of SmagorinskyLilly + some other closure doc fixes

glwagner:
I noticed it was a bit convoluted. I refactored the viscosity computation to hopefully be a bit more readable and understandable.

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_utils.jl
+++ b/src/TurbulenceClosures/turbulence_closure_utils.jl
@@ -26,9 +26,6 @@ function convert_diffusivity(FT, κ::NamedTuple; discrete_form=false, loc=(nothi
     return NamedTuple{κ_names}(Tuple(convert_diffusivity(FT, κi; discrete_form, loc, parameters) for κi in κ))
 end
 
-@inline geo_mean_Δᶠ(i, j, k, grid::AbstractGrid) =
-    cbrt(Δxᶜᶜᶜ(i, j, k, grid) * Δyᶜᶜᶜ(i, j, k, grid) * Δzᶜᶜᶜ(i, j, k, grid))
-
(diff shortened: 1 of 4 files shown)


navidcy:
Is the `SmagorinskyLilly` tested? If so I'm happy to approve this PR.

src/TurbulenceClosures/turbulence_closure_utils.jl
```diff
@@ -38,3 +35,4 @@ end
     i, j, k = @index(Global, NTuple)
     @inbounds κₑ[i, j, k] = calc_nonlinear_κᶜᶜᶜ(i, j, k, grid, closure, tracer, tracer_index, U)
 end
+
```
tomchor, line 38:
```suggestion
```

tomchor:
@glwagner nice. I think it's indeed clearer.

@navidcy I think the `SmagLilly` closure is tested in [time-stepping](https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_time_stepping.jl), [closures](https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_turbulence_closures.jl) and in [nonhydrostatic regression](https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_nonhydrostatic_regression.jl#L63-L72). But I think the only instance where we're testing actual values (and not just compilation of trivial (i.e. zero) values) is in 

https://github.com/CliMA/Oceananigans.jl/blob/8b42137e4224b86fe1f3f1744bd5076aa8821085/test/test_nonhydrostatic_regression.jl#L63-L72

Which does seem to pass, so I'm approving this.

tomchor (approved):

src/TurbulenceClosures/turbulence_closure_utils.jl
navidcy: Damn I missed this.

PR #3103, opened 2023-05-10
Mention native coords in `ScalarDiffusivity`/`ScalarBiharmonicDiffusivity` docs

navidcy:


src/TurbulenceClosures/turbulence_closure_implementations/scalar_diffusivity.jl
```diff
@@ -35,10 +35,20 @@ Arguments
...
+
+* `discrete_form`: `Boolean`; default: `False`.
+
+When prescribing the viscosities or diffusivities as functions, depending on the value of keyword argument
+`discrete_form`, the constructor expects:
+
+  - `discrete_form = false` (default): functions of the grid's native coordinates and time, e.g., `(x, y, z, t)` for a
+    `RectilinearGrid` or `(λ, φ, z, t)` for a `LatitudeLongitudeGrid`.
+
+  - `discrete_form = true`: functions of `(i, j, k, grid, ℓx, ℓy, ℓz)` with `ℓx`, `ℓy` and `ℓz` either `Face()` or `Center()`.
```
tomchor, line 51: I've never used the discrete form so I'm not sure how it works here, but how do the `ℓx`, `ℓy` and `ℓz` arguments work? Do we have consider all face combinations when providing the function? Should this be also explained here?

navidcy: You need to prescribe where you want the diffusivity to be computed. That’s what I understand. 

What do you mean by “consider all combinations”? Yes, all 8 combinations are in principle allowed. Is this what you mean?

tomchor: > What do you mean by “consider all combinations”? Yes, all 8 combinations are in principle allowed. Is this what you mean?

I mean more as is: are all combinations _required_ the be defined?

navidcy: Nah. Only the one you want.

navidcy: OK, I see what you mean! I added an example which I think clarifies it?

tomchor (approved):

glwagner (approved):

PR #3107, opened 2023-05-14
Fix small typoe in `tilted_bottom_boundary_layer.jl`

tomchor:
I'm also making a minor suggestion that instead indicating the grid dimensions with a comment, we do it with code.

### DIFF
--- a/examples/tilted_bottom_boundary_layer.jl
+++ b/examples/tilted_bottom_boundary_layer.jl
@@ -22,13 +22,13 @@
 #
 # ## The domain
 #
-# We create a ``400 × 100`` meter ``x, z`` grid with ``128 × 32`` cells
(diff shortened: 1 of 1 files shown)


examples/tilted_bottom_boundary_layer.jl
```diff
@@ -22,13 +22,13 @@
 #
 # ## The domain
 #
-# We create a ``400 × 100`` meter ``x, z`` grid with ``128 × 32`` cells
-# and finer resolution near the bottom,
+# We create an grid with finer resolution near the bottom,
```
navidcy, line 25:
```suggestion
# We create a grid with finer resolution near the bottom,
```

navidcy (approved):

PR #3111, opened 2023-05-22
Allowing `NonhydrostaticModel` with grids that are irregular in either `x` or `y`

tomchor:
The goal of this is PR to allow `NonhydrostaticModel` to have grids that are irregularly-spaced in either the `x` or the `y` directions (in addition to grids that are irregularly-spaced in the `z`, which we already allow).

The first step was https://github.com/CliMA/Oceananigans.jl/pull/3117, which is now merged. Now the rest of the work is to generalize `FourierTridiagonalPoissonSolver` to other dimensions.

Closes https://github.com/CliMA/Oceananigans.jl/issues/3116


tomchor:
cc @zhazorken

src/Solvers/fourier_tridiagonal_poisson_solver.jl
```diff
@@ -12,55 +13,94 @@ end
...
+
+    getindex(topology(grid), irregular_dimension(grid)) != Bounded && error("`FourierTridiagonalPoissonSolver` can only be used when the irregular direction's topology is `Bounded`.")
 
     Nx, Ny, Nz = size(grid)
 
     # Compute discrete Poisson eigenvalues
-    λx = poisson_eigenvalues(grid.Nx, grid.Lx, 1, TX())
-    λy = poisson_eigenvalues(grid.Ny, grid.Ly, 2, TY())
+    λ1 = poisson_eigenvalues(regular_siz1, regular_ext1, 1, regular_top1())
+    λ2 = poisson_eigenvalues(regular_siz2, regular_siz2, 2, regular_top2())
```
tomchor, line 80:
```suggestion
    λ2 = poisson_eigenvalues(regular_siz2, regular_ext2, 2, regular_top2())
```

glwagner:
I think we want to do this over a few PRs. The first task is to extend BatchedTridiagonalSolver

src/Models/NonhydrostaticModels/NonhydrostaticModels.jl
```diff
@@ -21,14 +22,16 @@ function PressureSolver(arch::DistributedArch, local_grid::RegRectilinearGrid)
...
+PressureSolver(arch, grid::XYRegRectilinearGrid) = FourierTridiagonalPoissonSolver(grid)
+PressureSolver(arch, grid::XZRegRectilinearGrid) = FourierTridiagonalPoissonSolver(grid)
+PressureSolver(arch, grid::YZRegRectilinearGrid) = FourierTridiagonalPoissonSolver(grid)
 
 # *Evil grin*
 PressureSolver(arch, ibg::ImmersedBoundaryGrid) = PressureSolver(arch, ibg.underlying_grid)
 
 # fall back
-PressureSolver(arch, grid) = error("None of the implemented pressure solvers for NonhydrostaticModel support horizontally-stretched grids.")
+PressureSolver(arch, grid) = error("None of the implemented pressure solvers for NonhydrostaticModel support both than one irregularly-spaced direction.")
```
glwagner, line 34:
```suggestion
PressureSolver(arch, grid) = error("None of the implemented pressure solvers for NonhydrostaticModel support more than one irregularly-spaced direction.")
```

src/Solvers/fourier_tridiagonal_poisson_solver.jl
```diff
@@ -25,7 +25,7 @@ architecture(solver::FourierTridiagonalPoissonSolver) = architecture(solver.grid
     D[N, m, n] = -1 / Δxᶠᵃᵃ(N, m, n, grid) - Δxᶜᵃᵃ(N, m, n, grid) * (λ1[m] + λ2[n])
 end 
 
-@kernel function compute_main_diagonals!(D, grid, λ1, λ2, dir::Val{2})
+@kernel function compute_main_diagonals!(D, grid, λ1, λ2, ::YDirection)
     m, n = @index(Global, NTuple)
     N = getindex(size(grid), 2)
```
glwagner:
```suggestion
    i, k = @index(Global, NTuple)
    Ny = getindex(size(grid), 2)
```

is that right?

tomchor: In this PR (which precedes https://github.com/CliMA/Oceananigans.jl/pull/3117) I'm still using the notation `N`, `m`, `n`, `q` instead of `Nx`, `Ny`, `Nz`, `i`, `j`, `k`. But I'll change it later to what you suggested.

tomchor:
I marked this as ready for review because I believe this is now working except for the case where one of the regularly-spaced dimensions is `Bounded` on a `GPU()`. I can't seem to get the transform plans right in that case and would appreciate pointers. Everything else should be fine though.

tomchor:
Is this line necessary? https://github.com/CliMA/Oceananigans.jl/blob/65d5c66b756cb8e8cf50eda821ba9e0db71f3161/src/Solvers/fourier_tridiagonal_poisson_solver.jl#L129

A few lines later we specifically copy only the real component of the array:
https://github.com/CliMA/Oceananigans.jl/blob/65d5c66b756cb8e8cf50eda821ba9e0db71f3161/src/Solvers/fourier_tridiagonal_poisson_solver.jl#L137

I removed it and didn't notice any difference in the output.

src/Solvers/plan_transforms.jl
```diff
@@ -11,7 +11,7 @@
 ##### efficient transforms. `A` will be mutated.
 #####
 
-using Oceananigans.Grids: XYRegRectilinearGrid, XZRegRectilinearGrid, YZRegRectilinearGrid, irregular_dimension
+using Oceananigans.Grids: XYRegRectilinearGrid, XZRegRectilinearGrid, YZRegRectilinearGrid, regular_dimensions, irregular_dimension
```
glwagner, line 14: We previously used "regular" and "stretched" to refer to these concepts. Have we decided to change nomenclature to "irregular" instead of "stretched"? I think it's ok to change terminology but it should be done uniformly throughtout the code rather than "by fiat" in a particular PR. The docs use "stretched grid" quite a bit so I think we would have to change this to "irregular grid"...

glwagner: Another note is that here we use "dimension" rather than "direction" as we use elsewhere. Do we want to use "dimension", or "direction"? We should pick one.

The function `irregular_dimension` is a little too specific I think, because grids can have between 0 and 3 irregular dimensions. For example, we have "regular_dimensions" --- not "regular_dimension" (the latter would only apply to a grid with a single regular dimension).

It makes sense to me to have two paired functions "regular" and "irregular" (or stretched), but they should both return _all_ of the dimensions that are either regular or stretched.

tomchor: Good points. I'm agnostic wrt stretch vs irregular grids, so I guess I'll revert back to stretched.

I'm only using `regular_dimension()` because there was already a function called just that, and I didn't wanna define a new one. However, I did try to follow the pattern where `direction` returns `XDirection` and so on, while `dimension` returns a number (3 and so on). I do think this distinction is good, but I'm open to ideas. Maybe we could use `axes`  when a number is returned and `directions` when an actual direction (e.g. `XDirection`) is returned?

And also good point about `regular_dimension`. I'll change it to plural.

glwagner: I don't mind "stretched". I also just realized that `show` uses "regularly spaced" and "variably spaced" 😂 

Anyways I think we should avoid introducing a 3rd term and probably `stretched_axes` is good. I didn't realize that `dimensions` meant it would return a number...

src/Solvers/plan_transforms.jl
```diff
@@ -174,45 +168,49 @@ function plan_transforms(grid::VerticallyStretched, storage, planner_flag)
...
         unflattened_topo = Tuple(T() isa Flat ? Bounded : T for T in topo)
         f_order = forward_orders(unflattened_topo...)
         b_order = backward_orders(unflattened_topo...)
 
-        # Extract third dimension
-        f_order = Tuple(f_order[i] for i in findall(d -> d != 3, f_order))
-        b_order = Tuple(b_order[i] for i in findall(d -> d != 3, b_order))
+        # Extract irregularly-spaced dimension
+        f_order = Tuple(f_order[i] for i in findall(d -> d != irreg_dim, f_order))
+        b_order = Tuple(b_order[i] for i in findall(d -> d != irreg_dim, b_order))
```
tomchor, line 213: I think this clause is the only thing left for this PR to be ready for review. However I'm having a hard time getting this right since I'm not very familiar with the transforms and for some reason `Revise` doesn't work for me when I'm on the GPU nodes. I've asked on slack about it, but so far I haven't been able to fix it, which means my workflow for GPU development right now is super slow...

I'd appreciate some help/insight here if possible.

@simone-silvestri I believe you've worked on similar code recently, no? Any tips/suggestions?

src/Solvers/plan_transforms.jl
```diff
@@ -174,45 +168,49 @@ function plan_transforms(grid::VerticallyStretched, storage, planner_flag)
...
+    elseif bounded_dims == ()
+        # We're on the GPU and either (Periodic, Periodic), (Flat, Periodic), or
+        # (Periodic, Flat) in the regular dimensions. So, we pretend like we need a 2D
+        # doubly-periodic transform (even if one dimension is Flat).
+        forward_periodic_plan = plan_forward_transform(storage, Periodic(), reg_dims, planner_flag)
+        backward_periodic_plan = plan_backward_transform(storage, Periodic(), reg_dims, planner_flag)
 
-        forward_transforms = tuple(DiscreteTransform(forward_periodic_plan, Forward(), grid, [1, 2]))
-        backward_transforms = tuple(DiscreteTransform(backward_periodic_plan, Backward(), grid, [1, 2]))
+        forward_transforms = tuple(DiscreteTransform(forward_periodic_plan, Forward(), grid, reg_dims))
```
simone-silvestri, line 189: can you batch here if you have reg_dims = [1, 3]? probably you cannot and have to handle this case as the one below

src/Solvers/plan_transforms.jl
simone-silvestri: I am not sure you always have to reshape to put the irregular direction at the end. You probably just have to expose the transform direction

src/Solvers/plan_transforms.jl
tomchor: So far this seems to be working and the only problems seem to happen with the case below. But I'll test more thoroughly

tomchor:
I think this PR is finally ready. Thanks to @simone-silvestri for the help!

Here a simple simulation (adapted from the [ocean wind mixing example](https://clima.github.io/OceananigansDocumentation/stable/generated/ocean_wind_mixing_and_convection/)) using stretched grids in the x, y, and z directions respectively:

https://github.com/CliMA/Oceananigans.jl/assets/13205162/fba0fa71-92b0-4f7b-978e-7bb24f7060fb

https://github.com/CliMA/Oceananigans.jl/assets/13205162/dea0309f-3b2f-499a-81b2-f5de819dc992

https://github.com/CliMA/Oceananigans.jl/assets/13205162/17e2c3e1-71be-4d9a-9bbe-378bbb657150

src/Solvers/fourier_tridiagonal_poisson_solver.jl
```diff
@@ -1,72 +1,114 @@
...
+    @unroll for j in 2:Ny-1
+        D[i, j, k] = - (1 / Δyᵃᶠᵃ(i, j+1, k, grid) + 1 / Δyᵃᶠᵃ(i, j, k, grid)) - Δyᵃᶜᵃ(i, j, k, grid) * (λx[i] + λz[k])
+    end
+    D[i, Ny, k] = -1 / Δyᵃᶠᵃ(i, Ny, k, grid) - Δyᵃᶜᵃ(i, Ny, k, grid) * (λx[i] + λz[k])
+end 
+
+@kernel function compute_main_diagonal!(D, grid, λx, λy, ::ZDirection)
     i, j = @index(Global, NTuple)
-    Nz = grid.Nz
+    Nz = getindex(size(grid), 3)
```
simone-silvestri, line 43:
```suggestion
    Nz = size(grid, 3)
```

src/Solvers/fourier_tridiagonal_poisson_solver.jl
```diff
@@ -1,72 +1,114 @@
...
 
-struct FourierTridiagonalPoissonSolver{G, B, R, S, β, T}
+struct FourierTridiagonalPoissonSolver{G, B, R, S, β, T, D}
                           grid :: G
     batched_tridiagonal_solver :: B
                    source_term :: R
                        storage :: S
                         buffer :: β
                     transforms :: T
+         tridiagonal_direction :: D
```
simone-silvestri, line 12:
```suggestion
```

we don't need this

src/Solvers/fourier_tridiagonal_poisson_solver.jl
```diff
@@ -1,72 +1,114 @@
-using Oceananigans.Operators: Δzᵃᵃᶜ, Δzᵃᵃᶠ
+using Oceananigans.Operators: Δxᶜᵃᵃ, Δxᶠᵃᵃ, Δyᵃᶜᵃ, Δyᵃᶠᵃ, Δzᵃᵃᶜ, Δzᵃᵃᶠ
+using Oceananigans.Grids: XYRegRectilinearGrid, XZRegRectilinearGrid, YZRegRectilinearGrid, stretched_dimensions
 import Oceananigans.Architectures: architecture
 
-struct FourierTridiagonalPoissonSolver{G, B, R, S, β, T}
+struct FourierTridiagonalPoissonSolver{G, B, R, S, β, T, D}
```
simone-silvestri, line 5:
```suggestion
struct FourierTridiagonalPoissonSolver{G, B, R, S, β, T}
```

simone-silvestri (approved):

src/Solvers/fourier_tridiagonal_poisson_solver.jl
tomchor: Fixed by https://github.com/CliMA/Oceananigans.jl/pull/3111/commits/1aa148d22c9764142ac7998b14329deaea5f5655

src/Solvers/fourier_tridiagonal_poisson_solver.jl
tomchor: Thanks, good catch. It was a leftover that I forgot to change back. Fixed by https://github.com/CliMA/Oceananigans.jl/pull/3111/commits/8f6c5410fab5183f0773528d6217156c7819ce58

tomchor: @simone-silvestri it turns out that we need `tridiagonal_direction` in the module `Models`. While technically we could get rid of `tridiagonal_direction` and just call `stretched_direction(solver.grid)` everywhere that's needed, I think this makes the code more complicated.

So if it's okay with you, I think it's best to keep that property since it makes the code simpler and easier to understand. Especially since the name "`TridiagonalPoissonSolver`" implies there's a "`tridiagonal_direction`" associated with it. So I changed this back in https://github.com/CliMA/Oceananigans.jl/pull/3111/commits/9507a1640656865845b7a27d3c2f5384a6a608d4. Let me know if you disagree.

simone-silvestri: is that not a property of the `batched_tridiagonal_solver` (i.e. `solver.batched_tridiagonal_solver.tridiagonal_direction`)? 
We could define
```
tridiagonal_direction(solver::TridiagonalPoissonSolver) = solver.batched_tridiagonal_solver.tridiagonal_direction
```
and pass it as a function. 
I would rather that than duplicating the information, but maybe it's just personal preference

tomchor: > is that not a property of the `batched_tridiagonal_solver` (i.e. `solver.batched_tridiagonal_solver.tridiagonal_direction`)? We could define
>
> ```
> tridiagonal_direction(solver::TridiagonalPoissonSolver) = solver.batched_tridiagonal_solver.tridiagonal_direction
> ```
> 
> and pass it as a function. I would rather that than duplicating the information, but maybe it's just personal preference

I guess it does come down to a personal preference. I'd rather have the information there in `FourierTridiagonalPoissonSolver`, than defining extra functions (however simple they may be). But it is a good point that that property can be accessed from `solver.batched_tridiagonal_solver`.

https://github.com/CliMA/Oceananigans.jl/pull/3111/commits/215d73160dabf7cff2f520be7352dd3c6e23207b removed `tridiagona_direciton` from the fourier solver, like you suggested, and accesses it via `solver.batched_tridiagonal_solver` when necessary. I think the code is indeed better that way :+1:

PR #3113, opened 2023-05-22
Find output files recursively after a doc build

### LINKED ISSUE #3112: Find a way to recursively search for `.nc`/`.jld2` files within the `docs` directory
At the moment we hardcode the locations where output files in docs builds are so we can delete them before deploying the docs. See:

https://github.com/CliMA/Oceananigans.jl/blob/4551a78b1f3fe4bb3b238676c128dc751be9b934/docs/make.jl#L165-L169

We should find a way to automatically search for output files recursively in the `docs` directory and delete them before deploying the docs.

I've asked in Julia slack how to do this and some ideas include:

```Julia
julia> function recursive_find(dir, pattern)
           mapreduce(vcat, walkdir(dir)) do (root, dirs, files)
               joinpath.(root, filter(contains(pattern), files))
           end
       end
recursive_find (generic function with 1 method)

julia> recursive_find(DEPOT_PATH[end], r"\.jl$")
988-element Vector{String}:
 "/opt/julia/julia-1.9.0/share/julia/julia-config.jl"
 "/opt/julia/julia-1.9.0/share/julia/base/Base.jl"
 "/opt/julia/julia-1.9.0/share/julia/base/Enums.jl"
 "/opt/julia/julia-1.9.0/share/julia/base/abstractarray.jl"
 ⋮
 "/opt/julia/julia-1.9.0/share/julia/test/testhelpers/llvmpasses.jl"
 "/opt/julia/julia-1.9.0/share/julia/test/testhelpers/withlocales.jl"
 "/opt/julia/julia-1.9.0/share/julia/test/unicode/utf8.jl"
```

navidcy:
Resolves #3112

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -176,7 +176,7 @@ files = []
 for pattern in [r".jld2", r".nc"]
     files = vcat(files, recursive_find(@__DIR__, pattern))
 end
-
+@show files
 for file in files
     rm(file)
 end
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -3,7 +3,6 @@ pushfirst!(LOAD_PATH, joinpath(@__DIR__, "..")) # add Oceananigans to environmen
 using Documenter
 using DocumenterCitations
 using Literate
-using Glob
 
 using CairoMakie # to avoid capturing precompilation output by Literate
 CairoMakie.activate!(type = "svg")
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -162,8 +162,13 @@ makedocs(bib,
 
 @info "Cleaning up temporary .jld2 and .nc output created by doctests or literated examples..."
 
-recursive_find(dir, pattern) =
-    mapreduce(vcat, walkdir(dir)) do (root, dirs, files)
+"""
+    recursive_find(directory, pattern)
+
+Return list of filepaths within `directory` that fall under the `pattern::Regex`, e.g., `pattern = r"\.jl"`.
+"""
+recursive_find(directory, pattern) =
+    mapreduce(vcat, walkdir(directory)) do (root, dirs, files)
         joinpath.(root, filter(contains(pattern), files))
     end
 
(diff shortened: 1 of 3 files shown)


docs/make.jl
```diff
@@ -162,11 +161,23 @@ makedocs(bib,
 
 @info "Cleaning up temporary .jld2 and .nc output created by doctests or literated examples..."
 
-for file in vcat(glob("docs/*.jld2"),
-                 glob("docs/*.nc"),
-                 glob("docs/build/generated/*.jld2"),
-                 glob("docs/build/generated/*.nc")
-                 )
```
tomchor, line 169: Honestly I prefer the old code. It's not as elegant a solution as this PR is proposing, but it's way clearer and more concise. I believe it also achieves the same result, no?

glwagner (approved):

docs/make.jl
glwagner: Right. The code is only the same under a very specific assumption.

navidcy:
seems like the docs error on buildkite but locally it works -- strange.

docs/make.jl
```diff
@@ -162,11 +161,23 @@ makedocs(bib,
...
+Return list of filepaths within `directory` that fall under the `pattern::Regex`, e.g., `pattern = r"\.jl"`.
+"""
+recursive_find(directory, pattern) =
+    mapreduce(vcat, walkdir(directory)) do (root, dirs, files)
+        joinpath.(root, filter(contains(pattern), files))
+    end
+
+files = []
+
+for pattern in [r"\.jld2", r"\.nc"]
```
tomchor, line 176: The error happens with escaping, so maybe the code below will work (works for me locally)

```suggestion
recursive_find(directory, pattern) =
    mapreduce(vcat, walkdir(directory)) do (root, dirs, files)
        joinpath.(root, filter(endswith(pattern), files))
    end

files = []

for pattern in [r".jld2", r".nc"]
```

navidcy: Thanks!!

navidcy:
> Same error... no idea what's happening since there are no more scaped characters. Would changing `r".jld2"` to `raw".jld2"` do anything?

error is in line 168 so I'm wondering if it's simply the docstring!

navidcy:
Let's see if this works. If so, I'll bring the examples back. :)

PR #3121, opened 2023-05-25
Bump compat entries for PencilFFTs and PencilArrays

navidcy:
Supercedes #3119 #3120


glwagner (approved):

glwagner:
Do these support `CuArray`?

navidcy:
Think so. GPU tests passed, only docs are failing now (but I think they just failed to start... I'm restarting!)

glwagner:
I don't think we use PencilArray for CuArray either in the source code or tests. My question is: does bumping PencilArrays and PencilFFTs mean that we can extend the distributed FFT algorithm (currently only implemented on CPU) to GPUs?

navidcy:
Oh I see. This was not at all clear to me from your previous question. I thought you were asking whether the new versions dropped support to CuArrays and therefore explain why tests were failing!

Let me get back on your question now that I understand!

navidcy:
I doubt it... I understood that the main change in the packages is https://github.com/jipolanco/PencilFFTs.jl/pull/65

navidcy:
But then this issue (https://github.com/jipolanco/PencilFFTs.jl/issues/56) implies that already these packages work for GPUs over distributed. No?

glwagner:
Yes, I think they do. I was wondering if the update here brought in that change (another possibility is that it's already supported)

PR #3124, opened 2023-06-01
Initialize `SplitExplicitFreeSurface` with a CFL condition

simone-silvestri:
This PR allows initializing the `SplitExplicitFreeSurface` with the following keyword arguments:

```julia
free_surface = SplitExplicitFreeSurface(; CFL = 0.7, Δt_max = 10minutes, grid)
```

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
@@ -50,8 +50,8 @@ Keyword Arguments
 
 - `timestepper`: Time stepping scheme used, either `ForwardBackwardScheme()` or `AdamsBashforth3Scheme()`.
 """
-SplitExplicitFreeSurface(; gravitational_acceleration = g_Earth, kwargs...) =
-    SplitExplicitFreeSurface(nothing, nothing, nothing, gravitational_acceleration, SplitExplicitSettings(; kwargs...))
+SplitExplicitFreeSurface(; gravitational_acceleration = g_Earth, kwargs...) = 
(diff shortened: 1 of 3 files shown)


glwagner:
`TimeStepWizard` uses `cfl` rather than `CFL`. So either we should change the wizard or use `cfl` here

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -247,9 +247,32 @@ Return `SplitExplicitSettings`. For a description of the keyword arguments, see
 the [`SplitExplicitFreeSurface`](@ref).
 """
 function SplitExplicitSettings(FT::DataType=Float64;
-                               substeps = 200, 
+                               substeps = nothing, 
+                               CFL    = nothing,
```
glwagner, line 251:
```suggestion
                               cfl    = nothing,
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -247,9 +247,32 @@ Return `SplitExplicitSettings`. For a description of the keyword arguments, see
...
+                               grid   = nothing,
+                               Δt_max = nothing,
```
glwagner, line 253:
```suggestion
                               max_Δt = nothing,
```

navidcy (approved):
Would it be useful to inform the user regarding how many substeps their cfl choice implied?

simone-silvestri:
`show(io::IO, free_surface::SplitExplicitFreeSurface)` shows the number of substeps, 
so that information can be retrieved with `@show free_surface` (or `@show model`)

PR #85, opened 2019-02-26
Support for user-defined forcing functions.

ali-ramadhan:
Just hacked something together that allows for user-defined forcing functions for the CPU. Have not tested on the GPU yet.

Basically there's a struct `Forcing` that stores the user-defined forcing functions. It will replace the old `ForcingFields` struct. See `examples/deep_convection_3d.jl` for how I switched to using a forcing function for T to enforce a cooling surface heat flux.

A big issue is that the current implementation slows down the time stepping by a factor of 2-3x. So we'll have to figure out why before merging.

The function must have a signature like `F(u, v, w, T, S, Nx, Ny, Nz, Δx, Δy, Δz, i, j, k)` right now so this won't produce a nice solution as we will have to figure out #59 before the function signature can look as nice as `surface_cooling_disk(grid, velocities, tracers, i, j, k)`. This is work for another branch.

Will keep working on this before merging. Just wanted to start something.

Resolves #73

PR #3125, opened 2023-06-01
(0.88.0) MPI communication and computation overlap in the `HydrostaticFreeSurfaceModel` and `NonhydrostaticModel`

### LINKED ISSUE #615: Overlapping computation and MPI halo communication

### LINKED ISSUE #1882: Multi GPU scaling is very poor

### LINKED ISSUE #3067: Abstraction for configuring kernels
@simone-silvestri's recent work on scalable performance for the `HydrostaticFreeSurfaceModel` used the concept of "inner" and "boundary" regions to overlap communication and computation.

The basic components are:

1) Reduce communication as much as possible (for example, compute diagnostic variables wherever needed rather than communicating results)
2) Compute the "interior" of the tendencies first, which do not require halo regions, while communication is occurring
3) Once communication is finished, compute the "boundary regions" of the tendency

Point (1) requires us to compute diagnostic variables inside halo regions. This requires both expanded sizes and offsets for each diagnostic variable. Different diagnostic variables have different sizes --- the width of the diagnostics kernels depends on how many points are accessed.

Points (2) and (3) require computations over parts of the grid (reduced size + offset).

These considerations demand a nice abstraction so that the code for launching kernels is readable and understandable (and also concise).

### LINKED ISSUE #3068: Offsetting kernel indices
Increasingly we have the need to offset kernel indices; ie to perform computations over "windows" for incomplete `Field` `indices`:

https://github.com/CliMA/Oceananigans.jl/blob/be00e364f9dcd712b3d0c3d48e32b94b181a02fc/src/AbstractOperations/computed_field.jl#L84-L92

or for @simone-silvestri's work to overlap communication and computation, eg #3067.

Possibly, such a feature can be supported by `KernelAbstractions` eg https://github.com/JuliaGPU/KernelAbstractions.jl/issues/384 which would simplify our code a lot (since we won't have to pass the offsets into the kernel explicitly).

simone-silvestri:
Overlapping MPI communication and computation in the HydrostaticFreeSurfaceModel and the NonhydrostaticModel.

In particular, this PR introduces two keyword arguments for the `fill_halo_regions!` function, active **only** in case of distributed halo passing boundary conditions:
- the`async::Bool` keyword argument that allows launching MPI operations without waiting for the communication to complete.
- the `only_local_halos::Bool` keyword argument, which fills only the halo in case of a local (i.e., Flux, Value, Gradient, Periodic, and, temporarily, MultiRegionCommunication)  boundary condition. This is required for having explicit boundary conditions (like Value or Flux) for turbulent diffusivities (we directly calculate diffusivities in the halos in the case of distributed BC).

This PR allows hiding MPI passing of barotropic auxiliary variables behind the implicit vertical solver and prognostic variables behind the tendency calculations. The latter is done by splitting the tendency kernels into an interior kernel that calculates tendencies between, e.g., `i = Hx` and `i = Nx - Hx`, and a boundary kernel, executed once communication is complete, that calculates tendencies adjacent to boundaries.
Before computing tendencies near the boundaries, boundary-adjacent auxiliary diagnostic variables are recalculated (hydrostatic pressure, vertical velocity and diffusivities for the hydrostatic model, hydrostatic pressure and diffusivities for the non-hydrostatic model)

Also, this PR introduces 
- 8-way halo passing for distributed grids (west, east, north, south, southwest, southeast, northwest, northeast), something that we will want to implement also for `MultiRegion`)
- Non uniform partitioning for distributed grids
- the `KernelParameters(size, offset)` to be passed to the `launch!` function to start a kernel of size ```size::Tuple``` offset by ```offset::Tuple```

todo:
- [ ] Implement Offsets in KernelAbstractions (at the moment implemented a fix in `src/Utils/kernel_launching.jl` to remove when offsets will be implemented)
- [x] Adapt that implementation in this PR
- [x] Generalize the current implementation of kernel offsets for Diffusivities, Pressures and W-velocity
- [x] Find a general way to tag MPI requests to allow large number of cores (currently `Nranks < 100`)
- [x] Remove `views` halo passing

#### API changes
- When calling `RectilinearGrid(arch::DistributedArch, size = (Nx, Ny, Nz), ....)`, `(Nx, Ny, Nz)` are the _per-rank_ local sizes, not the global size to be divided (easy way to specify non-uniform partitioning, see `validation/distributed/mpi_geostrophic_adjustment.jl`)
- added the `enable_overlapped_communication` keyword to `DistributedArch` (defaults to true)
- removed the `use_buffers` keyword to `DistributedArch` (_always_ use buffers, as views did not give significant speedup to justify maintaining two implementations)
- added the keyword argument `active_cells_map::Bool = false` to `ImmersedBoundaryGrid` (ex: `ImmersedBoundaryGrid(grid, ib, active_cells_map = true)`
- added a `required_halo_size::Int ` keyword argument to `ScalarDiffusivity` (defaults to 1) and `ScalarBiharmonicDiffusivity` (defaults to 2) to be specified by the user which sets the required halo for the specific `ν` or `κ` function (closures have now an explicitly required number of halos)
#### **Major** internals change
1) The tendencies are calculated at the **end** of a time step. Therefore at the end of a simulation `model.timestepper` will hold tendencies for the last time step completed 

2) Removed `fill_halo_regions!` for hydrostatic pressure in both the non-hydrostatic and the hydrostatic model and for w-velocity in the hydrostatic model. The halos are filled by enlarging the size of the kernels in `update_hydrostatic_pressure!` and `compute_w_from_continuity!` to incorporate the needed ghost points.

2) Removed `fill_halo_regions!` for diffusivities **only for halo-passing BC**; the halo calculation is now performed by launching the `calculate_diffusivity!` kernel inside the ghost nodes before recomputing the tendencies. This requires knowing how many halos each closure requires.

3) Added a _required_halo_ parameter to `AbstractTurbulenceClosure`. This means that each parameterization will have to specify explicitly the number of halos required to calculate the diffusivity: 
e.g 
```julia
abstract type AbstractTurbulenceClosure{TimeDiscretization, BoundaryBuffer} end
abstract type AbstractScalarDiffusivity{TD, F, N} <: AbstractTurbulenceClosure{TD, N} end
struct TwoDimensionalLeith{FT, CR, GM, M} <: AbstractScalarDiffusivity{ExplicitTimeDiscretization, ThreeDimensionalFormulation, 2}
``` 
Where Leith closure requires 2 halos (one for the vorticity calculation and an additional one for the vorticity derivative)

#### **Minor** internals change
removed the general `calculate_nonlinear_viscosity!` and `calculate_nonlinear_diffusivity!` kernels (to each turbulence closure their own kernel)

Requires https://github.com/JuliaGPU/KernelAbstractions.jl/pull/399 # On hold at the moment
Closes #615
Closes #1882
Closes #3067
Closes #3068 
Supersedes #2953


navidcy:
definitely bump a release (minor or patch; whatever more appropriate) before this is merged :)

navidcy:
Perhaps we wait for https://github.com/JuliaGPU/KernelAbstractions.jl/pull/399 to be merged and tag a patch release for KA before we merge this? @vchuravy?

simone-silvestri:
Sure! Anyways, this PR should be in its final form, so ready for review :) (will still wait for KA changes before merging)

simone-silvestri:
So, it looks like docs take more to build in this PR when compared to main (4:00 hrs vs 3:40 hrs). Probably good to benchmark a bit. I ll try some benchmarking on GPU and CPU

navidcy:
benchmarking sounds good!

but I thought this PR only made changes to the distributed grids, no?

simone-silvestri:
It also rearranges how time-stepping works by rendering operations to allow 
- minimal fill_halo_regions
- overlapped communication and computation
and removes `fill_halo_regions!` for computation of w velocity, hydrostatic pressure and diffusivities by enlarging the kernels to include the halo points

Manifest.toml
```diff
@@ -106,7 +106,7 @@ version = "4.6.1"
 [[deps.CompilerSupportLibraries_jll]]
 deps = ["Artifacts", "Libdl"]
 uuid = "e66e0078-7015-5450-92f7-15fbd957f2ae"
-version = "1.0.1+0"
+version = "0.5.2+0"
```
navidcy, line 109: is the required or just random chance?

src/Distributed/interleave_comm_and_comp.jl
```diff
@@ -0,0 +1,63 @@
+using Oceananigans: prognostic_fields
+using Oceananigans.Grids: halo_size
+
+function complete_communication_and_compute_boundary!(model, ::DistributedGrid, arch)
+
+    # We iterate over the fields because we have to clear _ALL_ architectures
+    # and split explicit variables live on a different grid
```
navidcy, line 7:
```suggestion
    # We iterate over the fields because we need to clear _ALL_ architectures
    # and split-explicit variables live on a different grids
```

src/Distributed/interleave_comm_and_comp.jl
```diff
@@ -0,0 +1,63 @@
...
+        arch.mpi_tag[1] -= arch.mpi_tag[1]
+    
+        # Reset MPI requests
+        empty!(arch.mpi_requests)
+    end
+    
+    recv_from_buffers!(field.data, field.boundary_buffers, field.grid)
+    
+    return nothing
+end
```
navidcy, line 63:
```suggestion
end

```

src/Distributed/multi_architectures.jl
```diff
@@ -59,11 +60,12 @@ Keyword arguments
...
-                         ranks,
-                         use_buffers = false,
-                         devices = nothing, 
-                         communicator = MPI.COMM_WORLD)
+                   topology = (Periodic, Periodic, Periodic), 
+                   ranks,
+                   use_buffers = true,
+                   devices = nothing, 
+                   enable_overlapped_computation = true,
+                   communicator = MPI.COMM_WORLD)
```
navidcy, line 68:
```suggestion
                         topology = (Periodic, Periodic, Periodic), 
                         ranks,
                         use_buffers = true,
                         devices = nothing, 
                         enable_overlapped_computation = true,
                         communicator = MPI.COMM_WORLD)
```

src/Distributed/multi_architectures.jl
```diff
@@ -140,17 +147,21 @@ end
...
-     south :: S
-       top :: T
-    bottom :: B
+struct RankConnectivity{E, W, N, S, T, B, SW, SE, NW, NE}
+         east :: E
+         west :: W
+        north :: N
+        south :: S
+          top :: T
+       bottom :: B
```
navidcy, line 156: do we split the domain in the vertical? why do we need top/bottom connectivity?

src/Distributed/partition_assemble.jl
```diff
@@ -1,32 +1,84 @@
...
+all_reduce(val, arch::DistributedArch; op = +) = 
+    MPI.Allreduce(val, op, arch.communicator)
+
+all_reduce(val, arch; kwargs...) = val
+
+"""
+    concatenate_local_sizes(n, arch::DistributedArch) 
+
+returns a 3-Tuple containing a vector of `size(grid, idx)` for each rank in 
+all 3 directions
```
navidcy, line 12:
```suggestion
Return a 3-Tuple containing a vector of `size(grid, idx)` for each rank in 
all 3 directions.
```

src/ImmersedBoundaries/ImmersedBoundaries.jl
```diff
@@ -105,24 +105,26 @@ struct ImmersedBoundaryGrid{FT, TX, TY, TZ, G, I, M, Arch} <: AbstractGrid{FT, T
...
     
-    function ImmersedBoundaryGrid{TX, TY, TZ}(grid::G, ib::I, wcm::M) where {TX, TY, TZ, G <: AbstractUnderlyingGrid, I, M}
+    # Internal interface
+    function ImmersedBoundaryGrid{TX, TY, TZ}(grid::G, ib::I, mi::M) where {TX, TY, TZ, G <: AbstractUnderlyingGrid, I, M}
         FT = eltype(grid)
         arch = architecture(grid)
         Arch = typeof(arch)
-
-        return new{FT, TX, TY, TZ, G, I, M, Arch}(arch, grid, ib, wcm)
+
```
navidcy, line 115:
```suggestion

```

src/ImmersedBoundaries/active_cells_map.jl
```diff
@@ -1,50 +1,124 @@
...
-    N = maximum(size(grid))
-    Type = N > MAXUInt8 ? (N > MAXUInt16 ? (N > MAXUInt32 ? UInt64 : UInt32) : UInt16) : UInt8
-    smaller_indices = getproperty.(full_indices, Ref(:I)) .|> Tuple{Type, Type, Type}
+    N = max(Nx, Ny)
+    IntType = N > MAXUInt8 ? (N > MAXUInt16 ? (N > MAXUInt32 ? UInt64 : UInt32) : UInt16) : UInt8
+    smaller_indices = getproperty.(full_indices, Ref(:I)) .|> Tuple{IntType, IntType}
     
     return smaller_indices
-end
+end
```
navidcy, line 124:
```suggestion
end

```

navidcy (commented):
just a few beautification remarks

Manifest.toml
simone-silvestri: random. I will update the manifest

src/Distributed/multi_architectures.jl
simone-silvestri: we cannot split in the vertical, that is legacy code. I could clean it all up here, or in a another PR. I am ok with both.

glwagner: isn't this code added by this PR? how is it legacy?

simone-silvestri: just changed the spacing for aesthetic purposes, this is in main
https://github.com/CliMA/Oceananigans.jl/blob/19dac0b8f6b621057e250ecf0510b3c4c91915e9/src/Distributed/multi_architectures.jl#L148-L149

Project.toml
```diff
@@ -45,7 +45,6 @@ Glob = "1.3"
 IncompleteLU = "0.2"
 IterativeSolvers = "0.9"
 JLD2 = "^0.4"
-KernelAbstractions = "0.9"
```
navidcy, line 48: we need the compat entry!

Manifest.toml
```diff
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.5"
+julia_version = "1.8.0"
```
navidcy, line 3: let's leave the manifest w 1.8.5 (which is the version the tests use)?

src/Advection/centered_reconstruction.jl
```diff
@@ -100,8 +100,8 @@ for (dir, ξ, val) in zip((:xᶠᵃᵃ, :yᵃᶠᵃ, :zᵃᵃᶠ), (:x, :y, :z),
 
     for buffer in [1, 2, 3, 4, 5, 6]
         @eval begin
-            @inline $stencil(i, j, k, grid, scheme::Centered{$buffer}, ψ, idx, loc, args...)           = @inbounds sum($(reconstruction_stencil(buffer, :symm, ξ, false)) .* retrieve_coeff(scheme, Val($val), idx, loc))
-            @inline $stencil(i, j, k, grid, scheme::Centered{$buffer}, ψ::Function, idx, loc, args...) = @inbounds sum($(reconstruction_stencil(buffer, :symm, ξ,  true)) .* retrieve_coeff(scheme, Val($val), idx, loc))
+            @inline $stencil(i, j, k, grid, scheme::Centered{$buffer, FT}, ψ, idx, loc, args...)           where FT = @inbounds sum($(reconstruction_stencil(buffer, :symm, ξ, false)) .* retrieve_coeff(scheme, Val($val), idx, loc))
+            @inline $stencil(i, j, k, grid, scheme::Centered{$buffer, FT}, ψ::Function, idx, loc, args...) where FT = @inbounds sum($(reconstruction_stencil(buffer, :symm, ξ,  true)) .* retrieve_coeff(scheme, Val($val), idx, loc))
```
navidcy, line 104: why do we need the parameter FT + where FT here since we don't use it?

src/Advection/reconstruction_coefficients.jl
```diff
@@ -154,16 +154,16 @@ julia> calc_reconstruction_stencil(3, :left, :x)
         c = n - buffer - 1
         if func
             stencil_full[idx] = dir == :x ? 
-                                :($coeff[$(order - idx + 1)] * ψ(i + $c, j, k, grid, args...)) :
+                                :(FT($coeff[$(order - idx + 1)]) * ψ(i + $c, j, k, grid, args...)) :
```
navidcy, line 157:
```suggestion
                                :(convert(FT, $coeff[$(order - idx + 1)]) * ψ(i + $c, j, k, grid, args...)) :
```
and same for the rest?

src/Advection/upwind_biased_reconstruction.jl
```diff
@@ -125,8 +125,8 @@ for (sd, side) in enumerate((:left, :right)), (dir, ξ, val) in zip((:xᶠᵃᵃ
 
     for buffer in [1, 2, 3, 4, 5, 6]
         @eval begin
-            @inline $stencil(i, j, k, grid, scheme::UpwindBiased{$buffer}, ψ, idx, loc, args...)           = @inbounds sum($(reconstruction_stencil(buffer, side, ξ, false)) .* retrieve_coeff(scheme, Val($sd), Val($val), idx, loc))
-            @inline $stencil(i, j, k, grid, scheme::UpwindBiased{$buffer}, ψ::Function, idx, loc, args...) = @inbounds sum($(reconstruction_stencil(buffer, side, ξ,  true)) .* retrieve_coeff(scheme, Val($sd), Val($val), idx, loc))
+            @inline $stencil(i, j, k, grid, scheme::UpwindBiased{$buffer, FT}, ψ, idx, loc, args...)           where FT = @inbounds sum($(reconstruction_stencil(buffer, side, ξ, false)) .* retrieve_coeff(scheme, Val($sd), Val($val), idx, loc))
+            @inline $stencil(i, j, k, grid, scheme::UpwindBiased{$buffer, FT}, ψ::Function, idx, loc, args...) where FT = @inbounds sum($(reconstruction_stencil(buffer, side, ξ,  true)) .* retrieve_coeff(scheme, Val($sd), Val($val), idx, loc))
```
navidcy, line 129: same as above: why the parameter FT if not used?

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -14,7 +14,7 @@ using CUDA
 using KernelAbstractions: @index, @kernel
 
 using Oceananigans.Architectures: CPU, GPU, device
-using Oceananigans.Utils: work_layout, launch!
+using Oceananigans.Utils: work_layout, launch!, KernelParameters
```
navidcy, line 17: let's move the
`using Oceananigans.Utils: KernelParameters` to 
[fill_halo_regions.jl](https://github.com/CliMA/Oceananigans.jl/pull/3125/files#diff-e6c8318633fd62c744bd0d83e9145622587263d81359648828794640d09d3766)

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -9,8 +9,8 @@ import Base
 ##### General halo filling functions
 #####
 
-fill_halo_regions!(::Nothing, args...) = nothing
-fill_halo_regions!(::NamedTuple{(), Tuple{}}, args...) = nothing
+fill_halo_regions!(::Nothing, args...; kwargs...) = nothing
+fill_halo_regions!(::NamedTuple{(), Tuple{}}, args...; kwargs...) = nothing
```
navidcy, line 13: why do we need `; kwargs...`? I don't wee were `fill_halo_regions!` uses kwargs. (not just here, in general).

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -163,28 +163,22 @@ fill_first(bc1, bc2)               = true
 ##### General fill_halo! kernels
 #####
 
-@kernel function _fill_west_and_east_halo!(c, west_bc, east_bc, offset, loc, grid, args) 
+@kernel function _fill_west_and_east_halo!(c, west_bc, east_bc, loc, grid, args)
```
navidcy, line 166: so I guess this is the part that is moved in the KA PR?
perhaps we wait for the to be merged before we review this?

Manifest.toml
navidcy, line 1: if this PR depends on a branch of KA then why doesn't the Manifest reflect that? seems that the tests passed just with using the tagged KA version? So do we need the KA branch? I'm bit confused.

navidcy (commented):
some comments (done w @glwagner)

src/BoundaryConditions/fill_halo_regions.jl
simone-silvestri: the `async` option in `update_hydrostatic_free_surface_model.jl`

Manifest.toml
simone-silvestri: the Manifest does reflect that

navidcy: Damn, I missed it. I thought the manifest changed ended where the first comment was. My bad, sorry.

src/BoundaryConditions/fill_halo_regions.jl
simone-silvestri: yeah we need that PR, we do not want to depend on a branch

src/BoundaryConditions/fill_halo_regions.jl
glwagner: should we pass `async`, or are there other kwargs? (it's easier to understand the code if we are more explicit)

src/Advection/upwind_biased_reconstruction.jl
simone-silvestri: right, here it is not really used

src/Advection/centered_reconstruction.jl
simone-silvestri: right here it's not really used

src/BoundaryConditions/fill_halo_regions.jl
simone-silvestri: there is also `reduced_dimensions`.

navidcy:
Is the PR on KA very crucial?

Would it be a lot of refactoring to have this merged atm with a tagged KA version? This way we don't need to wait for KA?

simone-silvestri:
closed it by mistake 😅

simone-silvestri:
> Is the PR on KA very crucial?
> 
> Would it be a lot of refactoring to have this merged atm with a tagged KA version? This way we don't need to wait for KA?

Sorry, I misunderstood. Yeah, it would be indeed a bit of refactoring.

src/Advection/upwind_biased_advective_fluxes.jl
```diff
@@ -9,7 +9,10 @@ const UpwindScheme = AbstractUpwindBiasedAdvectionScheme
 
 @inline upwind_biased_product(ũ, ψᴸ, ψᴿ) = ((ũ + abs(ũ)) * ψᴸ + (ũ - abs(ũ)) * ψᴿ) / 2
 
-@inline sign_val(u) = Val(Int(sign(u)))
+struct LeftUpwind end
+struct RightUpwind end
+
+@inline sign_val(u) = ifelse(u > 0, LeftUpwind(), RightUpwind())
```
glwagner, line 15:
```suggestion
@inline upwind_direction(u) = ifelse(u > 0, LeftUpwind(), RightUpwind())
```

src/MultiRegion/multi_region_models.jl
```diff
@@ -59,16 +59,18 @@ implicit_diffusion_solver(time_discretization::VerticallyImplicitTimeDiscretizat
...
                                           _getregion(t.vertical_scheme, r),
+                                          _getregion(t.ke_gradient_scheme, r),
                                           _getregion(t.upwinding, r))
 
-@inline _getregion(t::VectorInvariant{N, FT, Z, ZS, V, D, M}, r) where {N, FT, Z, ZS, V, D, M} = 
+@inline _getregion(t::VectorInvariant{N, FT, Z, ZS, V, K, D, M}, r) where {N, FT, Z, ZS, V, K, D, M} = 
                 VectorInvariant{N, FT, M}(getregion(t.vorticity_scheme, r), 
                                           getregion(t.vorticity_stencil, r), 
                                           getregion(t.vertical_scheme, r),
+                                          getregion(t.ke_gradient_scheme, r),
```
glwagner, line 73:
```suggestion
                                          getregion(t.kinetic_energy_gradient_scheme, r),
```

src/OutputWriters/checkpointer.jl
```diff
@@ -217,7 +217,7 @@ function set!(model, filepath::AbstractString)
             if string(name) ∈ keys(file) # Test if variable exist in checkpoint.
                 model_field = model_fields[name]
                 halo = halo_size(model_field.grid)
-                loc  = location(model_field)
+                loc  = map(instantiate, location(model_field))
```
glwagner, line 220: is this what `JLD2OutputWriter` does?

glwagner: this a major change... might be the right one but probably should writing something in the comments about why

simone-silvestri: Yeah I was just testing something, I have problem restarting with different halos. I have removed that behavior and hard coded the different halo loading in my repo manually, but I ll open an issue so we can discuss it, maybe it's something we want to support

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -306,7 +306,7 @@ const FNS = FixedSubstepNumber
 const FTS = FixedTimeStepSize
 
 @inline calculate_substeps(substepping::FNS, Δt) = length(substepping.averaging_weights)
-@inline calculate_substeps(substepping::FTS, Δt) = ceil(Int, 2 * Δt / substepping.Δt_barotropic)
+@inline calculate_substeps(substepping::FTS, Δt) = max(5, ceil(Int, 2 * Δt / substepping.Δt_barotropic))
```
glwagner, line 309: something like

```suggestion
const MINIMUM_SUBSTEPS = 5
@inline calculate_substeps(substepping::FTS, Δt) = max(MINIMUM_SUBSTEPS, ceil(Int, 2 * Δt / substepping.Δt_barotropic))
```

might be clearer. With a comment around `MINIMUM_SUBSTEPS`! even better

src/Distributed/partition_assemble.jl
```diff
@@ -110,10 +110,11 @@ end
 Partition a global array in local arrays of size `(nx, ny)` if 2D or `(nx, ny, nz)` is 3D.
 Usefull for boundary arrays, forcings and initial conditions.
 """
-partition_global_array(arch, c_global::Function, n) = c_global 
+partition_global_array(arch, c_global::AbstractArray, n) = c_global
+partition_global_array(arch, c_global::Function, n)      = c_global 
 
 # Here we assume that we cannot partition in z (we should remove support for that)
-function partition_global_array(arch, c_global::AbstractArray, n) 
+function partition_global_array(arch::DistributedArch, c_global::AbstractArray, n)
```
glwagner, line 117: should we call this `DistributedArchitecture`? Wouldn't want to confuse it with

image

which can cause trouble with Italians.

simone-silvestri: what about `MultiProcessor`?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.87.2"
+version = "0.88.0"
```
glwagner, line 4: What are the breaking changes? This PR seems to concern only MPI which doesn't really have user facing stuff?

src/Advection/vector_invariant_advection.jl
```diff
@@ -7,21 +7,24 @@ struct EnstrophyConservingScheme{FT} <: AbstractAdvectionScheme{1, FT} end
...
-    vorticity_stencil :: ZS # stencil used for assessing vorticity smoothness
-    vertical_scheme   :: V  # stencil used for assessing divergence smoothness
-    upwinding         :: D  # treatment of upwinding for divergence flux and kinetic energy gradient
+struct VectorInvariant{N, FT, Z, ZS, V, K, D, U, M} <: AbstractAdvectionScheme{N, FT}
+    vorticity_scheme   :: Z  # reconstruction scheme for vorticity flux
+    vorticity_stencil  :: ZS # stencil used for assessing vorticity smoothness
+    vertical_scheme    :: V  # recontruction scheme for vertical advection
+    ke_gradient_scheme :: K  # reconstruction scheme for kinetic energy gradient
+    divergence_scheme  :: D  # reconstruction scheme for divergence flux
+    upwinding          :: U  # treatment of upwinding for divergence flux and kinetic energy gradient
```
glwagner, line 16: There are changes to WENO vector invariant in this PR?

src/Advection/vector_invariant_advection.jl
```diff
@@ -77,18 +82,26 @@ Vector Invariant, Dimension-by-dimension reconstruction
...
-const VectorInvariantEnstrophyConserving      = VectorInvariant{<:Any, <:Any, <:EnstrophyConservingScheme}
-const VectorInvariantVerticalEnergyConserving = VectorInvariant{<:Any, <:Any, <:Any, <:Any, <:EnergyConservingScheme}
+const VectorInvariantEnergyConserving           = VectorInvariant{<:Any, <:Any, <:EnergyConservingScheme}
+const VectorInvariantEnstrophyConserving        = VectorInvariant{<:Any, <:Any, <:EnstrophyConservingScheme}
+const VectorInvariantVerticalEnergyConserving   = VectorInvariant{<:Any, <:Any, <:Any, <:Any, <:EnergyConservingScheme}
+const VectorInvariantKEGradientEnergyConserving = VectorInvariant{<:Any, <:Any, <:Any, <:Any, <:Any, <:EnergyConservingScheme}
 
 const VectorInvariantUpwindVorticity  = VectorInvariant{<:Any, <:Any, <:AbstractUpwindBiasedAdvectionScheme}
-const MultiDimensionalVectorInvariant = VectorInvariant{<:Any, <:Any, <:Any, <:Any, <:Any, <:Any, true}
+const MultiDimensionalVectorInvariant = VectorInvariant{<:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:Any, true}
```
glwagner, line 104: typically we recommend putting the dispatchable type parameters at the beginning of the type parameter list. Any reason why you are putting them at the very end? It will make this hard to debug in the future if they are at the end. It's also hard to read; eg we have to count up to 9 to figure out what "true" means in the above.

src/Advection/vector_invariant_advection.jl
```diff
@@ -281,21 +296,34 @@ const UX{N} = UpwindBiased{N, <:Any, <:Nothing}
...
+        @inline inner_right_biased_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme::UX{$buffer}, f::Function, idx, loc, ::AbstractSmoothnessStencil, args...) = inner_right_biased_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, f, idx, loc, args...)
+        @inline inner_right_biased_interpolate_yᵃᶠᵃ(i, j, k, grid, scheme::U{$buffer},  f::Function, idx, loc, ::AbstractSmoothnessStencil, args...) = inner_right_biased_interpolate_yᵃᶠᵃ(i, j, k, grid, scheme, f, idx, loc, args...)
+        @inline inner_right_biased_interpolate_yᵃᶠᵃ(i, j, k, grid, scheme::UY{$buffer}, f::Function, idx, loc, ::AbstractSmoothnessStencil, args...) = inner_right_biased_interpolate_yᵃᶠᵃ(i, j, k, grid, scheme, f, idx, loc, args...)
+        @inline inner_right_biased_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme::U{$buffer},  f::Function, idx, loc, ::AbstractSmoothnessStencil, args...) = inner_right_biased_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme, f, idx, loc, args...)
+        @inline inner_right_biased_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme::UZ{$buffer}, f::Function, idx, loc, ::AbstractSmoothnessStencil, args...) = inner_right_biased_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme, f, idx, loc, args...)
```
glwagner, line 327: These lines are long so it could help to shorten. For example `b` for `buffer and `ASS` for `AbstractSmoothnessStencil`.

src/Advection/weno_interpolants.jl
```diff
@@ -324,16 +324,16 @@ julia> calc_weno_stencil(2, :right, :x)
...
+            else    
                 stencil_point[idx] =  dir == :x ? 
-                                    :(@inbounds ψ[i + $c, j, k]) :
-                                    dir == :y ?
-                                    :(@inbounds ψ[i, j + $c, k]) :
-                                    :(@inbounds ψ[i, j, k + $c])
+                                      :(@inbounds ψ[i + $c, j, k]) :
+                                      dir == :y ?
+                                      :(@inbounds ψ[i, j + $c, k]) :
+                                      :(@inbounds ψ[i, j, k + $c])
```
glwagner, line 336: I've no idea what this code does so a comment could be warranted...

src/Architectures.jl
```diff
@@ -86,7 +93,7 @@ function unified_array(::GPU, arr::AbstractArray)
     return vec
 end
 
-## Only for contiguous data!! (i.e. only if the offset for pointer(dst::CuArrat, offset::Int) is 1)
+## Only for contiguous data!! (i.e. only if the offset for pointer(dst::CuArray, offset::Int) is 1)
```
glwagner, line 96: What does this comment mean? Is this a warning that the function can be used incorrectly or something? It might be better to enforce more strictly if possible.

src/Distributed/Distributed.jl
```diff
@@ -19,5 +18,6 @@ include("halo_communication_bcs.jl")
 include("distributed_fields.jl")
 include("halo_communication.jl")
 include("distributed_fft_based_poisson_solver.jl")
+include("interleave_comm_and_comp.jl")
```
glwagner, line 21:
```suggestion
include("interleave_communication_and_computation.jl")
```

Please don't use abbreviations.

src/Distributed/halo_communication.jl
```diff
@@ -28,46 +23,56 @@ import Oceananigans.BoundaryConditions:
...
-    :south => :north, :north => :south,
-    :bottom => :top, :top => :bottom
+    :west => :east, 
+    :east => :west,
+    :south => :north,
+    :north => :south,
+    :southwest => :northeast, 
+    :southeast => :northwest, 
+    :northwest => :southeast, 
+    :northeast => :southwest,
```
glwagner, line 41: Why do corners communicate?

src/Distributed/multi_architectures.jl
```diff
@@ -103,24 +96,26 @@ function DistributedArch(child_architecture = CPU();
...
-end
+    M = typeof(mpi_requests)
+    T = typeof(Ref(0))
 
-const ViewsDistributedArch = DistributedArch{<:Any, <:Any, <:Any, <:Any, <:Any, <:Any, false}
+    return DistributedArch{A, R, I, ρ, C, γ, M, T}(child_architecture, local_rank, local_index, ranks, local_connectivity, communicator, mpi_requests, Ref(0))
+end
 
-using_buffered_communication(::DistributedArch{A, R, I, ρ, C, γ, B}) where {A, R, I, ρ, C, γ, B} = B
+const BlockingDistributedArch = DistributedArch{<:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:Nothing}
```
glwagner, line 107:
```suggestion
const BlockingDistributedArch = DistributedArch{<:Nothing}
```

Why is the dispatched type parameter so far down the list? It's easier to read the code if the dispatched parameter is moved to the front as in the above.

src/Distributed/multi_architectures.jl
```diff
@@ -140,17 +135,19 @@ end
...
 #####
 
-struct RankConnectivity{E, W, N, S, T, B}
-      east :: E
-      west :: W
-     north :: N
-     south :: S
-       top :: T
-    bottom :: B
+struct RankConnectivity{E, W, N, S, SW, SE, NW, NE}
```
glwagner, line 138: docstring

src/Distributed/halo_communication.jl
simone-silvestri: in case of 2D partitioning, to avoid double pass.

src/Distributed/multi_architectures.jl
```diff
@@ -181,27 +178,32 @@ function decrement_index(i, R, topo)
...
+    r_east  = isnothing(i_east)  ? nothing : index2rank(i_east,  j, k, Rx, Ry, Rz)
+    r_west  = isnothing(i_west)  ? nothing : index2rank(i_west,  j, k, Rx, Ry, Rz)
     r_north = isnothing(j_north) ? nothing : index2rank(i, j_north, k, Rx, Ry, Rz)
     r_south = isnothing(j_south) ? nothing : index2rank(i, j_south, k, Rx, Ry, Rz)
-    r_top   = isnothing(k_top)   ? nothing : index2rank(i, j, k_top, Rx, Ry, Rz)
-    r_bot   = isnothing(k_bot)   ? nothing : index2rank(i, j, k_bot, Rx, Ry, Rz)
 
-    return RankConnectivity(east=r_east, west=r_west, north=r_north,
-                            south=r_south, top=r_top, bottom=r_bot)
+    r_northeast = isnothing(i_east) || isnothing(j_north) ? nothing : index2rank(i_east, j_north, k, Rx, Ry, Rz)
```
glwagner, line 196: should `r_northeast` be something like `northeastern_rank`? Not sure what `r` is.

src/Distributed/multi_architectures.jl
```diff
@@ -181,27 +178,32 @@ function decrement_index(i, R, topo)
     end
 end
 
-function RankConnectivity(model_index, ranks, topology)
-    i, j, k = model_index
+function RankConnectivity(local_index, ranks, topology)
```
glwagner, line 181:
```suggestion
function RankConnectivity(local_index, Nranks, topology)
```

src/Distributed/partition_assemble.jl
glwagner: A minor improvement on that would be `MultiProcess`. (When you say "multi-processor", I am thinking of the device itself, rather than an abstraction that represents the whole computation)

src/ImmersedBoundaries/ImmersedBoundaries.jl
```diff
@@ -105,24 +105,33 @@ struct ImmersedBoundaryGrid{FT, TX, TY, TZ, G, I, M, Arch} <: AbstractGrid{FT, T
     architecture :: Arch
     underlying_grid :: G
     immersed_boundary :: I
-    active_cells_map :: M
+    active_cells_interior :: M
```
glwagner, line 108:
```suggestion
    interior_active_cells :: M
```

Though I have to admit, I don't know what this object means. I just know that I would never use the phrase "active cells interior" in a sentence...

Project.toml
simone-silvestri: I changes a lot of the internals of the code so I think it is good to bump the version. There are some API changes such as:
- `active_cells_map` kwarg for `ImmersedBoundaryGrid`
- `boundary_buffer` for scalar diffusivities
and one breaking change
- size passed to distributed grids is the _local_ not the _global_ size

src/Models/HydrostaticFreeSurfaceModels/calculate_hydrostatic_free_surface_boundary_tendencies.jl
```diff
@@ -0,0 +1,57 @@
...
+    
+    w_kernel_parameters = boundary_w_kernel_parameters(grid, arch)
+    p_kernel_parameters = boundary_p_kernel_parameters(grid, arch)
+    κ_kernel_parameters = boundary_κ_kernel_parameters(grid, model.closure, arch)
+
+    for (wpar, ppar, κpar) in zip(w_kernel_parameters, p_kernel_parameters, κ_kernel_parameters)
+        compute_w_from_continuity!(model.velocities, arch, grid; parameters = wpar)
+        update_hydrostatic_pressure!(model.pressure.pHY′, arch, grid, model.buoyancy, model.tracers; parameters = ppar)
+        calculate_diffusivities!(model.diffusivity_fields, model.closure, model; parameters = κpar)
+    end
```
glwagner, line 36: Should all of this be in `update_state!`? Somehow all of this functionality is quite specific to the current form of update state; ie if we want to change how we conceive of the auxiliary state of the hydrostati cmodel, we need to change the stuff here _and_ the stuff in update state.

If we are going to create these algorithm dependencies we should write down / document what the algorithm is I think.

src/Models/HydrostaticFreeSurfaceModels/calculate_hydrostatic_free_surface_tendencies.jl
```diff
@@ -1,24 +1,35 @@
...
-using Oceananigans.ImmersedBoundaries: use_only_active_cells, ActiveCellsIBG, active_linear_index_to_ntuple
+import Oceananigans.Distributed: complete_communication_and_compute_boundary!
+import Oceananigans.Distributed: interior_tendency_kernel_parameters
+
+using Oceananigans.ImmersedBoundaries: use_only_active_interior_cells, ActiveCellsIBG, 
+                                       InteriorMap, active_linear_index_to_interior_tuple
 
 """
-    calculate_tendencies!(model::HydrostaticFreeSurfaceModel, callbacks)
+    compute_tendencies!(model::HydrostaticFreeSurfaceModel, callbacks)
```
glwagner, line 20: <3

src/Models/HydrostaticFreeSurfaceModels/calculate_hydrostatic_free_surface_tendencies.jl
```diff
@@ -76,21 +87,20 @@ top_tracer_boundary_conditions(grid, tracers) =
     NamedTuple(c => tracers[c].boundary_conditions.top for c in propertynames(tracers))
 
 """ Store previous value of the source term and calculate current source term. """
-function calculate_hydrostatic_free_surface_interior_tendency_contributions!(model)
+function calculate_hydrostatic_free_surface_tendency_contributions!(model, kernel_parameters; only_active_cells = nothing)
```
glwagner, line 90: calculate -> compute while you're at it...

src/Models/HydrostaticFreeSurfaceModels/single_column_model_mode.jl
```diff
@@ -42,12 +42,20 @@ validate_momentum_advection(momentum_advection, ::SingleColumnGrid) = nothing
...
 ##### Time-step optimizations
 #####
 
 calculate_free_surface_tendency!(::SingleColumnGrid, args...) = nothing
 
+# Disambiguation
+calculate_free_surface_tendency!(::SingleColumnGrid, ::ImplicitFreeSurfaceHFSM     , args...) = nothing
+calculate_free_surface_tendency!(::SingleColumnGrid, ::SplitExplicitFreeSurfaceHFSM, args...) = nothing
+
+
```
glwagner, line 58:
```suggestion
```

src/Models/HydrostaticFreeSurfaceModels/calculate_hydrostatic_free_surface_boundary_tendencies.jl
simone-silvestri: you mean maybe writing an `update_auxiliaries!(model::HydrostaticFreeSurfaceModel; kwargs...)` to call both in `update_state!` and here?

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
```diff
@@ -22,27 +22,26 @@ Update peripheral aspects of the model (auxiliary fields, halo regions, diffusiv
...
+    fill_halo_regions!(prognostic_fields(model), model.clock, fields(model); async = true)
 
     @apply_regionally replace_horizontal_velocity_halos!(model.velocities, model.grid)
     @apply_regionally compute_w_diffusivities_pressure!(model)
+    fill_halo_regions!(model.diffusivity_fields; only_local_halos = true)
 
-    fill_halo_regions!(model.velocities.w, model.clock, fields(model))
-    fill_halo_regions!(model.diffusivity_fields, model.clock, fields(model))
-    fill_halo_regions!(model.pressure.pHY′)
+    [callback(model) for callback in callbacks if isa(callback.callsite, UpdateStateCallsite)]
```
glwagner, line 38:
```suggestion
    [callback(model) for callback in callbacks if callback.callsite isa UpdateStateCallsite]
```

why doesn't that work

src/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -61,13 +62,21 @@ const VerticallyBoundedGrid{FT} = AbstractGrid{FT, <:Any, <:Any, <:Bounded}
...
 """
-abstract type AbstractTurbulenceClosure{TimeDiscretization} end
+abstract type AbstractTurbulenceClosure{TimeDiscretization, BoundaryBuffer} end
 
 # Fallbacks
 validate_closure(closure) = closure
 closure_summary(closure) = summary(closure)
 with_tracers(tracers, closure::AbstractTurbulenceClosure) = closure
-calculate_diffusivities!(K, closure::AbstractTurbulenceClosure, args...) = nothing
+calculate_diffusivities!(K, closure::AbstractTurbulenceClosure, args...; kwargs...) = nothing
```
glwagner, line 71:
```suggestion
compute_diffusivities!(K, closure::AbstractTurbulenceClosure, args...; kwargs...) = nothing
```

src/TurbulenceClosures/turbulence_closure_implementations/nothing_closure.jl
glwagner, line 1: we should delete this

src/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -61,13 +62,21 @@ const VerticallyBoundedGrid{FT} = AbstractGrid{FT, <:Any, <:Any, <:Bounded}
 
 Abstract supertype for turbulence closures.
 """
-abstract type AbstractTurbulenceClosure{TimeDiscretization} end
+abstract type AbstractTurbulenceClosure{TimeDiscretization, BoundaryBuffer} end
```
glwagner, line 65: What does "boundary buffer" mean

src/TurbulenceClosures/turbulence_closure_implementations/isopycnal_skew_symmetric_diffusivity.jl
```diff
@@ -1,4 +1,4 @@
-struct IsopycnalSkewSymmetricDiffusivity{TD, K, S, M, L} <: AbstractTurbulenceClosure{TD}
+struct IsopycnalSkewSymmetricDiffusivity{TD, K, S, M, L} <: AbstractTurbulenceClosure{TD, 1}
```
glwagner, line 1: Is `1` always right? This whole thing was set up so that users can implement arbitrary functions that calculate k_skew. Seems like users need to be able to set this based on their model for k_skew etc.

glwagner (approved):
This looks pretty good considering the insane complexity. I think we need to leave a summary of the flow of information in the hydrostatic model though because this PR adds a lot of inertia to make any changes to the hydrsotatic auxiliary state for example. Basically, it seems like we are committing pretty seriously to a particular pattern - compute w, compute hydrostatic pressure, compute diffusivities. If we have other auxiliary task to do, we can't just write them in `update_state!`; we also have to take care that the task is added in another file for the parallel algorithm. Something to keep in mind. Any way to make it easier to continue to develop the code --- because we are not done with our work --- will really help progress in the future.

glwagner:
> This PR allows hiding MPI passing of barotropic auxiliary variables behind the implicit vertical solver and prognostic variables behind the tendency calculations. The latter is done by splitting the tendency kernels into an interior kernel that calculates tendencies between, e.g., i = Hx and i = Nx - Hx, and a boundary kernel, executed once communication is complete, that calculates tendencies adjacent to boundaries.

@simone-silvestri I think this description is incomplete. Isn't there also a change to how the auxiliary computations are done?

simone-silvestri:
> This looks pretty good considering the insane complexity. I think we need to leave a summary of the flow of information in the hydrostatic model though because this PR adds a lot of inertia to make any changes to the hydrsotatic auxiliary state for example. Basically, it seems like we are committing pretty seriously to a particular pattern - compute w, compute hydrostatic pressure, compute diffusivities. If we have other auxiliary task to do, we can't just write them in `update_state!`; we also have to take care that the task is added in another file for the parallel algorithm. Something to keep in mind. Any way to make it easier to continue to develop the code --- because we are not done with our work --- will really help progress in the future.

To alleviate this problem we can maybe change the `recompute_auxiliaries` in line with how this PR treats `calculate_hydrostatic_free_surface_tendency_contributions!` i.e., the function is the same for interior and boundary computations, only the kernel parameters passed to the function change. In this way, we could define a `compute_auxiliaries(model; parameters)` which is then called again in the `complete_communication_and_compute_boundary!` function. What do you think?

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
simone-silvestri: It should work

src/Distributed/distributed_fields.jl
```diff
@@ -16,8 +14,22 @@ function Field((LX, LY, LZ)::Tuple, grid::DistributedGrid, data, old_bcs, indice
...
+        f_field = field(location(u), f, cpu_grid)
+        set!(u_cpu, f_field)
+        set!(u, u_cpu)
+    elseif child_architecture(u) isa CPU
+        f_field = field(location(u), f, u.grid)
+        set!(u, f_field)
+    end
+
+    return u
+end
```
glwagner, line 34: is this the same as the `set!` for non-distributed fields?

simone-silvestri: this is the same `set!` but it looks at `child_architecture` instead of `architecture`

src/TurbulenceClosures/TurbulenceClosures.jl
simone-silvestri: I changed it to `RequiredHalo`

src/TurbulenceClosures/turbulence_closure_implementations/isopycnal_skew_symmetric_diffusivity.jl
simone-silvestri: right, I'll change this

src/Advection/vector_invariant_advection.jl
simone-silvestri: It does not really change the default behavior, I introduced some flexibility to allow different schemes for the divergence flux and the kinetic energy gradient

src/ImmersedBoundaries/active_cells_map.jl
```diff
@@ -1,50 +1,121 @@
...
+
+function compute_active_cells_interior(ibg)
+    is_immersed_operation = KernelFunctionOperation{Center, Center, Center}(active_cell, ibg)
+    active_cells_field = Field{Center, Center, Center}(ibg, Bool)
     set!(active_cells_field, is_immersed_operation)
     return active_cells_field
 end
 
+function compute_active_cells_surface(ibg)
+    one_field = ConditionalOperation{Center, Center, Center}(OneField(Int), identity, ibg, NotImmersed(truefunc), 0.0)
```
navidcy, line 56:
```suggestion
    one_field = ConditionalOperation{Center, Center, Center}(OneField(Int), identity, ibg, NotImmersed(truefunc), 0)
```

to avoid type instability?

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -288,44 +284,44 @@ function split_explicit_free_surface_step!(free_surface::SplitExplicitFreeSurfac
...
     return nothing
 end
 
 # Change name
 const FNS = FixedSubstepNumber
 const FTS = FixedTimeStepSize
 
+# since weights can be negative in the first few substeps (as in the default averaging kernel), 
+# we set a minimum number of substeps to execute to avoid numerical issues
+const MINIMUM_SUBSTEPS = 5
```
navidcy, line 317: is the convention all caps for consts?

src/Solvers/batched_tridiagonal_solver.jl
```diff
@@ -214,4 +214,4 @@ end
             ϕ[i, j, k] -= t[i, j, k+1] * ϕ[i, j, k+1]
         end
     end
-end
+end
```
navidcy, line 217:
```suggestion
end

```

an empty line is nice

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
simone-silvestri: That's what we usually use (like in like 312 and 313)

navidcy: cool

src/Architectures.jl
simone-silvestri: I have changed the comment

src/Distributed/multi_architectures.jl
simone-silvestri: Move it to the front

navidcy:
oh dear, my PR created merge conflicts......

navidcy:
@simone-silvestri I think I resolve the conflicts alright but it wouldn't hurt if you have a look at [b51e681](https://github.com/CliMA/Oceananigans.jl/pull/3125/commits/b51e68180d062ea563f417481d811a3376a93ad5)...

glwagner:
do we use `arch = Distributed()` for distributed grids now? Might be good to summarize in the top description

tomchor:
> Requires [JuliaGPU/KernelAbstractions.jl#399](https://github.com/JuliaGPU/KernelAbstractions.jl/pull/399) # On hold at the moment

At the top comment it says this PR depends on another KA PR that isn't merged. However, this PR is merged (and tagged). Does this not depend on https://github.com/JuliaGPU/KernelAbstractions.jl/pull/399 anymore?

navidcy:
> > Requires [JuliaGPU/KernelAbstractions.jl#399](https://github.com/JuliaGPU/KernelAbstractions.jl/pull/399) # On hold at the moment
> 
> At the top comment it says this PR depends on another KA PR that isn't merged. However, this PR is merged (and tagged). Does this not depend on [JuliaGPU/KernelAbstractions.jl#399](https://github.com/JuliaGPU/KernelAbstractions.jl/pull/399) anymore?

Yeap, it doesn't. I edited the first post ;)

PetrKryslUCSD:
How do you guys handle MPI tracing? (A la MPE, for instance...)

glwagner:
@PetrKryslUCSD I think we use `nvtx`, but @simone-silvestri can say more.

simone-silvestri:
Using nsys it is possible to trace MPI with `--trace=mpi`, see:
https://docs.nvidia.com/nsight-systems/UserGuide/index.html

PetrKryslUCSD:
Thank you for your response. The question is: what do *you* use to trace MPI
in Julia programs? Do you use `nsys`? If so, how?

Petr Krysl
Prof. and Vice chair for undergraduate education
Department of Structural Engineering
<https://urldefense.com/v3/__https://www.linkedin.com/company/uc-san-diego-structural-engineering-department/__;!!Mih3wA!DXYUp152SRRo03xCfQJ9NlAXovNVk-zBYRalwekzmzf2bGwdDMmY8gy9t6iQo9ok2_dvk8m987An93w0pnT_6lM$>
University of California, San Diego
9500 Gilman Drive #0085
La Jolla, CA 92093

simone-silvestri:
Yep, we used nsys because our primary objective is to trace GPU execution. I think it will work also on CPU programs. 
There is nothing really specific about profiling julia with nsys, provided that MPI is correctly configured (i.e. your script works with MPI already). An example of a batch script that traces MPI calls is
```
#!/bin/bash
#SBATCH -N 2
#SBATCH --ntasks-per-node=4
#SBATCH --cpus-per-task=16
#SBATCH --mem=500GB
#SBATCH --time 24:00:00
#SBATCH --gres=gpus:4

cat > launch.sh << EoF_s
#! /bin/sh
export CUDA_VISIBLE_DEVICES=0,1,2,3
exec \$*
EoF_s
chmod +x launch.sh

srun nsys profile --trace=nvtx,cuda,mpi --output=report_%q{SLURM_PROCID} ./launch.sh julia --check-bounds=no --project scaling_experiments.jl 
```
Here, `nsys` will produce one report per processor.  You can use `mpirun` or `mpiexec` instead of `srun`.
If you want to insert NVTX annotations inside the code you need to add the environment variable ([ref](https://github.com/JuliaGPU/NVTX.jl)) 
```
export JULIA_NVTX_CALLBACKS=gc
```

PetrKryslUCSD:
Very cool. Many thanks!

Petr Krysl
Prof. and Vice chair for undergraduate education
Department of Structural Engineering
<https://urldefense.com/v3/__https://www.linkedin.com/company/uc-san-diego-structural-engineering-department/__;!!Mih3wA!DXYUp152SRRo03xCfQJ9NlAXovNVk-zBYRalwekzmzf2bGwdDMmY8gy9t6iQo9ok2_dvk8m987An93w0pnT_6lM$>
University of California, San Diego
9500 Gilman Drive #0085
La Jolla, CA 92093

PR #3127, opened 2023-06-01
Three dimensional difference operators

simone-silvestri:
instead of 3D conditional derivatives, this PR introduces 3D conditional differences (and as a result derivatives)

PR #3130, opened 2023-06-02
3D interpolation operators

simone-silvestri:
as #3127 but for interpolation operators.

In this case, these 3D interpolation operators assume no flux across the immersed boundary

discussed in #3158


glwagner:
I don't know what it means to "assume no flux". 

The behavior we likely want is to always return the "active" value. Considering two values `a` and `b` in adjacent cells, the rule we probably want is

1. If both cells involved in reconstruction are "active", return a centered second order reconstuction `(a + b) / 2`.
2. If `b` is in an inactive cell, return `a`.
3. If `a` is in an inactive cell, return `b`.
4. If both are inactive, return `0`.

simone-silvestri:
should I merge?

glwagner:
Shouldn't we have tests?

glwagner:
It's still a draft, needs to be unmarked.

simone-silvestri:
Oh sorry my comment was intended for #3127 not this PR

glwagner:

simone-silvestri:
I think this now changes the difference operators. I would remove the `\delta` operators from this PR and keep the interpolation which I would strip of the `B` superscript to match what the differences do.

PR #3131, opened 2023-06-03
Fix few bugs + improve docstring of `SplitExplicitFreeSurface` 

navidcy:
bugs:

- `Δy = minimum_xspacing(grid)`
- some leftover `Δt_max` was changed to `max_Δt`
- when computing `substeps` via cfl then we need to ensure it's an integer

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
@@ -36,20 +36,6 @@ end
 Return a `SplitExplicitFreeSurface` representing an explicit time discretization
 of oceanic free surface dynamics with `gravitational_acceleration`.
 
-The `SplitExplicitSettings` function call is:
-
-```julia
-    SplitExplicitSettings(FT::DataType=Float64;
-                          substeps = nothing,
-                          cfl    = nothing,
-                          grid   = nothing,
-                          max_Δt = nothing,
-                          gravitational_acceleration = g_Earth,
-                          barotropic_averaging_kernel = averaging_shape_function,
-                          timestepper = ForwardBackwardScheme())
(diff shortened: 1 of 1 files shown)


glwagner:
Might be better to contribute this to the broader user API overhaul in #3124 (won't this cause conflicts?)

navidcy:
#3124 was merged!

navidcy:
@simone-silvestri, what exactly is `max_Δt`? It's the maximum _baroclinic_ tilmestep, right?

glwagner:
> @simone-silvestri, what exactly is `max_Δt`? It's the maximum _baroclinic_ tilmestep, right?

That's right. Maybe we could have a better name for that.

navidcy:
Well I at least added a remark.

navidcy:
```Julia
if (!isnothing(substeps) && !isnothing(cfl)) || (isnothing(substeps) && isnothing(cfl))
``

that's twice the same condition, right?

navidcy:
Perhaps at
https://github.com/CliMA/Oceananigans.jl/blob/f273c54fa79e43e1fc3d473537effee2095b7550/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl#L280
we should use `Int(floor(...))` to be tad more conservative.

simone-silvestri:
> Perhaps at
> 
> https://github.com/CliMA/Oceananigans.jl/blob/f273c54fa79e43e1fc3d473537effee2095b7550/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl#L280
> 
> 
> we should use `Int(floor(...))` to be tad more conservative.

probably `ceil(Int, ...)`

simone-silvestri (approved):

simone-silvestri (approved):

PR #3094, opened 2023-05-01
PCG uses provided `grid`'s halo size instead of hardcoding `halo = (3, 3, 1)`

navidcy:
Resolve #3093

Issue #3076, opened 2023-04-20
Proposal for sustainable `validation` scripts


glwagner:
The `validation` scripts have proved useful for code development since they essentially provide convenient scientific / integration tests which would be hard or expensive to implement in CI, but which are still useful when implementing or evaluating a new feature. They also help us share knowledge and code.

Yet a major problem is that the `validation` scripts are not tested and therefore most of them are broken, since they aren't updated as Oceananigans syntax changes.

I was talking with @siddharthabishnu and we realized that a possible solution would be to follow the convention used by Flux's "model-zoo": https://github.com/FluxML/model-zoo

With this pattern, every validation "experiment" is a directory that includes a collection of scripts and a `Project.toml`. The `Project.toml` indicates the version of Oceananigans. If people want to upgrade the scripts + environment they can submit a PR.

We'd still be informally testing the validation scripts, but hopefully this would make them more useful in the future, especially to new users.

We'll have to select a handful of cases that we want to keep around in the process of transitioning to the new system.

navidcy:
Yes!

navidcy:
Adding a Manifest + Project toml files in each validation scripts directory is a good idea!
However, if we do it while we are working on them in a PR, then the Manifest comes with an Oceananigans dependency that points to the branch, e.g.,

navidcy:
https://github.com/CliMA/Oceananigans.jl/blob/1c2a6f8752b6425bf30d856f8ba0aa681c0ab818/validation/stokes_drift/Manifest.toml#L1202

and the branch gets deleted after the PR is merged. 

So what one would need to do is to make another PR with a Manifest pointing to the specific commit on `main` or to the next tagged release? Sounds like too much work to be sustainable...?

glwagner:
Hmm, maybe just `compat` entries instead of `Manifest`?

navidcy:
Doesn't ensure 100% reproducibility but definitely easier to do!

glwagner:
Do we want to have a goal of 100% reproducibility for the validation cases? I think their main purpose is expository, ie to show advanced usage of the code. Exact reproducibility isn't the highest priority for that purpose --- is it?

navidcy:
True true. I think a Project.toml with compat entries is good enough!

PR #3132, opened 2023-06-03
Add internal tide example in Docs

navidcy:
This is a simple example with immersed boundary.

https://github.com/CliMA/Oceananigans.jl/assets/7112768/b225a647-7f21-4dcc-8b12-1ae9d662d106


navidcy:
This PR requires #3094. At the moment it includes the changes there but ideally we want to conclude the discussion in #3094 and merge that one before this.

Or perhaps use the split-explicit free surface?

navidcy:
cc @djlikesdjs

tomchor:
@navidcy can you please include the link to the docs preview page so that we see how it looks before approving?

navidcy:
Could you build the docs locally?

Comment out all examples here
https://github.com/CliMA/Oceananigans.jl/blob/72a2399394d60fa22d77dc51822d6175ffcf07e3/docs/make.jl#L30-L43
_except_ the tide one and run:

```
$ rm -rf docs/build; rm -rf docs/src/generated; julia --project -e 'using Pkg; Pkg.instantiate()'; julia --project=docs/ -e 'using Pkg; Pkg.precompile()'; JULIA_DEBUG=Documenter julia --project=docs/ docs/make.jl
```
and then open, e.g., `docs/build/generated/barotropic_tide.html`.

If you even change
https://github.com/CliMA/Oceananigans.jl/blob/72a2399394d60fa22d77dc51822d6175ffcf07e3/docs/make.jl#L141-L151
to `doctest = false`, `strict = false` and `checkdocs = :none`, it'll speed things even more. 

(It took ~8 mins on my laptop.)

tomchor:
> Could you build the docs locally?

I know how do it, but I didn't try. I thought the best practice was to set `push_preview=true` temporarily for docs changes like this one as per your first comment in https://github.com/CliMA/Oceananigans.jl/pull/2870.

But fair enough. I'll build and view them locally for review.

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+# ```math
+# h(x) = -H + h_0 \exp(-x^2 / 2σ^2)
+# ```
+
+h₀ = 50 # m
+width = 5kilometers
+hill(x) = h₀ * exp(-x^2 / 2width^2)
+bottom(x, y) = - H + hill(x)
+
+grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom))
```
simone-silvestri, line 50: This example is probably the perfect test for `PartialCellBottom`

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+nothing #hide
+
+# ## Model
+
+# We built a `HydrostaticFreeSurfaceModel` with an `ImplicitFreeSurface` solver.
+
+using Oceananigans.Models.HydrostaticFreeSurfaceModels: FFTImplicitFreeSurfaceSolver
+
+fft_preconditioner = FFTImplicitFreeSurfaceSolver(grid)
+free_surface = ImplicitFreeSurface(solver_method=:PreconditionedConjugateGradient, preconditioner=fft_preconditioner)
```
simone-silvestri, line 121: `SplitExplicitFreeSurface(; grid, cfl = 0.7, max_Δt = 3minutes)` would give you about 6 barotropic steps, so probably it's faster

examples/barotropic_tide.jl
navidcy: Yes, or even better for shaved cells!
I'd be very interested to see if these streaks of w-velocities in the initial condition diminish with shaved cells.

navidcy:
> > Could you build the docs locally?
> 
> I know how do it, but I didn't try. I thought the best practice was to set `push_preview=true` temporarily for docs changes like this one as per your first comment in #2870.
> 
> But fair enough. I'll build and view them locally for review.

Thanks. It's quite cumbersome to delete the previews from the OceananigansDocumentation branch and thus I'm very reluctant doing it. Sorry :(

examples/barotropic_tide.jl
navidcy: There are a few bugs that didn't allow me to put SplitExplicit here the way you suggest. E.g, substep is not ensured to be integer.

#3131 deals with it; could you have a look there @simone-silvestri?

examples/barotropic_tide.jl
navidcy: cc @siddharthabishnu

examples/barotropic_tide.jl
navidcy: @simone-silvestri 

```julia
julia> free_surface = SplitExplicitFreeSurface(; grid, cfl = 0.7, max_Δt)
SplitExplicitFreeSurface with 51895 steps
```

navidcy: this happens, I believe, because for a Flat dimension, e.g., like y here

```Julia
julia> minimum_yspacing(grid)
1.0
```

navidcy: cc @glwagner remember why we made this choice? I believe there was a reason...

we should either change `minimum_ψspacing` for flat ψ coordinates or ensure the the `substeps` computation in the SplitExplicit takes this into account

navidcy: have a look at b395445

navidcy: omg, with the fix the SplitExplicit rips!

navidcy: @djlikesdjs we should change your experiments to use this free surface solver -- I can help you do it

navidcy:
So I tried with SplitExplicit and 1) it's *much faster* but 2) I see this very fast motion in the beginning...

Compare this

https://github.com/CliMA/Oceananigans.jl/assets/7112768/b098f251-0880-4ead-8f71-a3a5282540d4

with the animation at the top of this PR.

I suppose it's the gravity waves... Are they implicitly filtered out with the implicit free surface solver at the top?

glwagner:
This is an internal tide, not a barotropic tide...

examples/barotropic_tide.jl
glwagner: Why don't we use `PartialCellBottom` here?

docs/make.jl
```diff
@@ -31,6 +31,7 @@ examples = [
     "One-dimensional diffusion"        => "one_dimensional_diffusion",
     "Two-dimensional turbulence"       => "two_dimensional_turbulence",
     "Internal wave"                    => "internal_wave",
+    "Barotropic tide"                  => "barotropic_tide",
```
glwagner, line 34: I don't think "barotropic tide" is a good name for this example. People typically refer to these motions as "internal tide" (which contrasts barotropic tide; ie the internal tide consists of _baroclinic_ motions at the tidal frequency, whereas the barotropic tide is purely barotropic).

navidcy:
Agree! 
"Barotropic tide over seamount" is a good title
or
"Internal tide generation from seamount"

docs/make.jl
navidcy: agreed!

simone-silvestri:
> So I tried with SplitExplicit and 1) it's _much faster_ but 2) I see this very fast motion in the beginning...
> 
> Compare this
> 
>  barotropic_tide_splitexplicit.mp4 
> with the animation at the top of this PR.
> 
> I suppose it's the gravity waves... Are they implicitly filtered out with the implicit free surface solver at the top?

Yeah, the implicit solver is much more diffusive. You can also check out the validation case in `validation/implicit_free_surface/geostrophic_adjustment_test.jl` that shows quite some difference in the dynamics using the implicit solver vs split explicit solver

docs/make.jl
navidcy: see my comment below, which is better title?

internal_tide.jl + "Internal tide generation from a seamount" ?

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+# ```
+
+using Oceananigans
+using Oceananigans.Units
+
+# ## Grid
+
+# We use an immersed boundary two-dimensional grid (in ``x``--``z``) that is is periodic in
+# the ``x``-direction. To construct an immersed boundary grid we first need to create what
+# we refer to as "underlying grid", which the grid that encompasses the immersed boundary.
```
glwagner, line 21: "underlying" and "encompasses" in some way mean the opposite thing, so I think this explanation can be improved.

I would stick to describing how the code works. In the code, `ImmersedBoundaryGrid` is a wrapper around an underlying grid (not the other way around).

I would say something more concise like:

```
We create an `ImmersedBoundaryGrid` wrapped around an underlying two-dimensional `RectilinearGrid` that is periodic in `x` and bounded in `z`.
```

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+Nx, Nz = 200, 80
+
+H  = 2kilometers
+Lx = 2000kilometers
+
+underlying_grid = RectilinearGrid(size = (Nx, Nz),
+                                  x = (-Lx/2, Lx/2),
+                                  z = (-H, 0),
+                                  halo = (4, 4),
+                                  topology = (Periodic, Flat, Bounded))
```
glwagner, line 32: I think it's more readable to write

```julia
underlying_grid = RectilinearGrid(size = (200, 80),
                                  x = (-1000kilometer, 1000kilometers),
                                  z = (-2000, 0),
                                  halo = (4, 4),
```

Lz and Lx may be defined if they are needed later, however.

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+# coordinates and it returns the ``z`` of the bottom, or *(ii)* a two-dimensional array with the values
+# of ``z`` at the bottom at the cell centers.
+#
+# In this example we'd like to have a small Gaussian hill at the center of the domain.
+
+# ```math
+# h(x) = -H + h_0 \exp(-x^2 / 2σ^2)
+# ```
+
+h₀ = 50 # m
```
glwagner, line 45: why so tiny?

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+
+coriolis = FPlane(latitude = -45)
+
+# Now we have everything we require to construct the tidal forcing given a value of the
+# excursion parameter.
+#
+T₂ = 12.421hours
+const ω₂ = 2π / T₂ # radians/sec
+
+ε = 0.25 # the excursion parameter
```
glwagner, line 105:
```suggestion
ε = 0.25 # excursion parameter
```

we dont usually use varepsilon...

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+
+U_tidal = ε * ω₂ * width
+
+const tidal_forcing_amplitude = U_tidal * (ω₂^2 - coriolis.f^2) / ω₂
+
+@inline tidal_forcing(x, y, z, t) = tidal_forcing_amplitude * sin(ω₂ * t)
```
glwagner, line 111: It's probably better to explicitly use `Forcing` for an example. That will also allow you to use `parameters` and avoid `const`, and probalby avoid the `nothing` (since it's nice to `@show Forcing(...)`.

examples/barotropic_tide.jl
glwagner: It's unfortunate that the split explicit free surface always has to be configured (ie its not possible to make it default...)

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+# First, a utility to mask the region that is within the immersed boundary with `NaN`s and
+# also select the ``x``-``z`` interior slices of our fields.
+
+using Oceananigans.ImmersedBoundaries: mask_immersed_field!
+
+function mask_and_get_interior(φ_t, n; value=NaN)
+    mask_immersed_field!(φ_t[n], value)
+    return interior(φ_t[n], :, 1, :)
+end
+nothing #hide
```
glwagner, line 233: We shouldn't have to use this in an example... seems liek something is missing from the API?

examples/barotropic_tide.jl
navidcy: I was using `Lx` in plotting... but I can always switch to `grid.Lx`

examples/barotropic_tide.jl
navidcy: why not?

examples/barotropic_tide.jl
navidcy: I can use `\epsilon` :)
(it's just that ε is the default if I use the greek input from my keyboard...)

examples/barotropic_tide.jl
glwagner: Its hard to see in an animation and requires more vertical resolution. It's also less classical. A "seamount" is a large feature. This is just an isolated hill in an abyssal plain, which doesn't really occur in the ocean.

examples/barotropic_tide.jl
glwagner: Probably makes sense to be consistent, but if we want to change to varepsilon everywhere we should.

examples/barotropic_tide.jl
navidcy: Yes!! #3092 tries to sort this out!

glwagner: The example is a little early then.. convert to draft?

glwagner:
Just for the future I think adding an example is a big deal so I think it might be best to discuss such an addition on an issue before submitting a PR.

examples/barotropic_tide.jl
navidcy: This is not crucial -- I can delete it.

glwagner: Just don't want to encourage hacking 😂

glwagner:
We can delete this validation test too after this: https://github.com/CliMA/Oceananigans.jl/blob/main/validation/immersed_boundaries/internal_tide.jl

docs/make.jl
glwagner: where below?

glwagner: I thikn that sound good, but "by a seamount" not "from a seamount"

navidcy:
> Just for the future I think adding an example is a big deal so I think it might be best to discuss such an addition on an issue before submitting a PR.

Gotcha!

But also happy to close the PR! I had made this example for a demo and I thought it might be useful.

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+
+Nx, Nz = 200, 80
+
+H  = 2kilometers
+Lx = 2000kilometers
+
+underlying_grid = RectilinearGrid(size = (Nx, Nz),
+                                  x = (-Lx/2, Lx/2),
+                                  z = (-H, 0),
+                                  halo = (4, 4),
```
glwagner, line 31: It's too bad we need this. Should we make the default halo size 4?

examples/barotropic_tide.jl
navidcy: Yes, I agree -- I don't like that either!

navidcy: But it's not hacking! It's clean and honest.
But it's using non-user facing functionality which is *definitely* not ideal!

glwagner:
> > Just for the future I think adding an example is a big deal so I think it might be best to discuss such an addition on an issue before submitting a PR.
> 
> Gotcha!
> 
> But also happy to close the PR! I had made this example for a demo and I thought it might be useful.

I think we need more examples with immersed boundaries which was part of our motivation for doing the double gyre with bathymetry with @siddharthabishnu right? I was just also under the impression that we wanted to limit examples to keep the docs build relatively fast. (Or at least, until we had solved the problem with glacially slow docs builds, then we could afford to add more.) Maybe that has changed though, I'm not keeping track of the time it takes to build docs. It used to be 3h.

examples/barotropic_tide.jl
navidcy: well, I'm happy to change it... I wanted to see the internal wave beam

do you have a paper in mind? perhaps I can try to replicate that?

navidcy: I can also call the example "internal tide from an abyssal hill"

examples/barotropic_tide.jl
navidcy: I think we should.
Is there a downside to that?
Memory footprint is really minimal, right?

navidcy:
> > > Just for the future I think adding an example is a big deal so I think it might be best to discuss such an addition on an issue before submitting a PR.
> > 
> > 
> > Gotcha!
> > But also happy to close the PR! I had made this example for a demo and I thought it might be useful.
> 
> I think we need more examples with immersed boundaries which was part of our motivation for doing the double gyre with bathymetry with @siddharthabishnu right? I was just also under the impression that we wanted to limit examples to keep the docs build relatively fast. (Or at least, until we had solved the problem with glacially slow docs builds, then we could afford to add more.) Maybe that has changed though, I'm not keeping track of the time it takes to build docs. It used to be 3h.

It's still 3hrs... :(

Recently there was some suggestions in slack about parallelizing the run of the examples in the doc built but I haven't tried it yet.

examples/barotropic_tide.jl
```diff
@@ -0,0 +1,285 @@
...
+
+fft_preconditioner = FFTImplicitFreeSurfaceSolver(grid)
+free_surface = ImplicitFreeSurface(solver_method=:PreconditionedConjugateGradient, preconditioner=fft_preconditioner)
+
+model = HydrostaticFreeSurfaceModel(; grid, free_surface, coriolis,
+                                      buoyancy = BuoyancyTracer(),
+                                      tracers = :b,
+                                      momentum_advection = WENO(),
+                                      tracer_advection = WENO(),
+                                      forcing = (u = tidal_forcing,))
```
glwagner, line 128:
```suggestion
                                      forcing = (; u = tidal_forcing))
```

examples/barotropic_tide.jl
navidcy: \epsilon is great ;)

tomchor:
Great example @navidcy 

Do we want to mask the hill in the animations? (Especially since https://github.com/CliMA/Oceananigans.jl/pull/3092 isn't merged yet.) The top plot specifically keeps changing color where the hill is supposed to be, which might throw off some readers. This could probably be done with [`Makie.poly`](https://docs.makie.org/dev/examples/plotting_functions/poly/).

navidcy:
Thanks!! I also find the rainbow mountain at top annoying. That’s what the mask_immersed! was trying to deal with.

examples/barotropic_tide.jl
navidcy: > We can delete this validation test too after this: https://github.com/CliMA/Oceananigans.jl/blob/main/validation/immersed_boundaries/internal_tide.jl

The validation script includes a multi region example. Shall I still delete it?
I could also clean it up and add toml files in the validation script directory (see #3076).

examples/internal_tide.jl
```diff
@@ -0,0 +1,273 @@
...
+                                      momentum_advection = WENO(),
+                                      tracer_advection = WENO(),
+                                      forcing = (; u = u_forcing))
+
+# We initialize the model with the tidal flow and a linear stratification.
+
+uᵢ(x, y, z) = U_tidal
+
+Nᵢ² = 1e-4  # [s⁻²] initial buoyancy frequency / stratification
+bᵢ(x, y, z) = Nᵢ² * z
```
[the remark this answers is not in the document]
navidcy: Not sure I understand exactly what you mean. What would be the BC to impose?

examples/barotropic_tide.jl
navidcy: When I tried to change

```Julia
grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom))
```

to 

```Julia
using Oceananigans.ImmersedBoundaries: PartialCellBottom
```

the simulation `NaN`ed immediately. Even when I reduced the `max_Δt` from 6 minutes down to 2 minutes..

glwagner:
> > We can delete this validation test too after this: https://github.com/CliMA/Oceananigans.jl/blob/main/validation/immersed_boundaries/internal_tide.jl
> 
> The validation script includes a multi region example. Shall I still delete it? I could also clean it up and add toml files in the validation script directory (see #3076).

Yeah, that's a random place for a multi region example.

tomchor:
@navidcy I'm trying to add a drag flux to this example like we talked about in https://github.com/CliMA/Oceananigans.jl/issues/3148 but I'm getting an error that I can't figure out. Using `u` as an example, I'm trying to implement the BCs as 


```julia
@inline u_drag(x, y, t, u, v, w, p) = - p.cᵈ * u * √(u^2 + v^2 + w^2)
@inline u_drag(x, y, z, t, u, v, w, cᵈ) = u_drag(x, y, t, u, v, w, cᵈ)

u_drag_bc = FluxBoundaryCondition(u_drag, field_dependencies=(:u, :v, :w), parameters=(; cᵈ=1e-3))
boundary_conditions = (u = FieldBoundaryConditions(bottom=u_drag_bc),)
```

but when time-stepping the model I get

```
ERROR: LoadError: MethodError: no method matching field_arguments(::Int64, ::Int64, ::Int64, ::ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, ::NamedTuple{(:u, :v, :w, :b, :η), Tuple{Field{Face, Center, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Oceananigans.BoundaryConditions.ContinuousBoundaryFunction{Face, Center, Nothing, Oceananigans.BoundaryConditions.LeftBoundary, typeof(u_drag), NamedTuple{(:cᵈ,), Tuple{Float64}}, Tuple{Symbol, Symbol, Symbol}, Tuple{Int64, Int64, Nothing}, Tuple{typeof(Oceananigans.Operators.identity4), typeof(Oceananigans.Operators.ℑxyᶠᶜᵃ), typeof(Oceananigans.Operators.ℑxᶠᵃᵃ)}}}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Face, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, UnitRange{Int64}}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}}}, ::Tuple{typeof(Oceananigans.Operators.identity4), typeof(Oceananigans.Operators.ℑxyᶠᶜᵃ), typeof(Oceananigans.Operators.ℑxᶠᵃᵃ)}, ::Tuple{Int64, Int64, Nothing})

Closest candidates are:
  field_arguments(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Tuple{T, T} where T)
   @ Oceananigans ~/repos/Oceananigans.jl/src/Utils/user_function_arguments.jl:4
  field_arguments(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Tuple{T, T, T} where T)
   @ Oceananigans ~/repos/Oceananigans.jl/src/Utils/user_function_arguments.jl:8
  field_arguments(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Tuple{Any})
   @ Oceananigans ~/repos/Oceananigans.jl/src/Utils/user_function_arguments.jl:1
  ...

Stacktrace:
  [1] user_function_arguments
    @ ~/repos/Oceananigans.jl/src/Utils/user_function_arguments.jl:31 [inlined]
  [2] getbc
    @ ~/repos/Oceananigans.jl/src/BoundaryConditions/continuous_boundary_function.jl:129 [inlined]
  [3] apply_z_bottom_bc!
    @ ~/repos/Oceananigans.jl/src/BoundaryConditions/apply_flux_bcs.jl:125 [inlined]
  [4] macro expansion
    @ ~/repos/Oceananigans.jl/src/BoundaryConditions/apply_flux_bcs.jl:81 [inlined]
  [5] cpu__apply_z_bcs!
    @ ~/.julia/packages/KernelAbstractions/lhhMo/src/macros.jl:276 [inlined]
  [6] cpu__apply_z_bcs!(__ctx__::KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.StaticSize{(250, 1)}, KernelAbstractions.NDIteration.NoDynamicCheck, CartesianIndex{2}, Nothing, KernelAbstractions.NDIteration.NDRange{2, KernelAbstractions.NDIteration.StaticSize{(1, 1)}, KernelAbstractions.NDIteration.StaticSize{(250, 1)}, Nothing, Nothing}}, Gc::Field{Face, Center, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, loc::Tuple{Face, Center, Center}, grid::ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, bottom_bc::BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Oceananigans.BoundaryConditions.ContinuousBoundaryFunction{Face, Center, Nothing, Oceananigans.BoundaryConditions.LeftBoundary, typeof(u_drag), NamedTuple{(:cᵈ,), Tuple{Float64}}, Tuple{Symbol, Symbol, Symbol}, Tuple{Int64, Int64, Nothing}, Tuple{typeof(Oceananigans.Operators.identity4), typeof(Oceananigans.Operators.ℑxyᶠᶜᵃ), typeof(Oceananigans.Operators.ℑxᶠᵃᵃ)}}}, top_bc::BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, args::Tuple{Clock{Float64}, NamedTuple{(:u, :v, :w, :b, :η), Tuple{Field{Face, Center, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Oceananigans.BoundaryConditions.ContinuousBoundaryFunction{Face, Center, Nothing, Oceananigans.BoundaryConditions.LeftBoundary, typeof(u_drag), NamedTuple{(:cᵈ,), Tuple{Float64}}, Tuple{Symbol, Symbol, Symbol}, Tuple{Int64, Int64, Nothing}, Tuple{typeof(Oceananigans.Operators.identity4), typeof(Oceananigans.Operators.ℑxyᶠᶜᵃ), typeof(Oceananigans.Operators.ℑxᶠᵃᵃ)}}}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Face, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, UnitRange{Int64}}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}}}, Nothing, Buoyancy{BuoyancyTracer, Oceananigans.Grids.NegativeZDirection}})
    @ Oceananigans.BoundaryConditions ./none:0
  [7] __thread_run(tid::Int64, len::Int64, rem::Int64, obj::KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(250, 1)}, KernelAbstractions.NDIteration.StaticSize{(250, 1)}, typeof(Oceananigans.BoundaryConditions.cpu__apply_z_bcs!)}, ndrange::Nothing, iterspace::KernelAbstractions.NDIteration.NDRange{2, KernelAbstractions.NDIteration.StaticSize{(1, 1)}, KernelAbstractions.NDIteration.StaticSize{(250, 1)}, Nothing, Nothing}, args::Tuple{Field{Face, Center, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Tuple{Face, Center, Center}, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Oceananigans.BoundaryConditions.ContinuousBoundaryFunction{Face, Center, Nothing, Oceananigans.BoundaryConditions.LeftBoundary, typeof(u_drag), NamedTuple{(:cᵈ,), Tuple{Float64}}, Tuple{Symbol, Symbol, Symbol}, Tuple{Int64, Int64, Nothing}, Tuple{typeof(Oceananigans.Operators.identity4), typeof(Oceananigans.Operators.ℑxyᶠᶜᵃ), typeof(Oceananigans.Operators.ℑxᶠᵃᵃ)}}}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, Tuple{Clock{Float64}, NamedTuple{(:u, :v, :w, :b, :η), Tuple{Field{Face, Center, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Oceananigans.BoundaryConditions.ContinuousBoundaryFunction{Face, Center, Nothing, Oceananigans.BoundaryConditions.LeftBoundary, typeof(u_drag), NamedTuple{(:cᵈ,), Tuple{Float64}}, Tuple{Symbol, Symbol, Symbol}, Tuple{Int64, Int64, Nothing}, Tuple{typeof(Oceananigans.Operators.identity4), typeof(Oceananigans.Operators.ℑxyᶠᶜᵃ), typeof(Oceananigans.Operators.ℑxᶠᵃᵃ)}}}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Face, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Center, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, Tuple{Colon, Colon, UnitRange{Int64}}, OffsetArrays.OffsetArray{Float64, 3, Array{Float64, 3}}, Float64, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing}}}}, Nothing, Buoyancy{BuoyancyTracer, Oceananigans.Grids.NegativeZDirection}}}, dynamic::KernelAbstractions.NDIteration.NoDynamicCheck)
```


However, if I use the same formulation on a NohydrostaticModel this works

```julia
julia> @inline u_drag(x, y, t, u, v, w, p) = - p.cᵈ * u * √(u^2 + v^2 + w^2)
u_drag (generic function with 2 methods)

julia> u_drag_bc = FluxBoundaryCondition(u_drag, field_dependencies=(:u, :v, :w), parameters=(; cᵈ=1e-3))
FluxBoundaryCondition: ContinuousBoundaryFunction u_drag at (Nothing, Nothing, Nothing)

julia> boundary_conditions = (u = FieldBoundaryConditions(bottom=u_drag_bc),);

julia> model = NonhydrostaticModel(; grid, boundary_conditions);

julia> time_step!(model, 1)

julia>
```

So whatever's happening probably has to do with the HydrostaticModel, which I'm really not familiar with unfortunately.

What do you prefer that I do?

navidcy:
> So whatever's happening probably has to do with the HydrostaticModel, which I'm really not familiar with unfortunately.
> 
> What do you prefer that I do?

the hydrostatic model does not have `w` as state variable... try removing `w` and see if all is good.

navidcy:
I update the internal tide example. Let's merge this so we have an example with bathymetry in the docs?

The latest version of the movie is:

https://github.com/CliMA/Oceananigans.jl/assets/7112768/1bfdca5c-6d49-4101-9037-eadfacea6900

I'm not so sure about 979c2a7; is this a good idea?

cc @simone-silvestri, @glwagner

navidcy:
Without https://github.com/CliMA/Oceananigans.jl/commit/979c2a7cd10c1f2054cad996622cafd2e3fb3ffa the movie looks like

https://github.com/CliMA/Oceananigans.jl/assets/7112768/5b199c8c-0faf-4e73-97bf-b03edd40bb25

with the mountain changing colors in the first plot. A bit annoying...

tomchor:
> with the mountain changing colors in the first plot. A bit annoying...

Can't we just plot a fixed-color shape on top on the mount on all panels? [`Makie.poly`](https://docs.makie.org/stable/reference/plots/poly/) seems to be able to do that.

navidcy:
> Can't we just plot a fixed-color shape on top on the mount on all panels? [`Makie.poly`](https://docs.makie.org/stable/reference/plots/poly/) seems to be able to do that.

We can.

We can also mask the output after we load it, e.g, via

using something like

```Julia

function mask_and_get_interior(φ_t, n; value=NaN)
    mask_immersed_field!(φ_t[n], value)
    return interior(φ_t[n], :, 1, :)
end

u′ₙ = @lift mask_and_get_interior(u′_t, $n)
```

But these complicate the example a bit.

navidcy:
But I wouldn't mind adding the post-processing masking step with the hope we can eliminate it when masking is enabled for OutputWriters; see https://github.com/CliMA/Oceananigans.jl/pull/3092

simone-silvestri:
I am in favor of not doing 979c2a7. It hardcodes a change in the source code that is only required for aesthetic reasons. I would wait for #3092 and in the meantime mask the output in the example

navidcy:
> I am in favor of not doing [979c2a7](https://github.com/CliMA/Oceananigans.jl/commit/979c2a7cd10c1f2054cad996622cafd2e3fb3ffa). It hardcodes a change in the source code that is only required for aesthetic reasons. I would wait for #3092 and in the meantime mask the output in the example

Deal, done

francispoulin:
The movie looks great and I'm happy to see this example coming to life!

One question I have is in the plot for stratification, at the very top.  It looks like there is a black line and then there is a thin layer right below that.  Anyone know what that is?

glwagner:
> I am in favor of not doing [979c2a7](https://github.com/CliMA/Oceananigans.jl/commit/979c2a7cd10c1f2054cad996622cafd2e3fb3ffa). It hardcodes a change in the source code that is only required for aesthetic reasons. I would wait for #3092 and in the meantime mask the output in the example

I agree with @simone-silvestri that we want this feature, but that it should be done properly

glwagner:
> > Can't we just plot a fixed-color shape on top on the mount on all panels? [`Makie.poly`](https://docs.makie.org/stable/reference/plots/poly/) seems to be able to do that.
> 
> We can.
> 
> We can also mask the output after we load it, e.g, via
> 
> using something like
> 
> ```julia
> using Oceananigans.ImmersedBoundaries: mask_immersed_field!
> 
> function mask_and_get_interior(φ_t, n; value=NaN)
>     mask_immersed_field!(φ_t[n], value)
>     return interior(φ_t[n], :, 1, :)
> end
> 
> u′ₙ = @lift mask_and_get_interior(u′_t, $n)
> ```
> 
> which gives
> 
>  internal_tide.mp4 
> But either of these solutions complicate the example a bit. Ideally, `mask_immersed_field!` should not be user-facing. And plotting a mountain on top of the mountain is a bit of a hack and would require a bit of explaining and justifying why we do that.

I think we should mask the solution during visualization in the script. This is actually what users will need to do currently to make decent visualizations, so it is good to illustrate how to do it --- even if, hopefully, we will eventually have a better solution.

As for plotting a shape on top, I'm inclined to encourage visualization that directly represents the domain / data for the purposes of an Oceananigans example. I think its ok if people want to incorporate visualizations like that in their own work but we may not want to promote it as "the best" way to visualize complex domains in this example.

navidcy:
I also agree. That’s how the example is now. If you approve then please approve?

navidcy:
Masking is done:

https://github.com/CliMA/Oceananigans.jl/blob/ca437145013e82380bbbc8b1d700656eeba78607/examples/internal_tide.jl#L213-L217

navidcy:
> One question I have is in the plot for stratification, at the very top. It looks like there is a black line and then there is a thin layer right below that. Anyone know what that is?

@francispoulin nice catch!
You are correct... it's because `b` has no flux boundary conditions and therefore the first halo point outside the interior (top and bottom) have values exactly equal to the values just inside (so that the derivative is zero). And when we compute `N² = ∂z(b)` the top and bottom values are zero.


```Julia
using Oceananigans

Nx, Nz = 25, 12

underlying_grid = RectilinearGrid(size = (Nx, Nz), x = (-1000e3, 1000e3), z = (-2000, 0),
                                  halo = (4, 4), topology = (Periodic, Flat, Bounded))

h₀, width = 250, 20e3
bottom(x) = - H + h₀ * exp(-x^2 / 2width^2)

grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom))

model = HydrostaticFreeSurfaceModel(; grid, buoyancy = BuoyancyTracer(), tracers = :b)

Nᵢ² = 1e-4
bᵢ(x, z) = Nᵢ² * z
set!(model, b=bᵢ)

b = model.tracers.b
N² = Field(∂z(b))
compute!(N²)

using GLMakie

xb, _, zb = nodes(b)
heatmap(xb, zb, interior(b, :, 1, :))
```
b

```Julia
xN², _, zN² = nodes(N²)
heatmap(xN², zN², interior(N², :, 1, :))
```
N2

I don't know what would be the solution to this. I think it's physical in our domain to have no flux condition of `b` and therefore `∂z(b) = 0` at top and bottom, right?

francispoulin:
Makes sense.

In this case would it be better to specify a non zero flux on the buoyancy, the ambient stratification?  It won't change the nice results you are getting over the topography but makes sense to me.  No need to make this change.

navidcy:
Hm.... `HydrostaticFreeSurfaceModel` does not support `BackgroundFields`....

glwagner:
> Masking is done:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/ca437145013e82380bbbc8b1d700656eeba78607/examples/internal_tide.jl#L213-L217

That's great.

glwagner:
> Hm.... `HydrostaticFreeSurfaceModel` does not support `BackgroundFields`....



> Hm.... `HydrostaticFreeSurfaceModel` does not support `BackgroundFields`....

Right, we do not support that right now. I think there are some niche uses but, for the most part, when we go hydrostatic we are often doing large scale and a bit more complicated modeling with parameterizations, etc, where getting background fields right would be more of a pain than with NonhydrostaticModel (which still does not work properly with nonlinear subgrid-scale turbulence closures).

I think its better to go more mainstream for this example. I think the use cases that this example serves as a starting point for are probably not going to use background fields, typically.

navidcy:
> > Hm.... `HydrostaticFreeSurfaceModel` does not support `BackgroundFields`....
> 
> > Hm.... `HydrostaticFreeSurfaceModel` does not support `BackgroundFields`....
> 
> Right, we do not support that right now. I think there are some niche uses but, for the most part, when we go hydrostatic we are often doing large scale and a bit more complicated modeling with parameterizations, etc, where getting background fields right would be more of a pain than with NonhydrostaticModel (which still does not work properly with nonlinear subgrid-scale turbulence closures).
> 
> I think its better to go more mainstream for this example. I think the use cases that this example serves as a starting point for are probably not going to use background fields, typically.

True true.

Any idea how to deal with the dz(b) that vanishes at the top? Or should we live with it?

glwagner:
> > > Hm.... `HydrostaticFreeSurfaceModel` does not support `BackgroundFields`....
> > 
> > 
> > > Hm.... `HydrostaticFreeSurfaceModel` does not support `BackgroundFields`....
> > 
> > 
> > Right, we do not support that right now. I think there are some niche uses but, for the most part, when we go hydrostatic we are often doing large scale and a bit more complicated modeling with parameterizations, etc, where getting background fields right would be more of a pain than with NonhydrostaticModel (which still does not work properly with nonlinear subgrid-scale turbulence closures).
> > I think its better to go more mainstream for this example. I think the use cases that this example serves as a starting point for are probably not going to use background fields, typically.
> 
> True true.
> 
> Any idea how to deal with the dz(b) that vanishes at the top? Or should we live with it?

Because of the no-flux boundary condition? I think that's correct, ok to have. You could have a mixed layer at the top too!

navidcy:
hm... I don't know about a mixed layer...

let's just ignore the dz(b)=0 at the top&bottom _or_ we can add a remark in the example about it?

glwagner:
> hm... I don't know about a mixed layer...
> 
> let's just ignore the dz(b)=0 at the top&bottom _or_ we can add a remark in the example about it?

Yes definitely ignore.

What we want to focus on is providing an example that is a useful starting point for other, new simulations. This is the key priority.

This is why it makes sense to prioritize relatively simple examples (and also simple visualization) that can be quickly digested and are amenable to generalization.

We want to _deprioritize_ bespoke or niche cases that are a dead-end. For example, if we use `GradientBoundaryCondition` at the top in this case, then most users will have to delete / change that line in order to generalize this example into the case they are interested in using.

glwagner (approved):
This is a great start and I think in the future we can think about adding bottom drag using `ImmersedBoundaryCondition`

PR #3135, opened 2023-06-05
Use multiple processes to execute `examples` in make.jl

glwagner:
Let's see if this speeds up the docs build.


docs/make.jl
```diff
@@ -46,7 +46,10 @@ example_scripts = [ filename * ".jl" for (title, filename) in examples ]
 asyncmap(1:length(example_scripts)) do n
     example = example_scripts[n]
     example_filepath = joinpath(EXAMPLES_DIR, example)
-    Literate.markdown(example_filepath, OUTPUT_DIR; execute=true, flavor = Literate.DocumenterFlavor())
+    withenv("JULIA_DEBUG" => "Literate") do
+        Literate.markdown(example_filepath, OUTPUT_DIR;
+                         flavor = Literate.DocumenterFlavor(), execute = true)
```
glwagner, line 51:
```suggestion
                          flavor = Literate.DocumenterFlavor(), execute = true)
```

navidcy:
@fredrikekre, @mortenpi

We tried here what we were chatting on slack about parallelizing the execution of literated examples but we got this:

https://buildkite.com/clima/oceananigans/builds/11572#01888a04-bcd4-490b-a0d9-c69d6586c3ca

Any ideas?

(When I tried to build the docs locally I got the same error...)

navidcy:
> As we also discussed on slack, using threads (or tasks) won't work since to capture the output the global stdout variable is changed. You need `pmap` or launch new processes with `run`.

So what we did here was
https://github.com/CliMA/Oceananigans.jl/blob/5d9836398cb49cf2eea38ec6da505b03d9a94a66/docs/make.jl#L46-L53
which is equivalent to `pmap`, right?

But now I see. I thought the `run('...')` was only if I wanted to carry out global variables from one example to the other.
So you are saying that 

```Julia
 asyncmap(1:length(example_scripts)) do n 
     example = example_scripts[n] 
     example_filepath = joinpath(EXAMPLES_DIR, example) 
     withenv("JULIA_DEBUG" => "Literate") do 
         run(`julia --project=docs -e'Literate.markdown(example_filepath, OUTPUT_DIR;  flavor = Literate.DocumenterFlavor(), execute = true)'`
     end 
 end
```
right? I'll try that.

fredrikekre:
No, `asyncmap` just creates tasks in the same process. `Distributed.pmap` would run the code on a separate process.

navidcy:
This made a difference! 20-30% reduction in docs building time.

navidcy (approved):

fredrikekre:
Perhaps I am missing something, but where do you launch new workers?

glwagner:
I think here:

https://github.com/CliMA/Oceananigans.jl/blob/48f30b3244cff5ca287e32d8f4f35ee2558234ed/docs/make.jl#L47

I feel 20-30% speed up is not that great. It surprised me at first that launching new processes for every example is faster, because I believe we are dominated by compile time rather than run time here. But I might misunderstand --- is the `Literate.markdown` always executed in an independent process, so we always recompile every file?

glwagner:
Well, it looks like we can benefit from recompilation if we launch more than one example in a single process. So perhaps there is a sweet spot between purely serial execution and purely parallel.

glwagner:
> Could you add [`Distributed.nprocs`](https://docs.julialang.org/en/v1/stdlib/Distributed/#Distributed.nprocs) as an informational print? It seems like you're still running everything in a single process.. so not quite sure where you'd get any speedup from. What Fredrik was alluding to, I am pretty sure you'd need to do an explicit [`Distributed.addprocs`](https://docs.julialang.org/en/v1/stdlib/Distributed/#Distributed.addprocs) before the `pmap` (or give `julia` the `-p` option when starting).

Thank you @mortenpi and @fredrikekre, very helpful!

navidcy:
```
julia> using Distributed

help?> addprocs
search: addprocs

  addprocs(manager::ClusterManager; kwargs...) -> List of process identifiers


  Launches worker processes via the specified cluster manager.

  For example, Beowulf clusters are supported via a custom cluster manager implemented in the package ClusterManagers.jl.

  The number of seconds a newly launched worker waits for connection establishment from the master can be specified via variable
  JULIA_WORKER_TIMEOUT in the worker process's environment. Relevant only when using TCP/IP as transport.

  To launch workers without blocking the REPL, or the containing function if launching workers programmatically, execute addprocs in its own task.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  # On busy clusters, call `addprocs` asynchronously
  t = @async addprocs(...)


  # Utilize workers as and when they come online
  if nprocs() > 1   # Ensure at least one new worker is available
     ....   # perform distributed execution
  end


  # Retrieve newly launched worker IDs, or any error messages
  if istaskdone(t)   # Check if `addprocs` has completed to ensure `fetch` doesn't block
      if nworkers() == N
          new_pids = fetch(t)
      else
          fetch(t)
      end
  end


  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  addprocs(machines; tunnel=false, sshflags=``, max_parallel=10, kwargs...) -> List of process identifiers


  Add worker processes on remote machines via SSH. Configuration is done with keyword arguments (see below). In particular, the exename keyword
  can be used to specify the path to the julia binary on the remote machine(s).

  machines is a vector of "machine specifications" which are given as strings of the form [user@]host[:port] [bind_addr[:port]]. user defaults to
  current user and port to the standard SSH port. If [bind_addr[:port]] is specified, other workers will connect to this worker at the specified
  bind_addr and port.

  It is possible to launch multiple processes on a remote host by using a tuple in the machines vector or the form (machine_spec, count), where
  count is the number of workers to be launched on the specified host. Passing :auto as the worker count will launch as many workers as the number
  of CPU threads on the remote host.

  Examples:

  addprocs([
      "remote1",               # one worker on 'remote1' logging in with the current username
      "user@remote2",          # one worker on 'remote2' logging in with the 'user' username
      "user@remote3:2222",     # specifying SSH port to '2222' for 'remote3'
      ("user@remote4", 4),     # launch 4 workers on 'remote4'
      ("user@remote5", :auto), # launch as many workers as CPU threads on 'remote5'
  ])


  Keyword arguments:

    •  tunnel: if true then SSH tunneling will be used to connect to the worker from the master process. Default is false.

    •  multiplex: if true then SSH multiplexing is used for SSH tunneling. Default is false.

    •  ssh: the name or path of the SSH client executable used to start the workers. Default is "ssh".

    •  sshflags: specifies additional ssh options, e.g. sshflags=`-i /home/foo/bar.pem`

    •  max_parallel: specifies the maximum number of workers connected to in parallel at a host. Defaults to 10.

    •  shell: specifies the type of shell to which ssh connects on the workers.
       • shell=:posix: a POSIX-compatible Unix/Linux shell (sh, ksh, bash, dash, zsh, etc.). The default.
       • shell=:csh: a Unix C shell (csh, tcsh).
       • shell=:wincmd: Microsoft Windows cmd.exe.

    •  dir: specifies the working directory on the workers. Defaults to the host's current directory (as found by pwd())

    •  enable_threaded_blas: if true then BLAS will run on multiple threads in added processes. Default is false.

    •  exename: name of the julia executable. Defaults to "$(Sys.BINDIR)/julia" or "$(Sys.BINDIR)/julia-debug" as the case may be. It is
       recommended that a common Julia version is used on all remote machines because serialization and code distribution might fail
       otherwise.

    •  exeflags: additional flags passed to the worker processes.

    •  topology: Specifies how the workers connect to each other. Sending a message between unconnected workers results in an error.
       • topology=:all_to_all: All processes are connected to each other. The default.
       • topology=:master_worker: Only the driver process, i.e. pid 1 connects to the workers. The workers do not connect to each
       other.
       • topology=:custom: The launch method of the cluster manager specifies the connection topology via fields ident and
       connect_idents in WorkerConfig. A worker with a cluster manager identity ident will connect to all workers specified in
       connect_idents.

    •  lazy: Applicable only with topology=:all_to_all. If true, worker-worker connections are setup lazily, i.e. they are setup at the first
       instance of a remote call between workers. Default is true.

    •  env: provide an array of string pairs such as env=["JULIA_DEPOT_PATH"=>"/depot"] to request that environment variables are set on the
       remote machine. By default only the environment variable JULIA_WORKER_TIMEOUT is passed automatically from the local to the remote
       environment.

    •  cmdline_cookie: pass the authentication cookie via the --worker commandline option. The (more secure) default behaviour of passing the
       cookie via ssh stdio may hang with Windows workers that use older (pre-ConPTY) Julia or Windows versions, in which case
       cmdline_cookie=true offers a work-around.

  │ Julia 1.6
  │
  │  The keyword arguments ssh, shell, env and cmdline_cookie were added in Julia 1.6.

  Environment variables:

  If the master process fails to establish a connection with a newly launched worker within 60.0 seconds, the worker treats it as a fatal
  situation and terminates. This timeout can be controlled via environment variable JULIA_WORKER_TIMEOUT. The value of JULIA_WORKER_TIMEOUT on the
  master process specifies the number of seconds a newly launched worker waits for connection establishment.

  ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  addprocs(np::Integer=Sys.CPU_THREADS; restrict=true, kwargs...) -> List of process identifiers


  Launch np workers on the local host using the in-built LocalManager.

  Local workers inherit the current package environment (i.e., active project, LOAD_PATH, and DEPOT_PATH) from the main process.

  Keyword arguments:

    •  restrict::Bool: if true (default) binding is restricted to 127.0.0.1.

    •  dir, exename, exeflags, env, topology, lazy, enable_threaded_blas: same effect as for SSHManager, see documentation for
       addprocs(machines::AbstractVector).

  │ Julia 1.9
  │
  │  The inheriting of the package environment and the env keyword argument were added in Julia 1.9.
```

Ha! I was testing locally with v1.9 but the docs use v1.8.5
Seems like from v1.9 onwards the package environment is inherited so we don't need to instantiate multiple times.

navidcy:
With 4 workers docs built in 1hr (instead of 3hr)

navidcy:
@glwagner, turns out that most of the reduction was due to the modifications of the BCI example. In #3139 with just 3 workers we got 2h30.

glwagner:
That makes a lot of sense. For documentation the BCI example was taking 1.6 hours to run. I changed resolution, run time, and plotting to get it down to about 10-15 minutes (I think). It's still the most expensive example by far.

I would use 2 workers: one for the BCI example (and perhaps a second expensive one), and one for all the rest. If we can control threads-per-worker, we should put most of the threads on the heavy examples. Maybe that will speed things up?

docs/make.jl
```diff
@@ -1,19 +1,87 @@
...
+        example_filepath = joinpath(EXAMPLES_DIR, example)
+        withenv("JULIA_DEBUG" => "Literate") do
+            Literate.markdown(example_filepath, OUTPUT_DIR;
+                            flavor = Literate.DocumenterFlavor(), execute = true)
+        end
+    end
+
+    Distributed.rmprocs()
+
+elseif use_extra_workers == 0
```
glwagner, line 84: Is there a way to avoid repeating this code?

navidcy: this is ONLY for benchmarking

I want to see how long the docs take with 0, 1, 2 extra workers

and then we can clean up this of course.

navidcy: (I could benchmark locally but it beats the purpose a bit... I want to see whether these changes have an effect on tartarus)

navidcy (changes requested):
we need to bench first before merging

navidcy:
don't use workers: 1h48
`Distributed.addprocs(1)`: 1h41
`Distributed.addprocs(2)`: 
`Distributed.addprocs(3)`:

navidcy:
OK, in view of the benchmarks at https://github.com/CliMA/Oceananigans.jl/pull/3135#issuecomment-1589191799 seems that the additional workers do something but not that much? With 4 workers we can bring the docs from 1h45 -> 1h. Shall we do that for now @glwagner?

simone-silvestri:
I suggest putting back the complete testing suite and seeing if reducing documentation build time from 1.45 to 1hr decreases overall test time and how much. If it does not give a substantial decrease in test time, I would not bother not to overload Tartarus (already a 1.75 speedup with 4 cores is probably not worth so much)

simone-silvestri:
4 workers increase a lot the pressure on tartarus if we are testing multiple branches at the same time

glwagner:
> 4 workers increase the pressure on Tartarus a lot if we are testing multiple branches at the same time

what's wrong with that?

glwagner:
How about 2 workers: one for the most expensive example, and one for the rest?

navidcy:
> How about 2 workers: one for the most expensive example, and one for the rest?

I'm happy with this. But why specify, let both workers run examples and who ever is done with one example it'll start running one of the ones left.

glwagner:
> > How about 2 workers: one for the most expensive example, and one for the rest?
> 
> I'm happy with this. But why specify, let both workers run examples and who ever is done with one example it'll start running one of the ones left.

Good point

navidcy (approved):

navidcy:
I'm happy with this PR. I think we should merge when tests pass. How about others?

tomchor:
> I'm happy with this PR. I think we should merge when tests pass. How about others?

Agreed! Great PR.

tomchor (approved):

simone-silvestri (approved):

glwagner:
I'll take a look at this tomorrow

navidcy:
It’s merged. But open for alternations!

glwagner:
@navidcy did we order the examples so that the most expensive ones are built first? It looks like we are using:

https://github.com/CliMA/Oceananigans.jl/blob/87e3d770cf6b4502a891dbc587e1036968f36cb2/docs/make.jl#L70

which actually generates the least expensive examples first.

navidcy:
No we didn’t change the order. The order here is as they appear in the docs. 

Did you time the examples?
Also, does it matter if the expensive ones tun furst? What’s the difference?

PR #3139, opened 2023-06-07
use Distributed.pmap for docs examples

navidcy:


glwagner:
What's the point?

navidcy:
> What's the point?

Only to benchmark against the examples before you simplified them to see how faster the docs built. The other pr also includes a simpler bci example now.

glwagner:
I think we should test the difference between using pmap or not on the other PR

navidcy:
OK.

But as a point of reference, this PR only reduced docs built from 3h to 2h30 --> https://buildkite.com/clima/oceananigans/builds/11644#018897e6-4547-44d2-9644-d8eeed5d1957

glwagner:
Ok, good to know. But I think the strategy we use to distribute work should depend specifically on how much each example costs. So the optimal strategy for this code might be different than on the other PR.

PR #2990, opened 2023-03-20
Flips sign for `gravity_unit_vector` to match its description (attempt #2)

tomchor:
This is the second attempt at flipping the sign of `gravity_unit_vector`. The first attempt (https://github.com/CliMA/Oceananigans.jl/pull/2963) had an issue where the doctests were getting stuck with no apparent cause.

For this attempt, I proceeded more carefully, step by step to try and pinpoint what was happening. The good news is that the doctests are now passing locally. The bad news is that I still don't know why they were getting stuck in https://github.com/CliMA/Oceananigans.jl/pull/2963.

Since tests pass locally for me (I tested everything except the examples) I'm hopefully that this time they'll also pass on buildkite.

@navidcy your modifications to https://github.com/CliMA/Oceananigans.jl/pull/2963 are also included here!

PR #3142, opened 2023-06-08
(v0.85.0) Switching cells where fluxes are applied in `ImmersedBoundaryCondition` to match `FieldBoundaryConditions`' kwargs

### LINKED ISSUE #3141: `ImmersedBoundaryCondition` has the opposite convention of `FieldBoundaryConditions`
The current implementation of `ImmersedBondaryCondition` has opposite conventions as `FieldBoundaryConditions`.

for example, to have a bottom drag, you have to write
```
u_bc = FieldBoundaryConditions(bottom = drag_u)
```
but to have a drag above an immersed boundary
```
u_immersed = ImmersedBoundaryCondition(top = drag_u)
```

I think it is better to assume that immersed and non-immersed boundaries have the same conventions. (I ran for a lot of time an ocean simulation without drag because I haven't looked at the implementation and assumed the convention was the same)

I propose to change the convention to have 
```
u_immersed = ImmersedBoundaryCondition(bottom = drag_u)
```

(I guess we will need to invert `index_left` and `index_right` in 
https://github.com/CliMA/Oceananigans.jl/blob/6ab0c77ec7cb51b768ee30e248d633c98bb7e427/src/ImmersedBoundaries/immersed_boundary_condition.jl#L200-L201)

found by @Yixiao-Zhang

simone-silvestri:
see #3141 
closes #3141 

The results from `validation/immersed_boundaries/immersed_couette_flow.jl` from this branch

https://github.com/CliMA/Oceananigans.jl/assets/33547697/e50beac1-4d22-4581-9715-5f541e8dcf01


glwagner:
Wasn't immersed Couette flow doing the right thing previously? Which one is old and which is new?

simone-silvestri:
Those are `c` and `u` from the couette flow simulation in this branch. Couette flow was doing the right thing before also because immersed boundaries is implemented as
``` 
c_immersed_bc = ValueBoundaryCondition(1)
c_top_bc = ValueBoundaryCondition(-1)
c_bcs = FieldBoundaryConditions(immersed=c_immersed_bc, top=c_top_bc)
```
Effectively putting an immersed boundary on both the immersed top and the immersed bottom. 
it would have failed if 
```
c_imm_bc = ValueBoundaryCondition(1)
c_top_bc = ValueBoundaryCondition(-1)
c_immersed_bc = ImmersedBoundaryCondition(bottom = c_imm_bc)
c_bcs = FieldBoundaryConditions(immersed=c_immersed_bc, top=c_top_bc)
```

glwagner:
> Those are `c` and `u` from the couette flow simulation in this branch. Couette flow was doing the right thing before also because immersed boundaries is implemented there as
> 
> ```
> c_immersed_bc = ValueBoundaryCondition(1)
> c_top_bc = ValueBoundaryCondition(-1)
> c_bcs = FieldBoundaryConditions(immersed=c_immersed_bc, top=c_top_bc)
> ```
> 
> Effectively putting an immersed boundary on both the immersed top and the immersed bottom. it would have failed if
> 
> ```
> c_imm_bc = ValueBoundaryCondition(1)
> c_top_bc = ValueBoundaryCondition(-1)
> c_immersed_bc = ImmersedBoundaryCondition(bottom = c_imm_bc)
> c_bcs = FieldBoundaryConditions(immersed=c_immersed_bc, top=c_top_bc)
> ```

I'm confused. How does it get the right result in the first case?

simone-silvestri:
In main:

since we write the immersed boundary condition as 
```
c_immersed_bc = ValueBoundaryCondition(1)
c_top_bc = ValueBoundaryCondition(-1)
c_bcs = FieldBoundaryConditions(immersed=c_immersed_bc, top=c_top_bc)
```

immersed boundary conditions are defined on all facets, and in particular, what is important in this case, is that bc are assigned to the `top` and `bottom` fields. Now, in this case `bottom` assigns a flux to the cell _below_ the immersed boundary and, therefore, it does nothing (there is no immersed _ceiling_). 

The heavy lifting is done by the `top` boundary condition, which assigns a flux to the cell _above_ the immersed boundary.
Therefore, if instead of that we would have written:
```
c_imm_bc = ValueBoundaryCondition(1)
c_top_bc = ValueBoundaryCondition(-1)
c_immersed_bc = ImmersedBoundaryCondition(bottom = c_imm_bc)
c_bcs = FieldBoundaryConditions(immersed=c_immersed_bc, top=c_top_bc)
```
The simulation would not have had the intended outcome since in this case
```
julia> c_immersed_bc = ImmersedBoundaryCondition(bottom = c_imm_bc)
ImmersedBoundaryCondition:
├── west: Nothing
├── east: Nothing
├── south: Nothing
├── north: Nothing
├── bottom: ValueBoundaryCondition: 1
└── top: Nothing
```

with the `bottom` field being inactive

simone-silvestri:
on main in case of 
```
c_imm_bc = ValueBoundaryCondition(1)
c_top_bc = ValueBoundaryCondition(-1)
c_immersed_bc = ImmersedBoundaryCondition(bottom = c_imm_bc)
c_bcs = FieldBoundaryConditions(immersed=c_immersed_bc, top=c_top_bc)
```
and
```
u_imm_bc = ValueBoundaryCondition(-1)
u_top_bc = ValueBoundaryCondition(1)
u_immersed_bc = ImmersedBoundaryCondition(bottom = u_imm_bc)
u_bcs = FieldBoundaryConditions(immersed=u_immersed_bc, top=u_top_bc)
```
this is the result

https://github.com/CliMA/Oceananigans.jl/assets/33547697/46e13eef-5c39-4637-8c33-206196380a3a

while
```
c_imm_bc = ValueBoundaryCondition(1)
c_top_bc = ValueBoundaryCondition(-1)
c_immersed_bc = ImmersedBoundaryCondition(top = c_imm_bc)
c_bcs = FieldBoundaryConditions(immersed=c_immersed_bc, top=c_top_bc)
```
and
```
u_imm_bc = ValueBoundaryCondition(-1)
u_top_bc = ValueBoundaryCondition(1)
u_immersed_bc = ImmersedBoundaryCondition(top = u_imm_bc)
u_bcs = FieldBoundaryConditions(immersed=u_immersed_bc, top=u_top_bc)
```
gives

https://github.com/CliMA/Oceananigans.jl/assets/33547697/b3450270-e725-48e0-9ae8-92b206b98101

tomchor (approved):
Nice catch @Yixiao-Zhang and @simone-silvestri!

IMO this is kind of a serious bug. It affected several of my simulations, for example, and I'm just now finding out. So can we release a patch version with this fix soon please?

glwagner:
Should we discourage using `ImmersedBoundaryCondition`?

simone-silvestri:
I think it is good to use it in hydrostatic simulations where the aspect ratio is very high (for example drag on the sides should not be equal to drag on the bottom)

glwagner:
Yes, I see now that it is basically a convenience feature (ie friendly), which allows users to build models that satisfy global constraints without having to, for example, compute the bottom surface area of their discrete geometry. More discussion on #3141 .

My only misgiving is that I think it obscures the physics of a problem and can lead to confusion or modeling choices that don't make sense. That said, it's very difficult to achieve convenience without also allowing wrong modeling choices.

glwagner:
Probably makes sense to make sure this doesn't get reverted in the future too

docs/src/model_setup/boundary_conditions.md
```diff
@@ -487,8 +487,17 @@ Oceananigans.FieldBoundaryConditions, with boundary conditions
 └── immersed: ImmersedBoundaryCondition with west=Nothing, east=Nothing, south=Nothing, north=Nothing, bottom=Value, top=Nothing
 ```
 
-A boundary condition that depends on the fields may be prescribed using the `immersed` keyword argument in [`FieldBoundaryConditions`](@ref).
-We illustrate field-dependent boundary conditions with an example that imposes linear bottom drag on `u` on both immersed facets _and_ the bottom boundary of the underlying grid.
+!!! warning "`ImmersedBoundaryCondition`"
+    Manually building `ImmersedBoundaryCondition` should be done with care.
+    While convenient for some applications, using bespoke `ImmersedBoundaryCondition`s
+    can obscure the physical interpreation of a model setup. As with all model setups, it is important
+    to test important model choices in idealized situations where their impact can be easily discerned.
```
tomchor, line 494: I don't know about this. I guess I don't really understand what's bad about `ImmersedBoundaryCondition`. I agree, it's a bit less clear than the alternative, but I don't understand why it merits a warning. 

The same goes for the part about model choices. It's true, but I guess that's true of all other important choices and I don't see why it needs a warning.

Is there any reason why we don't just test `ImmersedBoundaryCondition` more thoroughly and treat it like any other BC in the docs?

tomchor:
> I think it is good to use it in hydrostatic simulations where the aspect ratio is very high (for example drag on the sides should not be equal to drag on the bottom)

Also if we use a Monin-Obukhov-based drag law the drag coefficients will depend on the grid-spacing in that direction, meaning they'd need to be applied separately if the spacing is different in different directions.

glwagner:
> Also if we use a Monin-Obukhov-based drag law the drag coefficients will depend on the grid-spacing in that direction, meaning they'd need to be applied separately if the spacing is different in different directions.

That's an interesting example. This wouldn't generalize to cut cells, right? Perhaps we need an abstraction that represents the distances to the boundary, for use within boundary conditions. That way we can ensure the distance to the boundary is computed correctly regardless of how the boundary is represented.

As an aside, hasn't it been shown that MOST is not valid in complex terrain? (That wouldn't prevent it from still somehow improving model fidelity, at least in principle.)

docs/src/model_setup/boundary_conditions.md
glwagner: I think we should include some comment that encourages using ordinary boundary conditions. I'm worried that the name is confusing: users could interpret `ImmersedBoundaryCondition` as being the only way to specify a condition on immersed boundaries.

More generally, I think that we can interpret cases that require this sort of detailed boundary condition construction (which typically would not generalize well) as a situation in which possibly more source code / more features are needed. In other words, we support this because it's not possible to support all boundary conditions models in a nice general fashion within Oceananigans. But I'd hope that through development efforts that lower-level aspects of the code, like `ImmersedBoundaryCondition`, would be used less often than higher-level, more generalizable (and testable) abstractions.

glwagner: The problem with complex implementations in user scripts is that they are harder to test and validate. Generally I think we try to design the code to encourage reproducible and high-quality science. I'm seeing a sort of gap here, where it's possible to "see between the cracks". Something similar goes on with `DiscreteForcing`, though our comment there is a bit more mild "Using discrete forcing functions may require understanding the staggered arrangement of velocity fields and tracers..."

https://clima.github.io/OceananigansDocumentation/stable/model_setup/forcing_functions/#%22Discrete-form%22-forcing-functions

tomchor:
> > Also if we use a Monin-Obukhov-based drag law the drag coefficients will depend on the grid-spacing in that direction, meaning they'd need to be applied separately if the spacing is different in different directions.
> 
> That's an interesting example. This wouldn't generalize to cut cells, right? Perhaps we need an abstraction that represents the distances to the boundary, for use within boundary conditions. That way we can ensure the distance to the boundary is computed correctly regardless of how the boundary is represented.

Agreed. Although I think that's for another PR, right?

> As an aside, hasn't it been shown that MOST is not valid in complex terrain? (That wouldn't prevent it from still somehow improving model fidelity, at least in principle.)

Yes, although the discrepancy is less severe for neutral stability, which I suspect encompasses most of the simulations people have been using Oceananigans for. Interestingly, there's been a very recent attempt at generalization that [looks pretty promising](https://journals.aps.org/prl/pdf/10.1103/PhysRevLett.130.124001)!

docs/src/model_setup/boundary_conditions.md
tomchor: > The problem with complex implementations in user scripts is that they are harder to test and validate. Generally I think we try to design the code to encourage reproducible and high-quality science. I'm seeing a sort of gap here, where it's possible to "see between the cracks". Something similar goes on with `DiscreteForcing`, though our comment there is a bit more mild "Using discrete forcing functions may require understanding the staggered arrangement of velocity fields and tracers..."
> 
> https://clima.github.io/OceananigansDocumentation/stable/model_setup/forcing_functions/#%22Discrete-form%22-forcing-functions

I think I understand your point a bit better. I guess part of your argument is that this API is in place only because we haven't been able to implement something clearer/more user-friendly yet?

If so, how about the following as a warning?:

```suggestion
!!! warning "`ImmersedBoundaryCondition`"
    `ImmersedBoundaryCondition` is considered an experimental API and may change in the future.
    Therefore, one should use it only when a finer level of control over the boundary conditions
    at the immersed boundary is required, and the user is familiar with the implementation of boundary 
    conditions on staggered grids. For all other cases , using the `immersed` argument of
    `FieldBoundaryConditions` is preferred.
```

simone-silvestri: I like this version, is a bit more clear IMO

tomchor:
@simone-silvestri bump patch version and merge?

tomchor:
Actually now that I think about it, this is a breaking release. Idk how many people use this feature, but I wonder if we should do something similar to https://github.com/CliMA/Oceananigans.jl/pull/2990 and throw a warning whenever `ImmersedBoundaryCondition` is used so that previously working user scripts don't break silently.

glwagner:
I'd be in favor of always emitting a warning for `ImmersedBoundaryCondition` that it's experimental (no published papers, few validation tests)

src/ImmersedBoundaries/immersed_boundary_condition.jl
```diff
@@ -51,6 +51,7 @@ function ImmersedBoundaryCondition(; west = nothing,
                                      bottom = nothing,
                                      top = nothing)
 
+    @warn "`ImmersedBoundaryCondition` is considered an experimental API and may change in the future."
```
glwagner, line 54:
```suggestion
    @warn "`ImmersedBoundaryCondition` is experimental."
```

PR #3145, opened 2023-06-14
Adaptive (CFL-based) time step in the SplitExplicit solver

### LINKED ISSUE #2982: Calculate number of barotropic substeps on the fly using CFL criterion and current time-step
Currently the number of barotropic substeps have to be calculated in advance, and the user must manually calculate the CFL constraint. However, the SplitExplicitFreeSurface _specifies_ gravitational acceleration (and knows the grid); therefore the CFL condition can be calculated internally. Moreover it should be possible to ensure integer substeps on the fly, rather than precalculating.

I think a nice user interface would be something like

```julia
free_surface = SplitExplicitFreeSurface(gravitational_acceleration=9.81, barotropic_cfl=0.7)
```

(with both defaults).

Then the number of barotropic time-steps required to (1) meet the specified CFL constraint and (2) ensure integer number of substeps relative to a baroclinic substep would be calculated on the fly.

cc @simone-silvestri

simone-silvestri:
closes #2982

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
@@ -226,8 +226,6 @@ struct SplitExplicitSettings{𝒩, ℳ, 𝒯, 𝒮}
     substeps :: 𝒩
     "`averaging_weights`: (`Vector`)"
     averaging_weights :: ℳ
-    "`mass_flux_weights`: (`Vector`)"
-    mass_flux_weights :: ℳ
     "fractional step: (`Number`), the barotropic time step is `Δτ ⋅ Δt`" 
     Δτ :: 𝒯
     "time-stepping scheme"
@@ -247,6 +245,28 @@ end
 @inline cosine_averaging_kernel(τ::FT) where FT = τ >= 0.5 && τ <= 1.5 ? FT(1 + cos(2π * (τ - 1))) : zero(FT)
 
 @inline constant_averaging_kernel(τ) = 1
+
+struct AdaptiveSubsteps{B, F}
+    Δtᴮ :: B
+    barotropic_averaging_kernel :: F
+end
+
+AdaptiveSubsteps() = AdaptiveSubsteps(nothing, nothing)
+
+@inline function weights_from_substeps(substeps, barotropic_averaging_kernel)
+
+    τᶠ = range(0, 2, length = substeps+1)
+    Δτ = τᶠ[2] - τᶠ[1]
+
+    averaging_weights = FT.(barotropic_averaging_kernel.(τᶠ[2:end]))
+    idx = searchsortedlast(averaging_weights, 0, rev=true)
+    substeps = idx
+
+    averaging_weights = averaging_weights[1:idx]
+    averaging_weights ./= sum(averaging_weights)
+
+    return Δτ, averaging_weights
+end
 
 function SplitExplicitSettings(FT::DataType=Float64;
                                substeps = nothing, 
@@ -277,32 +297,19 @@ function SplitExplicitSettings(FT::DataType=Float64;
         wave_speed = sqrt(gravitational_acceleration * grid.Lz)
         
         Δtᴮ = cfl * Δs / wave_speed
+        if substeps isa AdaptiveSubsteps
+            return SplitExplicitSettings(AdaptiveSubsteps(Δtᴮ, barotropic_averaging_kernel), 
+                                         nothing, 
+                                         nothing, 
+                                         timestepper)
+        end
         substeps = ceil(Int, 2 * max_Δt / Δtᴮ)
     end
 
-    τᶠ = range(0, 2, length = substeps+1)
-    Δτ = τᶠ[2] - τᶠ[1]
-
-    averaging_weights = FT.(barotropic_averaging_kernel.(τᶠ[2:end]))
-    idx = searchsortedlast(averaging_weights, 0, rev=true)
-    substeps = idx
-
-    averaging_weights = averaging_weights[1:idx]
-    mass_flux_weights = similar(averaging_weights)
-
-    M = searchsortedfirst(τᶠ, 1) - 1
-
-    averaging_weights ./= sum(averaging_weights)
-
-    for i in substeps:-1:1
-        mass_flux_weights[i] = 1 / M * sum(averaging_weights[i:substeps]) 
-    end
-
-    mass_flux_weights ./= sum(mass_flux_weights)
+    Δτ, averaging_weights = weights_from_substeps(substeps, barotropic_averaging_kernel)
 
     return SplitExplicitSettings(substeps,
                                  averaging_weights,
-                                 mass_flux_weights,
                                  Δτ,
                                  timestepper)
 end
--- a/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
@@ -171,7 +171,7 @@ end
     end
 end
 
-function split_explicit_free_surface_substep!(η, state, auxiliary, settings, arch, grid, g, Δτ, substep_index)
+function split_explicit_free_surface_substep!(η, state, auxiliary, settings, weights, arch, grid, g, Δτ, substep_index)
     # unpack state quantities, parameters and forcing terms 
     U, V             = state.U,    state.V
     Uᵐ⁻¹, Uᵐ⁻²       = state.Uᵐ⁻¹, state.Uᵐ⁻²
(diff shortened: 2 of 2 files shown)


src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -245,9 +239,35 @@ struct ForwardBackwardScheme end
 end
 
 @inline cosine_averaging_kernel(τ::FT) where FT = τ >= 0.5 && τ <= 1.5 ? FT(1 + cos(2π * (τ - 1))) : zero(FT)
-
 @inline constant_averaging_kernel(τ) = 1
 
+struct AdaptiveSubsteps{B, F}
+    Δtᴮ :: B
+    averaging_kernel :: F
+end
```
glwagner, line 247: Do we also need something like `minimum_substeps`?

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -245,9 +239,35 @@ struct ForwardBackwardScheme end
...
+
+struct FixedSubsteps{B, F}
+    Δτ :: B
+    averaging_weights :: F
+end
```
glwagner, line 252: More verbose names might help here... I guess `Δtᴮ` is `Δt_barotropic` ? Not exactly sure what `Δτ` is but maybe its a fractional step size?

Another improvement might be `AdaptiveSplitExplicitSubsteps` and `FixedSplitExplicitSubsteps`.

However, in the "fixed" case, it's just the _number_ that's fixed right? Not the step size?vThat isn't clear from the name, but could maybe just be fixed by a docstring.

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -272,38 +292,21 @@ function SplitExplicitSettings(FT::DataType=Float64;
...
-
-    mass_flux_weights ./= sum(mass_flux_weights)
+    Δτ, averaging_weights = weights_from_substeps(FT, substeps, barotropic_averaging_kernel)
 
-    return SplitExplicitSettings(substeps,
-                                 averaging_weights,
-                                 mass_flux_weights,
-                                 Δτ,
+    return SplitExplicitSettings(FixedSubsteps(Δτ, averaging_weights),
                                  timestepper)
```
glwagner, line 310: The above can be

```julia
if substeps isa AdaptiveSubsteps
    substeps = AdaptiveSubsteps(Δtᴮ, barotropic_averaging_kernel)
else
    substeps = FixedSubsteps(Δτ, averaging_weights)
end

return SplitExplicitSettings(substeps, timestepper)
```

PS why do we have a distinct type for "Settings" ? This code is quite different from most of the other code; I think following the tyle of the rest of the code, `substeps` and `timestepper` would simply be properties of `SplitExplicitFreeSurface`. Not sure the advantage of the `Settings` type.

src/MultiRegion/multi_region_split_explicit_free_surface.jl
```diff
@@ -43,8 +43,9 @@ function FreeSurface(free_surface::SplitExplicitFreeSurface, velocities, grid::M
 
         switch_device!(grid.devices[1])
         old_halos = halo_size(getregion(grid, 1))
+        substeps  = length(settings.substeps.averaging_weights)
```
glwagner, line 46:
```suggestion
        Nsubsteps  = length(settings.substeps.averaging_weights)
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -319,10 +327,15 @@ function iterate_split_explicit!(free_surface, grid, Δt)
     settings  = free_surface.settings
     g         = free_surface.gravitational_acceleration
 
-    Δτ = settings.Δτ * Δt  # we evolve for two times the Δt 
+    preliminary_substeps = calculate_substeps(settings, Δt)
```
glwagner, line 330:
```suggestion
    N_initial_substeps = calculate_substeps(settings, Δt)
```

using "N" helps remind us that this is a number, like Nx or Nt.

src/Models/HydrostaticFreeSurfaceModels/distributed_split_explicit_free_surface.jl
```diff
@@ -64,8 +64,9 @@ function FreeSurface(free_surface::SplitExplicitFreeSurface, velocities, grid::D
         settings  = free_surface.settings 
 
         old_halos = halo_size(grid)
+        substeps  = length(settings.substeps.averaging_weights)
 
-        new_halos = split_explicit_halos(old_halos, settings.substeps+1, grid)         
+        new_halos = split_explicit_halos(old_halos, substeps+1, grid)
```
glwagner, line 69:
```suggestion
        new_halos = distributed_split_explicit_halos(old_halos, substeps+1, grid)         
```

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -245,9 +239,35 @@ struct ForwardBackwardScheme end
 end
 
 @inline cosine_averaging_kernel(τ::FT) where FT = τ >= 0.5 && τ <= 1.5 ? FT(1 + cos(2π * (τ - 1))) : zero(FT)
```
glwagner, line 241: is this type stable?

glwagner, line 241: is this called in a hot inner loop? It uses short-circuited logic but I don't think that's necessary here

when converting a number to some floating point, I think it's usually easier to read if you use

convert(FT, number)

rather than FT(long_expression).

I think the latter form should only be used for converting literals, as in `FT(1.5)`.

glwagner (approved):

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
simone-silvestri: we can probably hardcode that.

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
simone-silvestri: It's to allow initialization with different settings (aka `cfl` and `grid` rather than `substeps`).

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
simone-silvestri: It's for CPU use only so I guess it should be ok

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -276,12 +276,12 @@ function FixedTimeStepSize(FT::DataType = Float64;
...
 end
 
-@inline function weights_from_substeps(FT, substeps, barotropic_averaging_kernel)
+@inline function weights_from_substeps(FT, substeps, averaging_kernel)
 
     τᶠ = range(0, 2, length = substeps+1)
     Δτ = τᶠ[2] - τᶠ[1]
 
-    averaging_weights = FT.(barotropic_averaging_kernel.(τᶠ[2:end]))
+    averaging_weights = FT.(averaging_kernel.(τᶠ[2:end]))
```
glwagner, line 282: Can you write this in more lines?

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl
```diff
@@ -276,10 +276,10 @@ end
 
 @inline function weights_from_substeps(FT, substeps, averaging_kernel)
 
-    τᶠ = range(0, 2, length = substeps+1)
+    τᶠ = range(FT(0), FT(2), length = substeps+1)
     Δτ = τᶠ[2] - τᶠ[1]
 
-    averaging_weights = FT.(averaging_kernel.(τᶠ[2:end]))
+    averaging_weights = averaging_kernel.(τᶠ[2:end])
```
glwagner, line 282:
```suggestion
    averaging_weights = map(averaging_kernel, τᶠ[2:end])
```

glwagner: The `.` is a little too subtle and short, easy to miss

glwagner:
Great! Now we can make the split-explicit default when we're ready for that

navidcy:
I'm ready for that. :) 

Shall I open a PR?

PR #3151, opened 2023-06-21
More doc built optimizations -- run computationally expensive examples first

navidcy:
with @glwagner

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,11 +1,3 @@
-pushfirst!(LOAD_PATH, joinpath(@__DIR__, "..")) # add Oceananigans to environment stack
-
-using Pkg
-Pkg.activate(joinpath(@__DIR__, ".."))
-Pkg.instantiate()
-Pkg.activate(@__DIR__)
-Pkg.instantiate()
-
 using Distributed
 
 Distributed.addprocs(2)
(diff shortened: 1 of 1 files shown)


navidcy:
OK, ready for review!

glwagner (approved):
did it help?

glwagner:
The docs build in 52 minutes, which is the longest individual job but not much longer than some others. And better than before?

navidcy:
before it was 52 or 56min... so perhaps it helped slightly but also sometimes depending on how many thing tartarus is doing at the moment you can have some variability in times.

navidcy:
(when I say "before" I mean: before this PR but after #3135.)

glwagner:
Ok so no help basically. But maybe this is still a better way to do things.

navidcy:
But it's helpful for future. We now have a way to see how much time each example needs to run so we can order them accordingly!

glwagner:
beautiful

navidcy:
I just noticed:

https://buildkite.com/clima/oceananigans/builds/11955#018909c3-4c60-4b9c-b4e2-c6260d8b2189/40-2662

```
[ Info: shallow_water_Bickley_jet.jl example took 28.452 minutes to build.
```

We should optimize that or make it run first!

Issue #3147, opened 2023-06-16
Highlighting important breaking changes in releases


tomchor:
Not sure if this is the best place to post this, but I was thinking it would be good if, every time there is a breaking release (or a release that's particularly significant even if it's non-breaking), there is a small text in the release indicating the important/breaking features. (Or maybe even something on slack?)

This is something that has implicitly come up in a few conversations I've had with Oceananigans users that don't also do development (and thus don't keep track of PRs, etc.). The advantages that I can see are:

- Users can have an idea if their code will break with updating, which might prompt them to update more often; (I think in practice lots of users don't update their code regularly since there's a looming possibility of their code to stop running and them having to deal with errors.)
- If their code will break, users will have an idea as to what they should change in order to fix it (as opposed to running it and fixing the code error-by-error).

The downside is that it requires someone to manually write something in most releases. But I think this is easy enough that it shouldn't be a problem.

PS: Btw I should say, I mentioned non-developer users above, but I very much include myself in the group that would benefit from this. There's often a lot going on in Oceananigans and it's easy to get overwhelmed with research and lose track of important PRs. For example v0.83.0 just dropped, implying a breaking change, and I cannot figure out which was the breaking change.

Curious to hear everyone's thoughts.

simone-silvestri:
I agree. Perhaps we could link releases that introduce breaking changes to a specific PR that's dedicated to updating the version and detailing the API breaking change. This might be tedious work (for us) because it involves tracking down all API changes in past PRs, but it might be very beneficial for users

glwagner:
I agree with this --- though this something similar is already our policy. PRs that have breaking changes are supposed to update the minor version, and also describe the PR in words in the top comment, which certainly includes describing a breaking change. It's true that we don't achieve this, however. We could be stricter in PR reviews?

To make it easier to find the specific PRs that has the change which updates the minor version within release notes, we could have a policy to put the new version in the name of the PR, ie preface the PR with

```
(0.83.0) Title of the PR
```

That way one can easily find the relevant PR if there is a long list of PRs within a particular release, only one of which is important. (Note this information can also be obtained via `git blame` on `Project.toml`.)

My only concern is that we struggle to enforce conventions already, like style conventions in source code, branch naming conventions, etc.

[...]

PR #3153, opened 2023-06-23
(0.84.0) Fix `NetCDFOutputWriter` compression kwarg

Yixiao-Zhang:
To use compression in `NCDatasets` , the keyword for `defVar` should be `deflatelevel` instead of `compression`. `defVar` does not show any warning message for unrecognized keywords.

### DIFF
--- a/src/OutputWriters/netcdf_output_writer.jl
+++ b/src/OutputWriters/netcdf_output_writer.jl
@@ -210,7 +210,7 @@ Keyword arguments
                         for more information about its `mode` option.
 
 - `deflatelevel`: Determines the NetCDF compression level of data (integer 0-9; 0 (default) means no compression
-                  and 9 means maximum compression). See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#Creating-a-variable).
+                  and 9 means maximum compression). See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#Creating-a-variable)
                   for more information.
 
 ## Miscellaneous keywords
--- a/src/OutputWriters/netcdf_output_writer.jl
+++ b/src/OutputWriters/netcdf_output_writer.jl
@@ -210,7 +210,7 @@ Keyword arguments
                         for more information about its `mode` option.
 
 - `deflatelevel`: Determines the NetCDF compression level of data (integer 0-9; 0 (default) means no compression
-                  and 9 means maximum compression). See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#Creating-a-variable)
+                  and 9 means maximum compression). See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#Creating-a-variable).
                   for more information.
 
 ## Miscellaneous keywords
--- a/src/OutputWriters/netcdf_output_writer.jl
+++ b/src/OutputWriters/netcdf_output_writer.jl
@@ -395,7 +395,7 @@ function NetCDFOutputWriter(model, outputs; filename, schedule,
     if mode == "c"
         for (dim_name, dim_array) in dims
             defVar(dataset, dim_name, array_type(dim_array), (dim_name,),
-                   compression=compression, attrib=default_dimension_attributes[dim_name])
+                   deflatelevel=compression, attrib=default_dimension_attributes[dim_name])
         end
 
         # DateTime and TimeDate are both <: AbstractTime
@@ -446,7 +446,7 @@ function define_output_variable!(dataset, output, name, array_type, compression,
     name ∉ keys(dimensions) && error("Custom output $name needs dimensions!")
 
     defVar(dataset, name, eltype(array_type), (dimensions[name]..., "time"),
-           compression=compression, attrib=output_attributes)
+           deflatelevel=compression, attrib=output_attributes)
 
     return nothing
 end
@@ -456,7 +456,7 @@ end
 define_output_variable!(dataset, output::AbstractField, name, array_type, compression, output_attributes, dimensions) =
     defVar(dataset, name, eltype(array_type),
            (netcdf_spatial_dimensions(output)..., "time"),
-           compression=compression, attrib=output_attributes)
+           deflatelevel=compression, attrib=output_attributes)
 
 """ Defines empty field variable for `WindowedTimeAverage`s over fields. """
 define_output_variable!(dataset, output::WindowedTimeAverage{<:AbstractField}, args...) =
@@ -573,7 +573,7 @@ function define_output_variable!(dataset, output::LagrangianParticles, name, arr
     particle_fields = eltype(output.properties) |> fieldnames .|> string
     for particle_field in particle_fields
         defVar(dataset, particle_field, eltype(array_type),
-               ("particle_id", "time"), compression=compression)
+               ("particle_id", "time"), deflatelevel=compression)
     end
 end
 
--- a/src/OutputWriters/netcdf_output_writer.jl
+++ b/src/OutputWriters/netcdf_output_writer.jl
@@ -164,7 +164,7 @@ end
                             output_attributes = Dict(),
                                    dimensions = Dict(),
                            overwrite_existing = false,
-                                  compression = 0,
+                                 deflatelevel = 0,
                                       verbose = false)
 
 Construct a `NetCDFOutputWriter` that writes `(label, output)` pairs in `outputs` (which should
@@ -209,7 +209,9 @@ Keyword arguments
                         Default: `false`. See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/)
                         for more information about its `mode` option.
 
-- `compression`: Determines the compression level of data (0-9; default: 0).
+- `deflatelevel`: Determines the NetCDF compression level of data (integer 0-9; 0 (default) means no compression
+                  and 9 means maximum compression). See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#Creating-a-variable)
+                  for more information.
 
 ## Miscellaneous keywords
 
@@ -334,7 +336,7 @@ function NetCDFOutputWriter(model, outputs; filename, schedule,
                             output_attributes = Dict(),
                                    dimensions = Dict(),
                            overwrite_existing = nothing,
-                                  compression = 0,
+                                 deflatelevel = 0,
                                       verbose = false)
 
     mkpath(dir)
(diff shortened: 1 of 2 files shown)


tomchor (approved):
Nice catch!
Do you have merge permission?

navidcy:
I'm wondering if we should expend the remark on the `NetCDFOutputWriter` docstring regarding what `compression` is?

glwagner:
Should we also use `deflatelevel` for the name of the `NetCDFOutputWriter` kwarg?

It'd be nice to have a test for this I think.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -395,7 +395,7 @@ function NetCDFOutputWriter(model, outputs; filename, schedule,
     if mode == "c"
         for (dim_name, dim_array) in dims
             defVar(dataset, dim_name, array_type(dim_array), (dim_name,),
-                   compression=compression, attrib=default_dimension_attributes[dim_name])
+                   deflatelevel=compression, attrib=default_dimension_attributes[dim_name])
```
glwagner, line 398: What is the motivation for using the name `compression` in the user interface for `NetCDFOutputWriter`, but assigning it to `deflatelevel`? Is this standard / intuitive? Or should we also change `compression` to `deflatelevel`.

navidcy: I believe we should change compression to `deflatelevel` and in the docstring refer to NCDatasets.jl docs

navidcy: Although `compression` is more intuitive to `deflatelevel`....

tomchor: IMO compression is way more intuitive than `deflatelevel`, so I vote for leaving it the way it is. Does JLD2 have a similar option? What keyword do they use?

navidcy: I don't think the JLD2OutputWriter has this option, but looking at the JLD2 package I found:

https://juliaio.github.io/JLD2.jl/dev/compression/

We could use `compress::Boolean = true` as a keyword for both output writers and for the NetCDFOutputWriter also have a `deflatelevel` kwarg with some default value which will be passed if `compress = true`?

navidcy: Hm...

```
Note that CodecZlib.jl must be explicitly installed for compression/decompression to work. If the package is not already loaded JLD2 will try to dynamically load it, but it is recommended to do it explicitly, e.g. using JLD2, CodecZlib.
```

OK, so not sure if we want to deal with the JLD2OutputWriter in this PR, but in anticipation of potentially enabling compression for JLD2OutputWriter I still vote for a `compress::Bool` kwarg and then an additional kwarg with a default value, e.g., `deflatelevel::Int=4`.

tomchor: I'm okay with this solution

glwagner: I think we should pass kwargs to NCDatasets without changing their name.

navidcy: OK, this is a very good point!

navidcy, line 398:
```suggestion
                   deflatelevel=deflatelevel, attrib=default_dimension_attributes[dim_name])
```

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -573,7 +573,7 @@ function define_output_variable!(dataset, output::LagrangianParticles, name, arr
     particle_fields = eltype(output.properties) |> fieldnames .|> string
     for particle_field in particle_fields
         defVar(dataset, particle_field, eltype(array_type),
-               ("particle_id", "time"), compression=compression)
+               ("particle_id", "time"), deflatelevel=compression)
```
navidcy, line 576:
```suggestion
               ("particle_id", "time"), deflatelevel=deflatelevel)
```

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -456,7 +456,7 @@ end
 define_output_variable!(dataset, output::AbstractField, name, array_type, compression, output_attributes, dimensions) =
     defVar(dataset, name, eltype(array_type),
            (netcdf_spatial_dimensions(output)..., "time"),
-           compression=compression, attrib=output_attributes)
+           deflatelevel=compression, attrib=output_attributes)
```
navidcy, line 459:
```suggestion
           deflatelevel=deflatelevel, attrib=output_attributes)
```

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -446,7 +446,7 @@ function define_output_variable!(dataset, output, name, array_type, compression,
     name ∉ keys(dimensions) && error("Custom output $name needs dimensions!")
 
     defVar(dataset, name, eltype(array_type), (dimensions[name]..., "time"),
-           compression=compression, attrib=output_attributes)
+           deflatelevel=compression, attrib=output_attributes)
```
navidcy, line 449:
```suggestion
           deflatelevel=deflatelevel, attrib=output_attributes)
```

navidcy:
@Yixiao-Zhang I made some suggestions but I couldn't make suggestions to everywhere so I also pushed some changes. Hope that this is alright!

navidcy (changes requested):
As per the discussion in the PR, let's change the kwarg from `compression` to `deflatelevel`.

This is a breaking API change so @Yixiao-Zhang can you also bump up a minor release? That is, 
change the version number here:
https://github.com/Yixiao-Zhang/Oceananigans.jl/blob/36961523d3b35bfb08e9af779504b99beb60d585/Project.toml#L4
to 0.84.0 ?

Yixiao-Zhang:
BTW, in the next release of `NCDatasets.jl`,`defVar` will throw an error if there is an unrecognized keyword \(See this [change](https://github.com/Alexander-Barth/NCDatasets.jl/issues/212)\).

That means that older versions of `Oceananigans.jl` \(\< 0.84\) will not work with newer versions of `NCDatasets.jl` \(\>0.12.17\).

Also for this reason, we want to release `Oceananigans.jl` v0.84 before the release of a newer version of  `NCDatasets.jl` \(v0.12.18 or 0.13.0\).

navidcy:
> BTW, in the next release of `NCDatasets.jl`,`defVar` will throw an error if there is an unrecognized keyword (See this [change](https://github.com/Alexander-Barth/NCDatasets.jl/issues/212)).
> 
> That means that older versions of `Oceananigans.jl` (< 0.84) will not work with newer versions of `NCDatasets.jl` (>0.12.17).

Sounds that NCDatasets is gonna have a breaking change but _not_ a minor release?

Yixiao-Zhang:
I do not mean that. I think it should be a major change.

When I wrote down that sentence, I did not realize that `compat` in `Project.toml` can prevent using newer versions of dependencies with breaking changes.

glwagner:
> BTW, in the next release of `NCDatasets.jl`,`defVar` will throw an error if there is an unrecognized keyword (See this [change](https://github.com/Alexander-Barth/NCDatasets.jl/issues/212)).
> 
> That means that older versions of `Oceananigans.jl` (< 0.84) will not work with newer versions of `NCDatasets.jl` (>0.12.17).

Helpful to know that we will need that compat, thank you!

glwagner:
> BTW, in the next release of `NCDatasets.jl`,`defVar` will throw an error if there is an unrecognized keyword (See this [change](https://github.com/Alexander-Barth/NCDatasets.jl/issues/212)).
> 
> That means that older versions of `Oceananigans.jl` (< 0.84) will not work with newer versions of `NCDatasets.jl` (>0.12.17).

That's good to know, thank you!

glwagner:
> As per the discussion in the PR, let's change the kwarg from `compression` to `deflatelevel`.
> 
> This is a breaking API change so @Yixiao-Zhang can you also bump up a minor release? That is, change the version number here: https://github.com/Yixiao-Zhang/Oceananigans.jl/blob/36961523d3b35bfb08e9af779504b99beb60d585/Project.toml#L4 to 0.84.0 ?

Also the title of the PR should be prepended with the new version ie `(0.84.0) Fix NetCDF...`

navidcy (approved):

navidcy:
OK, I'm merging this?

navidcy:
@Yixiao-Zhang ready to merge?

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -209,7 +209,9 @@ Keyword arguments
                         Default: `false`. See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/)
                         for more information about its `mode` option.
 
-- `compression`: Determines the compression level of data (0-9; default: 0).
+- `deflatelevel`: Determines the NetCDF compression level of data (integer 0-9; 0 (default) means no compression
+                  and 9 means maximum compression). See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#Creating-a-variable)
```
navidcy, line 213:
```suggestion
                  and 9 means maximum compression). See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#Creating-a-variable).
```

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -209,7 +209,9 @@ Keyword arguments
                         Default: `false`. See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/)
                         for more information about its `mode` option.
 
-- `compression`: Determines the compression level of data (0-9; default: 0).
+- `deflatelevel`: Determines the NetCDF compression level of data (integer 0-9; 0 (default) means no compression
+                  and 9 means maximum compression). See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#Creating-a-variable).
```
navidcy, line 213:
```suggestion
                  and 9 means maximum compression). See [NCDatasets.jl documentation](https://alexander-barth.github.io/NCDatasets.jl/stable/variables/#Creating-a-variable)
```

Yixiao-Zhang:
I think it is ready to merge.

glwagner:
This PR: https://github.com/CliMA/Oceananigans.jl/pull/3091 also claims to be (0.84.0). Which is right? 

@tomchor I'm going to need help enforcing the new conventions that were discussed on #3147

navidcy:
Both introduced breaking changes! And were merged the same day so we registered v0.84 after both were merged.

navidcy:
I thought that it wouldn’t make sense to register v0.84 and 20min later v0.85….

glwagner:
Ah ok

PR #3161, opened 2023-06-29
Fix issue with GL/CairoMakie not being able to compile on Apple's M1/2 chips

navidcy:
The current Manifest conflicts with FFMPEG v0.4 and thus enforces it to fallback to FFMPEG v0.2.
This, in principle, is not a problem since we don't depend on FFMPEG. But as soon as we try to use Makie/GLMakie which has FFMPEG as a dependency then it is forced to use FFMPEG v0.2.

FFMPEG v0.2 cannot compile on Apple’s M1/M2 chips and this creates errors and inconvenience. 

I tried to enforce installing `FFMPEG@v0.4` on the current environment and this resulted in downgrade of some `jll`s.

```
⌅ [0234f1f7] ↓ HDF5_jll v1.14.0+0 ⇒ v1.12.2+2
⌃ [7243133f] ↓ NetCDF_jll v400.902.208+0 ⇒ v400.902.5+1
⌅ [458c3c95] ↓ OpenSSL_jll v3.0.9+0 ⇒ v1.1.21+0
```

Informed by that, this PR ensures that those `jll`s satisfy `FFMPEG`s dep requirements so that people with Apple M1/2 chips (like myself) can still use Oceananigans + Makie for plotting.

I've been troubling with this all morning along with @glwagner and @siddharthabishnu. Not sure if this is the best solution possible, but it does work. Hopefully this is something that it will be fixed from other packages moving forward.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -6,9 +6,9 @@ project_hash = "c19e7e0747cdfab35ec6d8d8bd4d66290cc1b731"
 
 [[deps.AbstractFFTs]]
 deps = ["ChainRulesCore", "LinearAlgebra"]
-git-tree-sha1 = "16b6dbc4cf7caee4e1e75c49485ec67b667098a0"
+git-tree-sha1 = "8bc0aaec0ca548eb6cf5f0d7d16351650c1ee956"
 uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
-version = "1.3.1"
+version = "1.3.2"
 
 [[deps.Adapt]]
 deps = ["LinearAlgebra", "Requires"]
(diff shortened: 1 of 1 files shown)


siddharthabishnu (approved):

siddharthabishnu:
The issue has been fixed successfully by @navidcy! :-D

glwagner:
Note that this could be considered bad practice (ie we ran into similar issues with julia 1.7 + MPI on apple silicon). I'm ok with it as a bandaid though. We will have to update the jll in the future.

navidcy:
I know. But we need to plot! I think the issue goes away w Julia v1.9. I updated all deps no restrictions and don’t think there was an issue w Makie.

glwagner:
True but users can always fix the problematic jll themselves. This is just a convenience?

navidcy:
Yes. It is! Agreed

jagoosw:
Could we have a patch release soon so that this fix gets out?

navidcy:
do you get the same issue? you can forcefully downgrade the above `jll` files.

navidcy:
I don't know If this is an issue with the latest tagged release. is it?

jagoosw:
Yeah, I get this issue too with the latest tag and 0.83.0.

> do you get the same issue? you can forcefully downgrade the above `jll` files.

How do I do this?

glwagner:
Add the `_jll` package to your project and pin it to the desired version

PR #3167, opened 2023-07-01
Restore netCDF output writer tests

### LINKED ISSUE #2956: Running simulations with `Clock{DateTime}`
This function lives in the tests but is never used in the tests, right?

https://github.com/CliMA/Oceananigans.jl/blob/acb298b7f7447747e9f4bcb1ffbbfbe09d9f3d98/test/test_netcdf_output_writer.jl#L14-L46

In fact the whole of this file.

I remember some issues with NetCDF at some point on linux machines -- was this why this test was removed?

navidcy:
This PR restores some test on netCDF output.

(The tests were removed, if I recall correctly, because at some point netCDF was creating issues on unix machines. Things seems fine now!)

Closes #2956

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -67,6 +67,7 @@ CUDA.allowscalar() do
             include("test_simulations.jl")
             include("test_diagnostics.jl")
             include("test_output_writers.jl")
+            include("test_netcdf_output_writer.jl")
             include("test_output_readers.jl")
         end
     end


tomchor (approved):
I didn't know these were removed! Glad to see they're back.

I think keeping NetCDF up and running is important since I think a big chunk of users (myself included) do lots of post-processing outside of Julia. You can open JLD2 outside of Julia, but NetCDF is much easier.

PR #3168, opened 2023-07-02
More docs built optimization

navidcy:
The shallow water example needs ~25 min to run while all other examples need ~1min (2D) or up to ~7min (Baroclinic instability).

This PR forces the workers that run the examples for the docs to start from the shallow water example. This way we minimize the time that the one worker ends up running the swe example while the other one sits and waits.

This PR is related to #3169 but it does _not_ resolve the issue.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -43,6 +43,7 @@ end
     # The examples that take longer to run should be first. This ensures thats
     # docs built using extra workers is as efficient as possible.
     example_scripts = [
+        "shallow_water_Bickley_jet.jl",
         "baroclinic_adjustment.jl",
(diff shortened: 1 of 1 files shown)


navidcy:
Just with this reordering we get the docs built in 40min vs 52min e.g. in https://buildkite.com/clima/oceananigans/builds/11973#01891139-cbb4-4cc1-a2b5-25769f4e9978

tomchor (approved):
Wow, docs building in 40 min from like 3 hours a couple of weeks ago. That's crazy...

:tada:

navidcy:
Yes. It is!  And this example should be optimized. But until then I’m merging this so the developers have a slightly easier life ;)

PR #3170, opened 2023-07-02
Some simplifications in Docs/Simulations For GPUs

navidcy:


docs/src/simulation_tips.md
```diff
@@ -86,12 +86,13 @@ fixes the issue by indicating to the compiler that `T₀` will not change.
 Note that the _literal_ `2π / 86400` is not an issue -- it's only the
 _variable_ `T₀` that must be declared `const`.
 
-Alternatively, passing the variable as a parameter to `GradientBoundaryCondition` also works:
+Alternatively, passing the variable as a parameter to `GradientBoundaryCondition`. To do that
```
tomchor, line 89:
```suggestion
Alternatively, we can pass the variable as a parameter to `GradientBoundaryCondition`. To do that
```

docs/src/simulation_tips.md
```diff
@@ -188,8 +189,7 @@ For large simulations on the GPU, careful management of memory allocation may be
   ```
 
 - Try to use higher-order advection schemes. In general when you use a higher-order scheme you need
-  fewer grid points to achieve the same accuracy that you would with a lower-order one. Oceananigans
-  provides two high-order advection schemes: 5th-order WENO method (WENO) and 3rd-order upwind.
+  fewer grid points to achieve the same accuracy that you would with a lower-order one.
```
tomchor, line 192: Do we have a page discussing the advection schemes that are available? If so, we should probably link it here

navidcy: I don’t think we do. There is this —> https://clima.github.io/OceananigansDocumentation/stable/appendix/library/#Advection

navidcy: I added a reference to the above. It's better than nothing atm.

tomchor (approved):

PR #3173, opened 2023-07-06
Add `update_biogeochemical_state!` to `update_state!` for hydrostatic free surface single column mode

jagoosw:
Add `update_biogeochemical_state!` to `update_state!` for hydrostatic free surface single column mode because it's missing and therefore bgc doesn't work.


navidcy:
1) did we merge this before the tests passed?

2) how OceanBioME.jl was working without this?

3) I think this definitely deserves a patch release?

glwagner:
this only affects single column models

glwagner:
@jagoosw perhaps best practice is to add a test to make sure this doesn't break in the future

navidcy:
Yeap! That'd be good addition.

jagoosw:
Yeah that's a good idea, do you think I should add testing on lat/lon grids too?

glwagner:
yes

PR #3180, opened 2023-07-11
Make shallow water Bickley jet less expensive

### LINKED ISSUE #3169: `ShallowWaterModel` example in docs takes much longer to run, even when compared to 3D examples
In #3151 it was discovered that the shallow water example in the docs needs ~25 min to run while all other examples need from ~1min (2D) or up to ~7min (Baroclinic instability).

Was this always the case? Did something happen to impede the performance of the SWE model?

#3168 is relevant; it does not resolve the issue but nevertheless minimizes the toll that this issue has on the overall building time for the docs.

glwagner:
Closes #3169

### DIFF
--- a/examples/shallow_water_Bickley_jet.jl
+++ b/examples/shallow_water_Bickley_jet.jl
@@ -224,7 +224,7 @@ nothing # hide
 
 using Polynomials: fit
 
-I = 6000:7000
+I = 5000:6000
 
 degree = 1
 linear_fit_polynomial = fit(t[I], log.(norm_v[I]), degree, var = :t)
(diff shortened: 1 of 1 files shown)


glwagner:
After this PR, the shallow water Bickley jet takes 10 min to build: 

https://buildkite.com/clima/oceananigans/builds/12054#01894501-3758-4104-b295-b9ea601ffe6f/19-1630

looks like it takes about 20 min on `main`:

https://buildkite.com/clima/oceananigans/builds/12046#018942b3-7263-4631-bc50-98bf45c51932/17-2868

So this PR constitutes an improvement.

navidcy:
Did you run locally? How does the last figure looks like? (The one with the growth rate calculation.) If it's all good then happy to approve!

navidcy:
Note: It seems that it didn't really reduce the docs built overall time. I guess we've reached to the point where examples aren't the bottleneck anymore but rather doctests and compilation times are. :)

navidcy (approved):

PR #3182, opened 2023-07-14
Update Julia version warning

navidcy:
After discussion in #3177 I noticed that the warning only warns for Julia v1.7 and earlier... But it should warn for Julia v1.8 as well.

(Although, if we decide to add a compat entry for Julia v1.9 as a response to #3184 then the warning is rendered redundant.)


navidcy:
closing this as we are moving fwd with a compat entry that will render this message useless.

PR #3183, opened 2023-07-14
Warning for CFL > 0.8

simone-silvestri:


navidcy (approved):

glwagner:
This is an ok idea but the warning really needs to be better. It isn't clear who is "recommending" what; also it doesn't really make sense because the CFL depends on the time-step and in principle, someone can use `TimeStepWizard` with any time stepper (for example there are models external to Oceananigans that use the `Simulation`). Presumably 0.8 is some rule of thumb for RK3.

glwagner (changes requested):
This needs to be thought out better.

navidcy:
OK, let's close it for now?

glwagner:
The problem we want to solve is how to inform users about time-step restrictions associated with various time-steppers, which is noble...

glwagner:
What about designing an interface that's model-specific (since that's where the time-stepper lives). Something like

```julia
validate_time_step_size(model, dt)
```

Then we can implement

```julia
function change_simulation_dt!(simulation, new_dt)
    validate_time_step_size(simulation.model, dt)
    simulation.dt = dt
    return nothing
end
```

Now, model developers have the option to implement this function (and the fallback would be `nothing`).

glwagner:
Ok better idea. The _first_ time that time-step wizard is called as a callback, it can call into a function like `validate_CFL(model, cfl)`. 

Then model developers can implement that function if they would like to warn users. Since it's called at iteration 0, the warning is prominent. Moreover, the warning can be made specific depending on the type of model and time-stepper being used, so we can be more sure that we are giving accurate and relevant information.

glwagner:
This can be done with a custom `initialize!` for the time-step wizard:

https://github.com/CliMA/Oceananigans.jl/blob/1888b413aa89d187195d1b6834b5bb2f6f4761d0/src/Simulations/callback.jl#L18

glwagner:
Happy to tackle this is @simone-silvestri likes

simone-silvestri:
Sure go ahead! It would be nice to have a warning, if we can connect it to the timestepper used we can be more descriptive.

simone-silvestri:
Sure go ahead! It would be nice to have a warning, if we can connect it to the timestepper used we can be more descriptive.

glwagner:
I'm closing this in favor of #3183, open again if you disagree

PR #3188, opened 2023-07-17
Add `ImmersedPressureSolver` with `FFTBasedPoissonSolver` as preconditioner for complex domains

xkykai:
Using the `FFTBasedPoissonSolver` in complex domains on the `ImmersedBoundaryGrid` leads to resulting solution that either

1. Does not conserve tracers in the flow; or
2. Leads to divergent flow along the immersed boundary.

On the other hand, using the `PreconditionedConjugateGradientSolver` on complex domains leads to exact solution at a much slower speed on the GPU. This PR implements an `ImmersedPressureSolver` that uses the `FFTBasedPoissonSolver` as a preconditioner, then iterates `PreconditionedConjugateGradientSolver` to arrive at the correct solution. This approach is much faster than using a vanilla `PreconditionedConjugateGradientSolver` as the number of iterations scale much more slowly with grid size on the GPU.

This PR will include the `ImmersedPressureSolver`, a few validation examples for benchmarking purposes and a number of example use cases where this solver would be useful.

@glwagner @simone-silvestri


src/Models/NonhydrostaticModels/immersed_pressure_solver.jl
```diff
@@ -0,0 +1,152 @@
...
+    mask_immersed_field!(rhs, zero(grid))
+
+    # Solve pressure Pressure equation for pressure, given rhs
+    # @info "Δt before pressure solve: $(Δt)"
+    solve!(pressure, solver.pcg_solver, rhs)
+
+    return pressure
+end
+
+struct MITgcmPreconditioner end
```
navidcy, line 112: can we name this something bit more descriptive and less referential to MITgcm?

src/Models/NonhydrostaticModels/immersed_pressure_solver.jl
```diff
@@ -0,0 +1,152 @@
...
+    return P_r
+end
+
+# Kernels that calculate coefficients for the preconditioner
+@inline Ax⁻(i, j, k, grid) = @inbounds Axᶠᶜᶜ(i, j, k, grid) / Δxᶠᶜᶜ(i, j, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
+@inline Ay⁻(i, j, k, grid) = @inbounds Ayᶠᶜᶜ(i, j, k, grid) / Δyᶠᶜᶜ(i, j, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
+@inline Az⁻(i, j, k, grid) = @inbounds Azᶠᶜᶜ(i, j, k, grid) / Δzᶠᶜᶜ(i, j, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
+@inline Ax⁺(i, j, k, grid) = @inbounds Axᶠᶜᶜ(i+1, j, k, grid) / Δxᶠᶜᶜ(i+1, j, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
+@inline Ay⁺(i, j, k, grid) = @inbounds Ayᶠᶜᶜ(i, j+1, k, grid) / Δyᶠᶜᶜ(i, j+1, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
+@inline Az⁺(i, j, k, grid) = @inbounds Azᶠᶜᶜ(i, j, k+1, grid) / Δzᶠᶜᶜ(i, j, k+1, grid) / Vᶜᶜᶜ(i, j, k, grid)
```
navidcy, line 132:
```suggestion
@inline Ax⁻(i, j, k, grid) = Axᶠᶜᶜ(i, j, k, grid) / Δxᶠᶜᶜ(i, j, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
@inline Ay⁻(i, j, k, grid) = Ayᶠᶜᶜ(i, j, k, grid) / Δyᶠᶜᶜ(i, j, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
@inline Az⁻(i, j, k, grid) = Azᶠᶜᶜ(i, j, k, grid) / Δzᶠᶜᶜ(i, j, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
@inline Ax⁺(i, j, k, grid) = Axᶠᶜᶜ(i+1, j, k, grid) / Δxᶠᶜᶜ(i+1, j, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
@inline Ay⁺(i, j, k, grid) = Ayᶠᶜᶜ(i, j+1, k, grid) / Δyᶠᶜᶜ(i, j+1, k, grid) / Vᶜᶜᶜ(i, j, k, grid)
@inline Az⁺(i, j, k, grid) = Azᶠᶜᶜ(i, j, k+1, grid) / Δzᶠᶜᶜ(i, j, k+1, grid) / Vᶜᶜᶜ(i, j, k, grid)
```

navidcy: btw, why all these are computed at `fcc`, what's so special about `fcc`? I don't see that..

src/Models/NonhydrostaticModels/immersed_pressure_solver.jl
xkykai: Something like `SymmetricLocalPreconditioner`? I'm not entirely sure, could use some ideas too!

navidcy: Well I that sounds much better because at least it tries to describe something and not just refer to some other model.

I'm not sure what this preconditioner is doing so that's why I'm not in position to propose something. I don't even know what local refers to. But definitely a `SomethingThatDescribesMePreconditioner` is better than `SimilarToSomOtherModelOrSomebodyElsesPreconditioner`. That's what I wanted to convey with my comment.

src/Models/NonhydrostaticModels/immersed_pressure_solver.jl
xkykai: Good catch! Let me change it and test it, will update soon.

src/Models/NonhydrostaticModels/immersed_pressure_solver.jl
glwagner: I believe this preconditioner is some kind of perturbation about a diagonal preconditioner, is that right? `DiagonallyDominantPreconditioner` or something. In fact shoudln't we also delete the wrong version that's already present @simone-silvestri ?

simone-silvestri: Exactly, it should be the same one but corrected to be 3D and with the correct denominator (we should correct also the 2D version of it)

glwagner: Ok so maybe `DiagonallyDominantThreeDimensionalPreconditioner` and also `TwoDimensional`.

glwagner:
> I'm not sure what the state of this PR is, but last I checked it was working, but was about 5 times slower than the FFT-based solver, no?
> 
> Coming from the discussion in #3529, I'd argue that it'd be useful to include this solver in the main code, even if it's a bit slow for now. Then it'd be easy to check the influence of near-immersed-boundary divergence when investigating results such as #3529.
> 
> Furthermore, I can see situations where I'd be willing to spend 5 times longer running simulations in order to be more sure of the results. Oceananigans is can be so fast on GPUs that a 5x slower code might still be as good as a more traditional CPU code running the same problem 😂
> 
> What do you guys think?

There's no reason to wait on merging this as far as I'm concerned. @xkykai and @simone-silvestri can perhaps comment on their strategy / plan.

glwagner:
@xkykai are you ok if I send commits here?

glwagner:
> Furthermore, I can see situations where I'd be willing to spend 5 times longer running simulations in order to be more sure of the results. Oceananigans can be so fast on GPUs that a 5x slower code might still be as good as a more traditional CPU code running the same problem 😂

I agree with this. I also want to point out that if you want to make an apples-to-apples comparison, you need to compare this solver with a CPU code for _complex_ domains. So few people use those that we might not have a great idea what they cost, but I suspect they are quite a bit slower than the typical code for periodic boxes.

glwagner:
@xkykai I propose the following to move forward, let me know what you think:

1. Change the names a bit because technically, ImmersedPoissonSolver can be used as a general Poisson solver that works even on a lat lon grid. I think just `ConjugateGradientPoissonSolver` is good.
2. Make it default to an FFT-based preconditioner when possible, and otherwise to the "diagonally dominant preconditioner"
3. Since I feel we don't want to drop this just yet, let's not make it the default yet. However, I will put a note in the warning that currently accompanies NonhydrostaticModel on ImmersedBoundaryGrid about the possibility of using it.

glwagner:
It keeps tripping me up that NonhydrostaticModel calls it a "pressure solver" but everywhere else we use the term "Poisson solver". Should we call it `NonhydrostaticModel.poisson_solver`?

xkykai:
I agree with all 3 points. One thing that I think would be useful is to come up with one example to put in the docs about the choices of preconditioners and how to use it.  We can perhaps use a simple example of a staircase convection to illustrate the point, advertising the FFT-preconditioned solver at the same time as well. Currently there's no section about solver choices, and I suspect some users (me included) might not have been aware that different solver choices could be made.

tomchor:
> It keeps tripping me up that NonhydrostaticModel calls it a "pressure solver" but everywhere else we use the term "Poisson solver". Should we call it `NonhydrostaticModel.poisson_solver`?

I think it makes sense to use `pressure_solver` for the `NonhydrostaticModel` since that's literally what it's used for. The pressure "problem" just happens to also be a Poisson equation.

glwagner:
> > It keeps tripping me up that NonhydrostaticModel calls it a "pressure solver" but everywhere else we use the term "Poisson solver". Should we call it `NonhydrostaticModel.poisson_solver`?
> 
> I think it makes sense to use `pressure_solver` for the `NonhydrostaticModel` since that's literally what it's used for. The pressure "problem" just happens to also be a Poisson equation.

This statement merely repeats the original rationale for calling it pressure solver. What has to be addressed is the point about the names being easy to mix up. Perhaps you're dismissing that as unimportant, but you didn't say that explicitly.

glwagner:
> One thing that I think would be useful is to come up with one example to put in the docs about the choices of preconditioners and how to use it. We can perhaps use a simple example of a staircase convection to illustrate the point, advertising the FFT-preconditioned solver at the same time as well.

I like this idea though I am not sure it should be an example. There are no other examples like that. We are sort of in the process of revamping the docs to add more "tutorials". For example, this is a tutorial on grids: https://clima.github.io/OceananigansDocumentation/stable/grids/

I think what you're proposing is a "tutorial" on pressure solvers. What do you think?

I think we probably should wait though. It'd be nice to publish a paper about the solver first, and then put what we learned from the paper into the tutorial.

tomchor:
> > > It keeps tripping me up that NonhydrostaticModel calls it a "pressure solver" but everywhere else we use the term "Poisson solver". Should we call it `NonhydrostaticModel.poisson_solver`?
> > 
> > 
> > I think it makes sense to use `pressure_solver` for the `NonhydrostaticModel` since that's literally what it's used for. The pressure "problem" just happens to also be a Poisson equation.
> 
> This statement merely repeats the original rationale for calling it pressure solver. What has to be addressed is the point about the names being easy to mix up. Perhaps you're dismissing that as unimportant, but you didn't say that explicitly.

Okay... here's something more explicit: I think saying `pressure_solver = SpecificPoissonSolver()` makes sense since it's stating that a certain Poisson solver is  being used as a solver for the pressure equation. Personally, I don't think it's confusing. I also don't feel strongly either way though.

xkykai:
> I think what you're proposing is a "tutorial" on pressure solvers. What do you think?

Yes I think a tutorial is what I meant. We can do this after the paper is done so we encapsulate what we have learnt.

glwagner:
> Personally, I don't think it's confusing. I also don't feel strongly either way though.

Ah yes, totally agree that it isn't confusing. I wasn't arguing that it was confusing though. 

In terms of _reading_ these code statements I much prefer seeing `pressure_solver = poisson_solver`. Its even pedagogical because it teaches you that we solve the Poisson equation for pressure. I love that and that is exactly why we chose this design.

Please bear with me to understand the reason I am proposing this change. In practical usage / when coding new experiments it is easy to mix up "pressure" with "poisson". I've been working testing lots of different solvers these past few days, and keep accidentally typing `pressure_solver = ConjugateGradientPressureSolver(grid)` which is wrong. If it becomes more common to configure the pressure / poisson solver in the future (because we have better support for different solvers for irregular geometries) then I would like to also help others that may run into this issue. It's nice to have syntax that "rolls of the tongue", ie, is what you would type if you were typing without concentrating very hard.

glwagner (approved):

PR #3209, opened 2023-08-01
Toggle immersed differences with immersed_inactive_node

### LINKED ISSUE #3208: Immersed boundary conditions invalidate all non-immersed BCs?
I've been playing around with the `immersed_couette_flow.jl` validation experiment and think I discovered a bug in the immersed boundaries implementation which causes non-immersed boundary conditions to be ignored.

The BCs specified are here:
https://github.com/CliMA/Oceananigans.jl/blob/f7acd8d0bd30dbe1ccb72854b6ea0ccab1eae0b5/validation/immersed_boundaries/immersed_couette_flow.jl#L16-L23

It is clear from the movie produced by the script that the flow adjusts to the immersed value BCs of u(0.1)=1 and c(0.1)=1 but does not adjust to meet the top BCs of u(1)=-1 and c(1)=-1.

https://github.com/CliMA/Oceananigans.jl/assets/12971166/fb6ac2a8-7081-4edf-b363-87850b47fc6a

I ran into the same problem when adding an immersed boundary to the `horizontal_convection.jl` experiment. Adding an immersed boundary invalidated the surface buoyancy forcing, even when no `immersed` boundary condition was explicitly specified.

Has anyone run into this issue or know what might be going on?

glwagner:
This PR toggles conditional differences with `immersed_inactive_node`, rather than `inactive_node`. The difference is that `inactive_node` returns `true` if the node is outside the domain (either immersed _or_ outside the domain in a `Bounded` direction). `immersed_inactive_node` only returns true if the node is within the immersed boundary:

https://github.com/CliMA/Oceananigans.jl/blob/f7acd8d0bd30dbe1ccb72854b6ea0ccab1eae0b5/src/ImmersedBoundaries/ImmersedBoundaries.jl#L220-L221

This matters for applying `ValueBoundaryCondition` or `GradientBoundaryCondition` across non-immersed boundaries, on `ImmersedBoundaryGrid`, because both of these are enforced by filling the halo regions on the other side of a `Bounded` direction. Thus we have to be able to correctly evaluate differences / interpolate across non-immersed `Bounded` boundaries.

Closes #3208 

We might want to add a test so this doesn't break in the future. It'd also be great to make this code more understandable (suggestions welcome...)

@hdrake

### DIFF
--- a/src/ImmersedBoundaries/conditional_differences.jl
+++ b/src/ImmersedBoundaries/conditional_differences.jl
@@ -6,12 +6,29 @@ import Oceananigans.Operators:
     δzᶜᶜᶠ, δzᶜᶠᶠ, δzᶠᶜᶠ, δzᶠᶠᶠ,
     δzᶜᶜᶜ, δzᶜᶠᶜ, δzᶠᶜᶜ, δzᶠᶠᶜ
 
-@inline conditional_δx_f(LY, LZ, i, j, k, ibg::IBG, δx, args...) = ifelse(inactive_node(i, j, k, ibg, c, LY, LZ) | inactive_node(i-1, j, k, ibg, c, LY, LZ), zero(ibg), δx(i, j, k, ibg.underlying_grid, args...))
-@inline conditional_δx_c(LY, LZ, i, j, k, ibg::IBG, δx, args...) = ifelse(inactive_node(i, j, k, ibg, f, LY, LZ) | inactive_node(i+1, j, k, ibg, f, LY, LZ), zero(ibg), δx(i, j, k, ibg.underlying_grid, args...))
-@inline conditional_δy_f(LX, LZ, i, j, k, ibg::IBG, δy, args...) = ifelse(inactive_node(i, j, k, ibg, LX, c, LZ) | inactive_node(i, j-1, k, ibg, LX, c, LZ), zero(ibg), δy(i, j, k, ibg.underlying_grid, args...))
-@inline conditional_δy_c(LX, LZ, i, j, k, ibg::IBG, δy, args...) = ifelse(inactive_node(i, j, k, ibg, LX, f, LZ) | inactive_node(i, j+1, k, ibg, LX, f, LZ), zero(ibg), δy(i, j, k, ibg.underlying_grid, args...))
-@inline conditional_δz_f(LX, LY, i, j, k, ibg::IBG, δz, args...) = ifelse(inactive_node(i, j, k, ibg, LX, LY, c) | inactive_node(i, j, k-1, ibg, LX, LY, c), zero(ibg), δz(i, j, k, ibg.underlying_grid, args...))
-@inline conditional_δz_c(LX, LY, i, j, k, ibg::IBG, δz, args...) = ifelse(inactive_node(i, j, k, ibg, LX, LY, f) | inactive_node(i, j, k+1, ibg, LX, LY, f), zero(ibg), δz(i, j, k, ibg.underlying_grid, args...))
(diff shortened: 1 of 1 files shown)


src/ImmersedBoundaries/conditional_differences.jl
```diff
@@ -6,20 +6,55 @@ import Oceananigans.Operators:
...
+# Here we return `zero(ibg)` rather than `δx` (for example) when _one_ of the
+# nodes involved in the difference is `immersed_inactive_node`.
+#
+# These functions are used to generate all conditioned difference operators via metaprogramming.
+# There are 24 difference operators in all:
+# 1 operator for each of 8 locations (ᶠᶜᶜ, ᶠᶜᶠ, ...) × 3 directions (δx, δy, δz)).
+#
+# The operators depend on the location they end up at. For example, conditional_δx_f is used
+# to construct `δxᶠᶜᶜ`, `δxᶠᶜᶠ`, `δxᶠᶠᶜ`, `δxᶠᶠᶠ`, all of which difference values `Center`ed in `x`
+# at `Face`.
```
hdrake, line 19: This comment is helpful.

hdrake (commented):
Looks good to me. Both the `immersed_couette_flow.jl` example and `horizontal_convection.jl` with topography seem to be working as intended.

https://github.com/CliMA/Oceananigans.jl/assets/12971166/8ffc04c5-34ce-46b9-8ea3-fc79b39abdd5

glwagner:
> Looks good to me. Both the `immersed_couette_flow.jl` example and `horizontal_convection.jl` with topography seem to be working as intended.
> 
>  horizontal_convection_hill.mp4

That looks cool!!

Note that the pressure solver for `NonhydrostaticModel` in complex domains is approximate right now. (It's unclear which problems / kinds of questions are affected, but we know that the dynamics are not quite right.) On #3188 @xkykai is working on a proper pressure solver.

hdrake:
> Note that the pressure solver for `NonhydrostaticModel` in complex domains is approximate right now. (It's unclear which problems / kinds of questions are affected, but we know that the dynamics are not quite right.) On #3188 @xkykai is working on a proper pressure solver.

Yep, watching that closely. We're just at the exploratory stage for now, so we can wait for @xkykai's PR for production runs.

simone-silvestri:
I think this bug allows the option to restructure the code differently and implement the `ValueBoundaryCondition` like the immersed boundary. In that way, we can get rid of the fill halo regions for value, gradient, and flux boundary conditions.
I know this might be a bit more tedious, but I think it might pay off later.

glwagner:
> I think this bug allows the option to restructure the code differently and implement the `ValueBoundaryCondition` like the immersed boundary. In that way, we can get rid of the fill halo regions for value, gradient, and flux boundary conditions.
> I know this might be a bit more tedious, but I think it might pay off later.

I agree that's interesting. However, there are downsides for diagnostics since you can no longer evaluate gradients on the boundary. This means, for example, that we couldn't calculate the viscous momentum flux on a no-slip boundary in a straightforward manner (though it's still possible to figure it out).

glwagner:
> I think this bug allows the option to restructure the code differently and implement the `ValueBoundaryCondition` like the immersed boundary. In that way, we can get rid of the fill halo regions for value, gradient, and flux boundary conditions.
> I know this might be a bit more tedious, but I think it might pay off later.

@simone-silvestri how about merging this now and restructuring the code in a future PR? I think we also have to decide whether we want to accept the downsides of the restructuring, so we may not be ready to implement it yet.

simone-silvestri (approved):
ok, we can deal with that later on

PR #3212, opened 2023-08-08
Remove type instability in WENO β calculation for Float32

milankl:
@simone-silvestri and I replaced a hardcoded `0.5` with `/2` as the former caused a type instability which eventually crashed the baroclinic adjustment setup after 13 days. How exactly that happened we don't know, but it definitely caused a cascade to Float64 - we believe even into the whole tendency - we shouldn't happen.

With this little change Float32 runs stable even with 9th order WENO V-scheme (velocities used for smoothness estimation when reconstructing vorticity) which is Simone's preferred scheme also for global simulations. 

Simone can you throw in the video and the kinetic energy comparison to Float64?


simone-silvestri:
Float32 global simulations are near!

This is the u-velocity at the surface of a simple baroclinic adjustment case: left -> Float64, center -> Float32 right -> difference (u64 - u32)

https://github.com/CliMA/Oceananigans.jl/assets/33547697/4ed0e547-416f-4597-8e2c-a38de7cb9fa5

integrated KE energy time series

ke

more in-depth statistics from @milank incoming

milankl:
image

simone-silvestri:
Nice! looks like the precision does not affect the results nearly as much as the advection order.

milankl:
For a given time step at z=50 (surface) the rounding error in mantissa bits in the tendency addition. If red approaches 23 mantissa bits then there would be a stagnation due to rounding errors, but as long as this doesn't happen systematically it should be fine

image

milankl:
Same at the bottom `z=1`. Why is this similarly turbulent to the surface?

image

milankl:
Calculation via $$\log_2(\vert \frac{progn}{tend} \vert)$$

with $progn$ being the prognostic variable (zonal velocity above) and $tend$ being the tendency taken as two consecutive time steps from the output. It's not exactly that obv because the AB2 time stepper works a bit differently but it's probably good enough! So this quantity is essentially the orders of magnitude (as power of 2) that the prognostic variables are larger than the tendencies. This is equivalent to the mantissa bits of precision lost in the addition of the two.

milankl:
Now also with resting potential energy RPE and available potential energy APE

image

Float32 is more jaggedy, I reckon it's gravity waves that affect buoyancy? But note that there's no obvious bias otherwise. Which may explain why APE decreases less fast as with Float32 too.

milankl:
Buoyancy in constrast does have some stagnation going on (white dots which means we divide by zero in the formula above), but only very few grid cells in the inactive regions. Probably not an issue. Note also how Float64/32 stay really close but WENO7 is already different.

image

milankl:
And `z=1`

image

milankl:
Vertical velocity does not seem to cause any problems (precision loss very low)

image

simone-silvestri:
@milankl should I merge or do you want to add something?

milankl:
No from my side this can be merged, for me it was more the question why your latest run still blew up eventually. I thought we wanted to understand that first. Did you do any other Float32 runs afterwards?

navidcy:
perhaps @milankl mark it as "ready to review" then?

navidcy (approved):
lgtm

navidcy:
@simone-silvestri merge?

navidcy:
> @simone-silvestri merge?

bumping this up

glwagner:
I did it.

PR #3213, opened 2023-08-08
Remove type instability in AB2/3

milankl:
I don't know how relevant these changes are, but I found some more type instabilities in the split explicit free surface model's adams bashforth time integrator. Removing them similar to #3212 

Alternatively we could also hardcode the constants

https://github.com/CliMA/Oceananigans.jl/blob/9140a74536784396d368cf8205432433a9deea3d/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl#L11-L19

as Float32

```julia
const β = 0.281105f0
const α = 1.5f0 + β
```
etc. as this will directly upcast to Float64 if that's used, but not cause a type instability with Float32. But this will again cause problems if we ever want to use something else than Float32/64... Now just the `U★, η★` get a bit lengthier!


glwagner (approved):
Good find!

PR #3214, opened 2023-08-10
Float32 output as default for jld2 and netcdf

milankl:
In the output writers, this suggests to replace `array_type = Array{Float64}` with `type = Float32` to

1) make Float32 the default
2) output always `Array` 

`1)` is so that we save memory in big simulations, and I have the plan to add bitrounding too so that files get like 10x smaller or so. This would be by default in a very conservative way so that we don't accidentally throw away information. Once I've understood where to best do the bitrounding I'll add this to this PR and rename accordingly

`2)` is because I cannot think of a situation where you would want to output something different than Array? But please correct me if I'm wrong. I haven't found another example in this repository.


tomchor:
> `2)` is because I cannot think of a situation where you would want to output something different than Array? But please correct me if I'm wrong. I haven't found another example in this repository.

Does this matter if you're outputting an average in all 3 dimensions? Because that's important, but I'm not if changing the type would affect this situation.

glwagner:
> `2)` is because I cannot think of a situation where you would want to output something different than Array? But please correct me if I'm wrong. I haven't found another example in this repository.

We chose `array_type` to permit the flexibility of other array types. I don't know enough to say that we would _never_ want another array type. Better to be defensive than aggressively constraining user action?

`Float32` used to be the default. However, this produced a lot of pain in some testing situations where we wanted to show bitwise reproducability / accuracy in saving. I can't remember all the details, but after a few user issues (in addition to our own pain), we decided to switch to Float64. I agree that Float32 is better, but could be regarded as "premature optimization" in this case. Definitely open to discuss though.

src/OutputWriters/jld2_output_writer.jl
```diff
@@ -31,7 +31,7 @@ ext(::Type{JLD2OutputWriter}) = ".jld2"
                               dir = ".",
                           indices = (:, :, :),
                        with_halos = false,
-                       array_type = Array{Float64},
+                             type = Float32,
```
glwagner, line 34:
```suggestion
                             float_type = Float32,
```

a little more verbose, if we do go this direction. I think `array_type` is more explicitly regarding what actually happens in the code though, and provides more flexibility (I don't know what the application is, but I'm also not omniscient!)

milankl: What about both `array_type = Array` and `float_type = Float32`?

milankl:
I would need to know the details, but technically bitwise reproducibility is *easier* with Float32 if you compute in Float64 because you throw away 29 bits that could be different. In practice, however, I don't see a difference between Float32/64 regarding this or accuracy. Both are way too precise for what we're doing anyway?

glwagner:
> I would need to know the details, but technically bitwise reproducibility is _easier_ with Float32 if you compute in Float64 because you throw away 29 bits that could be different. In practice, however, I don't see a difference between Float32/64 regarding this or accuracy. Both are way too precise for what we're doing anyway?

It's not actually a question of easy vs hard, its a question of whether you need output in `Float64` for `a == b` to return `true`. If you forget that the default is `Float32` then `a` won't exactly equal `b` for this sort of test. But you might think there's a bug in the output writers or your own code before you realize that it's just because you need `Float64` for this particular kind of check.

I think one example was computing an average on the fly, versus computing it in post-processing by averaging the output?

src/OutputWriters/jld2_output_writer.jl
glwagner: ah, that will work provided `convert(array_type{float_type}, output)` works I think.

glwagner:
This is a clue:

https://github.com/CliMA/Oceananigans.jl/blob/fa5e280115f619d01a460f012328bd7e6d253b38/test/test_jld2_output_writer.jl#L57

or anywhere we hard-coded that

src/OutputWriters/jld2_output_writer.jl
milankl: That looks all possible
```julia
julia> struct Test
           array_type
           float_type
       end

julia> T = Test(Array,Float32)
Test(Array, Float32)

julia> T.array_type{T.float_type}(undef,2,2)
2×2 Matrix{Float32}:
 0.0  0.0
 0.0  0.0

julia> A = rand(2,2)
2×2 Matrix{Float64}:
 0.203629  0.73571
 0.576096  0.28416

julia> convert(T.array_type{T.float_type}, A)
2×2 Matrix{Float32}:
 0.203629  0.73571
 0.576096  0.28416
```

glwagner:
Might've been this one:

https://github.com/CliMA/Oceananigans.jl/blob/fa5e280115f619d01a460f012328bd7e6d253b38/test/test_netcdf_output_writer.jl#L552

But there were also some user issues which lead us to believe it wasn't just about making sure the tests were good, as a recall.

We can still take an opinionated stance that this is an important enough issue that it's worth some temporary user confusion.

glwagner:
I think we should do this...

milankl:
Thanks for the ping. I want to tackle this week after I submitted the JOSS paper!

glwagner:
should we revive this

tomchor:
This has apparently been superceded by https://github.com/CliMA/Oceananigans.jl/pull/4072 so I'm closing this. Feel free to reopen if I'm making a mistake.

PR #3215, opened 2023-08-15
(0.87.0) Fix advective forcing for upwind advection schemes

### LINKED ISSUE #2928: Inconsistent advection from "different" velocities

jagoosw:
Closes #2928

This PR fixes advective forcing so that the velocities are correctly added so that upwind and WENO work.

This changes the advective forcing API as it removes the argument `advection_scheme`.

It also removes the scheme and advected tracer properties of the `AdvectiveForcing` struct so scripts using these will no longer work.

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl
@@ -130,8 +130,7 @@ where `c = C[tracer_index]`.
                         v = SumOfArrays{2}(velocities.v, biogeochemical_velocities.v),
                         w = SumOfArrays{2}(velocities.w, biogeochemical_velocities.w))
 
-        total_velocities = with_advective_forcing(forcing, total_velocities)
-
+    total_velocities = with_advective_forcing(forcing, total_velocities)
 
     return ( - div_Uc(i, j, k, grid, advection, total_velocities, c)
              - ∇_dot_qᶜ(i, j, k, grid, closure, diffusivities, val_tracer_index, c, clock, model_fields, buoyancy)
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl
@@ -7,6 +7,7 @@ using Oceananigans.TurbulenceClosures: ∂ⱼ_τ₁ⱼ, ∂ⱼ_τ₂ⱼ, ∇_dot
 using Oceananigans.Biogeochemistry: biogeochemical_transition, biogeochemical_drift_velocity
 using Oceananigans.TurbulenceClosures: immersed_∂ⱼ_τ₁ⱼ, immersed_∂ⱼ_τ₂ⱼ, immersed_∂ⱼ_τ₃ⱼ, immersed_∇_dot_qᶜ
 using Oceananigans.Advection: div_Uc, U_dot_∇u, U_dot_∇v, boundary_buffer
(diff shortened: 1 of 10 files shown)


glwagner:
Does this eliminate a feature? What was the application of `advected_field`? Seems like we lose some flexibility if we delete it?

jagoosw:
I think it was only used for selecting the advection operator?

glwagner:
> I think it was only used for selecting the advection operator?

ah indeed

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl
```diff
@@ -129,6 +130,9 @@ where `c = C[tracer_index]`.
                         v = SumOfArrays{2}(velocities.v, biogeochemical_velocities.v),
                         w = SumOfArrays{2}(velocities.w, biogeochemical_velocities.w))
 
+        total_velocities = with_advective_forcing(forcing, total_velocities)
+
```
glwagner, line 134: one too many lines here
```suggestion
```

simone-silvestri:
We need `advected_field` because it contains the field that is advected by `AdvectiveForcing` (ad also `advective_kernel` which contains the kernel function used to advect)

https://github.com/CliMA/Oceananigans.jl/blob/fa5e280115f619d01a460f012328bd7e6d253b38/src/Forcings/advective_forcing.jl#L65-L66

jagoosw:
> We need `advected_field` because it contains the field that is advected by `AdvectiveForcing` (ad also `advection_kernel_function` which contains the kernel function used to advect)
> 
> https://github.com/CliMA/Oceananigans.jl/blob/fa5e280115f619d01a460f012328bd7e6d253b38/src/Forcings/advective_forcing.jl#L65-L66

Doing it this way we're using the advection kernel function in the normal tendency function instead though:
https://github.com/CliMA/Oceananigans.jl/blob/a33ba461d544a8eda40a26ded20b14f506d72e45/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl#L129-L135
so we don't need either any more right?

simone-silvestri:
You're right! I hadn't looked at the whole implementation

simone-silvestri (approved):
Looks good!

simone-silvestri:
Sure, let's bump minor version

jagoosw:
Merge when tests pass?

PR #3225, opened 2023-08-23
(0.87.1) Make `set!` a little more ambitious

### LINKED ISSUE #3224: Offline differential operators do not correctly apply boundary conditions
Presently, applying differential operators to fields offline (as opposed to using diagnosing them online using using an `OutputWriter`) yields erroneous results because derivatives seem to be naively using output halo region values (which seem to be filled with zeroes by default) and not overwriting them to satisfy boundary conditions.

One example impact of this is that the [Nusselt number calculation in the `horizontal_convection.jl` example script](https://clima.github.io/OceananigansDocumentation/stable/generated/horizontal_convection/#The-Nusselt-number) is totally meaningless because it is dominated by spuriously large buoyancy gradients in the boundary-adjacent cells.

@ikeshwani and I demonstrate this bug in [this `horizontal_diffusion.jl` script](https://github.com/hdrake/HorizontalConvection/blob/bf2c917e43a6198a55061a46b2097b6b4a9dda3b/horizontal_diffusion.jl), in which we turn off advection in the `horizontal_convection.jl` example and numerically integrate the solution to equilibrium. We compare timeseries of the volume-integrated buoyancy dissipation rates calculated online versus those calculated offline (as in the `horizontal_convection.jl` example). The results show that the online calculation correctly asymptotes to the numerical solution of the equilibrium boundary value problem while the offline calculation is erroneous and effectively yields a Nusselt number that is more than 6 times too high.

equilibration_ratio

The bug is also evident by comparing snapshots of the two buoyancy dissipation rate fields. The dissipation rates computed offline clearly do not satisfy the no-flux boundary conditions on the boundaries.

Screenshot 2023-08-22 at 12 38 27 PM

This bug is present in the live `main` Oceananigans.jl branch (circa `v0.86.0`), as is evident from the movie of the buoyancy dissipation rate field in the [`horizontal_convection.jl` example documentation](https://clima.github.io/OceananigansDocumentation/v0.84.1/generated/horizontal_convection/#Load-saved-output,-process,-visualize) and verified locally.

Screenshot 2023-08-22 at 1 13 30 PM

I am referring to this as a bug because it is contrary to the expected behavior of halos containing the necessary information for satisfying boundary conditions, as discussed in the horizontal convection documentation example:
https://github.com/CliMA/Oceananigans.jl/blob/a226b3efa7db7426ccee03884d610035314955e4/examples/horizontal_convection.jl#L143-L147

glwagner:
It was not copying halo regions, which are needed for correct offline diagnostics since `FieldTimeSeries` uses `set!`.

Closes #3224 

Thanks very much to @hdrake and @ikeshwani for finding this offline-diagnostics-ruining bug!

I think the implementation of `set!` between `Field`s has waffled over time. We should add a test if we want to ensure this behavior.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.86.1"
+version = "0.87.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.86.0"
+version = "0.86.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/Fields/set!.jl
+++ b/src/Fields/set!.jl
@@ -41,14 +41,29 @@ function set!(u::Field, f::Union{Array, CuArray, OffsetArray})
 end
 
 function set!(u::Field, v::Field)
-    # Note: we only copy interior points.
-    # To copy halos use `parent(u) .= parent(v)`.
+    # We implement some niceities in here that attempt to copy halo data,
+    # and revert to copying just interior points if that fails.
     
     if architecture(u) === architecture(v)
-        interior(u) .= interior(v)
+        # Note: we could try to copy first halo point even when halo
+        # regions are a different size. That's a bit more complicated than
+        # the below so we leave it for the future.
+        
+        try # to copy halo regions along with interior data
+            parent(u) .= parent(v)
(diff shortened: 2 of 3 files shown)


glwagner:
Behavior change:

Code snippet

```julia
using Oceananigans
grid = RectilinearGrid(size=1, z=(0, 1), topology=(Flat, Flat, Bounded))
a = CenterField(grid)
b = CenterField(grid)
parent(a) .= 1
set!(b, a)
```

# Consequences on `main`

```julia
julia> parent(a)[:]
7-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0

julia> parent(b)[:]
7-element Vector{Float64}:
 0.0
 0.0
 0.0
 1.0
 0.0
 0.0
 0.0
```

# This PR

```julia
julia> parent(a)[:]
7-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0

julia> parent(b)[:]
7-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
```

simone-silvestri (approved):

glwagner:
Note: we probably need more tests for `FieldTimeSeries` specifically as well to ensure more robust behavior. These are a little more annoying to design though, I will leave for the future.

navidcy (approved):

Issue #3205, opened 2023-07-28
Eliminate `LatitudeLongitudeGrid` in place of the `OrthogonalSphericalShellGrid`


navidcy:
The `OrthogonalSphericalShellGrid` is more general than the `LatitudeLongitudeGrid`. We should add functionality so that _any_ grid with horizontal coordinates/metrics that depend in both horizontal dimensions can be constructed using the `OrthogonalSphericalShellGrid`. When we do that and when we ensure that we can replicate any `LatitudeLongitudeGrid` using the `OrthogonalSphericalShellGrid` we will be able to eliminate the `LatitudeLongitudeGrid`.

navidcy:
cc @glwagner

glwagner:
cc @simone-silvestri

PR #3230, opened 2023-08-24
Generalization of `OrthogonalSphericalShellGrid` to be able to construct any grid with coords-metrics that vary in both `i, j`

### LINKED ISSUE #3200: Generalization of `OrthogonalSphericalShellGrid` to be able to construct _any_ grid with coords-metrics that vary in both `i, j`
At the moment the `OrthogonalSphericalShellGrid` constructor is only specialized to conformal cubed sphere mapping. But `OrthogonalSphericalShellGrid` can be more general to that. We should add a `conformal_cubed_sphere_panel` constructor for the cubed sphere panel. We should also gather all conformal-cubed-sphere-related properties like `ξₗ, ξᵣ, ηₗ, ηᵣ` into a property e.g. `conformal_cubed_sphere`.

navidcy:
Closes #3200

### DIFF
--- a/src/Grids/Grids.jl
+++ b/src/Grids/Grids.jl
@@ -10,6 +10,7 @@ export XFlatGrid, YFlatGrid, ZFlatGrid
 export XRegRectilinearGrid, YRegRectilinearGrid, ZRegRectilinearGrid, HRegRectilinearGrid, RegRectilinearGrid
 export LatitudeLongitudeGrid, XRegLatLonGrid, YRegLatLonGrid, ZRegLatLonGrid
 export OrthogonalSphericalShellGrid, ConformalCubedSphereGrid, ZRegOrthogonalSphericalShellGrid
+export conformal_cubed_sphere_panel
 export node, nodes
 export xnode, ynode, znode, λnode, φnode
 export xnodes, ynodes, znodes, λnodes, φnodes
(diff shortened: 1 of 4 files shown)


glwagner:
Maybe show us what the syntax would look like to construct a lat lon grid? or a lat lon grid with a displaced pole?

navidcy:
I’ll add an example in the docstring.

glwagner:
and paste here?

navidcy:
> and paste here?
For sure

navidcy:
> Maybe show us what the syntax would look like to construct a lat lon grid? or a lat lon grid with a displaced pole?

OK, so what you are asking here implicitly is that this PR also deals (at least partly) with #3205?
Because as it stands now, there is no convenience constructor to create a lat-lon `OrthogonalSphericalShellGrid`; you can just create it using the `struct` constructor. Is this what you'd like to see an example of? It will involve manually creating the arrays with the lat, lon values and the metrics and then provide it to the `OrthogonalSphericalShellGrid` constructor.

navidcy:
I think I'd rather deal with #3205 in a different PR.

glwagner:
Oh no problem. I think there could even be more than one PR. First we just want to show that we can build a latitude longitude grid using OSSG. next we can test dynamics on OSSG and make sure its identical (that's a better test than just visualizing a splash). Then we can design a user interface that makes it easy to build latitdue longitude grid or any other grid on OSSG. Then we can deprecate latitude longitude grid. So there's a lot of ways to break it up. I just thought if OSSG works it'd be nice to demonstrate by running a case on lat lon grid.

navidcy:
I'll do that on a new PR! I'm merging this which is mostly name-change for the constructor and some sorting of conformal mapping related properties together to make OSSG grid more generic.

glwagner (approved):

PR #3233, opened 2023-08-25
FieldTimeSeries for time-dependent boundary condition

simone-silvestri:
The intention of this PR is to extend the capability of `FieldTimesSeries` such that it can be used in time-dependent boundary conditions and forcing.

This PR develops over 2 main points, but it is still a work in progress so suggestions are welcome

1) Allow writing down files with `set!`ing the `OnDisk` flavor of a field time series so that it aligns with the format of our jld2 output writer. In this way, it is possible to easily format BC from different file types (such as binary or text) to be used with Oceananigans. 
An example of this is:
```julia
grid = LatitudeLongitudeGrid(size = (16, 16, 10), latitude = (-60, -40), longitude = (-10, 10), z = (0, 1))
u = XFaceField(grid)
fts = FieldTimeSeries{location(u)...}(grid, 1:100, backend = OnDisk(), path = "testfile.jld2", name = "u")
for i in 1:length(fts.times)
    set!(u, 2i)
    set!(fts, u, i)
end
```
This will generate a file called `testfile.jld2` with the following structure
```
julia> f = jldopen("testfile.jld2")
JLDFile /Users/simonesilvestri/development/Oceananigans.jl/testfile.jld2 (read-only)
 ├─📂 serialized
 │  └─🔢 grid
 └─📂 timeseries
    ├─📂 u
    │  ├─🔢 1
    │  ├─📂 serialized
    │  │  ├─🔢 location
    │  │  └─ ⋯ (2 more entries)
    │  └─ ⋯ (99 more entries)
    └─📂 t (100 entries)
```
which can be easily read by the other field time series types.

2) To do linear interpolation we need at least 2 fields in memory, so `OnDisk` will not do. On the other hand, we might not want _all_ fields in memory as if we are dealing with forcings that might overwhelm the memory (especially on the GPU). So the proposal is to implement a `Chunked` abstraction that only keeps in memory a "chunk" of the data. The details of this implementation are still open do be decided, especially if we want an automatic update of the chunk if we index into an index not existing in memory or if we want the user to be responsible in updating the data in memory through something like a callback
```julia
julia> fts3 = FieldTimeSeries("testfile.jld2", "u", backend = Chunked(chunk_size = 4))
17×16×10×100 FieldTimeSeries{Chunked} located at (Face, Center, Center) on CPU
├── grid: 16×16×10 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×3 halo and with precomputed metrics
├── indices: (Colon(), Colon(), Colon())
└── data: 23×22×16×4 OffsetArray(::Array{Float64, 4}, -2:20, -2:19, -2:13, 1:4) with eltype Float64 with indices -2:20×-2:19×-2:13×1:4
    └── max=8.0, min=0.0, mean=1.67984
```

This is not final and all open to suggestions/changes/improvement

maybe interesting for @yuchenma23


glwagner:
How would we make this work with `time::DateTime`? I don't think we need that feature right now (it matters only if you have observational time-series that span more than one year)

simone-silvestri:
Hmm yeah

simone-silvestri:
Hmmm yeah, do `AbstractTimes` support operations?

src/OutputReaders/field_time_series.jl
```diff
@@ -184,16 +184,16 @@ end
...
+    return compute!(Field(fts[t₂] * (t - t₁) + fts[t₁] * (t₂ - t)))
 end
 
 # Linear time interpolation
 function Base.getindex(fts::FieldTimeSeries, i::Int, j::Int, k::Int, time::Float64)
     Ntimes = length(fts.time)
     t₁, t₂ = index_binary_search(fts.times, time, Ntimes)
-
-    return getindex(fts, i,  j, k, t₂) * (time - t₁) + getindex(fts, i,  j, k, t₁) * (t₂ - time)
+    t = (t₂ - t₁) / (fts.times[t₂] - fts.times[t₁]) * time + t₁
```
glwagner, line 195:
```suggestion
    t = @inbounds (t₂ - t₁) / (fts.times[t₂] - fts.times[t₁]) * time + t₁
```

src/OutputReaders/field_time_series.jl
```diff
@@ -184,16 +184,16 @@ end
...
-
-    return compute!(Field(fts[t₂] * (time - t₁) + fts[t₁] * (t₂ - time)))
+    t = (t₂ - t₁) / (fts.times[t₂] - fts.times[t₁]) * time + t₁
+    return compute!(Field(fts[t₂] * (t - t₁) + fts[t₁] * (t₂ - t)))
 end
 
 # Linear time interpolation
 function Base.getindex(fts::FieldTimeSeries, i::Int, j::Int, k::Int, time::Float64)
     Ntimes = length(fts.time)
     t₁, t₂ = index_binary_search(fts.times, time, Ntimes)
```
glwagner, line 194:
```suggestion
    n₁, n₂ = index_binary_search(fts.times, time, Ntimes)
```

glwagner: most places we use `n` for the time index

glwagner: hmm maybe we can extend `index_binary_search` to work for `x::AbstactRange`. Might make sense to call it "search_index_bracket" after that or something.

simone-silvestri:
at the moment for chunked arrays the user can change with `set!`. For the above example of `fts3` which has a chunk size of 4:

```
julia> fts3[1]
17×16×10 Field{Face, Center, Center} on LatitudeLongitudeGrid on CPU
├── grid: 16×16×10 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×3 halo and with precomputed metrics
├── boundary conditions: Nothing
└── data: 23×22×16 OffsetArray(view(::Array{Float64, 4}, :, :, :, 1), -2:20, -2:19, -2:13) with eltype Float64 with indices -2:20×-2:19×-2:13
    └── max=2.0, min=2.0, mean=2.0

julia> fts3[7]
ERROR: BoundsError: attempt to access 23×22×16×4 Array{Float64, 4} at index [1:23, 1:22, 1:16, 7]

julia> set!(fts3, 7:10)

julia> fts3[7]
17×16×10 Field{Face, Center, Center} on LatitudeLongitudeGrid on CPU
├── grid: 16×16×10 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×3 halo and with precomputed metrics
├── boundary conditions: Nothing
└── data: 23×22×16 OffsetArray(view(::Array{Float64, 4}, :, :, :, 1), -2:20, -2:19, -2:13) with eltype Float64 with indices -2:20×-2:19×-2:13
    └── max=14.0, min=14.0, mean=14.0

julia> fts3[1]
ERROR: BoundsError: attempt to access 23×22×16×4 Array{Float64, 4} at index [1:23, 1:22, 1:16, -5]
Stacktrace:
```

So at the moment the chunked field does not "auto-load". I am not sure whether it is better to autoload or not.

src/OutputReaders/OutputReaders.jl
```diff
@@ -6,14 +6,33 @@ export FieldTimeSeries, FieldDataset
...
-Base.summary(odd::OnDiskData) = "OnDiskData($(odd.path), $(odd.name))"
+struct ChunkedData{D, I}
+    path :: String
+    name :: String
+    data :: D
+    index_range :: I
+end
+
+Base.summary(odd::OnDiskData)  = "OnDiskData($(odd.path), $(odd.name))"
+Base.summary(odd::ChunkedData) = "ChunkedData($(odd.path), $(odd.name), with indices $(index_range))"
```
glwagner: Thought this was for another PR 👀

simone-silvestri: I think I got carried away 😅, I can open another PR

glwagner:
Maybe combine `Chunked` and `InMemory`? eg the default is like `InMemory(chunk=:)`

src/OutputReaders/adapted_field_time_series.jl
```diff
@@ -0,0 +1,17 @@
+using Adapt
+
+struct AdaptedFieldTimeSeries{T, D, χ} <: AbstractArray{T, 4}
+                   data :: D
+                  times :: χ
```
glwagner, line 5: the indents here are huge

src/OutputReaders/adapted_field_time_series.jl
```diff
@@ -0,0 +1,17 @@
+using Adapt
+
+struct AdaptedFieldTimeSeries{T, D, χ} <: AbstractArray{T, 4}
```
glwagner, line 3:
```suggestion
struct GPUAdaptedFieldTimeSeries{T, D, χ} <: AbstractArray{T, 4}
```

It's important to put that to understand the purpose of this

src/OutputReaders/OutputReaders.jl
```diff
@@ -11,15 +11,17 @@ struct InMemory{I} <: AbstractDataBackend
...
+InMemory(; chunk_size = Colon()) = chunk_size isa Colon ? 
+                                          InMemory("", "", chunk_size) :
+                                          InMemory("", "", 1:chunk_size)
 
 struct OnDisk <: AbstractDataBackend 
     path :: String
     name :: String
 end
 
+OnDisk() = OnDisk("", "")
```
glwagner, line 23: What does it mean when you have no path but are `OnDisk`?

src/OutputReaders/field_time_series_update.jl
```diff
@@ -49,6 +49,10 @@ function update_time_series!(fts::InMemoryFieldTimeSeries, n::Int)
     return nothing
 end
 
+# Update _all_ `FieldTimeSeries` in an `AbstractModel`. Loop 
+# over all propery names and extract any of them which is a `FieldTimeSeries`.
+# Flatten the resulting tuple by extracting unique values and set! them to the 
+# correct time range by looping over them
 function update_time_series!(model::AbstractModel, clock::Clock)
```
glwagner, line 56: Maybe this belongs in `Models`?

glwagner: Ah but we have to shuffle the code that, I see

src/OutputReaders/adapted_field_time_series.jl
```diff
@@ -0,0 +1,52 @@
...
+end
+
+Adapt.adapt_structure(to, fts::FieldTimeSeries{LX, LY, LZ}) where {LX, LY, LZ} = 
+    GPUAdaptedFieldTimeSeries{eltype(fts.grid), LX, LY, LZ}(adapt(to, fts.data),
+                                                            adapt(to, fts.times))
+
+@propagate_inbounds Base.lastindex(fts::GPUAdaptedFieldTimeSeries) = lastindex(fts.data)
+@propagate_inbounds Base.lastindex(fts::GPUAdaptedFieldTimeSeries, dim) = lastindex(fts.data, dim)
+
+const XYFTS = FieldTimeSeries{<:Any, <:Any, <:Any, Nothing}
```
glwagner, line 20: add a comment to explain why we need all this

src/OutputReaders/field_time_series_update.jl
glwagner, line 1: the file should be `update_field_time_series.jl`

src/OutputReaders/field_time_series_update.jl
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    fts.data.index_range .= index_range
+    set!(fts, fts.backend.path, fts.backend.name)
+
+    return nothing
+end
+
+# Update the `fts` to contain the time `time_index.time`.
+function update_time_series!(fts::InMemoryFieldTimeSeries, time_index::Time)
```
glwagner, line 33: the file is "update_field_time_series" but this function is "update_time_series". Let's use the same name throughout

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -3,11 +3,12 @@ using PencilArrays: Permutation
 
 import FFTW 
 
+using Oceananigans.Solvers: AbstractSolver
 import Oceananigans.Solvers: poisson_eigenvalues, solve!
 import Oceananigans.Architectures: architecture
 
 
-struct DistributedFFTBasedPoissonSolver{P, F, L, λ, S, I}
+struct DistributedFFTBasedPoissonSolver{P, F, L, λ, S, I} <: AbstractSolver
```
glwagner: What's the point of AbstractSolver?

glwagner:
What is remaining to do on this PR? Do you need help @simone-silvestri ?

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
simone-silvestri: It was a test IO was doing, I found a better way to solve the same problem

src/OutputReaders/OutputReaders.jl
simone-silvestri: This is a convenience constructor to allow passing `path` and `name` in the `FieldTimeSeries` along with `OnDisk`. 

aka
```julia 
fts = FieldTimeSeries(path, name; backend = OnDisk())`
```

glwagner: Add `path` and `name` as properties of `FieldTimeSeries`? `path = name = nothing` when there's no corresponding file.

simone-silvestri:
Suggestions implemented! I ll merge this PR

Issue #1483, opened 2021-03-16
Fusing kernels for calculating diagnostics to improve performance


tomchor:
So, I've been running some simulations with a bunch of diagnostics, which make the simulation slower. I think some of the diagnostics that I'm calculating are already being calculated by Oceananigans (like derivatives). If I understand correctly, in each of those cases Oceananigans calculates that variable once to get the tendencies and then does the same calculation again when for my diagnostics, which seems wasteful.

Is it possible (or desirable) to create a way for the code not to do the calculation twice if the user wants that diagnostic specifically? Maybe pass options when creating the model like:

```julia
model = IncompressibleModel(grid, computed_fields=(:dudz, :dvdz,))
```
so that those fields get stored just like the velocity and the tracers do?

Thoughts?

francispoulin:
The library computes the derivatives that are required to compute the tendencies, but they are not stored since that would not be very efficient.  Getting some of these values but I don't know what that would look like.  Also, if you are not computing this field at every time step, the cost of computing it sepratelyl might not be that high, but that of course depends on the particular problem you are dealing with.

glwagner:
For better or for worse, Oceananigans currently does not store intermediate terms in the computation of a PDE's right hand side (with notable exceptions hydrostatic pressure and eddy diffusivities). In other words, a single, sometimes large kernel that evaluates the right hand side at each grid point `i, j, k` is compiled for each PDE in a user's model, and there's no way to pull out intermediate steps in that calculation to use elsewhere.

It's important to note when considering optimization strategies that our computations are probably memory-limited, rather than compute-limited. In other words, we think the process of transferring data from global memory to local thread memory is a bottleneck for our computations (we can really only know this through profiling a particular application, however, since all models are different...) Storing intermediate components of the tendency terms would probably create more memory accesses overall (since rather than immediately using intermediate results for subsequent calculations, we would have to send them to global memory, and then back, to complete the evaluation of a tendency) --- and thus could slow down tendency evaluations that are performed 1-3 times per time-step. For example, our best idea for speeding up tendency evaluations is to better manage memory movement using GPU shared memory (unfortunately, we haven't had the time to explore such optimization strategies...)

I think there may be other ways to optimize diagnostics calculations, however.

# Fusing `ComputedField` kernels

One possibility to speed up diagnostics is to "fuse" kernels for different `ComputedField` diagnostics. The kernel for a `ComputedField` is

[...]

PR #3234, opened 2023-08-25
Implementing kernel fusion on computed fields to address #1483

tztsai:
I implemented a struct `FusedComputedFields` containing multiple ComputedFields on the same grid, and a single kernel to do the computation of these fields together.


glwagner:
Nice! I have a few comments but this looks like great progress.

I think for this to be useful, we have to have a way for it to work with output writing. It looks like that would require some non-trivially changes to the output writer. We can maybe just prototype the ideas with `JLD2OutputWriter` and see how that goes.

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,62 @@ end
 
     @inbounds data[i′, j′, k′] = operand[i′, j′, k′]
 end
+
+struct FusedComputedFields
+    fields::Vector{ComputedField}
+    grid::AbstractGrid
+    datas::Vector{AbstractArray}
+    operands::Vector{AbstractOperation}
+    indices::Tuple
```
glwagner, line 98: This probably should be type parameterized, ie

```julia
indices :: I
```

and

```julia
FusedComputedFields{I}
```

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,62 @@ end
 
     @inbounds data[i′, j′, k′] = operand[i′, j′, k′]
 end
+
+struct FusedComputedFields
+    fields::Vector{ComputedField}
+    grid::AbstractGrid
+    datas::Vector{AbstractArray}
+    operands::Vector{AbstractOperation}
```
glwagner, line 97: we don't need these to be stored explicitly right?

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,62 @@ end
 
     @inbounds data[i′, j′, k′] = operand[i′, j′, k′]
 end
+
+struct FusedComputedFields
+    fields::Vector{ComputedField}
+    grid::AbstractGrid
```
glwagner, line 95: This needs to be type parameterized, otherwise `FusedComputedFields` is not concretely typed:

```julia
grid :: G
```

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,62 @@ end
 
     @inbounds data[i′, j′, k′] = operand[i′, j′, k′]
 end
+
+struct FusedComputedFields
+    fields::Vector{ComputedField}
```
glwagner, line 94: I think we need this to be either `Tuple` or `NamedTuple`. Probably the best thing is to leave it as an abstract type so we can use either Tuple or NamedTuple as needed (`NamedTuple` may be useful for output)

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,62 @@ end
 
     @inbounds data[i′, j′, k′] = operand[i′, j′, k′]
 end
+
+struct FusedComputedFields
```
glwagner, line 93: in summary here's my suggestion:

```julia
struct FusedComputedFields{G, F, I}
    grid :: G
    fields :: F
    indices :: I
end
```

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,62 @@ end
...
+    function FusedComputedFields(fields::Vector{ComputedField})
+        grid = fields[1].grid
+        indices = fields[1].indices
+        arch = architecture(fields[1])
+        sz = size(fields[1])
+        for field in fields[2:end]
+            @assert field.grid === grid
+            @assert field.indices == indices
+            @assert architecture(field) === arch
+            @assert size(field) == sz
```
glwagner, line 109: We typically like to give more explicit error messages, so we don't use `@assert` to throw errors for incorrect input. (In other words, its more helpful to the user to say something like "one of your grids is different"). There's also the concern that it can be disabled

https://docs.julialang.org/en/v1/base/base/#Base.@assert

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,62 @@ end
...
+    grid::AbstractGrid
+    datas::Vector{AbstractArray}
+    operands::Vector{AbstractOperation}
+    indices::Tuple
+
+    function FusedComputedFields(fields::Vector{ComputedField})
+        grid = fields[1].grid
+        indices = fields[1].indices
+        arch = architecture(fields[1])
+        sz = size(fields[1])
```
glwagner, line 104: We usually like to use `first` rather than `list[1]`, but that's a minor point...

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,62 @@ end
...
+        fill_halo_regions!(comp)
+    end
+
+    return fused
+end
+
+function compute_fused_computed_fields!(fused)
+    arch = architecture(fused.fields[1])
+    launch!(arch, fused.grid, size(fused.fields[1]), _fused_compute!, 
+            fused.data, fused.operands, fused.indices)
```
glwagner, line 136: If the struct changes are adopted then

```suggestion
    fused_data = Tuple(f.data for f in fused.fields)
    fused_operands = Tuple(f.operand for f in fused.field)
    launch!(arch, fused.grid, size(first(fused.fields)), _fused_compute!,  fused_data, fused_operands, fused.indices)
```

glwagner:
Few TODO:

1. Add a minimal test (probably only need one or two to start, we can build the tests up gradually if this gets adopted)
2. Brainstorm how to use this with output writers (eg `JLD2OutputWriter(model, outputs; fuse_computations=true)`) or something like that

@simone-silvestri curious what you think

simone-silvestri:
Oh, it looks good, it would be interesting to perform some complex operation and see if we get a speedup

simone-silvestri:
We could also support `compute!(fields::NTuple{N, <:Fields})` which would use `FusedComputedFields` in the background

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,62 @@ end
...
+        arch = architecture(fields[1])
+        sz = size(fields[1])
+        for field in fields[2:end]
+            @assert field.grid === grid
+            @assert field.indices == indices
+            @assert architecture(field) === arch
+            @assert size(field) == sz
+        end
+        datas = [field.data for field in fields]
+        operands = [field.operand for field in fields]
```
simone-silvestri, line 112:
```suggestion
        datas = [field.data for field in fields]
        operands = [field.operand for field in fields]
```
tuples are better

simone-silvestri: actually this is already included in @glwagner's suggestion

simone-silvestri:
I would also add a check that the fields are "full" fields, aka no `Windowed` fields (with particular indices different from the one we use to launch the kernel) and no `Reduced` fields with lower than 3 dimensions. We want to separate out those fields and treat them differently. 
We have an implementation like that for `fill_halo_regions!` here
https://github.com/CliMA/Oceananigans.jl/blob/29a99a0c235b2f6bf0cec525f2249125ad254ccc/src/Fields/field_tuples.jl#L54-L78

What I would probably do actually, instead of a new struct is to extent `compute!` for tuple fields as we do for `fill_halo_regions!` and use your newly implemented kernels for tuples of full fields.

On the other hand, if defining a new type is useful for OutputWriters I am ok with it

src/AbstractOperations/computed_field.jl
TZTsai: The new commit throws `ArgumentError` instead of using `@assert`.

src/AbstractOperations/computed_field.jl
TZTsai: Updated.

src/AbstractOperations/computed_field.jl
TZTsai: updated

TZTsai:
> I would also add a check that the fields are "full" fields, aka no `Windowed` fields (with particular indices different from the one we use to launch the kernel) and no `Reduced` fields with lower than 3 dimensions. We want to separate out those fields and treat them differently. We have an implementation like that for `fill_halo_regions!` here
> 
> https://github.com/CliMA/Oceananigans.jl/blob/29a99a0c235b2f6bf0cec525f2249125ad254ccc/src/Fields/field_tuples.jl#L54-L78
> 
> What I would probably do actually, instead of a new struct is to extend `compute!` for tuple fields as we do for `fill_halo_regions!` and use your newly implemented kernels for tuples of full fields.
> 
> On the other hand, if defining a new type is useful for OutputWriters I am ok with it

The new commit follows the suggestion of overloading the `compute!` method to apply it on a tuple of `ComputedField`. I also added a check to make sure that all these fields are `FullField`.

glwagner:
> The new commit follows the suggestion of overloading the compute! method to apply it on a tuple of ComputedField.

Good point that we may not need an independent type (except for the very minor performance optimization of precomputing the compatibility between fields -- which probably doesn't matter)

src/AbstractOperations/computed_field.jl
```diff
@@ -89,3 +89,87 @@ end
...
+            throw(ArgumentError("All fields to compute must have the same architecture!"))
+        end
+        if size(comp) != sz
+            throw(ArgumentError("All fields to compute must have the same size!"))
+        end
+        if (comp isa FullField) && !(comp isa ReducedField)
+            # push!(ordinary_field_ids, i)
+            throw(ArgumentError("All fields to compute must be FullField!"))
+        end
+    end
```
glwagner, line 151: this looks great!! Just little more compact syntax suggestion:

```suggestion
    for (i, comp) in enumerate(comps)
        comp.grid !== grid || throw(ArgumentError("All fields to compute must have the same grid!"))
        comp.indices != indices || throw(ArgumentError("All fields to compute must have the same indices!"))
        architecture(comp) !== arch || throw(ArgumentError("All fields to compute must have the same architecture!"))
        size(comp) != sz || throw(ArgumentError("All fields to compute must have the same size!"))
        (comp isa FullField) && !(comp isa ReducedField) || throw(ArgumentError("All fields to compute must be FullField!"))
           
        end
    end
```

PR #3236, opened 2023-08-26
Linear time interpolation in `FieldTimeSeries`

simone-silvestri:


navidcy:
If time is `DateTime` the _only_ integer indices work but it fails for fractional indices. Correct?

simone-silvestri:
yep. I am not super familiar with what date-times supports in term of operations, if it supports summation and division it is straightforward to extend this to date times

navidcy:
I think it can be done but it’s definitely not as straightforward as +, /. 

Happy to approve now and open issue to generalize the method for TimeDates later.

navidcy:
Out of curiosity @simone-silvestri could you check

```Julia
fts2[1, 2, 3, 4] == fts2[1, 2, 3, 4.0]
```

in your example with the `.jld2` file above?

glwagner:
We wouldn't have `Clock` available in post-processing.

simone-silvestri:
So I would propose a clock implementation for easy internal manipulation which calls the `at_time` function available for postprocessing

glwagner:
`Clock` does save the definition of a new type. However, I think it is misleading because it contains extra information like the `iteration` and `stage`. Here, the time-interpolation is independent of those.

Why not a new type `Time`? What is your opposition to that?

glwagner:
Basically I think the "convenience" of not having to define a new type is trivial. You can define `Time(clock::Clock)`.

simone-silvestri:
I am thinking about the ease of use in bc, for example
```julia
@inline time_interpolated_bc(i, j, grid, clock, fields, p) = p.time_array[i, j, 1, clock]
```
vs
```julia
@inline function time_interpolated_bc(i, j, grid, clock, fields, p) 
   time = Time(clock.time)
   return p.time_array[i, j, 1, time]
end
```
I find the first implementation a bit more straightforward, but we can also have both

glwagner:
also worth checking out 

https://github.com/rafaqz/DimensionalData.jl

the `Time` object is a "selector" in their terminology. IF we have something similar here, we could eventually consider sub typing `AbstractDimArray` (which has been proposed before), which gives us a lot of powerful functionality. Someone passionate about that stuff has to take that on for a full implementation but at least we are consistent with it, which helps

src/TimeSteppers/clock.jl
```diff
@@ -36,6 +36,9 @@ next_time(clock::Clock{<:AbstractTime}, Δt) = clock.time + Nanosecond(round(Int
 tick_time!(clock, Δt) = clock.time += Δt
 tick_time!(clock::Clock{<:AbstractTime}, Δt) = clock.time += Nanosecond(round(Int, 1e9 * Δt))
 
+Time(clock::Clock{<:Number}) = Time(clock.time)
+Time(clock::Clock{<:AbstractTime}) = Time(clock.time)
```
glwagner, line 40:
```suggestion
Time(clock::Clock) = Time(clock.time)
```

src/Units.jl
```diff
@@ -120,4 +120,15 @@ A `Float64` constant equal to 1024`GiB`. Useful for increasing the clarity of sc
...
 
+"""
+    struct Time{T}
+        time_in_seconds::T 
+    end
+
+A ``selector'' type to perform indexing in time
+"""
+struct Time{T}
+    time_in_seconds :: T
```
glwagner, line 131:
```suggestion
    time :: T
```

so that we can generalize to `DateTime` later (where we will need the date not just seconds, in order to interpolate `Vector{DateTime}`)

glwagner: As well as non-dimensional time units

src/OutputReaders/field_time_series.jl
```diff
@@ -183,9 +183,9 @@ function Base.getindex(fts::FieldTimeSeries{LX, LY, LZ, OnDisk}, n::Int) where {
 end
 
 # Linear time interpolation
-function Base.getindex(fts::FieldTimeSeries, time::Time)
+function Base.getindex(fts::FieldTimeSeries, time_index::Time)
     Ntimes = length(fts.times)
-    time = time.time_in_seconds
+    time = time_index.time
```
glwagner, line 188: perfect

src/Units.jl
```diff
@@ -120,4 +120,15 @@ A `Float64` constant equal to 1024`GiB`. Useful for increasing the clarity of sc
 """
 const TiB = 1024GiB
 
+"""
+    struct Time{T}
+        time_in_seconds::T 
+    end
+
+A ``selector'' type to perform indexing in time
```
glwagner, line 128:
```suggestion
    Time(t)

Return a time "selector" at the continuous time `t` for linearly interpolating `FieldTimeSeries`.

Examples
=======

```julia
# Interpolate `field_time_series` to `t=0.1`, returning `interpolated::Field`
interpolated = field_time_series[Time(0.1)]

# Interpolate `field_time_series` at `i, j, k` and `t=0.1`
interpolated_ijk = field_time_series[i, j, k, Time(0.1)]
```
```

File: src/Units.jl [hunk 5]
```diff
@@ -121,11 +121,19 @@ A `Float64` constant equal to 1024`GiB`. Useful for increasing the clarity of sc
...
+
+Examples
+=======
+
+```julia
+# Interpolate `field_time_series` to `t=0.1`, returning `interpolated::Field`
+interpolated = field_time_series[Time(0.1)]
+
+# Interpolate `field_time_series` at `i, j, k` and `t=0.1`
+interpolated_ijk = field_time_series[i, j, k, Time(0.1)]
```
glwagner, line 136:
```suggestion
interpolated_ijk = field_time_series[i, j, k, Time(0.1)]
```
```

Comment (glwagner) [reply src/Units.jl:136]: It's hard to add those triple ticks with a suggestion!

File: src/TimeSteppers/clock.jl [hunk 6]
```diff
@@ -36,6 +36,8 @@ next_time(clock::Clock{<:AbstractTime}, Δt) = clock.time + Nanosecond(round(Int
 tick_time!(clock, Δt) = clock.time += Δt
 tick_time!(clock::Clock{<:AbstractTime}, Δt) = clock.time += Nanosecond(round(Int, 1e9 * Δt))
 
+Time(clock::Clock) = Time(clock.time)
```
glwagner, line 39: do we need to import `Time`?

simone-silvestri: no we should export it

simone-silvestri: actually it is exported

glwagner: yea but to extent the constructor `Time`, we need to import it here since it's defined in `Units`

glwagner:
We will also have to figure out (perhaps in a future PR) how to insert boundary conditions to `Time`, eg `Cyclic` (so forcing / specification repeats) or `Continued`

src/OutputReaders/field_time_series.jl
```diff
@@ -180,6 +182,34 @@ function Base.getindex(fts::FieldTimeSeries{LX, LY, LZ, OnDisk}, n::Int) where {
...
+    end
+    
+    # fractional index
+    @inbounds n = (n₂ - n₁) / (fts.times[n₂] - fts.times[n₁]) * (time - fts.times[n₁]) + n₁
+    return compute!(Field(fts[n₂] * (n - n₁) + fts[n₁] * (n₂ - n)))
+end
+
+# Linear time interpolation
+function Base.getindex(fts::FieldTimeSeries, i::Int, j::Int, k::Int, time_index::Time)
+    Ntimes = length(fts.times)
```
glwagner, line 201: `length` is slow for big vectors right? Should we allow this to be passed into `Time`?

src/OutputReaders/field_time_series.jl
```diff
@@ -180,6 +182,34 @@ function Base.getindex(fts::FieldTimeSeries{LX, LY, LZ, OnDisk}, n::Int) where {
...
+    time = time_index.time
+    n₁, n₂ = index_binary_search(fts.times, time, Ntimes)
+    if n₁ == n₂ # no interpolation
+        return getindex(fts, i, j, k, n₁)
+    end
+    
+    # fractional index
+    @inbounds n = (n₂ - n₁) / (fts.times[n₂] - fts.times[n₁]) * (time - fts.times[n₁]) + n₁
+    return getindex(fts, i, j, k, n₂) * (n - n₁) + getindex(fts, i, j, k, n₁) * (n₂ - n)
+end
```
glwagner, line 211:
```suggestion
   
    # fractional index
    @inbounds n = (n₂ - n₁) / (fts.times[n₂] - fts.times[n₁]) * (time - fts.times[n₁]) + n₁
   fts_interpolated = getindex(fts, i, j, k, n₂) * (n - n₁) + getindex(fts, i, j, k, n₁) * (n₂ - n)
   
   # Don't interpolate if n = 0.
   return ifelse(n₁ == n₂, getindex(fts, i, j, k, n₁), fts_interpolated)
end
```

src/OutputReaders/field_time_series.jl
glwagner: then we can implement something like

```julia
maybe_diagnose_length(time::Time{<:Int}, times) = time.Ntimes
maybe_diagnose_length(::Time, times) = length(times)
```

or mahybe you have a better name

src/OutputReaders/field_time_series.jl
simone-silvestri: I prefer not to do the interpolation if it is not required

glwagner: Why?

simone-silvestri: Ah, I was actually looking at the other getindex, no in the kernel it's ok

glwagner: (It's not good practice to resolve other people's comments, unless they are not responding or something --- the reviewer gets priority to decide whether their comment has been resolved or not.)

glwagner: The main concern here is performance for time interpolation of `FieldTimeSeries` within a `Forcing`, which enters into the expensive 3D kernels to calculate tendencies. This will be an important case; for example for driving LES using 3D forcing derived from GCM. We can't know the flip side either way without profiling or benchmarking, but `ifelse` is simply a conservative choice to avoid performance cliffs (or at least, that has been our experience time and time again in the past --- maybe there is new information we should consider?)

simone-silvestri: yeah, the kernel is fast so it shouldn't be a problem. On the other hand, a simple if condition should be okay here as there should not be problems of branch divergence (all threads should look at the same time). Still, I m okay with interpolating in the kernel

glwagner: What kernel is fast? I'm referring to the situation where this is embedded in another kernel. For example, this `getindex` call can be embedded in the kernels that calculate model tendencies.

glwagner: > On the other hand, a simple if condition should be okay here as there should not be problems of branch divergence (all threads should look at the same time).

Got it. Is that still the case in very complicated kernels, for example kernels that calculate 3D tendencies?

simone-silvestri: sorry, linear interpolation in GPU kernels is fast. That is if threads access the same time index. There could be a situation in which different threads access different time indexes but it doesn't immediately come to mind.

glwagner: My question is whether that principle is still useful when the linear interpolation is surrounded by lots of other expensive operations, as can happen when this is used inside a tendency kernel.

glwagner: Here's another example: `FieldTimeSeries` is used to store data for an advecting velocity field. This is the application you'd get if you're interested in, for example, advecting biogeochemical tracers with a velocity field derived from the ECCO state estimate. In that case the linear interpolation is inserted into an advection scheme.

PR #3240, opened 2023-08-29
Better validation for `LatitudeLongitudeGrid` + add test for `sum(Azᶜᶜᵃ) ≈ 4πr²`

### LINKED ISSUE #3241: `LatitudeLongitudeGrid` validation seems to be leaky
Why the lat-lon validation _only_ happens if `isnothing(topology)`? See here:

https://github.com/CliMA/Oceananigans.jl/blob/29a99a0c235b2f6bf0cec525f2249125ad254ccc/src/Grids/latitude_longitude_grid.jl#L240-L262

This way, if I prescribe latitudes beyond +/-90 no error is given......

```Julia
julia> using Oceananigans

julia> grid = LatitudeLongitudeGrid(topology=(Bounded, Bounded, Bounded), size = (5, 5, 2), longitude = [0, 1, 3, 6, 12, 20], latitude = [10, 12, 20, 40, 100, 170], z = (-1, 0))
5×5×2 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×3 halo and with precomputed metrics
├── longitude: Bounded  λ ∈ [0.0, 20.0]   variably spaced with min(Δλ)=1.0, max(Δλ)=8.0
├── latitude:  Bounded  φ ∈ [10.0, 170.0] variably spaced with min(Δφ)=2.0, max(Δφ)=70.0
└── z:         Bounded  z ∈ [-1.0, 0.0]   regularly spaced with Δz=0.5
```

navidcy:
Closes #3241

### DIFF
--- a/src/Grids/latitude_longitude_grid.jl
+++ b/src/Grids/latitude_longitude_grid.jl
@@ -238,22 +238,22 @@ function with_precomputed_metrics(grid)
 end
 
 function validate_lat_lon_grid_args(FT, latitude, longitude, z, size, halo, topology, precompute_metrics)
+    Nλ, Nφ, Nz = N = size
+    
+    λ₁, λ₂ = get_domain_extent(longitude, Nλ)
+    @assert λ₁ <= λ₂ && λ₂ - λ₁ ≤ 360
 
+    φ₁, φ₂ = get_domain_extent(latitude, Nφ)
+    @assert -90 <= φ₁ <= φ₂ <= 90
+
+    (φ₁ == -90 || φ₂ == 90) &&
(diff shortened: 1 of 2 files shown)


glwagner (approved):
Nice!

Should we implement this test for OrthogonalSphericalShellGrid instead, considering we are going to deprecate LatitutdeLongitudeGrid?

navidcy:
> Nice!
> 
> Should we implement this test for OrthogonalSphericalShellGrid instead, considering we are going to deprecate LatitutdeLongitudeGrid?

That's already implemented!

navidcy:
I also improved the validation according to the discussion in #3241.

PR #3243, opened 2023-08-30
Fix latex rendering in Ocean wind and mixing example

navidcy:
`minutes` shouldn't be italic as is atm:

https://clima.github.io/OceananigansDocumentation/dev/generated/ocean_wind_mixing_and_convection/#Turbulence-visualization


glwagner (approved):
🚀

PR #3245, opened 2023-08-31
Provide a better explanation of the `max_change` and `min_change` of `TimeStepWizard`

iuryt:
closes #3244

Issue #3244, opened 2023-08-30
Add to the docs a better explanation about `max_change` and `min_change` of the `TimeStepWizard`


iuryt:
Hi,

I have been confused about what the `max_change` and `min_change` means.
I understand from the documentation that this is to avoid the time step to change too quickly, but how is it used?

Is it in time units? e.g., taking the default `max_change` of 1.1, that means that for dated simulation it will change the time step at max in 1.1 seconds? 

Is it a multiplication? e.g., `new_time_step = 1.1*time_step` ?

I think we can try to be more clear about that in the docs.

glwagner:
It's the latter: `max_change` is the maximum relative change of the time step and is therefore non-dimensional. Thus the new time step satisfies the criteria

```julia
new_dt <= max_change * old_dt
```

It's implemented here:

https://github.com/CliMA/Oceananigans.jl/blob/0d39f5ce93b75e7df2b896fbd8326214dbf5cea7/src/Simulations/time_step_wizard.jl#L96-L99

Improving the docs is always a good idea!

iuryt:
I can do that to the docs.

PR #3248, opened 2023-08-31
Update `TimeStepWizard`

iuryt:
Reopens #3244 and #3245 

If I understand it correctly, it does not make sense to have `min_change > 1` or `max_change < 1`. Am I right?
If so, should we add at least a warning for those cases?

Please, let me know if I 'reopened' the PR correctly.

### DIFF
--- a/src/Simulations/time_step_wizard.jl
+++ b/src/Simulations/time_step_wizard.jl
@@ -74,7 +74,7 @@ function TimeStepWizard(FT=Float64;
     # check if user gave max_change or min_change values that are invalid
     min_change ≥ 1 && throw(ArgumentError("min_change must be < 1. You provided min_change = $min_change."))
   
-    max_change ≥ 1 && throw(ArgumentError("max_change must be > 1. You provided max_change = $max_change."))
+    max_change ≤ 1 && throw(ArgumentError("max_change must be > 1. You provided max_change = $max_change."))
   
     # user wants to limit by diffusive CFL and did not provide custom function to calculate timescale
     if isfinite(diffusive_cfl) && (cell_diffusion_timescale === infinite_diffusion_timescale)
--- a/src/Simulations/time_step_wizard.jl
+++ b/src/Simulations/time_step_wizard.jl
@@ -34,8 +34,8 @@ Callback for adapting simulation to maintain the advective Courant-Freidrichs-Le
 number to `cfl`, the `diffusive_cfl`, while also maintaining `max_Δt`, `min_Δt`, and
 satisfying `max_change` and `min_change` criteria so that the simulation's timestep
 `simulation.Δt` is not adapted "too quickly". In other words, `max_change` is the maximum
-and `min_change` is the minimum relative change of the time step and are therefore non-dimensional
-(`min_change * old_Δt <= new_Δt <= max_change * old_Δt`).
+and `min_change` is the minimum relative change of the time step and are, therefore, non-dimensional
+(`min_change * old_Δt ≤ new_Δt ≤ max_change * old_Δt`).
 
 For more information on the CFL number, see its [wikipedia entry]
 (https://en.wikipedia.org/wiki/Courant%E2%80%93Friedrichs%E2%80%93Lewy_condition).
--- a/src/Simulations/time_step_wizard.jl
+++ b/src/Simulations/time_step_wizard.jl
@@ -33,7 +33,9 @@ Base.summary(wizard::TimeStepWizard) = string("TimeStepWizard(",
 Callback for adapting simulation to maintain the advective Courant-Freidrichs-Lewy (CFL)
 number to `cfl`, the `diffusive_cfl`, while also maintaining `max_Δt`, `min_Δt`, and
 satisfying `max_change` and `min_change` criteria so that the simulation's timestep
-`simulation.Δt` is not adapted "too quickly".
+`simulation.Δt` is not adapted "too quickly". In other words, `max_change` is the maximum
+and `min_change` is the minimum relative change of the time step and are therefore non-dimensional
+(`min_change * old_Δt <= new_Δt <= max_change * old_Δt`).
 
 For more information on the CFL number, see its [wikipedia entry]
 (https://en.wikipedia.org/wiki/Courant%E2%80%93Friedrichs%E2%80%93Lewy_condition).
--- a/src/Simulations/time_step_wizard.jl
+++ b/src/Simulations/time_step_wizard.jl
@@ -72,10 +72,9 @@ function TimeStepWizard(FT=Float64;
                         cell_diffusion_timescale = infinite_diffusion_timescale)
     
     # check if user gave max_change or min_change values that are invalid
-    if min_change >= 1.0 || max_change <= 1.0
-        throw(ArgumentError("Invalid value: min_change should be < 1 and max_change should be > 1. " *
-                            "You provided min_change = $min_change and max_change = $max_change."))
-    end
+    min_change ≥ 1 && throw(ArgumentError("min_change must be < 1"))
+
+    max_change ≥ 1 && throw(ArgumentError("max_change must be > 1"))
   
     # user wants to limit by diffusive CFL and did not provide custom function to calculate timescale
     if isfinite(diffusive_cfl) && (cell_diffusion_timescale === infinite_diffusion_timescale)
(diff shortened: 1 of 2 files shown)


glwagner:
> If I understand it correctly, it does not make sense to have `min_change > 1` or `max_change < 1`. Am I right?
> If so, should we add at least a warning for those cases?

Right, it doesn't make sense. But I think we should `throw(ArgumentError` rather than merely warn because that can do bad things like spuriously increase your time-step.

glwagner:
Some stuff later in the docstring is also weird. For example it says

> To use `TimeStepWizard`, adapt in a [`Callback`](@ref) and add it to a `Simulation`:

I don't know what it means to "adapt" a function in a callback. There's no adaptation, `TimeStepWizard` was _born_ for `Callback`!! Better to say that we "wrap" or "insert it" into `Callback`?

iuryt:
> Some stuff later in the docstring is also weird. For example it says
> 
> > To use `TimeStepWizard`, adapt in a [`Callback`](@ref) and add it to a `Simulation`:
> 
> I don't know what it means to "adapt" a function in a callback. There's no adaptation, `TimeStepWizard` was _born_ for `Callback`!! Better to say that we "wrap" or "insert it" into `Callback`?

Good catch! I will work on the docstring and error after the weekend.

iuryt:
I believe I implemented everything we discussed. Please, let me know if that's seems fine to be merged.

src/Simulations/time_step_wizard.jl
```diff
@@ -30,20 +30,23 @@ Base.summary(wizard::TimeStepWizard) = string("TimeStepWizard(",
...
-(`min_change * old_Δt ≤ new_Δt ≤ max_change * old_Δt`).
+Callback function that adjusts the simulation time step to meet specified target values 
+for advective and diffusive Courant-Friedrichs-Lewy (CFL) numbers (`cfl` and `diffusive_cfl`), 
+subject to the limits
+
+```julia
+    max(min_Δt, min_change * previous_Δt) ≤ new_Δt ≤ min(max_Δt, max_change * previous_Δt)
+```
+
+where `new_Δt` is the new time step calculated by `TimeStepWizard`.
```
navidcy, line 41:
```suggestion
where `new_Δt` is the new time step calculated by the `TimeStepWizard`.
```

src/Simulations/time_step_wizard.jl
```diff
@@ -30,20 +30,23 @@ Base.summary(wizard::TimeStepWizard) = string("TimeStepWizard(",
...
-satisfying `max_change` and `min_change` criteria so that the simulation's timestep
-`simulation.Δt` is not adapted "too quickly". In other words, `max_change` is the maximum
-and `min_change` is the minimum relative change of the time step and are, therefore, non-dimensional
-(`min_change * old_Δt ≤ new_Δt ≤ max_change * old_Δt`).
+Callback function that adjusts the simulation time step to meet specified target values 
+for advective and diffusive Courant-Friedrichs-Lewy (CFL) numbers (`cfl` and `diffusive_cfl`), 
+subject to the limits
+
+```julia
+    max(min_Δt, min_change * previous_Δt) ≤ new_Δt ≤ min(max_Δt, max_change * previous_Δt)
```
navidcy, line 38:
```suggestion
max(min_Δt, min_change * previous_Δt) ≤ new_Δt ≤ min(max_Δt, max_change * previous_Δt)
```

src/Simulations/time_step_wizard.jl
```diff
@@ -67,7 +70,13 @@ function TimeStepWizard(FT=Float64;
                         min_Δt = 0.0,
                         cell_advection_timescale = cell_advection_timescale,
                         cell_diffusion_timescale = infinite_diffusion_timescale)
-
+    
+    # check if user gave max_change or min_change values that are invalid
+    if min_change >= 1.0 || max_change <= 1.0
+        throw(ArgumentError("Invalid value: min_change should be < 1 and max_change should be > 1. " *
+                            "You provided min_change = $min_change and max_change = $max_change."))
+    end
```
navidcy, line 78: is this simpler and cleaner?
```suggestion
    min_change ≥ 1 && throw(ArgumentError("min_change must be < 1"))

    max_change ≥ 1 && throw(ArgumentError("max_change must be > 1"))
```

src/Simulations/time_step_wizard.jl
```diff
@@ -70,7 +70,13 @@ function TimeStepWizard(FT=Float64;
                         min_Δt = 0.0,
                         cell_advection_timescale = cell_advection_timescale,
                         cell_diffusion_timescale = infinite_diffusion_timescale)
-
+    
+    # check if user gave max_change or min_change values that are invalid
+    if min_change >= 1.0 || max_change <= 1.0
```
glwagner:
```suggestion
    if min_change >= 1 || max_change <= 1
```

glwagner: I think it's better to make these separate checks. Ie check `min_change` and then afterwards check `max_change`. The error message will be more specific.

navidcy: Yes, that aligns with what I suggested -- already implemented ;)

src/Simulations/time_step_wizard.jl
```diff
@@ -30,20 +30,23 @@ Base.summary(wizard::TimeStepWizard) = string("TimeStepWizard(",
...
+where `new_Δt` is the new time step calculated by the `TimeStepWizard`.
 
 For more information on the CFL number, see its [wikipedia entry]
 (https://en.wikipedia.org/wiki/Courant%E2%80%93Friedrichs%E2%80%93Lewy_condition).
 
 Example
 =======
 
-To use `TimeStepWizard`, adapt in a [`Callback`](@ref) and add it to a `Simulation`:
+To use `TimeStepWizard`, insert it into a [`Callback`](@ref) and add it to a `Simulation`:
```
glwagner, line 49:
```suggestion
To use `TimeStepWizard`, insert it into a [`Callback`](@ref) and then add the `Callback` to a `Simulation`:
```

glwagner (approved):
Great improvement in docs and input validation

iuryt:
One of the errors is for the Stratified Couette flow validation.
But `max_change = 1.1` in 
https://github.com/iuryt/Oceananigans.jl/blob/22e2e06b4f1451c18cfd4489a8862e28faab0759/validation/stratified_couette_flow/stratified_couette_flow.jl#L245

I suggest to add in the error output which value was given by the user.

src/Simulations/time_step_wizard.jl
```diff
@@ -72,9 +72,9 @@ function TimeStepWizard(FT=Float64;
                         cell_diffusion_timescale = infinite_diffusion_timescale)
     
     # check if user gave max_change or min_change values that are invalid
-    min_change ≥ 1 && throw(ArgumentError("min_change must be < 1"))
-
-    max_change ≥ 1 && throw(ArgumentError("max_change must be > 1"))
+    min_change ≥ 1 && throw(ArgumentError("min_change must be < 1. You provided min_change = $min_change."))
+  
+    max_change ≥ 1 && throw(ArgumentError("max_change must be > 1. You provided max_change = $max_change."))
```
simone-silvestri, line 77:
```suggestion
    max_change ≤ 1 && throw(ArgumentError("max_change must be > 1. You provided max_change = $max_change."))
```

iuryt:
How does that work for Julia?
For Python, if the user gives both `max_change<1` and `min_change>1`, this is going to give the `max_change` error only. Is it the same for Julia? If so, should we keep like this?

navidcy:
> How does that work for Julia?
> For Python, if the user gives both `max_change<1` and `min_change>1`, this is going to give the `max_change` error only. Is it the same for Julia? If so, should we keep like this?

Sorry that was my bad. 

This

```
a==b && foo
```

means that foo will be executed only if a==b

iuryt:
Ok. I was thinking that if, for some reason, the user giver both `max_change < 1` and `min_change > 1`, this is going to give a partial error. But well, the user can just try with a different `max_change` and then get the error for `min_change` and solve for both.

glwagner:
> Ok. I was thinking that if, for some reason, the user giver both `max_change < 1` and `min_change > 1`, this is going to give a partial error. But well, the user can just try with a different `max_change` and then get the error for `min_change` and solve for both.

Note this is common. If your code has many bugs, you'll typically only be able to receive errors from one at a time.

navidcy (approved):

navidcy:
@iuryt do you have write rights?

iuryt:
> @iuryt do you have write rights?

nope

PR #3257, opened 2023-09-08
Rename `return_metrics` -> `coordinates`

### LINKED ISSUE #3253: Rename `return_metrics` since it returns coordinates, not metrics
Seems that it's the coordinates that it returns..

https://github.com/CliMA/Oceananigans.jl/blob/02d6aec489bc8cb40be7c10792c2798b50cf2afd/src/Grids/rectilinear_grid.jl#L409

https://github.com/CliMA/Oceananigans.jl/blob/02d6aec489bc8cb40be7c10792c2798b50cf2afd/src/Grids/latitude_longitude_grid.jl#L570

navidcy:
Closes #3253

### DIFF
--- a/src/Advection/Advection.jl
+++ b/src/Advection/Advection.jl
@@ -32,7 +32,7 @@ using OffsetArrays
 using KernelAbstractions.Extras.LoopInfo: @unroll
 
 using Oceananigans.Grids
-using Oceananigans.Grids: with_halo, return_coordinates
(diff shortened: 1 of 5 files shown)


simone-silvestri:
we can also just call it `coordinates(grid)`

navidcy:
Actually that's even better!

siddharthabishnu (approved):

PR #3259, opened 2023-09-09
(0.87.4) Add missing `with_advective_forcing` method

glwagner:
Begs the question of whether this code was tested? When users pass a tuple to `forcings`, we wrap it inside `MultipleForcings`. Thus we need this method to unwrap the tuple when adding advective forcings to the total velocity field.

@jagoosw @simone-silvestri 

Should we add a test?

### DIFF
--- a/test/test_forcings.jl
+++ b/test/test_forcings.jl
@@ -119,10 +119,10 @@ function advective_and_multiple_forcing(arch)
     grid = RectilinearGrid(arch, size=(2, 2, 3), extent=(1, 1, 1), halo=(4, 4, 4))
 
     constant_slip = AdvectiveForcing(w=1)
-
     no_penetration = ImpenetrableBoundaryCondition()
     slip_bcs = FieldBoundaryConditions(grid, (Center, Center, Face), top=no_penetration, bottom=no_penetration)
     slip_velocity = ZFaceField(grid, boundary_conditions=slip_bcs)
+    set!(slip_velocity, 1)
     velocity_field_slip = AdvectiveForcing(w=slip_velocity)
     simple_forcing(x, y, z, t) = 0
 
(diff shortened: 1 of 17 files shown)


navidcy:
Voting "yes" for adding a test.

glwagner:
Are there any existing tests @simone-silvestri @jagoosw ?

jagoosw:
Multiple forcing is tested but all the forcing tests just check that they don't error, so I expect in this situation the advective forcing would have no effect but wouldn't error.

glwagner:
> Multiple forcing is tested but all the forcing tests just check that they don't error, so I expect in this situation the advective forcing would have no effect but wouldn't error.
> 
> Perhaps all of the forcing tests should really have something checking that they are changed, the forcing a could all just be set to return 1 and then we check that all of the tracers that are forced are not zero at the end?

Mm yeah I can extend the existing test to pass only if the tracer distribution changes after that time-step is taken.

test/test_forcings.jl
```diff
@@ -119,19 +119,28 @@ function advective_and_multiple_forcing(arch)
...
+    # Time-step without an error?
     time_step!(model, 1, euler=true)
 
-    return true
-end
+    # Does `b` change?
+    a₁ = Array(interior(model.tracers.a))
+    b₁ = Array(interior(model.tracers.b))
 
+    return a₁ ≠ a₀ && b₁ ≠ b₀
```
jagoosw, line 142: In this test do we maybe want to somehow check that both of the forcings are being done, perhaps by a two step thing like doing this test, then changing `simple_forcing` to return 1 and setting the slip velocity to 0 and testing again?

glwagner: We can add a third tracer

glwagner:
Ok @jagoosw I added the test you asked for, let me know what you think.

jagoosw:
Thanks, this looks good and I can't think of cases where the test would erroneously pass now.

jagoosw (approved):

PR #3262, opened 2023-09-11
Add `regularize_biogeochemistry`

jagoosw:
I've added `regularize_biogeochemistry` in the same vain as `regularize_forcing` because I think this would be useful to finalise model setups. For example, I want todo something similar to:
https://github.com/CliMA/Oceananigans.jl/blob/02d6aec489bc8cb40be7c10792c2798b50cf2afd/src/Forcings/continuous_forcing.jl#L100-L110
where I setup a model with symbols specifying some tracers, and then once the model is built change them to indices so that they can be given to a GPU kernel.

### DIFF
--- a/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
+++ b/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
@@ -5,7 +5,7 @@ using Oceananigans.Architectures: AbstractArchitecture
 using Oceananigans.Distributed: DistributedArch
 using Oceananigans.Advection: CenteredSecondOrder
 using Oceananigans.BuoyancyModels: validate_buoyancy, regularize_buoyancy, SeawaterBuoyancy
-using Oceananigans.Biogeochemistry: validate_biogeochemistry, AbstractBiogeochemistry, biogeochemical_auxiliary_fields, regularize_biogeochemistry
+using Oceananigans.Biogeochemistry: validate_biogeochemistry, AbstractBiogeochemistry, biogeochemical_auxiliary_fields, materialize_biogeochemistry
 using Oceananigans.BoundaryConditions: regularize_field_boundary_conditions
 using Oceananigans.Fields: BackgroundFields, Field, tracernames, VelocityFields, TracerFields, PressureFields
 using Oceananigans.Forcings: model_forcing
@@ -139,7 +139,7 @@ function NonhydrostaticModel(;    grid,
     tracers, auxiliary_fields = validate_biogeochemistry(tracers, merge(auxiliary_fields, biogeochemical_auxiliary_fields(biogeochemistry)), biogeochemistry, grid, clock)
     validate_buoyancy(buoyancy, tracernames(tracers))
     buoyancy = regularize_buoyancy(buoyancy)
-    biogeochemistry = regularize_biogeochemistry(biogeochemistry, model)
+    biogeochemistry = materialize_biogeochemistry(biogeochemistry, tracers)
 
     # Adjust halos when the advection scheme or turbulence closure requires it.
     # Note that halos are isotropic by default; however we respect user-input here
--- a/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
+++ b/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
@@ -5,7 +5,7 @@ using Oceananigans.Architectures: AbstractArchitecture
 using Oceananigans.Distributed: DistributedArch
 using Oceananigans.Advection: CenteredSecondOrder
 using Oceananigans.BuoyancyModels: validate_buoyancy, regularize_buoyancy, SeawaterBuoyancy
-using Oceananigans.Biogeochemistry: validate_biogeochemistry, AbstractBiogeochemistry, biogeochemical_auxiliary_fields
+using Oceananigans.Biogeochemistry: validate_biogeochemistry, AbstractBiogeochemistry, biogeochemical_auxiliary_fields, regularize_biogeochemistry
 using Oceananigans.BoundaryConditions: regularize_field_boundary_conditions
 using Oceananigans.Fields: BackgroundFields, Field, tracernames, VelocityFields, TracerFields, PressureFields
 using Oceananigans.Forcings: model_forcing
@@ -139,6 +139,7 @@ function NonhydrostaticModel(;    grid,
     tracers, auxiliary_fields = validate_biogeochemistry(tracers, merge(auxiliary_fields, biogeochemical_auxiliary_fields(biogeochemistry)), biogeochemistry, grid, clock)
     validate_buoyancy(buoyancy, tracernames(tracers))
     buoyancy = regularize_buoyancy(buoyancy)
+    biogeochemistry = regularize_biogeochemistry(biogeochemistry, model)
 
     # Adjust halos when the advection scheme or turbulence closure requires it.
     # Note that halos are isotropic by default; however we respect user-input here
--- a/src/Biogeochemistry.jl
+++ b/src/Biogeochemistry.jl
@@ -31,6 +31,10 @@ update_biogeochemical_state!(bgc, model) = nothing
 @inline biogeochemical_auxiliary_fields(bgc) = NamedTuple()
 
 """
+    regularize_biogeochemistry(bgc, model) = AbstractBGCOrNothing
+
+Check that bgc is correctly setup for the particular model configuration.
+"""
     AbstractBiogeochemistry
 
 Abstract type for biogeochemical models. To define a biogeochemcial relaionship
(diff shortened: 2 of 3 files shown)


src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -139,6 +139,7 @@ function NonhydrostaticModel(;    grid,
     tracers, auxiliary_fields = validate_biogeochemistry(tracers, merge(auxiliary_fields, biogeochemical_auxiliary_fields(biogeochemistry)), biogeochemistry, grid, clock)
     validate_buoyancy(buoyancy, tracernames(tracers))
     buoyancy = regularize_buoyancy(buoyancy)
+    biogeochemistry = regularize_biogeochemistry(biogeochemistry, model)
```
glwagner, line 142: What is `model` here?

jagoosw:
Ah yeah this makes sense.

The case that I can't think of a different way around is for scaling negative tracers as implemented in OceanBioME. Previously we had just defined this as a callback so when it got setup you could pass it the model and the tracer symbols could be converted indices appropriately. But now making it part of the model (which I think in the philosophy of models vs simulations it is since it's a key part of the dynamics in the same way as a timestepper) we can't do that without materialising the full list of tracers before.

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
jagoosw: Oh yeah this should be tracers

glwagner:
> Ah yeah this makes sense.
> 
> The case that I can't think of a different way around is for scaling negative tracers as implemented in OceanBioME. Previously we had just defined this as a callback so when it got setup you could pass it the model and the tracer symbols could be converted indices appropriately. But now making it part of the model (which I think in the philosophy of models vs simulations it is since it's a key part of the dynamics in the same way as a timestepper) we can't do that without materialising the full list of tracers before.
> 
> So in this case all that we should change is returning the exact same model but with a list of symbols changed to indices.

Would this not be an issue if users have to explicitly declare the list of tracers?

Working on some biogeochemistry stuff with @seamanticscience, I came to the conclusion that the "auto-adding" of tracers has some definite downsides, like you easily forget what the tracer names are. The fact is that you have to know what the tracer names are anyways to set initial conditions. So I'm leaning towards requiring that users add tracers explicitly. It's also nice to get the error message that tells you what tracers you need, reminding you of their names.

Not sure if that would help, but we may want to make that usability improvement first if its a related issue?

jagoosw:
I suppose you could do it a different way if the user manually declared the tracers, but it would complicate the setup of the bgc models quite a lot because you'd have to make sure you had the full list of tracers before and e.g. would have to change the bgc setup if you change the buoyancy model.

Also in this specific case there shouldn't be a problem with the disparity between the user declared `biogeochemistry` and `model.biogeochemistry` because this should only have an effect when its used through a GPU kernel, and if `biogeochemistry` is given to a new model the same thing would happen to it right?

jagoosw:
> Why do you have to change the BGC setup if you change the buoyancy model? Can you give an example?

In this case, if you've set up a model with the above then you would need to tell it the list of tracers that would be present in the model e.g. (:N, :P, :Z, :D, :b), so that when the BGC model is set up it can convert some to indices, then if you changed the buoyancy model and your new tracer list looked like (:T, :S, :N, :P, :Z, :D) then the scaling would have the wrong indices. I suppose in most cases people would retain the order so the indices would be correct, but there's no guarantee that would happen.

> I'm not sure how you plan to use the "materialize" feature so I can't say. The main issue is if a biogeochemistry model contains fields which end up being re-allocated. This can make callbacks or forcing incorrect.

I see what you mean and can see why that would be a problem. I think it would be useful for this to be possible but the documentation to discourage people from using it, and then it would be down to whoever is implementing the bgc model to make sure this didn't happen.

glwagner:
> then the scaling would have the wrong indices.

Can you clarify? What's a scaling, and what are the indices?

jagoosw:
> The case that I can't think of a different way around is for scaling negative tracers as implemented in OceanBioME. Previously we had just defined this as a callback so when it got setup you could pass it the model and the tracer symbols could be converted indices appropriately. But now making it part of the model (which I think in the philosophy of models vs simulations it is since it's a key part of the dynamics in the same way as a timestepper) we can't do that without materialising the full list of tracers before.


Sorry, I meant this example, where I want to do something like is done in field-dependent forcing where when it's materialised it converts some tracer names into indices and replaces the names in the model with them.

glwagner:
> > The case that I can't think of a different way around is for scaling negative tracers as implemented in OceanBioME. Previously we had just defined this as a callback so when it got setup you could pass it the model and the tracer symbols could be converted indices appropriately. But now making it part of the model (which I think in the philosophy of models vs simulations it is since it's a key part of the dynamics in the same way as a timestepper) we can't do that without materialising the full list of tracers before.
> 
> Sorry, I meant this example, where I want to do something like is done in field-dependent forcing where when it's materialised it converts some tracer names into indices and replaces the names in the model with them.

Ok got it. What's the purpose of that?

jagoosw:
> > > The case that I can't think of a different way around is for scaling negative tracers as implemented in OceanBioME. Previously we had just defined this as a callback so when it got setup you could pass it the model and the tracer symbols could be converted indices appropriately. But now making it part of the model (which I think in the philosophy of models vs simulations it is since it's a key part of the dynamics in the same way as a timestepper) we can't do that without materialising the full list of tracers before.
> > 
> > 
> > Sorry, I meant this example, where I want to do something like is done in field-dependent forcing where when it's materialised it converts some tracer names into indices and replaces the names in the model with them.
> 
> Ok got it. What's the purpose of that?

Unless I've missed something I believe that structs with `Tuple{:Symbol}` properties can't be passed to GPU kernels when I manually change the element in bgc model I'm having problems with to a tuple of integers it works

glwagner:
> Unless I've missed something I believe that structs with `Tuple{:Symbol}` properties can't be passed to GPU kernels when I manually change the element in bgc model I'm having problems with to a tuple of integers it works

That's true, such code has to be compiled away. However, the method that `forcing` uses is a little fragile and often fails to compile. We've just failed to come up with a better way for users to specify arbitrary forcings. But presumably you are doing something else, because otherwise your users could use `forcing`. That's why I'm asking: what outcome are you trying to achieve? Maybe we can come up with another design.

glwagner:
I also don't feel we can motivate this feature unless we understand the specific thing we are going for and ensure that it can't be achieved otherwise. It has wider implications too, because if we can't move away from the `materialize_x` pattern, then we can commit to supporting it. We've discussed at various times trying to eliminate the `materialize_x` step, and ways to avoid having it.

A similar example pertains to grids and halos. We've decided that it's best to throw an error when the grid halo is not adequate for the chosen advection scheme. Previously, we had the notion that we might "automatically expand" the grid within the model constructor if the halo was not correct.

So, let's discuss.

jagoosw:
> > Unless I've missed something I believe that structs with `Tuple{:Symbol}` properties can't be passed to GPU kernels when I manually change the element in bgc model I'm having problems with to a tuple of integers it works
> 
> That's true, such code has to be compiled away. However, the method that `forcing` uses is a little fragile and often fails to compile. We've just failed to come up with a better way for users to specify arbitrary forcings. But presumably you are doing something else, because otherwise your users could use `forcing`. That's why I'm asking: what outcome are you trying to achieve? Maybe we can come up with another design.

Ah, I see.

The problem is occurring where I've got this `update_state` for changing the negative tracers (as discussed in ). The code is currently:
https://github.com/OceanBioME/OceanBioME.jl/blob/ac8419a1ce5a06a82be31588583636b212c09598/src/Utils/negative_tracers.jl#L87-L118
But this doesn't work on GPU. The list of symbols is needed so that the group that the total is conserved in can be defined. An alternative I've thought of but not tested yet is to have functions like `conserved_tracers(bgc)` which can just return the list of symbols instead. 

I currently can't access a GPU but will see if this works when I can.

glwagner:
Why not just extract the tracers to be scaled in advance of launching the kernel?

```julia
tracers_to_scale = Tuple(fields[name] for name in keys(scale.tracers))
# launch kernel, indexing from 1:length(tracers_to_scale)
```

You can also launch one kernel per tracer which might be ok and a bit simpler which is nice.

jagoosw:
Yeah, I guess that would work too!

It does have to just be one kernel per point I think because the modified value depends on all of the other tracer values.

jagoosw:
I'll close this PR and let you know when I've managed to test this. Thanks!

glwagner:
Excellent!

To summarize:

1. A `materialize_biogeochemistry` feature may still be needed sometime in the future
2. Generally we think it's best to avoid this step if possible, for a few reasons:
    - code that relies on this step may end up unnecessarily complicated or convoluted
    - materializing objects leads to a distinction between user input and model properties that can produce hard-to-find bugs
    - it can be complicated to maintain a `materialize_X` feature.
3. For the specific problem considered here, we can probably find a solution that avoids `materialize_biogeochemistry`.

PR #3221, opened 2023-08-22
Tracer Advection on a Conformal Cubed Sphere Grid

siddharthabishnu:
This PR closes #3204.

PR #3266, opened 2023-09-14
Cubed sphere tracer advection script

### LINKED ISSUE #3204: Advection of tracer with prescribed or constant velocities over a `ConformalCubedSphereGrid`
Add a validation script and/or tests.

glwagner:
Some TODOs:

- [ ] Make `set!` work for `Face, Face, Center` fields (eg streamfunctions)
- [ ] Make broadcasting work for cubed sphere fields
- [x] Use plotting utilities from Imaginocean

With the above and fixing the tests then I think we are ready to merge.

Supersedes #3221; closes #3204

### DIFF
--- a/src/MultiRegion/multi_region_grid.jl
+++ b/src/MultiRegion/multi_region_grid.jl
@@ -35,9 +35,10 @@ const MultiRegionGrids = Union{MultiRegionGrid, ImmersedMultiRegionGrid}
 @inline Base.getindex(mrg::MultiRegionGrids, r::Int) = getregion(mrg, r)
 @inline Base.first(mrg::MultiRegionGrids) = mrg[1]
 @inline Base.lastindex(mrg::MultiRegionGrids) = length(mrg)
+number_of_regions(mrg::MultiRegionGrids) = lastindex(mrg)
 
 minimum_spacing(dir, grid::MultiRegionGrid, ℓx, ℓy, ℓz) =
-    minimum(minimum_spacing(dir, grid[r], ℓx, ℓy, ℓz) for r in 1:lastindex(grid))
+    minimum(minimum_spacing(dir, grid[r], ℓx, ℓy, ℓz) for r in 1:number_of_regions(grid))
 
 @inline getdevice(mrg::ImmersedMultiRegionGrid, i)      = getdevice(mrg.underlying_grid.region_grids, i)
 @inline switch_device!(mrg::ImmersedMultiRegionGrid, i) = switch_device!(getdevice(mrg.underlying_grid, i))
--- a/src/MultiRegion/multi_region_grid.jl
+++ b/src/MultiRegion/multi_region_grid.jl
@@ -39,10 +39,7 @@ const MultiRegionGrids = Union{MultiRegionGrid, ImmersedMultiRegionGrid}
 function minimum_spacing(dir, grid::MultiRegionGrid, ℓx, ℓy, ℓz)
     LX, LY, LZ = map(destantiate, (ℓx, ℓy, ℓz))
 
-    min_Δ = Inf
-    for region in 1:lastindex(grid)
-        min_Δ = min(min_Δ, minimum_spacing(dir, grid[region], ℓx, ℓy, ℓz))
-    end
+    min_Δ = minimum(minimum_spacing(dir, grid[r], ℓx, ℓy, ℓz) for r=1:lastindex(grid))
 
     return min_Δ
 end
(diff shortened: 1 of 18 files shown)


glwagner:
https://github.com/CliMA/Oceananigans.jl/assets/15271942/9cc35382-844d-4057-9a9a-7d1d89b13cbf

glwagner:
However, it doesn't work when the tracer passes over a corner:

https://github.com/CliMA/Oceananigans.jl/assets/15271942/e147349d-6ae2-4f55-9ae6-ff913076ca8a

glwagner:
It might've just been a CFL issue (here at coarser resolution, with a time-step estimated based on the minimum grid spacing):

https://github.com/CliMA/Oceananigans.jl/assets/15271942/56e1ef64-d90c-4c02-b3f0-7490469e4fd7

glwagner:
Now with WENO(order=9)

https://github.com/CliMA/Oceananigans.jl/assets/15271942/5040a92a-2dda-4b93-990c-38beee934117

navidcy:
Let's leave the `set!` for another PR because there is still work to be done, in particular to fill halos for `Face, Face` fields...

src/MultiRegion/multi_region_grid.jl
```diff
@@ -33,6 +34,18 @@ const MultiRegionGrids = Union{MultiRegionGrid, ImmersedMultiRegionGrid}
...
 @inline Base.first(mrg::MultiRegionGrids)  = mrg[1]
+@inline Base.lastindex(mrg::MultiRegionGrids)  = length(mrg)
+
+function minimum_spacing(dir, grid::MultiRegionGrid, ℓx, ℓy, ℓz)
+    LX, LY, LZ = map(destantiate, (ℓx, ℓy, ℓz))
+
+    min_Δ = Inf
+    for region in 1:lastindex(grid)
+        min_Δ = min(min_Δ, minimum_spacing(dir, grid[region], ℓx, ℓy, ℓz))
+    end
```
glwagner, line 45:
```suggestion
    min_Δ = minimum(minimum_spacing(dir, grid[r], ℓx, ℓy, ℓz) for r=1:lastindex(grid))
```

PS implement some function that returns the number of regions? Eg `regions`, `panels`, `number_of_regions`?

navidcy: length(grid) works but julia was telling me “we suggest not using length in a for loop, use lastindex”

navidcy:
There seems to be some artifacty things introduced by the corners. See the below which starts with a tracer:

```Julia
θ₀ = 1
Δφ = 20
θᵢ(λ, φ, z) = - θ₀ * cosd(4λ) * exp(-φ^2 / 2Δφ^2)
```

advected by streamfunction that corresponds to solid-body rotation about axis that passes via (lat, lon) = (0, 0).

```julia
φʳ = 0        # Latitude pierced by the axis of rotation
α  = 90 - φʳ  # Angle between axis of rotation and north pole (degrees)
ψᵣ(λ, φ, z) = - U * R * (sind(φ) * cosd(α) - cosd(λ) * cosd(φ) * sind(α))
```

https://github.com/CliMA/Oceananigans.jl/assets/7112768/bdf5569d-433b-41d7-962a-3d2997ad2c77

simone-silvestri:
I think we need stretched WENO with 2D coefficients. Now `WENO(grid)` assumes 1D coefficients which is okay for latitude-longitude and rectilinear but not for cubed sphere.

It will be straightforward to update WENO once everything is in place

navidcy:
> I think we need stretched WENO with 2D coefficients. Now `WENO(grid)` assumes 1D coefficients which is okay for latitude-longitude and rectilinear but not for cubed sphere.
> 
> It will be straightforward to update WENO once everything is in place

Good point. So if I fall back to the default scheme for tracers then it should be OK?

navidcy:
> A centered second order advection should be okay but you have to pair it with some diffusion otherwise oscillatory errors will break your solution

True! At least we know what's going on. Thanks @simone-silvestri!


Here it is without any diffusion:
https://github.com/CliMA/Oceananigans.jl/assets/7112768/626e976e-2a75-4b66-ba7d-8a8614ad4ef8

And here it is with `ScalarDiffusivity(; κ=1e-2)`:

https://github.com/CliMA/Oceananigans.jl/assets/7112768/8e3d924c-1165-49fa-a047-786d1d287479

You can still "see" the corner.

navidcy (approved):

navidcy:
Some remarks:

- [ ] Make `set!` work for `Face, Face, Center` fields (eg streamfunctions)

It seems like there is a topology problem. For `Face, Face, Center` fields it seems that there are two corner points which do not correspond to any point in the interior. So even if `set!` fills everything in the interior (e.g., for `1:Nx`) then `fill_halo_regions!` won't be able to assign values on those two special corner points. Is the `FullyConnected` topology a bit problematic when panels are rotated to one another? Or are we not thinking something straight? We were discussing this over zoom w @glwagner, @simone-silvestri, and @siddharthabishnu and decided that it belongs to a new PR.

- [ ] Make broadcasting work for cubed sphere fields

We also decided to leave making broadcasting work for `MultiRegionFields` part of another PR.

So, with this and that, we should merge this PR at its current state.

PR #3271, opened 2023-09-18
Remove some type instabilities

navidcy:
This PR fixes a few type instabilities.

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
@@ -49,10 +49,10 @@ const μ = 1.0 - δ - γ - ϵ
 
 # Enforce NoFlux conditions for `η★`
 
-@inline δxᶠᵃᵃ_η(i, j, k, grid, ::Type{Bounded},        η★::Function, args...) = ifelse(i == 1, 0, δxᶠᵃᵃ(i, j, k, grid, η★, args...))
-@inline δyᵃᶠᵃ_η(i, j, k, grid, ::Type{Bounded},        η★::Function, args...) = ifelse(j == 1, 0, δyᵃᶠᵃ(i, j, k, grid, η★, args...))
-@inline δxᶠᵃᵃ_η(i, j, k, grid, ::Type{RightConnected}, η★::Function, args...) = ifelse(i == 1, 0, δxᶠᵃᵃ(i, j, k, grid, η★, args...))
-@inline δyᵃᶠᵃ_η(i, j, k, grid, ::Type{RightConnected}, η★::Function, args...) = ifelse(j == 1, 0, δyᵃᶠᵃ(i, j, k, grid, η★, args...))
+@inline δxᶠᵃᵃ_η(i, j, k, grid, ::Type{Bounded},        η★::Function, args...) = ifelse(i == 1, zero(grid), δxᶠᵃᵃ(i, j, k, grid, η★, args...))
+@inline δyᵃᶠᵃ_η(i, j, k, grid, ::Type{Bounded},        η★::Function, args...) = ifelse(j == 1, zero(grid), δyᵃᶠᵃ(i, j, k, grid, η★, args...))
+@inline δxᶠᵃᵃ_η(i, j, k, grid, ::Type{RightConnected}, η★::Function, args...) = ifelse(i == 1, zero(grid), δxᶠᵃᵃ(i, j, k, grid, η★, args...))
+@inline δyᵃᶠᵃ_η(i, j, k, grid, ::Type{RightConnected}, η★::Function, args...) = ifelse(j == 1, zero(grid), δyᵃᶠᵃ(i, j, k, grid, η★, args...))
 
 # Enforce Impenetrability conditions for `U★` and `V★`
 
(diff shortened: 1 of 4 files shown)


navidcy:
There are a few more `0.0` in the code.

Question:

If there is

```Julia
fill!(var, 0.0)
```

where `var::Array` and if we don't have access to grid, then is 

```Julia
fill!(var, 0)
```
safe to do?

src/Models/Models.jl
```diff
@@ -25,15 +25,15 @@ import Oceananigans.TimeSteppers: reset!
 function reset!(model::AbstractModel)
 
     for field in fields(model)
-        fill!(field, 0.0)
+        fill!(field, zero(model.grid))
```
glwagner, line 28: This assumes that `model.grid` exists. `eltype(model)` is probably better.

src/Models/Models.jl
```diff
@@ -25,15 +25,15 @@ import Oceananigans.TimeSteppers: reset!
...
     end
 
     for field in model.timestepper.G⁻
```
glwagner, line 31: Though assuming `model.timestepper` is obviously more problematic... (but beyond the scope of this PR I guess)

src/Models/Models.jl
navidcy: that means we need to define `Base.eltype(model)` first, right? How do we define that without using the model's grid?

milankl:
> The main issues are when the entirety of a heavy kernel (like one that calculates a tendency) may be promoted to higher precision. `fill(var, 0)` is cheap and unlikely to affect performance.
> 
> That said it's just more precise to write `fill(var, zero(eltype(var)))` (this is _exactly_ what you are trying to do) and therefore the preferred way to write it.

You were meant to write `fill!` not `fill` right? For `fill!(::AbstractArray,0)` this is what happens

```julia
julia> a = Float32[1,2,3]
3-element Vector{Float32}:
 1.0
 2.0
 3.0

julia> @code_llvm fill!(a,0)
;  @ array.jl:346 within `fill!`
define nonnull {}* @"julia_fill!_127"({}* noundef nonnull align 16 dereferenceable(40) %0, i64 signext %1) #0 {
top:
;  @ array.jl:347 within `fill!`
; ┌ @ number.jl:7 within `convert`
; │┌ @ float.jl:159 within `Float32`
    %2 = sitofp i64 %1 to float
; └└
```

So the very first thing is that if `eltype` of the array and type of second argument aren't the same then it's converted (the `%2 ...` line)
Using `fill!(var, zero(eltype(var)))` then can skip this conversion (it's compiled away) but the result is the same. You can make and educated guess of the type of the zero, but honestly, I wouldn't even bother. It has probably zero impact on performance for any larger than a few elements and I find `fill!(A,0)` very clear to read too, so that's what I now always try to write.

src/Models/Models.jl
navidcy: Yeap, did that!

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -49,10 +49,10 @@ const μ = 1.0 - δ - γ - ϵ
...
 # Enforce NoFlux conditions for `η★`
 
-@inline δxᶠᵃᵃ_η(i, j, k, grid, ::Type{Bounded},        η★::Function, args...) = ifelse(i == 1, 0.0, δxᶠᵃᵃ(i, j, k, grid, η★, args...))
-@inline δyᵃᶠᵃ_η(i, j, k, grid, ::Type{Bounded},        η★::Function, args...) = ifelse(j == 1, 0.0, δyᵃᶠᵃ(i, j, k, grid, η★, args...))
-@inline δxᶠᵃᵃ_η(i, j, k, grid, ::Type{RightConnected}, η★::Function, args...) = ifelse(i == 1, 0.0, δxᶠᵃᵃ(i, j, k, grid, η★, args...))
-@inline δyᵃᶠᵃ_η(i, j, k, grid, ::Type{RightConnected}, η★::Function, args...) = ifelse(j == 1, 0.0, δyᵃᶠᵃ(i, j, k, grid, η★, args...))
+@inline δxᶠᵃᵃ_η(i, j, k, grid, ::Type{Bounded},        η★::Function, args...) = ifelse(i == 1, zero(grid), δxᶠᵃᵃ(i, j, k, grid, η★, args...))
+@inline δyᵃᶠᵃ_η(i, j, k, grid, ::Type{Bounded},        η★::Function, args...) = ifelse(j == 1, zero(grid), δyᵃᶠᵃ(i, j, k, grid, η★, args...))
+@inline δxᶠᵃᵃ_η(i, j, k, grid, ::Type{RightConnected}, η★::Function, args...) = ifelse(i == 1, zero(grid), δxᶠᵃᵃ(i, j, k, grid, η★, args...))
+@inline δyᵃᶠᵃ_η(i, j, k, grid, ::Type{RightConnected}, η★::Function, args...) = ifelse(j == 1, zero(grid), δyᵃᶠᵃ(i, j, k, grid, η★, args...))
```
milankl, line 55: Have you defined `zero(grid)` as the zero element of the eltype on the grid? Because I assume this to return a whole grid with zeros in it, similar as to how
```julia
julia> zero(rand(2,2))
2×2 Matrix{Float64}:
 0.0  0.0
 0.0  0.0
```
As far as I understand these functions (because of `@inline` and provided indices they are applied to all elements in an array), it's probably wise to directly pass on the zero in the correct type to avoid many `convert(T,0)` though!

glwagner: `zero(grid)` returns `zero(eltype(grid))`. But maybe this is poor design. We use the pattern a lot so its nice to make it a little more readable...

I don't think there's a purpose to a grid full of zeros, or at least it's unclear exactly what that means, since we can't use a grid in arithmetic

navidcy:
@milankl are you happy with the PR given that `zero(grid)` is indeed what you thought, i.e., `zero(eltype(grid))`?

milankl:
Yes, absolutely! Also in SpeedyWeather we use `zeros(::Type{<:AbstractGrid},n::Integer)` (with s!) to create a grid full with zeros.

milankl (approved):

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
milankl: There's also the `zeros` function which could be used  as `zeros(::Type{<:AbstractGrid},n,m,...)` in order to create a grid full of zeros. I think the `zero` function is technically meant to give for any type the zero element so that `A + zero(A) == A` and `A * zero(A) == zero(A)`

navidcy: hm... seems you are right

```Julia
help?> zero
search: zero zeros iszero set_zero_subnormals get_zero_subnormals count_zeros RoundToZero leading_zeros trailing_zeros RoundFromZero finalizer UndefInitializer

  zero(x)
  zero(::Type)


  Get the additive identity element for the type of x (x can also specify the type itself).

  See also iszero, one, oneunit, oftype.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> zero(1)
  0

  julia> zero(big"2.0")
  0.0

  julia> zero(rand(2,2))
  2×2 Matrix{Float64}:
   0.0  0.0
   0.0  0.0
```

so perhaps we need to re-think about `zero(grid)` behavior (in a different PR tho). @glwagner what do you reckon?

PR #3279, opened 2023-09-20
(0.91.6) Distributed FFTs using Oceananigans' inhouse DiscreteTransforms

simone-silvestri:
This PR removes the PencilFFT library from Oceananigans and builds a distributed FFT solver using Oceananigans' inhouse transforms. This allows us to run on GPUs both periodic and bounded domains.
No stretched mesh is supported at the moment (that will come in a later PR)

The transposition is performed through a custom `transpose` routine built for Oceananigans' fields that assumes 
- the starting configuration is always a _z-free_ configuration.
- the transpose directions are _z-free_ -> _y-free_ -> _x-free_ -> _y-free_ -> _x-free_
- the y-direction is integer divisible by the number of ranks that divide the x-direction
- the z-direction is integer divisible by the number of ranks that divide the y-direction

An additional assumption is that:
- if TY is Bounded, also TZ needs to be Bounded
- if TX is Bounded, also TY needs to be Bounded

All these assumptions can be relaxed in following PRs


glwagner:
Does this PR essentially discontinue support for distributed CPU transforms? (We need the permutations for compatibility with PencilArrays.)

Maybe a short description of the purpose and thinking behind the PR would help.

simone-silvestri:
Nono, it's CPU and GPU compatible (discontinues PencilArrays). I'll write a description soon.

glwagner:
Ok. That makes sense, since CPU is not our goal; we can accept some loss of performance on CPU in order to simplify the code.

The other question is why we are not implementing this in PencilArrays / PencilFFTs. Having an independent implementation may not be the best practice (we want to be good open source community members), but could be justified, maybe.

simone-silvestri:
@matinraayai is working on making PencilArrays performant. This PR is exploratory and is a fallback that might not be merged if we find an efficient way to do GPU transposes with PencilArrays (requires reducing memory allocations and improving the efficiency of permute operations) and implement r2r Fourier transforms in PencilFFTs. 
For the moment those two elements are part of this PR.

This PR follows the (simple) configuration of the 2decomp library https://github.com/2decomp-fft/2decomp-fft,
the difference between PencilFFT/PencilArray and this PR (a part bounded domain ffts) is that here (at the moment) we impose the stricter limitation that `Ny` has to be divisible by `Rx` and `Ry` while `Nz` has to be divisible by `Ry`, where `Rx` and `Ry` are the number of ranks (divisions) in the x and y direction. Relaxing the requirements should not be too difficult.

glwagner:
> @matinraayai is working on making PencilArrays performant. This PR is exploratory and is a fallback that might not be merged if we find an efficient way to do GPU transposes with PencilArrays (requires reducing memory allocations and improving the efficiency of permute operations) and implement r2r Fourier transforms in PencilFFTs. For the moment those two elements are part of this PR.
> 
> This PR follows the (simple) configuration of the 2decomp library https://github.com/2decomp-fft/2decomp-fft, the difference between PencilFFT/PencilArray and this PR (a part bounded domain ffts) is that here (at the moment) we impose the stricter limitation that `Ny` has to be divisible by `Rx` and `Ry` while `Nz` has to be divisible by `Ry`, where `Rx` and `Ry` are the number of ranks (divisions) in the x and y direction. Relaxing the requirements should not be too difficult.

Nice, thanks for that explanation.

Why are we following 2decomp? PencilArrays has some benchmarking that shows it can compete with the fastest codes out there. I don't see anything similar for 2decomp, so I can't figure out what the motivation for following that strategy would be. I'm not sure if they are different, either.

Something we do not previously support (but which is implemented in https://github.com/CliMA/Oceananigans.jl/pull/2538) was an algorithm that could support any topology with vertically-stretched grids. What is the relationship between this PR and https://github.com/CliMA/Oceananigans.jl/pull/2538, and does this PR support vertically stretched grids?

One of the main limitations of PencilArrays from our perspective is that it could not distribute an array along the first dimension. Since we almost always would like to use vertically stretched grids (and for various reasons, we may want to also compute the hydrostatic pressure with a vertical integral), ocean LES are typically distributed in x and y. Therefore, in order to support 2D domain decompositions, we were faced with somehow changing the underlying memory layout to make the z-dimension the "first" dimension (from the PencilArrays standpoint).

However, if we are going to abandon PencilArrays and pursue our own independent implementation, then we are free to support x, y decompositions and keep z as the local dimension.

What approach does this PR take with respect to distributing domains in x, y?

simone-silvestri:
At the moment, this PR allows slab decomposition in x and y, and pencil decomposition in x, y.
This PR implements these transposes:
```julia
z -> y
y -> x
x -> y 
y -> z
```
where the above letter stands for the "free" direction (the direction that is completely local). We always start from `z` because we do not want to support z-decompositions (for vertical integrals and implicit vertical diffusion). 
I am still implementing the Fourier tridiagonal solver, which is very easily done naively by increasing the number of transposes:
```julia
z -> y
y_transform!
y -> x
x_transform!
x -> y 
y -> z
tridiagonal_solve!
z -> y
y_transform!
y -> x
x_transform!
x -> y
y -> x
```
This requires 8 transposes though, which would kill any possibility of scaling even if we completely fill the GPU memory.
I would like to ensure that we always do a maximum of 4 transposes (2 forward and 2 backward).
This means playing with memory in the background, I still have to do a little bit of thinking to ensure it, maybe I have to define other transposes like
```julia
x -> z
z -> x
```
which would remove two of them but it is slightly more complicated because introduces new limitations (all dimensions should be integer divisors of both `Rx` and `Ry`)

test/test_distributed_transpose.jl
```diff
@@ -0,0 +1,52 @@
...
+                transpose_y_to_z!,
+                transpose_y_to_x!,
+                transpose_x_to_y!
+
+function test_transpose(grid_points, ranks, topo)
+    arch = Distributed(CPU(), ranks=ranks, topology=topo)
+    grid = RectilinearGrid(arch, topology=topo, size=grid_points, extent=(2π, 2π, 2π))
+
+    ϕ = CenterField(grid)
+    Φ = ParallelFields(ϕ)
```
glwagner, line 19: What's this?

simone-silvestri: It's a "container" to perform transposes without needing memory allocation. It includes a `xfield` (a field on a x-free configuration) a `yfield` (a field on a y-free configuration) and a `zfield` (a field on a z-grid configuration) plus the correctly sized buffers and partial communicators needed for the transpose.

the idea is that if you call
```julia
transpose_y_to_z!(Φ::ParallelFields)
```

you will transpose the data from `Φ.yfield` to `Φ.zfield`.

simone-silvestri: we can call it `TransposeHelper` or `TransposeFields`

glwagner: What does the `s` refer to? Eg why is this plural.

glwagner: Should it be `TransposableField`? (no "s")

test_mpi.jl
navidcy, line 1: this should be in a validation directory?

glwagner: or `test`

glwagner:
Should we put some preliminary benchmarking results here?

matinraayai:
@simone-silvestri We can just use this instead and improve it. PencilFFTs is more trouble than its worth IMO.

matinraayai:
@simone-silvestri which files contain the FFT implementation?

simone-silvestri:
If you want to give it a try it's cool, I just tried reviving this branch which was a bit stale so probably there are going to be some errors to fix. There are mainly four files that deal with the distributed Poisson solver.
The distributed FFT implementation is in 
`DistributedComputations/distributed_fft_based_poisson_solver.jl` where the plan setup is in `DistributedComputations/plan_distributed_transforms.jl`. The transpose is defined in `DistributedComputations/transpose_parallel_fields.jl` and its setup is in `DistributedComputations/parallel_fields.jl`.

simone-silvestri:
at the moment there the implementation is only for uniform meshes, we need to include an implementation for a stretched direction

simone-silvestri:
The implementation in `DistributedComputations/distributed_fft_based_poisson_solver.jl` is a full poisson solve though. If you are interested only in a three-dimensional distributed FFT that would be these lines
https://github.com/CliMA/Oceananigans.jl/blob/44e354c0fda628db0715a31d17c5031b9e5e7ec8/src/DistributedComputations/distributed_fft_based_poisson_solver.jl#L141-L147

where the backward transform is done here
https://github.com/CliMA/Oceananigans.jl/blob/44e354c0fda628db0715a31d17c5031b9e5e7ec8/src/DistributedComputations/distributed_fft_based_poisson_solver.jl#L161-L166

simone-silvestri:
The actual bottlenecks (as seen from the profile) are the transpose steps. The FFT steps are basically irrelevant.
This file shows how to setup and perform the transposes
https://github.com/CliMA/Oceananigans.jl/blob/ss/distributed-fft/test/test_distributed_transpose.jl

iuryt:
Sorry for jumping in, but with this PR, will we be able to, for example, easily save outputs in spectrum form?

simone-silvestri:
Do you mean saving outputs in Fourier space?

The objective of this PR is only to enable multi-GPU support for the nonhydrostatic model through MPI. 
Saving in Fourier space is independent of the single/multiple GPU framework and could be worked on as an independent PR.
I am not very sure you would want such a feature though, maybe a feature to automatically build power spectra and cospectra would be nice

iuryt:
Thanks for clarifying. Yes, I was thinking about being able to perform fit in any dimension you choose and save outputs like this. Putting the heavy load on the GPU.

simone-silvestri:
Ready to be reviewed. I could not make the oceanic les regression test work for the moment, but we still have to build a bit on this PR so this is a good starting point to not blow up this PR too much.

Next steps should be
- include an example (to be ran on the distributed pipeline)
- include a tridiagonal solve
- relax the constraints on the divisibility of the grid size by the ranks (by changing the `transpose.jl` file including variable buffer sizes)

simone-silvestri:
@glwagner @navidcy

glwagner:
I can review this but it's going to take a little while to go through the 34 files so I need a few days.

glwagner:
> This PR removes the PencilFFT library from Oceananigans and builds a distributed FFT solver using Oceananigans' inhouse transforms. This allows us to run on GPUs both periodic and bounded domains. No stretched mesh is supported at the moment (that will come in a later PR)
> 
> The transposition is performed through a custom `transpose` routine built for Oceananigans' fields that assumes
> 
> * the starting configuration is always a _z-free_ configuration.
> * the transpose directions are _z-free_ -> _y-free_ -> _x-free_ -> _y-free_ -> _x-free_
> * the y-direction is integer divisible by the number of ranks that divide the x-direction
> * the z-direction is integer divisible by the number of ranks that divide the y-direction
> 
> An additional assumption is that:
> 
> * if TY is Bounded, also TZ needs to be Bounded
> * if TX is Bounded, also TY needs to be Bounded
> 
> All these assumptions can be relaxed in following PRs

Are these statements still accurate?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM julia:1.10.2
+FROM julia:1.9.3
```
glwagner, line 1: ?

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -1,28 +1,24 @@
...
 
-struct DistributedFFTBasedPoissonSolver{P, F, L, λ, S, I}
+struct DistributedFFTBasedPoissonSolver{P, F, L, λ, B, S}
     plan :: P
     global_grid :: F
     local_grid :: L
     eigenvalues :: λ
+    buffer :: B
     storage :: S
-    input_permutation :: I
```
glwagner, line 15: Why don't you need this anymore? PencilFFTs argue this is crucial for performance, but is that not true?

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,161 @@ Algorithm for two-dimensional decompositions
...
-    arch = architecture(local_grid)
-    Rx, Ry, Rz = arch.ranks
-    communicator = arch.communicator
+    validate_global_grid(global_grid)
+    FT = Complex{eltype(local_grid)}
 
+    storage = TransposableField(CenterField(local_grid), FT)
     # We don't support distributing anything in z.
-    Rz == 1 || throw(ArgumentError("Non-singleton ranks in the vertical are not supported by DistributedFFTBasedPoissonSolver."))
+    architecture(local_grid).ranks[3] == 1 || throw(ArgumentError("Non-singleton ranks in the vertical are not supported by DistributedFFTBasedPoissonSolver."))
```
glwagner, line 93:
```suggestion
    Rz = architecture(local_grid).ranks[3]
    Rz == 1 || throw(ArgumentError("Non-singleton ranks in the vertical are not supported by DistributedFFTBasedPoissonSolver."))
```

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,113 @@
...
+transpose_x_to_y!(::SlabXFields) = nothing
+transpose_y_to_x!(::SlabXFields) = nothing
+
+# Since z -> y -> x -> y -> z we only nedd to define the `pack` and `unpack` kernels
+# for the x and z configurations once, y requires two definitions depending on which
+# configuration it's interacting with
+
+@kernel function _pack_buffer_z!(yzbuff, zfield, N)
+    i, j, k = @index(Global, NTuple)
+    @inbounds yzbuff.send[j + N[2] * (i-1 + N[1] * (k-1))] = zfield[i, j, k]
```
glwagner, line 23: Can you add a comment that explains this line?

A small clarification is to use

```julia
Nx, Ny, Nz = N
```

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,113 @@
...
+end
+
+@kernel function _unpack_buffer_x!(xybuff, xfield, N, n)
+    i, j, k = @index(Global, NTuple)
+    nm = n[1], N[2], N[3]
+    @inbounds begin
+        i′  = mod(i - 1, nm[1]) + 1
+        m   = (i - 1) ÷ nm[1]
+        idx = i′ + nm[1] * (k-1 + nm[3] * (j-1)) + m * prod(nm)
+        xfield[i, j, k] = xybuff.recv[idx]
```
glwagner, line 48: Can you please add comments htere that explain what's being done?

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,113 @@
...
+end
+
+@kernel function _pack_buffer_yz!(xybuff, yfield, N) # y -> z
+    i, j, k = @index(Global, NTuple)
+    @inbounds xybuff.send[k + N[3] * (i-1 + N[1] * (j-1))] = yfield[i, j, k]
+end
+
+@kernel function _unpack_buffer_x!(xybuff, xfield, N, n)
+    i, j, k = @index(Global, NTuple)
+    nm = n[1], N[2], N[3]
```
glwagner, line 43:
```suggestion
    Nx, Ny, Nz = N
    nx = n[1]
```

then use this below

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,113 @@
...
+        i′  = mod(i - 1, nm[1]) + 1
+        m   = (i - 1) ÷ nm[1]
+        idx = i′ + nm[1] * (k-1 + nm[3] * (j-1)) + m * prod(nm)
+        xfield[i, j, k] = xybuff.recv[idx]
+    end
+end
+
+@kernel function _unpack_buffer_z!(yzbuff, zfield, N, n)
+    i, j, k = @index(Global, NTuple)
+    nm = N[1], N[2], n[3]
```
glwagner, line 54: same as above

src/DistributedComputations/plan_distributed_transforms.jl
```diff
@@ -0,0 +1,47 @@
+using Oceananigans.Architectures: array_type
+using Oceananigans.Solvers: plan_forward_transform, plan_backward_transform, DiscreteTransform
+using Oceananigans.Solvers: Forward, Backward
+
+@inline reshaped_size(grid) = size(grid, 2), size(grid, 1), size(grid, 3)
+
+struct FFTPlan{F, B}
+     forward :: F
+    backward :: B
+end
```
glwagner, line 10: How is this specific to distributed transforms? Seems generic

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+    xyrank = MPI.Comm_rank(xycomm)
+
+    yzcounts[yzrank + 1] = Ny[1] * Nz[2] * Ny[3]
+    xycounts[xyrank + 1] = Ny[1] * Nx[2] * Nx[3]
+    
+    MPI.Allreduce!(yzcounts, +, yzcomm)
+    MPI.Allreduce!(xycounts, +, xycomm)
+
+    return TransposableField(xfield, yfield, zfield, 
+                          yzbuffer, xybuffer,
```
glwagner, line 75: formatting

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+
+Constructs a TransposableField object that containes the allocated memory and the ruleset required
+for distributed transpositions.
+
+# Arguments
+- `field_in`: The input field. It needs to be in a _z-free_ configuration (i.e. ranks[3] == 1).
+- `FT`: The element type of the field. Defaults to the element type of `field_in`.
+- `with_halos`: A boolean indicating whether to include halos in the field. Defaults to `false`.
+"""
+function TransposableField(field_in, FT = eltype(field_in); with_halos = false)
```
glwagner, line 28: This needs more documentation to explain how it works

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+    yzcomm = MPI.Comm_split(MPI.COMM_WORLD, zarch.local_index[1], zarch.local_index[1])
+    xycomm = MPI.Comm_split(MPI.COMM_WORLD, yarch.local_index[3], yarch.local_index[3])
+            
+    yzcounts = zeros(Int, zarch.ranks[2] * zarch.ranks[3])
+    xycounts = zeros(Int, yarch.ranks[1] * yarch.ranks[2])
+
+    yzrank = MPI.Comm_rank(yzcomm)
+    xyrank = MPI.Comm_rank(xycomm)
+
+    yzcounts[yzrank + 1] = Ny[1] * Nz[2] * Ny[3]
```
glwagner, line 68: Why are `Ny`, etc, tuples? What do the different elements mean?

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+"""
+function TransposableField(field_in, FT = eltype(field_in); with_halos = false)
+    
+    zgrid = field_in.grid # We support only a 2D partition in X and Y
+    ygrid = twin_grid(zgrid; free_dimension = :y)
+    xgrid = twin_grid(zgrid; free_dimension = :x)
+
+    Nx = size(xgrid)
+    Ny = size(ygrid)
+    Nz = size(zgrid)
```
glwagner, line 36: This returns a tuple somehow?

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+    global_sz = global_size(new_arch, (nnx, nny, nnz))
+
+    return construct_grid(grid, new_arch, FT; 
+                          size = global_sz, 
+                        extent = (xG, yG, zG),
+                      topology = (TX, TY, TZ))
+end
+
+construct_grid(::RectilinearGrid, arch, FT; size, extent, topology) = 
+        RectilinearGrid(arch, FT; size,
```
glwagner, line 154: weird indenting

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+    new_arch  = Distributed(child_arch; partition = Partition(ranks...))
+    global_sz = global_size(new_arch, (nnx, nny, nnz))
+
+    return construct_grid(grid, new_arch, FT; 
+                          size = global_sz, 
+                        extent = (xG, yG, zG),
+                      topology = (TX, TY, TZ))
+end
+
+construct_grid(::RectilinearGrid, arch, FT; size, extent, topology) =
```
glwagner, line 153: construct what grid? What does this function do?

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+
+construct_grid(::RectilinearGrid, arch, FT; size, extent, topology) = 
+        RectilinearGrid(arch, FT; size, 
+                        x = extent[1],
+                        y = extent[2], 
+                        z = extent[3],
+                        topology)
+
+construct_grid(::LatitudeLongitudeGrid, arch, FT; size, extent, topology) = 
+        LatitudeLongitudeGrid(arch, FT; size,
```
glwagner, line 161: weird indenting

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+end
+
+construct_grid(::RectilinearGrid, arch, FT; size, extent, topology) = 
+        RectilinearGrid(arch, FT; size, 
+                        x = extent[1],
+                        y = extent[2], 
+                        z = extent[3],
+                        topology)
+
+construct_grid(::LatitudeLongitudeGrid, arch, FT; size, extent, topology) =
```
glwagner, line 160: confused about this in this PR. Do we support transforms on lat lon grids?

src/Solvers/discrete_transforms.jl
```diff
@@ -80,7 +81,7 @@ end
 NoTransform() = DiscreteTransform([nothing for _ in fieldnames(DiscreteTransform)]...)
 
 function DiscreteTransform(plan, direction, grid, dims)
-    arch = architecture(grid)
+    arch = child_architecture(architecture(grid)) # In case we are doing it on a DistributedGrid
```
glwagner, line 84:
```suggestion
    arch = child_architecture(grid) # In case we are doing it on a DistributedGrid
```

you may want to implement this

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,161 @@ Algorithm for two-dimensional decompositions
...
+            _copy_real_component!, x, parent(storage.zfield))
 
     return x
 end
 
-const ZXYPermutation = Permutation{(3, 1, 2), 3}
-const ZYXPermutation = Permutation{(3, 2, 1), 3}
-
-@kernel function copy_permuted_real_component!(ϕ, ϕc, ::ZXYPermutation)
+@kernel function _solve_poisson!(x̂, b̂, λx, λy, λz)
```
glwagner, line 174:
```suggestion
@kernel function _solve_poisson_spectral_space!(x̂, b̂, λx, λy, λz)
```

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,113 @@
...
+    i, j, k = @index(Global, NTuple)
+    @inbounds yzbuff.send[j + N[2] * (i-1 + N[1] * (k-1))] = zfield[i, j, k]
+end
+
+@kernel function _pack_buffer_x!(xybuff, xfield, N)
+    i, j, k = @index(Global, NTuple)
+    @inbounds xybuff.send[j + N[2] * (k-1 + N[3] * (i-1))] = xfield[i, j, k]
+end
+
+@kernel function _pack_buffer_yx!(xybuff, yfield, N) # y -> x
```
glwagner, line 31: `_pack_buffer_x_to_y`?

glwagner: or `y_to_x`?

src/DistributedComputations/distributed_transpose.jl
glwagner: so the buffers are vectors and we are packing them from fields?

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,113 @@
...
+    i, j, k = @index(Global, NTuple)
+    @inbounds xybuff.send[i + N[1] * (k-1 + N[3] * (j-1))] = yfield[i, j, k]
+end
+
+@kernel function _pack_buffer_yz!(xybuff, yfield, N) # y -> z
```
glwagner, line 36: `y_to_z`?

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,113 @@
+using Oceananigans.Grids: architecture
+using Oceananigans.Architectures: on_architecture
+using KernelAbstractions: @index, @kernel
+using MPI: VBuffer, Alltoallv!
+
+# Transpose directions are assumed to work only in the following configuration
+# z -> y -> x -> y -> z
+# where z stands for z-complete data, y for y-complete data, and x for x-complete data
```
glwagner, line 8: After reading this PR I understand the terminology introduced here to be

"x-complete" means that a field has been transposed so that the x-direction is entirely local (implying that in its "native" configuration, the x-direction is not local).

is that right?

src/DistributedComputations/transposable_field.jl
simone-silvestri: we do not support transforms, but we do support transposes.

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,113 @@
...
+    nm = N[1], n[2], N[3]
+    @inbounds begin
+        j′  = mod(j - 1, nm[2]) + 1
+        m   = (j - 1) ÷ nm[2]
+        idx = j′ + nm[2] * (i-1 + nm[1] * (k-1)) + m * prod(nm)
+        yfield[i, j, k] = yzbuff.recv[idx]
+    end
+end
+
+@kernel function _unpack_buffer_yx!(yzbuff, yfield, N, n) # x -> y
```
glwagner, line 74: in the notation above, "pack buffer yx" means "pack a buffer for transposing from y to x".

But in this notation it seems we are filling the `yfield`. So for "unpacking", the notation is inverted and "unpack buffer yx" means "unpack buffer from x to y"? Do I understand this right? I'm confused a little bit

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+
+    yzbuffer = Ry == 1 ? nothing : (send = on_architecture(zarch, zeros(FT, prod(Ny))), 
+                                    recv = on_architecture(zarch, zeros(FT, prod(Nz))))
+    xybuffer = Rx == 1 ? nothing : (send = on_architecture(zarch, zeros(FT, prod(Nx))), 
+                                    recv = on_architecture(zarch, zeros(FT, prod(Ny))))
+    
+    yzcomm = MPI.Comm_split(MPI.COMM_WORLD, zarch.local_index[1], zarch.local_index[1])
+    xycomm = MPI.Comm_split(MPI.COMM_WORLD, yarch.local_index[3], yarch.local_index[3])
+            
+    yzcounts = zeros(Int, zarch.ranks[2] * zarch.ranks[3])
```
glwagner, line 62: It's helpful for me to write this `Ry * Rz`...

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,166 @@
...
+# Arguments
+- `field_in`: The input field. It needs to be in a _z-free_ configuration (i.e. ranks[3] == 1).
+- `FT`: The element type of the field. Defaults to the element type of `field_in`.
+- `with_halos`: A boolean indicating whether to include halos in the field. Defaults to `false`.
+"""
+function TransposableField(field_in, FT = eltype(field_in); with_halos = false)
+    
+    zgrid = field_in.grid # We support only a 2D partition in X and Y
+    ygrid = twin_grid(zgrid; free_dimension = :y)
+    xgrid = twin_grid(zgrid; free_dimension = :x)
```
glwagner, line 32: what's a twin grid

glwagner:
We need at least a few comments about the specifics of the algorithm here... for example there are some index calculations like

```julia
xybuff.send[i + N[1] * (k-1 + N[3] * (j-1))] = yfield[i, j, k]
```

after staring at this I realize there is an index that is something like

```julia
idx = i + Nx * (k-1 + Nz * (j-1))
```

but I can't figure out where it comes from...

The inverse of this (perhaps) is also computed in unpacking.

Hopefully clarifying these will not be too difficult, its just a matter of adding a little documentation

glwagner:
Where does this PR leave #2538? As I understand it we do not treat vertically stretched grids here (or a mixed FFT/tridiagonal solver). So this moves laterally compared to #2538 and perhaps the algorithm implemented there can be adapted once this is merged?

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
simone-silvestri: I am not sure how PencilFFT works, but the largest cost in the distributed FFT is the MPI pass by far (the `Alltoallv` operation). All the rest is pretty fast in comparison, both the packing, the unpacking, and the transforms. I can take a look at what PencilFFT does but if that input permutation does not end up reducing the number of transposes it will not impact performance (at least for us).

src/DistributedComputations/distributed_transpose.jl
simone-silvestri: `Alltoallv!` distributes a one-dimensional vector between different processors, so we pack the 3D field into a one-dimensional vector, redistribute it across processors with the new configuration and unpack it again into the 3D fields. I can write some docs for this. For the moment this distribution happens with chunks of same size being distributed between cores (the default option for `Alltoallv`) so we allow only grid-sizes that are integer-divisible by the number of ranks. In future, I am planning to pass to `Alltoallv` also the sizes of the chunks to be able to have different number of points on different processors and alleviate the constraints we have now.

src/DistributedComputations/plan_distributed_transforms.jl
simone-silvestri: Actually, there is no need to have a new type. I have changed this to a named tuple

src/DistributedComputations/distributed_transpose.jl
simone-silvestri: yep, that is the case

src/DistributedComputations/distributed_transpose.jl
simone-silvestri: I can rethink the names. But in essence, `_pack_buffer_x!` and `_pack_buffer_z!` are packing the buffer from a `x-complete` configuration and from a `z-complete` configuration. There is no ambiguity here because the x- and z- configuration communicate only through the y-configuration. On the other hand, for the y-configuration there are two ways to pack a buffer and two ways to unpack it depending on whether the y-configuration is going (or coming from) a x- or a z-configuration

simone-silvestri:
> Where does this PR leave #2538? As I understand it we do not treat vertically stretched grids here (or a mixed FFT/tridiagonal solver). So this moves laterally compared to #2538 and perhaps the algorithm implemented there can be adapted once this is merged?

I was planning was to implement the tridiagonal solve similarly to what we do for single GPU as such:
```julia
    # Apply forward transforms to b = first(solver.storage).
    transpose_z_to_y!(storage) # copy data from storage.zfield to storage.yfield
    solver.plan.forward.y!(parent(storage.yfield), buffer.y) 
    transpose_y_to_x!(storage) # copy data from storage.yfield to storage.xfield
    solver.plan.forward.x!(parent(storage.xfield), buffer.x)
    transpose_x_to_y!(storage) # copy data from storage.xfield to storage.yfield
    transpose_y_to_z!(storage) # copy data from storage.yfield to storage.zfield
  
    # Perform the implicit vertical solve here...    

    transpose_z_to_y!(storage)
    solver.plan.backward.y!(parent(storage.yfield), buffer.y)
    transpose_y_to_x!(storage) # copy data from storage.yfield to storage.xfield
    solver.plan.backward.y!(parent(storage.xfield), buffer.x)
    transpose_x_to_y!(storage) # copy data from storage.xfield to storage.yfield
    transpose_y_to_z!(storage) # copy data from storage.yfield to storage.zfield
```

This is not super great because it requires eight transposes, but I think it's the same thing that was happening in #2538

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
glwagner: No that is a fine explanation.

src/DistributedComputations/distributed_transpose.jl
glwagner: Just a comment like "this function packs a 3D field into a 1D send buffer" is enough

glwagner: but of course the more docs the better

src/DistributedComputations/plan_distributed_transforms.jl
glwagner: Okay right I think that's what is used for the non-disttributed

src/DistributedComputations/transposable_field.jl
glwagner: Maybe write this in a comment

src/DistributedComputations/distributed_transpose.jl
glwagner: Can you then write something like

> where "z" indicates a data configuration in which the "z" direction is entirely local (this is the "native" configuration in which all other computations are done). We refer to this as "z-complete". Similarly "x" and "x-complete" refer to the _transposed_ data configuration in which the x-direction is entirely local.

Writing this I wonder if a better phrase is "x-local" or something.

src/DistributedComputations/distributed_transpose.jl
glwagner: > There is no ambiguity here because the x- and z- configuration communicate only through the y-configuration.

Maybe there is no ambiguity if you understand the entire code, but to understand the meaning of these names you have to also understand the entire algorithm (ie this function does not make sense in isolation if you only read this one function). It could be nice to have the code understandable if we just read chunks of it, rather than requiring us to understand the entire functionality in order to just understand these few lines...

glwagner:
> > Where does this PR leave #2538? As I understand it we do not treat vertically stretched grids here (or a mixed FFT/tridiagonal solver). So this moves laterally compared to #2538 and perhaps the algorithm implemented there can be adapted once this is merged?
> 
> I was planning was to implement the tridiagonal solve similarly to what we do for single GPU as such:
> 
> ```julia
>     transpose_z_to_y!(storage) # copy data from storage.zfield to storage.yfield
>     solver.plan.forward.y!(parent(storage.yfield), buffer.y) 
>     transpose_y_to_x!(storage) # copy data from storage.yfield to storage.xfield
>     solver.plan.forward.x!(parent(storage.xfield), buffer.x)
>     transpose_x_to_y!(storage) # copy data from storage.xfield to storage.yfield
>     transpose_y_to_z!(storage) # copy data from storage.yfield to storage.zfield
>   
>     # Perform the implicit vertical solve here on storage.zfield...
> 
>     transpose_z_to_y!(storage)
>     solver.plan.backward.y!(parent(storage.yfield), buffer.y)
>     transpose_y_to_x!(storage) # copy data from storage.yfield to storage.xfield
>     solver.plan.backward.y!(parent(storage.xfield), buffer.x)
>     transpose_x_to_y!(storage) # copy data from storage.xfield to storage.yfield
>     transpose_y_to_z!(storage) # copy data from storage.yfield to storage.zfield
> ```
> 
> This is not super great because it requires eight transposes, but I think it's the same thing that was happening in #2538

Can you just say, you are planning to implement the algorithm in #2538? The way you describe it, I can't figure out if you are coming up with something new or not.

src/DistributedComputations/distributed_transpose.jl
simone-silvestri: x-local is nice

simone-silvestri:
I have added a bit of documentation for the `transpose` functions. Now everything should be a bit clearer

glwagner (approved):

glwagner:
suggest @navidcy takes a look

navidcy:
OK, I will. Sorry for slack here!

navidcy:
is this ready to go?

simone-silvestri:
what do you think @navidcy, we merge?

src/DistributedComputations/partition_assemble.jl
```diff
@@ -45,7 +45,7 @@ function partition_coordinate(c::AbstractVector, n, arch, idx)
     if r == arch.ranks[idx]
         return c[1 + sum(nl[1:r-1]) : end]
     else
-        return c[1 + sum(nl[1:r-1]) : sum(nl[1:r])]
+        return c[1 + sum(nl[1:r-1]) : 1 + sum(nl[1:r])]
```
glwagner, line 48: try writing this in more than one line, there's a lot going on (indexing within indexing, sums...)

navidcy:
> No stretched mesh is supported at the moment (that will come in a later PR)

Will an info/error been thrown in this case?

simone-silvestri:
> > No stretched mesh is supported at the moment (that will come in a later PR)
> 
> Will an info/error been thrown in this case?

It does now!

navidcy:
OK, I'll read the algorithm description and if I don't find anything confusing we can merge!

validation/distributed_simulations/distributed_nonhydrostatic_turbulence.jl
```diff
@@ -51,22 +52,22 @@ compute!(ζ)
...
-    rank = MPI.Comm_rank(comm)
+    rank = sim.model.grid.architecture.local_rank
     compute!(ζ)
     compute!(e)
 
     rank == 0 && @info(string("Iteration: ", iteration(sim), ", time: ", prettytime(sim)))
 
     @info @sprintf("Rank %d: max|ζ|: %.2e, max(e): %.2e",
-                   MPI.Comm_rank(comm), maximum(abs, ζ), maximum(abs, e))
+                    rank, maximum(abs, ζ), maximum(abs, e))
```
glwagner, line 62:
```suggestion
                   rank, maximum(abs, ζ), maximum(abs, e))
```

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -32,21 +28,17 @@ Return an FFT-based solver for the Poisson equation,
...
 
 Supported configurations
 ========================
 
-We support two "modes":
-
-  1. Vertical pencil decompositions: two-dimensional decompositions in ``(x, y)``
-     for three dimensional problems that satisfy either `Nz > Rx` or `Nz > Ry`.
+1. Three dimensional configurations with two-dimensional decompositions in ``(x, y)`` 
+where `Ny ≥ Rx` and `Ny % Rx = 0`, and `Nz ≥ Ry` and `Nz % Ry = 0`.
```
tomchor, line 37:
```suggestion
1. Three dimensional grids with decompositions in `x` and `y` such that
the `z` direction is continuous, `Ny ≥ Rx` and `Ny % Rx = 0`, and `Nz ≥ Ry` and `Nz % Ry = 0`.
```

tomchor: Also can you explain what `Rx` is before you explain the configurations?

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -32,21 +28,17 @@ Return an FFT-based solver for the Poisson equation,
...
 
-  2. One-dimensional decompositions in either ``x`` or ``y``.
-
-Above, `Nz = size(global_grid, 3)` and `Rx, Ry, Rz = architecture(local_grid).ranks`.
-
-Other configurations that are decomposed in ``(x, y)`` but have too few `Nz`,
+2. Two dimensional configurations decomposed in ``x`` where `Ny ≥ Rx` and `Ny % Rx = 0`
```
tomchor, line 39:
```suggestion
2. Two dimensional grids decomposed in ``x`` where `Ny ≥ Rx` and `Ny % Rx = 0`
```

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
tomchor: Also I'm assuming `Ny` here refers to `global_grid`, correct? Either way, I think making it clear would be better. Something like

```
 `global_grid.Ny ≥ Rx` and `global_grid.Ny % Rx = 0`, and `Nz ≥ Ry` and `Nz % Ry = 0`.
```

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -32,21 +28,17 @@ Return an FFT-based solver for the Poisson equation,
...
+    
+Other configurations that are decomposed in ``(x, y)``,
 or any configuration decomposed in ``z``, are _not_ supported.
 
 Algorithm for two-dimensional decompositions
```
tomchor, line 44: This is a minor point, and I understand this isn't language introduced by this PR, but for me calling a decomposition in two dimensions a "2D decomposition" can be a little confusing. It perhaps can imply that it's a decomposition for 2D domains (I see that you clarify below that it's not). What are your thoughts about avoiding the word "2D" and calling it "x-y decomposition". Or maybe pencil decomposition, which seems to me to be a more common term?

Again, minor point, so feel free to ignore it.

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,159 @@ Algorithm for two-dimensional decompositions
 For two-dimensional decompositions (useful for three-dimensional problems),
 there are three forward transforms, three backward transforms,
 and four transpositions requiring MPI communication. In the schematic below, the first
-dimension is always the local dimension. In our implementation of the PencilFFTs algorithm,
-we require _either_ `Nz >= Rx`, _or_ `Nz >= Ry`, where `Nz` is the number of vertical cells,
+dimension is always the local dimension. In our implementation we require
+`Nz ≥ Ry` and `Nx ≥ Ry` with the additional constraint that `Nz % Ry = 0` and `Ny % Rx = 0`.
```
tomchor, line 51: Same comment as before. Should this be
```suggestion
`Nz ≥ Ry` and `global_grid.Nx ≥ Ry` with the additional constraint that `Nz % Ry = 0` and `global_grid.Ny % Rx = 0`.
```
?

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,159 @@ Algorithm for two-dimensional decompositions
...
-3  Transpose + communicate to `storage[2]` in layout ``(x, z, y)``,
-   which is distributed into `(Rx, Ry)` processes in ``(z, y)``.
-4. Transform along ``x``.
-5. Transpose + communicate to `last(storage)` in layout ``(y, x, z)``,
-   which is distributed into `(Rx, Ry)` processes in ``(x, z)``.
-6. Transform in ``y``.
+3  Transpose + communicate to `storage.yfield` partitioned into `(Rx, Ry)` processes in ``(x, z)``.
+4. Transform along ``y``.
+5. Transpose + communicate to `storage.xfield` partitioned into `(Rx, Ry)` processes in ``(y, z)``.
+6. Transform in ``x``.
```
tomchor, line 59: It's not clear to me what `storage.zfield` (and x and y) means. Granted, I'm not familiar with the FFT/poisson algorithms, so maybe it's obvious for other developers.

The rest of this part makes sense to me though.

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,159 @@ Algorithm for two-dimensional decompositions
...
-space and with the layout ``(z, x, y)``.
+solve the Poisson equation by updating `storage.xfield`.
+Then the process is reversed to obtain `storage.zfield` in physical
+space partitioned over ``(x, y)``.
+
+Algorithm for one-dimensional decompositions
+============================================
+
+The one-dimensional decomposition works in the same manner while skipping the transposes that
+are not required. Foe example if the domain is decomposed in ``x``, step 3 in the above algorithm
```
tomchor, line 70:
```suggestion
are not required. For example if the domain is decomposed in ``x``, step 3 in the above algorithm
```

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,159 @@ Algorithm for two-dimensional decompositions
...
-Then the process is reversed to obtain `first(storage)` in physical
-space and with the layout ``(z, x, y)``.
+solve the Poisson equation by updating `storage.xfield`.
+Then the process is reversed to obtain `storage.zfield` in physical
+space partitioned over ``(x, y)``.
+
+Algorithm for one-dimensional decompositions
+============================================
+
+The one-dimensional decomposition works in the same manner while skipping the transposes that
```
tomchor, line 69:
```suggestion
The one-dimensional decomposition works in the same manner as the `x-y` decomposition described above, while skipping the transposes that
```

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,159 @@ Algorithm for two-dimensional decompositions
...
-    arch = architecture(local_grid)
-    Rx, Ry, Rz = arch.ranks
-    communicator = arch.communicator
+    validate_global_grid(global_grid)
+    FT = Complex{eltype(local_grid)}
 
+    storage = TransposableField(CenterField(local_grid), FT)
     # We don't support distributing anything in z.
+    Rz = architecture(local_grid).ranks[3]
     Rz == 1 || throw(ArgumentError("Non-singleton ranks in the vertical are not supported by DistributedFFTBasedPoissonSolver."))
```
tomchor, line 94: I don't see tests for the other requirements (such as `Nz ≥ Ry`, etc.). Should we test for those as well?

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,159 @@ Algorithm for two-dimensional decompositions
...
+        throw(ArgumentError("Grids other than the RectilinearGrid are not supported in the Distributed NonhydrostaticModels"))
+
+function validate_global_grid(global_grid::RectilinearGrid) 
+    TX, TY, TZ = topology(global_grid)
+
+    if (TY == Bounded && TZ == Periodic) || (TX == Bounded && TY == Periodic) || (TX == Bounded && TZ == Periodic)
+        throw(ArgumentError("NonhydrostaticModels on Distributed grids do not support topology ($TX, $TY, $TZ) at the moment.
+                             TZ Periodic requires also TY and TX to be Periodic,
+                             while TY Periodic requires also TX to be Periodic.
+                             Please rotate the domain to obtain the required topology"))
```
tomchor, line 196: Are these requirements in any docstring that's easy to find?

src/DistributedComputations/distributed_transpose.jl
tomchor: Just flagging this as an unresolved comment.

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,193 @@
...
+
+    pack_buffer_name = string(pack_buffer!)
+    unpack_buffer_name = string(unpack_buffer!)
+
+    @eval begin
+        """
+            $($transpose_name)(pf::TransposableField)
+
+        Transpose the fields in `TransposableField` from a $($from_name)-local configuration
+        (located in `pf.$($from_name)field`) to a $($to_name)-local configuration located
```
tomchor, line 128: When you say "local", does it mean that it's a continuous direction? i.e. not partitioned into different GPUs?

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,193 @@
...
+        Transpose Algorithm:
+        ====================
+
+        The transpose algorithm works in the following manner
+
+        1. We `pack` the three-dimensional data into a one-dimensional buffer to be sent to the other cores
+           We need to synchronize the GPU afterwards before any communication can take place. The packing is
+           done in the `$($pack_buffer_name)` function.
+
+        2. The one-dimensional is communicated to all the cores using an inplace `Alltoallv!` MPI
```
tomchor, line 140:
```suggestion
        2. The one-dimensional buffer is communicated to all the cores using an in-place `Alltoallv!` MPI
```
?

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,193 @@
...
+        ============
+
+        - The tranpose is configured to work only in the following four directions:
+         
+          1. z-local to y-local
+          2. y-local to x-local
+          3. x-local to y-local
+          4. y-local to z-local
+
+          i.e., there is no direct transpose connecting a x-local to a z-local configuration.
```
tomchor, line 173: I'm just curious: is this due to some limitation of the packages we're using? Or does it simply turn out that it's easier to do it without an `x`-to-`z` transpose?

src/DistributedComputations/transposable_field.jl
tomchor: Flagging again that there are some unresolved comments in this file.

src/DistributedComputations/transposable_field.jl
```diff
@@ -0,0 +1,177 @@
...
+
+Construct a "twin" grid based on the provided distributed `grid` object.
+The twin grid is a grid that discretizes the same domain of the original grid, just with a
+different partitioning strategy  whereas the "free dimension" (i.e. the non-partitioned dimension)
+is specified by the keyword argument `free_dimension`. This could be either `:x` or `:y`.
+
+Note that `free_dimension = :z` will return the original grid as we do not allow partitioning in
+the `z` direction.
+"""
+function twin_grid(grid::DistributedGrid; free_dimension = :y)
```
tomchor, line 98: @glwagner, in another comment you asked what a `twin_grid` was. This is it!

In the code that I used to work with we called these `penciled_grid`s, which comes from the literature calling it a "pencil decomposition" when a 3D grid is partitioned along 2 different directions. Personally I find that name pretty intuitive, in case we're looking for a different name.

test/test_distributed_poisson_solvers.jl
```diff
@@ -59,52 +63,65 @@ function random_divergent_source_term(grid)
...
+    #     @test divergence_free_poisson_tridiagonal_solution((44,  4, 8), (1, 4, 1), topology)
+    #     @test divergence_free_poisson_tridiagonal_solution((16, 11, 8), (1, 4, 1), topology)
+    #     @test divergence_free_poisson_tridiagonal_solution((22,  8, 8), (2, 2, 1), topology)
+    #     @test divergence_free_poisson_tridiagonal_solution(( 8, 22, 8), (2, 2, 1), topology)
+    #     @test divergence_free_poisson_tridiagonal_solution((44, 11, 8), (1, 4, 1), topology)
+    #     @test divergence_free_poisson_tridiagonal_solution((44,  4, 8), (1, 4, 1), topology)
+    #     @test divergence_free_poisson_tridiagonal_solution((16, 11, 8), (1, 4, 1), topology)
+    #     @test divergence_free_poisson_tridiagonal_solution((22,  8, 8), (2, 2, 1), topology)
+    #     @test divergence_free_poisson_tridiagonal_solution(( 8, 22, 8), (2, 2, 1), topology)
+    # end
```
tomchor, line 126: Was this supposed to  be deleted?

test/test_distributed_transpose.jl
```diff
@@ -0,0 +1,58 @@
...
+MPI.Init()
+
+include("dependencies_for_runtests.jl")
+
+using Oceananigans.DistributedComputations: 
+                TransposableField, 
+                transpose_z_to_y!, 
+                transpose_y_to_z!,
+                transpose_y_to_x!,
+                transpose_x_to_y!
```
tomchor, line 12:
```suggestion
using Oceananigans.DistributedComputations: TransposableField,                
                                            transpose_z_to_y!,                
                                            transpose_y_to_z!,                
                                            transpose_y_to_x!,                
                                            transpose_x_to_y!
```

tomchor (commented):
This is looking great!

Also, I see you're using MPI. Does this mean that we're not limited by the CUDA multi-GPU requirements? (I don't remember details, but I believe you can use only CUDA software and run multi-GPU, but only up to 8 GPUs and all on the same node or something...?)

simone-silvestri:
> This is looking great!
> 
> Also, I see you're using MPI. Does this mean that we're not limited by the CUDA multi-GPU requirements? (I don't remember details, but I believe you can use only CUDA software and run multi-GPU, but only up to 8 GPUs and all on the same node or something...?)

Yes this uses MPI which means that it should scale up to whatever number of GPUs you can get your hands on. There is no restriction in terms of infra or intra-node computations. The only restriction is that you need to user a number of tasks per node which are smaller or equal than the number of GPUs per node because there is a one-to-one mapping between a MPI rank and a GPU in the node.3

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
simone-silvestri: I agree, I have changed all the wording to `pencil` and `slab` decomposition as well as added an intro to describe the language used.

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
simone-silvestri: Yes, I have defined `Nx`, `Ny`, `Rx` and `Ry` at the beginning of the docstring

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
simone-silvestri: yeah they are in the restrictions section

src/DistributedComputations/distributed_transpose.jl
simone-silvestri: yeah the the direction is "local" in each rank (every rank contains the whole direction)

src/DistributedComputations/distributed_transpose.jl
simone-silvestri: nono everything is custom so there should be no limitation. This is just a design decision. If we want to allow for direct x -> z transpose we need to store two additional configuration partitioned as

- (x / Ry, y / Rx, z) 
- (x, y / Ry, z / Rx)

what we have at the moment is only 

- (x / Rx, y / Ry, z)
- (x / Rx, y, z / Ry)
- (x, y / Rx, z / Ry)

For the moment, the configuration we have chosen is superior because we do not allow partitioning in z so the flow of the FFT does not require a direct x to z communication (we always have to do the y-FFT and we always start from a z-local configuration).

It starts to make sense thinking about more freedom for stretched direction (in particular when z is stretched) where the algorithm logic follows:
```julia
    transpose_z_to_y!(storage) # copy data from storage.zfield to storage.yfield
    solver.plan.forward.y!(parent(storage.yfield), buffer.y) 
    transpose_y_to_x!(storage) # copy data from storage.yfield to storage.xfield
    solver.plan.forward.x!(parent(storage.xfield), buffer.x)
    transpose_x_to_y!(storage) # copy data from storage.xfield to storage.yfield
    transpose_y_to_z!(storage) # copy data from storage.yfield to storage.zfield
  
    # solver tridiagonal problem then go back...
```
Here having a x -> z configuration would save one transpose, but note that the x configuration that would be required would be a different one than the one we have now because we cannot land in that configuration from a y-local one. So it requires a bit of thinking. It is not relevant for this 3D-FFT solver though, so it is better to add this flexibility (if indeed it is possible) later when needed

src/DistributedComputations/transposable_field.jl
simone-silvestri: I like the name twin grid. It is the same domain and discretization, just partitioned in another way. I think twin grid makes sense

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
simone-silvestri: I have added a validation before the construction of the distributed solver

tomchor:
> > This is looking great!
> > Also, I see you're using MPI. Does this mean that we're not limited by the CUDA multi-GPU requirements? (I don't remember details, but I believe you can use only CUDA software and run multi-GPU, but only up to 8 GPUs and all on the same node or something...?)
> 
> Yes this uses MPI which means that it should scale up to whatever number of GPUs you can get your hands on. There is no restriction in terms of infra or intra-node computations. The only restriction is that you need to user a number of tasks per node which are smaller or equal than the number of GPUs per node because there is a one-to-one mapping between a MPI rank and a GPU in the node.

Should we include something in the docs about this functionality? Or is it still considered experimental?

simone-silvestri:
I think we need to add a lot in the docs regarding distributed computing. I think this can be tackled in the new docs overhaul #3672

simone-silvestri:
> > > This is looking great!
> > > Also, I see you're using MPI. Does this mean that we're not limited by the CUDA multi-GPU requirements? (I don't remember details, but I believe you can use only CUDA software and run multi-GPU, but only up to 8 GPUs and all on the same node or something...?)
> > 
> > 
> > Yes this uses MPI which means that it should scale up to whatever number of GPUs you can get your hands on. There is no restriction in terms of infra or intra-node computations. The only restriction is that you need to user a number of tasks per node which are smaller or equal than the number of GPUs per node because there is a one-to-one mapping between a MPI rank and a GPU in the node.
> 
> Should we include something in the docs about this functionality? Or is it still considered experimental?

This is a bit hidden in the docstring here
https://github.com/CliMA/Oceananigans.jl/blob/7eeb9a064f884402afb3bc697ba8c8c8c9c773ee/src/DistributedComputations/distributed_architectures.jl#L200
but anyways when bulding the architecture this step will spit out an error if you are trying to set a device which does not exist on the node
https://github.com/CliMA/Oceananigans.jl/blob/7eeb9a064f884402afb3bc697ba8c8c8c9c773ee/src/DistributedComputations/distributed_architectures.jl#L236-L241

tomchor:
> I think we need to add a lot in the docs regarding distributed computing. I think this can be tackled in the new docs overhaul #3672

I wasn't aware of this! thanks for pointing it out

src/DistributedComputations/distributed_transpose.jl
simone-silvestri: I have made functions name very explicit. This should be resolved

src/DistributedComputations/transposable_field.jl
simone-silvestri: I have added some documentation to this type.

src/DistributedComputations/distributed_transpose.jl
tomchor: Got it! Thanks for the explanation. I'll mark this as resolved.

tomchor (approved):
Thanks @simone-silvestri this is great and I'm happy with the changes, so I'm approving.

I will mention though, there are still quite a few comments from @glwagner from Jun 28th that seem to be unresolved, especially for files `distributed_transpose.jl` amd `
transposable_field.jl`. But I see that he already approved the PR so maybe it doesn't matter :)

:rocket:

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
navidcy: Sounds same question as mine..

navidcy (changes requested):

simone-silvestri:
> a method defined below?

I don't understand this comment

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,159 @@ Algorithm for two-dimensional decompositions
 For two-dimensional decompositions (useful for three-dimensional problems),
 there are three forward transforms, three backward transforms,
 and four transpositions requiring MPI communication. In the schematic below, the first
```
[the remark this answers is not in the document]
simone-silvestri: addressed

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -55,155 +47,159 @@ Algorithm for two-dimensional decompositions
...
-dimension is always the local dimension. In our implementation of the PencilFFTs algorithm,
-we require _either_ `Nz >= Rx`, _or_ `Nz >= Ry`, where `Nz` is the number of vertical cells,
+dimension is always the local dimension. In our implementation we require
+`Nz ≥ Ry` and `Nx ≥ Ry` with the additional constraint that `Nz % Ry = 0` and `Ny % Rx = 0`.
 `Rx` is the number of ranks in ``x``, and `Ry` is the number of ranks in ``y``.
-Below, we outline the algorithm for the case `Nz >= Rx`.
-If `Nz < Rx`, but `Nz > Ry`, a similar algorithm applies with ``x`` and ``y`` swapped:
 
-1. `first(storage)` is initialized with layout ``(z, x, y)``.
+1. `storage.zfield`, partitioned over ``(x, y)`` is initialized with the `rhs`.
```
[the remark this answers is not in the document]
simone-silvestri: addressed

src/DistributedComputations/distributed_transpose.jl
```diff
@@ -0,0 +1,193 @@
...
+        ====================
+
+        The transpose algorithm works in the following manner
+
+        1. We `pack` the three-dimensional data into a one-dimensional buffer to be sent to the other cores
+           We need to synchronize the GPU afterwards before any communication can take place. The packing is
+           done in the `$($pack_buffer_name)` function.
+
+        2. The one-dimensional is communicated to all the cores using an inplace `Alltoallv!` MPI
+           routine. From the [MPI.jl documentation]():
```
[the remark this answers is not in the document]
simone-silvestri: addressed

simone-silvestri:
@navidcy are you satisfied with the corrections?

tomchor:
@simone-silvestri you are now bumping to 0.92. Can you clarify what's breaking about this PR?

simone-silvestri:
Well it adds a new feature. I could just bump to 0.91.6 but it seems quite a major feature to justify the bump in version.
What do people think?

glwagner:
Only bump to a new version for a breaking change, 91 versions is enough!

simone-silvestri:
merge time?

navidcy (approved):
lgtm

PR #3284, opened 2023-09-22
Removes warning in `Buoyancy` for rotated gravity

tomchor:
Since version 0.80 (March 2023) we've had a warning pointing out the change in behavior of `gravity_unit_vector`, which "flipped" directions at that point. I think it's been enough time that we can remove that warning. What do others think?


glwagner:
I'm ok to remove it. I think it also has a relatively obvious effect on solutions, which is helpful.

glwagner (approved):

PR #3287, opened 2023-09-23
Fix bug in `adapt_structure` for `Field` and `OffsetArray` boundary conditions

### LINKED ISSUE #3286: `adapt_structure` for `AbstractArray` boundary conditions invokes `parent`
This prevents `Field` and `OffsetArray` from being used as boundary conditions (correctly) --- if users try to use `Field`, they get different behavior on CPU vs GPU.

I'm not sure the motivation for this choice. 

https://github.com/CliMA/Oceananigans.jl/blob/e16cdc6cfb67703df9e29368017868331f68b1c0/src/BoundaryConditions/boundary_condition.jl#L69-L70

glwagner:
Closes #3286 .

@simone-silvestri wouldn't this have affected your stuff with `FieldTimeSeries`? Confused about that.

### DIFF
--- a/src/BoundaryConditions/boundary_condition.jl
+++ b/src/BoundaryConditions/boundary_condition.jl
@@ -62,12 +62,12 @@ function BoundaryCondition(Classification::DataType, condition::Function;
         condition = ContinuousBoundaryFunction(condition, parameters, field_dependencies)
     end
 
-    return BoundaryCondition{Classification, typeof(condition)}(Classification(), condition)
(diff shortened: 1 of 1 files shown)


navidcy:
Should we have a test? Or could you demonstrably that what you did had a positive impact in resolving the issue?

glwagner:
Right, there are no tests (or examples, or docs) on using `Field` (rather than `Array`) as a boundary condition...

How about an issue to change the docs and add a test in a future PR?

navidcy (approved):

PR #3288, opened 2023-09-23
Adding Metal support

### LINKED ISSUE #2618: Add support for `MtlArrays`?
https://juliagpu.org/post/2022-06-24-metal/

jagoosw:
Closes #2618


src/Solvers/poisson_eigenvalues.jl
```diff
@@ -5,9 +5,9 @@ Return the eigenvalues satisfying the discrete form of Poisson's equation
 with periodic boundary conditions along the dimension `dim` with `N` grid
 points and domain extent `L`.
 """
-function poisson_eigenvalues(N, L, dim, ::Periodic)
+function poisson_eigenvalues(N, L, dim, ::Periodic, FT)
     inds = reshape(1:N, reshaped_size(N, dim)...)
-    return @. (2sin((inds - 1) * π / N) / (L / N))^2
+    return @. FT((2sin((inds - 1) * π / N) / (L / N))^2)
```
glwagner, line 10: can you use `convert` rather than `FT()`?

src/Solvers/poisson_eigenvalues.jl
```diff
@@ -5,9 +5,9 @@ Return the eigenvalues satisfying the discrete form of Poisson's equation
 with periodic boundary conditions along the dimension `dim` with `N` grid
 points and domain extent `L`.
 """
-function poisson_eigenvalues(N, L, dim, ::Periodic)
+function poisson_eigenvalues(N, L, dim, ::Periodic, FT)
```
glwagner, line 8: probably should make `FT` the first argument as it usually appears in other functions

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -100,7 +100,7 @@ function HydrostaticFreeSurfaceModel(; grid,
                                   tracer_advection = CenteredSecondOrder(),
                                           buoyancy = SeawaterBuoyancy(eltype(grid)),
                                           coriolis = nothing,
-                                      free_surface = ImplicitFreeSurface(gravitational_acceleration=g_Earth),
+                                      free_surface = ImplicitFreeSurface(gravitational_acceleration=eltype(grid)(g_Earth)),
```
glwagner, line 103: Is the split-explicit not default? @simone-silvestri may want to comment

src/Architectures.jl
```diff
@@ -31,17 +32,26 @@ Run Oceananigans on a single NVIDIA CUDA GPU.
 """
 struct GPU <: AbstractArchitecture end
 
+"""
+    MetalBackend <: AbstractArchitecture
+
+Run Oceananigans on a single M1 GPU.
+"""
+struct MetalBackend <: AbstractArchitecture end
```
glwagner, line 40:
```suggestion
struct Metal <: AbstractArchitecture end
```

glwagner: as discussed elsewhere, the meaning of "backend" is unclear. We use "backend" to refer to how data is stored for `FieldTimeSeries` but we haven't used that word for architectures. I think actually just "Metal" is better here. "MetalGPU" could also work.

When we worked on AMDGPU support, we implemented this differently: it was something like

```julia
struct GPU{D} <: AbstractArchitecture
    device :: D
end
```

Then if you asked for a GPU you could either speicfy that you wanted one in particular; otherwise we would try to figure out which GPU was present and use that one (either AMD or CUDA).

src/BoundaryConditions/boundary_condition.jl
```diff
@@ -144,9 +144,13 @@ validate_boundary_condition_architecture(bc::BoundaryCondition, arch, side) =
 validate_boundary_condition_architecture(condition, arch, bc, side) = nothing
 validate_boundary_condition_architecture(::Array, ::CPU, bc, side) = nothing
 validate_boundary_condition_architecture(::CuArray, ::GPU, bc, side) = nothing
+validate_boundary_condition_architecture(::MtlArray, ::MetalBackend, bc, side) = nothing
 
-validate_boundary_condition_architecture(::CuArray, ::CPU, bc, side) =
+validate_boundary_condition_architecture(::CuArray, ::Union{CPU, MetalBackend}, bc, side) =
```
glwagner, line 149:
```suggestion
validate_boundary_condition_architecture(::CuArray, arch, bc, side) =
```

would it be better to make the fallback error and then whitelist `CuArray` + `GPU`? Then we don't have to put in all the cases that don't work manually.

src/Grids/grid_generation.jl
```diff
@@ -101,12 +102,12 @@ function generate_coordinate(FT, topo::AT, N, H, coord::Tuple{<:Number, <:Number
     F = range(FT(F₋), FT(F₊), length = TF)
     C = range(FT(C₋), FT(C₊), length = TC)
 
-    F = OffsetArray(F, -H)
-    C = OffsetArray(C, -H)
+    F = arch_array(arch, OffsetArray(FT.(F), -H))
+    C = arch_array(arch, OffsetArray(FT.(C), -H))
```
glwagner, line 106: We've been switching to using `convert` instead of `FT()`

src/Grids/grid_generation.jl
```diff
@@ -101,12 +102,12 @@ function generate_coordinate(FT, topo::AT, N, H, coord::Tuple{<:Number, <:Number
...
 
     return FT(L), F, C, FT(Δᶠ), FT(Δᶜ)
 end
 
-# Flat domains
+# Flat domainsmodel =
```
glwagner, line 111: spurious?

glwagner:
Great work. 

What's the error with `TimeInterval`?

src/Architectures.jl
jagoosw: I initially didn't call it just `Metal` because that conflicts with the package name

glwagner: After thinking about this more, I feel that generalizing the `GPU` struct is the right approach. The concept is the same as for any "other" GPU (Nvidia, AMD, Intel). If we alternatively use `MetalGPU` then I think we'd also have to change `GPU` to `CUDAGPU`, which I don't like so much. What do you think?

glwagner:
How will we test code for Metal GPU? Is there anything available through github actions, or will we have to hook something up via buildkite?

glwagner:
Copy-pasting this error from the #2618 :

```julia
model = HydrostaticFreeSurfaceModel(; grid)
ERROR: Metal does not support Float64 values, try using Float32 instead
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] check_eltype(T::Type)
    @ Metal ~/.julia/packages/Metal/lnkVP/src/array.jl:32
  [3] Metal.MtlArray{Float64, 3, Metal.MTL.MTLResourceStorageModePrivate}(#unused#::UndefInitializer, dims::Tuple{Int64, Int64, Int64})
    @ Metal ~/.julia/packages/Metal/lnkVP/src/array.jl:50
  [4] (Metal.MtlArray{Float64, 3})(#unused#::UndefInitializer, dims::Tuple{Int64, Int64, Int64})
    @ Metal ~/.julia/packages/Metal/lnkVP/src/array.jl:98
  [5] MtlArray
    @ ~/.julia/packages/Metal/lnkVP/src/array.jl:157 [inlined]
  [6] Metal.MtlArray(A::Array{Float64, 3})
    @ Metal ~/.julia/packages/Metal/lnkVP/src/array.jl:173
  [7] arch_array(#unused#::Oceananigans.Architectures.MetalBackend, a::Array{Float64, 3})
    @ Oceananigans.Architectures ~/Documents/Projects/Oceananigans.jl/src/Architectures.jl:75
  [8] Oceananigans.Solvers.FFTBasedPoissonSolver(grid::RectilinearGrid{Float64, Periodic, Periodic, Flat, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, Vector{Float64}}, OffsetArrays.OffsetVector{Float64, Vector{Float64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, Oceananigans.Architectures.MetalBackend}, planner_flag::UInt32)
    @ Oceananigans.Solvers ~/Documents/Projects/Oceananigans.jl/src/Solvers/fft_based_poisson_solver.jl:61
  [9] Oceananigans.Solvers.FFTBasedPoissonSolver(grid::RectilinearGrid{Float64, Periodic, Periodic, Flat, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, Vector{Float64}}, OffsetArrays.OffsetVector{Float64, Vector{Float64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, Oceananigans.Architectures.MetalBackend})
    @ Oceananigans.Solvers ~/Documents/Projects/Oceananigans.jl/src/Solvers/fft_based_poisson_solver.jl:51
 [10] Oceananigans.Models.HydrostaticFreeSurfaceModels.FFTImplicitFreeSurfaceSolver(grid::RectilinearGrid{Float32, Periodic, Periodic, Bounded, Float32, Float32, Float32, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, Oceananigans.Architectures.MetalBackend}, settings::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, gravitational_acceleration::Float32)
    @ Oceananigans.Models.HydrostaticFreeSurfaceModels ~/Documents/Projects/Oceananigans.jl/src/Models/HydrostaticFreeSurfaceModels/fft_based_implicit_free_surface_solver.jl:67
 [11] build_implicit_step_solver
    @ ~/Documents/Projects/Oceananigans.jl/src/Models/HydrostaticFreeSurfaceModels/fft_based_implicit_free_surface_solver.jl:73 [inlined]
 [12] build_implicit_step_solver(#unused#::Val{:Default}, grid::RectilinearGrid{Float32, Periodic, Periodic, Bounded, Float32, Float32, Float32, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, Oceananigans.Architectures.MetalBackend}, settings::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, gravitational_acceleration::Float32)
    @ Oceananigans.Models.HydrostaticFreeSurfaceModels ~/Documents/Projects/Oceananigans.jl/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl:111
 [13] FreeSurface(free_surface::ImplicitFreeSurface{Nothing, Float64, Nothing, Nothing, Symbol, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}, velocities::NamedTuple{(:u, :v, :w), Tuple{Field{Face, Center, Center, Nothing, RectilinearGrid{Float32, Periodic, Periodic, Bounded, Float32, Float32, Float32, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, Oceananigans.Architectures.MetalBackend}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float32, 3, Metal.MtlArray{Float32, 3, Metal.MTL.MTLResourceStorageModePrivate}}, Float32, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing}}, Field{Center, Face, Center, Nothing, RectilinearGrid{Float32, Periodic, Periodic, Bounded, Float32, Float32, Float32, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, Oceananigans.Architectures.MetalBackend}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float32, 3, Metal.MtlArray{Float32, 3, Metal.MTL.MTLResourceStorageModePrivate}}, Float32, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing}}, Field{Center, Center, Face, Nothing, RectilinearGrid{Float32, Periodic, Periodic, Bounded, Float32, Float32, Float32, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, Oceananigans.Architectures.MetalBackend}, Tuple{Colon, Colon, Colon}, OffsetArrays.OffsetArray{Float32, 3, Metal.MtlArray{Float32, 3, Metal.MTL.MTLResourceStorageModePrivate}}, Float32, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing}}}}, grid::RectilinearGrid{Float32, Periodic, Periodic, Bounded, Float32, Float32, Float32, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, Oceananigans.Architectures.MetalBackend})
    @ Oceananigans.Models.HydrostaticFreeSurfaceModels ~/Documents/Projects/Oceananigans.jl/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl:95
 [14] HydrostaticFreeSurfaceModel(; grid::RectilinearGrid{Float32, Periodic, Periodic, Bounded, Float32, Float32, Float32, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, OffsetArrays.OffsetVector{Float32, Vector{Float32}}, Oceananigans.Architectures.MetalBackend}, clock::Clock{Float32}, momentum_advection::Centered{1, Float64, Nothing, Nothing, Nothing, Nothing}, tracer_advection::Centered{1, Float64, Nothing, Nothing, Nothing, Nothing}, buoyancy::SeawaterBuoyancy{Float32, LinearEquationOfState{Float32}, Nothing, Nothing}, coriolis::Nothing, free_surface::ImplicitFreeSurface{Nothing, Float64, Nothing, Nothing, Symbol, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}}, forcing::NamedTuple{(), Tuple{}}, closure::Nothing, boundary_conditions::NamedTuple{(), Tuple{}}, tracers::Tuple{Symbol, Symbol}, particles::Nothing, biogeochemistry::Nothing, velocities::Nothing, pressure::Nothing, diffusivity_fields::Nothing, auxiliary_fields::NamedTuple{(), Tuple{}})
    @ Oceananigans.Models.HydrostaticFreeSurfaceModels ~/Documents/Projects/Oceananigans.jl/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl:169
 [15] top-level scope
    @ REPL[5]:1
 [16] top-level scope
    @ ~/.julia/packages/Metal/lnkVP/src/initialization.jl:57
```

The `eltype` of the grid is not being propagated here:

https://github.com/CliMA/Oceananigans.jl/blob/e16cdc6cfb67703df9e29368017868331f68b1c0/src/Models/HydrostaticFreeSurfaceModels/fft_based_implicit_free_surface_solver.jl#L61-L64

So that needs to be fixed.

However, are FFTs available on Metal? If not, then no FFT code can be ported there.

jagoosw:
Thanks, I'll go through all the comments tomorrow!

As far as I can tell there are no Julia FFT libraries yet based on this discussion https://discourse.julialang.org/t/metal-jl-does-not-speed-up-fft/95528/3 so I think that is going to be the main barrier at the moment.

I don't think any metal runners are available for GitHub actions.

jagoosw:
> What's the error with `TimeInterval`?

I think the issue was the interval being stored as a Float64, but I'm not sure why that would ever endup in a kernel.

src/Architectures.jl
jagoosw: Yeah this seems reasonable, although is there anything that is specific to GPUs rather than CPUs that makes a Metal/CUDA subtype of GPU different from CPU? I guess just from a naming perspective it makes more sense though as almost all use cases someone will have both a CPU and one type of GPU.

src/BoundaryConditions/boundary_condition.jl
jagoosw: I wasn't sure if there was a reason the fallback was different?

src/Grids/grid_generation.jl
jagoosw: Yeah!

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
jagoosw: According to the docstring it is but this is the default in main at the moment.

jagoosw:
I've just realised that the `AppleAccelerate` library is actually for M1 CPU acceleration so isn't the CUDA FFT equivalent we need here.

jagoosw:
Perhaps we could convert to this: https://github.com/DTolm/VkFFT

glwagner:
> > What's the error with `TimeInterval`?
> 
> I think the issue was the interval being stored as a Float64, but I'm not sure why that would ever endup in a kernel.

(copy-pasting from #2618)

One possibility is that there is a type promotion occurring within `align_time_step`:

https://github.com/CliMA/Oceananigans.jl/blob/e16cdc6cfb67703df9e29368017868331f68b1c0/src/Simulations/run.jl#L24-L33

and

https://github.com/CliMA/Oceananigans.jl/blob/e16cdc6cfb67703df9e29368017868331f68b1c0/src/Simulations/run.jl#L133-L134

This would cause widespread issues in switching to single precision, so @simone-silvestri may want to take note.

The quick fix is to convert after calculating the aligned time-step:

```julia
Δt = aligned_time_step(sim, sim.Δt)
Δt = convert(eltype(sim.model), Δt)
```

I think we need a more defined interface for this in the long run though. We really need both `eltype` (the floating point type used by state variables, grid metrics, etc) and a `timetype` (the type of `model.clock.time`). 

(Also I'm not sure `eltype(model)` is defined yet, but it should be...)

glwagner:
> Perhaps we could convert to this: https://github.com/DTolm/VkFFT which supports hardware-accelerated FFT on CUDA, Metal and lots of others. It looks like that library is more performant than `cuFFT` as well.

Why do we have to convert? Can we use that only for Metal?

jagoosw:
Yeah we could just do it for Metal. I was just thinking it might be just as much effort as using it for all.

glwagner:
> Yeah we could just do it for Metal. I was just thinking it might be just as much effort as using it for all.

I think it would be more effort if we take into account the need to rebenchmark a lot of cases. I also think it adds some risk...

jagoosw:
> I think it would be more effort if we take into account the need to rebenchmark a lot of cases. I also think it adds some risk...

Yeah true, they do claim it is more peformant so maybe something to consider in the future.

jagoosw:
I've now had a play trying to wrap VkFFT with https://github.com/JuliaInterop/Clang.jl/tree/master but it is proving difficult given my inexperience with C. 

Does anyone working on Oceananigans have experience doing that sort of thing?

glwagner:
> I've now had a play trying to wrap VkFFT with https://github.com/JuliaInterop/Clang.jl/tree/master but it is proving difficult given my inexperience with C.
> 
> Does anyone working on Oceananigans have experience doing that sort of thing?

Could be worth asking on julia slack! You'll have to ship an independent wrapper package like `VkFFT.jl` and figure out how to precompile the binaries, right (so we can install everything from the REPL).

Could be good for this PR to focus on getting explicit free surface to work, then build up the rest of the features after that. Doing this for real will also require figuring testing out I think.

jagoosw:
weird that all the tests fail, when I run e.g. `RectilinearGrid(CPU(), FT, size=(16, 16, 16), extent=(1, 1, 1))` locally it works fine

src/Architectures.jl
```diff
@@ -27,21 +28,49 @@ struct CPU <: AbstractArchitecture end
...
+
+function GPU()
+    if CUDA.has_cuda()
+        return GPU(CUDA.CUDABackend(; always_inline = true))
+    elseif has_metal_device()
+        return GPU(Metal.MetalBackend())
+    else
+        error("No compatible GPU detected")
+    end
+end
```
glwagner, line 45: Make the device selectable:

```suggestion
function GPU(device=nothing)
    if isnothing(device) # try to find one
        device = if CUDA.has_cuda()
            CUDA.CUDABackend(; always_inline = true)
        elseif has_metal_device()
            Metal.MetalBackend()
        else
            throw(ArgumentError("Cannot construct GPU architecture because no compatible GPU was found!"))
        end
    end
        
    return device
end
```

It also needs a docstring.

src/Architectures.jl
```diff
@@ -27,21 +28,49 @@ struct CPU <: AbstractArchitecture end
...
+Returns true if a metal device is present (not a current function available from Metal.jl)
+"""
+function has_metal_device()
+    try 
+        Metal.current_device()
+        return true
+    catch
+        return false
+    end
+end
```
glwagner, line 59: Sorta piracy here --- probably should be added to Metal.jl via PR (or at least raise an issue)

src/Architectures.jl
```diff
@@ -27,21 +28,49 @@ struct CPU <: AbstractArchitecture end
...
 
 #####
 ##### These methods are extended in DistributedComputations.jl
 #####
 
 device(::CPU) = KernelAbstractions.CPU()
-device(::GPU) = CUDA.CUDABackend(; always_inline=true)
+device(::GPU{<:CUDABackend}) = CUDA.CUDABackend(; always_inline=true)
+device(::GPU{<:MetalBackend}) = Metal.MetalBackend()
```
glwagner, line 67: Not sure why we rebuild --- is there reason? try this:

```suggestion
device(gpu::GPU) = gpu.device
```

also useful

```julia
const CUDAGPU = GPU{<:CUDABackend}
const MetalGPU = GPU{<:MetalBackend}
```

src/Architectures.jl
```diff
@@ -27,21 +28,49 @@ struct CPU <: AbstractArchitecture end
...
 
 architecture() = nothing
 architecture(::Number) = nothing
 architecture(::Array) = CPU()
-architecture(::CuArray) = GPU()
+architecture(::CuArray) = GPU(CUDA.CUDABackend(; always_inline=true))
```
glwagner, line 72: This pattern:

```julia
CUDA.CUDABackend(; always_inline=true)
```

appears many times. This should be defined:

```julia
default_cuda_backend = CUDA.CUDABackend(; always_inline=true)
```

and then used throughout (rather then rewriting `always_inline=true` many times) to throw a bone to future developers.

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -11,6 +11,7 @@ export
     fill_halo_regions!
 
 using CUDA
+using Metal
```
glwagner, line 14: This isn't used I don't think

src/BoundaryConditions/boundary_condition.jl
glwagner: Previously there were only two cases, so there's no distinction between a "whitelist" or "blacklist". But with 3 cases (this PR), that distinction emerges

truedichotomy:
Curious what's the status of this effort to add Metal support to Oceananigans.

simone-silvestri:
Wondering about the status of this PR. I think having Metal implemented will speed up prototyping a ton and force us to deal with Float32.

@jagoosw do you need help on this PR? 
If the FFT does not work we could still start just supporting the Hydrostatic model

jagoosw:
> Wondering about the status of this PR. I think having Metal implemented will speed up prototyping a ton and force us to deal with Float32.
> 
> @jagoosw do you need help on this PR? If the FFT does not work we could still start just supporting the Hydrostatic model

Sorry I stopped working on this and didn't keep the branch up to date!

If you want to tackle this it would probably be easiest to copy the changes from the architecture file to a new branch (assuming that the other float-type problems have been fixed elsewhere). As far as I'm aware there is still no FFT library for Metal in Julia but initially only supporting hydrostatic models sounds like a good step.

glwagner:
Superceded by https://github.com/CliMA/Oceananigans.jl/pull/4124

PR #3293, opened 2023-09-26
Propagate eltype through FFT-based implicit free surface solver

glwagner:
This PR propagates the eltype of the "3D grid" into the horizontal grid that's used by the FFT-based implicit free surface solver.

Should help down the line with one of the issues that cropped up on #3288.

@jagoosw


navidcy (approved):

navidcy:
Nice catch!

PR #3294, opened 2023-09-27
Fixes docstring for free surface in `HydrostaticFreeSurfaceModel`

### LINKED ISSUE #3289: Default free surface solver: `SplitExplicit` or `Implicit`
I noticed that the default free surface model is different than what the docstring says:
https://github.com/CliMA/Oceananigans.jl/blob/e16cdc6cfb67703df9e29368017868331f68b1c0/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl#L59

vs

https://github.com/CliMA/Oceananigans.jl/blob/e16cdc6cfb67703df9e29368017868331f68b1c0/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl#L103

I'm not sure which it is meant to be

navidcy:
This PR deals with the wrong docstring mentioned in #3289.

If in the discussion in #3289 we decide to change the default free-surface solver to Split-Explicit then we can make the change in this PR. ;)

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
@@ -2,7 +2,7 @@ using CUDA: has_cuda
 using OrderedCollections: OrderedDict
 
 using Oceananigans.DistributedComputations
-using Oceananigans.Architectures: AbstractArchitecture, GPU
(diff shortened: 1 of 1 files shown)


glwagner (approved):

navidcy:
OK, I'm merging this so the docstring is fixed and we can continue the discussion in #3289.

PR #3300, opened 2023-09-28
Allow grids with Flat dimensions to use functions with fewer than 3 args in `set!`

### LINKED ISSUE #3285: `node` should omit nodes from `Flat` dimensions
I think it'd be cleaner to change `node` so that `Flat` dimensions are omitted.

For example, this would mean that on a grid that is flat in `z` (two-dimensional in `x, y`, we would write initial conditions

```julia
ui(x, y) = # initial u
set!(model, u=ui)
```

Basically, grids that are `Flat` in `z` cannot depend on `z`. So it would be clearer if initial conditions also did not depend on `z`.

navidcy:
Closes #3285


navidcy:
(This is a placeholder PR for now.)

glwagner:
Nice

navidcy:
superseded by #3355

Issue #3238, opened 2023-08-28
Some issue(s) with `SplitExplicitFreeSurface`


navidcy:
So initially I was trying to use the `SplitExplicitFreeSurface` with the adaptive barotropic step based on CFL.

This:

```Julia
using Oceananigans

grid = RectilinearGrid(size = (10, 10), x = (-100, 100), z = (-100, 0), topology = (Periodic, Flat, Bounded))

free_surface = SplitExplicitFreeSurface(; grid, cfl = 0.7)
```

errors

```Julia
ERROR: ArgumentError: either specify a cfl or a number of substeps
Stacktrace:
 [1] Oceananigans.Models.HydrostaticFreeSurfaceModels.SplitExplicitSettings(FT::DataType; substeps::Int64, cfl::Float64, grid::RectilinearGrid{Float64, Periodic, Flat, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, fixed_Δt::Nothing, gravitational_acceleration::Float64, averaging_kernel::Function, timestepper::Oceananigans.Models.HydrostaticFreeSurfaceModels.ForwardBackwardScheme)
   @ Oceananigans.Models.HydrostaticFreeSurfaceModels ~/.julia/packages/Oceananigans/pbNSE/src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface.jl:306
[...]

PR #3308, opened 2023-10-05
Add HydrostaticFreeSurfaceModel quickstart to README

### LINKED ISSUE #3289: Default free surface solver: `SplitExplicit` or `Implicit`

glwagner:
The objective of this PR is to more clearly and quickly advertise the setting up of Oceananigans models in more than one configuration. We currently have one "quickstart", which sets up `NonhydrostaticModel` to do decaying turbulence. I propose that we add two more for `HydrostaticFreeSurfaceModel`: a rectangular configuration (say baroclinic adjustment) and a spherical configuration, maybe with bathymetry.

I think we should also change defaults to grease the wheels a bit. So this PR will change the free surface to split explicit. I also think we should use WENO advection schemes and increase the default halo size to (4, 4, 4) so that 5th order WENO works on immersed boundary grid. Basically, I think the default should be useful: second-order advection with nothing closure is not useful. 

I'd also like to add the quick start setups to the docs, in addition to the README.

A few more miscellaneous ideas:

- A bulleted list of the examples with a short description of what each example demonstrates (so that one does not have to wade through all of them to find what they are looking for)
- A new page in the docs that summarizes the "major capabilities" of each model

More docs refactoring is ultimately needed but I think this is a good start.

Closes #3289

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
@@ -94,23 +94,24 @@ Keyword arguments
   - `auxiliary_fields`: `NamedTuple` of auxiliary fields. Default: `nothing`.
 """
 function HydrostaticFreeSurfaceModel(; grid,
-                                             clock = Clock{eltype(grid)}(0, 0, 1),
-                                momentum_advection = CenteredSecondOrder(),
-                                  tracer_advection = CenteredSecondOrder(),
-                                          buoyancy = SeawaterBuoyancy(eltype(grid)),
-                                          coriolis = nothing,
-                                      free_surface = ImplicitFreeSurface(gravitational_acceleration = g_Earth),
-                               forcing::NamedTuple = NamedTuple(),
-                                           closure = nothing,
-                   boundary_conditions::NamedTuple = NamedTuple(),
-                                           tracers = (:T, :S),
-                     particles::ParticlesOrNothing = nothing,
-             biogeochemistry::AbstractBGCOrNothing = nothing,
-                                        velocities = nothing,
-                                          pressure = nothing,
-                                diffusivity_fields = nothing,
-                                  auxiliary_fields = NamedTuple()
-    )
+                                     clock = Clock{eltype(grid)}(0, 0, 1),
+                                     momentum_advection = CenteredSecondOrder(),
+                                     tracer_advection = CenteredSecondOrder(),
+                                     buoyancy = SeawaterBuoyancy(eltype(grid)),
+                                     coriolis = nothing,
+                                     free_surface = SplitExplicitFreeSurface(; grid,
+                                                                              cfl = 0.2,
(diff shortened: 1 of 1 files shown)


src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -94,23 +94,24 @@ Keyword arguments
...
-                                  auxiliary_fields = NamedTuple()
-    )
+                                     clock = Clock{eltype(grid)}(0, 0, 1),
+                                     momentum_advection = CenteredSecondOrder(),
+                                     tracer_advection = CenteredSecondOrder(),
+                                     buoyancy = SeawaterBuoyancy(eltype(grid)),
+                                     coriolis = nothing,
+                                     free_surface = SplitExplicitFreeSurface(; grid,
+                                                                              cfl = 0.2,
+                                                                              gravitational_acceleration = g_Earth),
```
navidcy, line 104: change the docstring also please

glwagner:
The show method for split explicit free surface needs work because it has units (not guaranteed) and there's no good show method for `FixedTimeStepSize`:

```julia
free surface: SplitExplicitFreeSurface with gravitational acceleration 9.80665 m s⁻²
│   └── substepping: Oceananigans.Models.HydrostaticFreeSurfaceModels.FixedTimeStepSize{Float64, typeof(Oceananigans.Models.HydrostaticFreeSurfaceModels.averaging_shape_function)}(3.3059697913387924, Oceananigans.Models.HydrostaticFreeSurfaceModels.averaging_shape_function)
```

navidcy:
Tests error with `UndefVarError: `settings` not defined`, e.g.,

https://buildkite.com/clima/oceananigans/builds/12688#018afd6b-fb54-446d-a3cc-c4d1c4f99dd1/18-366

This is related I believe to #3238.

glwagner:
ok, so we can't default to split explicit for multi region?

navidcy:
> ok, so we can't default to split explicit for multi region?

I don't think it's MultiRegion issue. It's something more basic. I'll have a look tomorrow.

navidcy:
Let's try to deal with #3238. There are some glitches with the interface that, e.g., @djlikesdjs and @siddharthabishnu have been stumbling upon.

glwagner:
Ok two PRs that this depends on now: #3238 and #3314 . I'll mark as draft to indicate that this is on hold for a little

glwagner:
Let's revisit this later

PR #3310, opened 2023-10-05
Update dependencies + use Documenter v1

navidcy:


glwagner (approved):

examples/convecting_plankton.jl
```diff
@@ -214,13 +214,13 @@ avg_P_timeseries = FieldTimeSeries(filepath, "avg_P")
 
 times = w_timeseries.times
 buoyancy_flux_time_series = [buoyancy_flux(0, 0, t, buoyancy_flux_parameters) for t in times]
-nothing # hide
+nothing #hide
```
glwagner, line 217: fyi I prefer `# hide` with a space. 

https://documenter.juliadocs.org/stable/man/syntax/#@example-block

navidcy: For Literate.jl I think we _need_ to avoid the space.
https://fredrikekre.github.io/Literate.jl/v2/fileformat/#Filtering-lines

Looking a the docs, the lines with `# hide` were not being hidden...

glwagner: oh ok

navidcy:
Distributed tests fail on cpu. Something related to CUDA? Any ideas? @simone-silvestri?

PR #3323, opened 2023-10-10
Fix various bugs in auto adding bgc tracers and setting up field dependant forcing

jagoosw:
There is a bug when trying to auto add bgc tracers when a user has specified a namedtuple of tracers.

@glwagner I recall discussing dropping support for auto adding tracers, I find it quite useful but can't remember where we landed. 

This PR:
- fixes an issue with auto adding bgc tracers
- fixes an issue with auxiliary field dependant forcing in hydrostatic free surface models and boundary conditions
- adds an error when a user tries to define a forcing or functional boundary condition which depends on a field not found in the model, previously this would just raise a mysterious error like the one in https://github.com/OceanBioME/OceanBioME.jl/pull/146#issuecomment-1755771558

### DIFF
--- a/src/Biogeochemistry.jl
+++ b/src/Biogeochemistry.jl
@@ -126,7 +126,8 @@ tracernames(tracers) = keys(tracers)
 tracernames(tracers::Tuple) = tracers
 
 add_biogeochemical_tracer(tracers::Tuple, name, grid) = tuple(tracers..., name)
-add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, NamedTuple((name => CenterField(grid), )))
+add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, 
+                                                                   NamedTuple{(name, )}((CenterField(grid), )))
 
 @inline function has_biogeochemical_tracers(fields, required_fields, grid)
     user_specified_tracers = [name in tracernames(fields) for name in required_fields]
--- a/src/Biogeochemistry.jl
+++ b/src/Biogeochemistry.jl
@@ -126,8 +126,7 @@ tracernames(tracers) = keys(tracers)
 tracernames(tracers::Tuple) = tracers
 
 add_biogeochemical_tracer(tracers::Tuple, name, grid) = tuple(tracers..., name)
-add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, 
-                                                                   NamedTuple{(name, )}((CenterField(grid), )))
+add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, (; name => CenterField(grid)))
 
 @inline function has_biogeochemical_tracers(fields, required_fields, grid)
     user_specified_tracers = [name in tracernames(fields) for name in required_fields]
--- a/src/Biogeochemistry.jl
+++ b/src/Biogeochemistry.jl
@@ -126,7 +126,7 @@ tracernames(tracers) = keys(tracers)
 tracernames(tracers::Tuple) = tracers
 
 add_biogeochemical_tracer(tracers::Tuple, name, grid) = tuple(tracers..., name)
-add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, NamedTuple(name => CenterField(grid)))
+add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, NamedTuple((name => CenterField(grid), )))
 
 @inline function has_biogeochemical_tracers(fields, required_fields, grid)
     user_specified_tracers = [name in tracernames(fields) for name in required_fields]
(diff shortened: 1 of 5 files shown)


src/Biogeochemistry.jl
```diff
@@ -126,7 +126,7 @@ tracernames(tracers) = keys(tracers)
 tracernames(tracers::Tuple) = tracers
 
 add_biogeochemical_tracer(tracers::Tuple, name, grid) = tuple(tracers..., name)
-add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, NamedTuple(name => CenterField(grid)))
+add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, NamedTuple((name => CenterField(grid), )))
```
glwagner, line 129: This is hard to read, can you find a better way to write this?

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -175,7 +175,7 @@ function HydrostaticFreeSurfaceModel(; grid,
                               G⁻ = HydrostaticFreeSurfaceTendencyFields(velocities, free_surface, grid, tracernames(tracers)))
 
     # Regularize forcing for model tracer and velocity fields.
-    model_fields = hydrostatic_prognostic_fields(velocities, free_surface, tracers)
+    model_fields = (hydrostatic_prognostic_fields(velocities, free_surface, tracers)..., keys(auxiliary_fields)...)
```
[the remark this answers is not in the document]
jagoosw: Ah yes this isn't doing what I thought it was

jagoosw:
I guess when we made it so auxiliary fields can be field dependencies for forcings we a) only tested on nonhydrostatic models, and b) didn't implement for functional boundary conditions.

I also have no idea how this didn't cause OceanBioME models I've run on HydrostaticFreeSurface models to error.

src/Biogeochemistry.jl
```diff
@@ -126,7 +126,8 @@ tracernames(tracers) = keys(tracers)
 tracernames(tracers::Tuple) = tracers
 
 add_biogeochemical_tracer(tracers::Tuple, name, grid) = tuple(tracers..., name)
-add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, NamedTuple(name => CenterField(grid)))
+add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, 
+                                                                   NamedTuple{(name, )}((CenterField(grid), )))
```
glwagner, line 130:
```suggestion
add_biogeochemical_tracer(tracers::NamedTuple, name, grid) = merge(tracers, (; name => CenterField(grid)))
```

glwagner (approved):

simone-silvestri:
nono we are adding another pipeline for distributed as part of #3326. You can disregard the erroring of `buildkite/oceananigans-distributed`

jagoosw:
> nono we are adding another pipeline for distributed as part of #3326. You can disregard the erroring of `buildkite/oceananigans-distributed`

I see, thank you!

jagoosw:
> I've just thought, in the same vein as using total velocities, should `fields(model)` not always return sums of prognostic and background fields?
> 
> For example if you have a function that is dependant on temperature and a background temperature, it should be given the sum.

Thought more about this and I think it might be quite complicated to change to this so I will leave for now

jagoosw:
Does anyone know what the issue with the CPU environment initialization is?

glwagner:
> > I've just thought, in the same vein as using total velocities, should `fields(model)` not always return sums of prognostic and background fields?
> > For example if you have a function that is dependant on temperature and a background temperature, it should be given the sum.
> 
> Thought more about this and I think it might be quite complicated to change to this so I will leave for now

That isn't what `fields` means right now (in fact, it doesn't have much of a concrete meaning at all and is defined rather vaguely, as it arbitrary omits some model fields while including others). Possibly, there is scope for another utility that returns something like "prognostic + background fields, and selected auxiliary fields). Perhaps `forcing_fields` or something.

However, I would pause a moment to consider whether this is the most useful abstraction for users. It may actually prove more limiting than helpful from the user perspective, as it is typically the case that a forcing function (such as a sponge layer) should depend only on perturbation fields rather than a "total" field representing perturbation + background...

glwagner:
Also, I think major changes to the user API should really be proposed in an issue, with a thoughtful discussion of the problem and various alternatives. It wouldn't be appropriate to put such a change into a PR like this.

jagoosw:
Yeah, I thought a bit more about it and agree it doesn't seem like a good idea, and if needed a user could pass the background field separately.

jagoosw:
@glwagner do you know what the problems with the tests are?

glwagner:
> @glwagner do you know what the problems with the tests are?

No idea, I've restarted the tests.

PR #3306, opened 2023-10-04
Momentum dynamics validation on the CubedSphere

navidcy:
with @glwagner

PR #3325, opened 2023-10-10
Correct vorticity computation at the corners of cubed sphere panels

### LINKED ISSUE #1584: Circulation and vorticity operators need to be updated at the cubed sphere corners
https://github.com/CliMA/Oceananigans.jl/blob/28dadbcc52a2badd242e89ba30c8cc34f9117327/src/Operators/vorticity_operators.jl#L1-L2

At the cubed sphere corners the grid is not actually orthogonal and so the vector-invariant formulation breaks down unless we compute the circulation carefully for each corner.

image

I've attached the appropriate reference (Adcroft et al., 2004):

[Adcroft et al, 2004, Monthly Weather Review, Implementation of an Atmosphere–Ocean General Circulation Model on the Expanded Spherical Cube.pdf](https://github.com/CliMA/Oceananigans.jl/files/6311899/Adcroft.et.al.2004.Monthly.Weather.Review.Implementation.of.an.Atmosphere.Ocean.General.Circulation.Model.on.the.Expanded.Spherical.Cube.pdf)

Here is the MITgcm implementation: https://github.com/MITgcm/MITgcm/blob/master/pkg/mom_common/mom_calc_relvort3.F

Once this is fixed, the Rossby-Haurwitz test case in PR #1570 with momentum advection should stop blowing up at the corners.

siddharthabishnu:
Closes #1584.


src/Operators/vorticity_operators.jl
```diff
@@ -18,25 +32,10 @@
...
-        return - Δy_qᶜᶠᶜ(i-1, j, k, grid, v) - Δx_qᶠᶜᶜ(i, j, k, grid, u) + Δx_qᶠᶜᶜ(i, j-1, k, grid, u)
-
-    # Not a corner
-    else
-        return δxᶠᵃᵃ(i, j, k, grid, Δy_qᶜᶠᶜ, v) - δyᵃᶠᵃ(i, j, k, grid, Δx_qᶠᶜᶜ, u)
-    end
-end
+@inline Γᶠᶠᶜ(i, j, k, grid::OrthogonalSphericalShellGrid, u, v) = 
+
+    ifelse(on_south_west_corner(i, j) | on_north_west_corner(i, j, grid),
```
navidcy, line 37:
```suggestion
    ifelse(on_south_west_corner(i, j, grid) | on_north_west_corner(i, j, grid),
```

src/Operators/vorticity_operators.jl
```diff
@@ -8,6 +8,20 @@
  """
 @inline ζ₃ᶠᶠᶜ(i, j, k, grid, u, v) = Γᶠᶠᶜ(i, j, k, grid, u, v) / Azᶠᶠᶜ(i, j, k, grid)
 
+@inline function ζ₃ᶠᶠᶜ(i, j, k, grid::OrthogonalSphericalShellGrid{FT}, u, v) where FT
+    scaling = ifelse(on_corner(i, j, grid), convert(FT, 4/3), one(grid))
```
navidcy, line 12:
```suggestion
    scaling = ifelse(on_corner(i, j, grid), convert(FT, 4/3), 1)
```

navidcy, line 12:
```suggestion
    scaling = ifelse(on_corner(i, j, grid), convert(FT, 4/3), 1)
```

navidcy:
I think we can't move forward with this until we have a working validation script that uses it. So let's wait until #3306 is merged.

src/Operators/vorticity_operators.jl
```diff
@@ -8,6 +8,20 @@
...
+    return scaling * Γᶠᶠᶜ(i, j, k, grid, u, v) / Azᶠᶠᶜ(i, j, k, grid)
+end
+
+# Corner
+@inline on_corner(i, j, grid) = (i == 1 & j == 1) | (i == grid.Nx+1 & j == 1) | (i == 1 & j == grid.Ny+1) | (i == grid.Nx+1 & j == grid.Ny+1)
```
navidcy, line 17:
```suggestion
@inline on_corner(i, j, grid) = on_south_west_corner(i, j, grid) | on_south_east_corner(i, j, grid) | on_north_west_corner(i, j, grid) | on_north_east_corner(i, j, grid)
```

src/Operators/vorticity_operators.jl
navidcy, line 37:
```suggestion
    ifelse(on_south_west_corner(i, j) | on_north_west_corner(i, j, grid),
```

navidcy:
As it is now, it assumes that the `ConformalCubedSphereGrid` has 1 region per panel. Could we generalize it?

Perhaps these are useful?

https://github.com/CliMA/Oceananigans.jl/blob/b0ba4939dd20cf181620881726304257b22bc66c/src/MultiRegion/cubed_sphere_partitions.jl#L61-L83

navidcy:
Closing this in favor of #3306

siddharthabishnu:
Reopening this to align with the goal of keeping commits (#3306 in this case) smaller and more manageable.

siddharthabishnu:
**Final Commit Details**

Remove 4/3 factor, && --> &, ζ=NaN where undefined

1. Remove 4/3 Factor in Vorticity Computation: The 4/3 factor was initially introduced in the vorticity calculation at the corners because the area used to divide the circulation (in the denominator) was reduced to three-quarters of its original value. By adopting the Haversine formula for calculating the quadrilateral area on a sphere---which simplifies to a triangle when two points coincide—--the 4/3 factor is inherently accounted for. Therefore, the explicit inclusion of this factor is no longer necessary.

2. Replace Short-Circuiting && with Logical &: Changed the short-circuiting && operator to the logical & operator to ensure compatibility with GPU operations.

3. Set Circulation Γ and Vorticity ζ to NaN when Undefined: Updated the code to assign NaN (Not a Number) to circulation Γ and vorticity ζ in cases where they are undefined.

@glwagner, @navidcy, could one of you please review this PR once the tests pass, so I can proceed with merging? Thanks!

src/Operators/vorticity_operators.jl
```diff
@@ -30,11 +22,20 @@ end
...
-                  - Δy_qᶜᶠᶜ(i-1, j, k, grid, v) - Δx_qᶠᶜᶜ(i, j, k, grid, u) + Δx_qᶠᶜᶜ(i, j-1, k, grid, u), 
-                  δxᶠᵃᵃ(i, j, k, grid, Δy_qᶜᶠᶜ, v) - δyᵃᶠᵃ(i, j, k, grid, Δx_qᶠᶜᶜ, u)))
\ No newline at end of file
+The vertical circulation associated with horizontal velocities ``u`` and ``v``.
+"""
+
+@inline function Γᶠᶠᶜ(i, j, k, grid::OrthogonalSphericalShellGrid, u, v)
+    Hx, Hy = grid.Hx, grid.Hy
+    Γ = ifelse((i == 1 - Hx) | (j == 1 - Hy),
+               NaN,
```
glwagner, line 31: Putting NaN's can slow the model down, are we sure we want to do this?

siddharthabishnu: I was using NaNs to aid in visualization, but they aren't necessary. Let me go ahead and remove them to avoid any potential slowdown.

src/Operators/vorticity_operators.jl
```diff
@@ -16,27 +22,17 @@
...
-    elseif i == grid.Nx+1 && j == grid.Ny+1
-        return - Δy_qᶜᶠᶜ(i-1, j, k, grid, v) - Δx_qᶠᶜᶜ(i, j, k, grid, u) + Δx_qᶠᶜᶜ(i, j-1, k, grid, u)
+The vertical circulation associated with horizontal velocities ``u`` and ``v``.
+"""
 
-    # Not a corner
-    else
-        return δxᶠᵃᵃ(i, j, k, grid, Δy_qᶜᶠᶜ, v) - δyᵃᶠᵃ(i, j, k, grid, Δx_qᶠᶜᶜ, u)
-    end
+@inline function Γᶠᶠᶜ(i, j, k, grid::OrthogonalSphericalShellGrid, u, v)
```
navidcy, line 28: do we need a new type? eg ConformalCubedSpherePanel? because now OrthogonalSphericaShellGrid is used more broadly, right?

cc: @simone-silvestri

glwagner: Yes

navidcy: @siddharthabishnu can you deal with this?

siddharthabishnu: I can, but as of now, `conformal_cubed_sphere_panel` is a function, not a struct.

glwagner: You should embed the conformal map within the grid:

https://github.com/CliMA/Oceananigans.jl/blob/9c6ed92edf212daaa4c709c4441ce770cdd0bae3/src/Grids/orthogonal_spherical_shell_grid.jl#L46

This is important for two reasons. First, we would like to be able to reproduce the creation of the grid given the grid object. So the parameters that generate the panel should be embedded.

Second, you can use the type of `conformal_mapping` to distinguish between orthogonal grids generated by different mappings. In this way you can tag the grid as being part of the "cubed sphere", thus warranting a different way to compute the circulation.

This was part of changes to orthogonal spherical shell grid a few months ago, concurrent with design of the tripolar grid. The comformal map for the tripolar grid is called `Tripolar`:

https://github.com/CliMA/OrthogonalSphericalShellGrids.jl/blob/379d0de985a0927c026e6c3e43b2ebd163e054aa/src/tripolar_grid.jl#L325

I suggest working first on generating the grid in a clean and understandable way, designing the conformal mapping type. Merge that to main with tests. Next, use that PR to dispatch on the special grid type for computing circulation at the panel corners.

navidcy: @glwagner, what do you mean embed the map within the grid?

navidcy: However, at the moment we are only dispatching on that parameter being `Something` or `Nothing`. You are suggesting we create a type for the conformal map and dispatch over that type, right? (And similarly, we'll create a type for the tripolar map over the OrthogonalSphericalShellGrids.jl package..)

glwagner: > what do you mean "embed the map within the grid"? it's already a field and a parameter of the OrthogonalSphericalShellGrid type, no?

Then why can't you simply dispatch on this special kind of `OrthogonalSphericalShellGrid` to compute the cubed sphere circulation?

navidcy: Yeap, we should do that.

glwagner: > And similarly, we'll create a type for the tripolar map over the OrthogonalSphericalShellGrids.jl package

This already exists

siddharthabishnu: @glwagner, @navidcy please check out PR #3832.

navidcy (approved):

PR #3329, opened 2023-10-10
Add `SeawaterDensity` to `Oceananigans.Models`

jbisits:
See #3316.

This PR adds `SeawaterDensity` which returns a `KernelFunctionOperation` to compute either the in-situ seawater density or a potential density a some reference geopotential height. To calculate the seawater density `SeawaterPolynomials.ρ` is used. This means that at this stage you can only compute the seawater density if the equation of state is a `BoussinesqEquationOfState` because a reference density is required. I was not sure how/if to do this for other `SeawaterBuoyancy` models so have not tried to implement yet.

I made `density_model.jl` a new script so that all the new code could be easily viewed and moved to another location if necessary.


jbisits:
I will fix these tests later today (Sydney time..)!

src/Models/Models.jl
```diff
@@ -136,5 +136,7 @@ function default_nan_checker(model::OceananigansModels)
     return nan_checker
 end
 
-end # module
+# This is here so that `NonhydrostaticModel` and  `HydrsostaticFreeSurfaceModel`
+include("density_model.jl")
```
glwagner, line 140:
```suggestion
include("seawater_density.jl")
```

jbisits:
@glwagner I have simplified the test script (good idea as it was clunky for me to sieve through and I wrote it..) and the new tests pass on both `CPU()` and `GPU()`. The error from the docs build does not look related to this PR as far as I can tell! The docs built on a previous version and I do not think I changed any code related to the docstring.

This PR is already likely enough but one thing that could be done is to move `src/Buoyancy/buoyancy_field.jl`

```julia
# TODO: move to Models
buoyancy(model) = buoyancy(model.buoyancy, model.grid, model.tracers)
buoyancy(b, grid, tracers) = KernelFunctionOperation{Center, Center, Center}(buoyancy_perturbationᶜᶜᶜ, grid, b.model, tracers)
BuoyancyField(model) = Field(buoyancy(model))
```

 to `Oceananigans.Models` and alter this so instead of returning `BuoyancyField` it returns the `KernelFunctionOperation`

```julia
sewater_buoyancy(model) = buoyancy(model.buoyancy, model.grid, model.tracers)
sewater_buoyancy(b, grid, tracers) = KernelFunctionOperation{Center, Center, Center}(buoyancy_perturbationᶜᶜᶜ, grid, b.model, tracers)
SewaterBuoyancyPerturbation(model) = sewater_buoyancy(model)
```

to match the behaviour of `SeawaterDensity`.
Likely better to do another day but if you think it worth doing in this PR let me know.

src/Models/Models.jl
```diff
@@ -136,5 +136,7 @@ function default_nan_checker(model::OceananigansModels)
     return nan_checker
 end
 
-end # module
+# This is here so that `NonhydrostaticModel` and  `HydrsostaticFreeSurfaceModel`
```
glwagner, line 139:
```suggestion
# A utility for computing density in `NonhydrostaticModel` and  `HydrostaticFreeSurfaceModel`
```

src/Models/seawater_density.jl
```diff
@@ -0,0 +1,132 @@
...
+@inline ρ(i, j, k, grid, eos, T, S, Z) = @inbounds ρ(T[i, j, k], S[i, j, k], Z[i, j, k], eos)
+
+"Return a `KernelFunctionOperation` to compute the in-situ `seawater_density`."
+seawater_density(grid, eos, temperature, salinity, geopotential_height) =
+    KernelFunctionOperation{Center, Center, Center}(ρ, grid, eos, temperature, salinity, geopotential_height)
+
+const ModelsWithBuoyancy = Union{NonhydrostaticModel, HydrostaticFreeSurfaceModel}
+
+validate_model_eos(eos) = eos isa BoussinesqEquationOfState ? nothing :
+                                                              throw(ArgumentError("The equation of state must be a `BoussinesqEquationOfState` to compute the density."))
```
glwagner, line 21:
```suggestion
validate_model_eos(eos) = eos isa BoussinesqEquationOfState ? nothing :
                                                              throw(ArgumentError("seawater_density is not defined for $eos."))
```

src/Models/seawater_density.jl
```diff
@@ -0,0 +1,132 @@
...
+    eos = model.buoyancy.model.equation_of_state
+    validate_model_eos(eos)
+    # Convert function or constant user input to AbstractField
+    grid = model.grid
+    loc = (Center, Center, Center)
+    temperature = field(loc, temperature, grid)
+    salinity = field(loc, salinity, grid)
+    # Preferable here is to leave `geopotential_height` as an `AbstractOperation` rather than creating a `Field`
+    geopotential_height = geopotential_height isa KernelFunctionOperation ? Field(geopotential_height) :
+                                                                            field(loc, geopotential_height, grid)
```
glwagner, line 129:
```suggestion
    # Preferable here is to leave `geopotential_height` as an `AbstractOperation` rather than creating a `Field`
    geopotential_height = geopotential_height isa KernelFunctionOperation ? geopotential_height :
                                                                            field(loc, geopotential_height, grid)
```

was this intended?

test/runtests.jl
```diff
@@ -24,6 +24,7 @@ CUDA.allowscalar() do
             include("test_stokes_drift.jl")
             include("test_utils.jl")
             include("test_schedules.jl")
+            include("test_sewater_density.jl")
```
glwagner, line 27: I don't think this belongs in unit tests because it requires a model. Perhaps `time_stepping_3`?

src/Models/seawater_density.jl
jbisits: Yep that is what I wanted. I thought I tried it and it did not work but seems it does!

src/Models/seawater_density.jl
```diff
@@ -0,0 +1,132 @@
...
+├── operand: KernelFunctionOperation at (Center, Center, Center)
+├── status: time=0.0
+└── data: 38×38×1 OffsetArray(::Array{Float64, 3}, -2:35, -2:35, 1:1) with eltype Float64 with indices -2:35×-2:35×1:1
+    └── max=1027.7, min=1027.7, mean=1027.7
+```
+
+Values for `temperature`, `salinity` and `geopotential_height` can be passed to
+`SeawaterDensity` to override the defaults that are obtained from the `model`.
+"""
+function SeawaterDensity(model::ModelsWithBuoyancy;
```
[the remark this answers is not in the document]
jbisits: Cool, thanks for pointing this out. If/when I notice others I will make a PR to change.

test/test_sewater_density.jl
```diff
@@ -0,0 +1,198 @@
...
+                        constant_temperature = nothing, constant_salinity = nothing)
+
+    grid = RectilinearGrid(arch, FT, size=(3, 3, 3), extent=(1, 1, 1))
+    buoyancy =  if !isnothing(constant_temperature)
+                    SeawaterBuoyancy(equation_of_state = eos; constant_temperature)
+                elseif !isnothing(constant_salinity)
+                    SeawaterBuoyancy(equation_of_state = eos; constant_salinity)
+                else
+                    SeawaterBuoyancy(equation_of_state = eos)
+                end
```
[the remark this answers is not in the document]
jbisits: Yes, you are right and it is much nicer!

test/runtests.jl
jbisits: Yep, moved to `time_stepping_3`

src/Models/seawater_density.jl
glwagner: Possibly we should check `AbstractOperation` rather than just `KernelFunctionOperation` (a minor niceity)

jbisits: Done, tests also now check `seawater_density(model) isa AbstractOperation` as well. Is that preferable?

glwagner (approved):

PR #3328, opened 2023-10-10
Regression tests for the Distributed module 

simone-silvestri:
This PR adapts the hydrostatic regression tests to be run in distributed mode.

an important step in this direction is to allow array partitioning under the hood in `set!` if the array size coincides with the global size of the field (implemented here)

a more important step is to remove the topology from the `Distributed` architecture (it should not know anything about topology which is a grid property)

Issue #3202, opened 2023-07-28
Add testing for MultiRegionGrids with `YPartition`


navidcy:
At the moment only `XPartition` is tested:

https://github.com/CliMA/Oceananigans.jl/blob/2447ea7c15d552fb6a50d3fd347d6534af0018c7/test/test_multi_region_unit.jl#L14-L15

PR #3338, opened 2023-10-13
Add Y-partition and XY-partition tests

simone-silvestri:
The code should allow all types of partitioning but there is a bug preventing the tests from passing for y- and xy-partitions.

This PR aims to fix the bug (and consequently add the excluded y- and xy-partitioning tests) 

rework `fill_halo_regions!` to split out non-communicating and communicating boundary conditions such that the latter are executed alone. 

To give an example, for 
```julia 
julia> boundary_conditions = FieldBoundaryConditions(west = NoFluxBoundaryCondition(), east = NoFluxBoundaryCondition(), south = ImpenetrableBoundaryCondition(), north = DistributedCommunicationBoundaryCondition(), bottom = nothing, top = nothing)
├── west: FluxBoundaryCondition: Nothing
├── east: FluxBoundaryCondition: Nothing
├── south: OpenBoundaryCondition: Nothing
├── north: DistributedBoundaryCondition: Nothing
├── bottom: Nothing
├── top: Nothing
└── immersed: DefaultBoundaryCondition (FluxBoundaryCondition: Nothing)
```

in main:
```julia
julia> halo_tuple = permute_boundary_conditions(boundary_conditions);

julia> for i in 1:length(halo_tuple[1])
              @info "operation $(halo_tuple[1][i]) with bcs $((halo_tuple[2][i], halo_tuple[3][i]))"
       end
[ Info: operation fill_bottom_and_top_halo! with bcs (nothing, nothing)
[ Info: operation fill_west_and_east_halo! with bcs (FluxBoundaryCondition: Nothing, FluxBoundaryCondition: Nothing)
[ Info: operation fill_south_and_north_halo! with bcs (OpenBoundaryCondition: Nothing, BoundaryCondition{Oceananigans.BoundaryConditions.DistributedCommunication, Nothing})
```

in this PR
```julia
julia> halo_tuple = permute_boundary_conditions(boundary_conditions);

julia> for i in 1:length(halo_tuple[1])
              @info "operation $(halo_tuple[1][i]) with bcs $(halo_tuple[2][i])"
       end
[ Info: operation fill_bottom_and_top_halo! with bcs (nothing, nothing)
[ Info: operation fill_south_halo! with bcs (OpenBoundaryCondition: Nothing,)
[ Info: operation fill_west_and_east_halo! with bcs (FluxBoundaryCondition: Nothing, FluxBoundaryCondition: Nothing)
[ Info: operation fill_north_halo! with bcs (DistributedBoundaryCondition: Nothing,)
```


src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
```diff
@@ -282,7 +282,7 @@ end
 
 function split_explicit_free_surface_step!(free_surface::SplitExplicitFreeSurface, model, Δt, χ)
 
-    grid = model.grid
+    grid = free_surface.η.grid
```
glwagner, line 285: add a comment here. Are these the same?

test/utils_for_runtests.jl
```diff
@@ -18,11 +18,13 @@ function test_architectures()
...
-               # Should work but tests are failing for some reason
-               # Distributed(child_arch; partition = Partition(1, 4)))
-               # Distributed(child_arch; partition = Partition(2, 2)))
+        return (Distributed(child_arch; partition = Partition(4)),
+                Distributed(child_arch; partition = Partition(1, 4)),
+                Distributed(child_arch; partition = Partition(2, 2))) 
+               # TODO: add support for Non uniform partitioning
+               # Distributed(child_arch; partition = Partition(Rx = [0.2, 0.1, 0.5, 0.3])))
+               # Distributed(child_arch; partition = Partition(Ry = [0.2, 0.1, 0.5, 0.3])))
+               # Distributed(child_arch; partition = Partition(Rx = [0.3, 0.7], Ry = [0.6, 0.4])))
```
glwagner, line 27: not sure this is the syntax we want, but a TODO makes sense.

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
simone-silvestri: They are not the same for `Distributed` because the `free_surface` has enlarged halos. I had changed it in #3328 because I thought it was a bug but actually, there was an error cancellation with another bug found in this PR (the radius of the LatitudeLongitudeGrid was not propagated through `with_halo`). 
In this PR I fixed the latter and I saw the correct grid to use is the free surface one

simone-silvestri:
I think I caught two small bugs that were disallowing y-partitioning. If the tests pass now we should be ready to merge

simone-silvestri:
Actually, the problem persists... tomorrow I ll figure out the issue

simone-silvestri:
This PR includes a solution to #3342

navidcy:
@simone-silvestri should we have a look at #3202 here? E.g., if you figure out something that was causing issues with Y-partition the same may be the culprit for MultiRegion Y-Partition.

simone-silvestri:
I think so, the issue #3342 applies also to multi-region, so I think in a next PR we can extend this halo splitting to multi region boundary conditions

simone-silvestri:
I think so, issue #3342 applies also to multi-region, so I think in another PR we can extend this fill_halo splitting to multi-region boundary conditions

simone-silvestri:
The Y-partition tests with Periodic longitude do not pass because of one point near the boundary where values are really low. I am kind of confident that the implementation is correct, we might be hitting a Floating point ordering problem that does not allow perfect reproducibility.

I am considering relaxing the relative tolerance for distributed tests because I do not really see a way out for regression tests.

simone-silvestri:
especially because we are comparing `Float64` model output with `Float32` saved data.
I propose to regenerate regression data in `Float64` in another PR so we can make the regression tests more repetible

glwagner:
> especially because we are comparing `Float64` model output with `Float32` saved data. I propose to regenerate regression data in `Float64` in another PR so we can make the regression tests more repetible

The regression test data should be saved in Float64.

Regenerate in serial with Float64?

glwagner:
If you regenerate the regression test data in serial, it's still a valid test of the distributed code.

test/data_dependencies.jl
```diff
@@ -16,7 +16,7 @@ datadep"cubed_sphere_32_grid"
 
 # Downloading the regression fields
 
-path = "https://github.com/CliMA/OceananigansArtifacts.jl/raw/main/data_for_regression_tests/"
+path = "https://github.com/CliMA/OceananigansArtifacts.jl/raw/ss/new-regression-Float64/data_for_regression_tests/"
```
glwagner, line 19: why not merge into main?

simone-silvestri: just wanted to test it first

simone-silvestri:
Since I have regenerated the data I thought to add some data for `SplitExplicitFreeSurface`. If we want the test I ll keep it

simone-silvestri:
Ready to merge (then we can merge to `OceananigansArtifacts.jl/main` the new regression data)

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -22,29 +22,24 @@ conditions, possibly recursing into `fields` if it is a nested tuple-of-tuples.
...
+@inline extract_bc(bc, ::Val{:west})   = tuple(extract_west_bc(bc))
+@inline extract_bc(bc, ::Val{:east})   = tuple(extract_east_bc(bc))
+@inline extract_bc(bc, ::Val{:south})  = tuple(extract_south_bc(bc))
+@inline extract_bc(bc, ::Val{:north})  = tuple(extract_north_bc(bc))
+@inline extract_bc(bc, ::Val{:bottom}) = tuple(extract_bottom_bc(bc))
+@inline extract_bc(bc, ::Val{:top})    = tuple(extract_top_bc(bc))
+
+@inline extract_bc(bc, ::Val{:west_and_east})   = (extract_west_bc(bc), extract_east_bc(bc))
+@inline extract_bc(bc, ::Val{:south_and_north}) = (extract_south_bc(bc), extract_north_bc(bc))
+@inline extract_bc(bc, ::Val{:bottom_and_top})  = (extract_bottom_bc(bc), extract_top_bc(bc))
```
glwagner, line 41: What's all this for?

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -56,60 +51,91 @@ function fill_halo_regions!(c::MaybeTupledData, boundary_conditions, indices, lo
     halo_tuple  = permute_boundary_conditions(boundary_conditions)
 
     # Fill halo in the three permuted directions (1, 2, and 3), making sure dependencies are fulfilled
-    for task in 1:3
+    for task in 1:length(halo_tuple[1])
```
glwagner, line 54: What's `length(halo_tuple[1])`?

Might be nice to write this a little differently:

```julia
meaningful_name_for_number_of_tasks = length(halo_tuple[1])
for task in 1:meaningful_name_for_number_of_tasks0
```

where `meaningful_name_for_number_of_tasks` describes the content of `length(halo_tuple[1])`.

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -56,60 +51,91 @@ function fill_halo_regions!(c::MaybeTupledData, boundary_conditions, indices, lo
...
 function fill_halo_event!(task, halo_tuple, c, indices, loc, arch, grid, args...; kwargs...)
-    fill_halo!  = halo_tuple[1][task]
-    bc_left     = halo_tuple[2][task]
-    bc_right    = halo_tuple[3][task]
+    fill_halo! = halo_tuple[1][task]
+    bcs        = halo_tuple[2][task]
 
     # Calculate size and offset of the fill_halo kernel
-    size   = fill_halo_size(c, fill_halo!, indices, bc_left, loc, grid)
+    size   = fill_halo_size(c, fill_halo!, indices, bcs[1], loc, grid)
```
glwagner, line 66: Isn't there an assumption hidden in using `bcs[1]`? What is that assumption? Document your assumption with a comment.

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -118,10 +144,10 @@ const PBCT  = Union{PBC,  NTuple{<:Any, <:PBC}}
...
 const DCBCT = Union{DCBC, NTuple{<:Any, <:DCBC}}
 
-# Distributed halos have to be filled for last in case of 
-# buffered communication. Hence, we always fill them last
-
-# The reasoning for filling Periodic after Flux, Value, Gradient 
+# Distributed halos have to be filled last because of
+# buffered communication. 
+ 
+# Periodic is handled after Flux, Value, Gradient because
```
glwagner, line 150: I thought this was fixed in this PR?

src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -118,10 +144,10 @@ const PBCT  = Union{PBC,  NTuple{<:Any, <:PBC}}
 const MCBCT = Union{MCBC, NTuple{<:Any, <:MCBC}}
 const DCBCT = Union{DCBC, NTuple{<:Any, <:DCBC}}
 
-# Distributed halos have to be filled for last in case of 
-# buffered communication. Hence, we always fill them last
-
-# The reasoning for filling Periodic after Flux, Value, Gradient 
+# Distributed halos have to be filled last because of
+# buffered communication.
```
glwagner, line 148: This doesn't explain it to me completely. Why would buffered communication require halos to be filled last? What do buffers have to do with it? Is that point that you have to complete other halo filling tasks before communicating, if you want to avoid filling the halos _again_ after communication?

It seems like we could communicate first before filling halos, right?

src/BoundaryConditions/fill_halo_regions_periodic.jl
```diff
@@ -106,3 +106,14 @@ end
...
+#####
+##### Throw error if single-sided periodic boundary conditions are used
+#####
+
+fill_west_halo!(c, ::PBCT, args...; kwargs...)   = throw(ArgumentError("Periodic boundary conditions must be applied to both sides"))
+fill_east_halo!(c, ::PBCT, args...; kwargs...)   = throw(ArgumentError("Periodic boundary conditions must be applied to both sides"))
+fill_south_halo!(c, ::PBCT, args...; kwargs...)  = throw(ArgumentError("Periodic boundary conditions must be applied to both sides"))
+fill_north_halo!(c, ::PBCT, args...; kwargs...)  = throw(ArgumentError("Periodic boundary conditions must be applied to both sides"))
+fill_bottom_halo!(c, ::PBCT, args...; kwargs...) = throw(ArgumentError("Periodic boundary conditions must be applied to both sides"))
+fill_top_halo!(c, ::PBCT, args...; kwargs...)    = throw(ArgumentError("Periodic boundary conditions must be applied to both sides"))
```
glwagner, line 119:
```suggestion
fill_top_halo!(c, ::PBCT, args...; kwargs...)    = throw(ArgumentError("Periodic boundary conditions must be applied to both sides"))

```

src/DistributedComputations/halo_communication.jl
```diff
@@ -101,11 +101,11 @@ function fill_halo_regions!(c::OffsetArray, bcs, indices, loc, grid::Distributed
     arch       = architecture(grid)
     halo_tuple = permute_boundary_conditions(bcs)
 
-    for task = 1:3
+    for task = 1:length(halo_tuple[1])
```
glwagner, line 104: What is the meaning of "halo_tuple[1]"?

src/Fields/field_boundary_buffers.jl
```diff
@@ -132,6 +132,25 @@ function fill_send_buffers!(c::OffsetArray, buff::FieldBoundaryBuffers, grid, ::
...
+fill_send_buffers!(c::OffsetArray, buff::FieldBoundaryBuffers, grid, ::Val{:west}) = 
+    _fill_west_send_buffer!(parent(c), buff, buff.west, halo_size(grid)[[1, 2]]..., size(grid)[[1, 2]]...)
+fill_send_buffers!(c::OffsetArray, buff::FieldBoundaryBuffers, grid, ::Val{:east}) = 
+    _fill_east_send_buffer!(parent(c), buff, buff.east, halo_size(grid)[[1, 2]]..., size(grid)[[1, 2]]...)
+fill_send_buffers!(c::OffsetArray, buff::FieldBoundaryBuffers, grid, ::Val{:south}) = 
+    _fill_south_send_buffer!(parent(c), buff, buff.south, halo_size(grid)[[1, 2]]..., size(grid)[[1, 2]]...)
+fill_send_buffers!(c::OffsetArray, buff::FieldBoundaryBuffers, grid, ::Val{:north}) = 
+    _fill_north_send_buffer!(parent(c), buff, buff.north, halo_size(grid)[[1, 2]]..., size(grid)[[1, 2]]...)
+fill_send_buffers!(c::OffsetArray, buff::FieldBoundaryBuffers, grid, ::Val{:bottom}) = nothing
+fill_send_buffers!(c::OffsetArray, buff::FieldBoundaryBuffers, grid, ::Val{:top}) = nothing
```
glwagner, line 148: What advantage is using `Val{:top}`?

Usually one would just write something like `fill_top_buffers!`.

src/Models/HydrostaticFreeSurfaceModels/split_explicit_free_surface_kernels.jl
glwagner: You may want to call it `free_surface_grid`.

test/data_dependencies.jl
glwagner: Still testing?

src/BoundaryConditions/fill_halo_regions.jl
simone-silvestri: That is a much more disruptive change which is nice to have but I believe we should approach it in another dedicated PR (see issue #3342)

src/BoundaryConditions/fill_halo_regions.jl
simone-silvestri: I will update the comment. If you want asynchronous communication there are two options:

1) Fill all local halos - fill the buffers and start the communication - perform any required computation - complete the communication.
2) Fill the buffers and start the communication - perform any required computation - complete the communication - Fill any local halos.

with the first option, nothing changes in the case of non-distributed fields so I went in that direction in the #3125.

src/Fields/field_boundary_buffers.jl
simone-silvestri: that function is needed here 
https://github.com/CliMA/Oceananigans.jl/blob/af8450fe65f993f89411fdfa8ace4b1a4bf291b2/src/DistributedComputations/halo_communication.jl#L184
where we do not necessarily know which direction is being filled so we infer it from the `fill_halo!` function

src/BoundaryConditions/fill_halo_regions.jl
glwagner: By "complete the communication" you mean "wait until communication finishes" right?

simone-silvestri: completing the communication requires waiting until the communication is finished and then filling the relevant halos from the received buffer. In practice, we can think about it as waiting until the communication is finished (assuming that filling the relevant halos is part of the communication process).

src/Fields/field_boundary_buffers.jl
glwagner: Ok got it. My main comment then is remember that by YASGuide the function arguments have the ordering:

1. Anything being mutated (in a mutating function)
2. Functions being dispatched on (ie best not to bury dispatchable arguments deep in the argument list)
3. Everything else in a nice or logical order

We break this only for the operator form functions which need to have first arguments `i, j, k, grid`.

src/BoundaryConditions/fill_halo_regions.jl
glwagner: Oh okay, so this will extract a "tuple of west bcs" and a "tuple of east bcs".

Can you write that in a comment?

src/Fields/field_boundary_buffers.jl
simone-silvestri: Changed it

src/BoundaryConditions/fill_halo_regions.jl
glwagner: Okay so the first step is to initiate communication, and the second step is to fill halos from the buffer. The code is misleading, because the initiation of communication is currently called "fill halo regions!". But halo filling is actually occurring in the second step, right?

src/BoundaryConditions/fill_halo_regions.jl
glwagner: Alright

src/BoundaryConditions/fill_halo_regions.jl
simone-silvestri: in asynchronous communication `fill_halo_regions!(f; async = true)` yes, the communication is completed in the `synchronize_communication!(f)` otherwise everything is completed in the `fill_halo_regions!` function

glwagner: Perhaps `finish_fill_halo_regions!(f)` or something would be helpful for the second step

glwagner: And probably neesd a comment in the code to indicate that the halo regions are not actually filled into the second step in the case of a distributed computation

simone-silvestri: We assume that the second step completes everything: i.e. wait for communication to complete and fill the halos
https://github.com/CliMA/Oceananigans.jl/blob/d0b7ec8f98c860ce49927e0a7214961d2f47fb75/src/DistributedComputations/distributed_fields.jl#L56-L77

src/DistributedComputations/halo_communication.jl
simone-silvestri: changed to `number_of_tasks`

simone-silvestri:
I think I should have addressed all comments

glwagner (approved):

PR #3339, opened 2023-10-13
Non uniform partitioning for `Distributed` architectures

simone-silvestri:
This PR tweaks the API to simplify non-uniform partitioning which should already be supported by the algorithm.

This PR also extends the tests to include non-uniform distributed partitioning 

The proposal of this PR (up to discussion and tweaking) is to allow calling
```julia
arch = Distributed(CPU(); partition = Partition(Rx = [0.3, 0.1, 0.6])
```
which allows to distributed the domain over 3 workers which hold 30%, 10% and 60% of the computation, respectively


glwagner:
> This PR tweaks the API to simplify non-uniform partitioning which should already be supported by the algorithm.
> 
> This PR also extends the tests to include non-uniform distributed partitioning
> 
> The proposal of this PR (up to discussion and tweaking) is to allow calling
> 
> ```julia
> arch = Distributed(CPU(); partition = Partition(Rx = [0.3, 0.1, 0.6])
> ```
> 
> which allows to distributed the domain over 3 workers which hold 30%, 10% and 60% of the computation, respectively

How do you manually specify the grid points for each partition with this API?

glwagner:
Here's another concept for an API. As I understand this, the basic operations we want to support divide a 2D plane into a rectilinear grid. Based on this principle, we can support only a dimension-by-dimension specification (ie x and y are specified independently). Example use cases are:

* equal partitioning of one or both dimensions among `R` ranks
* non-uniform partitioning based on the number of grid points carried by each rank
* non-uniform partitioning based on the relative size (necessarily approximate) associated with each rank

Is there anything else we want to support?

Here's a syntax that achieves those:

```julia
p = Partition(x=Equal()) # computation equally partitioned among available ranks, however many are available.
p = Partition(x=4) # computation equally partitioned between 4 ranks. There must be 4 available ranks
p = Partition(x=2, y=Equal()) # computational divided among 2 ranks in x, and the remaining distributed equally in y. There must be at least 2 ranks.
p = Partition(x=Relative(0.1, 0.3, 0.6)) # computational unevenly distributed among 3 ranks in x each given grids that contain 10%, 30%, 60% of the total grid points. There must be 3 ranks.
p = Partition(x=Relative(1, 3, 6)) # same as above (eg we normalize by the total amount, not requiring =1)
p = Partition(x=Equal(), y=Relative(1, 3, 6)) # computation unevenly distributed in y among 3 ranks, but evenly distributed in x amoung remaining ranks. There must be at least 3 ranks.
p = Partition(x=[16, 32, 64]) # computational distributed among 3 ranks each with 16, 32, and 64 grid points total. The total size of the grid must be 16 + 32 + 64 = 112 and there must be 3 ranks.
```

simone-silvestri:
I like this API, my only reservation is for specifying grid points. This would reverse the flow of information between architecture and the grid. But I guess we can have that as a fine-tuning setup as long as `Relative` is there...

glwagner:
> I like this API, my only reservation is for specifying grid points. This would reverse the flow of information between architecture and the grid. But I guess we can have that as a fine-tuning setup as long as `Relative` is there...

The "priority" within API could be reversed by implementing

```julia
p = Partition(x=Sizes(16, 32, 64)) # manually specified sizes (provided only for fine-tuning, discouraged from typical use)
```

glwagner:
One thought --- `Relative` is also defined by `Makie` so the names will clash:

https://docs.makie.org/stable/reference/blocks/gridlayout/#relative

Maybe another name would be better like `Fractional` or something.

simone-silvestri:
ready to merge after #3338

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -51,10 +23,46 @@ Return `Partition` representing the division of a domain into
...
+Partition(Rx, Ry) = Partition(validate_partition(Rx, Ry, 1)...)
+
+function Partition(; x = 1, y = 1, z = 1)
+    x, y, z = validate_partition(x, y, z)
+    return Partition(x, y, z)
+end
+
+struct Equal end
+struct Fractional{S} end
+struct Sizes{S} end
```
glwagner, line 42: Why `Sizes{S}` and not

```julia
struct Sizes{S}
    sizes :: S
end
```

it's a little simpler if the struct can be interrogated via `partition.x.sizes`?

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -10,38 +10,10 @@ import Oceananigans.Utils: sync_device!, tupleit
...
-    function Partition(sizes::S) where S
-        Rx = size(sizes, 1)
-        Ry = size(sizes, 2)
-        Rz = size(sizes, 3)
-        return new{S, Rx, Ry, Rz}(sizes)
-    end
+struct Partition{Sx, Sy, Sz}
+    Rx :: Sx
+    Ry :: Sy
+    Rz :: Sz
```
glwagner, line 16:
```suggestion
    x :: Sx
    y :: Sy
    z :: Sz
```

the "R" used to mean "Ranks". I don't think that's what `Rx` means anymore.

src/DistributedComputations/distributed_architectures.jl
glwagner: ditto for `Fractional`

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -51,10 +23,46 @@ Return `Partition` representing the division of a domain into
...
+struct Sizes{S} end
+
+ranks(p::Partition) = (ranks(p.Rx), ranks(p.Ry), ranks(p.Rz))
+ranks(r::Int)      = r
+ranks(r::Sizes)    = length(r)
+ranks(r::Fractional) = length(r)
+
+Base.size(p::Partition) = ranks(p)
+
+Fractional(args...) = sum(args) != 1 ? Fractional{args ./ sum(args)}() : Fractional{args}()
```
glwagner, line 51: Why the if statement? If `sum(args) == 1` then `args ./ sum(args) == args` right?

simone-silvestri: to normalize the arguments to fractional values. Maybe we can issue a warning before so that the user knows we are normalizing the partitions?

glwagner: This code does the same thing:

```julia
Fractional(args...) = Fractional{args ./ sum(args)}()
```

there is no need for an `if`-statement. It doesn't change anything.

simone-silvestri: oh yeah right

simone-silvestri:
This PR should be complete.

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -51,10 +25,47 @@ Return `Partition` representing the division of a domain into
...
+`Rx`, `Ry` and `Rz` can be `Int` or `Equal`, `Fractional` or `Sizes` 
+(see below)
 """
-Partition(Rx::Number, Ry::Number=1, Rz::Number=1) = Partition{Rx, Ry, Rz}()
-Partition(lengths::Array{Int, 1}) = Partition(reshape(lengths, length(lengths), 1))
-Base.size(::Partition{<:Any, Rx, Ry, Rz}) where {Rx, Ry, Rz} = (Rx, Ry, Rz)
+Partition(Rx)     = Partition(validate_partition(Rx, 1, 1)...)
+Partition(Rx, Ry) = Partition(validate_partition(Rx, Ry, 1)...)
+
+function Partition(; x = 1, y = 1, z = 1)
```
glwagner, line 35: I think we need a docstring with a few examples of uniform, non-uniform, etc partitioning?

glwagner:
Can you update the top-level description, and add a docstring for `Partition` with a few examples that enumerates the various possible syntaxes for common important cases?

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -6,55 +6,109 @@ import Oceananigans.Architectures: device, cpu_architecture, arch_array, array_t
...
+
+- `x`: partitioning of the first dimension 
+- `y`: partitioning of the second dimension
+- `z`: partitioning of the third dimension
+
+if supplied as positional arguments `x` will be the first argument, 
+`y` the second and `z` the third
+
+`x`, `y` and `z` can be `Int`, `Equal`, `Fractional` or `Sizes` 
+(see below)
```
glwagner, line 38: I think we need to explain what each option means, something like:

```suggestion
`x`, `y` and `z` can be:
    `x::Int`: allocate `x` processors to the first dimension
    `Equal()`: divide the domain in `x` equally among the remaining processes
    `Fractional(ϵ1, ϵ2, ..., ϵN):` divide the domain unequally among `N` processes. The total work is `W = sum(ϵi)`, and each process is then allocated `ϵi / W` of the domain.
    `Sizes`:
```

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -6,55 +6,109 @@ import Oceananigans.Architectures: device, cpu_architecture, arch_array, array_t
...
+    print(io,
+    "Domain partitioning with $(ranks(p)) ranks", "\n",
+    "├── x-partitioning: $(ranks(p.x) == 1 ? "none" : p.x)", "\n",
+    "├── y-partitioning: $(ranks(p.y) == 1 ? "none" : p.y)", "\n",
+    "└── z-partitioning: $(ranks(p.z) == 1 ? "none" : p.z)")
 
-# For 3D partitions:
-# regularize_size(sz::Tuple{<:Int, <:Int, <:Int}, Rx, Ry) = sz
+"""type representing equal domain partitioning (not supported for more than one direction)"""
+struct Equal end
```
glwagner, line 73: I think we need a real docstring because this is user facing, something like

```suggestion
"""
    Equal()

Return a type that partitions a direction equally among remaining processes.

`Equal()` can be used for only one direction. Other directions must either be unspecified, or
specifically defined by `Int`, `Fractional`, or `Sizes`.
"""
```

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -6,55 +6,109 @@ import Oceananigans.Architectures: device, cpu_architecture, arch_array, array_t
...
+"""
+Partition(x)       = Partition(validate_partition(x, 1, 1)...)
+Partition(x, y)    = Partition(validate_partition(x, y, 1)...)
+
+Partition(; x = 1, y = 1, z = 1) = Partition(validate_partition(x, y, z)...)
+
+Base.show(io::IO, p::Partition) =
+    print(io,
+    "Domain partitioning with $(ranks(p)) ranks", "\n",
+    "├── x-partitioning: $(ranks(p.x) == 1 ? "none" : p.x)", "\n",
```
glwagner, line 68: Should we say "1" rather than "none"?

The purpose of a display is to give as much code-relevant information about the content of a type as possible. So I think "1" is more accurate, while "none" is not right, it doesn't correspond to any julia type (unless we are using "nothing"), but in that case we should say "nothing" not "none").

test/utils_for_runtests.jl
```diff
@@ -19,8 +19,11 @@ function test_architectures()
...
-                Distributed(child_arch; partition = Partition(1, 4), synchronized_communication = true),
-                Distributed(child_arch; partition = Partition(2, 2))) 
+                Distributed(child_arch; partition = Partition(1, 4)),
+                Distributed(child_arch; partition = Partition(2, 2)),
+                Distributed(child_arch; partition = Partition(x = Fractional(1, 2, 3, 4))),
+                Distributed(child_arch; partition = Partition(y = Fractional(1, 2, 3, 4))),
+                Distributed(child_arch; partition = Partition(x = Fractional(1, 2), y = Equal()))) 
                # TODO: add support for Non uniform partitioning
                # Distributed(child_arch; partition = Partition(Rx = [0.2, 0.1, 0.5, 0.3])))
                # Distributed(child_arch; partition = Partition(Ry = [0.2, 0.1, 0.5, 0.3])))
```
glwagner, line 29:
```suggestion
```

src/DistributedComputations/distributed_architectures.jl
simone-silvestri: I like to differentiate from 1 because with 1 rank there is no partitioning. `nothing` is a good option

simone-silvestri: probably it is just better to omit the directions in which rank == 1

glwagner: what's the difference between "1" and "nothing"? If you write `x=1`, isn't that the same thing as no partition?

glwagner: It's important that it matches the underlying code, if `partition.x` should match what we claim the `x` partitioning is

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -39,7 +39,7 @@ if supplied as positional arguments `x` will be the first argument,
     `Equal()`: divide the domain in `x` equally among the remaining processes (not supported for multiple directions)
     `Fractional(ϵ1, ϵ2, ..., ϵN):` divide the domain unequally among `N` processes. The total work is `W = sum(ϵi)`, 
                                    and each process is then allocated `ϵi / W` of the domain.
-    `Sizes(ϵ1, ϵ2, ..., ϵN)`: divide the domain unequally. EThe total work is `W = sum(ϵi)`, 
+    `Sizes(ϵ1, ϵ2, ..., ϵN)`: divide the domain unequally. The total work is `W = sum(ϵi)`,
```
glwagner, line 42:
```suggestion
    `Sizes(n1, n2, ..., nN)`: divide the direction by number of grid points, where `ni` is the number of grid points allocated to process `i`. The total size of the direction is `N = sum(ni)`, 
```

Maybe a different letter than `N` for the number of processors, maybe `P` is better?

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -83,18 +83,29 @@ specifically defined by `Int`, `Fractional`, or `Sizes`.
...
+    Fractional(ϵ1, ϵ2, ..., ϵN)
+
+Return a type that partitions a direction unequally. The total work is `W = sum(ϵi)`, 
+and each process is then allocated `ϵi / W` of the domain.
+"""
 Fractional(args...) = Fractional(tuple(args ./ sum(args)...))  # We need to make sure that `sum(R) == 1`
-     Sizes(args...) = Sizes(tuple(args...))
+
+"""
+`Sizes(ϵ1, ϵ2, ..., ϵN)`
```
glwagner, line 103: I used `ϵ` for fractions because that made sense but we should use a different letter for sizes.

It's good to get this right because users will copy the notation we use in docstrings, so it propagates everywhere.

src/DistributedComputations/distributed_architectures.jl
simone-silvestri: ok let's go with nothing then

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -6,55 +6,131 @@ import Oceananigans.Architectures: device, cpu_architecture, arch_array, array_t
...
+Base.show(io::IO, s::Fractional) = print(io, "domain fractions: $(s.sizes)")
+
+ranks(p::Partition)  = (ranks(p.x), ranks(p.y), ranks(p.z))
+ranks(r::Int)        = r
+ranks(r::Sizes)      = length(r.sizes)
+ranks(r::Fractional) = length(r.sizes)
+
+Base.size(p::Partition) = ranks(p)
+
+validate_partition(x, y, z) = (x, y, z)
```
glwagner, line 124: This has to convert `1` to `nothing` right?

src/DistributedComputations/distributed_architectures.jl
glwagner: ok, I think we need to change `validate_partition` as well in that case

glwagner (approved):

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -6,55 +6,135 @@ import Oceananigans.Architectures: device, cpu_architecture, arch_array, array_t
...
+
+validate_partition(x, y, z) = validate_partition.(x, y, z)
+validate_partition(::Equal, y, z) = remaining_workers(y, z), y, z
+
+validate_partition(x, ::Equal, z) = x, remaining_workers(x, z), z
+validate_partition(x, y, ::Equal) = x, y, remaining_workers(x, y)
+
+function remaining_workers(r1, r2)
+    MPI.Initialized() || MPI.Init()    
+    return MPI.Comm_size(MPI.COMM_WORLD) ÷ (ranks(r1)*ranks(r2))
```
glwagner, line 136:
```suggestion
    return MPI.Comm_size(MPI.COMM_WORLD) ÷ (ranks(r1) * ranks(r2))
```

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -6,55 +6,135 @@ import Oceananigans.Architectures: device, cpu_architecture, arch_array, array_t
...
+julia> Partition(x = Fractional(1, 2, 3, 4))
+Domain partitioning with (4, 1, 1) ranks
+└── x-partitioning: domain fractions: (0.1, 0.2, 0.3, 0.4)
+
+```
+"""
+Partition(x)    = Partition(validate_partition(x, 1, 1)...)
+Partition(x, y) = Partition(validate_partition(x, y, 1)...)
+
+Partition(; x = 1, y = 1, z = 1) = Partition(validate_partition(x, y, z)...)
```
glwagner, line 64:
```suggestion
Partition(; x = nothing, y = nothing, z = nothing) = Partition(validate_partition(x, y, z)...)
```

?

src/DistributedComputations/distributed_architectures.jl
```diff
@@ -6,55 +6,135 @@ import Oceananigans.Architectures: device, cpu_architecture, arch_array, array_t
...
+└── y-partitioning: 4
+
+julia> Partition(x = Fractional(1, 2, 3, 4))
+Domain partitioning with (4, 1, 1) ranks
+└── x-partitioning: domain fractions: (0.1, 0.2, 0.3, 0.4)
+
+```
+"""
+Partition(x)    = Partition(validate_partition(x, 1, 1)...)
+Partition(x, y) = Partition(validate_partition(x, y, 1)...)
```
glwagner, line 62:
```suggestion
Partition(x)    = Partition(validate_partition(x, nothing, nothing)...)
Partition(x, y) = Partition(validate_partition(x, y, nothing)...)
```

PR #3340, opened 2023-10-14
Refactor vertical vorticity to return operation rather than Field

glwagner:
Following the newfound insight from implementing `seawater_density`, I believe we always want to define these diagnostics to return `KernelFunctionOperation` rather than `Field`. This PR also fixes the incorrect use of TitleCase.


navidcy:
Sounds good. Is this tested, do you know?

src/Models/HydrostaticFreeSurfaceModels/vertical_vorticity.jl
```diff
@@ -0,0 +1,17 @@
+using Oceananigans.Grids: Face, Face, Center
+using Oceananigans.Operators: ζ₃ᶠᶠᶜ
+using Oceananigans.AbstractOperations: KernelFunctionOperation
+
+"""
+    vertical_vorticity(model)
```
navidcy, line 6:
```suggestion
    vertical_vorticity((model::HydrostaticFreeSurfaceModel)
```

glwagner:
I was thinking we'd find out via the tests

navidcy (approved):

glwagner:
> Sounds good. Is this tested, do you know?

looks like it is

PR #3341, opened 2023-10-14
(0.90.0) Define `WENOVectorInvariant` + some cleanup

glwagner:
This PR defines and exports "WENO" `VectorInvariant`, which is a convenience constructor for `VectorInvariant` wherein the vorticity, divergence, and vertical schemes are all WENO. It also defines the proper type alias and refactors the code to make it slightly more readable.


navidcy (approved):
I like this

src/Advection/Advection.jl
```diff
@@ -20,7 +20,7 @@ export
     Centered, CenteredSecondOrder, CenteredFourthOrder,
     UpwindBiased, UpwindBiasedFirstOrder, UpwindBiasedThirdOrder, UpwindBiasedFifthOrder,
     WENO, WENOThirdOrder, WENOFifthOrder,
-    VectorInvariant,
+    VectorInvariant, WENOVectorInvariant,
     EnergyConservingScheme,
     EnstrophyConservingScheme
```
navidcy, line 25:
```suggestion
    EnergyConserving,
    EnstrophyConserving
```

navidcy (commented):
This is a breaking change, right?

glwagner:
> With the name changes, e.g., `EnergyConservingScheme -> EnergyConserving`, this PR introduces a breaking change, right?

True. We can revert that.

navidcy:
Leave it. I approve. Just bump minor release.

src/Advection/vector_invariant_advection.jl
```diff
@@ -116,27 +163,65 @@ Base.show(io::IO, a::VectorInvariant{N, FT}) where {N, FT} =
...
+                                   multi_dimensional_stencil = false,
+                                   weno_kw...)
+
+"""
+function WENOVectorInvariant(; upwinding = nothing,
+                               vorticity_stencil = VelocityStencil(),
+                               multi_dimensional_stencil = false,
+                               weno_kw...)
+
+    weno = WENO(; weno_kw...)
```
simone-silvestri, line 184: most of the time we would like a higher order for vorticity. Can we have a `vorticity_order` keyword argument?

navidcy: what should be the default order?

simone-silvestri: I found that 9 for vorticity and 5 for the other gave good results with high enough resolution.

glwagner: But not order 9 for the others?

glwagner:
> Leave it. I approve. Let's just bump a minor release.

We should have rules about what is "official" API and what isn't. Because it doesn't really make sense to bump here.

navidcy:
Hm. I thought anything that might require you to change your scripts was a breaking change. But you are right that this is bottomless in a sense.

navidcy:
OK, shall we not bump minor for this you reckon?

I think the advection scheme is something that most users choose in the model constructor so for me it seems like a worthy-enough API change for a breaking change. But we can have this discussion elsewhere if you prefer. Happy for you to decide on this one, either way I like the changes in this PR!

src/Advection/vector_invariant_advection.jl
simone-silvestri: with 9 for the others, the scheme becomes heavy without really affecting the solution much.
It is also nice to keep a lower order scheme for the vertical advection because it is more of a "difficult" term when using low vertical resolution typical of a hydrostatic model.

glwagner:
> Hm. I thought anything that might require you to change your scripts was a breaking change. But you are right that this is bottomless in a sense.

Our informal definition of "user API" consists of all exported functions. So, if you don't export something, it's free to change without version bump. That's consistent with Julia's convention.

navidcy:
These are exported types that got renamed: `EnergyConservingScheme` -> `EnergyConserving`, `EnstrophyConservingScheme` -> `EnstrophyConserving`,

src/Advection/vector_invariant_advection.jl
```diff
@@ -116,27 +163,65 @@ Base.show(io::IO, a::VectorInvariant{N, FT}) where {N, FT} =
               "$(a.vertical_scheme isa AbstractUpwindBiasedAdvectionScheme ? 
               "\n └── upwinding treatment: $(a.upwinding)" : "")")
 
+#####
+##### Convenience for WENO Vector Invariant
+#####
+
+#                           VectorInvariant{N,     FT,    M,     Z,      ZS,    V,      K (kinetic energy gradient scheme)
```
simone-silvestri, line 170:
```suggestion
#                           VectorInvariant{N,     FT,    M,     Z (vorticity scheme),      ZS,    V (vertical scheme),      K (kinetic energy gradient scheme)
```

PR #3343, opened 2023-10-15
Excise `LatitudeLongitudeGrid`

### LINKED ISSUE #3205: Eliminate `LatitudeLongitudeGrid` in place of the `OrthogonalSphericalShellGrid`

simone-silvestri:
This PR removes the `LatitudeLongitudeGrid` type in favor of `OrthogonalSphericalShellGrid`

API (and hopefully internals) do not change

Closes #3205


simone-silvestri:
As part of this PR, I need to come up with a name for the `OrthogonalSphericalShellGrid` with a conformal cubed sphere mapping (or _classification_ how we call it here). 

I have named it here `ConformalCubedSpherePanelGrid` with the corresponding `ConformalCubedSpherePanel` classification.

@navidcy @siddharthabishnu any comments or opinions?

simone-silvestri:
I think we want to keep the lat-lon grid, one of the advantages is the reduced parameter space when compared to an orthogonal spherical shell grid. 
I will close this PR, but happy to reopen it if the consensus is still to pursue this direction.

PR #3344, opened 2023-10-16
Modify `NonhydrostaticModel` to include different advection schemes for velocities and tracers

xkykai:
This PR would allow the user to choose the advection schemes they wish to use for velocities and tracers separately. 
This improves customizability as well as allowing the user to run other arbitrary sets of equations that requires timestepping.

### DIFF
--- a/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
+++ b/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
@@ -11,13 +11,14 @@ using Oceananigans.Fields: BackgroundFields, Field, tracernames, VelocityFields,
 using Oceananigans.Forcings: model_forcing
 using Oceananigans.Grids: inflate_halo_size, with_halo, architecture
 using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid
-using Oceananigans.Models: AbstractModel, NaNChecker, extract_boundary_conditions
+using Oceananigans.Models: AbstractModel, NaNChecker, extract_boundary_conditions, validate_tracer_advection
 using Oceananigans.Solvers: FFTBasedPoissonSolver
 using Oceananigans.TimeSteppers: Clock, TimeStepper, update_state!, AbstractLagrangianParticles
 using Oceananigans.TurbulenceClosures: validate_closure, with_tracers, DiffusivityFields, time_discretization, implicit_diffusion_solver
 using Oceananigans.TurbulenceClosures.CATKEVerticalDiffusivities: FlavorOfCATKE
 using Oceananigans.Utils: tupleit
 using Oceananigans.Grids: topology
+using Oceananigans.Models.HydrostaticFreeSurfaceModels: validate_momentum_advection
 
 import Oceananigans.Architectures: architecture
 import Oceananigans.Models: total_velocities, default_nan_checker, timestepper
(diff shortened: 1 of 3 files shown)


simone-silvestri:
Even if curvilinear grids are not supported yet for the non-hydrostatic model, I like the idea of moving `validate_momentum_advection` to `Models` like `validate_tracer_advection`.

xkykai:
> Even if curvilinear grids are not supported yet for the non-hydrostatic model, I like the idea of moving `validate_momentum_advection` to `Models` like `validate_tracer_advection`.

does this mean that I should actually write separate versions of `validate_momentum_advection` for `HydrostaticFreeSurfaceModels` and 'NonhydrostaticModel` instead?

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -192,6 +198,16 @@ function NonhydrostaticModel(;    grid,
...
 
+    default_tracer_advection, tracer_advection = validate_tracer_advection(tracer_advection, grid)
+
+    # Advection schemes
+    tracer_advection_tuple = with_tracers(tracernames(tracers),
+                                          tracer_advection,
+                                          (name, tracer_advection) -> default_tracer_advection,
+                                          with_velocities=false)
+
+    advection = merge((momentum=momentum_advection,), tracer_advection_tuple)
```
glwagner, line 209:
```suggestion
    advection = merge((; momentum=momentum_advection), tracer_advection_tuple)
```

simone-silvestri:
> > Even if curvilinear grids are not supported yet for the non-hydrostatic model, I like the idea of moving `validate_momentum_advection` to `Models` like `validate_tracer_advection`.
> 
> does this mean that I should actually write separate versions of `validate_momentum_advection` for `HydrostaticFreeSurfaceModels` and 'NonhydrostaticModel` instead?

Just moving `validate_momentum_advection` to the `Models` module is fine. In the end, `validate_momentum_advection` only depends on the grid and the advection scheme.

simone-silvestri:
This is a breaking change in user API so we need to bump the major version (in the `Project.toml` file) and change the examples to reflect the change.

navidcy (changes requested):
let's bump + change examples that use NonhydrostaticModel as @simone-silvestri suggested

(I just put this "Request changes" review so that the PR cannot be accidentally merged.)

xkykai:
I can change all instances in `./examples`. Should I change everything else too? If not which ones should I change?

glwagner:
Shouldn't we change _everything_? The things we don't change won't work, right? Are we stil planning to support the `advection` kwarg, or is it required that we separate `momentum_advection` and `tracer_advection` ?

glwagner:
It might help to write up in detail what the proposed API change is, and give a few examples of new + old usages

glwagner:
@xkykai I think this PR is stale. I'd be interested in taking it up though. But it might make sense to start from scratch? Because I would prefer to preserve the existing API (eg allow a single `advection` argument), and merely add the _option_ of separate momentum/tracer advection schemes.

glwagner:
I'll close it but please re-open if you'd like to continue

xkykai:
> @xkykai I think this PR is stale. I'd be interested in taking it up though. But it might make sense to start from scratch? Because I would prefer to preserve the existing API (eg allow a single `advection` argument), and merely add the _option_ of separate momentum/tracer advection schemes.

Perhaps we can do it like the `tracers` argument? ie: allow for `advection=WENO()` type input which applies to all tracers, or `advection = (u=WENO(), T=CenteredSecondOrder())` for custom choices for example.

glwagner:
Yes for sure, it is straightforward to support `advection=WENO()` I think. Do you agree we should start from scratch here? If not please reopen!

PR #3327, opened 2023-10-10
Add Enzyme extension

wsmoses:
This PR adds an extension package to Oceananigans to describing the custom derivative behavior of Oceananigans types and functions.

PR #3347, opened 2023-10-16
Move Enzyme extension tests to dedicated test group

glwagner:
Continues from #3327

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -554,6 +554,33 @@ steps:
     depends_on: "init_cpu"
 
 #####
+##### Enzyme extension tests
+#####
+
+  - label: "👺 gpu Enzyme extension tests"
+    env:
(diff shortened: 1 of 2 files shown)


glwagner:
cc @wsmoses

wsmoses:
FYI, the GPU Enzyme tests probably fail, because GPU reverse-mode rules are not added to KA (forward mode, however exists).

glwagner:
> FYI, the GPU Enzyme tests probably fail, because GPU reverse-mode rules are not added to KA (forward mode, however exists).

They did pass previously though! (GPU tests ran on #3327.) We'll see what happens here.

wsmoses:
Oh, actually that makes sense since the unit test here just does the field allocation, and doesn't hit a KA call that would result in an issue.

navidcy (approved):

PR #3349, opened 2023-10-18
Fix the problem that particles may be out of bounds when velocity is too large

Yixiao-Zhang:


src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -6,8 +6,17 @@ using Oceananigans.Models: total_velocities
 #####
 
 # Functions for bouncing particles off walls to the right and left
-@inline  bounce_left(x, xᴿ, Cʳ) = xᴿ - Cʳ * (x - xᴿ)
-@inline bounce_right(x, xᴸ, Cʳ) = xᴸ + Cʳ * (xᴸ - x)
+@inline function bounce_left(x, xᴸ, xᴿ, Cʳ)
+    xᵢ = xᴿ - Cʳ * (x - xᴿ)
+    # Limit the particle to the left boundary when it leaves from there
```
glwagner, line 11:
```suggestion
    # Keep the particle from bouncing so far left it leaves the domain
```

is that right?

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -6,8 +6,17 @@ using Oceananigans.Models: total_velocities
...
+    return ifelse(xᵢ > xᴸ, xᵢ, xᴸ)
```
glwagner, line 12:
```suggestion
    return ifelse(xᵢ < xᴸ, xᴸ, xᵢ)
```

I think this logic is a little more clear, as in

"if [bad things happen] then [do something special] else [proceed as normal]"

glwagner:
Thank you!

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -6,8 +6,17 @@ using Oceananigans.Models: total_velocities
...
+@inline function bounce_left(x, xᴸ, xᴿ, Cʳ)
+    xᵢ = xᴿ - Cʳ * (x - xᴿ)
+    # Keep the particle from bouncing so far left it leaves the domain
+    return ifelse(xᵢ < xᴸ, xᴸ, xᵢ)
+end
+
+@inline function bounce_right(x, xᴸ, xᴿ, Cʳ)
+    xᵢ = xᴸ + Cʳ * (xᴸ - x)
+    # Limit the particle to the right boundary when it leaves from there
+    return ifelse(xᵢ < xᴿ, xᵢ, xᴿ)
```
glwagner, line 18:
```suggestion
    # Keep the particle from bouncing so far right it leaves the domain
    return ifelse(xᵢ > xᴿ, xᴿ, xᵢ)
```

PR #3350, opened 2023-10-19
Update Docs/Model/Output Writers to replace `FieldSlicer` -> `indices`

xkykai:
replacing `FieldSlicer` in docs to `indices` so that it is up to date.


navidcy (approved):
lgtm

navidcy:
However, I think this page includes duplicate information from the output writers docstrings and thus makes maintaining the docs doubly difficult. We should consider somehow using the docstrings in this page to showcase the usage of the output writers. But until then, this PR is definitely an improvement.

navidcy:
thanks @xkykai !

PR #3353, opened 2023-10-20
Explore trying to warn users if the CFL is too big

glwagner:
cc @simone-silvestri


navidcy:
Is this supposed to supersede #3183?

glwagner:
not going anywhere so closing

Issue #3191, opened 2023-07-18
Docs don't build for tagged releases


navidcy:
See, e.g., recent v0.85.

Possibly something related with buildkite settings. E.g. for v0.84.1 the tag triggered [a buildkite run](https://buildkite.com/clima/oceananigans/builds/12019)

<img width="1215" alt="Screenshot 2023-07-18 at 1 57 24 pm" src="https://github.com/CliMA/Oceananigans.jl/assets/7112768/f0b0f59a-f909-4858-aaf1-3536ed485d91">

<img width="1234" alt="Screenshot 2023-07-18 at 1 58 19 pm" src="https://github.com/CliMA/Oceananigans.jl/assets/7112768/818d2fb1-086f-40ac-a109-52833375aa40">

@glwagner, something related to buildkite setting we were fiddling with lately?

glwagner:
Ok so we fixed the problem with the dev docs, but created a new one?

navidcy:
> Ok so we fixed the problem with the dev docs, but created a new one?

I think so… :(

navidcy:
@glwagner could you try to resolve this? sorry for the burden...

navidcy:
bumping this

glwagner:
Huh, I have no idea what to do to solve this

navidcy:
OK, no worries. We can always look at the `dev` version.

navidcy:
> Huh, I have no idea what to do to solve this

It must be buildkite-related.

glwagner:
I wonder if this is the problem:

https://github.com/CliMA/OceananigansDocumentation/actions/runs/6101820665

<img width="816" alt="image" src="https://github.com/CliMA/Oceananigans.jl/assets/15271942/26b5926b-7f8c-4889-b1c5-0f7923b74d31">

navidcy:
I think this is an additional independent problem. I can try fix this one and we see. I think though that the stable tagged versions don’t ever trigger docs built in the first place.

glwagner:
Something seemed to change with 0.81.1

<img width="928" alt="image" src="https://github.com/CliMA/Oceananigans.jl/assets/15271942/81cae28b-12ee-480f-b40c-1502701b97d5">

navidcy:
Deploy of gh-pages at OceananigansDocumentation repo seems OK now. I think it was just a glitch...
https://github.com/CliMA/OceananigansDocumentation/actions/runs/6109843276

[...]

PR #3355, opened 2023-10-23
(0.90.0) Omit entries in Flat direction from `node`

### LINKED ISSUE #3285: `node` should omit nodes from `Flat` dimensions

glwagner:
Closes #3285

This is a big API change so we'll have to bump a minor version.

This PR changes the behavior of the internal function `node`, which returns a tuple of coordinates at index `i, j, k` and location `lx, ly, lz`, to omit coordinates from the tuple along `Flat` directions. Previously, `node` omitted coordinates only for `Nothing` locations. Now, `Flat` directions are treated similarly to `Nothing` locations.

This change changes the API for using `set!` with functions, and therefore for setting initial conditions, building immersed boundaries, using boundary condition functions, and background fields.

Before this PR, one might write

```julia
grid = RectilinearGrid(size=64, z=(0, 1), topology=(Flat, Flat, Grid))
c = CenterField(grid)
random_numbers(x, y, z) = rand()
set!(c, random_numbers)
```

After this PR, this code has to change because the `x, y` directions are `Flat`. Thus after this PR we will write

```julia
grid = RectilinearGrid(size=64, z=(0, 1), topology=(Flat, Flat, Grid))
c = CenterField(grid)
random_numbers(z) = rand()
set!(c, random_numbers)
```

As a rule, coordinates within functions should never be provided for `Flat` directions or for `Nothing` locations.

There's more discussion about the pros and cons of this choice on #3285.


glwagner:
Should we also try things like

```julia
node(i, j, k, grid, ::Nothing, ℓy, ℓz) = _node(i, j, k, grid, nothing, , ℓy, ℓz)[1:2]
```

I think we determined there could be a tiny performance loss but it would make the code a little simpler and also easier to read since we don't have to define every combination of locations for `_node`.

I still want to clean up and streamline `xspacings` and `xnodes` (etc) but that'll have to wait for another PR I think.

glwagner:
> Should we also try things like
> 
> ```julia
> node(i, j, k, grid, ::Nothing, ℓy, ℓz) = _node(i, j, k, grid, nothing, , ℓy, ℓz)[1:2]
> ```
> 
> I think we determined there could be a tiny performance loss but it would make the code a little simpler and also easier to read since we don't have to define every combination of locations for `_node`.

Hmm this is more annoying than I thought so I'll leave it for later.

glwagner:
Ok, we now emit warnings like

```julia
julia> include("one_dimensional_diffusion.jl")
┌ Warning: An error was encountered within set! while setting the field
│
│     1×1×128 Field{Center, Center, Center} on RectilinearGrid on CPU
│
│ Note that to use set!(field, func::Function) on a field at location (Center, Center, Center)
│ and on a grid with topology (Flat, Flat, Bounded), func must be callable via
│
│      func(z)
└ @ Oceananigans.Fields ~/Projects/Oceananigans.jl/src/Fields/set!.jl:59
```

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -1171,6 +1159,14 @@ end
 @inline xnode(i, j, k, grid::OSSG, ℓx, ℓy, ℓz) = xnode(i, j, grid, ℓx, ℓy)
 @inline ynode(i, j, k, grid::OSSG, ℓx, ℓy, ℓz) = ynode(i, j, grid, ℓx, ℓy)
 
+# Definitions for node
+@inline ξnode(i, j, k, grid::OSSG, ℓx, ℓy, ℓz) = λnode(i, j, grid, ℓx, ℓy)
+@inline ηnode(i, j, k, grid::OSSG, ℓx, ℓy, ℓz) = φnode(i, j, grid, ℓx, ℓy)
+@inline rnode(i, j, k, grid::OSSG, ℓx, ℓy, ℓz) = znode(k, grid, ℓz)
+
+ξname(::OSSG) = :λ
+ηname(::OSSG) = :η
```
navidcy, line 1168: should this be `:φ`?

src/Grids/latitude_longitude_grid.jl
```diff
@@ -574,7 +574,33 @@ coordinates(::LatitudeLongitudeGrid) = (:λᶠᵃᵃ, :λᶜᵃᵃ, :φᵃᶠᵃ
 ##### Grid nodes
 #####
 
-function nodes(grid::LatitudeLongitudeGrid, ℓx, ℓy, ℓz; reshape=false, with_halos=false)
+ξname(::LLG) = :λ
+ηname(::LLG) = :η
```
navidcy, line 578:
```suggestion
ηname(::LLG) = :φ
```
?

navidcy (approved):
noice!

src/Fields/set!.jl
```diff
@@ -19,16 +19,51 @@ function set!(u::Field, v)
...
+        loc_str = tuple_string(location(u))
+        topo_str = tuple_string(topology(u.grid))
+
+        msg = string("An error was encountered within set! while setting the field", '\n', '\n',
+                     "    ", prettysummary(u), '\n', '\n',
+                     "Note that to use set!(field, func::Function) on a field at location ",
+                     "(", loc_str, ")", '\n',
+                     "and on a grid with topology (", topo_str, "), func must be ",
+                     "callable via", '\n', '\n',
+                     "     func(", arg_str, ")", '\n')
```
tomchor, line 59: :heart:

tomchor:
Since this is such a major change in the interface, I think we should get the docs back up again before registering 0.90 (i.e. fix https://github.com/CliMA/Oceananigans.jl/issues/3191).

navidcy:
We don’t know how to do that unfortunately

glwagner:
Happy to hold off registering 0.90.0.

tomchor:
> That’s a good point. Unfortunately, we don’t know how to do that unfortunately. We tried bunch of things. I’ll give it a go.

Can't we simply revert the commit/PR that introduced the issue?

glwagner:
It's not due to a commit (as far as we can tell). It's because we changed a buildkite setting.

navidcy:
It bothers me this issue... I'll try to get to the bottom of it.

navidcy:
I changed a setting. I don't know how to test it though other than registering a new version... But what if it doesn't work?

glwagner:
I think it's not the end of the world if we go a few days without docs. It's a risk but it's not worth losing sleep and mental health over this

navidcy:
Shall I register and see?

navidcy:
I'm doing it.

PR #3359, opened 2023-10-23
Explicit `OffsetNDRange` vs `NDRange`

simone-silvestri:
previously `OffsetNDRange` (in Oceananigans) was assuming that KernelAbstractions' `NDRange` could not have a `Tuple` as the last parameter. 

This PR introduces an Oceananigans-specific `KernelOffsets` type that is necessary to unambiguously identify Oceananigans' offset `NDRange`s from KernelAbstractions' native `NDRange`s

cc @glwagner


glwagner (approved):
perfect, thank you 🙏

PR #3362, opened 2023-10-25
Fix advection timescale calculation

### LINKED ISSUE #3357: Should the time step wizard take into account of background velocity in calculation of CFL number?
### Discussed in https://github.com/CliMA/Oceananigans.jl/discussions/3354

<div type='discussions-op-text'>

<sup>Originally posted by **zhihua-zheng** October 22, 2023</sup>
Hi,

I'm running a case similar to this [example](https://github.com/OceanBioME/OceanBioME.jl/blob/cdb79a3236461e5c32a916639b10dc8da997eae1/paper/figures/eady.jl), which has a background velocity from the thermal wind shear. However, I find this setup usually has some numerical instabilities at the beginning. I was wondering if that's related to the time step wizard's adjustment of `Δt` being too large. From the calculation of `cell_advection_timescale` here,
https://github.com/CliMA/Oceananigans.jl/blob/d0b7ec8f98c860ce49927e0a7214961d2f47fb75/src/Advection/cell_advection_timescale.jl#L13-L27
it seems like it does not use the background velocity. 

For this use case, should the time step wizard include the background velocity to calculate a new `Δt`?</div>

zhihua-zheng:
This PR improves the calculation of `cell_advection_timescale` by including velocities in `background_field`. This fixes #3357, as discussed in #3354.

Code change suggested by @glwagner.


src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -221,8 +221,9 @@ function inflate_grid_halo_size(grid, tendency_terms...)
...
-@inline total_velocities(m::NonhydrostaticModel) =
-    (u = SumOfArrays{2}(m.velocities.u, m.background_fields.velocities.u),
-     v = SumOfArrays{2}(m.velocities.v, m.background_fields.velocities.v),
-     w = SumOfArrays{2}(m.velocities.w, m.background_fields.velocities.w))
-
+@inline function total_velocities(m::NonhydrostaticModel)
+    velocities = (u = SumOfArrays{2}(m.velocities.u, m.background_fields.velocities.u),
+                  v = SumOfArrays{2}(m.velocities.v, m.background_fields.velocities.v),
+                  w = SumOfArrays{2}(m.velocities.w, m.background_fields.velocities.w))
+    return with_advective_forcing(m.forcing, velocities)
```
glwagner, line 228: I don't think this will work. `m.forcing` is a `NamedTuple` that contains different `Forcing` for every prognostic field (`forcing.u`, `forcing.v`, etc). When we call `with_advective_forcing`, we add the forcing per component, ie:

https://github.com/CliMA/Oceananigans.jl/blob/cc51ff0ee0a1326cebcbfa207ec0ed29ae768648/src/Models/NonhydrostaticModels/nonhydrostatic_tendency_kernel_functions.jl#L69

To be strictly correct we might have to consider all "effective" velocities experienced by each component, and take the minimum of all of those.

Since that's complicated (and unclear whether we even want to spare the expense for that kind of calculation), I'd leave it for the future. Typically, "additional" forcing-associated velocities (like settling velocities) are known in advance and could be incorporated into the user determination of the cfl.

glwagner, line 228:
```suggestion
    return velocities
```

glwagner: I guess it's also unclear what the "total velocities" mean when different tracers can have different effective velocities. Maybe it's better to leave the forcing-related velocity out of total velocities for clarity...

glwagner:
Awesome, thank you! I think we probably want to leave `total_velocities` as it was previously, as noted in my review. Otherwise it looks like a great improvement to `cell_advection_timescale`.

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
zhihua-zheng: That sounds good, I will remove the second commit related to `AdvectiveForcing`.

zhihua-zheng:
Change on `total_velocities` removed.

glwagner (approved):

glwagner:
Thanks @zhihua-zheng !

PR #3365, opened 2023-10-27
Better defaults for WENOVectorInvariant

glwagner:
This PR incorporates some of @simone-silvestri's suggestions for default that didn't make it into #3341

### DIFF
--- a/src/Advection/vector_invariant_advection.jl
+++ b/src/Advection/vector_invariant_advection.jl
@@ -170,6 +170,8 @@ Base.show(io::IO, a::VectorInvariant{N, FT}) where {N, FT} =
 #                           VectorInvariant{N,     FT,    M,     Z (vorticity scheme),      ZS,    V (vertical scheme),      K (kinetic energy gradient scheme)
 const WENOVectorInvariant = VectorInvariant{<:Any, <:Any, <:Any, <:WENO, <:Any, <:WENO, <:WENO}
 
+nothing_to_default(user_value, default) = isnothing(user_value) ? default : user_value
+
 """
     function WENOVectorInvariant(; upwinding = nothing,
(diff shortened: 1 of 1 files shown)


src/Advection/vector_invariant_advection.jl
```diff
@@ -178,19 +180,34 @@ const WENOVectorInvariant = VectorInvariant{<:Any, <:Any, <:Any, <:WENO, <:Any,
...
+        kinetic_energy_gradient_order = nothing_to_default(kinetic_energy_gradient_order, default=order)
     end
 
+    vorticity_scheme               = WENO(; order=vorticity_order, weno_kw...)
+    vertical_scheme                = WENO(; order=vertical_order, weno_kw...)
+    kinetic_energy_gradient_scheme = WENO(; order=kinetic_energy_gradient_order, weno_kw...)
+    divergence_scheme              = WENO(; order=divergence_order, weno_kw...)
+
+    default_upwinding = OnlySelfUpwinding(cross_scheme=divergence_scheme)
+    upwinding = nothing_to_default(upwinding; default = default_upwinding)
```
glwagner, line 209: @simone-silvestri it looks like we only have one cross scheme, but it is used for _both_ kinetic energy and divergence. Is that right? Does it still make sense to treat the advection scheme for kinetic energy and divergence separately?

simone-silvestri: We could give that option for experimenting.

simone-silvestri (approved):

src/Advection/vector_invariant_advection.jl
glwagner: Ok, not for this PR then

PR #3367, opened 2023-10-29
Fixes bug with pre-defining diffusivity fields

jagoosw:
Currently the user may pass `diffusivity_fields` into model constructors, but `Oceananigans.TurbulantClosures.DiffusivityFields` resets these to the value of `validate_field_tuple_grid` which nominally is `nothing`. I have changed this to return the field values if `validate_field_tuple_grid` does not error instead.

E.g.

```julia

using Oceananigans
using Oceananigans.TurbulenceClosures: DiffusivityFields

grid = RectilinearGrid(size = (64, 64, 16), extent = (150, 150, 10))

tracers = (:T, :S)

closure = AnisotropicMinimumDissipation()

diffusivity_fields = diffusivity_fields = DiffusivityFields(grid, tracers, NamedTuple(), closure)

                              tracers,
                              closure,
                              diffusivity_fields)
```


src/TurbulenceClosures/diffusivity_fields.jl
```diff
@@ -4,8 +4,8 @@ using Oceananigans.Fields: validate_field_tuple_grid
 ##### Forms for NonhydrostaticModel constructor
 #####
 
-DiffusivityFields(diffusivities::NamedTuple, grid, tracer_names, bcs, closure) =
-    validate_field_tuple_grid("diffusivities", diffusivities, grid)
+DiffusivityFields(diffusivities::NamedTuple, grid, tracer_names, bcs, closure) = 
+    isnothing(validate_field_tuple_grid("diffusivities", diffusivities, grid)) && diffusivities
```
glwagner, line 8: can you write this on more than one line?

src/TurbulenceClosures/diffusivity_fields.jl
```diff
@@ -4,8 +4,11 @@ using Oceananigans.Fields: validate_field_tuple_grid
 ##### Forms for NonhydrostaticModel constructor
 #####
 
-DiffusivityFields(diffusivities::NamedTuple, grid, tracer_names, bcs, closure) =
+function DiffusivityFields(diffusivities::NamedTuple, grid, tracer_names, bcs, closure)
     validate_field_tuple_grid("diffusivities", diffusivities, grid)
+    
+    return diffusivities
```
glwagner, line 10: Does this have to be 

```julia
diffusivities = validate_field_tuple_grid("diffusivities", diffusivities, grid)
```

I haven't seen this `validate_field_tuple_grid` function before, it looks kinda sketchy... (I can't figure out what it does based on its name)

@simone-silvestri

jagoosw: No that was the problem, like the other validation functions it just returns nothing if they're valid or errors if not.

As far as I can see this is the only instance where this function is used and the string is just used for the error message:
https://github.com/CliMA/Oceananigans.jl/blob/f3996c2703bb23b2e84f80c6f1564fc7965bdad1/src/Fields/field_tuples.jl#L119-L132

If it were to remain like that it probably would be a good idea to change the string to `diffusivity_fields` since that's what they're called in the model constructor.

glwagner: Yes, with that change I think we can merge this PR.

Some of the `validation` functions return a new copy of the validated objects, especially in grid constructors...

src/TurbulenceClosures/diffusivity_fields.jl
```diff
@@ -4,8 +4,11 @@ using Oceananigans.Fields: validate_field_tuple_grid
 ##### Forms for NonhydrostaticModel constructor
 #####
 
-DiffusivityFields(diffusivities::NamedTuple, grid, tracer_names, bcs, closure) =
+function DiffusivityFields(diffusivities::NamedTuple, grid, tracer_names, bcs, closure)
     validate_field_tuple_grid("diffusivities", diffusivities, grid)
```
glwagner, line 8:
```suggestion
    validate_field_tuple_grid("diffusivity_fields", diffusivities, grid)
```

src/TurbulenceClosures/diffusivity_fields.jl
```diff
@@ -4,8 +4,11 @@ using Oceananigans.Fields: validate_field_tuple_grid
 ##### Forms for NonhydrostaticModel constructor
 #####
 
-DiffusivityFields(diffusivities::NamedTuple, grid, tracer_names, bcs, closure) =
+function DiffusivityFields(diffusivities::NamedTuple, grid, tracer_names, bcs, closure)
```
glwagner, line 7:
```suggestion
function DiffusivityFields(diffusivity_fields::NamedTuple, grid, tracer_names, bcs, closure)
```

src/TurbulenceClosures/diffusivity_fields.jl
glwagner, line 10:
```suggestion
    return diffusivity_fields
```

glwagner (approved):

src/TurbulenceClosures/diffusivity_fields.jl
navidcy: that's good

navidcy:
Should this be merged?

glwagner:
go for it @jagoosw

PR #3376, opened 2023-11-04
Default NaN-Checker for Particle Tracking models

simone-silvestri:
Particle tracking models have no prognostic fields and no possibility of generating a NaN. Therefore, this PR disables the nan-checker


src/Models/Models.jl
```diff
@@ -162,13 +162,25 @@ end
 # Check for NaNs in the first prognostic field (generalizes to prescribed velocities).
 function default_nan_checker(model::OceananigansModels)
     model_fields = prognostic_fields(model)
+
+    if isempty(model_fields) 
+        @warn "The NaNChecker is disabled because this model does not evolve prognostic variables."
```
glwagner, line 167:
```suggestion
```

Don't think we need a warning...

src/Models/Models.jl
```diff
@@ -162,13 +162,25 @@ end
...
     nan_checker = NaNChecker(field_to_check_nans)
     return nan_checker
 end
 
-# This is here so that `NonhydrostaticModel` and  `HydrsostaticFreeSurfaceModel`
+using Oceananigans.Models.HydrostaticFreeSurfaceModels: OnlyParticleTrackingModel
+
+# Particle tracking models with prescribed velocities (and no tracers) 
+# have no prognostic fields and no chance to producing a NaN.
+default_nan_checker(::OnlyParticleTrackingModel) = nothing
```
glwagner, line 181: Why do we need this _and_ the code above? Also is this PR only about particle tracking models or is more going on?

src/Models/Models.jl
```diff
@@ -162,13 +162,25 @@ end
...
+
+# General seawater density calculation applicable to both `NonhydrostaticModel` and  `HydrostaticFreeSurfaceModel`
```
glwagner, line 183:
```suggestion
# Implementation of a `seawater_density` `KernelFunctionOperation
# applicable to both `NonhydrostaticModel` and  `HydrostaticFreeSurfaceModel`
```

glwagner (approved):

src/Models/Models.jl
simone-silvestri: I can think only at `OnlyParticleTrackingModel` as a case of a model without prognostic fields but there might be other instances. It is nice to explicitly state that `OnlyParticleTrackingModel` does not have a NaNChecker

glwagner: I don't think statements about the NaNChecker are important. I think other aspects of the simulation are much more important. Since printing comes at a cost, if we want to print more, we should focus on the things that matter.

glwagner: I would think about the existing of the NaNChecker differently. If you wanted to print something, you should print when you _add_ the NaNChecker (not when you remove it). Printing something when you remove the NaNChecker is probably noise for most people, they have no idea what the NaNChecker is to begin with.

src/Models/Models.jl
glwagner: Also, why is this a warning? Is there some reason to be careful about? With no prognostic fields, I'm confused what a NaNChecker would do.

src/Models/Models.jl
simone-silvestri: Ok, you convinced me, I have removed the warining

glwagner (approved):

PR #3387, opened 2023-11-13
Fix docs for `@apply_regionally`

### LINKED ISSUE #3386: `compute_regionally` must be deprecated
@simone-silvestri, what is `compute_regionally` that the docstring here is referring to?

https://github.com/CliMA/Oceananigans.jl/blob/70536571523ad2eb71fd9a2200121eca63998ac4/src/Utils/multi_region_transformation.jl#L196

Legacy/Deprecated perhaps?

navidcy:
Closes #3386

### DIFF
--- a/src/Utils/multi_region_transformation.jl
+++ b/src/Utils/multi_region_transformation.jl
@@ -192,9 +192,10 @@ end
 """
     @apply_regionally expr
     
-Use `@apply_regionally` to distribute locally the function calls.
-Call `compute_regionally` in case of a returning value and `apply_regionally!` 
-in case of no return.
+Use `@apply_regionally` to distribute locally the function calls in `expr`ession
+when the functions do not return anything.
+
+In case the function in `expr` returns something, then use [`construct_regionally`](@ref).
 """
 macro apply_regionally(expr)
     if expr.head == :call
@@ -209,7 +210,7 @@ macro apply_regionally(expr)
     elseif expr.head == :(=)
         ret = expr.args[1]
         Nret = 1
-        if expr.args[1] isa Expr 
+        if expr.args[1] isa Expr
             Nret = length(expr.args[1].args)
         end
         exp = expr.args[2]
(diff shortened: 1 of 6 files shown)


navidcy:
@simone-silvestri  could you have a look?
I think we should add a docstring to `calculate_regionally`.

src/Utils/multi_region_transformation.jl
```diff
@@ -192,9 +192,10 @@ end
 """
     @apply_regionally expr
     
-Use `@apply_regionally` to distribute locally the function calls.
-Call `compute_regionally` in case of a returning value and `apply_regionally!` 
-in case of no return.
+Use `@apply_regionally` to distribute locally the function calls in `expr`ession
+when the functions do not return anything.
+
+In case the function in `expr` returns something, then use [`construct_regionally`](@ref).
```
simone-silvestri, line 198:
```suggestion
Distributes locally the function calls in `expr`ession

It calls [`apply_regionally!`](@ref) when the functions do not return anything.

In case the function in `expr` returns something, `@apply_regionally` calls [`construct_regionally`](@ref).
```

navidcy: Hm... so I'm confused. Actually the `@apply_regionally` will call `construct_regionally` if the function does not return anything? Is it better to say "fallsback to `construct_regionally` if the function in `expr` returns something."?

Isn't this but prone for bugs?

What will happen with this expression:

```Julia
function foo!(bar)
    bar = 2bar
end
```
versus this:

```Julia
function fooo!(bar)
    bar = 2bar
    return nothing
end
```
?

simone-silvestri: the difference is between
```julia
@apply_regionally ret = returning_function(args)
```
and
```julia
@apply_regionally non_returning_function!(args)
```
the first one calls `construct_regionally` because there is an equal sign, implying a return value. The second is a function call without a return value so it will call `apply_regionally!`

PR #3388, opened 2023-11-13
Bugfix for particles in `ImmersedBoundaryGrid`s

simone-silvestri:
Particles were not working in `ImmersedBoundaryGrid`s.
This PR provides a fix


glwagner (approved):

glwagner:
I think you should add a simple test so this doesn't regress again

simone-silvestri:
I believe this is stale

PR #3391, opened 2023-11-14
Changes property "longname" to "long_name" for NetCDFOutputWriter

tomchor:
As far as I can tell the property `longname` isn't really standard. According to [CF-coventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03s03.html) the property should be called `long_name`. This will make several more things work out of box now, like `xarray` plot naming (as expected when sticking to conventions).


glwagner:
> (as expected when sticking to conventions).

It's worth pointing out this was probably a mistake rather than an intentional violation of convention. Thanks for the improvement.

PR #3392, opened 2023-11-15
Correction of three-dimensional Stokes drift implementation

BrodiePearson:
@glwagner This PR completes (hopefully!) the implementation of three-dimensional Stokes drift implementation within non-hydrostatic simulations by adding the momentum forcing terms that arise from horizontal gradients of the Stokes drift via $(\nabla \times \mathbf{u^s}) \times \mathbf{v}$. 

I also added an example code that utilizes the new `StokesDrift` function (and retains the previous `UniformStokesDrift` function). It is a little convoluted because I wanted to implement a Stokes drift that varied in all 3 dimensions to ensure that the code was well-behaved. The example is as follows:

- Is based on the Langmuir turbulence example
- The new example imposes a spatially-varying x-aligned Stokes drift that is a separable function of the three spatial dimensions

```math
u^s(x, y, z) = f(x)g'(y)h(z)
```
- The vertical variation is a classic exponential decay with depth over a vertical length scale $\delta$. Note this function is defined as the z-derivative of another function $h(z)$ to make the description of the vertical Stokes drift easier below.

```math
h'(z) = U^s e^{(z/\delta)}
```

- The x-variation is sinusoidal modulation on top of a background that peaks in the middle of the x-domain.

```math
f(x) = \frac{1}{2}\left( 1 + 0.1cos\left(\frac{\pi(2x-L_x)}{L_x}\right)\right)
```

- The y-variation represents a piece-wise function consisting of a _jet_ region of width $L_{jet}$ centered at $y_{jet}$ where $u^s$ is constant, transition regions of width $L_{transition}$ on either side of the jet where Stokes drift follows a sinusoidal decay from its jet value to zero, and a _no wave_ region where there is no Stokes drift outside of the jet and transition zones.
```math
 g(y) =    1 + cos\left(\pi \frac{y - y_{jet} + L_{jet}/2 }{L_{transition}} \right)       \quad \text{if } y_{jet} - L_{jet}/2  > y >  y_{jet} - L_{jet}/2 - L_{transition} 
```

```math
 g(y) = 1 + cos\left(\pi \frac{y - y_{jet} - L_{jet}/2 }{L_{transition}} \right)       \quad \text{if } y_{jet} + L_{jet}/2 + L_{transition}  > y >  y_{jet} + L_{jet}/2
```

```math
  g(y) = 0  \quad \text{at all other } y
```

- This x-aligned Stokes drift has $\partial_x u^s\neq0$ so incompressibility ($\nabla \cdot \mathbf{u^s}$) requires another Stokes drift component. Since we do not impose a y-oriented wave field, we achieve this through a vertical Stokes drift component $w^s(x,y,z)$. We diagnose this from the $u^s$ field and by imposing a surface boundary condition of no vertical motion: $w^s(x,y,0)=0$ . It then follows that:

```math
w^s(x, y, z) = f'(x)g(y)\left[h(z)-h(0)\right]
```

### DIFF
--- a/examples/Spatially_varying_stokes_drift.jl
+++ b/examples/Spatially_varying_stokes_drift.jl
@@ -1,0 +1,539 @@
+# # Langmuir turbulence example
+#
+# This example implements a Langmuir turbulence simulation reported in section
+# 4 of
+#
+# > [Wagner et al., "Near-inertial waves and turbulence driven by the growth of swell", Journal of Physical Oceanography (2021)](https://journals.ametsoc.org/view/journals/phoc/51/5/JPO-D-20-0178.1.xml)
+#
+# This example demonstrates
+#
+#   * How to run large eddy simulations with surface wave effects via the
+#     Craik-Leibovich approximation.
+#
+#   * How to specify time- and horizontally-averaged output.
+
+# ## Install dependencies
+#
+# First let's make sure we have all required packages installed.
+
+# ```julia
+# using Pkg
+# pkg"add Oceananigans, CairoMakie"
+# ```
+
+using Oceananigans
+using Oceananigans.Units: minute, minutes, hours
+
+# ## Model set-up
+#
+# To build the model, we specify the grid, Stokes drift, boundary conditions, and
+# Coriolis parameter.
+#
+# ### Domain and numerical grid specification
+#
+# We use a modest resolution and the same total extent as Wagner et al. 2021,
+
+grid = RectilinearGrid(size=(32, 64, 32), extent=(128, 256, 64))
+
+# ### The Stokes Drift profile
+#
+# The surface wave Stokes drift profile prescribed in Wagner et al. 2021,
+# corresponds to a 'monochromatic' (that is, single-frequency) wave field.
+#
+# A monochromatic wave field is characterized by its wavelength and amplitude
+# (half the distance from wave crest to wave trough), which determine the wave
+# frequency and the vertical scale of the Stokes drift profile.
+
+using Oceananigans.BuoyancyModels: g_Earth
+
+ amplitude = 0.8 # m
+wavelength = 60  # m
+wavenumber = 2π / wavelength # m⁻¹
+ frequency = sqrt(g_Earth * wavenumber) # s⁻¹
+
+## The vertical scale over which the Stokes drift of a monochromatic surface wave
+## decays away from the surface is `1/2wavenumber`, or
+const vertical_scale = wavelength / 4π
+
+## Stokes drift velocity at the surface
+const Uˢ = amplitude^2 * wavenumber * frequency # m s⁻¹
+
+stokes_jet_center = 70
+stokes_jet_central_width = 40
+stokes_jet_edge_width = 40
+
+# The `const` declarations ensure that Stokes drift functions compile on the GPU.
+# To run this example on the GPU, include `GPU()` in the
+# constructor for `RectilinearGrid` above.
+#
+# The Stokes drift profile is
+
+#uˢ(x, y, z, t) = Uˢ * exp(z / vertical_scale) * exp( - (y - stokes_jet_center)^2 / (2 * stokes_jet_width^2) )
+
+# Create a Stokes drift field that is a cosine function within a subregion of the domain.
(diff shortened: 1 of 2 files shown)


BrodiePearson:
@glwagner Sorry for the additional PR, I was having trouble with permissions stopping me from contributing directly to your PR.  

Here is a snapshot of the simulation with Stokes drift details:

image

navidcy:
I'll add you as collaborator so you can commit to PRs next time!

navidcy (approved):
I'm approving this PR so we can merge it with #3384 and continue working over there!

navidcy:
As soon as tests pass I'll merge and then we can continue in #3384, how does that sound @BrodiePearson?

You should have rights to push commits there now! I sent you an invite for project collaborator.

glwagner:
Nice! Formally, I think those horizontal derivative terms have to be small for the validity of the CL derivation. In other words, if `L` is the scale at which the envelope varies, then we need `kL << 1` for slowly modulated waves where `k` is the wavenumber.

That said I do think we have to include them to be able to model the "return flow" of McIntyre 1981...

BrodiePearson:
That sounds great to me @navidcy - thanks for adding me and for the collaborator invite!

glwagner:
I don't think we should add a new example for this feature though. Perhaps we can include a script as a reference in `validation`? Another possibility is to include an example in the docstring that sets up the Stokes drift profile described here, which will preserve it for future generations as well.

glwagner:
Ah, but since this PR doesn't go to main, we can keep the example for now and switch things up in the other PR!

glwagner (approved):

glwagner:
I may have misunderstood... but the Stokes drift is not divergence free, so we can't compute the vertical component from incompressibility, right? See https://royalsocietypublishing.org/doi/pdf/10.1098/rsta.2021.0032.

glwagner:
Actually, I am now realizing that this paper:

https://royalsocietypublishing.org/doi/pdf/10.1098/rsta.2021.0032

may contain the key to describing what we are doing here.

However, we have to read that carefully to understand how to prescribe the "solenoidal Stokes drift" for modulated waves. For `UniformStokesDrift` the only thing we miss is a mean vertical advection of the whole domain, which I believe is dynamically inert. For modulated waves, we may need to carefully re-define the solenoidal Stokes drift (relative to its classical definition) for these simulations to correctly represent Craik-Leibovich.

BrodiePearson:
@glwagner That's a good point about the CL theory limits (did you mean $kL \gg 1$ for validity?). Is it feasible that people could want to simulate a large domain where that condition is satisfied (i.e., Stokes drift variation over kms)? 

Also, based on your last comment should we just remove the vertical component of Stokes drift? It only appears in the Stokes-related forcing through its horizontal derivatives

glwagner:
Ok... see here

image

this shows what to do.

@BrodiePearson I think it could be nice to have a worked example. However, rather than starting from the definition of the Stokes drift, we need to start from a definition of the wave field. Next, we can compute the solenoidal Stokes drift as defined above. Note that for the solenoidal Stokes drift, it _is_ in fact valid to use non-divergence to compute the vertical component --- it is non-divergent by definition.

The solenoidal Stokes drift is precisely what we want, because our pressure solver enforced div (uL) = 0. So, using the solenoidal Stokes drift is consistent with our solver.

I think the differences will be small for slowly modulated waves, and probably negligible for cases where turbulence is the main thing going on. For other cases where we are investigating the "wave-induced flow", the differences may be noticeable.

glwagner:
Ah well I retract my comments about the validity of theory, based on this new information! We are good to go --- provided that the Stokes drift we use is the solenoidal Stokes drift. In that case, we can cope with rapid horizontal variation of the Stokes drift it seems, and still model a non-divergence Lagrangian-mean flow. But we should be clear about this in the documentation for this new type StokesDrift.

> Also, based on your last comment should we just remove the vertical component of Stokes drift? It only appears in the Stokes-related forcing through its horizontal derivatives

I think we _need_ the vertical component to retain consistency in the case of modulated waves.

glwagner:
ok let's continue on the other PR

PR #3393, opened 2023-11-17
Add link to arXiv for scalings in README

navidcy:
What do you both think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@
 
 Oceananigans is a fast, friendly, flexible software package for finite volume simulations of the nonhydrostatic
 and hydrostatic Boussinesq equations on CPUs and GPUs.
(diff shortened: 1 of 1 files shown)


glwagner:
seems relevant

glwagner (approved):

glwagner:
It looks like GPU tests did not run for this PR. Any idea why it was merged?

navidcy:
They were run!
https://buildkite.com/clima/oceananigans/builds/13516

I'm not sure why there are two builds....
@glwagner probably you are referring to 
https://buildkite.com/clima/oceananigans/builds/13574
right?

glwagner:
Hmm ok I didn't notice that. It makes sense too because I think the current problems on `main` are from https://github.com/CliMA/Oceananigans.jl/commit/1b4736614d3528d6a7b5cad00459bacad580052c, which succeeded this PR.

PR #3395, opened 2023-11-17
(0.90.5) Refactor `Fields.interpolate` to make sense for `Nothing` locations and `Flat` directions - take 2

simone-silvestri:
By mistake it reverted the previous PR, so the work in #3368 will continue here

PR #3396, opened 2023-11-17
Add custom forcing velocity to particles

simone-silvestri:
With @xkykai we were looking at ways to add a sinking velocity to particles to be added to the advecting velocity.

This PR introduces a framework to do this by adding a `particle_forcing_u` function.

The usage will be something like this:
```julia
struct SinkingParticle{T} <: AbstractParticle
    x :: T
    y :: T
    z :: T
    sinking_velocity :: T
end

import Oceananigans: particle_advective_forcing_w

@inline particle_advective_forcing_w(particles::SinkingParticle, p) = particles.sinking_velocity[p]
```


glwagner:
Wasn't this already possible?

glwagner:
You should add a property `forcing` to `LagrangianParticles`:

https://github.com/CliMA/Oceananigans.jl/blob/7f5ddf7eb675ddf5029878ebef01fc1c6cd5f96f/src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl#L38

simone-silvestri:
right, `particle` is a list of particles though. `p` is the index that we have to pass in. 
we can call it forcing but it is specifically an advective forcing (or velocity forcing) because it is added to the velocity

jagoosw:
> But this deviates from the other interfaces we provide.
> 

I think we kind of have a difference in the interface between particles and body forces. For tracers etc we always launch a kernel over every cell, get the arguments for the forcing at that point, and pass that to forcing functions. But for particles, the only way to add forcings (without making custom structs) is through the `custom_dynamics` where we just give the users forcing function all of the particles together and leave them to e.g. launch a kernel over all of them. 

It might be helpful to add this kind of forcing more similar to the first interface and ask users to define slip velocity functions with arbitrary arguments e.g. `w(x, y, z, t, $\rho$, d)`, and then we make the inside of the lagrangian particle advection kernel pass the correct particle properties to the function.

I don't think it would be as simple to generalise particle forcings because we don't know what users want to do with the properties and in the current setup couldn't assume that they want every particle property integrated from some tendency function.

simone-silvestri:
The problem with custom forcing is that the `dynamics` function is not implemented where the particle is actually advected (inside the particle advection kernel). 
We need a way to calculate a particle velocity that differs from the fluid velocity (i.e. imagine particles with inertia).

glwagner:
> right, `particle` is a list of particles though. `p` is the index that we have to pass in. we can call it forcing but it is specifically an advective forcing (or velocity forcing) because it is added to the velocity

That's a detail and can be implemented regardless of the type of interface.

The difference is

1. Users _extend_ an Oceananigans function to implement a forcing
2. Users pass a function into `ParticleAdvectiveForcing` to implement a forcing

The first choice is fine. However, the rest of our interface uses option 2. I think it makes sense to be consistent. It's also easier for people who don't know julia very well.

glwagner:
> I think we kind of have a difference in the interface between particles and body forces. For tracers etc we always launch a kernel over every cell, get the arguments for the forcing at that point, and pass that to forcing functions. But for particles, the only way to add forcings (without making custom structs) is through the `custom_dynamics` where we just give the users forcing function all of the particles together and leave them to e.g. launch a kernel over all of them.

Agree, but what point are you trying to make? I don't understand.

> 
> It might be helpful to add this kind of forcing more similar to the first interface and ask users to define slip velocity functions with arbitrary arguments e.g. `w(x, y, z, t, $\rho$, d)`, and then we make the inside of the lagrangian particle advection kernel pass the correct particle properties to the function.

This is basically what I was suggesting.

> 
> I don't think it would be as simple to generalise particle forcings because we don't know what users want to do with the properties and in the current setup couldn't assume that they want every particle property integrated from some tendency function.

I don't understand what you're trying to say. Can you elaborate or give an example?

glwagner:
> The problem with custom forcing is that the `dynamics` function is not implemented where the particle is actually advected (inside the particle advection kernel). We need a way to calculate a particle velocity that differs from the fluid velocity
> 
> for example, particles with inertia that follow dvdt=1τp(u−v) where v is the particle velocity and u is the fluid velocity can be implemented as:

First of all, I think it's great to implement a better interface for forcing particles. So I'm on board with this.

You could probably implement this kind of thing by storing the previous particle velocity inside the particle struct. So I'm pretty sure you could also implement these dynamics with the current interface.

As I said though, I agree it's a good idea to make this kind of thing easier.

glwagner:
> > The problem with custom forcing is that the `dynamics` function is not implemented where the particle is actually advected (inside the particle advection kernel). We need a way to calculate a particle velocity that differs from the fluid velocity
> > for example, particles with inertia that follow dvdt=1τp(u−v) where v is the particle velocity and u is the fluid velocity can be implemented as:
> 
> First of all, I think it's great to implement a better interface for forcing particles. So I'm on board with this.
> 
> You could probably implement inertia (and many other things) by storing the previous particle velocity (or whatever properties you needed for correct dynamics) inside the particle struct. So I'm pretty sure you could also implement these dynamics with the current interface.
> 
> As I said though, I agree it's a good idea to make this kind of thing easier.

Actually I realized that `dynamics` is called _before_ advection, which makes it more annoying:

https://github.com/CliMA/Oceananigans.jl/blob/dd893745e50fd7aef945f765df929d2a7510db6e/src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl#L123-L132

That makes it a little trickier. It's still possible, since you can compute a "fake particle update", and then add the inertia term, before the "real update" that occurs in `advect_lagrangian_particles!`. But it's complicated.

Why do we call `dynamics` before advection? Should we switch the order of those two?

glwagner:
> > > The problem with custom forcing is that the `dynamics` function is not implemented where the particle is actually advected (inside the particle advection kernel). We need a way to calculate a particle velocity that differs from the fluid velocity
> > > for example, particles with inertia that follow dvdt=1τp(u−v) where v is the particle velocity and u is the fluid velocity can be implemented as:
> > 
> > 
> > First of all, I think it's great to implement a better interface for forcing particles. So I'm on board with this.
> > You could probably implement inertia (and many other things) by storing the previous particle velocity (or whatever properties you needed for correct dynamics) inside the particle struct. So I'm pretty sure you could also implement these dynamics with the current interface.
> > As I said though, I agree it's a good idea to make this kind of thing easier.
> 
> Actually I realized that `dynamics` is called _before_ advection, which makes it more annoying:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/dd893745e50fd7aef945f765df929d2a7510db6e/src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl#L123-L132
> 
> That makes it a little trickier. It's still possible, since you can compute a "fake particle update", and then add the inertia term, before the "real update" that occurs in `advect_lagrangian_particles!`. But it's complicated.
> 
> Why do we call `dynamics` before advection? Should we switch the order of those two? It also seems like we should update particle properties _after_ we advect the particles.

simone-silvestri:
We could think about a `ParticleAdvectiveForcing` or also a `ParticleVelocityTendency` function that specifies the rhs of
$$\frac{d \boldsymbol{v}}{dt} = \boldsymbol{G}$$

simone-silvestri:
What is the use of the `dynamics` function?

xkykai:
> We could think about a `ParticleAdvectiveForcing` or also a `ParticleVelocityTendency` function that specifies the rhs of dvdt=G

One example that this might be tricky for this approach is where the particles have a constant sinking velocity on top of the fluid velocity.

glwagner:
> What is the use of the `dynamics` function?
> 
> ok I see, if we want to remove or add particles or other non-advective phenomena. I think, yes it should be after advection

It's a catch all that can be used to implement a wide variety of physics. Basically a callback.

glwagner:
> We could think about a `ParticleAdvectiveForcing` or also a `ParticleVelocityTendency` function that specifies the rhs of

What is the difference between this and what we are talking about?

glwagner:
> > We could think about a `ParticleAdvectiveForcing` or also a `ParticleVelocityTendency` function that specifies the rhs of dvdt=G
> 
> One example that this might be tricky for this approach is where the particles have a constant sinking velocity on top of the fluid velocity.

A constant sinking velocity is easy to implement using `dynamics`, since it's constant everywhere and doesn't depend on the position / fluid velocity experienced by the particle.

glwagner:
There used to be an example of using `dynamics` to implement particles that played "rock, paper, scissors"; ie near-neighbors were identified, and then made to interact in some way that changed particle properties.

There is scope for building out a user interface for more generic particle interactions of course. I'm not convinced that this is so important that we should spoon-feed users with an interface though. In principle, `dynamics` can do everything and advanced users can write their own code.

It depends on how important the uses get, and how clever we can be with an interface (so it is easy to maintain and doesn't screw things up for future development)

glwagner:
> However, I think that advection forcing should be run at the same time as advect_lagrangian_particles!. This would be important for cases where the additional advection is a function of its location for example.

Agreed

xkykai:
I added a `ParticleAdvectionForcing` that behaves similarly to `dynamics(particles, model, Δt)` which allows user to specify any velocities it wishes to add to the model by passing in `NonHydrostaticModel(..., advective_forcing = ParticleAdvectiveForcing(u=some_u, v=some_v, w=some_w))` . The arguments are functions that take `(particles, model, Δt)` as inputs. 

When I am trying to test my code I ran into a problem:
```
ERROR: UndefVarError: `flattened_node` not defined
```
@simone-silvestri I see that you added this and I couldn't find anywhere in the code where `flattened_node` is defined. What is it supposed to be?

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -183,7 +183,7 @@ function advect_lagrangian_particles!(particles, model, Δt)
     worksize = length(particles)
 
     advect_particles_kernel! = _advect_particles!(device(arch), workgroup, worksize)
-    advect_particles_kernel!(particles.properties, particles.restitution, model.grid, Δt, total_velocities(model))
+    advect_particles_kernel!(particles.properties, model, particles.restitution, model.grid, Δt, total_velocities(model))
```
simone-silvestri, line 186: we cannot pass the `model` to a GPU kernel because we don't have an `adapt` function for the model. 

You should extract the fields that you need and pass those to the kernel individually

simone-silvestri:
> I added a `ParticleAdvectionForcing` that behaves similarly to `dynamics(particles, model, Δt)` which allows user to specify any velocities it wishes to add to the model by passing in `NonHydrostaticModel(..., advective_forcing = ParticleAdvectiveForcing(u=some_u, v=some_v, w=some_w))` . The arguments are functions that take `(particles, model, Δt)` as inputs.
> 
> When I am trying to test my code I ran into a problem:
> 
> ```
> ERROR: UndefVarError: `flattened_node` not defined
> ```
> 
> https://github.com/CliMA/Oceananigans.jl/blob/a73e845a7a7bce8e22e9453670c1c20b67dbfc3a/src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl#L92
> 
> @simone-silvestri I see that you added this and I couldn't find anywhere in the code where `flattened_node` is defined. What is it supposed to be?

This is from #3395. We should wait for that PR to be merged before merging this one

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
xkykai: I see. Is `particles` fine? What do you think the `ParticleAdvectiveForcing` arguments should contain? I'm imagining `(particles, fluid_velocities, background_velocities, tracers, Δt)` perhaps.

simone-silvestri: sounds good! maybe also auxiliary fields to allow customization?

simone-silvestri:
to test the code you can merge #3395 in this PR

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
glwagner: Put the same stuff in it that we put into the tendency kernel forcing functions.

glwagner: > and the buoyancy model to allow for buoyant particles

That's an interesting idea. Since no buoyancy model is grid-dependent, usually a user can include things like the equation of state as a parameter to a forcing function.

I think you should design this feature to be identical to forcing for tendencies. Literally, identical. that will make it easy to understand.

xkykai: Sorry I am slightly confused: what do you mean by buoyancy model as an argument? The way I am thinking about it whatever buoyancy model one wish to use for the lagrangian particles can be defined in the function to be passed into `ParticleAdvectionForcing`

xkykai: In line with the forcing for tendencies, which I am assuming is this:
https://github.com/CliMA/Oceananigans.jl/blob/76c32f65bbf1035686515fcc039a45cafcd8c3ff/src/Forcings/discrete_forcing.jl#L46C1-L46C1
The arguments for the advective forcing would be as such
```
u_advective_forcing(particles, p, grid, clock, Δt, model_fields)
```
Is that correct?

simone-silvestri: we probably also need `i, j, k` as arguments

glwagner: There is no `i, j, k` though because this is particles right? The equivalent is `p`, the particle index. So the discrete form arguments would be `p, grid, clock, model_fields`.

Shouldn't you multiply this by `Δt` (because its forward Euler forcing)? I don't think you need to pass in `Δt`.

glwagner: The forcings are called here:

https://github.com/CliMA/Oceananigans.jl/blob/76c32f65bbf1035686515fcc039a45cafcd8c3ff/src/Models/NonhydrostaticModels/nonhydrostatic_tendency_kernel_functions.jl#L80

for example.

`DiscreteForcing` is a simple wrapper function that allows users to also store "parameters" in the forcing object. 

Eventually, it makes sense to use the same here, and have "discreet form" and "continuous form" for `ParticleForcing`. It's up to you whether you want to go all the way. Those wrappers are pure convenience and do not add functionality.

glwagner: > to be passed into ParticleAdvectionForcing

Should we call it simply `ParticleForcing`? It's true that this is a forcing on particle _position_ and is therefore, in some sense, "advective". But I'm not sure it actually helps clarify how the forcing is used.

simone-silvestri: We would need `i, j, k` for `model_fields` unless we want to interpolate them and pass them

simone-silvestri:
I am still not super convinced about the `ParticleForcing` implementation. 
This would not work well in case we want to solve the tendency for the particle-velocity evolution equation.

For example, there is no non-convoluted way to implement a particle with drag where $\frac{d \boldsymbol{v}}{dt} = \frac{C_d}{\tau}(\boldsymbol{u} - \boldsymbol{v})$ with an "advecting forcing" (we would have to remove the fluid velocity to not count it twice).  

In my opinion, I would prefer to call it `ParticleVelocities` with fields `u_velocity`, `v_velocity`, and `w_velocity` and `parameters` which are functions that default to the fluid velocity if not specified (where fluid velocity is passed to the function) with more complicated implementations possible such as in the above case with drag

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
glwagner: Hmm, but what would `i, j, k` be? I guess to use `model_fields`, the user here would have to call `interpolate` manually, and they would also need to know the staggered location...

I do like the idea of interpolating the model fields, but to put that in we probably want to design an API where the interpolation only happens if the users want to use those fields, eg similar to how we have `field_dependencies` in `ContinuousForcing`. But as a first pass, we can simply avoid interpolating if `isnothing(particle_forcing)`.

glwagner:
What does i, j, k refer to?

simone-silvestri:
in this case they would be the signature of the function as in the particle_advecting_forcing implementation

simone-silvestri:
these are if we want to make the particle velocity depend on other _eulerian_ fields (like forcing for our prognostic variables). An example can be the buoyancy field for buoyant particles.  If we decide that that role can be undertaken by the `dynamics` kernel, we can remove the `i, j, k` from the signature.

We should probably come up with a couple of examples to include in this PR that show the implementation of different particle dynamics. 
I would include:
- buoyant particles (sinking or rising with a density-dependent vertical velocity)
- particles with drag

glwagner:
> these are if we want to make the particle velocity depend on other _eulerian_ fields (like forcing for our prognostic variables). An example can be the buoyancy field for buoyant particles. If we decide that that role can be undertaken by the `dynamics` kernel, we can remove the `i, j, k` from the signature.
> 
> We should probably come up with a couple of examples to include in this PR that show the implementation of different particle dynamics. I would include:
> 
> * buoyant particles (sinking or rising with a density-dependent vertical velocity)
> * particles with drag

I understand the idea. But to use Eulerian fields we have to interpolate them to the particle location. The particles do not have indices `i, j, k`. Read what I said carefully.

glwagner (commented):
Why are there 25 files changed? The changes don't seem related to this PR.

simone-silvestri:
If we want to mimic the forcing implementation and pass the model fields to particle velocity we also necessarily need to pass `i, j, k`: the particle index is not enough to interpolate Eulerian fields to the particle location.

The second option is to forget passing Eulerian fields and restrict flexibility.

simone-silvestri:
right, sorry I confused `i, j, k` with particle position, I meant we have to pass the particle position `x, y, z`

glwagner:
> right, sorry I confused `i, j, k` with particle position, I meant we have to pass the particle position `x, y, z`

Ah ok, that makes more sense!

The particle position is available via

```julia
x = particles[p].x
```

etc right?

We should implement a helper function,

```julia
position(p) = (p.x, p.y, p.z)
```

Note that particles are also annoying on `Flat` grids. One needs to beware the particle position in the `Flat` direction, but they shouldn't have to. The same issue will affect `interpolate` once we merge #3395

xkykai:
I've updated the code to `ParticleVelocities` and changed the function signature to 

```julia
@inline particle_u_velocity(x, y, z, u_fluid, particles, p, advective_velocity::ParticleVelocities, grid, clock, Δt, model_fields) = advective_velocity.u(x, y, z, u_fluid, particles, p, grid, clock, Δt, model_fields)
```

 I also swapped around `advect_lagrangian_particles!` and `dynamics`, so now the particles are advected before the `dynamics` function is applied. I found this to be easier when implementing an example with the drag.

I've implemented a simple example with a drag in the form of $\frac{d \boldsymbol{v}}{dt} = \frac{C_d}{\tau}(\boldsymbol{u} - \boldsymbol{v})$.

xkykai:
Upon thinking about it I think perhaps it makes more sense to run `dynamics` first before `advect_lagrangian_particles!`, one example of which might be that particles sink depending on the radius of the particle, which changes with time. In such a case it is perhaps better to evolve the particle radius, then compute the sinking velocity given the new radius before advective it.

Also provided a draft example of how one could set up a problem where the particle sinks with a drag in the form of $\frac{d \boldsymbol{v}}{dt} = \frac{C_d}{\tau}(\boldsymbol{u} - \boldsymbol{v})$. 
Note: in the calculation the velocity of the particle itself needs to be tracked. This is done in `u_particle`, `v_particle`, and `w_particle` in `particles.properties` where `particles::LagrangianParticles`. The particle velocities are computed and updated in the  `dynamics` step, then `ParticleVelocities` only has functions that access the particle properties to grab the particle velocity. It is slightly clunky but unless we keep track of the particle velocities right out of the box and update them during the `advect_lagrangian_particles!` step, this is the way I could think of.

Since particle velocities are not updated when the particle is bounced, it will not work if the particles bounce from the boundaries back into the interior during the advection step, but for doubly-periodic domian and sinking particles it might not be very important.

glwagner:
> > > In such a case it is perhaps better to evolve the particle radius, then compute the sinking velocity given the new radius before advective it.
> > 
> > 
> > Are you sure? This could mean that the advecting velocity is extracted at time-step `n` but we are using a radius from `n+1`.
> 
> I think that depends on how one defines the advective velocity eg. if it is defined as an explicit function of time then it is better that dynamics act first to compute the radius at time t before it is being advected to time t+Δt.

But in that case, when a time-step is complete, the radius has been computed at time `t` but the particle position is updated to time `t + Δt`. This means that if the user asks for output, the two are inconsistent with one another.

glwagner:
Actually, I think we need to change the time-stepping loop:

https://github.com/CliMA/Oceananigans.jl/blob/3e2650373e9c73231681535aadb5b720a830dc97/src/TimeSteppers/quasi_adams_bashforth_2.jl#L93-L100

We should step particles forward before stepping the velocity field?

It also seems that we need to separate particle advection from "dynamics", and compute the dynamics once the time-stepping is complete (probably it makes sense to compute dynamics within `update_state!`?)

xkykai:
> > > > In such a case it is perhaps better to evolve the particle radius, then compute the sinking velocity given the new radius before advective it.
> > > 
> > > 
> > > Are you sure? This could mean that the advecting velocity is extracted at time-step `n` but we are using a radius from `n+1`.
> > 
> > 
> > I think that depends on how one defines the advective velocity eg. if it is defined as an explicit function of time then it is better that dynamics act first to compute the radius at time t before it is being advected to time t+Δt.
> 
> But in that case, when a time-step is complete, the radius has been computed at time `t` but the particle position is updated to time `t + Δt`. This means that if the user asks for output, the two are inconsistent with one another.

Makes sense. `advect_lagrangian_particles!` first before `dynamics` it is!

xkykai:
> We should step particles forward before stepping the velocity field?

> It also seems that we need to separate particle advection from "dynamics", and compute the dynamics once the time-stepping is complete (probably it makes sense to compute dynamics within `update_state!`?)

Agreed. @simone-silvestri what do you think?

glwagner:
This PR is very stale, so I'm closing it.

PR #3401, opened 2023-12-06
Added adapt method for scalar and scalar biharmonic diffusvity + test

jagoosw:
Scalar diffusivity fails on GPU if you use fields for the viscosity/diffusivity since there is no adapt method. 

I've added them in this PR along with a test, feel free to remove the test if its too much.


glwagner (approved):
Seems legit

glwagner:
This PR was merged even though tests never ran.

PR #3402, opened 2023-12-13
Fix CUDA compat entry

jagoosw:
There was a bug in the earlier v4 releases so Oceananigans won't work with them (this has come up as CUDAv4.0 is the final version compatible with CUDA 10.2).

```julia
julia> using Oceananigans
┌ Warning: The NVIDIA driver on this system only supports up to CUDA 10.2.0.
│ For performance reasons, it is recommended to upgrade to a driver that supports CUDA 11.2 or higher.
└ @ CUDA ~/.julia/packages/CUDA/ZdCxS/src/initialization.jl:71
julia> grid = RectilinearGrid(GPU(); size = (128, 128, 128), extent = (1, 1, 1))
mo128×128×128 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on GPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.0078125
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=0.0078125
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=0.0078125

julia> model = NonhydrostaticModel(; grid)
ERROR: UndefVarError: `CUDABackend` not defined
Stacktrace:
  [1] getproperty
    @ ./Base.jl:31 [inlined]
  [2] device(#unused#::GPU)
    @ Oceananigans.Architectures ~/.julia/packages/Oceananigans/f5Cpw/src/Architectures.jl:39
[...]

PR #3406, opened 2023-12-18
Testing this change from type-stable-with-tracers

jlk9:
Just seeing if the change to top_boundary_conditions breaks anything.


wsmoses:
@glwagner can you review this? The failures seem unrelated, e.g. `Reason: unsupported dynamic function invocation (call to Base.Fix1{typeof(Adapt.adapt), Float64})` and I'm not sure what already is expected to fail

glwagner:
> The failures seem unrelated, e.g. `Reason: unsupported dynamic function invocation (call to Base.Fix1{typeof(Adapt.adapt), Float64})`

wut

> I'm not sure what already is expected to fail

Nothing is expected to fail

Where does `Fix1` come from?

glwagner:
`Fix1` is a julia function:

```julia
help?> Base.Fix1
  Fix1(f, x)

  A type representing a partially-applied version of the two-argument function
  f, with the first argument fixed to the value "x". In other words, Fix1(f,
  x) behaves similarly to y->f(x, y).

  See also Fix2.
```

I don't see where this gets created.

glwagner:
Something fishy is happening. The error is

```julia
Reason: unsupported dynamic function invocation (call to Base.Fix1{typeof(Adapt.adapt), Float64})
```

This means we created a partially-applied function via `Fix1(adapt, x)`. 

This looks like a bug that has to do with some package update, maybe to `Adapt`?

glwagner:
This issue crept in on https://github.com/CliMA/Oceananigans.jl/pull/3402. That PR should not have been merged. So, we have to rewind it I guess.

glwagner (approved):

Issue #1679, opened 2021-05-20
Re-implementing the HydrostaticFreeSurfaceModel with Generalized Vertical Coordinates (GVC)


glwagner:
The equations of motion for most or all major ocean modeling software are implemented using a "Generalized Vertical Coordinate" (GVC). Generalized vertical coordinates contain a "fixed z" coordinate as a limiting case, but generalize to vertical coordinates that

*  Are fixed in time but conform to topographic variations (`σ` coordinates);
* Vary in time in a diagnostic manner (for example, _z-star_ coordinates, which vary with the sea surface displacement);
* Vary in time according to some prognostic equation (fully Lagrangian isopycnal coordinates, semi-Lagrangian hybrid coordinates / _z-tilde_ coordinates that vary only with high frequency motion, and adaptive methods that prescribe GVC dynamics to obtain favorable properties like concentration in regions of strong stratification).

A fully general GVC typically also requires a "Lagrangian remapping" step to avoid extreme grid distortions in regions of persistent vertical velocities.

The implementation of GVC is likely a major refactoring of `HydrostaticFreeSurfaceModel` because it will change the equations of motion and could even potentially change the nature of its prognostic variables. For time-dependent GVC, tracer equations must be reformulated in terms of the "thickness-weighted" tracer concentration, which in our case means the tracer concentration normalized by the local grid spacing. This could mean either using the thickness-weighted tracer conservation as a state variable, or it could mean rewriting the time-stepping algorithm so that unweighted tracer conservation can be updated according to the conservation of thickness-weighted tracer.

A preliminary roadmap towards GVC in `HydrostaticFreeSurfaceModel` is

1. Introduce `AbstractVerticalCoordinate` and refactor the `HydrostaticFreeSurfaceModel` to integrate thickness-weighted equations. When using `ZCoordinate`, the resulting model produces identical results to the current implementation of `HydrostaticFreeSurfaceModel`. I think we will also want a rectilinear grid to use for testing to avoid modifying `RegularRectilinearGrid`, such as a refactored or reimplemented three-dimensionally `StretchedRectilinearGrid` from #1532 which accepts the use of a generalized vertical coordinates.

[...]

PR #3411, opened 2023-12-21
Z-star vertical coordinate

simone-silvestri:
This PR introduces a ZStar, free surface following vertical coordinate. 

the vertical coordinate evolves in time and is implemented following the recipe in 
[Rescaled height coordinates for accurate representation of free-surface flows in ocean circulation models ](https://www.sciencedirect.com/science/article/abs/pii/S1463500303000544) that allows the implementation of a general vertical coordinate where the only specification needed is the vertical derivative of the vertical coordinate.

Closes https://github.com/CliMA/Oceananigans.jl/issues/1679


simone-silvestri:
global tracer conservation for the hydrostatic lock-release configuration in `validation/moving_coordinate/z_star_lock_release.jl`. Better than the z-coordinate but still not machine-precision conserving.

Screenshot 2023-12-21 at 11 00 02 AM

glwagner:
Please describe your design in words, the challenges that exist, and how you plan to meet those challenges.

glwagner:
There are two things at play here. First, there is a refactoring of the equation set to allow for a generalized vertical coordinate. This actually involves rewriting the equations. To merge that generalization, I think we would also like to see the new equations that are being solved written up in the docs. When the equations are generalized, this would also involve implementing a new abstraction for the vertical coordinate, along with a `vertical_coordinate = ZVerticalCoordinate()` default, which preserves existing behavior.

The second PR should then implement a `ZStarVerticalCoordinate()` (other better names are fine too).

So to me this feels like we need two PRs, one with docs update and API design, and then a second with z* implementation.

src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
```diff
@@ -33,7 +33,7 @@ function GeneralizedSpacingGrid(grid::AbstractUnderlyingGrid{FT, TX, TY, TZ}, ::
     
     # Initial "at-rest" conditions
     launch!(architecture(grid), grid, :xy, _update_zstar!, 
-            sⁿ, s⁻, ΔzF, ΔzC, ZeroField(grid), grid, Val(grid.Nz))
+            sⁿ, s⁻, ΔzF, ΔzC, ZeroField(), grid, Val(grid.Nz))
```
glwagner, line 36: Should we use Oceananigans notation for `Δz`?

src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
```diff
@@ -86,6 +86,10 @@ Return a flattened `NamedTuple` of the prognostic fields associated with `Hydros
 @inline prognostic_fields(model::HydrostaticFreeSurfaceModel) =
     hydrostatic_prognostic_fields(model.velocities, model.free_surface, model.tracers)
 
+@inline prognostic_fields(model::HydrostaticFreeSurfaceModel{<:Any, <:Any, <:Any, <:Any, ZStarSpacingGrid}) =
+    merge(hydrostatic_prognostic_fields(model.velocities, model.free_surface, model.tracers), 
+          (; ∂t_∂s = grid.Δzᵃᵃᶠ.∂t_∂s))
```
glwagner, line 91: What does `∂t_∂s` mean?

glwagner: You probably want to define an interface to `grid.Δzᵃᵃᶠ.∂t_∂s`. This is specific to the grids we have because it requires the property `Δzᵃᵃᶠ` to exist. We discourage writing assumptive code like this.

simone-silvestri: right, I will add that (it actually already exists, I just forgot to add it there)

src/ImmersedBoundaries/ImmersedBoundaries.jl
```diff
@@ -150,8 +150,11 @@ Adapt.adapt_structure(to, ibg::IBG{FT, TX, TY, TZ}) where {FT, TX, TY, TZ} =
                                      nothing, 
                                      nothing)
 
-with_halo(halo, ibg::ImmersedBoundaryGrid) =
-    ImmersedBoundaryGrid(with_halo(halo, ibg.underlying_grid), ibg.immersed_boundary)
+function with_halo(halo, ibg::ImmersedBoundaryGrid)
+    new_grid = with_halo(halo, ibg.underlying_grid)
+    active_cells_map = !isnothing(ibg.interior_active_cells)
+    return ImmersedBoundaryGrid(new_grid, ibg.immersed_boundary; active_cells_map)
+end
```
glwagner, line 157: does this have anything to do with the present PR?

src/ImmersedBoundaries/ImmersedBoundaries.jl
```diff
@@ -238,6 +241,10 @@ As well as
 @inline immersed_inactive_node(i, j, k, ibg::IBG, LX, LY, LZ) =  inactive_node(i, j, k, ibg, LX, LY, LZ) &
                                                                 !inactive_node(i, j, k, ibg.underlying_grid, LX, LY, LZ)
 
+# Underlying grids are never immersed!                                                                
+@inline immersed_peripheral_node(i, j, k, grid::AbstractUnderlyingGrid, LX, LY, LZ) = false
+@inline   immersed_inactive_node(i, j, k, grid::AbstractUnderlyingGrid, LX, LY, LZ) = false
```
glwagner, line 246: relation to present PR?

src/Grids/rectilinear_grid.jl
```diff
@@ -19,29 +19,31 @@ struct RectilinearGrid{FT, TX, TY, TZ, FX, FY, FZ, VX, VY, VZ, Arch} <: Abstract
...
+                                                                          TX, TY, TZ,
+                                                                          FX, VX, FY,
+                                                                          VY, FZF, FZC, VZ}
+                                                                                           
+        return new{FT, TX, TY, TZ, FX, FY, FZF, FZC, VX, VY, VZ, Arch}(arch, Nx, Ny, Nz,
+                                                                       Hx, Hy, Hz, Lx, Ly, Lz, 
+                                                                       Δxᶠᵃᵃ, Δxᶜᵃᵃ, xᶠᵃᵃ, xᶜᵃᵃ,
+                                                                       Δyᵃᶠᵃ, Δyᵃᶜᵃ, yᵃᶠᵃ, yᵃᶜᵃ,
+                                                                       Δzᵃᵃᶠ, Δzᵃᵃᶜ, zᵃᵃᶠ, zᵃᵃᶜ)
+    end
```
glwagner, line 46: This could be its own PR.

src/ImmersedBoundaries/grid_fitted_bottom.jl
```diff
@@ -8,6 +8,7 @@ using Oceananigans.BoundaryConditions: FBC
 using Printf
 
 import Oceananigans.TurbulenceClosures: z_bottom
+import Oceananigans.Architectures: on_architecture
```
glwagner, line 11: relation to present PR?

src/Models/HydrostaticFreeSurfaceModels/generalized_vertical_spacing.jl
```diff
@@ -0,0 +1,195 @@
...
+
+spacings for a generalized vertical coordinate system. The reference (non-moving) spacings are stored in `Δr`. 
+`Δ` contains the spacings associated with the moving coordinate system.
+`s⁻`, `sⁿ` and `∂t_∂s` fields are the vertical derivative of the vertical coordinate (∂Δ/∂Δr)
+at timestep `n-1` and `n` and it's time derivative.
+`denomination` contains the "type" of generalized vertical coordinate (the only one implemented is `ZStar`)
+"""
+abstract type AbstractVerticalSpacing{R} end
+
+const AbstractVerticalSpacingRG  = RectilinearGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:AbstractVerticalSpacing}
```
glwagner, line 27: why is the property all the way at the end of the list of properties of the grid? makes this harder to read

src/Models/HydrostaticFreeSurfaceModels/generalized_vertical_spacing.jl
```diff
@@ -0,0 +1,195 @@
...
+retrieve_static_grid(grid) = grid
+
+function retrieve_static_grid(grid::AbstractVerticalSpacingImmersedGrid) 
+    underlying_grid  = retrieve_static_grid(grid.underlying_grid)
+    active_cells_map = !isnothing(grid.interior_active_cells)
+    return ImmersedBoundaryGrid(underlying_grid, grid.immersed_boundary; active_cells_map)
+end
+
+reference_Δzᵃᵃᶠ(grid) = grid.Δzᵃᵃᶠ
+reference_Δzᵃᵃᶜ(grid) = grid.Δzᵃᵃᶜ
```
glwagner, line 48: does zstar not work for a deep sphere case?

src/Models/HydrostaticFreeSurfaceModels/generalized_vertical_spacing.jl
```diff
@@ -0,0 +1,195 @@
...
+
+generalized_spacing_grid(grid, coord) = grid
+update_vertical_spacing!(model, grid; kwargs...) = nothing
+
+##### 
+##### Vertical spacings for a generalized vertical coordinate system
+#####
+
+# Very bad for GPU performance!!! (z-values are not coalesced in memory for z-derivatives anymore)
+# TODO: make z-direction local in memory by not using Fields
```
glwagner, line 107: This comment makes me worried. There are better ways to make the z-direction local; the `Field` abstraction has nothing to do with it. Remember that `getindex` is divorced from the memory layout. The 3rd index can be local. The thing that has to change is simply how the arrays are constructed via `newdata` in the grids module.

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -100,30 +101,43 @@ Keyword arguments
...
+                                   tracers = (:T, :S),
+                       forcing::NamedTuple = NamedTuple(),
+                                   closure = nothing,
+           boundary_conditions::NamedTuple = NamedTuple(),
+             particles::ParticlesOrNothing = nothing,
+     biogeochemistry::AbstractBGCOrNothing = nothing,
+                                velocities = nothing,
+                                  pressure = nothing,
+                        diffusivity_fields = nothing,
+           generalized_vertical_coordinate = nothing,
```
glwagner, line 123: `generalized` is implied fine here, so I would go with the name `vertical_coordinate`.

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl
```diff
@@ -82,10 +83,11 @@ implicitly during time-stepping.
     
     model_fields = merge(hydrostatic_fields(velocities, free_surface, tracers), auxiliary_fields)
 
-    return ( - U_dot_∇v(i, j, k, grid, advection, velocities)
+    return ( - U_dot_∇v(i, j, k, grid, advection, velocities) 
              - explicit_barotropic_pressure_y_gradient(i, j, k, grid, free_surface)
              - y_f_cross_U(i, j, k, grid, coriolis, velocities)
              - ∂yᶜᶠᶜ(i, j, k, grid, hydrostatic_pressure_anomaly)
+             - grid_slope_contribution_y(i, j, k, grid, free_surface, buoyancy, model_fields)
```
glwagner, line 90: is this a pressure contribution? I can't quite tell

src/Models/HydrostaticFreeSurfaceModels/generalized_vertical_spacing.jl
glwagner: it's unclear to me whether this is a "reference" or "resting"

src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
```diff
@@ -0,0 +1,213 @@
...
+    sᶜᶜⁿ :: SCC
+    sᶠᶜⁿ :: SFC
+    sᶜᶠⁿ :: SCF
+    sᶠᶠⁿ :: SFF
+    sᶜᶜ⁻ :: SCC
+ ∂t_∂s   :: SCC
+end
+
+Adapt.adapt_structure(to, coord::ZStarSpacing) = 
+            ZStarSpacing(Adapt.adapt(to, coord.Δr),
```
glwagner, line 36: is that an 8 character indent

src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
```diff
@@ -0,0 +1,213 @@
...
+The vertical spacing is defined by a reference spacing `Δr` and a scaling `s` that obeys
+```math
+s = (η + H) / H
+```
+where ``η`` is the free surface height and ``H`` the vertical depth of the water column
+
+# Fields
+- `Δr`: reference vertical spacing with `η = 0`
+- `sⁿ`: scaling of the vertical coordinate at time step `n`
+- `s⁻`:scaling of the vertical coordinate at time step `n - 1`
```
glwagner, line 22:
```suggestion
- `s⁻`: scaling of the vertical coordinate at time step `n - 1`
```

src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
```diff
@@ -0,0 +1,213 @@
...
+A vertical spacing for the hydrostatic free surface model that follows the free surface.
+The vertical spacing is defined by a reference spacing `Δr` and a scaling `s` that obeys
+```math
+s = (η + H) / H
+```
+where ``η`` is the free surface height and ``H`` the vertical depth of the water column
+
+# Fields
+- `Δr`: reference vertical spacing with `η = 0`
+- `sⁿ`: scaling of the vertical coordinate at time step `n`
```
glwagner, line 21: Are there any options other than `s`? It doesn't produce the most obvious expressions and also `s` is used for salinity.

src/Models/HydrostaticFreeSurfaceModels/generalized_vertical_spacing.jl
```diff
@@ -0,0 +1,195 @@
...
+@inline Δzᶜᶠᶠ(i, j, k, grid::AbstractVerticalSpacingGrid) = ℑyᵃᶠᵃ(i, j, k, grid, Δzᶜᶜᶠ)
+@inline Δzᶜᶠᶜ(i, j, k, grid::AbstractVerticalSpacingGrid) = ℑyᵃᶠᵃ(i, j, k, grid, Δzᶜᶜᶜ)
+
+@inline Δzᶠᶜᶠ(i, j, k, grid::AbstractVerticalSpacingGrid) = ℑxᶠᵃᵃ(i, j, k, grid, Δzᶜᶜᶠ)
+@inline Δzᶠᶜᶜ(i, j, k, grid::AbstractVerticalSpacingGrid) = ℑxᶠᵃᵃ(i, j, k, grid, Δzᶜᶜᶜ)
+
+@inline Δzᶠᶠᶠ(i, j, k, grid::AbstractVerticalSpacingGrid) = ℑxyᶠᶠᵃ(i, j, k, grid, Δzᶜᶜᶠ)
+@inline Δzᶠᶠᶜ(i, j, k, grid::AbstractVerticalSpacingGrid) = ℑxyᶠᶠᵃ(i, j, k, grid, Δzᶜᶜᶜ)
+
+@inline Δz_reference(i, j, k, Δz::Number) = Δz
```
glwagner, line 119: I'm not a huge fan of this notation because it mixes symbols (`Δz`) with language (`reference`)

src/Grids/rectilinear_grid.jl
```diff
@@ -1,4 +1,4 @@
-struct RectilinearGrid{FT, TX, TY, TZ, FX, FY, FZ, VX, VY, VZ, Arch} <: AbstractUnderlyingGrid{FT, TX, TY, TZ, Arch}
+struct RectilinearGrid{FT, TX, TY, TZ, FX, FY, FZF, FZC, VX, VY, VZ, Arch} <: AbstractUnderlyingGrid{FT, TX, TY, TZ, Arch}
```
glwagner, line 1: Why do FX, FY take priority over FZF and FZC (not to mention Arch) here?

src/Advection/vector_invariant_cross_upwinding.jl
```diff
@@ -21,16 +21,19 @@
     @inbounds û = u[i, j, k]
     δ_stencil = scheme.upwinding.divergence_stencil
 
-    δᴿ = _biased_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, scheme.divergence_scheme, bias(û), flux_div_xyᶜᶜᶜ, δ_stencil, u, v) 
+    δᴿ  =    _biased_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, scheme.divergence_scheme, bias(û), flux_div_xyᶜᶜᶜ, δ_stencil, u, v) 
+    ∂ts = _symmetric_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, cross_scheme, V_times_∂t_∂s_grid)
```
glwagner, line 25: Is `∂t_∂s` right? Or is this just `∂t_s`? Note, this should be consistent everywhere. (I see a `∂ts` below.) We use the form `∂t_s` as a shorthand for `∂s/∂t` (the time-derivative of `s` with respect to `t`), elsewhere we have written `∂z_u` for example.

src/Models/HydrostaticFreeSurfaceModels/generalized_vertical_spacing.jl
simone-silvestri: this is actually superseded by the fact that we do not allocate anymore a 3d field but compute \Delta z locally by doing \Delta r * s where s is a two-dimensional field, so memory access should be coalesced now

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl
simone-silvestri: yep, it's the hydrostatic pressure contribution coming from the fact that the grid is sloped. What about `grid_slope_hydrostatic_pressure_y`?

src/Models/HydrostaticFreeSurfaceModels/generalized_vertical_spacing.jl
```diff
@@ -116,23 +116,27 @@ update_vertical_spacing!(model, grid; kwargs...) = nothing
...
 
-@inline Δz_reference(i, j, k, Δz::AbstractVerticalSpacing{<:Number}) = Δz.Δr
-@inline Δz_reference(i, j, k, Δz::AbstractVerticalSpacing) = @inbounds Δz.Δr[k]
+@inline Δr(i, j, k, Δz::AbstractVerticalSpacing{<:Number}) = Δz.Δr
+@inline Δr(i, j, k, Δz::AbstractVerticalSpacing) = @inbounds Δz.Δr[k]
 
-@inline Δzᶜᶜᶠ_reference(i, j, k, grid) = Δz_reference(i, j, k, grid.Δzᵃᵃᶠ)
-@inline Δzᶜᶜᶜ_reference(i, j, k, grid) = Δz_reference(i, j, k, grid.Δzᵃᵃᶜ)
+@inline Δrᶜᶜᶠ(i, j, k, grid) = Δr(i, j, k, grid.Δzᵃᵃᶠ)
+@inline Δrᶜᶜᶜ(i, j, k, grid) = Δr(i, j, k, grid.Δzᵃᵃᶜ)
```
glwagner, line 130: this is better!

I would like to know, does this mean we use the opposite notation as MOM6? I believe in MOM6 `r` is the generalized vertical coordinate and `z` is the height coordinate. It would seem with these definitions that `z` is the name of the "generalized" coordinate and `r` is the reference (which at least for z* is the height coordinate, though for other coordinates I suppose it might be something else).

simone-silvestri: I was thinking the reference `r` is the `z*` (always fixed between 0 and -H), while the z is the moving grid defined between -H and eta. We probably need to also add an `rnode` for this grid.

glwagner: Yes for sure. It's a notation battle. `rnode` implies that "r" means the "third coordinate" (not specifically the reference coordinate).

It'd require more herculean effort for us to be able to denote "z" as "height", rather than the generalized vertical coordinate, though.

Still something to ponder. It's not impossible. I don't think we need it here.

"r" for reference makes a lot of sense. Perhaps, we can choose a different name for `rnode`. Like `ζnode` or something.

src/Advection/vector_invariant_cross_upwinding.jl
simone-silvestri: ok make sense, I will switch everything to `∂t_s`. Then we can think about a better name for `s`. In MITgcm they use `hfac`. In NEMO they use `e3`.

src/Grids/rectilinear_grid.jl
simone-silvestri: I don't think we ever dispatch on Arch. FZF and FZC can definitely take priority. I ll open a new PR to do that.

glwagner:
@simone-silvestri discussed a bit the user interface for building models with a z* vertical coordinate. Here's a prototype design we came up with:

```julia
reference_levels = exponential_z_faces(Nz=40, depth=6000)

Nx = 1440
Ny = 600
Nz = length(reference_levels) - 1

grid = LatitudeLongitudeGrid(arch;
                             size = (Nx, Ny, Nz),
                             halo = (7, 7, 7),
                             z = ZStarVerticalCoordinate(reference_levels),
                             latitude  = (-75, 75),
                             longitude = (0, 360))
```

We like this because with z*, the znodes, and zspacings are always changing. We think this makes that semantically clear.

Note there are some challenges for output that we have to solve, because the grid has to be updated to do computations. But I think we can solve these with `FieldTimeSeries`.

glwagner:
We can break some of this up into at least 3 PRs:

1. Compute "numerical bottom height" when building immersed boundary grid
2. Implement a nonlinear free surface
3. Do z*

simone-silvestri:
superseded by #3956

PR #3413, opened 2023-12-21
(0.90.3) Fix bug in adapting `ScalarDiffusivity` and `ScalarBiharmonicDiffusivity`

glwagner:
This PR fixes a bug that (apparently) crept in on #3401 --- I think. It's hard to know for sure because GPU tests don't pass right now. There may be more to fix.

@jagoosw I believe that `adapt` is missing its first argument. Note the PRs were merged without tests passing... that's why we have this issue.

@wsmoses @jlk9 we may need this to pass for your test PRs to be useful.

@navidcy

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.90.2"
+version = "0.90.3"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/TurbulenceClosures/discrete_diffusion_function.jl
+++ b/src/TurbulenceClosures/discrete_diffusion_function.jl
@@ -81,13 +81,13 @@ const UnlocalizedDDF                 = DiscreteDiffusionFunction{<:Nothing, <:No
 const UnlocalizedUnparametrizedDDF   = DiscreteDiffusionFunction{<:Nothing, <:Nothing, <:Nothing, <:Nothing}
 
 @inline function getdiffusivity(dd::DiscreteDiffusionFunction{LX, LY, LZ},
-                              i, j, k, grid, location, clock, fields) where {LX, LY, LZ} 
+                                i, j, k, grid, location, clock, fields) where {LX, LY, LZ} 
     from = (LX(), LY(), LZ())
     return ℑxyz(i, j, k, grid, from, location, dd.func, clock, fields, dd.parameters)
 end
 
 @inline function getdiffusivity(dd::UnparameterizedDDF{LX, LY, LZ},
-                              i, j, k, grid, location, clock, fields) where {LX, LY, LZ}
+                                i, j, k, grid, location, clock, fields) where {LX, LY, LZ}
     from = (LX(), LY(), LZ())
     return ℑxyz(i, j, k, grid, from, location, dd.func, clock, fields)
 end
@@ -99,5 +99,5 @@ end
         dd.func(i, j, k, grid, location..., clock, fields)
 
 Adapt.adapt_structure(to, dd::DiscreteDiffusionFunction{LX, LY, LZ}) where {LX, LY, LZ} =
-     DiscreteBoundaryFunction{LX, LY, LZ}(Adapt.adapt(to, dd.func),
+    DiscreteDiffusionFunction{LX, LY, LZ}(Adapt.adapt(to, dd.func),
                                           Adapt.adapt(to, dd.parameters))
--- a/src/TurbulenceClosures/turbulence_closure_implementations/scalar_biharmonic_diffusivity.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/scalar_biharmonic_diffusivity.jl
@@ -103,6 +103,9 @@ end
 
 Base.show(io::IO, closure::ScalarBiharmonicDiffusivity) = print(io, summary(closure))
 
-Adapt.adapt_structure(to, closure::ScalarBiharmonicDiffusivity) =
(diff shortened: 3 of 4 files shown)


glwagner:
Possibly, we should start using a merge queue that would disallow PRs from being merged unless tests pass.

glwagner:
PS from the docstring for `adapt`:

```julia
help?> Adapt.adapt
  adapt(to, x)

  Adapt a value x according to to. If no specific adaptions have been registered for to, this call will be a no-op.

  To alter the behavior, implement methods of adapt_structure and adapt_storage to respectively define how to adapt structures, and the leaves of those structures.

  For example, defining an adaptor for an environment where we can't have integers, and adding a method to adapt_storage to appropriately convert those to floating-point numbers:

  julia> struct IntegerLessAdaptor end

  julia> Adapt.adapt_storage(::IntegerLessAdaptor, x::Int64) = Float64(x)

  julia> adapt(IntegerLessAdaptor(), 42)
  42.0

  This will automatically work on known types too:

  julia> adapt(IntegerLessAdaptor(), tuple(1,2,3))
  (1.0, 2.0, 3.0)

  If we want this to work with custom structures, we need to extend adapt_structure:

  julia> struct MyStructure
           x
         end

  julia> Adapt.adapt_structure(to, obj::MyStructure) = MyStructure(adapt(to, obj.x))

  julia> adapt(IntegerLessAdaptor(), MyStructure(42))
  MyStructure(42.0)

  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  adapt(to)

  Create a function that adapts its argument according to to. If no specific adaptions have been registered for to, the returned function will be equivalent to identity.
```

The "1-argument" version of `adapt` actually returns a `Fix1`, I think. That's why we are getting errors about calling `Fix1`.

glwagner:
```julia
julia> f = Adapt.adapt(0)
(::Base.Fix1{typeof(adapt), Int64}) (generic function with 1 method)

julia> typeof(f)
Base.Fix1{typeof(adapt), Int64}
```

navidcy:
Adapt got a breaking change recently from v3 to v4. I'm not sure what the breaking change was tbh.

We are using v3 up to now. @glwagner, is the help from Adapt you point out from v3 or v4?

glwagner:
> Adapt got a breaking change recently from v3 to v4. I'm not sure what the breaking change was tbh.
> 
> We are using v3 up to now. @glwagner, is the help from Adapt you point out from v3 or v4?

That help was from Adapt 3.6.2, which is in the Oceananigans Manifest. I'll check v4 now. But I don't think it matters. The first argument `to` is necessary --- that indicates the device you are adapting to. There are two pieces of information required to adapt an object: the device we're adapting to, and the object being adapted.

glwagner:
> Adapt got a breaking change recently from v3 to v4. I'm not sure what the breaking change was tbh.
> 
> We are using v3 up to now. @glwagner, is the help from Adapt you point out from v3 or v4?

The syntax for `adapt` on v4.0.0 is the same:

```julia
help?> adapt
search: adapt Adapt InvalidStateException

  adapt(to, x)

  Adapt a value x according to to. If no specific adaptions have been registered for to, this call will be a no-op.

  To alter the behavior, implement methods of adapt_structure and adapt_storage to respectively define how to adapt structures, and the leaves of those structures.

  For example, defining an adaptor for an environment where we can't have integers, and adding a method to adapt_storage to appropriately convert those to floating-point numbers:

  julia> struct IntegerLessAdaptor end

  julia> Adapt.adapt_storage(::IntegerLessAdaptor, x::Int64) = Float64(x)

  julia> adapt(IntegerLessAdaptor(), 42)
  42.0

  This will automatically work on known types too:

  julia> adapt(IntegerLessAdaptor(), tuple(1,2,3))
  (1.0, 2.0, 3.0)

  If we want this to work with custom structures, we need to extend adapt_structure:

  julia> struct MyStructure
           x
         end

  julia> Adapt.adapt_structure(to, obj::MyStructure) = MyStructure(adapt(to, obj.x))

  julia> adapt(IntegerLessAdaptor(), MyStructure(42))
  MyStructure(42.0)

  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  adapt(to)

  Create a function that adapts its argument according to to. If no specific adaptions have been registered for to, the returned function will be equivalent to identity.
```

glwagner:
Why are the tests being run twice? It's confusing to interpret the test results.

navidcy:
> Why are the tests being run twice? It's confusing to interpret the test results.

I know! I’m confused. Perhaps yet some other buildkite setting? I’ll try to figure it out…..

glwagner:
These tests still fail: https://buildkite.com/clima/oceananigans/builds/13791#018c8e8a-a41b-43de-a387-49157d3ccc32/6-46

navidcy:
Is 

https://github.com/CliMA/Oceananigans.jl/blob/c5b030908793d9b49faba6648980abb1932638fc/src/TurbulenceClosures/discrete_diffusion_function.jl#L102

correct?

If it should be `DiscreteBoundaryFunction` then we need to import it. Or should this be `DiscreteDiffusionFunction`? The latter makes the test pass.

navidcy:
(I changed it but revert back if this wasn't supposed to be like that!)

navidcy:
Bump patch release?

navidcy (approved):

navidcy:
> Possibly, we should start using a merge queue that would disallow PRs from being merged unless tests pass.

Yeap. I added these in https://github.com/CliMA/Oceananigans.jl/settings/branches

Screenshot 2023-12-23 at 1 10 26 pm

How does this sound?

jagoosw:
> @jagoosw I believe that `adapt` is missing its first argument. Note the PRs were merged without tests passing... that's why we have this issue.

Sorry! I hadn't realised they could be merged without the tests passing so when I saw the PR approved just hit merge. Sorry for all the problems!

Issue #3412, opened 2023-12-21
Bugs in the Tilted Bottom Boundary Layer Example Code


loganpknudsen:
When running the Tilted Bottom Boundary Layer example code on the latest version of Oceananigans and Julia 1.9.2, I found the following error popped up:
```MethodError: no method matching constant_stratification(::Float64, ::Float64, ::Float64, ::Float64, ::NamedTuple{(:ĝ, :N²), Tuple{Vector{Float64}, Float64}})

Closest candidates are:
  constant_stratification(::Any, ::Any, ::Any, ::Any)
   @ Main In[7]:1


Stacktrace:
  [1] call_func
    @ ~/.julia/packages/Oceananigans/0y15B/src/Fields/function_field.jl:54 [inlined]
  [2] getindex
    @ ~/.julia/packages/Oceananigans/0y15B/src/Fields/function_field.jl:63 [inlined]
  [3] inner_left_biased_interpolate_xᶠᵃᵃ
    @ ~/.julia/packages/Oceananigans/0y15B/src/Advection/upwind_biased_reconstruction.jl:114 [inlined]
  [4] left_biased_interpolate_xᶠᵃᵃ
    @ ~/.julia/packages/Oceananigans/0y15B/src/Advection/reconstruction_coefficients.jl:11 [inlined]
  [5] _left_biased_interpolate_xᶠᵃᵃ
    @ ~/.julia/packages/Oceananigans/0y15B/src/Advection/topologically_conditional_interpolation.jl:56 [inlined]
  [6] advective_tracer_flux_x
    @ ~/.julia/packages/Oceananigans/0y15B/src/Advection/upwind_biased_advective_fluxes.jl:106 [inlined]
  [7] _advective_tracer_flux_x
    @ ~/.julia/packages/Oceananigans/0y15B/src/Advection/tracer_advection_operators.jl:4 [inlined]
  [8] δxᶜᵃᵃ
    @ ~/.julia/packages/Oceananigans/0y15B/src/Operators/difference_operators.jl:20 [inlined]
  [9] div_Uc
    @ ~/.julia/packages/Oceananigans/0y15B/src/Advection/tracer_advection_operators.jl:31 [inlined]
 [10] tracer_tendency
    @ ~/.julia/packages/Oceananigans/0y15B/src/Models/NonhydrostaticModels/nonhydrostatic_tendency_kernel_functions.jl:258 [inlined]
 [11] cpu_compute_Gc!
[...]

PR #3414, opened 2023-12-29
Few fixes in few examples

navidcy:
This PR fixes a few latex rendering related issues in the tilted BBL and HC example.

Let's make sure we resolve any issues related with #3412 here before merging.


navidcy:
https://github.com/CliMA/Oceananigans.jl/blob/6a78d74065bd74d80f6d98cb022e5db97d3a2e0b/src/BuoyancyModels/buoyancy.jl#L47-L53

So, I'm a bit confused. Where is ĝ supposed to point, up or down for non-tilted? @tomchor?

navidcy:
The tilted BBL example seems to built just fine so I don't understand what's the issue that https://github.com/CliMA/Oceananigans.jl/issues/3412 points to. I believe this PR is ready to be reviewed and merged.

PR #3404, opened 2023-12-14
Scaling improvement for the distributed `HydrostaticFreeSurfaceModel`

simone-silvestri:
This PR contains some improvements that allow better scaling on multiple devices. 

In particular:
- Splitting the interior active cell map into `west` `east` `north` `south` and `interior` to allow for overlapping communication and computations with _active_ cells.
- Advancements for the `SplitExplicitFreeSurface` to reduce kernel launch latency

PR #3434, opened 2024-01-19
MPData advection scheme

simone-silvestri:
This PR implements a three-dimensional positivity preserving advection scheme called multidimensional positive definite advection transport algorithm (MPDATA), which follows this [paper](https://www.researchgate.net/publication/241972841_Antidiffusive_Velocities_for_Multipass_Donor_Cell_Advection).

We don't need to merge this PR, but I'll leave it here in case, in the future, we need a multidimensional positivity-preserving advection scheme (for example for wetting and drying).


src/ImmersedBoundaries/ImmersedBoundaries.jl
```diff
@@ -102,26 +102,27 @@ abstract type AbstractImmersedBoundary end
...
 #####
 
-struct ImmersedBoundaryGrid{FT, TX, TY, TZ, G, I, M, Arch} <: AbstractGrid{FT, TX, TY, TZ, Arch}
+struct ImmersedBoundaryGrid{FT, TX, TY, TZ, G, I, M, S, Arch} <: AbstractGrid{FT, TX, TY, TZ, Arch}
     architecture :: Arch
     underlying_grid :: G
     immersed_boundary :: I
     interior_active_cells :: M
-    
+    surface_active_cells :: S
```
glwagner, line 110: what surface is this, top, bottom, north, south?

simone-silvestri: That is meant to be the top boundary. It comes from PR #3404. I have to correct it there.

simone-silvestri: Actually, to be precise, it's the active map of reduced fields in the z-direction

glwagner: That is only relevant for the hydrostatic model right? It wouldn't apply to the non-hydrostatic model, or probably other models that are isotropic and 3D.

francispoulin:
@simone-silvestri Have you already done tests to ensure that it is positive preserving?

simone-silvestri:
Indeed, it is positivity preserving. Not bounds preserving though. And it seems to be too diffusive.

francispoulin:
> Indeed, it is positivity preserving. Not bounds preserving though. And it seems to be too diffusive.

That's a big step forward!  I know there are ways to iterate to improve the accuracy.  I presume your test was the low order option?  

Do you think we should try more iterations to improve the upper bound and general accuracy?

simone-silvestri:
This is the output of the 1D advection test in `validation/advection/simple_one_dimensional_advection.jl`

Screenshot 2024-02-13 at 12 41 25 PM

simone-silvestri:
with a 10 times smaller time step
Screenshot 2024-02-13 at 12 45 03 PM

src/TimeSteppers/runge_kutta_3.jl
```diff
@@ -104,6 +105,8 @@ function time_step!(model::AbstractModel{<:RungeKutta3TimeStepper}, Δt; callbac
     calculate_pressure_correction!(model, first_stage_Δt)
     pressure_correct_velocities!(model, first_stage_Δt)
 
+    correct_advection!(model, Δt)
```
glwagner, line 108: what's wrong about it

simone-silvestri: that it is only upwind first order, definitely incorrect 😄

simone-silvestri:
Nono, it is not bounds preserving. It is only intended for positive definite quantities and prevents negative values.

src/TimeSteppers/runge_kutta_3.jl
glwagner: maybe "refine advective tendencies" is the right word?

francispoulin:
I produced a few more figures that might be of interest.

First, it's the same plot as @simone-silvestri produced but with fifth order upwinding.

N100

francispoulin:
The second is with double the spatial resolution, and therefore halving the time step. 
N200

francispoulin:
The third and final one is with doubling the spatial resolution again (total of x4) and calculating the time step in the same way as  usual. 

N400

francispoulin:
Some thoughts:

The first case (`N=100`) doesn't look great but it shows that the MPData schemes do not differ very much whether we have 2 or 5 iteractions, and they are somewhere between 1st and 3rd order upwinding.

The second case (`N=200`) looks better and I can't really see any difference between 2 and 5 iterations.

The third case (`N=400`) has all the schemes performing well, even first order upwinding.  Again we can't differentiate between the two MPData schemes and it's even hard to differentiate between 3rd and 5th order upwinding.

I lesson that I learned from here is to use as much spatial resolution as we can as this does much more compared to the number of iterations.  

@simone-silvestri : do you think we should do a similar comparison for 2D advection?

simone-silvestri:
Thanks for running these tests.  It would be nice to see how MPData performs in a 2D flow. 

It's interesting that tweaking the number of iterations doesn't seem to make much difference. I'm a bit skeptical about this scheme though; it seems pretty diffusive and doesn't even beat third-order upwind which is way simpler. I might have introduced a bug although I went through the code again and it looks correct to me.

Another possibility is that the high diffusivity is the trade-off for keeping the scheme positivity-preserving?
Godunov's theorem sets some tough standards -- it's not possible to obtain monotonic advection with a linear scheme beyond first order.

francispoulin:
I looked into MPData a long time ago and I recall it being very diffusive, so it might be the nature of the beast.  It beats first order upwinding, which is also positive preserving, but not by much.

The WENO schemes seem to be a more complicated way to ensure positivity more accurately but I remember you saying there were problems doing this properly in multiple dimensions.  There is no perfect scheme for sure.

This is a[ paper](https://core.ac.uk/download/pdf/82262632.pdf) that presents a positive preserving WENO scheme for population models.  Example 3 is a 2D problem but the plots are only 1D.  How similar is this to what you have tried already?

ali-ramadhan:
I feel like this could be quite useful, especially for OceanBioME.jl users! Is there any appetite for positivity-preserving advection schemes?

@simone-silvestri How much work do you think it would be to get this PR ready to be merged? I'm happy to do the work, but probably won't revisit for at least a few weeks.

simone-silvestri:
This PR is quite old, however, the advection module did not change much, so I guess it would not take too much to revamp this. I remember I was not quite satisfied with the results though. Have you tried with WENO using the `bounds = (min, max)` kwarg? I think the results might be better even if I am not 100% sure it is bounds preserving.

ali-ramadhan:
Oh wow I obviously don't read docstrings because I had no idea `WENO` has a `bounds` kwarg. I guess passing something like `bounds = (0, typemax(FT))` should provide a positivity-preserving WENO scheme. I'll give it a try!

Yeah if this makes WENO positivity-preserving and it's better than MPData in every way then I suppose there's no need to introduce extra code by merging this PR.

ali-ramadhan:
Checked out this branch and reproduced the figure with with WENO-5 with bounds as well. Not sure why MPData looks worse but there have been some changes since the latest figure in this PR. WENO-5 with bounds is looking good in 1D!

advection

francispoulin:
I would also like to see this get done and happy to help if/when I can.  I am pretty sure this is easier to ensure positivity than WENO.  I remember wanting to look at it but never did. I'll try and look at it again.  Happy to have a chat sometime if people like.

glwagner:
go WENO 💪 boo MPDATA 👎 

just kidding let the best scheme win!

glwagner:
> Checked out this branch and reproduced the figure with with WENO-5 with bounds as well. Not sure why MPData looks worse but there have been some changes since the latest figure in this PR. WENO-5 with bounds is looking good in 1D!
> 
> advection

are you using the same time-stepper? There may have been some changes there.

How expensive is an MPData iteration? Is it pretty cheap on GPU so we can afford to increase that?

ali-ramadhan:
@simone-silvestri @francispoulin I think you guys are right. I tried 2D diagonal advection and WENO-5 does not actually preserve bounds with multiple dimensions (although it's not very negative). MPData does a better job but is not 100% positivity-preserving. MPData is both quite diffusive and dispersive though.

Here any negative value is colored in red, and any value greater than 1 is colored in yellow.

https://github.com/user-attachments/assets/79e7bb2f-727f-44a8-8754-aa08c7edad68

Generated using this AI-generated validation script:

```julia
using Oceananigans
using Oceananigans.Advection: MPData, WENO
using Oceananigans.Models.ShallowWaterModels: ConservativeFormulation
using CairoMakie
using Colors
using Printf
using Adapt

"""
This simulation is a simple 2D advection to test and compare
advection schemes with diagonal flow (u=1, v=1)
"""

arch = CPU()

# Parameters
N = 100

# 2D grid construction
Nx, Ny = N, N
grid = RectilinearGrid(arch,
                       size = (Nx, Ny),
                       x = (-1, 1),
                       y = (-1, 1),
                       z = (0, 1),
                       halo = (7, 7, 0),
                       topology = (Periodic, Periodic, Flat))

# Initial condition helper functions
@inline G2D(x, y, β, cx, cy) = exp(-β*((x - cx)^2 + (y - cy)^2))
@inline F2D(x, y, α, cx, cy) = √(max(1 - α^2*((x-cx)^2 + (y-cy)^2), 0.0))
@inline cone2D(x, y, cx, cy, r) = max(1.0 - √((x-cx)^2 + (y-cy)^2) / r, 0.0)

# 2D initial condition with multiple shapes
@inline function c₀_2D(x, y)
    # Parameters for different shapes
    β = 100.0  # Gaussian width parameter
    α = 10.0   # Smooth bump parameter

    # 2D Gaussian bump (centered at (-0.7, -0.7))
    if (x + 0.7)^2 + (y + 0.7)^2 <= 0.1^2
        return G2D(x, y, β, -0.7, -0.7)

    # 2D square (region: -0.4 ≤ x ≤ -0.2, -0.4 ≤ y ≤ -0.2)
    elseif x >= -0.4 && x <= -0.2 && y >= -0.4 && y <= -0.2
        return 1.0

    # 2D cone/pyramid (centered at (0.1, 0.1))
    elseif (x - 0.1)^2 + (y - 0.1)^2 <= 0.1^2
        return cone2D(x, y, 0.1, 0.1, 0.1)

    # 2D smooth bump (centered at (0.5, 0.5))
    elseif (x - 0.5)^2 + (y - 0.5)^2 <= 0.1^2
        return F2D(x, y, α, 0.5, 0.5)

    else
        return 0.0
    end
end

@info "Setting up 2D advection test"

# Time stepping parameters
Δt_max   = 0.2 * minimum_xspacing(grid)
end_time = 2.0

@show tot_iter = end_time ÷ Δt_max

# Store initial condition
c_real = CenterField(grid)
set!(c_real, c₀_2D)

solution = Dict()
time_series = Dict()  # Store all time steps for animation

# Advection schemes to test
schemes = [Centered(; order = 4),
           UpwindBiased(; order = 1),
           UpwindBiased(; order = 3),
           MPData(grid; iterations = 2),
           MPData(grid; iterations = 5),
           WENO(; order = 5, bounds = (0, typemax(Int)))]
names   = [:C, :U1, :U3, :M2, :M5, :W5]
labels  = ["Centered order 4",
           "Upwind order 1",
           "Upwind order 3",
           "MPData 2 iterations",
           "MPData 5 iterations",
           "WENO5 with bounds"]

# Run simulations for each scheme
for (tracer_advection, name, label) in zip(schemes, names, labels)
    @info "Running scheme: $(summary(tracer_advection))"

    model = HydrostaticFreeSurfaceModel(;
        grid = grid,
        tracers = :c,
        velocities = PrescribedVelocityFields(
            u = (x, y, z) -> 1.0,  # Diagonal advection
            v = (x, y, z) -> 1.0,
            w = (x, y, z) -> 0.0
        ),
        buoyancy = nothing,
        tracer_advection
    )

    set!(model, c=c₀_2D)
    c = model.tracers.c

    # Store time series
    snapshots = []
    times = Float64[]

    # Store initial condition
    push!(snapshots, Array(interior(c, :, :, 1)))
    push!(times, model.clock.time)

    # Time integration with snapshots
    for i = 1:Int(tot_iter)
        time_step!(model, Δt_max)

        # Store snapshot
        push!(snapshots, Array(interior(c, :, :, 1)))
        push!(times, model.clock.time)
    end

    # Store final result and time series
    csim = snapshots[end]
    solution[name] = csim
    time_series[name] = (snapshots = snapshots, times = times)

    @info "  Min: $(minimum(csim)), Max: $(maximum(csim))"
end

# Create modified solution for exact cutoffs visualization
# Clamp out-of-bounds values to flag values for exact color mapping
solution_viz = Dict()
for name in names
    data = copy(solution[name])
    # Track violations
    n_neg = sum(data .< 0)
    n_over = sum(data .> 1)

    # Clamp to flag values for exact cutoff visualization
    data[data .< 0] .= -0.05  # Flag for negative (will be red)
    data[data .> 1] .= 1.05   # Flag for overshoot (will be yellow)

    solution_viz[name] = data

    if n_neg > 0 || n_over > 0
        @info "  Scheme $name: $n_neg negative values, $n_over overshoot values"
    end
end

@info "Creating visualization"

# Create figure with 2×3 subplot grid
fig = Figure(size=(1800, 1200))

# Create custom colormap with exact cutoffs at 0 and 1
# Using thermal-style gradient for 0-1 range
function create_exact_cutoff_colormap()
    # Colors for violations
    neg_color = colorant"red"
    over_color = colorant"yellow"

    # Thermal-style gradient for 0-1 range (dark to bright, high contrast)
    # Using colors similar to cmocean's thermal: dark blue-black through red to yellow-white
    n_neg = 10      # Region for negative values (-0.05 to 0)
    n_normal = 236  # Region for valid range (0 to 1)
    n_over = 10     # Region for overshoot (1 to 1.05)

    # Thermal gradient with high dynamic range
    thermal_gradient = range(
        colorant"#041526",  # Very dark blue-black (0.0)
        colorant"#FFFAD4",  # Pale yellow-white (1.0)
        length = n_normal
    )

    # Alternative with more color variety (uncomment to use):
    # thermal_gradient = [range(colorant"#041526", colorant"#8B1A1A", length=n_normal÷3);
    #                    range(colorant"#8B1A1A", colorant"#FF8C00", length=n_normal÷3);
    #                    range(colorant"#FF8C00", colorant"#FFFAD4", length=n_normal - 2*(n_normal÷3))]

    colors = vcat(
        fill(neg_color, n_neg),
        collect(thermal_gradient),
        fill(over_color, n_over)
    )

    return colors
end

custom_cmap = create_exact_cutoff_colormap()
crange = (-0.05, 1.05)

# Plot each scheme
for (idx, (name, label)) in enumerate(zip(names, labels))
    row = (idx - 1) ÷ 3 + 1
    col = (idx - 1) % 3 + 1

    local ax = Axis(fig[row, col],
              title = label,
              xlabel = "x",
              ylabel = "y",
              aspect = DataAspect())

    hm = heatmap!(ax, solution_viz[name],
                  colormap = custom_cmap,
                  colorrange = crange)

    # Add colorbar only to the last plot in each row
    if col == 3
        cb = Colorbar(fig[row, 4], hm, label = "Tracer concentration",
                      ticks = ([-0.05, 0.0, 0.25, 0.5, 0.75, 1.0, 1.05],
                               ["<0", "0", "0.25", "0.5", "0.75", "1", ">1"]))
    end
end

# Save figure
output_file = "advection_2d_comparison.png"
save(output_file, fig)

@info "Saved figure to $output_file"

# Create animation showing time evolution
@info "Creating animation..."

# Prepare time series data with clamped values for visualization
time_series_viz = Dict()
for name in names
    snapshots = time_series[name].snapshots
    times = time_series[name].times

    # Clamp each snapshot
    clamped_snapshots = []
    for snapshot in snapshots
        data = copy(snapshot)
        data[data .< 0] .= -0.05
        data[data .> 1] .= 1.05
        push!(clamped_snapshots, data)
    end

    time_series_viz[name] = (snapshots = clamped_snapshots, times = times)
end

# Create animation with all schemes in a 2×3 grid
anim_fig = Figure(size=(1800, 1200))

# Create observables for each scheme's data
data_observables = []
for name in names
    push!(data_observables, Observable(time_series_viz[name].snapshots[1]))
end

# Create axes and heatmaps for each scheme
for (idx, (name, label)) in enumerate(zip(names, labels))
    row = (idx - 1) ÷ 3 + 1
    col = (idx - 1) % 3 + 1

    local ax = Axis(anim_fig[row, col],
              title = label,
              xlabel = "x",
              ylabel = "y",
              aspect = DataAspect())

    # Create heatmap with observable
    hm = heatmap!(ax, data_observables[idx],
                  colormap = custom_cmap,
                  colorrange = crange)

    # Add colorbar only to the last plot in each row
    if col == 3
        cb = Colorbar(anim_fig[row, 4], hm, label = "Tracer concentration",
                      ticks = ([-0.05, 0.0, 0.25, 0.5, 0.75, 1.0, 1.05],
                               ["<0", "0", "0.25", "0.5", "0.75", "1", ">1"]))
    end
end

# Add time text as an observable
time_observable = Observable("Time: 0.00")
time_text = Label(anim_fig[0, :], time_observable, fontsize = 24)

# Create animation
n_frames = length(time_series_viz[names[1]].times)
fps = 20  # Frames per second

@info "Recording animation with $n_frames frames..."

record(anim_fig, "advection_2d_animation.mp4", 1:n_frames; framerate = fps) do frame
    # Update time label
    t = time_series_viz[names[1]].times[frame]
    time_observable[] = @sprintf("Time: %.3f", t)

    # Update each heatmap data
    for (idx, name) in enumerate(names)
        data_observables[idx][] = time_series_viz[name].snapshots[frame]
    end
end

@info "Saved animation to advection_2d_animation.mp4"
```

ali-ramadhan:
> are you using the same time-stepper? There may have been some changes there.

Ah I just checked out this branch so I'm actually on Oceananigans v0.96.26. But yeah there have been some changes to this PR (including multiple merge commits) since @simone-silvestri posted the latest figure before mine.

> How expensive is an MPData iteration? Is it pretty cheap on GPU so we can afford to increase that?

I have not benchmarked but earlier in this thread it was suggested that more iterations doesn't do much, so I didn't go beyond the 5 used in the 1D advection validation.

ali-ramadhan:
Better animation and also comparing WENO5 with and without bounds. It's better with bounds but it's bounds-preserving is not guaranteed.

https://github.com/user-attachments/assets/5b583ec8-ed2e-46cc-a4bc-7335f5b8f2a1

Updated script:

```julia
using Oceananigans
using Oceananigans.Advection: MPData, WENO
using Oceananigans.Models.ShallowWaterModels: ConservativeFormulation
using CairoMakie
using Colors
using Printf
using Adapt

"""
This simulation is a simple 2D advection to test and compare
advection schemes with diagonal flow (u=1, v=1)
"""

arch = CPU()

# Parameters
N = 100

# 2D grid construction
Nx, Ny = N, N
grid = RectilinearGrid(arch,
                       size = (Nx, Ny),
                       x = (-1, 1),
                       y = (-1, 1),
                       z = (0, 1),
                       halo = (7, 7, 0),
                       topology = (Periodic, Periodic, Flat))

# Initial condition helper functions
@inline G2D(x, y, β, cx, cy) = exp(-β*((x - cx)^2 + (y - cy)^2))
@inline F2D(x, y, α, cx, cy) = √(max(1 - α^2*((x-cx)^2 + (y-cy)^2), 0.0))
@inline cone2D(x, y, cx, cy, r) = max(1.0 - √((x-cx)^2 + (y-cy)^2) / r, 0.0)

# 2D initial condition with multiple shapes
@inline function c₀_2D(x, y)
    # Parameters for different shapes
    β = 100.0  # Gaussian width parameter
    α = 10.0   # Smooth bump parameter

    # 2D Gaussian bump (centered at (-0.7, -0.7))
    if (x + 0.7)^2 + (y + 0.7)^2 <= 0.1^2
        return G2D(x, y, β, -0.7, -0.7)

    # 2D square (region: -0.4 ≤ x ≤ -0.2, -0.4 ≤ y ≤ -0.2)
    elseif x >= -0.4 && x <= -0.2 && y >= -0.4 && y <= -0.2
        return 1.0

    # 2D cone/pyramid (centered at (0.1, 0.1))
    elseif (x - 0.1)^2 + (y - 0.1)^2 <= 0.1^2
        return cone2D(x, y, 0.1, 0.1, 0.1)

    # 2D smooth bump (centered at (0.5, 0.5))
    elseif (x - 0.5)^2 + (y - 0.5)^2 <= 0.1^2
        return F2D(x, y, α, 0.5, 0.5)

    else
        return 0.0
    end
end

@info "Setting up 2D advection test"

# Time stepping parameters
Δt_max   = 0.2 * minimum_xspacing(grid)
end_time = 2.0

@show tot_iter = end_time ÷ Δt_max

# Store initial condition
c_real = CenterField(grid)
set!(c_real, c₀_2D)

solution = Dict()
time_series = Dict()  # Store all time steps for animation

# Advection schemes to test
schemes = [Centered(; order = 4),
           UpwindBiased(; order = 1),
           UpwindBiased(; order = 3),
           MPData(grid; iterations = 5),
           WENO(; order = 5),
           WENO(; order = 5, bounds = (0, 1))]
names   = [:C, :U1, :U3, :M5, :W5, :W5B]
labels  = ["Centered order 4",
           "Upwind order 1",
           "Upwind order 3",
           "MPData 5 iterations",
           "WENO5",
           "WENO5 with bounds"]

# Run simulations for each scheme
for (tracer_advection, name, label) in zip(schemes, names, labels)
    @info "Running scheme: $(summary(tracer_advection))"

    model = HydrostaticFreeSurfaceModel(;
        grid = grid,
        tracers = :c,
        velocities = PrescribedVelocityFields(
            u = (x, y, z) -> 1.0,  # Diagonal advection
            v = (x, y, z) -> 1.0,
            w = (x, y, z) -> 0.0
        ),
        buoyancy = nothing,
        tracer_advection
    )

    set!(model, c=c₀_2D)
    c = model.tracers.c

    # Store time series
    snapshots = []
    times = Float64[]

    # Store initial condition
    push!(snapshots, Array(interior(c, :, :, 1)))
    push!(times, model.clock.time)

    # Time integration with snapshots
    for i = 1:Int(tot_iter)
        time_step!(model, Δt_max)

        # Store snapshot
        push!(snapshots, Array(interior(c, :, :, 1)))
        push!(times, model.clock.time)
    end

    # Store final result and time series
    csim = snapshots[end]
    solution[name] = csim
    time_series[name] = (snapshots = snapshots, times = times)

    @info "  Min: $(minimum(csim)), Max: $(maximum(csim))"
end

# Create modified solution for exact cutoffs visualization
# Clamp out-of-bounds values to flag values for exact color mapping
solution_viz = Dict()
for name in names
    data = copy(solution[name])
    # Track violations
    n_neg = sum(data .< 0)
    n_over = sum(data .> 1)

    # Clamp to flag values for exact cutoff visualization
    data[data .< 0] .= -0.05  # Flag for negative (will be red)
    data[data .> 1] .= 1.05   # Flag for overshoot (will be yellow)

    solution_viz[name] = data

    if n_neg > 0 || n_over > 0
        @info "  Scheme $name: $n_neg negative values, $n_over overshoot values"
    end
end

@info "Creating visualization"

# Create figure with 2×3 subplot grid
fig = Figure(size=(1800, 1200))

# Create custom colormap with exact cutoffs at 0 and 1
# Using thermal-style gradient for 0-1 range
function create_exact_cutoff_colormap()
    # Colors for violations
    neg_color = colorant"red"
    over_color = colorant"yellow"

    # Thermal-style gradient for 0-1 range (dark to bright, high contrast)
    # Using colors similar to cmocean's thermal: dark blue-black through red to yellow-white
    n_neg = 10      # Region for negative values (-0.05 to 0)
    n_normal = 236  # Region for valid range (0 to 1)
    n_over = 10     # Region for overshoot (1 to 1.05)

    # Thermal gradient with high dynamic range
    thermal_gradient = range(
        colorant"#041526",  # Very dark blue-black (0.0)
        colorant"#FFFAD4",  # Pale yellow-white (1.0)
        length = n_normal
    )

    # Alternative with more color variety (uncomment to use):
    # thermal_gradient = [range(colorant"#041526", colorant"#8B1A1A", length=n_normal÷3);
    #                    range(colorant"#8B1A1A", colorant"#FF8C00", length=n_normal÷3);
    #                    range(colorant"#FF8C00", colorant"#FFFAD4", length=n_normal - 2*(n_normal÷3))]

    colors = vcat(
        fill(neg_color, n_neg),
        collect(thermal_gradient),
        fill(over_color, n_over)
    )

    return colors
end

custom_cmap = create_exact_cutoff_colormap()
crange = (-0.05, 1.05)

# Plot each scheme
for (idx, (name, label)) in enumerate(zip(names, labels))
    row = (idx - 1) ÷ 3 + 1
    col = (idx - 1) % 3 + 1

    local ax = Axis(fig[row, col],
              title = label,
              xlabel = "x",
              ylabel = "y",
              aspect = DataAspect())

    hm = heatmap!(ax, solution_viz[name],
                  colormap = custom_cmap,
                  colorrange = crange)

    # Add colorbar only to the last plot in each row
    if col == 3
        cb = Colorbar(fig[row, 4], hm, label = "Tracer concentration",
                      ticks = ([-0.05, 0.0, 0.25, 0.5, 0.75, 1.0, 1.05],
                               ["<0", "0", "0.25", "0.5", "0.75", "1", ">1"]))
    end
end

# Save figure
output_file = "advection_2d_comparison.png"
save(output_file, fig)

@info "Saved figure to $output_file"

# Create animation showing time evolution
@info "Creating animation..."

# Prepare time series data with clamped values for visualization
# Also track min/max before clamping for display
time_series_viz = Dict()
min_max_values = Dict()
for name in names
    snapshots = time_series[name].snapshots
    times = time_series[name].times

    # Track min/max before clamping
    mins = Float64[]
    maxs = Float64[]

    # Clamp each snapshot
    clamped_snapshots = []
    for snapshot in snapshots
        # Store actual min/max BEFORE clamping
        push!(mins, minimum(snapshot))
        push!(maxs, maximum(snapshot))

        # Now clamp for visualization
        data = copy(snapshot)
        data[data .< 0] .= -0.05
        data[data .> 1] .= 1.05
        push!(clamped_snapshots, data)
    end

    time_series_viz[name] = (snapshots = clamped_snapshots, times = times)
    min_max_values[name] = (mins = mins, maxs = maxs)
end

# Create animation with all schemes in a 2×3 grid
anim_fig = Figure(size=(1800, 1200))

# Create observables for each scheme's data and titles
data_observables = []
title_observables = []
for (name, label) in zip(names, labels)
    push!(data_observables, Observable(time_series_viz[name].snapshots[1]))
    # Initial title with min/max
    min_val = min_max_values[name].mins[1]
    max_val = min_max_values[name].maxs[1]
    initial_title = @sprintf("%s (min=%.5f, max=%.5f)", label, min_val, max_val)
    push!(title_observables, Observable(initial_title))
end

# Create axes and heatmaps for each scheme
for (idx, name) in enumerate(names)
    row = (idx - 1) ÷ 3 + 1
    col = (idx - 1) % 3 + 1

    local ax = Axis(anim_fig[row, col],
              title = title_observables[idx],
              xlabel = "x",
              ylabel = "y",
              aspect = DataAspect())

    # Create heatmap with observable
    hm = heatmap!(ax, data_observables[idx],
                  colormap = custom_cmap,
                  colorrange = crange)

    # Add colorbar only to the last plot in each row
    if col == 3
        cb = Colorbar(anim_fig[row, 4], hm, label = "Tracer concentration",
                      ticks = ([-0.05, 0.0, 0.25, 0.5, 0.75, 1.0, 1.05],
                               ["<0", "0", "0.25", "0.5", "0.75", "1", ">1"]))
    end
end

# Add time text as an observable
time_observable = Observable("Time: 0.00")
time_text = Label(anim_fig[0, :], time_observable, fontsize = 24)

# Create animation
n_frames = length(time_series_viz[names[1]].times)
fps = 20  # Frames per second

@info "Recording animation with $n_frames frames..."

record(anim_fig, "advection_2d_animation.mp4", 1:n_frames; framerate = fps) do frame
    # Update time label
    t = time_series_viz[names[1]].times[frame]
    time_observable[] = @sprintf("Time: %.3f", t)

    # Update each heatmap data and title with min/max
    for (idx, name) in enumerate(names)
        data_observables[idx][] = time_series_viz[name].snapshots[frame]

        # Update title with current min/max values
        min_val = min_max_values[name].mins[frame]
        max_val = min_max_values[name].maxs[frame]
        title_observables[idx][] = @sprintf("%s (min=%.5f, max=%.5f)", labels[idx], min_val, max_val)
    end
end

@info "Saved animation to advection_2d_animation.mp4"
```

glwagner:
Pretty awesome animation. For WENO a lot of the red is in regions with little variance, which is interesting.

simone-silvestri:
A way to easily extend the 1D bounds preserving property to 2D and 3D is to do successive 1D reconstructions a la MITgcm https://mitgcm.org/sealion/online_documents/node80.html in this way we are guaranteed that the 1D properties apply also to multi-dimensional reconstructions.

It is kind of annoying to do it because we have to split the kernels and do successive interleaved fill halos, but I might have a stab at it when I finish with the timestepper.

francispoulin:
Very nice plots @ali-ramadhan !  Interesting that AI did all that.  It works and I believe it but wow, is that a lot of lines of code!

I believe that MPData should preserve the bounds exactly at least when you have low order methods.  Not sure why it doesn't.   But the plot clearly shows how diffusive it is.

So far first order upwinding is winning!

glwagner:
> A way to easily extend the 1D bounds preserving property to 2D and 3D is to do successive 1D reconstructions a la MITgcm https://mitgcm.org/sealion/online_documents/node80.html in this way we are guaranteed that the 1D properties apply also to multi-dimensional reconstructions.
> 
> It is kind of annoying to do it because we have to split the kernels and do successive interleaved fill halos, but I might have a stab at it when I finish with the timestepper.

We can add a property of the advection scheme to turn this on and off, similar to what we do with vertically implicit time discretization.

ali-ramadhan:
> Very nice plots @ali-ramadhan !  Interesting that AI did all that.  It works and I believe it but wow, is that a lot of lines of code!

@francispoulin Claude Code can so some pretty cool stuff with the right context! I basically pointed it to the 1D script then asked it to turn it into a 2D validation script (it also had access to the entire codebase). Then iterated by asking for a movie, then improved the color bars then the plot subtitles.

francispoulin:
Very impressive!  I should try it sometime. Thanks.

simone-silvestri:
@glwagner, a positive definite advection might also be very interesting for moisture in the atmospheric model.

glwagner:
> @glwagner, a positive definite advection might also be very interesting for moisture in the atmospheric model.

The benefits of that are definitely not lost on me!

glwagner:
Something that might matter, but is also annoying, is that it looks like the MITgcm scheme requires saving at least one copy of the tracer fields that using dimension-by-dimension advection. So a nice implementation would actually only save extra data for those tracers that are being advected like this. In turn that motivates generalizing the `TimeStepper` to handle this (similar to how we have `implicit_solver`. All doable, but adds a few moving pieces. I could look into designing something if there is interest.

glwagner:
I think this paper describes a flux limiting method that does not require splitting up the time step as done in MITgcm: https://journals.ametsoc.org/view/journals/mwre/117/11/1520-0493_1989_117_2626_copdas_2_0_co_2.pdf

francispoulin:
> I think this paper describes a flux limiting method that does not require splitting up the time step as done in MITgcm: https://journals.ametsoc.org/view/journals/mwre/117/11/1520-0493_1989_117_2626_copdas_2_0_co_2.pdf

I believe that the author of this paper is the person who created MPData, but I'm not sure if that's exactly what this PR is doing.

glwagner:

simone-silvestri:
I would probably close it. If we need it we can reopen it

PR #3441, opened 2024-01-24
Add missing `adapt_structure` for stokes drift

jagoosw:
(Btw I came across this trying to pass parameters that aren't just a named tuple of numbers.)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.90.4"
+version = "0.90.5"
 
 [deps]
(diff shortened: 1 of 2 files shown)


glwagner (approved):

jagoosw:
@glwagner could I be added to the buildkite organisation so I can rerun the test please?

glwagner:
> @glwagner could I be added to the buildkite organisation so I can rerun the test please?

Unfortunately I think we have to limit access to that to Clima folks, but it affects all Clima software (not just Oceananigans). As an example, none of us have permissions to add people to the buildkite.

jagoosw:
> Unfortunately I think we have to limit access to that to Clima folks, but it affects all Clima software (not just Oceananigans). As an example, none of us have permissions to add people to the buildkite.

No worries makes sense!

navidcy:
Reminder to add the `(0.90.5)` at the beginning of titles of PR that change the version number...

PR #3450, opened 2024-01-29
(0.90.8) Rules for `Time` interpolation / extrapolation for FieldTimeSeries and `InterpolatedField`

### LINKED ISSUE #3472: `FieldTimeSeries` does not adapt correctly on GPU
I can't quite identify where this issue is coming from but if you load a `FieldTimeSeries` in the normal way, i.e.:
```julia
u = FieldTimeSeries("path.jld2", "u");
```
when it gets adapted the `data` and `times` are not adapted to GPU arrays if the `FieldTimeSeries` is adapted to `GPUAdaptedFieldTimeSeries` when it's passed to a model etc. If instead I specify the grid like:
```julia
u = FieldTimeSeries("path.jld2", "u"; grid);
```
then the data is correctly adapted but the `times` are still a regular vector.

I can of course manually create the adapted time series like:
```julia
u = FieldTimeSeries("path.jld2", "u"; grid);
u_adapt = GPUAdaptedFieldTimeSeries{Face, Center, Center, eltype(grid)}(arch_array(arch, u.data), arch_array(arch, u.times))
```
and also defining:
```julia
adapt_structure(to, fts::GPUAdaptedFieldTimeSeries{LX, LY, LZ, FT}) where {LX, LY, LZ, FT} = 
    GPUAdaptedFieldTimeSeries{LX, LY, LZ, FT}(adapt(to, fts.data),
                                              adapt(to, fts.times))
```
but this should probably be part of the default behaviour.

glwagner:
This PR will implement boundary conditions _in time_ for `FieldTimeSeries` (ie so we can cycle through a `FieldTimeSeries`) and also a field that interpolates into another field on the fly as it is indexed.

with @simone-silvestri 

Resolves #3472


src/OutputReaders/set_field_time_series.jl
```diff
@@ -0,0 +1,91 @@
...
+"""
+function set!(fts::OnDiskFTS, field::Field, n::Int, time=fts.times[n])
+    fts.grid == field.grid || error("The grids attached to the Field and \
+                                    FieldTimeSeries appear to be different.")
+    path = fts.path
+    name = fts.name
+    jldopen(path, "a+") do file
+        initialize_file!(file, name, fts)
+        maybe_write_property!(file, "timeseries/t/$n", time)
+        maybe_write_property!(file, "timeseries/$name/$n", parent(field))
```
simone-silvestri, line 78:
```suggestion
        maybe_write_property!(file, "timeseries/$name/$n", Array(parent(field)))
```

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/CATKEVerticalDiffusivities.jl
```diff
@@ -304,23 +304,56 @@ end
...
+        # So
+        #
+        #       Lⁱ e = wb - ϵ
+        #            = (wb / e - ω) e,
+        #               ↖--------↗
+        #                  = Lⁱ
+        #
+        # where ω = ϵ / e ∼ √e / ℓ.
+        
+        diffusivities.Lᵉ[i, j, k] = wb_e - ω + div_Jᵉ_e
```
simone-silvestri, line 356: shall we put all this in the CATKE PR?

simone-silvestri (approved):

PR #3452, opened 2024-02-01
Remove Enzyme from core dependency

navidcy:


wsmoses (approved):

navidcy:
@wsmoses quick question: how does the `gpu Enzyme extensions tests` set work when `test_enzyme.jl` hardcodes architectures as CPU()?

https://github.com/CliMA/Oceananigans.jl/blob/b8918c7d33089eafb2395fba3a9f5f3e5ad42d87/test/test_enzyme.jl#L12

PR #3453, opened 2024-02-02
Some CATKE performance optimizations

simone-silvestri:
build up on #3404.


simone-silvestri:
disregard the clipping of TKE to zero, it is just a stability test. In general this branch will remain a experimental for a little bit

src/Advection/tracer_advection_operators.jl
```diff
@@ -1,6 +1,34 @@
...
+    y :: B
+    z :: C
+
+    TracerAdvection{N, FT}(x::A, y::B, z::C) where {N, FT, A, B, C} = new{N, FT, A, B, C}(x, y, z)
+end
+
+"""
+    function TracerAdvection(; x, y, z)
+
+builds a `TracerAdvection` type with different reconstructions in `x`, `y`, and `z`
```
glwagner, line 15: This is wrong --- I don't think there is any enforcement of having different reconstructions. The description should simply mention that `x, y, z` represent reconstruction in the `x, y, z` direction.

src/Advection/tracer_advection_operators.jl
```diff
@@ -12,7 +12,7 @@ end
 """
     function TracerAdvection(; x, y, z)
 
-builds a `TracerAdvection` type with different reconstructions in `x`, `y`, and `z`
+Builds a `TracerAdvection` type with reconstructions schemes in `x`, `y`, and `z`
```
glwagner, line 15:
```suggestion
Builds a `TracerAdvection` type with reconstructions schemes in `x`, `y`, and `z`.
```

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
```diff
@@ -121,17 +119,17 @@ end
...
-    ω_numerical = 1 / closure.negative_turbulent_kinetic_energy_damping_time_scale
-    ω_physical = sqrt(abs(eᵢ)) / ℓᴰ
+    ω = sqrt(abs(eᵢ)) / ℓᴰ
 
-    return ifelse(eᵢ < 0, ω_numerical, ω_physical)
+    return ω
 end
 
-# Fallbacks for explicit time discretization
+# Dissipation: if e is negative treat it implicitly, otherwise explicit (we do not want to subtract
```
glwagner:
```suggestion
# Dissipation: if e is negative treat it explictly, otherwise implicit (we do not want to subtract
```

glwagner: is that what you meant?

src/Advection/tracer_advection_operators.jl
```diff
@@ -1,6 +1,34 @@
...
+
+    FT = eltype(x)
+
+    return TracerAdvection{max(Nx, Ny, Nz), FT}(x, y, z)
+end
+
+Adapt.adapt_structure(to, scheme::TracerAdvection{N, FT}) where {N, FT} =
+    TracerAdvection{N, FT}(Adapt.adapt(to, scheme.x),
+                           Adapt.adapt(to, scheme.y),
+                           Adapt.adapt(to, scheme.z))
```
glwagner, line 30: What does this stuff have to do with CATKE performance optimization?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/CATKEVerticalDiffusivities.jl
```diff
@@ -315,12 +349,13 @@ end
         on_bottom = !inactive_cell(i, j, k, grid) & inactive_cell(i, j, k-1, grid)
         Δz = Δzᶜᶜᶜ(i, j, k, grid)
         Cᵂϵ = closure_ij.turbulent_kinetic_energy_equation.Cᵂϵ
-        Q_e = - Cᵂϵ * turbulent_velocityᶜᶜᶜ(i, j, k, grid, closure_ij, tracers.e) / Δz * on_bottom
+        Q_e = - Cᵂϵ * w★[i, j, k] / Δz * on_bottom
```
glwagner, line 352: We should discuss whether it would be better to keep the design that uses the function `turbulent_velocity`. I tried to write this code so that we could experiment with putting TKE at cell interfaces. All of these changes hard code the TKE location, making it harder to modify in the future...

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/CATKEVerticalDiffusivities.jl
```diff
@@ -249,47 +254,76 @@ function compute_diffusivities!(diffusivities, closure::FlavorOfCATKE, model; pa
...
+    e  = tracers.e
+    S² = diffusivities.S²
+    N² = diffusivities.N²
+    w★ = diffusivities.w★
+    u, v, _ = velocities
+
+    @inbounds begin
+        S²[i, j, k] = shearᶜᶜᶠ(i, j, k, grid, u, v)
+        N²[i, j, k] = ∂z_b(i, j, k, grid, buoyancy, tracers)
+        w★[i, j, k] = turbulent_velocityᶜᶜᶜ(i, j, k, grid, closure, tracers.e)
```
glwagner, line 284: Does this really need to be precomputed? It's just the sqrt of `e`.

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/mixing_length.jl
```diff
@@ -145,30 +141,27 @@ end
...
+    N²_above = ℑzᵃᵃᶜ(i, j, k+1, grid, N²)
 
     # "Convective length"
     # ℓᶜ ∼ boundary layer depth according to Deardorff scaling
     ℓᶜ = Cᶜ * w★³ / (Qᵇ + Qᵇᵋ)
     ℓᶜ = ifelse(isnan(ℓᶜ), zero(grid), ℓᶜ)
 
     # Figure out which mixing length applies
-    convecting = (Qᵇ > Qᵇᵋ) & (N² < 0)
+    convecting = (Qᵇ > Qᵇᵋ) & (N²_local < 0)
```
glwagner, line 161: "local" implies there is such a thing as a "non-local buoyancy gradient". But I don't think that's what you mean..

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/CATKEVerticalDiffusivities.jl
glwagner: It seems like the big one here is precomputing the buoyancy gradient, since that's potentially expensive.

Rather than completely nuking the existing code, it might make sense to generalize it instead so we can avoid this memory allocation if desired.

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
simone-silvestri: ah yeah 😅

src/Advection/tracer_advection_operators.jl
simone-silvestri: that is code from another PR #3404 we have to merge before this one

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/CATKEVerticalDiffusivities.jl
simone-silvestri: We probably don't even need that extra memory. It can be precomputed inside the `_compute_CATKE_diffusivities!` kernel (at centers and faces).

glwagner:
This is stale right?

simone-silvestri:
Yeah, the concepts introduced here can be applied to the new formulation of CATKE but since the implementation changed it is better to do it from scratch in another PR

PR #3458, opened 2024-02-05
Prevent duplicate tendency computation when simulaiton is running

jagoosw:
See https://github.com/CliMA/Oceananigans.jl/discussions/3457

Made PR to see if tests will pass.


jagoosw:
I've just realised I missed the point

PR #3467, opened 2024-02-08
Fix bug for lat-lon grid with Periodic `y`-dimension

### LINKED ISSUE #3465: `LatitudeLongitudeGrid` may have incorrect metrics for `y`-Periodic domains
This code:

```Julia
using Oceananigans
grid = LatitudeLongitudeGrid(longitude = (215, 215.25),
                             latitude = (50, 50.25),
                             size = (1, 1, 50),
                             halo = (1, 1, 1),
                             z = (-200, 0),
```

produces a `BoundsError`, if run with `julia --check-bounds=yes`.

The problem is the worksize specified here:

https://github.com/CliMA/Oceananigans.jl/blob/3b02321a350003fa9596c4f40839b031a5cc7ad1/src/Grids/latitude_longitude_grid.jl#L443

I also don't like this code because we cannot read it to understand it. For example, what is `length(grid.φᵃᶜᵃ)`? Is this `Ny + Hy`, or `Ny + Hy + 1`? I think it would be better to be explicit.

It appears that for both `Periodic` and `Bounded` in `y`, then we have

```julia
length(grid.φᵃᶜᵃ) = 3
```

Which looks like `Ny + 2Hy`.

I think both the worksize and the metric size is wrong in general --- I think we have hardcoded a `Bounded` y-topology. I think the right thing to do is to go in and use `total_length` to determine both the correct size of the metrics for any topology, as well as the size of the kernel that is launched to compute them.

navidcy:
Closes #3465

### DIFF
--- a/src/Grids/latitude_longitude_grid.jl
+++ b/src/Grids/latitude_longitude_grid.jl
@@ -437,10 +437,17 @@ end
 ##### Kernels that precompute the z- and x-metric
 #####
 
-@inline metric_worksize(grid::LatitudeLongitudeGrid)  = (length(grid.Δλᶜᵃᵃ), length(grid.φᵃᶜᵃ) - 1) 
+@inline function metric_worksize(grid::LatitudeLongitudeGrid)
+    TX, TY, _ = topology(grid)
+    return (total_length(Face(), TX(), grid.Nx, grid.Hx) - 2, total_length(Face(), TY(), grid.Ny, grid.Hy) - 2)
+end
 @inline metric_workgroup(grid::LatitudeLongitudeGrid) = (16, 16) 
 
-@inline metric_worksize(grid::XRegularLLG)  = length(grid.φᵃᶜᵃ) - 1 
+@inline function metric_worksize(grid::XRegularLLG)
+    _, TY, _ = topology(grid)
+    return total_length(Face(), TY(), grid.Ny, grid.Hy) - 2
+end
+
 @inline metric_workgroup(grid::XRegularLLG) = 16
 
 function precompute_curvilinear_metrics!(grid, Δxᶠᶜ, Δxᶜᶠ, Δxᶠᶠ, Δxᶜᶜ, Azᶠᶜ, Azᶜᶠ, Azᶠᶠ, Azᶜᶜ)
@@ -498,7 +505,9 @@ end
 
 function precompute_Δy_metrics(grid::LatitudeLongitudeGrid, Δyᶠᶜ, Δyᶜᶠ)
     arch = grid.architecture
-    precompute_Δy! = precompute_Δy_kernel!(Architectures.device(arch), 16, length(grid.Δφᵃᶜᵃ) - 1)
+    _, TY, _ = topology(grid)
+
+    precompute_Δy! = precompute_Δy_kernel!(Architectures.device(arch), 16, total_length(Face(), TY(), grid.Ny, grid.Hy) - 2)
     precompute_Δy!(grid, Δyᶠᶜ, Δyᶜᶠ)
     
     return Δyᶠᶜ, Δyᶜᶠ
--- a/src/Grids/latitude_longitude_grid.jl
+++ b/src/Grids/latitude_longitude_grid.jl
@@ -507,7 +507,7 @@ end
 
 function precompute_Δy_metrics(grid::LatitudeLongitudeGrid, Δyᶠᶜ, Δyᶜᶠ)
     arch = grid.architecture
-    _, TY, _ = topology(grid)
+    TY = topology(grid, 2)
 
     precompute_Δy! = precompute_Δy_kernel!(Architectures.device(arch), 16, total_length(Face(), TY(), grid.Ny, grid.Hy) - 2)
     precompute_Δy!(grid, Δyᶠᶜ, Δyᶜᶠ)
--- a/src/Grids/latitude_longitude_grid.jl
+++ b/src/Grids/latitude_longitude_grid.jl
@@ -90,7 +90,7 @@ regular_dimensions(::ZRegularLLG) = tuple(3)
                           precompute_metrics = true,
                           halo = nothing)
 
-Creates a `LatitudeLongitudeGrid` with coordinates `(λ, φ, z)` denoting longitude, latitude,
+Create a `LatitudeLongitudeGrid` with coordinates `(λ, φ, z)` denoting longitude, latitude,
 and vertical coordinate respectively.
 
 Positional arguments
@@ -208,7 +208,7 @@ function LatitudeLongitudeGrid(architecture::AbstractArchitecture = CPU(),
                                                          Δλᶠᵃᵃ, Δλᶜᵃᵃ, λᶠᵃᵃ, λᶜᵃᵃ,
                                                          Δφᵃᶠᵃ, Δφᵃᶜᵃ, φᵃᶠᵃ, φᵃᶜᵃ,
                                                          Δzᵃᵃᶠ, Δzᵃᵃᶜ, zᵃᵃᶠ, zᵃᵃᶜ,
-                                                         (nothing for i=1:10)..., FT(radius))
+                                                         (nothing for _ in 1:10)..., FT(radius))
 
     return !precompute_metrics ? preliminary_grid : with_precomputed_metrics(preliminary_grid)
 end
@@ -260,9 +260,9 @@ function validate_lat_lon_grid_args(topology, size, halo, FT, latitude, longitud
     λ₁ <= λ₂      || throw(ArgumentError("Longitudes must increase west to east."))
 
     φ₁, φ₂ = get_domain_extent(latitude, Nφ)
-    -90 <= φ₁ || throw(ArgumentError("The southernmost latitude cannot be less than -90 degrees."))
-    φ₂ <= 90  || throw(ArgumentError("The northern latitude cannot be less than -90 degrees."))
-    φ₁ <= φ₂  || throw(ArgumentError("Latitudes must increase south to north."))
+    -90 ≤ φ₁ || throw(ArgumentError("The southernmost latitude cannot be less than -90 degrees."))
+     φ₂ ≤ 90 || throw(ArgumentError("The northern latitude cannot be more than +90 degrees."))
+     φ₁ ≤ φ₂ || throw(ArgumentError("Latitudes must increase south to north."))
 
     if TX == Flat || TY == Flat 
         precompute_metrics = false
@@ -652,9 +652,9 @@ const C = Center
     view(grid.φᵃᶜᵃ, interior_indices(ℓy, topology(grid, 2)(), size(grid, 2)))
 
 @inline xnodes(grid::LLG, ℓx, ℓy; with_halos=false) =
-    grid.radius * deg2rad.(λnodes(grid, ℓx; with_halos=with_halos))' .* hack_cosd.(φnodes(grid, ℓy; with_halos=with_halos))
+    grid.radius * deg2rad.(λnodes(grid, ℓx; with_halos))' .* hack_cosd.(φnodes(grid, ℓy; with_halos))
 @inline ynodes(grid::LLG, ℓy; with_halos=false)     =
-    grid.radius * deg2rad.(φnodes(grid, ℓy; with_halos=with_halos))
+    grid.radius * deg2rad.(φnodes(grid, ℓy; with_halos))
 
 @inline znodes(grid::LLG, ℓz::F; with_halos=false) = with_halos ? grid.zᵃᵃᶠ :
     view(grid.zᵃᵃᶠ, interior_indices(ℓz, topology(grid, 3)(), size(grid, 3)))
--- a/src/Grids/latitude_longitude_grid.jl
+++ b/src/Grids/latitude_longitude_grid.jl
@@ -540,7 +540,7 @@ end
 
 function allocate_metrics(grid::LatitudeLongitudeGrid)
     FT = eltype(grid)
-    
+
     # preallocate quantities to ensure correct type and size
     grid_metrics = (:Δxᶠᶜ,
                     :Δxᶜᶠ,
@@ -583,7 +583,7 @@ function allocate_metrics(grid::LatitudeLongitudeGrid)
         Δyᶠᶜ    = OffsetArray(arch_array(arch, parentC), grid.Δφᵃᶜᵃ.offsets[1])
         Δyᶜᶠ    = OffsetArray(arch_array(arch, parentF), grid.Δφᵃᶜᵃ.offsets[1])
     end
-    
+
     return Δxᶠᶜ, Δxᶜᶠ, Δxᶠᶠ, Δxᶜᶜ, Δyᶠᶜ, Δyᶜᶠ, Azᶠᶜ, Azᶜᶠ, Azᶠᶠ, Azᶜᶜ
 end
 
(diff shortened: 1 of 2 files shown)


navidcy:
there are other places that this can change -- just wanna see if all tests are OK first

src/Grids/latitude_longitude_grid.jl
```diff
@@ -437,10 +437,17 @@ end
 ##### Kernels that precompute the z- and x-metric
 #####
 
-@inline metric_worksize(grid::LatitudeLongitudeGrid)  = (length(grid.Δλᶜᵃᵃ), length(grid.φᵃᶜᵃ) - 1) 
+@inline function metric_worksize(grid::LatitudeLongitudeGrid)
+    TX, TY, _ = topology(grid)
+    return (total_length(Face(), TX(), grid.Nx, grid.Hx) - 2, total_length(Face(), TY(), grid.Ny, grid.Hy) - 2)
```
glwagner, line 442:
```suggestion
    Tx = total_length(Face(), TX(), grid.Nx, grid.Hx)
    Ty = total_length(Face(), TY(), grid.Ny, grid.Hy)
    return (Tx - 2, Ty - 2)
```
And the same below

src/Grids/latitude_longitude_grid.jl
```diff
@@ -437,10 +437,17 @@ end
...
+end
 @inline metric_workgroup(grid::LatitudeLongitudeGrid) = (16, 16) 
 
-@inline metric_worksize(grid::XRegularLLG)  = length(grid.φᵃᶜᵃ) - 1 
+@inline function metric_worksize(grid::XRegularLLG)
+    _, TY, _ = topology(grid)
```
glwagner, line 447:
```suggestion
    TY = topology(grid, 2)
```

src/Grids/latitude_longitude_grid.jl
```diff
@@ -498,7 +505,9 @@ end
 
 function precompute_Δy_metrics(grid::LatitudeLongitudeGrid, Δyᶠᶜ, Δyᶜᶠ)
     arch = grid.architecture
-    precompute_Δy! = precompute_Δy_kernel!(Architectures.device(arch), 16, length(grid.Δφᵃᶜᵃ) - 1)
+    _, TY, _ = topology(grid)
```
glwagner, line 508:
```suggestion
    TY = topology(grid, 2)
```

navidcy:
I believe this PR is ready.

navidcy:
Interestingly this breaks the distributed tests!
https://buildkite.com/clima/oceananigans-distributed/builds/1218#018da64e-30be-4ecb-9b10-32cdb3daa89c

glwagner:
That's annoying. There must be some implicit assumptions that we can't see

glwagner:
@simone-silvestri can you help?

simone-silvestri:
this is the assumption
https://github.com/CliMA/Oceananigans.jl/blob/c55878eb6de62d8568bd162dfe4718d9bb772a67/src/DistributedComputations/distributed_grids.jl#L140-L160

glwagner:
I don't understand that text on first read...

I think it's just wrong that the meridional direction is always `Bounded`. We need `Periodic` too. How can we implement this?

simone-silvestri:
Hmmm, I am not sure how you solved this in the serial grid, but I guess in the same way. By removing that assumption and calculating the coordinate accordingly?

glwagner:
> Hmmm, I am not sure how you solved this in the serial grid, but I guess in the same way. By removing that assumption and calculating the coordinate accordingly?

Since we removed the assumption, do we also eliminate the need for this code which avoids the assumption?

simone-silvestri:
If the metrics are correctly calculated on the boundary for `FullyConnected`, `RightConnected`, and `LeftConnected` topologies yes. 

Note that in the latitude direction, the metrics for `FullyConnected` and `Periodic` are very different!

glwagner:
> If the metrics are correctly calculated on the boundary for `FullyConnected`, `RightConnected`, and `LeftConnected` topologies yes.
> 
> Note that in the latitude direction, the halo metrics for `FullyConnected` and `Periodic` are very different! In the sense that `FullyConnected` is like a `Bounded` in the halos (`Periodic` in latitude is a bit wonky)

`Periodic` in latitude is precisely defined...

simone-silvestri:
sure it is possible to define it, still it is conceptually wrong, you can easily see it if you visualize the grid:
```julia
julia> grid = LatitudeLongitudeGrid(size = (1, 2, 1), topology = (Periodic, Periodic, Bounded), latitude = (50, 52), longitude = (10, 11), z = (0, 1))
1×2×1 LatitudeLongitudeGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo and with precomputed metrics
├── longitude: Periodic λ ∈ [10.0, 11.0) regularly spaced with Δλ=1.0
├── latitude:  Periodic φ ∈ [50.0, 52.0) regularly spaced with Δφ=1.0
└── z:         Bounded  z ∈ [0.0, 1.0]   regularly spaced with Δz=1.0

julia> grid.Δxᶜᶠᵃ
8-element OffsetArray(::Vector{Float64}, -2:5) with eltype Float64 with indices -2:5:
     0.0
 74403.92868970236
 72950.43560309989
 71474.721107126
 69977.2347187117
 68458.43258671739
 66918.77735298542
     0.0
```
Note, the delta x in the halos. They are inconsistent with the periodic assumption. Even if it were, it is impossible to reconcile `Δxᶠᵃᵃ` at `Ny + 1` and `Δxᶠᵃᵃ` at 1 which should be exactly the same (but they end up being different)! This means that velocities and tracers that you use to "fill" the periodic halo will be exchanged between grid volumes of different size.
Another example, these volumes should be the same
```julia
julia> Vᶜᶜᶜ(1, 3, 1, grid)
7.526820531985937e9

julia> Vᶜᶜᶜ(1, 1, 1, grid)
7.864569567312662e9
```

glwagner:
> sure it is possible to define it, still it is conceptually wrong, you can easily see it if you visualize the grid:
> 
> ```julia
> julia> grid = LatitudeLongitudeGrid(size = (1, 2, 1), topology = (Periodic, Periodic, Bounded), latitude = (50, 52), longitude = (10, 11), z = (0, 1))
> 1×2×1 LatitudeLongitudeGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo and with precomputed metrics
> ├── longitude: Periodic λ ∈ [10.0, 11.0) regularly spaced with Δλ=1.0
> ├── latitude:  Periodic φ ∈ [50.0, 52.0) regularly spaced with Δφ=1.0
> └── z:         Bounded  z ∈ [0.0, 1.0]   regularly spaced with Δz=1.0
> 
> julia> grid.Δxᶜᶠᵃ
> 8-element OffsetArray(::Vector{Float64}, -2:5) with eltype Float64 with indices -2:5:
>      0.0
>  74403.92868970236
>  72950.43560309989
>  71474.721107126
>  69977.2347187117
>  68458.43258671739
>  66918.77735298542
>      0.0
> ```
> 
> Note, the delta x in the halos. They are inconsistent with the periodic assumption. Even if it were, it is impossible to reconcile `Δxᶠᵃᵃ` at `Ny + 1` and `Δxᶠᵃᵃ` at 1 which should be exactly the same (but they end up being different)! This means that velocities and tracers that you use to "fill" the periodic halo will be exchanged between grid volumes of different size. Another example, these volumes should be the same
> 
> ```julia
> julia> Vᶜᶜᶜ(1, 3, 1, grid)
> 7.526820531985937e9
> 
> julia> Vᶜᶜᶜ(1, 1, 1, grid)
> 7.864569567312662e9
> ```

A single column is fine.

simone-silvestri:
Shouldn't a single column use a `Flat` topologies? I think that would probably be more appropriate

simone-silvestri:
In a periodic single column, again, the problem would be that `Δxᶠᵃᵃ[1]` is not equal to `Δxᶠᵃᵃ[2]` which it should. 

A consequence is that, since there should be zero derivatives in the y direction, `u[1] != u[2]` because of the difference in areas.

glwagner:
> Shouldn't a single column use a `Flat` topologies? I think that would probably be more appropriate

Not when we want to interpolate data onto a particular location.

glwagner:
> In a periodic single column, again, the problem would be that `Δxᶠᵃᵃ[1]` is not equal to `Δxᶠᵃᵃ[2]` which it should.
> 
> A consequence is that, since there should be zero derivatives in the y direction, `v[1] != v[2]` because of the difference in areas.

The halo region metrics need to enforced to be periodic.

simone-silvestri:
Nono, these are not halo metrics, they are part of the actual grid, periodic assumes that `Δxᶠᵃᵃ[1] == Δxᶠᵃᵃ[2]` (for a single column) I am not sure how you reconcile the fact that they are not the same. In practice, if the grid is not symmetric about the equator, I think that a single column cell (with associated periodic halos) would look like this:
Screenshot 2024-02-27 at 5 21 29 PM
So I guess the only way to be consistent would be to straighten out that cells

glwagner:
Where in the code does this crop up?

glwagner:
Changes in both grid construction and also interpolation would be needed to support `Flat, Flat, Bounded` with non-trivial location:

```
julia> flat_grid = LatitudeLongitudeGrid(size=grid.Nz, longitude=(215, 215.25), latitude=(50, 50.25), z=(-400, 0), topology=(Flat, Flat, Bounded))
1×1×100 LatitudeLongitudeGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo and without precomputed metrics
├── longitude: Flat λ
├── latitude:  Flat φ
└── z:         Bounded  z ∈ [-400.0, 0.0]    regularly spaced with Δz=4.0

julia> flat_grid.λᶜᵃᵃ
StepRangeLen(1.0, 0.0, 1)

julia> non_flat_grid = LatitudeLongitudeGrid(size=(8, 8, grid.Nz), longitude=(214, 216), latitude=(49, 51), z=(-400, 0), topology=(Bounded, Bounded, Bounded))
8×8×100 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×3 halo and with precomputed metrics
├── longitude: Bounded  λ ∈ [214.0, 216.0] regularly spaced with Δλ=0.25
├── latitude:  Bounded  φ ∈ [49.0, 51.0]   regularly spaced with Δφ=0.25
└── z:         Bounded  z ∈ [-400.0, 0.0]  regularly spaced with Δz=4.0

julia> c_src = CenterField(non_flat_grid)
8×8×100 Field{Center, Center, Center} on LatitudeLongitudeGrid on CPU
├── grid: 8×8×100 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×3 halo and with precomputed metrics
│   └── west: ZeroFlux, east: ZeroFlux, south: ZeroFlux, north: ZeroFlux, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 14×14×106 OffsetArray(::Array{Float64, 3}, -2:11, -2:11, -2:103) with eltype Float64 with indices -2:11×-2:11×-2:103

julia> set!(c_src, (λ, φ, z) -> rand())
8×8×100 Field{Center, Center, Center} on LatitudeLongitudeGrid on CPU
├── grid: 8×8×100 LatitudeLongitudeGrid{Float64, Bounded, Bounded, Bounded} on CPU with 3×3×3 halo and with precomputed metrics
│   └── west: ZeroFlux, east: ZeroFlux, south: ZeroFlux, north: ZeroFlux, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 14×14×106 OffsetArray(::Array{Float64, 3}, -2:11, -2:11, -2:103) with eltype Float64 with indices -2:11×-2:11×-2:103
    └── max=0.999998, min=1.2255e-5, mean=0.497663

julia> c_dst = CenterField(flat_grid)
1×1×100 Field{Center, Center, Center} on LatitudeLongitudeGrid on CPU
├── grid: 1×1×100 LatitudeLongitudeGrid{Float64, Flat, Flat, Bounded} on CPU with 0×0×3 halo and without precomputed metrics
│   └── west: Nothing, east: Nothing, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 1×1×106 OffsetArray(::Array{Float64, 3}, 1:1, 1:1, -2:103) with eltype Float64 with indices 1:1×1:1×-2:103

julia> Oceananigans.Fields.interpolate!(c_dst, c_src)
ERROR: BoundsError: attempt to access Tuple{Float64} at index [2]
Stacktrace:
  [1] indexed_iterate
    @ ./tuple.jl:92 [inlined]
  [2] _fractional_indices
    @ ~/Projects/Oceananigans.jl/src/Fields/interpolate.jl:163 [inlined]
  [3] fractional_indices
    @ ~/Projects/Oceananigans.jl/src/Fields/interpolate.jl:152 [inlined]
  [4] interpolate
    @ ~/Projects/Oceananigans.jl/src/Fields/interpolate.jl:222 [inlined]
  [5] macro expansion
    @ ~/Projects/Oceananigans.jl/src/Fields/interpolate.jl:302 [inlined]
```

glwagner:
Another possibility is to build two grids, then use one for interpolation and then manually copy the data onto a new field generated on the `Flat` grid. But this is complicated. 

Another question is whether this should work for grids that are symmetric about the equator. It seems like it should.

glwagner:
I think we can close this PR and #3465 because the fix was already merged in #3450 (sorry for putting that in silently, I was just trying to get the code working). As @simone-silvestri points out, simulations that are Periodic-in-latitude may be physically incorrect. However, the code will run and does not necessary blow up immediately, for example:

https://github.com/CliMA/Oceananigans.jl/assets/15271942/96a91c80-87ad-49b2-89f9-f8107124bed0

I think we could consider adding a warning for this case in a future PR, but we can discuss that elsewhere.

glwagner:
An inertial oscillation solution at a midlatitude location also illustrates potential concerns about physical validity:

image

generated by this code:

```julia
using Oceananigans
using Oceananigans.Units
using GLMakie

φ = 50
δ = 0.01

grid = LatitudeLongitudeGrid(size = (1, 1, 1),
                             longitude = (-δ, δ),
                             latitude = (φ-δ, φ+δ),
                             z = (-1, 0),

model = HydrostaticFreeSurfaceModel(; grid,
                                    tracers = (),
                                    buoyancy = nothing,
                                    coriolis = HydrostaticSphericalCoriolis(),
                                    tracer_advection = nothing,
                                    momentum_advection = nothing)

@show model

set!(model, u=1)

simulation = Simulation(model, Δt=20minutes, stop_iteration=200)

ut = Float64[]
vt = Float64[]
t = Float64[]

function save_uvt(sim)
    push!(ut, first(sim.model.velocities.u))
    push!(vt, first(sim.model.velocities.v))
    push!(t, time(sim))
    return nothing
end

add_callback!(simulation, save_uvt)

run!(simulation)

Ω = model.coriolis.rotation_rate
f₀ = 2Ω * sind(φ)

fig = Figure()
ax = Axis(fig[1, 1])
lines!(ax, t, ut, linestyle=:dash, label="u")
lines!(ax, t, vt, linestyle=:dash, label="v")
lines!(ax, t, cos.(f₀ * t), linewidth=4, label="cos(ft)")
axislegend(ax)
```

though a bit more thinking about the math would be needed for a complete interpretation. Nevertheless, it's not an f-plane.

navidcy:
Since https://github.com/CliMA/Oceananigans.jl/issues/3465 was closed, do we still need this PR open?

glwagner:
no we don't

PR #3469, opened 2024-02-09
More illustrative example for checkpointing

navidcy:


docs/src/model_setup/checkpointing.md
```diff
@@ -1,51 +1,61 @@
...
 time to files of the form `model_checkpoint_iteration12500.jld2` where `12500` is the iteration
-number (automatically filled in)
+number (automatically filled in).
 
-```@meta
-DocTestSetup = quote
-    using Oceananigans
-end
-```
+Here's an example where we checkpoint every 5 iterations. This is _way_ too often, we only do it here for
```
glwagner, line 12:
```suggestion
Here's an example where we checkpoint every 5 iterations. This is far more often that one wants for a typical application: we only do it here for
```

glwagner (approved):

docs/src/model_setup/checkpointing.md
glwagner: probably good to more specific here

PR #3471, opened 2024-02-12
Replace type unstable tuple call with type-stable ntuple call

jlk9:
Just replaced the tuple line in `permute_boundary_conditions` (in file `src/BoundaryConditionsfill_halo_regions.jl`) with an ntuple call that makes it type-stable and eases use of Enzyme.


src/BoundaryConditions/fill_halo_regions.jl
```diff
@@ -115,9 +115,12 @@ function permute_boundary_conditions(boundary_conditions)
...
 
-    boundary_conditions = Tuple(extract_bc(boundary_conditions, Val(side)) for side in sides)
-
-    return fill_halos!, boundary_conditions
+    boundary_conditions2 = ntuple(Val(length(sides))) do i
+        Base.@_inline_meta
+        extract_bc(boundary_conditions, Val(sides[i]))
+    end
+    
+    return fill_halos!, boundary_conditions2
```
glwagner, line 123:
```suggestion
    boundary_conditions_tuple = ntuple(Val(length(sides))) do n
        Base.@_inline_meta
        extract_bc(boundary_conditions, Val(sides[n]))
    end
    
    return fill_halos!, boundary_conditions_tuple
```

just a little clarity. Usually `i` is the `x`-index.

PR #3484, opened 2024-02-27
(0.90.9) Fix a sign error in CATKE's implicit dissipative buoyancy flux

glwagner:
This PR fixes a sign error in the implicit dissipative buoyancy flux. We're also recalibrating parameters and will update them here before merging the PR.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.90.8"
+version = "0.90.9"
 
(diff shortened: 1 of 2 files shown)


navidcy:
OK, I approve (and can also formally approve) but we also need to update the CATKE default parameters, correct?

navidcy (approved):
A new era for CATKE

glwagner:
It's doing reasonable things

https://github.com/CliMA/Oceananigans.jl/assets/15271942/2cd1311d-3130-4e51-8c1c-7a9eb00347ed

glwagner:
> OK, I approve (and can also formally approve) but we also need to update the CATKE default parameters, correct?

Yes, and as you probably noticed, done.

Issue #3051, opened 2023-04-05
Bug in `ShallowWaterModel` treatment of bathymetric contribution to pressure


glwagner:
These lines multiply `h` (a field at ccc) with objects at `fcc` and `cfc` respectively

https://github.com/CliMA/Oceananigans.jl/blob/bcc34f07b3f949ea6fb34c7814f4b856d24924c2/src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl#L16-L17

This doesn't look correct, but if there is some logic that makes it correct, it should be documented with a comment at least. I think if the bathymetric height is defined at cell centers, then the bathymetric height at a cell interface might need to be defined as the maximum of the height of the adjacent cells.

navidcy:
btw, these lines could use an `@inbounds` :)

glwagner:
Another issue.

simone-silvestri:
This definitely seems wrong to me, the bathymetric height is defined at cells centers, so we should interpolate `h` to velocity points

Issue #3394, opened 2023-11-17
Fix up `ShallowWaterModel.jl`


francispoulin:
This issue will help to combine several outstanding issues involving the `ShallowWaterModel.jl`.  I suggest the following in this order:

- [ ] Fix inconsistent advection from the two models (@2928)
- [ ] Fix up shallow water regression tests (@3049)
- [ ] Speed up shallow water example when building docs (@3151)
- [ ] Make shallow water Bickley jet less expensive (@3169)
- [ ] Update immersed bondaries validation scripts (@2985)
- [ ] Introduce multi-layer shallow water model (@2507)
- [ ] Validate positive preserving WENO schemes

For the first part, I put together this document that shows and derives the equations for the two different models.  Next I will look at the tendencies.  

Any suggestions on what else to do, or what to do differently, are welcome!

[ShallowWaterModel_eqns.pdf](https://github.com/CliMA/Oceananigans.jl/files/13392634/ShallowWaterModel_eqns.pdf)

francispoulin:
@simone-silvestri and I created a new branch to try and start tackle these issues, `fp-ss/shallow-water-version2`.

When ran the `shallow_water_Bickley_jet.jl` on this branch there was an error.  We were using superscripts `fcc` and `cfc` instead if `faa` and `cfc` on the averaging operators.  I have since fixed that.  

Now I'm going to try and merge main into this branch.

francispoulin:
@simone-silvestri, I'm trying to merge `main` into `fp-ss/shallow-water-version2` and having some problems with `src/Grids/latitude_longitude-grid.jl`.

The version you put together has this in line 189 https://github.com/CliMA/Oceananigans.jl/blob/6bd880053dc865c58a1667cfd2649624cb3e8aa8/src/Grids/latitude_longitude_grid.jl#L189

Main however has a very different syntax on the left hand side of equals.  

https://github.com/CliMA/Oceananigans.jl/blob/0391b3a90d0dddd67625cd4373e80012e9806df6/src/Grids/latitude_longitude_grid.jl#L189

I remember that you worked on the sizes when we chatted.  Any suggestions as to how we should resolve the conflicts?

francispoulin:
@simone-silvestri , any advice on how to resolve this merger with main?

glwagner:
@francispoulin `Nλ, Nφ, Nz = size` and `Hλ, Hφ, Hz = halo`

francispoulin:
[...]

PR #2985, opened 2023-03-18
Update SWE immersed boundaries validation scripts

navidcy:
@simone-silvestri I tried to update this script but still it won't run. Could you have a look?

PR #3486, opened 2024-02-29
Fixing tendencies in shallow water model and bringing back regression tests

francispoulin:
Closes https://github.com/CliMA/Oceananigans.jl/issues/3051

Following up on #3394, this PR tries to do the first two tasks, fix the tendencies in the shallow water model and bring back the regression tests.


validation/shallow_water_model/near_global_shallow_water_quarter_degree.jl
navidcy, line 1: @francispoulin did you run this? it had an ACC going the wrong way previously... I think because the bathymetry was inverted or something... I tried to sort this out in https://github.com/CliMA/Oceananigans.jl/pull/2985 but I lost steam at some point...

francispoulin: I have not run it yet but thanks for the reminder.  I will run it in the next do or so.  I remember it was a big simulation so probably best to do it on a server and not my laptop.  Thanks for the reminder!

francispoulin:
I'm trying to run the near global shallow water model and having difficulties with the latitude longitude grid.  The line that calls it is here

https://github.com/CliMA/Oceananigans.jl/blob/5ba82e5786d1b7725d6bfa3e45072d2f6fad487a/validation/shallow_water_model/near_global_shallow_water_quarter_degree.jl#L95

And the error that I get is the following,

```
julia> include("near_global_shallow_water_quarter_degree.jl")
┌ Warning: Over-writing registration of the datadep
│   name = "quarter_degree_near_global_lat_lon"
└ @ DataDeps ~/.julia/packages/DataDeps/Y2lje/src/registration.jl:15
ERROR: LoadError: UndefVarError: `Nλ` not defined
Stacktrace:
 [1] validate_lat_lon_grid_args(topology::Tuple{…}, size::Tuple{…}, halo::Tuple{…}, FT::Type, latitude::Tuple{…}, longitude::Tuple{…}, z::Nothing, precompute_metrics::Bool)
   @ Oceananigans.Grids ~/Software/Oceananigans.jl/src/Grids/latitude_longitude_grid.jl:257
 [2] LatitudeLongitudeGrid(architecture::GPU, FT::DataType; size::Tuple{…}, longitude::Tuple{…}, latitude::Tuple{…}, z::Nothing, radius::Float64, topology::Tuple{…}, precompute_metrics::Bool, halo::Tuple{…})
   @ Oceananigans.Grids ~/Software/Oceananigans.jl/src/Grids/latitude_longitude_grid.jl:189
 [3] macro expansion
   @ show.jl:1181 [inlined]
 [4] top-level scope
   @ ~/Software/Oceananigans.jl/validation/shallow_water_model/near_global_shallow_water_quarter_degree.jl:95
 [5] include(fname::String)
   @ Base.MainInclude ./client.jl:489
 [6] top-level scope
   @ REPL[70]:1
 [7] top-level scope
   @ ~/.julia/packages/CUDA/nbRJk/src/initialization.jl:205
in expression starting at /home/fpoulin/Software/Oceananigans.jl/validation/shallow_water_model/near_global_shallow_water_quarter_degree.jl:95
```

Should the lat-lon grid be able to handle flat in the vertical?  @simone-silvestri, maybe this is something you tried to fix the last time we chatted?

navidcy:
on `#main`:

```Julia
julia> grid = LatitudeLongitudeGrid(size = (10, 10), latitude = (-60, 60), longitude = (-180, 180), topology = (Periodic, Bounded, Flat))
10×10×1 LatitudeLongitudeGrid{Float64, Periodic, Bounded, Flat} on CPU with 3×3×0 halo and with precomputed metrics
├── longitude: Periodic λ ∈ [-180.0, 180.0)  regularly spaced with Δλ=36.0
├── latitude:  Bounded  φ ∈ [-60.0, 60.0]    regularly spaced with Δφ=12.0
└── z:         Flat z
```

src/Grids/latitude_longitude_grid.jl
```diff
@@ -239,11 +239,10 @@ function with_precomputed_metrics(grid)
                                              Δxᶠᶜᵃ, Δxᶜᶠᵃ, Δxᶠᶠᵃ, Δxᶜᶜᵃ, Δyᶠᶜᵃ, Δyᶜᶠᵃ,
                                              Azᶠᶜᵃ, Azᶜᶠᵃ, Azᶠᶠᵃ, Azᶜᶜᵃ, grid.radius)
 end
-
+    
 function validate_lat_lon_grid_args(topology, size, halo, FT, latitude, longitude, z, precompute_metrics)
     if !isnothing(topology)
-        TX, TY, TZ = validate_topology(topology)
-        Nλ, Nφ, Nz = size = validate_size(TX, TY, TZ, size)
+        TX, TY, TZ = topology
```
navidcy, line 245: this change here, why? this is what causes your trouble with Flat z coord @francispoulin

francispoulin: Agreed!  @simone-silvestri made this change and I don't pretend to understand the details.  

Have worked on this for a while now and it seems clear this is a tricky problem.  I remember there were some recent discussions about lat-lon grid and wanting to allow for Flat.  Shallow water only needs flat in z but if we want to allow for flat in x and y as well, which seems reasoanble, then some care should be taken to do this well.

Maybe this part should be part of another PR?

simone-silvestri: I am not sure about this change here. On main everything seems to work correctly 

```julia
julia> grid = LatitudeLongitudeGrid(size = (100, 10), latitude = (-10, 10), longitude = (-10, 10), topology = (Periodic, Bounded, Flat))
100×10×1 LatitudeLongitudeGrid{Float64, Periodic, Bounded, Flat} on CPU with 3×3×0 halo and with precomputed metrics
├── longitude: Periodic λ ∈ [-10.0, 10.0)    regularly spaced with Δλ=0.2
├── latitude:  Bounded  φ ∈ [-10.0, 10.0]    regularly spaced with Δφ=2.0
└── z:         Flat z

julia> model = ShallowWaterModel(; grid, gravitational_acceleration=9.0, formulation=VectorInvariantFormulation(), momentum_advection=VectorInvariant())
┌ Warning: The ShallowWaterModel is currently unvalidated, subject to change, and should not be used for scientific research without adequate validation.
└ @ Oceananigans.Models.ShallowWaterModels ~/development/Oceananigans.jl/src/Models/ShallowWaterModels/shallow_water_model.jl:129
ShallowWaterModel{typename(CPU), Float64}(time = 0 seconds, iteration = 0)
├── grid: 100×10×1 LatitudeLongitudeGrid{Float64, Periodic, Bounded, Flat} on CPU with 3×3×0 halo and with precomputed metrics
├── tracers: ()
└── coriolis: Nothing
```
we can probably revert these changes

francispoulin: I should revert `latitude_longitude_grid.jl` to what we have on `main`?

I can give that a try and see if I can get this going but it seems that the lat-lon grid is a bit fussy when you declare a direction to be `Flat`.

francispoulin: I reverted to the old file and that seemed to work.  Thanks!  Now will work on the validation example and see if that runs.

francispoulin:
I am able to run the validatoin example.  Unfortunately, it produces `NaNs`  after just over an hour.  Sigh.

Even if I reduce the time step from 20 seconds to 1 second, it still goes unstable at about the same time.  I need to look more closely as to why this might be happening.  I would share an animation but one hour doesn't seem to be enough time for much to happen so I don't see any changes.

simone-silvestri:
There must be a simple bug to fix. I remember the simulation was quite stable. A simple configuration might snoop out the issue

simone-silvestri:
I think in this PR we changed the definition of bathymetry to how it was defined before. Also, the initial height cannot be zero (because of wind stress forcing begin $\tau / h$). On main, it looks like it runs stably (correcting the momentum advection scheme that changed in the meantime)

francispoulin:
I have created a new validation script called `shallow_jet_topography.jl` that evolves the barotropically unstable jet (like in the shallow water example) but now over a shelflike topography.  It can be found in `valiation/shallow_water`.  

Some preliminary results are shown below.

First, this is the case with a flat bottom and it does produce the same results as what we have in the example.

Geostrophic_State

Second, this is an example of prograde topography with the shallow water on the right of the propogating jet.

Geostrophic_State

Third, this is an example of retrograde topopgraphy with the shallow water on the right of the propogating jet.

Geostrophic_State

The growth rates differ but I want to update my linear stability code to compute the growth rates so that we can validate that the results are quantitatively close.

francispoulin:
This is a sample plot of the growth rate in the case of a flat bottom, and it agrees with what I expecte to see.  Also the growth rate is 0.139, which matches with the theory to at least three significant digits.  

growth_rate

This is what I get with some topography, it doesn't matter which direction, they are both very similar to this.  The large oscillations at the beginning are a cluse that something is wrong.  There is a rapid initial growth that shouldn't be there either.

growth_rate

This is using the conservative form.  I'll try the vector invariant form and see if this happens in both cases or not.

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
```diff
@@ -4,21 +4,15 @@ using Oceananigans.Coriolis
...
 using Oceananigans.TurbulenceClosures: ∇_dot_qᶜ, ∂ⱼ_τ₁ⱼ, ∂ⱼ_τ₂ⱼ
 
-@inline half_g_h²(i, j, k, grid, h, g)  = @inbounds 1/2 * g * h[i, j, k]^2
+# bathymetry (hB) is assumed to be a negative value equal to - depth.
 @inline h_plus_hB(i, j, k, grid, h, hB) = @inbounds h[i, j, k] + hB[i, j, k]
 
-@inline x_pressure_gradient(i, j, k, grid, g, h, hB, formulation) = ∂xᶠᶜᶜ(i, j, k, grid, half_g_h², h, g)
-@inline y_pressure_gradient(i, j, k, grid, g, h, hB, formulation) = ∂yᶜᶠᶜ(i, j, k, grid, half_g_h², h, g)
+@inline x_pressure_gradient(i, j, k, grid, g, h, hB, ::ConservativeFormulation) = g * ℑxᶠᵃᵃ(i, j, k, grid, h) * ∂xᶠᶜᶜ(i, j, k, grid, h_plus_hB, h, hB)
+@inline y_pressure_gradient(i, j, k, grid, g, h, hB, ::ConservativeFormulation) = g * ℑyᵃᶠᵃ(i, j, k, grid, h) * ∂yᶜᶠᶜ(i, j, k, grid, h_plus_hB, h, hB)
```
navidcy, line 11: these lines deal with #3051, right?

validation/shallow_water_model/near_global_shallow_water_quarter_degree.jl
```diff
@@ -182,7 +181,7 @@ model = ShallowWaterModel(grid = grid,
 ##### Initial condition:
 #####
 
-h_init = deepcopy(1e1 .+ maximum(bat) .- bat) 
+h_init = - deepcopy(bat) .+ 10 # we should add a positive value to avoid a vanishing layer
```
navidcy, line 184: Do we need to initialize `h` with `-bathymetry`? (btw the variable names here could be improved...)

navidcy:
> I have created a new validation script called `shallow_jet_topography.jl` that evolves the barotropically unstable jet (like in the shallow water example) but now over a shelflike topography. It can be found in `valiation/shallow_water`.

I don't see it; did you push?

francispoulin:
> > I have created a new validation script called `shallow_jet_topography.jl` that evolves the barotropically unstable jet (like in the shallow water example) but now over a shelflike topography. It can be found in `valiation/shallow_water`.
> 
> I don't see it; did you push?

Sorry!  I have added it now.

https://github.com/CliMA/Oceananigans.jl/blob/fp-ss/shallow-water-version2/validation/shallow_water_model/shallow_water_jet_topography.jl

One of my mitakes is that I didn't define the bathymetry. I need to figure out how to do that.  Thanks for the very helpful feedback!

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
francispoulin: Yes

validation/shallow_water_model/near_global_shallow_water_quarter_degree.jl
francispoulin: Good question and observation.  

I defined `h` to be the total depth, which seems to be the correct thing to do.

I am happy to change the names after we get something working.

navidcy: I meant really that `bat` is that mammal that flies or the stick we play baseball.

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
navidcy: Can we link this PR with that issue then. Edit the first comment pf the PR and write “Closes #3051”

simone-silvestri:
you need to pass it to the model as a keyword argument. In this case with
`bathymetry = b`

src/Models/ShallowWaterModels/solution_and_tracer_tendencies.jl
francispoulin: Thanks for the suggestion and done!

francispoulin:
> you need to pass it to the model as a keyword argument. In this case with `bathymetry = b`

Nice!  I will give it a try today.

francispoulin:
This is the case of prograde topography, which has a growth rate that looks reasoanble and is 0.147 > 0.139 as we have in the flat bottom case.  This is consistent with theory.
growth_rate

This is the case of retrograde topography, which has a growth rate that looks reasoanble and is 0.115 < 0.139 as we have in the flat bottom case.  This is consistent with theory.

growth_rate

This is all very promising and I will confirm we get the same values with the vector invariant form.

francispoulin:
I am happy to say that the `VectorInvariant` formulation yields similar growth rates `0.144` and `0.123`.   Given the numerical methods are different I'm not bothered by this difference. They should converge with increased resolution.

Any more questions/concerns on this simple validation problem before I return to the global one?

navidcy:
I have a bit of trouble and confusion regarding conventions for setting up the bathymetry. I'll put a few comments on the script and hopefully that will showcase where my confusion is.

validation/shallow_water_model/shallow_water_jet_topography.jl
```diff
@@ -0,0 +1,167 @@
...
+
+grid = RectilinearGrid(size = ( Nx, Ny),
+                          x = (  0, Lx),
+                          y = (-Ly, Ly),
+                   topology = (Periodic, Bounded, Flat))
+
+η(x, y) = Δη * tanh(y)
+b(x, y) = Hb * tanh(y/Lb)
+
+h̄(x, y) = H₀ + η(x, y) + b(x, y)
```
navidcy, line 21: so this is the total fluid depth, right?

validation/shallow_water_model/shallow_water_jet_topography.jl
```diff
@@ -0,0 +1,167 @@
...
+
+ū̄h(x, y) = ū(x, y) * h̄(x, y)
+
+bathymetry(x,y) = -b(x,y)
+
+model = ShallowWaterModel(; grid, coriolis = FPlane(f=f), 
+                gravitational_acceleration = g,
+                               timestepper = :RungeKutta3,
+                        momentum_advection = WENO(),
+                                bathymetry = bathymetry
```
navidcy, line 38: what is expected to provide here? here we give `-b`, but what should we give? Why `-b`? is this a convention hidden somewhere? Why not `b` or `-H0+b` or whatever else?

validation/shallow_water_model/shallow_water_jet_topography.jl
```diff
@@ -0,0 +1,167 @@
...
+#ωb(x, y) = - 2 * U * sech(y)^2 * tanh(y)
+
+small_amplitude = 1e-4
+ 
+ uⁱ(x, y) = ū(x, y) + small_amplitude * exp(-y^2) * randn()
+uhⁱ(x, y) = uⁱ(x, y) * h̄(x, y)
+
+ū̄h(x, y) = ū(x, y) * h̄(x, y)
+
+bathymetry(x,y) = -b(x,y)
```
navidcy, line 32: why we define something that is `b(x, y)` and then `bathymetry = - b`? what is `b` then? let's define the bathymetry from the beginning?

validation/shallow_water_model/shallow_water_jet_topography.jl
navidcy: I think I'd like to see a sketch  with the free surface and the bathymetry so that I'm clear on what each of these are...

all this confusion of mine started from #2985; see discussion there and, for example, comment https://github.com/CliMA/Oceananigans.jl/pull/2985#issuecomment-1476028985

validation/shallow_water_model/shallow_water_jet_topography.jl
francispoulin: Yes.  In order to get a reasonable result I needed to set `bathymetry = - b`.  This was not my first guess and I want to make sure I understand why this is the correct choice.  The PDF that I posted with the equations does not follow this convention, so I need to fix that up, after I understand this better.

validation/shallow_water_model/shallow_water_jet_topography.jl
francispoulin: This is not direct and should be fixed.  Agreed!

validation/shallow_water_model/shallow_water_jet_topography.jl
francispoulin: @simone-silvestri , since both @navidcy and I are confused about the convention used here, might you be able to shed some light on the subject?

validation/shallow_water_model/shallow_water_jet_topography.jl
francispoulin: Yes, very good idea as clearly this is a point of confusion.

navidcy: I guess the docs are clear:

> The elevation of the bottom bathymetry, measured with respect to the free-surface at rest.

francispoulin: Agreed!  Now to understand why this doesn't work like we expect.  Or, at least how i expect.

validation/shallow_water_model/shallow_water_jet_topography.jl
glwagner: I think the name is confusing because it's vague. Is this the bottom height, where the resting surface displacement is 0? Or is this the thickness of the fluid at rest?

francispoulin: I thought it was the bottom height but I could be wrong.

glwagner: I think something like `depth` or `resting_depth` would be less ambiguous maybe

glwagner: Or `resting_layer_height` which is the same thing

validation/shallow_water_model/shallow_water_jet_topography.jl
navidcy: Well there is a good chance the docs say one thing but the source code does something else!

francispoulin:
Thanks everyone for the feedback.  I did some debugging and I'm happy to say that bathymetry is now specified as we expected, no negative sign, and it gives results that are consistent with previous findings: prograde is more unstable and retrograde is more stable. 

You can see the code [here](https://github.com/CliMA/Oceananigans.jl/blob/fp-ss/shallow-water-version2/validation/shallow_water_model/shallow_water_jet_topography.jl). 

Any other thoughts on this before I move on to the global model?

validation/shallow_water_model/shallow_water_jet_topography.jl
```diff
@@ -0,0 +1,164 @@
...
+
+u = uh / h
+v = vh / h
+
+# Plot the Geostrophic State 
+
+using CairoMakie
+
+x = grid.xᶜᵃᵃ[1:Nx]
+y = grid.yᵃᶜᵃ[1:Ny]
```
navidcy, line 50: lets use nodes here?

```suggestion
x = xnodes(grid, Center())
y = ynodes(grid, Center())
```

validation/shallow_water_model/shallow_water_jet_topography.jl
```diff
@@ -0,0 +1,164 @@
...
+
+fig = Figure()
+Axis(fig[1,1], xlabel = "y", ylabel = "depth", title = "Geostrophic State")
+lines!(y,  h[ 1,1:Ny,1] .+ b.(0,y), linewidth=4, linestyle = :solid, color=:blue, label="η(y)")
+lines!(y,  0*y, linewidth=2, linestyle = :dash, color=:blue, label="0")
+lines!(y,  b.(0,y), linewidth=4, color=:black, label="b(y)")
+lines!(y, -H₀ .+ 0*y, linewidth=2, linestyle = :dash, color=:black, label="H₀")
```
navidcy, line 57:
```suggestion
Axis(fig[1, 1], xlabel = "y", ylabel = "depth", title = "Geostrophic State")
lines!(y, h[1, 1:Ny, 1] .+ b.(0, y), linewidth=4, linestyle=:solid, color=:blue, label="η(y)")
lines!(y, 0*y, linewidth=2, linestyle=:dash, color=:blue, label="0")
lines!(y, b.(0, y), linewidth=4, color=:black, label="b(y)")
lines!(y, -H₀ .+ 0*y, linewidth=2, linestyle = :dash, color=:black, label="H₀")
```

validation/shallow_water_model/shallow_water_jet_topography.jl
```diff
@@ -0,0 +1,164 @@
...
+f, g, U  = 1, 1, 1
+Δη = -f * U / g 
+
+grid = RectilinearGrid(size = ( Nx, Ny),
+                          x = (  0, Lx),
+                          y = (-Ly, Ly),
+                   topology = (Periodic, Bounded, Flat))
+
+η(x, y) =   Δη * tanh(y)
+b(x, y) = - H₀ + Hb * tanh(y/Lb)  # FJP: sign in front of Hb?
```
navidcy, line 19:
```suggestion
bathymetry(x, y) = - H₀ + Hb * tanh(y/Lb)  # FJP: sign in front of Hb?
```

validation/shallow_water_model/shallow_water_jet_topography.jl
```diff
@@ -0,0 +1,164 @@
...
+h̄(x, y) = η(x, y) - b(x, y)
+ū(x, y) =  U * sech(y)^2
+ 
+ uⁱ(x, y) = ū(x, y) + 1e-4 * exp(-y^2) * randn()
+uhⁱ(x, y) = uⁱ(x, y) * h̄(x, y)
+
+ū̄h(x, y) = ū(x, y) * h̄(x, y)
+
+bathymetry(x,y) = b(x,y)
+
```
navidcy, line 30:
```suggestion
```

validation/shallow_water_model/shallow_water_jet_topography.jl
```diff
@@ -0,0 +1,164 @@
...
+model = ShallowWaterModel(; grid, coriolis = FPlane(f=f), 
+                gravitational_acceleration = g,
+                               timestepper = :RungeKutta3,
+                        momentum_advection = WENO(),
+                                bathymetry = bathymetry
```
navidcy, line 35:
```suggestion
                                bathymetry = b
```

francispoulin:
> So in the end bathymetry is defined as a negative distance from the surface?

Yes, exactly that.

I still think that we should change `h_plus_hB` in the tendencies to `free_surface` or something, as it is exactly that and more descriptive.

simone-silvestri:
> > So in the end bathymetry is defined as a negative distance from the surface?
> 
> Yes, exactly that.
> 
> I still think that we should change `h_plus_hB` in the tendencies to `free_surface` or something, as it is exactly that and more descriptive.

Sure!

validation/shallow_water_model/shallow_water_jet_topography.jl
```diff
@@ -0,0 +1,164 @@
...
+H₀     = 10
+α, β   = 0.1, -0.1
+Lb, Hb = α * Ly, β * H₀
+
+f, g, U  = 1, 1, 1
+Δη = -f * U / g 
+
+grid = RectilinearGrid(size = ( Nx, Ny),
+                          x = (  0, Lx),
+                          y = (-Ly, Ly),
```
navidcy, line 15: it's a bit counter-intuitive that we define `Ly = 10` and then `grid.Ly = 20`

navidcy: can we change to

```
Ly = 20
...
grid = RectilinearGrid(..., y = (-Ly/2, Ly/2), ...)
```

francispoulin: Yes, a much better idea!

francispoulin: Done!

francispoulin:
Lots of the checks have failed and I don't understand why.  Anyone have an idea as to what's going wrong?

francispoulin:
Thanks @glwagner for merging main into this branch.

I have returned to the global simulations and have learned something.

Using all the same parameters as before, I get `NaNs` after just over an hour.

Upon closer inspection, I see that the height becomes negative after 56 minutes or so.  I have tried reducing the time step by a factor of 20 and that didn't help.  

Do you think I should play around with other advection schemes, say some high order upwinding schemes?

navidcy:
> Thanks @glwagner for merging main into this branch.
> 
> I have returned to the global simulations and have learned something.
> 
> Using all the same parameters as before, I get `NaNs` after just over an hour.
> 
> Upon closer inspection, I see that the height becomes negative after 56 minutes or so. I have tried reducing the time step by a factor of 20 and that didn't help.
> 
> Do you think I should play around with other advection schemes, say some high order upwinding schemes?

Seems to me (but I don't know for sure) that the problem is more basic than the advection scheme...
You sure that the initial condition is consistent with bathymetry?

test/test_shallow_water_models.jl
```diff
@@ -3,6 +3,8 @@ include("dependencies_for_runtests.jl")
 using Oceananigans.Models.ShallowWaterModels
 using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBoundary
 
+CUDA.allowscalar(true)
+
```
navidcy, line 7: why this?

test/dependencies_for_runtests.jl
```diff
@@ -75,7 +75,8 @@ float_types = (Float32, Float64)
 
 include("utils_for_runtests.jl")
 
-archs = test_architectures()
+#archs = test_architectures()
+archs = [CPU(), GPU()]
```
navidcy, line 79: Test fail because of this! You hardcoded that `archs` is always both CPU and GPU so then the CI tries to test GPU-related methods on CPU clusters. Can we undo this?

simone-silvestri:
> Thanks @glwagner for merging main into this branch.
> 
> I have returned to the global simulations and have learned something.
> 
> Using all the same parameters as before, I get `NaNs` after just over an hour.
> 
> Upon closer inspection, I see that the height becomes negative after 56 minutes or so. I have tried reducing the time step by a factor of 20 and that didn't help.
> 
> Do you think I should play around with other advection schemes, say some high order upwinding schemes?

You need to make sure that `h` is never zero, that will give you a `NaN` because the forcing is divided by `h`. It is kind of annoying but you have to set a "minimum height" in the initial conditions (cannot start with `h = 0` somewhere). I remember setting `hmin = 10` was enough to get a stable simulation

test/test_shallow_water_models.jl
francispoulin: To run it on a GPU I needed to add this to work.  But I haven't tested it and this runs on a CPU so I"m happy to delete it if you prefer.  Probably better to remove it.

test/dependencies_for_runtests.jl
francispoulin: Ah!  Great to know and thanks for pointing this out.

I well retry this without the allowscalar on my computer and see what happens.  Thanks!

francispoulin: I undid my silly change.

test/test_shallow_water_models.jl
francispoulin: I removed this line.

francispoulin:
> > Thanks @glwagner for merging main into this branch.
> > I have returned to the global simulations and have learned something.
> > Using all the same parameters as before, I get `NaNs` after just over an hour.
> > Upon closer inspection, I see that the height becomes negative after 56 minutes or so. I have tried reducing the time step by a factor of 20 and that didn't help.
> > Do you think I should play around with other advection schemes, say some high order upwinding schemes?
> 
> You need to make sure that `h` is never zero, that will give you a `NaN` because the forcing is divided by `h`. It is kind of annoying but you have to set a "minimum height" in the initial conditions (cannot start with `h = 0` somewhere). I remember setting `hmin = 10` was enough to get a stable simulation

Thanks @simone-silvestri.  

The validation script was set up with `hmin = 10` and that did go unstable.  I tried `20` and `30` and they both go unstable.  I'm tring `40` and it seems better.  I thought if this at least runs, and shows us the ACC going on the right direction, thsi would be a big leap forward.  Then we can worry abou thow to minimize the minimum h, if we like.  

I will keep you posted.

simone-silvestri:
Does the script crash on main?

navidcy:
Smells like there might be something else happening also here @francispoulin...

simone-silvestri:
I would expect the simulation on main to be unstable when using `VectorInvariant()` and no dissipation method. I can take a look today at the differences between main and this branch

francispoulin:
Even  though it shouldn't run it does run for at least a day.  I will see later how far it got.

What's strange is that the changes that we made were mostly for the conservative scheme.  We only had a cosmetic change for the vector invariant form.  The fact that the vector invariant scheme is doing anything different is troubling.  I'll also take a look this afternoon when I have time.

francispoulin:
This branch had some weird issues that I didn't understand so I decided to copy the relevant files to a new branch, called `fp-ss-shallow-water-version3`.  I'm happy to say that one is behaving much better.  Hence, I will close this and create a draft PR for the new branch.

PR #3496, opened 2024-03-05
Default partition in `Distributed`

simone-silvestri:
on main `arch = Distributed(CPU())` would not work without having an  beforehand`MPI.Init()`. This PR corrects it be doing `MPI.Init()` before specifying a default partition


glwagner:
One step closer

glwagner:
What's left to solve on this PR?

simone-silvestri:
should be finished. We need to clean up Sverdrup, seems to be full

navidcy (approved):

PR #3497, opened 2024-03-05
Alternative stencil for shear production in CATKE's TKE equation

glwagner:
This formulation may be more friendly to stretched grid runs, since it more closely captures the exact discrete TKE produced by discrete mean momentum diffusion. Note that this is still not exact because it makes an approximation in _time_ (an exact formulation requires incorporating velocities at both the previous and current time-step).

@simone-silvestri


glwagner:
This is stale

PR #3499, opened 2024-03-06
Allow interpolation on `Distributed` architectures

simone-silvestri:
this change is necessary to interpolate on `Distributed` architectures


navidcy:
should we include an interpolation test on the distributed CI?

PR #3500, opened 2024-03-06
Change default scheme for `HydrostaticSphericalCoriolis`

glwagner:
The default should be `ActiveCellEnstrophyConserving` according to @simone-silvestri

### DIFF
--- a/src/Coriolis/Coriolis.jl
+++ b/src/Coriolis/Coriolis.jl
@@ -12,7 +12,9 @@ using Oceananigans.Operators
 
 # Physical constants for constructors.
 using Oceananigans.Grids: R_Earth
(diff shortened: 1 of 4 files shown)


navidcy (approved):

navidcy (changes requested):
you only seem to have changed the docstring function signature.
let's change the description below to agree with the signature...

and also, the constructor at the moment seems to suggests that it's actually the EnergyConserving that is the default, no?

navidcy (approved):

navidcy:
OK, some regression tests break with this change (probably because the data was produced with the default scheme which was different). Also some unit tests. I'll have a look.

glwagner:
oh sorry

navidcy:
No worries!
it brings up an issue that regression tests and generation scripts should not rely on default args/kwargs

navidcy (approved):

PR #3505, opened 2024-03-12
Enhance immersed boundary grid docs

navidcy:


src/ImmersedBoundaries/grid_fitted_bottom.jl
```diff
@@ -63,34 +63,35 @@ function Base.show(io::IO, ib::GridFittedBottom)
...
     bottom_field = Field{Center, Center, Nothing}(grid)
-    set!(bottom_field, ib.bottom_height)
+    set!(bottom_field, immersed_boundary.bottom_height)
     fill_halo_regions!(bottom_field)
-    new_ib = GridFittedBottom(bottom_field, ib.immersed_condition)
+    new_immersed_boundary = GridFittedBottom(bottom_field, immersed_boundary.immersed_condition)
     TX, TY, TZ = topology(grid)
-    return ImmersedBoundaryGrid{TX, TY, TZ}(grid, new_ib)
+    return ImmersedBoundaryGrid{TX, TY, TZ}(grid, new_immersed_boundary)
 end
```
simone-silvestri, line 79: I kinda prefer it how it was before. In my opinion "verbosity" is not required when everything is self-contained and clear as in this function (`ib` here already clearly comes in as a `GridFittedBottom`)

navidcy: okie!

navidcy:
how about now @simone-silvestri?

src/ImmersedBoundaries/grid_fitted_bottom.jl
```diff
@@ -24,18 +24,18 @@ struct InterfaceImmersedCondition end
...
     bottom_height :: H
     immersed_condition :: I
 end
 
 const GFBIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:GridFittedBottom}
 
+"""
+    GridFittedBottom(bottom_height, [immersed_condition=CenterImmersedCondition()])
+
+Return an immersed boundary with an irregular bottom fit to the underlying grid.
```
glwagner, line 37: the bottom isn't necessarily "irregular". The docstring should actually say what `bottom_height` is (it's an array or function that gives the height of the bottom in absolute `z` coordinates (ie not relative to the lowest level or anything like that)

navidcy: true -- thanks!

src/ImmersedBoundaries/grid_fitted_bottom.jl
```diff
@@ -34,7 +34,19 @@ const GFBIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:GridFit
...
+=================
+
+* `bottom_height`: an array or function that gives the height of the
+                   bottom in absolute ``z`` coordinates.
+
+* `immersed_condition`: determine whether the part of the domain that is 
+                        immersed are all the cell centers that lie below
+                        `bottom_height` (`CenterImmersedCondition()`; default)
+                        or all the cell faces that lie below `bottom_height`
+                        (`InterfaceImmersedCondition()`).
```
glwagner, line 49: Probably it should be noted that the only purpose of this is to allow `GridFittedBottom` and `PartialCellBottom` to have the same behavior when the minimum fractional cell height for partial cells is set to 0.

glwagner: Hard to concept to convey simply, but perhaps the main point is that most people don't care about this and shouldn't expend too much energy trying to understand it.

navidcy: okie!

glwagner (approved):

PR #3508, opened 2024-03-13
Added `last_Δt` to `Clock`

jagoosw:


glwagner:
I like `last` better than `previous`. But it's not perfect because "last" can also mean "final". Can anyone think of a better name?

Also, do we need `previous_dt` in the time-stepper anymore?

glwagner:
I think we also need to add something to `reset!`:

https://github.com/CliMA/Oceananigans.jl/blob/3bb62a647a55a7dadf5f37331321bf0020a78c4d/src/Simulations/simulation.jl#L163-L175

glwagner:
Also 

https://github.com/CliMA/Oceananigans.jl/blob/3bb62a647a55a7dadf5f37331321bf0020a78c4d/src/Simulations/simulation.jl#L164

should just be

```julia
 sim.model.clock.time = 0
```

(at least --- this will also have to be changed to support dates @navidcy)

src/TimeSteppers/clock.jl
```diff
@@ -6,30 +6,31 @@ import Base: show
...
 
-Keeps track of the current `time`, `iteration` number, and time-stepping `stage`.
+Keeps track of the current `time`, `last_Δt`, `iteration` number, and time-stepping `stage`.
 The `stage` is updated only for multi-stage time-stepping methods. The `time::T` is
 either a number or a `DateTime` object.
 """
-mutable struct Clock{T}
+mutable struct Clock{T, FT}
          time :: T
+      last_Δt :: FT
```
glwagner, line 17:
```suggestion
      last_Δt :: DT
```

This will stand the test of time, when we support date times and time-deltas for the time-step.

src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -112,7 +112,7 @@ Keyword arguments
 function ShallowWaterModel(;
                            grid,
                            gravitational_acceleration,
-                               clock = Clock{eltype(grid)}(0, 0, 1),
+                               clock = Clock{eltype(grid), eltype(grid)}(0, Inf, 0, 1),
```
glwagner, line 115: Personally I would add a constructor for `Clock` that takes one type rather than writing this. But you can decide.

glwagner: Also note that we might reasonably expect users to provide custom clocks with their own time-step and iteration --- but that the stage and time-step probably won't commonly be provided when building a clock. So I think it would make more sense to have a syntax that looks like

```julia
Clock{TT}(time, iteration, [optional args])
```

glwagner: I think the type of the time-step can probably be inferred from the time type `TT` (ie if the time type is either a floating point or a DateTime or something like that)

jagoosw: There actually is already a constructor I can just change it a bit to infer the `DT` https://github.com/CliMA/Oceananigans.jl/blob/3bb62a647a55a7dadf5f37331321bf0020a78c4d/src/TimeSteppers/clock.jl#L27

jagoosw: It seems like `Dates` only stores things as integers and is never directly used but the timestep is always `eltype(grid)` so I'm not sure it can be directly infered.

jagoosw:
Okay, so I've got a way we could do this. When a user specifies an `AbstractTime` we can then use the type they've given to set the timestep type like:
```julia
clock = Clock{eltype(grid)}(time = DateTime(2020))
```

src/Models/ShallowWaterModels/shallow_water_model.jl
glwagner: We don't support dates right now -- to support it, we have to generalize how the time-step specification works.

src/TimeSteppers/clock.jl
```diff
@@ -6,30 +6,33 @@ import Base: show
...
-    Clock(; time, iteration=0, stage=1)
+    Clock(; time, last_Δt = Inf, iteration=0, stage=1)
 
 Returns a `Clock` object. By default, `Clock` is initialized to the zeroth `iteration`
-and first time step `stage`.
+and first time step `stage` with `last_Δt`.
 """
-Clock(; time::T, iteration=0, stage=1) where T = Clock{T}(time, iteration, stage)
+Clock(; time::TT, last_Δt::DT=Inf, iteration=0, stage=1) where {TT, DT} = Clock{TT, DT}(time, last_Δt, iteration, stage)
+Clock{TT}(; time, last_Δt=Inf, iteration=0, stage=1) where TT = Clock{TT, TT}(time, last_Δt, iteration, stage)
```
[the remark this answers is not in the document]
glwagner: I thikn taht's good

glwagner:
> Okay, so I've got a way we could do this. When a user specifies an `AbstractTime` we can then use the type they've given to set the timestep type like:
> 
> ```julia
> clock = Clock{eltype(grid)}(time = DateTime(2020))
> ```

I think it'll work a little differently, because we will want the time-step to also be a period. We don't support it yet, but we will have to convert the time-step to a floating point somewhere internally.

src/TimeSteppers/clock.jl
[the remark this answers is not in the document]
glwagner: It is a little bit of a conundrum --- I need to understand all the possibilities with `DateTime` a bit more. I think when we use `DateTime`, there will be many possibilities for the time-step type (`Second`, `Minute`, etc...).  But we can figure it out when the time comes though and no need to overworry right now.

src/Models/ShallowWaterModels/shallow_water_model.jl
glwagner: > the timestep is always eltype(grid) so I'm not sure it can be directly infered.

We will have to change how time-step specification works to support DateTime properly.

src/TimeSteppers/clock.jl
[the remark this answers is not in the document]
glwagner: Oh I missed the line below

src/TimeSteppers/clock.jl
```diff
@@ -6,30 +6,33 @@ import Base: show
...
+Clock{DT}(; time::AbstractTime, last_Δt=Inf, iteration=0, stage=1) where DT = Clock{AbstractTime, DT}(time, last_Δt, iteration, stage)
```
glwagner, line 30:
```suggestion
# TODO: when supporting DateTime, this function will have to be extended
time_step_type(TT) = TT

function Clock{TT}(; time, last_Δt=Inf, iteration=0, stage=1) where TT
    DT = time_step_type(TT)
    last_Δt = convert(DT, last_Δt)
    return Clock{TT, DT}(time, last_Δt, iteration, stage)
end
```

And get rid of the `AbstractTime` function since I don't think it's what we want (also we don't properly support `AbstractTime` anyways)

src/Simulations/simulation.jl
```diff
@@ -161,7 +161,8 @@ run_wall_time(sim::Simulation) = prettytime(sim.run_wall_time)
 Reset `sim`ulation, `model.clock`, and `model.timestepper` to their initial state.
 """
 function reset!(sim::Simulation)
-    sim.model.clock.time = 0.0
+    sim.model.clock.time = 0
+    sim.model.clock.last_Δt = Inf
```
glwagner, line 165: excellent

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -50,7 +50,7 @@ end
 
 """
     HydrostaticFreeSurfaceModel(; grid,
-                                             clock = Clock{eltype(grid)}(0, 0, 1),
+                                             clock = Clock{eltype(grid)}(time = 0),
```
glwagner, line 53: An improvement I think!

src/TimeSteppers/clock.jl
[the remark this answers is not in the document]
jagoosw: Oh thank you good to know

glwagner:
Also going to change the title of this PR to use our convention --- try to do that in the future too!

glwagner (approved):

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
navidcy: for sure!

src/TimeSteppers/clock.jl
```diff
@@ -63,4 +74,4 @@ function tick!(clock, Δt; stage=false)
 end
 
 "Adapt `Clock` to work on the GPU via CUDAnative and CUDAdrv."
-Adapt.adapt_structure(to, clock::Clock) = (time=clock.time, iteration=clock.iteration, stage=clock.stage)
+Adapt.adapt_structure(to, clock::Clock) = (time=clock.time, last_Δt=clock.last_Δt, iteration=clock.iteration, stage=clock.stage)
```
navidcy, line 77: just for clarity:
```suggestion
Adapt.adapt_structure(to, clock::Clock) =
    (time=clock.time, last_Δt=clock.last_Δt, iteration=clock.iteration, stage=clock.stage)

```

navidcy (approved):
I like this

jagoosw:
These test failures are just testing problems not actual code problems right?

glwagner:
I think so, I'll retry them.

glwagner:
we close now

glwagner:
doesn't this mean that we can get rid of previous_dt from `QuasiAdamsBashforth2TimeStepper` ?

glwagner:
It's fine if we do it in another PR.

jagoosw:
Could you try and run the docs again?

jagoosw:
> doesn't this mean that we can get rid of previous_dt from `QuasiAdamsBashforth2TimeStepper` ?

I would have thought so

src/TimeSteppers/clock.jl
```diff
@@ -6,30 +6,39 @@ import Base: show
 import Oceananigans.Units: Time
 
 """
-    mutable struct Clock{T<:Number}
+    mutable struct Clock{T, FT}
```
navidcy, line 9:
```suggestion
    mutable struct Clock{T, DT}

```

PR #3518, opened 2024-03-21
Improve kernel efficiency of the WENO algorithm

simone-silvestri:
This PR tries to improve the GPU efficiency of the WENO algorithm by
- using fast math for smoothness calculation (this should not be a problem since weights are normalized)
- resorting the WENO algorithm to enforce register reuse by accumulating the solution instead of computing all the stencils at tge same time

WENO-Z weights are calculated as $$\alpha_s = C_s \left( 1 + \left(\frac{\tau}{\beta_s +\varepsilon}\right)^2 \right)$$
and the interpolation is calculated as $$\psi =\frac{1}{\sum \alpha_s} \sum \psi_s \alpha_s$$
so if we reorder we can calculate $\psi$ as $$\psi = \frac{ \tau^2 \hat{\psi}_1  + \hat{\psi}_2}{ \tau^2 \sum \alpha^{\star}_s  + 1}$$
where $$\hat{\psi}_1 = \sum\psi_s \alpha^{\star}_s$$ and $$\hat{\psi}_2 = \sum \psi_s C_s$$ and $\alpha^{\star}_s$ are the WENO-JS coefficients that depend only on the local stencil: $$\alpha^{\star}_s = \frac{C_s}{(\beta_s + \varepsilon)^2}$$
We can then calculate stencils one by one by accumulating the results and "throwing away" registers we don't need after the computation.

This PR is a draft because, despite it works, everything is written down manually unrolled and maybe there is a way to express the same concept with metaprogramming


src/AbstractOperations/conditional_operations.jl
```diff
@@ -1,6 +1,6 @@
 using Oceananigans.Fields: OneField
 using Oceananigans.Grids: architecture
-using Oceananigans.Architectures: on_architecture
+import Oceananigans.Architectures: on_architecture
```
glwagner, line 3: should we put this in another PR? this seems important...

src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
```diff
@@ -1,4 +1,4 @@
-using Oceananigans.Architectures: architecture, on_architecture
```
glwagner, line 1: all these `on_architecture` changes create some noise in this PR which is making it harder to review, might make sense to put them in another PR

src/Advection/weno_default_stencils.jl
```diff
@@ -0,0 +1,685 @@
+using Printf
+
+@inline getvalue(i, j, k, grid, ψ, args...) = @inbounds ψ[i, j, k]
+@inline getvalue(i, j, k, grid, ψ::Function, args...) = ψ(i, j, k, grid, args...)
```
glwagner, line 4: don't we have an `identity` function that does this same thing?

glwagner: https://github.com/CliMA/Oceananigans.jl/blob/427c92fac963ee0883fa2766df15ba6a48e62736/src/Operators/interpolation_utils.jl#L39-L41

glwagner:
I think the algorithm for saving register usage could be easier to understand if it is written abstractly (ie within a loop that goes to WENO order `N` rather than written manually. 

The main advantage of using metaprogramming is that it will be easier to maintain if this code needs to change in the future (ie even for the trivial reason that julia syntax changes). Rather than having to inspect and change 7 functions we can change one. It'll also main we can probably get away with fewer regression tests. Otherwise, to prevent the code from returning wrong results when/if it needs to be updated in the future, we need to test every WENO order...

These seem like pretty significant advantages, but I understand that everyone is busy.

glwagner:
Is this stale?

simone-silvestri:
yep, I ll close it

PR #3523, opened 2024-03-26
Tests for splitting output files using `TimeInterval`

josuemtzmo:
Adds tests for splitting output based on time-interval, e.g., 

```Julia
ow = JLD2OutputWriter(model, (; u);
                      filename = "test.jld2",
                      ...
                      file_splitting = TimeInterval(3seconds)
                      )
```


glwagner:
We should definitely get `TimeInterval` to work. This is important! Why doesn't it work? I don't understand the reason.

glwagner (changes requested):
Let's get `TimeInterval` to work.

test/test_jld2_output_writer.jl
```diff
@@ -88,6 +88,52 @@ function test_jld2_file_splitting(arch)
...
+    end
+    ow = JLD2OutputWriter(model, (; u=model.velocities.u);
+                          dir = ".",
+                          filename = "test.jld2",
+                          schedule = IterationInterval(1),
+                          init = fake_bc_init,
+                          including = [:grid],
+                          array_type = Array{Float64},
+                          with_halos = true,
+                          file_splitting = FileTimeSplit(3seconds),
```
glwagner, line 107:
```suggestion
                          file_splitting = TimeInterval(3seconds),
```

start with this change then we can see the errors

test/test_netcdf_output_writer.jl
```diff
@@ -90,6 +90,46 @@ function test_netcdf_file_splitting(arch)
...
+    fake_attributes = Dict("fake_attribute"=>"fake_attribute")
+
+    ow = NetCDFOutputWriter(model, (; u=model.velocities.u);
+                            dir = ".",
+                            filename = "test.nc",
+                            schedule = IterationInterval(2),
+                            array_type = Array{Float64},
+                            with_halos = true,
+                            global_attributes = fake_attributes,
+                            file_splitting = FileTimeSplit(4seconds),
```
glwagner, line 107:
```suggestion
                            file_splitting = TimeInterval(4seconds),
```

josuemtzmo:
> We should definitely get `TimeInterval` to work. This is important! Why doesn't it work? I don't understand the reason.

Yes, I was also surprised. I retried and now it passed the tests, so I guess at some point I was doing something weird with `TimeInterval`, but I'm not sure what it was. It seems to work now.

glwagner:
Awesome! And to be more clear, I think the reason we want it to work is so that we can also reuse all the other schedules. Another reason to use it is because we may need to change `TimeInterval` in the future if we support different time types (for example `DateTime`, or other time types that solve the annoying problem of rounding error). So in all those cases it will be nice not to have to worry about different implementations of time interval schedules.

src/OutputWriters/output_writer_utils.jl
```diff
@@ -12,6 +12,12 @@ using Oceananigans.Utils: AbstractSchedule
 ##### Output writer utilities
 #####
 
+struct NoFileSplitting end
+(::NoFileSplitting)(model) = false
+Base.summary(::NoFileSplitting) = "NoFileSplitting" 
+Base.show(io::IO, nfs::NoFileSplitting) = print(io, summary(nfs))
+
+
```
navidcy, line 20:
```suggestion
```

test/test_jld2_output_writer.jl
```diff
@@ -88,6 +88,52 @@ function test_jld2_file_splitting(arch)
...
+                          init = fake_bc_init,
+                          including = [:grid],
+                          array_type = Array{Float64},
+                          with_halos = true,
+                          file_splitting = TimeInterval(3seconds),
+                          overwrite_existing = true)
+
+    push!(simulation.output_writers, ow)
+
+    # 531 KiB of output will be written which should get split into 3 files.
```
navidcy, line 112: the comment about file size is not relevant here, right?

test/test_netcdf_output_writer.jl
```diff
@@ -90,6 +90,46 @@ function test_netcdf_file_splitting(arch)
...
+                            schedule = IterationInterval(2),
+                            array_type = Array{Float64},
+                            with_halos = true,
+                            global_attributes = fake_attributes,
+                            file_splitting = TimeInterval(4seconds),
+                            overwrite_existing = true)
+
+    push!(simulation.output_writers, ow)
+
+    # 531 KiB of output will be written which should get split into 3 files.
```
navidcy, line 112: comment about filesize is not relevant, right?

test/test_netcdf_output_writer.jl
```diff
@@ -891,4 +932,4 @@ for arch in archs
         test_netcdf_vertically_stretched_grid_output(arch)
         test_netcdf_regular_lat_lon_grid_output(arch)
     end
-end
+end
```
navidcy, line 935: we usually have an empty line at the end of file

```suggestion
end

```

navidcy (approved):
lgtm

navidcy:
@josuemtzmo could you update the first PR comment to reflect the latest status?

test/test_jld2_output_writer.jl
```diff
@@ -266,11 +312,12 @@ for arch in archs
...
-        #####
-        ##### File splitting
-        #####
+        ####
+        #### File splitting
+        ####
 
-        test_jld2_file_splitting(arch)
+        test_jld2_file_splitting_size(arch)
+        test_jld2_file_splitting_time(arch)
```
glwagner, line 320:
```suggestion
        test_jld2_size_file_splitting(arch)
        test_jld2_time_file_splitting(arch)
```

this doesn't matter, but does maintain our "english language style" where adjectives proceed nouns (yes, different than the way many codes are written, or French 😆 )

glwagner (approved):
awesome contribution! I agree with @navidcy's comments but except for the cosmetic stuff it looks good

test/test_netcdf_output_writer.jl
josuemtzmo: Yes, I deleted it

test/test_jld2_output_writer.jl
josuemtzmo: Yes, I deleted it.

navidcy:
Great! When all tests pass you can merge.

navidcy:
excellent @josuemtzmo !

PR #3524, opened 2024-03-27
Add interpolation for fields on `ImmersedBoundaryGrid`s

simone-silvestri:


glwagner (approved):
The name of this branch is "correct immersed", but to be clear there is nothing wrong with immersed interpolation right (this PR merely re-translates functions for `ImmersedBoundaryGrid`). Or is more needed here?

simone-silvestri:
yep interpolation for immersed boundary fields does not currently work because of these missing functions

navidcy:
merge?

PR #3530, opened 2024-04-01
Minor bug fix so that `Δzᶜᶜᶜ(... , ibg::PCBIBG)` works for grids with flat dims

navidcy:
There was a

```Julia
x, y, z = node(...)
```

but only `z` was needed. And `node` returns a different-size tuple depending on how many dimensions of `grid` are non-flat. Thus there was issues sometimes. Now that part of the code became

```Julia
z = znode(...)
```

which is much more robust!

### DIFF
--- a/src/ImmersedBoundaries/partial_cell_bottom.jl
+++ b/src/ImmersedBoundaries/partial_cell_bottom.jl
@@ -91,14 +91,14 @@ on_architecture(to, ib::PartialCellBottom) = PartialCellBottom(on_architecture(t
           ∘   k  | Δz
       k --x--    ↓
       
-Criterion is h >= z - ϵ Δz
+Criterion is h ≥ z - ϵ Δz
 
 """
 @inline function _immersed_cell(i, j, k, underlying_grid, ib::PartialCellBottom)
     # Face node above current cell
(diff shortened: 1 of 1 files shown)


glwagner:
Nice! I think the `node` was a legacy from when the bottom height could be a function of x, y, z. This is better now that we know it's pretty much 100% of the time better to precompute that.

glwagner (approved):

francispoulin (dismissed):
Looks good!

I think in line 122 you need a .- instead of -, for the code to work.

navidcy:
I don't think so.
Now both `h` and `z` are just floats so the `.` you suggest is redundant.

glwagner:
I don't think you want `.-` --- the metric functions are invoked inside kernels, where broadcasting doesn't work on the GPU.

navidcy:
@francispoulin do you still insist it needs broadcasting? Note that you need to approve the PR otherwise it can’t be merged.

francispoulin:
> @francispoulin do you still insist it needs broadcasting? Note that you need to approve the PR otherwise it can’t be merged. 
> 

Sorry for holding things up and please merge when you like.

PR #3531, opened 2024-04-01
Add a bottom boundary layer example using CATKE

glwagner:


validation/vertical_mixing_closures/bottom_boundary_layer.jl
```diff
@@ -0,0 +1,66 @@
...
+using Printf
+
+using Oceananigans.TurbulenceClosures.CATKEVerticalDiffusivities:
+    CATKEVerticalDiffusivity, MixingLength
+
+# Parameters
+Δz = 4          # Vertical resolution
+Lz = 256        # Extent of vertical domain
+Nz = Int(Lz/Δz) # Vertical resolution
+f₀ = 0.0        # Coriolis parameter (s⁻¹)
```
navidcy, line 13: Why so verbose to explicitly denote that coriolis parameter is zero?

validation/vertical_mixing_closures/bottom_boundary_layer.jl
```diff
@@ -0,0 +1,66 @@
...
+                                                      overwrite_existing = true)
+
+progress(sim) = @info string("Iter: ", iteration(sim), " t: ", prettytime(sim),
+                             ", max(u): ", maximum(model.velocities.u))
+simulation.callbacks[:progress] = Callback(progress, IterationInterval(100))
+
+@info "Running a simulation of $model..."
+
+run!(simulation)
+
```
navidcy, line 66: How about a plot?

navidcy (approved):
Made minor comments

validation/vertical_mixing_closures/bottom_boundary_layer.jl
glwagner: To show users how it might be changed

glwagner:
https://github.com/CliMA/Oceananigans.jl/assets/15271942/b41d7837-61bc-4da5-8b2c-6f2c740c307f

validation/vertical_mixing_closures/bottom_boundary_layer.jl
navidcy: Good point!

navidcy:
Great -- thanks!
The `e` looks a bit strange with 1 value very off the rest. Does that look OK to you?
(We don't need to deal with this in this PR -- I was just wondering.)

francispoulin:
I also thought that the `e` plot looked odd.  Is it just one point or is better resolved than that?

PR #3532, opened 2024-04-02
Add info about models that support BackgroundFields

navidcy:


docs/src/model_setup/background_fields.md
```diff
@@ -36,6 +36,9 @@ velocity field.
 Other possible terms associated with the Coriolis force, buoyancy, turbulence closures,
 and surface waves acting on background fields are neglected.
 
+!!! compat "Model compatibility"
+    At the moment _only_ the [`NonhydrostaticModel`](@ref) supports `BackgroundFields`.
```
glwagner, line 40:
```suggestion
    `BackgroundFields` are only supported by [`NonhydrostaticModel`](@ref).
```

glwagner: I feel like saying "at the moment" implies there are plans to implement support, but I don't think there are...

navidcy: True. No need to imply anything about the future. Talk about the present.

francispoulin (approved):
Good addition.

PR #3540, opened 2024-04-05
Use adaptive timestep in two-dimensional turbulence example

### LINKED ISSUE #3538: 2D turbulence example is outputting `NaNs`
I was just looking at the docs and noticed that the video for the 2D turbulence example goes blank, presumably because the output is `NaN`: https://clima.github.io/OceananigansDocumentation/stable/generated/two_dimensional_turbulence/.

The last version that appears to work is 0.90.9 so unless there's some change that might have cause this it might be worth setting the seed manually so that the initial conditions are always the same so it doesn't occasionally NaN?

navidcy:
Closes #3538

### DIFF
--- a/examples/two_dimensional_turbulence.jl
+++ b/examples/two_dimensional_turbulence.jl
@@ -139,8 +139,6 @@ nothing #hide
 using CairoMakie
 set_theme!(Theme(fontsize = 24))
 
-@info "Making a neat movie of vorticity and speed..."
-
 fig = Figure(size = (800, 500))
 
 axis_kwargs = (xlabel = "x",
--- a/examples/two_dimensional_turbulence.jl
+++ b/examples/two_dimensional_turbulence.jl
@@ -37,10 +37,6 @@ model = NonhydrostaticModel(; grid,
 
 using Statistics
 
-using Random
-
-Random.seed!(1234)
-
 u, v, w = model.velocities
 
 uᵢ = rand(size(u)...)
@@ -51,23 +47,18 @@ vᵢ .-= mean(vᵢ)
 
 set!(model, u=uᵢ, v=vᵢ)
 
-# ## Simulation
+# ## Setting up a simulation
 #
-# Next we construct a simulation. To choose an appropriate timestep we check the
-# CFL number. In this case, the initial flow is highly variable and given
-# that there is no forcing in the model we expect flow to smoothen with time.
-# Thus, the CFL that corresponds to the initial time should be a close upper bound.
-#
-# For the `:RungeKutta3` timestepper used in this model, a timestep `Δt` that
-# gives CFL ⪅ 1 should be adequate; the `:QuasiAdamsBashforth2` requires CFL ⪅ 0.25.
+# We set-up a simulation that stops at 50 time units, with an initial
+# time-step of 0.1, and with adaptive time-stepping and progress printing.
 
-Δt = 0.03
-cfl = AdvectiveCFL(Δt)
-cfl(model)
+simulation = Simulation(model, Δt=0.2, stop_time=50)
 
-# Now we create the simulation.
+# The `TimeStepWizard` helps ensure stable time-stepping
+# with a Courant-Freidrichs-Lewy (CFL) number of 0.7.
 
-simulation = Simulation(model, Δt=Δt, stop_time=50)
+wizard = TimeStepWizard(cfl=0.7, max_change=1.1, max_Δt=0.5)
+simulation.callbacks[:wizard] = Callback(wizard, IterationInterval(10))
 
 # ## Logging simulation progress
 #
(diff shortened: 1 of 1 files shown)


simone-silvestri:
wait a sec. The fact that this example NaNs worries me a bit. What is the CFL of the simulation? a Runge Kutta timestepper should be able to handle up to 0.7 easily

simone-silvestri:
I am trying it locally and it works on `main`

navidcy:
well it does not always NaN... but once it did and the guess was that it was because of the random initialization of velocities...

navidcy:
with the setup currently on `#main` the initial CFL is > 2!
of course it falls pretty quickly... but I guess there is a good chance it blows up if the randomness in the initial velocities conspires in that way?

francispoulin:
Reducing the time step so that the CFL condition is less than 2 seems like a good idea.  The fact that it behaves differently on different computers is odd.

simone-silvestri:
We are well above the recommended CFL number for |u| > 0.2 which will most likely happen with a random velocity field. 
(for |u| > 0.9 the CFL is 3.66)
I think the correct thing to do is to
(1) set a random seed (should be always done for reproducibility regardless)
(2) reduce the time step to satisfy CFL < 0.7
(3) Explicitly write in the example how to calculate the time step and the recommendation for time stepping (setting a time step size is not super trivial and many users could benefit from an explanation in the examples)

navidcy:
Deal!

examples/two_dimensional_turbulence.jl
```diff
@@ -37,6 +37,8 @@ model = NonhydrostaticModel(; grid,
 
 using Statistics
 
+using Random; Random.seed!(1234) #hide
```
glwagner, line 40:
```suggestion
using Random

Random.seed!(1234)
```

I think we want to show this!

examples/two_dimensional_turbulence.jl
```diff
@@ -47,14 +49,37 @@ vᵢ .-= mean(vᵢ)
...
+# CFL number. In this case, the initial flow is highly variable and given
+# that there is no forcing in the model we expect flow to smoothen with time.
+# Thus, the CFL that corresponds to the initial time should be a close upper bound.
+#
+# For the `:RungeKutta3` timestepper used in this model, a timestep `Δt` that
+# gives CFL ⪅ 1 should be adequate; the `:QuasiAdamsBashforth2` requires CFL ⪅ 0.25.
+
+Δt = 0.03
+cfl = AdvectiveCFL(Δt)
+cfl(model)
```
glwagner, line 64: Maybe it's better to simply compute the time-step from the CFL.

Also, for the purpose of pedagogy, I think it is nice not to "tune" numbers too much. eg why not 0.01 or 0.05 rather than 0.03? When parameters appear tuned, they draw attention from the reader into the details and away from the bigger picture, which is more important. Perhaps choosing the time-step to be 0.5 or 0.2 of the initial CFL will help?

The instability we saw previously did not occur at the first time step though. So I think using the initial condition is rather conservative. Still, the physical explanation does make sense. We might want to mention that larger time-steps can be stable too.

examples/two_dimensional_turbulence.jl
```diff
@@ -47,14 +49,37 @@ vᵢ .-= mean(vᵢ)
...
 
-progress(sim) = @info string("Iteration: ", iteration(sim), ", time: ", time(sim))
-add_callback!(simulation, progress, IterationInterval(100))
+using Printf
+
+progress_message(sim) = @printf("Iteration: %04d, time: %s, cfl: %s, wall time: %s\n",
+                                iteration(sim),
+                                time(sim),
+                                round(AdvectiveCFL(sim.Δt)(sim.model), digits=4),
+                                prettytime(sim.run_wall_time))
```
glwagner, line 80: Maybe make this a function and compute things on different lines for clarity

glwagner (approved):

examples/two_dimensional_turbulence.jl
navidcy: Ok. I resonate. 

Let’s just use the cfl-based criterion and adaptive timestep?

glwagner: I think we wanted to avoid the wizard to keep things simple. But I guess if we are computing CFL, maybe the point is moot. Maybe we want to encourage people to always use the wizard, too...

glwagner: In summary, using the wizard sounds good to me 🥲

navidcy: I also think we want to encourage using the wizard...

examples/two_dimensional_turbulence.jl
navidcy: done

PR #3488, opened 2024-03-01
Fill halos of conformal cubed sphere grid coordinates and metrics in a single pass

siddharthabishnu:
This PR fills the halos of the grid metrics of the conformal cubed sphere panels at face-face locations using a dedicated function instead of hard-coded values, similar to the existing approach for center-center, face-center, and center-face locations.

cc @navidcy

PR #3547, opened 2024-04-09
Implement kernel functions to fill cubed sphere halos

siddharthabishnu:
This PR implements kernel functions to fill halos of the cubed sphere, enabling cubed sphere simulations to run on a single GPU.


src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
navidcy, line 1: I see a lot of new functions here that I presume they belong in the `cubed_sphere_boundary_conditions.jl`?

navidcy:
What does `find_neighboring_panels` do that is not already included in the `grid.connectivity`?
`find_neighboring_panels` seems like a bit of a "hack" to me that only works for 1-region-per-panel configs? `grid.connectivity` is general. Why don't we use the grid's connectivity to read out the rank of the neighboring panels.

src/MultiRegion/cubed_sphere_boundary_conditions.jl
```diff
@@ -32,39 +33,81 @@ function fill_halo_regions!(field::CubedSphereField{<:Center, <:Center})
...
-    for region in 1:6
+    multiregion_field = Reference(field.data.regional_objects)
+    region = Iterate(1:6)
 
-        region_E, region_N, region_W, region_S = find_neighboring_panels(grid, region)
+    kernel_parameters = KernelParameters((Hc, Nc, Nz), (0, 0, 0))
+    @apply_regionally begin
+        launch!(grid.architecture, grid, kernel_parameters, _fill_cubed_sphere_center_center_field_east_west_halo_regions!,
+                field, multiregion_field, n_regions, region, Hc, Nc)
+    end
```
simone-silvestri, line 43:
```suggestion
    @apply_regionally begin
        launch!(grid.architecture, grid, (Hc, Nc, Nz), _fill_cubed_sphere_center_center_field_east_west_halo_regions!,
                field, multiregion_field, n_regions, region, Hc, Nc)
    end
```

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
siddharthabishnu: Yes, it's unclear why this script wasn't updated after merging the cs-grid-dynamics branch into main and subsequently the main branch into this one. Regardless, this issue has been taken care of in commit 57516c3.

siddharthabishnu:
> What does `find_neighboring_panels` do that is not already included in the `grid.connectivity`? `find_neighboring_panels` seems like a bit of a "hack" to me that only works for 1-region-per-panel configs? `grid.connectivity` is general. Why don't we use the grid's connectivity to read out the rank of the neighboring panels.
> 
> (I really should have made this comment in #3488, but it's never too late..)

It's been taken care of in commit [5cffc13](https://github.com/CliMA/Oceananigans.jl/pull/3547/commits/5cffc139eb93c1afbd68c520a95e21ba56d73990).

src/MultiRegion/cubed_sphere_boundary_conditions.jl
siddharthabishnu: It turns out I actually need the offset in the z direction for multi-layer halo fillings.

src/MultiRegion/cubed_sphere_boundary_conditions.jl
```diff
@@ -2,25 +2,7 @@ using Oceananigans.MultiRegion: number_of_regions
...
-        region_N = mod(region + 0, 6) + 1
-        region_W = mod(region + 4, 6) + 1
-        region_S = mod(region + 3, 6) + 1
-    end
-
-    return (; region_E, region_N, region_W, region_S)
-end
-
-function fill_halo_regions!(field::CubedSphereField{<:Center, <:Center})
+function fill_halo_regions!(field::CubedSphereField{<:Center, <:Center}; multilayer = true)
```
navidcy, line 5: what is a `multilayer`?
let's find a way to dispatch

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
navidcy: how do you merge?

src/MultiRegion/cubed_sphere_boundary_conditions.jl
navidcy: introducing a kwarg boolean and an if-else statement within a method affects performance

src/MultiRegion/cubed_sphere_boundary_conditions.jl
```diff
@@ -39,22 +20,23 @@ function fill_halo_regions!(field::CubedSphereField{<:Center, <:Center})
...
         launch!(grid.architecture, grid, kernel_parameters, _fill_cubed_sphere_center_center_field_east_west_halo_regions!,
-                field, multiregion_field, n_regions, region, Hc, Nc)
+                field, multiregion_field, region, grid.connectivity.connections, Hc, Nc)
     end
 
     kernel_parameters = KernelParameters((Nc, Hc, Nz), (0, 0, 0))
     @apply_regionally begin
         launch!(grid.architecture, grid, kernel_parameters, _fill_cubed_sphere_center_center_field_north_south_halo_regions!,
-                field, multiregion_field, n_regions, region, Nc, Hc)
+                field, multiregion_field, region, grid.connectivity.connections, Nc, Hc)
```
navidcy: I don't see why we need to give `grid.architecture`, `grid`, and `grid.connectivity.connections` as separate args. Why not just provide `grid` as argument and then `grid.architecture` and `grid.connectivity.connections` are all included?

navidcy: oh now I see: `launch!` needs `(arch, grid, params, kernel, args...)`...
so OK, I get why `grid.architecture is first arg; but why do we need to provide `grid.connectivity.connections` as separate argument?

simone-silvestri: probably no actually, because `getregion` happens on the arguments of `launch!`, not before

navidcy: gotcha!

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
siddharthabishnu: Let's say I want to merge the main branch into this one. Then I type:
```
git fetch origin main
git merge origin/main
```
I resolve the conflicts on VS Code by typically selecting `Accept Incoming` instead of `Accept Current`.

src/MultiRegion/cubed_sphere_boundary_conditions.jl
siddharthabishnu: I am sure there is a better way to handle this e.g. dispatch. I use the multilayer kwarg to distinguish between arrays that include a third dimension, like velocities, from those that do not, such as cubed sphere grid metrics.

src/MultiRegion/cubed_sphere_boundary_conditions.jl
siddharthabishnu: Correct. Since `@apply_regionally` happens on the arguments of `launch!`, I end up passing `grid.connectivity.connections[region]` for each `region`, which is defined only for `ConformalCubedSphereGrid`, and not for `OrthogonalSphericalShellGrid`.

siddharthabishnu: Even though `test_multiregion_cubed_sphere` runs smoothly on the CPU, I do get multiple errors on the GPU. 🙁

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
navidcy: OK, that's how I do it.

You'll have to be very careful in resolving the conflicts then. There is no one-answer-for-all, all conflicts have the be thought through.

src/MultiRegion/cubed_sphere_boundary_conditions.jl
simone-silvestri: looks like the problem is solved. There is a out-of-bound access in the metrics though

src/MultiRegion/cubed_sphere_boundary_conditions.jl
```diff
@@ -32,38 +14,84 @@ function fill_halo_regions!(field::CubedSphereField{<:Center, <:Center})
...
     Hc = Hx
 
-    #-- one pass: only use interior-point values:
-    for region in 1:6
+    multiregion_field = Reference(field.data.regional_objects)
+    region = Iterate(1:6)
 
-        region_E, region_N, region_W, region_S = find_neighboring_panels(grid, region)
+    multilayer ? nZ = Nz + 2Hz : nZ = Nz
+    multilayer ? hZ = -Hz : hZ = 0
```
simone-silvestri, line 21:
```suggestion
    nZ = multilayer ?Nz + 2Hz : Nz
    hZ = multilayer ?  -Hz : 0
```

src/MultiRegion/cubed_sphere_boundary_conditions.jl
navidcy: why don't you use their type?

src/MultiRegion/cubed_sphere_boundary_conditions.jl
navidcy: btw, if we are gonna call `fill_halo_regions!` with `@apply_regionally` in front, then how will the methods with typed args as `CubedSphereField`s, e.g., 

```
fill_halo_regions!(field::CubedSphereField{<:Face, <:Face})
```

ever are used?

simone-silvestri: `fill_halo_regions!` is not a local function because it requires communication so it should not be called with `@apply_regionally` but it should contain apply regionally inside

siddharthabishnu:
Replacing
```julia
@apply_regionally begin
    launch!(grid.architecture, grid, kernel_parameters,
            _fill_cubed_sphere_center_center_field_east_west_halo_regions!, field, multiregion_field, region,
            grid.connectivity.connections, Nc, Hc)
end
```
by
```julia
for region in number_of_regions(grid)
    launch!(grid.architecture, grid[region], kernel_parameters,
            _fill_cubed_sphere_center_center_field_east_west_halo_regions!, field[region], field.data.regional_objects, region,
            grid.connectivity.connections[region], Nc, Hc)
end 
```
reduces the wall time of filling only the east-west halos of a `CenterField` on a `ConformalCubedSphereGrid` with one region of 12x128 cells per panel from
```julia
julia> @btime fill_halo_regions!(c)
  711.639 μs (714 allocations: 950.66 KiB)
```
to
```julia
julia> @btime fill_halo_regions!(c)
  13.371 μs (61 allocations: 14.83 KiB)
```
Upcoming commits will incorporate modifications of this type.

navidcy:
I'm wondering if the loop:

```julia
for region in number_of_regions(grid)
...
end
```

can be done in parallel.

glwagner:
@siddharthabishnu do you need help with this PR?

glwagner:
closing as stale, re-open if needed

PR #3549, opened 2024-04-11
Fix `interpolator` for left side of periodic domain

### LINKED ISSUE #3114: Interpolation over periodic dimension returns strange result?
In an attempt to debug #3100 I ran onto this

```julia
using Oceananigans

grid = RectilinearGrid(CPU(), size=(4, 4, 4), extent=(1, 1, 1), topology=(Periodic, Periodic, Bounded))

buoyancy = SeawaterBuoyancy(gravitational_acceleration=1, equation_of_state = LinearEquationOfState(thermal_expansion=1, haline_contraction=1))

model = NonhydrostaticModel(; grid, buoyancy, tracers = (:T, :S))

set!(model.velocities.u, 1)

model.velocities.u
```

gives

```julia
4×4×4 Field{Face, Center, Center} on RectilinearGrid on CPU
```

But then, if I define a new `u` interpolated on cell centers:

```julia
new_u = @at (Center, Center, Center) u
u_new = Field(new_u)
compute!(u_new)
interior(u_new)
```

I get

```Julia
4×4×4 view(::Array{Float64, 3}, 4:7, 4:7, 4:7) with eltype Float64:
[:, :, 1] =
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 0.5  0.5  0.5  0.5

[:, :, 2] =
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 0.5  0.5  0.5  0.5

[:, :, 3] =
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 0.5  0.5  0.5  0.5

[:, :, 4] =
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 0.5  0.5  0.5  0.5
```

I would expect that the interpolated `u_new` still would have values 1 everywhere. No?

If I change the topology to Bounded then the interpolated `u_new` is all 1s.

(The above behaviour occurs with either Julia v1.8 or v1.9.)

jagoosw:
Fixes #3114

### DIFF
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -252,7 +252,7 @@ left bound `i⁻`, such that `ξ ∈ [0, 1)`.
 
     i⁻ = Base.unsafe_trunc(Int, fractional_idx)
     i⁻ = Int(i⁻ + 1) # convert to "proper" integer?
-    i⁺ = i⁻ + 1
+    i⁺ = i⁻ + 1 * Int(sign(fractional_idx))
     ξ = mod(fractional_idx, 1)
 
     return (i⁻, i⁺, ξ)
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -252,7 +252,8 @@ left bound `i⁻`, such that `ξ ∈ [0, 1)`.
 
     i⁻ = Base.unsafe_trunc(Int, fractional_idx)
     i⁻ = Int(i⁻ + 1) # convert to "proper" integer?
-    i⁺ = i⁻ + 1 * Int(sign(fractional_idx))
+    shift = Int(sign(fractional_idx))
+    i⁺ = i⁻ + shift
     ξ = mod(fractional_idx, 1)
 
     return (i⁻, i⁺, ξ)
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.90.12"
+version = "0.90.13"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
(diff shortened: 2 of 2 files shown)


navidcy (commented):
Hi @jagoosw! Could you explain a bit why this fixes the issue and demonstrate it with an example?

jagoosw:
Sure, I'll use the example of how I found it which was interpolating a Center field to a face field. 

The fractional index of the i=1 face point is -0.5, and of the i=2 point is 0.5. Truncating these returns 0 for both so the indices for interpolation are 1 and 2 which is correct for the i=2 point but wrong for the i=1 point because the code assumed that's the fractional index is larger than the truncated index which is only true when it is positive.

jagoosw:
Example here: https://github.com/CliMA/Oceananigans.jl/issues/3114#issuecomment-2049679195

src/Fields/interpolate.jl
```diff
@@ -252,7 +252,7 @@ left bound `i⁻`, such that `ξ ∈ [0, 1)`.
 
     i⁻ = Base.unsafe_trunc(Int, fractional_idx)
     i⁻ = Int(i⁻ + 1) # convert to "proper" integer?
-    i⁺ = i⁻ + 1
+    i⁺ = i⁻ + 1 * Int(sign(fractional_idx))
```
glwagner, line 255:
```suggestion
    shift = Int(sign(fractional_idx))
    i⁺ = i⁻ + shift
```

Not sure I understand what this is doing, but I don't think we need to multiply anything by `1` right?

jagoosw: yeah not sure why I wrote it that way before

simone-silvestri (approved):

navidcy:
should we merge this?

jagoosw:
Sorry forgot about this, merged now!

src/Fields/interpolate.jl
```diff
@@ -269,7 +269,8 @@ left bound `i⁻`, such that `ξ ∈ [0, 1)`.
 
     i⁻ = Base.unsafe_trunc(Int, fractional_idx)
     i⁻ = Int(i⁻ + 1) # convert to "proper" integer?
-    i⁺ = i⁻ + 1
+    shift = Int(sign(fractional_idx))
```
glwagner, line 272: It'd be good to add a comment that explains why this is needed.

glwagner: @simone-silvestri since you approved this, you can help @jagoosw if he is busy.

PR #3553, opened 2024-04-15
Update deps

navidcy:
@glwagner I'm testing in this PR using https://github.com/glwagner/OceananigansArtifacts.jl

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -34,14 +34,15 @@ version = "1.1.1"
 
 [[deps.ArrayInterface]]
 deps = ["Adapt", "LinearAlgebra", "SparseArrays", "SuiteSparse"]
-git-tree-sha1 = "44691067188f6bd1b2289552a23e4b7572f4528d"
+git-tree-sha1 = "133a240faec6e074e07c31ee75619c90544179cf"
 uuid = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-version = "7.9.0"
+version = "7.10.0"
 
     [deps.ArrayInterface.extensions]
     ArrayInterfaceBandedMatricesExt = "BandedMatrices"
     ArrayInterfaceBlockBandedMatricesExt = "BlockBandedMatrices"
     ArrayInterfaceCUDAExt = "CUDA"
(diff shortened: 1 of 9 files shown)


glwagner (approved):

navidcy:
I'm quite confused that this test from `test_seawater.jl` fails:

https://buildkite.com/clima/oceananigans/builds/15457#018eec75-847a-4323-b8e9-9f104d750297/37-1020

On tartarus everything passes...

navidcy:
> I'm quite confused that this test from `test_seawater.jl` fails:
> 
> https://buildkite.com/clima/oceananigans/builds/15457#018eec75-847a-4323-b8e9-9f104d750297/37-1020
> 
> On tartarus everything passes...

OK I can reproduce the fail on tartarus. I'm on it.

navidcy:
The differences are O(1e-13).

jbisits:
> This test:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/63cb6713efed46d47372038cc75c11192c75b73b/test/test_seawater_density.jl#L90
> 
> is failing on `GPU()` with `Float64` and `RoquetEquationOfState(:SimplestRealistic)`. I don't know if something changed -- I think nothing has changed.
> 
> I suggest `.==` -> `.\approx`
> 
> (cc'ing @jbisits, @glwagner in case you have an idea of something that changed in SeaWaterPolynomials or you can see a flaw in the tests, eg FT not passed in properly etc)

Just curious, does the test for the `potential_density` function

https://github.com/CliMA/Oceananigans.jl/blob/63cb6713efed46d47372038cc75c11192c75b73b/test/test_seawater_density.jl#L118

also fail? (when on `GPU()` with `Float64` and `RoquetEquationOfState(:SimplestRealistic)`).

navidcy:
Test of potential density pass. 
The insitundensity difference for k=2 is instead of 0 is 2e-13…

jbisits:
> Test of potential density pass. The insitu density difference for k=2 is instead of 0 is 2e-13…

In that case I think your suggestion  `.== `-> `.\approx` should do the trick!

PR #3558, opened 2024-04-23
Bugfix on FieldTimeSeries interpolation

simone-silvestri:


src/OutputReaders/field_time_series_indexing.jl
```diff
@@ -198,27 +198,28 @@ function interpolate!(target_fts::FieldTimeSeries, source_fts::FieldTimeSeries)
     source_location = map(instantiate, location(source_fts))
     target_location = map(instantiate, location(target_fts))
 
+    target_times = target_fts.times
```
glwagner, line 201:
```suggestion
    target_times = map(Time, target_fts.times)
```

src/OutputReaders/field_time_series_indexing.jl
```diff
@@ -198,27 +198,28 @@ function interpolate!(target_fts::FieldTimeSeries, source_fts::FieldTimeSeries)
...
+
     launch!(arch, target_grid, size(target_fts),
             _interpolate_field_time_series!,
-            target_fts.data, target_grid, target_location,
-            source_fts.data, source_grid, source_location)
+            target_fts.data, target_grid, target_location, Time.(target_times),
```
glwagner, line 205:
```suggestion
            target_fts.data, target_grid, target_location, target_times,
```

remember that we want to use `map` instead of broadcasting where applicable.

PR #3559, opened 2024-04-24
Adding tracer advection in a Single column model

simone-silvestri:
At the moment, tracer advection is `nothing` in a single-column model, which is consistent with zero velocities and saves computational time.

However, now that we have background velocities and biogeochemical velocities implemented using the same advection scheme as the tracer advection we need to change this behavior so that we can have tracers being advected by background (or sinking) vertical velocities in a single column model.

found by @hengdiliang


glwagner:
Hmm true. We can still omit horizontal advection though I guess, but maybe having `Flat` is enough.

simone-silvestri:
right, I think `Flat` assumes that `advective_tracer_flux = zero(grid)`

glwagner (approved):

PR #3567, opened 2024-04-28
Extend operations for two `ZeroField`s

zhihua-zheng:
Code suggested by @glwagner.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.90.12"
+version = "0.90.13"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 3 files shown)


tomchor (commented):
Can you add tests?

zhihua-zheng:
> Can you add tests?

New tests passed.

tomchor:
> > Can you add tests?
> 
> New tests passed.

Thanks! Some of the tests segfaulted so I restarted them. I'll approve it once tests pass and then we should be ready to go! Feel free to ping me if I'm taking too long ;)

glwagner (approved):

tomchor (approved):
Tests passed!

tomchor:
@glwagner are you okay if we register a new patch release version so that we can move forward with https://github.com/tomchor/Oceanostics.jl/pull/169 in Oceanostics?

glwagner:
Go for it

tomchor:
> Go for it

Thanks! `Project.toml` is already at 0.90.14, while the last registered version is still at 0.90.13, so I'll just register a new version without bumping the patch number.

PR #3570, opened 2024-04-30
Extract major source code modifications from PR 3306

siddharthabishnu:
This PR extracts the major source code modifications from PR #3306, especially the elimination of replace_horizontal_vector_halos!, refinement of existing methods and addition of new ones to take the cubed sphere grid into consideration, and the improved vorticity computation function. A new test will be implemented to verify that the norm of the prognostic variables of the Rossby-Haurwitz wave remains almost constant as it propagates over the cubed sphere.

### DIFF
--- a/src/Fields/field.jl
+++ b/src/Fields/field.jl
@@ -122,10 +122,10 @@ julia> grid = RectilinearGrid(size=(2, 3, 4), extent=(1, 1, 1));
 
 julia> ω = Field{Face, Face, Center}(grid)
 2×3×4 Field{Face, Face, Center} on RectilinearGrid on CPU
-├── grid: 2×3×4 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 2×3×3 halo
+├── grid: 2×3×4 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
 ├── boundary conditions: FieldBoundaryConditions
 │   └── west: Periodic, east: Periodic, south: Periodic, north: Periodic, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
-└── data: 6×9×10 OffsetArray(::Array{Float64, 3}, -1:4, -2:6, -2:7) with eltype Float64 with indices -1:4×-2:6×-2:7
+└── data: 8×9×10 OffsetArray(::Array{Float64, 3}, -2:5, -2:6, -2:7) with eltype Float64 with indices -2:5×-2:6×-2:7
     └── max=0.0, min=0.0, mean=0.0
 ```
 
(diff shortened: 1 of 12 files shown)


navidcy:
Let me know when this is ready to review.

Could you rename this to a title the just refers to what has happened in this PR?

src/Fields/field.jl
```diff
@@ -245,6 +245,10 @@ function offset_windowed_data(data, Loc, grid, indices)
         windowed_parent = parent(data)
     else
         parent_indices = map(parent_index_range, indices, loc, topo, halo)
+        if size(data)[3] == 1 && first.(axes(data))[3] == grid.Nz + 1 # take ssh into consideration
+            parent_indices = collect(parent_indices)
+            parent_indices[3] = 1:1
+        end
```
navidcy, line 251: this seems like a hack

also, what does `ssh` refers to?

glwagner:
@siddharthabishnu, what capability does this PR add?

siddharthabishnu:
> Let me know when this is ready to review.
> 
> Could you rename this to a title the just refers to what has happened in this PR?

I just renamed the title to reflect the main functionality of this PR. I will let you know it's ready for review.

siddharthabishnu:
> @siddharthabishnu, what capability does this PR add?

@glwagner, this PR uses the new halo-filling functions to update halos for prognostic variables and refines the vorticity computation function on the cubed sphere. Additional features introduced in this PR are detailed in the description.

src/Fields/field.jl
siddharthabishnu: Yes, it is a temporary hack. I have outlined the issue arising in the absence of this hack in issue #3572 and created PR #3573 to resolve it. In the comment, 'ssh' refers to variables like shh, for which the parent array takes the z-halo into consideration even though it does not exist.

navidcy: OK. Well we won't merge a hack, right?

siddharthabishnu: No, once I fix it in PR #3573, we will import the fix (not hack) in this PR.

navidcy: OK, I see. This PR requires first #3573. Gotcha!

navidcy:
Does this PR resolved #1584?

src/Fields/field.jl
navidcy: I don't understand though why filling halos in z for free surface has a problem now; that was OK up to now, right? Why is #3573 holding us here?

navidcy: @siddharthabishnu could you merge `main`? I think then this comment is not relevant any more and can be resolved ;)

siddharthabishnu:
> Does this PR resolved #1584 (amongst other things)?

Yes, it does.

siddharthabishnu:
To compare the numerical solution of the Rossby-Haurwitz wave on the cubed sphere against an established benchmark, I chose the exact solution of the non-linear barotropic vorticity equation on the sphere (not the solution of the non-linear shallow water equations). This was achieved by replacing the initial phase Rλ with the time-dependent phase Rλ−νt in Equations (143)–(146) of [Williamson et al.](https://www.sciencedirect.com/science/article/pii/S0021999105800166), using the angular velocity ν given by Equation (142). The results at a resolution of 128x128 cells per panel are as follows:

- Since we prescribe the initial stream function and numerically take its derivatives to specify the initial velocities on the cubed sphere, numerical errors are present even in the initial velocities and the diagnosed initial vorticity. The vorticity error is significantly pronounced near the cubed sphere corners due to the deviation from orthogonality. For instance, in the domain's interior, the vorticity error norm is three orders of magnitude less than the vorticity norm, whereas at the corners, it is nearly the same order of magnitude.
- As the simulation progresses, the amplified error from the corners propagates into the interior as expected. After 500 time steps, the vorticity error in the interior increases from three orders of magnitude less to one order of magnitude less than the vorticity norm.

I have attached panel-wise plots of the exact vorticity and the error of its numerical counterpart on the cubed sphere for the initial condition and after 500 time steps for reference. Given these circumstances, @navidcy and I are contemplating whether we should continue using the exact solution for verification in the test script or consider alternative verification methods, such as ensuring that the norm of the numerical solution of the Rossby-Haurwitz wave remains bounded within a specified range (as mentioned in the introduction to this PR). @glwagner, @simone-silvestri and @jm-c, let us know your thoughts.

cubed_sphere_rossby_haurwitz_wave_ζ₀_exact

cubed_sphere_rossby_haurwitz_wave_ζ₀_error

cubed_sphere_rossby_haurwitz_wave_ζ_exact

cubed_sphere_rossby_haurwitz_wave_ζ_error

navidcy:
so to be honest I'm not sure whether this can be a dynamics test; something for which the error is just an order of magnitude less than the solution itself seems a bit loose for me...

isn't there anything else we can use for a dynamics test? feels like we've been stuck on this for ages... we need a dynamics test to merge the cubed sphere code and move on I think. Perhaps there is no dynamics test we can have?

navidcy:
What if we have `free_surface=nothing`? Would that mean that we are solving the barotropic vortucity eq with Nz=1?

glwagner:
closing as stale, re-open if needed

PR #3574, opened 2024-05-01
(0.91.0) Make hydrostatic pressure anomaly optional in `NonhydrostaticModel`

### LINKED ISSUE #3364: bug with background fields?
Hi all,
We have come across what seems to be a bug in background_fields.  This arose in another context, but the code below is the simplest way that I have come up with so far that demonstrates the problem.  Here there is a stable linear buoyancy profile (imposed as a background gradient) with random noise for the initial velocity.  This generates extra turbulence at the top and bottom.  I wonder if the background gradient in the halo region isn't playing nicely with the initial conditions.  Has anyone else seen something like this or have any ideas for where to look for the problem?
I am using Julia 1.9.2 and Oceananigans 0.85.0
Thanks!
John

```julia
using Oceananigans
using JLD2
using Plots
using Printf

Lx = 1
Lz = 1
Nx = 64
Nz = 64

grid = RectilinearGrid(size=(Nx, Nz), x=(0, Lx), z=(0, Lz), topology=(Periodic, Flat, Periodic))

# Create a stable background buoyancy gradient
N = 1       # buoyancy frequency [s⁻¹]
B_func(x, y, z, t, N) = N^2 * z
B = BackgroundField(B_func, parameters=N)

                            advection = WENO(),
                            timestepper = :RungeKutta3,
                            closure = ScalarDiffusivity(ν=1e-6, κ=1e-6),
                            tracers = :b,
                            buoyancy = BuoyancyTracer(),
                            background_fields = (; b=B)) # `background_fields` is a `NamedTuple`

A=0.01
u_ic(x,y,z) = A * randn()
w_ic(x,y,z) = A * randn()

set!(model, u=u_ic, w=w_ic)

simulation = Simulation(model, Δt = 0.01, stop_iteration=200)

wizard = TimeStepWizard(cfl=0.3, diffusive_cfl=0.3, max_change=1.2, max_Δt=1)
simulation.callbacks[:wizard] = Callback(wizard, IterationInterval(1))

progress(sim) = @printf("i: % 6d, sim time: % 1.3f, wall time: % 10s, Δt: % 1.4f, advective CFL: %.2e, diffusive CFL: %.2e\n",
                        iteration(sim), time(sim), prettytime(sim.run_wall_time),
                        sim.Δt, AdvectiveCFL(sim.Δt)(sim.model), DiffusiveCFL(sim.Δt)(sim.model))

filename = "background_field_test"
simulation.output_writers[:velocities] = JLD2OutputWriter(model, model.velocities,
                                                          filename = filename * ".jld2",
                                                          overwrite_existing = true)

run!(simulation)

xu, yu, zu = nodes(model.velocities.u) 
xw, yw, zw = nodes(model.velocities.w)

file = jldopen(filename * ".jld2")

iterations = parse.(Int, keys(file["timeseries/t"]))

anim = @animate for (i,iter) in enumerate(iterations)
    @info "Drawing frame $i from iteration $iter..."

    u = file["timeseries/u/$iter"][:,1,:];
    w = file["timeseries/w/$iter"][:,1,:];
    t = file["timeseries/t/$iter"];

    ## Plot the heatmaps
    u_plot = Plots.heatmap(xu, zu, u', color=:thermal, xlabel="x", ylabel="z", aspect_ratio = :equal, xlims = (0, Lx), ylims = (0, Lz));
    w_plot = Plots.heatmap(xw, zw, w', color=:thermal, xlabel="x", ylabel="z", aspect_ratio = :equal, xlims = (0, Lz), ylims = (0, Lz));
    u_title = @sprintf("u, t=%s",round(t));
    w_title = @sprintf("w, t=%s",round(t));
	
    Plots.plot(u_plot,w_plot,layout=(1,2),size=(1200,600),title =[u_title w_title]);

    iter == iterations[end] && close(file) # close the file if we reach the end

    end

mp4(anim, filename * ".mp4", fps = 10);
```

glwagner:
This PR makes the separation computation of the hydrostatic pressure anomaly an optional feature of `NonhydrostaticModel`. It's also off by default, which hopefully means that triply periodic simulations should work by default (@johnryantaylor a while ago you wanted to do a triply periodic simulation, hopefully this PR makes that possible).

@tomchor ready to review, let me know what you think about the design.

This PR supercedes #3080, where there is an extended description of this change plus a discussion about its pros and cons.

Closes #3364 (I think...)

### DIFF
--- a/src/Fields/field_tuples.jl
+++ b/src/Fields/field_tuples.jl
@@ -203,42 +203,6 @@ TracerFields(::Union{Tuple{}, Nothing}, grid, bcs) = NamedTuple()
 "Shortcut constructor for empty tracer fields."
 TracerFields(::NamedTuple{(), Tuple{}}, grid, bcs) = NamedTuple()
 
-#####
-##### Pressure fields tuples
-#####
-
-"""
-    PressureFields(grid, bcs::NamedTuple)
-
-Return a `NamedTuple` with pressure fields `pHY′` and `pNHS` initialized as
-`CenterField`s on `grid`.  Boundary conditions `bcs`
-may be specified via a named tuple of `FieldBoundaryCondition`s.
-"""
-function PressureFields(grid, bcs=NamedTuple())
-
-    default_pressure_boundary_conditions =
-        (pHY′ = FieldBoundaryConditions(grid, (Center, Center, Center)),
-         pNHS = FieldBoundaryConditions(grid, (Center, Center, Center)))
-
-    bcs = merge(default_pressure_boundary_conditions, bcs)
-
-    pHY′ = CenterField(grid, boundary_conditions=bcs.pHY′)
-    pNHS = CenterField(grid, boundary_conditions=bcs.pNHS)
-
-    return (pHY′=pHY′, pNHS=pNHS)
-end
-
-function PressureFields(grid::AbstractGrid{<:Any, <:Any, <:Any, <:Flat}, bcs=NamedTuple())
-    default_pressure_boundary_conditions =
-        (pHY′ = FieldBoundaryConditions(grid, (Center, Center, Center)),
-         pNHS = FieldBoundaryConditions(grid, (Center, Center, Center)))
-
-    bcs = merge(default_pressure_boundary_conditions, bcs)
-    pNHS = CenterField(grid, boundary_conditions=bcs.pNHS)
-
-    return (; pHY′=nothing, pNHS=pNHS)
-end
-
 """
     TendencyFields(grid, tracer_names;
                    u = XFaceField(grid),
--- a/src/Fields/field_tuples.jl
+++ b/src/Fields/field_tuples.jl
@@ -232,7 +232,6 @@ end
 #####
 
 VelocityFields(::Nothing, grid, bcs) = VelocityFields(grid, bcs)
-PressureFields(::Nothing, grid, bcs) = PressureFields(grid, bcs)
 
 """
     VelocityFields(proposed_velocities::NamedTuple{(:u, :v, :w)}, grid, bcs)
@@ -265,19 +264,4 @@ function TracerFields(proposed_tracers::NamedTuple, grid, bcs)
     tracer_fields = Tuple(CenterField(grid, boundary_conditions=bcs[c], data=proposed_tracers[c].data) for c in tracer_names)
 
     return NamedTuple{tracer_names}(tracer_fields)
-end
-
-"""
-    PressureFields(proposed_pressures::NamedTuple{(:pHY′, :pNHS)}, grid, bcs)
-
-Return a `NamedTuple` of pressure fields with, overwriting boundary conditions
-in `proposed_tracer_fields` with corresponding fields in the `NamedTuple` `bcs`.
-"""
-function PressureFields(proposed_pressures::NamedTuple{(:pHY′, :pNHS)}, grid, bcs)
-    validate_field_tuple_grid("pressures", proposed_pressures, grid)
-
-    pHY′ = CenterField(grid, boundary_conditions=bcs.pHY′, data=proposed_pressures.pHY′.data)
-    pNHS = CenterField(grid, boundary_conditions=bcs.pNHS, data=proposed_pressures.pNHS.data)
-
-    return (pHY′=pHY′, pNHS=pNHS)
 end
--- a/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
+++ b/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
@@ -7,7 +7,7 @@ using Oceananigans.Advection: CenteredSecondOrder
 using Oceananigans.BuoyancyModels: validate_buoyancy, regularize_buoyancy, SeawaterBuoyancy
 using Oceananigans.Biogeochemistry: validate_biogeochemistry, AbstractBiogeochemistry, biogeochemical_auxiliary_fields
 using Oceananigans.BoundaryConditions: regularize_field_boundary_conditions
-using Oceananigans.Fields: BackgroundFields, Field, tracernames, VelocityFields, TracerFields
+using Oceananigans.Fields: BackgroundFields, Field, tracernames, VelocityFields, TracerFields, CenterField
 using Oceananigans.Forcings: model_forcing
 using Oceananigans.Grids: inflate_halo_size, with_halo, architecture
 using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid
@@ -161,12 +161,12 @@ function NonhydrostaticModel(; grid,
     # First, we extract boundary conditions that are embedded within any _user-specified_ field tuples:
     embedded_boundary_conditions = merge(extract_boundary_conditions(velocities),
                                          extract_boundary_conditions(tracers),
-                                         extract_boundary_conditions(pressures),
                                          extract_boundary_conditions(diffusivity_fields))
 
     # Next, we form a list of default boundary conditions:
     prognostic_field_names = (:u, :v, :w, tracernames(tracers)..., keys(auxiliary_fields)...)
-    default_boundary_conditions = NamedTuple{prognostic_field_names}(FieldBoundaryConditions() for name in prognostic_field_names)
+    default_boundary_conditions = NamedTuple{prognostic_field_names}(FieldBoundaryConditions()
+                                                                     for name in prognostic_field_names)
 
     # Finally, we merge specified, embedded, and default boundary conditions. Specified boundary conditions
     # have precedence, followed by embedded, followed by default.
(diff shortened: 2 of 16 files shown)


src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -161,12 +161,12 @@ function NonhydrostaticModel(; grid,
     # First, we extract boundary conditions that are embedded within any _user-specified_ field tuples:
     embedded_boundary_conditions = merge(extract_boundary_conditions(velocities),
                                          extract_boundary_conditions(tracers),
-                                         extract_boundary_conditions(pressures),
```
tomchor, line 157: I don't fully understand the implications of this change. We're not trying to support user-defined BCs for pressure? Is that it?

tomchor:
This is looking pretty good! My only comment is that it's not intuitive for a given user to figure out how to opt for a hydrostatic pressure separation. A flag called `separate_hydrostatic_pressure` that takes `true`/`false` would be much more intuitive, although we'd need a little more code. Should we prioritize user-friendliness here?

glwagner:
> This is looking pretty good! My only comment is that it's not intuitive for a given user to figure out how to opt for a hydrostatic pressure separation. A flag called `separate_hydrostatic_pressure` that takes `true`/`false` would be much more intuitive, although we'd need a little more code. Should we prioritize user-friendliness here?

This is more flexible, because sometimes users want to have access to the pressure field prior to model construction. If we use a flag, then we either can't support that or have to put some annoying logic in the constructor. Since I feel it'll be rare that people want to change this kwarg, I think the trade-offs work out that its better to have a simpler constructor even if those rare users that want to separate hydrostatic pressure have to build `CenterField`.

tomchor:
> > This is looking pretty good! My only comment is that it's not intuitive for a given user to figure out how to opt for a hydrostatic pressure separation. A flag called `separate_hydrostatic_pressure` that takes `true`/`false` would be much more intuitive, although we'd need a little more code. Should we prioritize user-friendliness here?
> 
> This is more flexible, because sometimes users want to have access to the pressure field prior to model construction. If we use a flag, then we either can't support that or have to put some annoying logic in the constructor. Since I feel it'll be rare that people want to change this kwarg, I think the trade-offs work out that its better to have a simpler constructor even if those rare users that want to separate hydrostatic pressure have to build `CenterField`.

Got it! Agreed. Should we test that `hydrostatic_pressure_anomaly` is either `CenterField(grid)` or `nothing` and throw a helpful error otherwise?

tomchor:
CC @zhazorken

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
glwagner: Yeah, before this PR, this line allowed users to (try) to put boundary conditions on pressure through the `boundary_conditions` kwarg. But after thinking about it, this can't have any effect because the velocity field is not prognostic on the boundaries. So it's useless...

This PR discontinuous the useless feature.

glwagner:
> > > This is looking pretty good! My only comment is that it's not intuitive for a given user to figure out how to opt for a hydrostatic pressure separation. A flag called `separate_hydrostatic_pressure` that takes `true`/`false` would be much more intuitive, although we'd need a little more code. Should we prioritize user-friendliness here?
> > 
> > 
> > This is more flexible, because sometimes users want to have access to the pressure field prior to model construction. If we use a flag, then we either can't support that or have to put some annoying logic in the constructor. Since I feel it'll be rare that people want to change this kwarg, I think the trade-offs work out that its better to have a simpler constructor even if those rare users that want to separate hydrostatic pressure have to build `CenterField`.
> 
> Got it! Agreed. Should we test that `hydrostatic_pressure_anomaly` is either `CenterField(grid)` or `nothing` and throw a helpful error otherwise?

Sure.

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -97,7 +98,12 @@ Keyword arguments
   - `particles`: Lagrangian particles to be advected with the flow. Default: `nothing`.
   - `biogeochemistry`: Biogeochemical model for `tracers`.
   - `velocities`: The model velocities. Default: `nothing`.
-  - `pressures`: Hydrostatic and non-hydrostatic pressure fields. Default: `nothing`.
+  - `nonhydrostatic_pressure`: The nonhydrostatic pressure field. Default: `CenterField(grid)`.
+  - `hydrostatic_pressure_anomaly`: An optional field that stores the part of the nonhydrostatic pressure
+                                    in hydrostatic balance with the buoyancy field. If `nothing` (default), the anomaly
+                                    is not computed. If `CenterField(grid)`, the anomaly is precomputed by
+                                    vertically-integrating the buoyancy field, and `nonhydrostatic_pressure` represents
+                                    only the part of pressure that deviates from the hydrostsatic anomaly.
```
navidcy, line 106:
```suggestion
                                    only the part of pressure that deviates from the hydrostatic anomaly.
```

src/Fields/field_tuples.jl
```diff
@@ -203,42 +203,6 @@ TracerFields(::Union{Tuple{}, Nothing}, grid, bcs) = NamedTuple()
 "Shortcut constructor for empty tracer fields."
 TracerFields(::NamedTuple{(), Tuple{}}, grid, bcs) = NamedTuple()
 
-#####
-##### Pressure fields tuples
-#####
```
navidcy, line 208: do we even use PressureFields anywhere now? should we delete in general? I can't see this being used anywhere but it's still being exported from Fields module and there are another 1-2 methods here and there...

navidcy (changes requested):

src/Fields/field_tuples.jl
glwagner: It was supposed to be deleted. Is this the only stray comment?

glwagner: Can you make these changes please instead of just requesting changes

navidcy: I was mostly asking for confirmation. I’ll deal with them don’t worry. 😉

navidcy: These lines are comments but there are PressureField related methods (not comments) elsewhere.

navidcy: Great. Done it!

navidcy (approved):
lgtm!

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -119,7 +125,8 @@ function NonhydrostaticModel(; grid,
             particles::ParticlesOrNothing = nothing,
     biogeochemistry::AbstractBGCOrNothing = nothing,
                                velocities = nothing,
-                                pressures = nothing,
+             hydrostatic_pressure_anomaly = nothing,
+                  nonhydrostatic_pressure = CenterField(grid),
```
tomchor, line 129: Just to follow the same order of args as in the docstrings:

```suggestion
                  nonhydrostatic_pressure = CenterField(grid),
             hydrostatic_pressure_anomaly = nothing,
```

tomchor: Should we just rename `hydrostatic_pressure_anomaly` to `hydrostatic_pressure` for simplicity?

glwagner: I'm open to that, let's discuss.

The possible confusion of that simplification is that there are other parts of the hydrostatic pressure (eg associated with reference density). The distinction between physical pressure and dynamic / modeled pressure is a persistent source of confusions so...

Even "hydrostatic" is a little confusing because this is, of course, dynamic.

glwagner: So sorry to be more specific: we call it the "anomaly" because the total hydrostatic pressure includes, for example, `rho0 * g * z` (there is also an additional component to the true pressure that we ignore, which is associated with compressibility effects when using a nonhydrostatic equation of state)

tomchor: Yeah, I agree with that. I suggested the change mainly because when we say `hydrostatic_pressure`, people will correctly assume what we're calculating there. I think when users see `hydrostatic_pressure_anomaly`, it won't be obvious to them what it is. But maybe that's for the best...?

I don't feel strongly about it though.

glwagner: You saying that if uses write

```julia
hydrostatic_pressure = CenterField(grid)
```

that it will be obvious that `hydrostatic_pressure` isn't the total hydrostatic pressure but rather just the anomaly associated with buoyancy forces?

navidcy:
@glwagner the distributed tests are broken :(

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
navidcy: I don’t really have a strong opinion. But personally with “hydrostatic” I would thing the ρ g z part was included

glwagner:
> @glwagner the distributed tests are broken :(

easy to fix

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
glwagner: I think we should keep anomaly for now since its consistent with the docs and whatnot. In the future we can consider revising the docs + this kwarg.

navidcy:
nice job!!

Issue #3460, opened 2024-02-06
Can we calculate a `Field` at every `N` grid points?


tomchor:
I'd like to compute a `Field` (ideally in order to write it to a NetCDF file) but only at every `N` grid points. Something like the following example, which tries to compute `u` at every 2 grid points in the vertical direction:

```julia
using Oceananigans
grid = RectilinearGrid(size = (1, 1, 8), extent = (1,1,1));
model = NonhydrostaticModel(; grid,)
u_slices = Field(model.velocities.u, indices=(:, :, 1:2:grid.Nz))
```

However, the above code fails since `Field` currently doesn't accept `StepRange`s as `indices`, just (I think) `UnitRange`s and `Int`s:

```
ERROR: LoadError: MethodError: no method matching isinteger(::StepRange{Int64, Int64})

Closest candidates are:
  isinteger(::Integer)
   @ Base number.jl:20
  isinteger(::Complex)
   @ Base complex.jl:148
  isinteger(::Rational)
   @ Base rational.jl:281
  ...

Stacktrace:
 [1] validate_index(idx::StepRange{Int64, Int64}, loc::Center, topo::Bounded, N::Int64, H::Int64)
   @ Oceananigans.Grids ~/repos/Oceananigans.jl/src/Grids/input_validation.jl:196
 [2] map (repeats 3 times)
   @ ./tuple.jl:318 [inlined]
 [3] validate_indices(indices::Tuple{Colon, Colon, StepRange{Int64, Int64}}, loc::Tuple{DataType, DataType, DataType}, topo::Tuple{DataType, DataType, DataType}, sz::Tuple{Int64, Int64, Int64}, halo_sz::Tuple{Int64, Int64, Int64})
```

Is there a workaround?

CC @iuryt

glwagner:
I think the best way to implement this is to write a function that returns a subsampled array, rather than trying to use the `Field` machinery.

If you would like to preserve grid information in the file automatically, you can generate the subsampled grid + subsampled `Field`, and then write a function that computes the subsampled `Field`.

Another way to implement this is to use `interpolate` or the conservative `regrid!` from the fine to coarse field.

I think the concept of a field with subsampled indices is a little tricky. For example, what should we do if we try to index that field in cells that lie between the existing data?

iuryt:
Hi!
Thanks for the explanation and reply.
Do you have any example that uses `interpolate` of `regrid!` I could use as a base for our case?

[...]

PR #3576, opened 2024-05-02
Passes `grid` argument to `NetCDFOutputWriter`

tomchor:
Related to https://github.com/CliMA/Oceananigans.jl/issues/3460

### DIFF
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -175,7 +175,7 @@ coarse_u = Field{Face, Center, Center}(coarse_grid)
 interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
 outputs = (; u = interpolate_u)
 
-output_writer = NetCDFOutputWriter(model, outputs, coarse_grid;
+output_writer = NetCDFOutputWriter(model, outputs; grid = coarse_grid;
                                    filename = "coarse_u.nc",
                                    schedule = IterationInterval(1))
 
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -163,38 +163,27 @@ To use this functionality, the `output_grid` must be passed explicitly when cons
 
 ```jldoctest
 using Oceananigans
+using Oceananigans.Fields: interpolate!
 
-grid = RectilinearGrid(size = (1, 1, 8), extent = (1,1,1));
-model = NonhydrostaticModel(; grid, closure = ScalarDiffusivity(ν=1e-2))
+grid = RectilinearGrid(size=(1, 1, 8), extent=(1, 1, 1));
+model = NonhydrostaticModel(; grid)
+simulation = Simulation(model, Δt=1, stop_iteration=1)
 
-set!(model, u=(x, y, z,) -> z)
-
-simulation = Simulation(model,
-                        Δt=0.5*maximum(zspacings(grid, Center())) / maximum(abs, model.velocities.u),
-                        stop_time=20)
-
-simulation.output_writers[:fullfields] = NetCDFOutputWriter(model, (; model.velocities.u),
-                                                            filename = "fullfields.nc",
-                                                            schedule = TimeInterval(5),
-                                                            overwrite_existing = true,)
-
-coarse_grid = RectilinearGrid(size = (grid.Nx, grid.Ny, grid.Nz÷2), extent = (grid.Lx, grid.Ly, grid.Lz))
+coarse_grid = RectilinearGrid(size=(grid.Nx, grid.Ny, grid.Nz÷2), extent=(grid.Lx, grid.Ly, grid.Lz))
 coarse_u = Field{Face, Center, Center}(coarse_grid)
 
-using Oceananigans.Fields: interpolate!
-update_coarse_u(simulation) = interpolate!(coarse_u, simulation.model.velocities.u)
-simulation.callbacks[:update_interp] = Callback(update_coarse_u)
+interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
+outputs = (; u = interpolate_u)
 
-simulation.output_writers[:coarsefields] = NetCDFOutputWriter(model, (; coarse_u,), coarse_grid;
-                                                              filename="coarsefields.nc",
-                                                              schedule=TimeInterval(5),
-                                                              overwrite_existing=true,)
+simulation.output_writers[:coarse_u] = NetCDFOutputWriter(model, outputs, coarse_grid;
+                                                          filename = "coarse_u.nc",
+                                                          schedule = IterationInterval(1))
 
 # output
-NetCDFOutputWriter scheduled on TimeInterval(5 seconds):
-├── filepath: ./coarsefields.nc
+NetCDFOutputWriter scheduled on IterationInterval(1):
+├── filepath: ./coarse_u.nc
 ├── dimensions: zC(4), zF(5), xC(1), yF(1), xF(1), yC(1), time(0)
-├── 1 outputs: coarse_u
+├── 1 outputs: u
 └── array type: Array{Float64}
 ├── file_splitting: NoFileSplitting
 └── file size: 14.6 KiB
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -175,7 +175,8 @@ coarse_u = Field{Face, Center, Center}(coarse_grid)
 interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
 outputs = (; u = interpolate_u)
 
-output_writer = NetCDFOutputWriter(model, outputs; grid = coarse_grid;
+output_writer = NetCDFOutputWriter(model, outputs;
+                                   grid = coarse_grid,
                                    filename = "coarse_u.nc",
                                    schedule = IterationInterval(1))
 
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -158,8 +158,9 @@ NetCDFOutputWriter scheduled on IterationInterval(1):
 └── file size: 17.8 KiB
 ```
 
-`NetCDFOutputWriter` can also be configured for `outputs` that are interpolated or regridded to a different grid than `model.grid`.
-To use this functionality, the `output_grid` must be passed explicitly when constructing `NetCDFOutputWriter` along with the regridded / interpolated `outputs`.
+`NetCDFOutputWriter` can also be configured for `outputs` that are interpolated or regridded
+to a different grid than `model.grid`. To use this functionality, include the keyword argument
+`grid = output_grid`.
 
 ```jldoctest
 using Oceananigans
@@ -312,3 +313,4 @@ JLD2OutputWriter scheduled on TimeInterval(4 days):
 ├── file_splitting: NoFileSplitting
 └── file size: 26.5 KiB
 ```
+
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -167,7 +167,6 @@ using Oceananigans.Fields: interpolate!
 
 grid = RectilinearGrid(size=(1, 1, 8), extent=(1, 1, 1));
 model = NonhydrostaticModel(; grid)
-simulation = Simulation(model, Δt=1, stop_iteration=1)
 
 coarse_grid = RectilinearGrid(size=(grid.Nx, grid.Ny, grid.Nz÷2), extent=(grid.Lx, grid.Ly, grid.Lz))
 coarse_u = Field{Face, Center, Center}(coarse_grid)
@@ -175,9 +174,9 @@ coarse_u = Field{Face, Center, Center}(coarse_grid)
 interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
 outputs = (; u = interpolate_u)
 
-simulation.output_writers[:coarse_u] = NetCDFOutputWriter(model, outputs, coarse_grid;
-                                                          filename = "coarse_u.nc",
-                                                          schedule = IterationInterval(1))
+output_writer = NetCDFOutputWriter(model, outputs, coarse_grid;
+                                    filename = "coarse_u.nc",
+                                    schedule = IterationInterval(1))
 
 # output
 NetCDFOutputWriter scheduled on IterationInterval(1):
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -122,7 +122,7 @@ simulation = Simulation(model, Δt=1.25, stop_iteration=3)
 
 f(model) = model.clock.time^2; # scalar output
 
-g(model) = model.clock.time .* exp.(znodes(Center, grid)) # single-column profile output (vector)
+g(model) = model.clock.time .* exp.(znodes(grid, Center())) # single-column profile output (vector)
 
 xC, yF = xnodes(grid, Center()), ynodes(grid, Face())
 
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -158,8 +158,46 @@ NetCDFOutputWriter scheduled on IterationInterval(1):
 └── file size: 17.8 KiB
 ```
 
-`NetCDFOutputWriter` also accepts an arbitrary `grid` for input that can be used to interpolate or subset the output fields. The output `grid` must be in the same architecture as the original grid.
+`NetCDFOutputWriter` can also be configured for `outputs` that are interpolated or regridded to a different grid than `model.grid`. To use this functionality, the `output_grid` must be passed explicitly when constructing `NetCDFOutputWriter` along with the regridded / interpolated `outputs`.
 
+```jldoctest
+using Oceananigans
+
+grid = RectilinearGrid(size = (1, 1, 8), extent = (1,1,1));
+model = NonhydrostaticModel(; grid, closure = ScalarDiffusivity(ν=1e-2))
+
+set!(model, u=(x, y, z,) -> z)
+
+simulation = Simulation(model,
+                        Δt=0.5*maximum(zspacings(grid, Center())) / maximum(abs, model.velocities.u),
+                        stop_time=20)
+
+simulation.output_writers[:fullfields] = NetCDFOutputWriter(model, (; model.velocities.u),
+                                                            filename = "fullfields.nc",
+                                                            schedule = TimeInterval(5),
+                                                            overwrite_existing = true,)
+
+coarse_grid = RectilinearGrid(size = (grid.Nx, grid.Ny, grid.Nz÷2), extent = (grid.Lx, grid.Ly, grid.Lz))
+coarse_u = Field{Face, Center, Center}(coarse_grid)
+
+using Oceananigans.Fields: interpolate!
+update_coarse_u(simulation) = interpolate!(coarse_u, simulation.model.velocities.u)
+simulation.callbacks[:update_interp] = Callback(update_coarse_u)
+
+simulation.output_writers[:coarsefields] = NetCDFOutputWriter(model, (; coarse_u,), coarse_grid;
+                                                              filename="coarsefields.nc",
+                                                              schedule=TimeInterval(5),
+                                                              overwrite_existing=true,)
+
+# output
+NetCDFOutputWriter scheduled on TimeInterval(5 seconds):
+├── filepath: ./coarsefields.nc
+├── dimensions: zC(4), zF(5), xC(1), yF(1), xF(1), yC(1), time(0)
+├── 1 outputs: coarse_u
+└── array type: Array{Float64}
+├── file_splitting: NoFileSplitting
+└── file size: 14.6 KiB
+```
 
 See [`NetCDFOutputWriter`](@ref) for more information.
 
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -175,8 +175,8 @@ interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
 outputs = (; u = interpolate_u)
 
 output_writer = NetCDFOutputWriter(model, outputs, coarse_grid;
-                                    filename = "coarse_u.nc",
-                                    schedule = IterationInterval(1))
+                                   filename = "coarse_u.nc",
+                                   schedule = IterationInterval(1))
 
 # output
 NetCDFOutputWriter scheduled on IterationInterval(1):
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -158,6 +158,9 @@ NetCDFOutputWriter scheduled on IterationInterval(1):
 └── file size: 17.8 KiB
 ```
 
+`NetCDFOutputWriter` also accepts an arbitrary `grid` for input that can be used to interpolate or subset the output fields. The output `grid` must be in the same architecture as the original grid.
+
+
 See [`NetCDFOutputWriter`](@ref) for more information.
 
 ## JLD2 output writer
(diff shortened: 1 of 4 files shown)


tomchor:
Unfortunately this example

```julia
using Oceananigans

grid = RectilinearGrid(size = (1, 1, 8), extent = (1,1,1));
model = NonhydrostaticModel(; grid, closure = ScalarDiffusivity(ν=1e-2))

set!(model, u=(x, y, z,) -> z)

simulation = Simulation(model,
                        Δt=0.5*maximum(zspacings(grid, Center())) / maximum(abs, model.velocities.u),
                        stop_time=20)
# Create regular output
simulation.output_writers[:fullfields] = NetCDFOutputWriter(model, (; model.velocities.u),
                                                            filename = "fullfields.nc",
                                                            schedule = TimeInterval(5),
                                                            overwrite_existing = true,)

# Create interpolated u on coarse grid
coarse_grid = RectilinearGrid(size = (grid.Nx, grid.Ny, grid.Nz÷2), extent = (grid.Lx, grid.Ly, grid.Lz))
coarse_u = Field{Face, Center, Center}(coarse_grid)

using Oceananigans.Fields: interpolate!
update_coarse_u(simulation) = interpolate!(coarse_u, simulation.model.velocities.u)
simulation.callbacks[:update_interp] = Callback(update_coarse_u)

# Create coarse output
simulation.output_writers[:coarsefields] = NetCDFOutputWriter(model, (; coarse_u,), coarse_grid;
                                                              filename="coarsefields.nc",
                                                              schedule=TimeInterval(5),
                                                              overwrite_existing=true,)

run!(simulation)
```

Throws the following error:

```
ERROR: LoadError: DimensionMismatch: new dimensions (1, 1, 8, 1) must be consistent with array size 4
Stacktrace:
  [1] (::Base.var"#throw_dmrsa#328")(dims::NTuple{4, Int64}, len::Int64)
    @ Base ./reshapedarray.jl:41
  [2] reshape(a::Array{Float64, 3}, dims::NTuple{4, Int64})
    @ Base ./reshapedarray.jl:45
  [3] setindex_disk!(::NCDatasets.Variable{Float64, 4, NCDatasets.NCDataset{Nothing, Missing}}, ::Array{Float64, 3}, ::Function, ::Vararg{Any})
    @ DiskArrays ~/.julia/packages/DiskArrays/bZBJE/src/diskarray.jl:56
  [4] setindex!
    @ ~/.julia/packages/DiskArrays/bZBJE/src/diskarray.jl:229 [inlined]
  [5] setindex!(::CommonDataModel.CFVariable{…}, ::Array{…}, ::Colon, ::Colon, ::Colon, ::UnitRange{…})
    @ CommonDataModel ~/.julia/packages/CommonDataModel/GGvMn/src/cfvariable.jl:477
  [6] save_output!(ds::NCDatasets.NCDataset{…}, output::Field{…}, model::NonhydrostaticModel{…}, ow::NetCDFOutputWriter{…}, time_index::Int64, name::String)
    @ Oceananigans.OutputWriters ~/repos/Oceananigans.jl/src/OutputWriters/netcdf_output_writer.jl:482
  [7] write_output!(ow::NetCDFOutputWriter{…}, model::NonhydrostaticModel{…})
    @ Oceananigans.OutputWriters ~/repos/Oceananigans.jl/src/OutputWriters/netcdf_output_writer.jl:525
  [8] initialize!(sim::Simulation{…})
    @ Oceananigans.Simulations ~/repos/Oceananigans.jl/src/Simulations/run.jl:212
  [9] time_step!(sim::Simulation{…})
    @ Oceananigans.Simulations ~/repos/Oceananigans.jl/src/Simulations/run.jl:112
 [10] run!(sim::Simulation{…}; pickup::Bool)
    @ Oceananigans.Simulations ~/repos/Oceananigans.jl/src/Simulations/run.jl:97
 [11] run!(sim::Simulation{NonhydrostaticModel{…}, Float64, Float64, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}})
    @ Oceananigans.Simulations ~/repos/Oceananigans.jl/src/Simulations/run.jl:85
 [12] top-level scope
    @ ~/repos/Oceananigans.jl/sandbox/mwe.jl:31
 [13] include(fname::String)
    @ Base.MainInclude ./client.jl:489
```

So it's not as trivial as the single change I just made. From glancing at the code we at least have to modify `initialize_nc_file!()` to take a `grid` option as well:

https://github.com/CliMA/Oceananigans.jl/blob/6730e6f6b2c8f1695e20b95ef467b5b14fdc4c5f/src/OutputWriters/netcdf_output_writer.jl#L625-L636

plus a couple of other things. Still pretty easy, but more work/time that I have right now.

@iuryt feel free to jump in here and make these changes if you feel it's necessary, since creating a whole separate model can be a bit onerous and wastes precious GPU memory.

glwagner:
> Still pretty easy, but more work/time that I have right now.

Reading the code this is just one more line. Can you be more specific about how much more time this will take?

glwagner:
Ok, the above example works by adding `grid` as a property to the output writer, and then passing `grid` explicitly to the initialization.

Making this change required a couple minutes. The main technique was to search the file for `model.grid`, and fix that. I noticed that initialization has to occur outside the output writer constructor, which means that the "output writer grid" (which is now different from the model grid) must be stored within the output writer.

Any other changes needed?

tomchor:
> > Still pretty easy, but more work/time that I have right now.
> 
> Reading the code this is just one more line. Can you be more specific about how much more time this will take?

I wasn't specific about time because I wasn't sure. I often find myself spending way longer on PRs than I initially anticipate so I'm generally not very good at assessing these things lol

But I'm glad it's working now, pending docs. If the example above works without needing to create a second `model`, then I don't think anything else is needed on the coding part. @iuryt ?

glwagner:
> I wasn't specific about time because I wasn't sure. I often find myself spending way longer on PRs than I initially anticipate so I'm generally not very good at assessing these things lol

ok ok! Here's a rule of thumb: a bugfix or refactor is the least committment, because you can get away with no new tests or docs.

A new _feature_ takes more time because of documentation. One should expect to spend most of their time on docs and tests. If you're spending a lot of time implementing a feature, then either the feature is very complicated / hard, or your workflow can use some improvement.

Among new features, different types of features require different amount of effort. This case is one of the easier cases --- it's extending functionality without changing existing functionality, also its fairly niche (at this point) so simple documentation will suffice. The work required to test the feature also has already been partially completed (the script you posted).

Other features, like adding new _physics_ will take much longer, because often you'll have to run a full validation case + analysis to assess whether things work as expected. So even if the source code change is small to implement new physics, the validation will take a while and dominate the development effort.

If a new feature also requires changing existing functionality / refactoring, that's going to take the most amount of time, because you will probably also have to change existing tests. And many tests are poorly written, so updating test code is a potential rabbit hole.

glwagner:
Tests pass, so documentation in your courts @iuryt @tomchor. I think for the docstring we can just say that the `grid` needs to be provided if `outputs` are on a different `grid` than `model`.

But it could be nice to also put an example in the docs, since I think the whole idea of output on a different grid from the model takes a bit to explain. Eg here: https://clima.github.io/OceananigansDocumentation/stable/model_setup/output_writers/

I can contribute a test if there's desire to put in the documentation.

iuryt:
I can work on the docs

docs/src/model_setup/output_writers.md
```diff
@@ -158,6 +158,9 @@ NetCDFOutputWriter scheduled on IterationInterval(1):
 └── file size: 17.8 KiB
 ```
 
+`NetCDFOutputWriter` also accepts an arbitrary `grid` for input that can be used to interpolate or subset the output fields. The output `grid` must be in the same architecture as the original grid.
```
glwagner, line 161: This isn't quite right.

The `grid` _must_ be consistent with output. So it's misleading to say that the output writer "accepts" an arbitrary grid. The grid is not arbitrary at all, its the opposite: the `grid` has to be consistent with `outputs`.

The right statement is something like "NetCDFOutputWriter" can also be configured for `outputs` that are interpolated or regridded to a different grid than `model.grid`. To use this functionalitt, the `output_grid` must be passed explicitly when constructing `NetCDFOutputWriter` along with the regridded / interpolated `outputs`.

glwagner: Someday this statement could be true if the `grid` is actually used to generate automatically interpolated outputs 😄 by the way (but this PR doesn't go so far)

docs/src/model_setup/output_writers.md
```diff
@@ -158,8 +158,46 @@ NetCDFOutputWriter scheduled on IterationInterval(1):
 └── file size: 17.8 KiB
 ```
 
-`NetCDFOutputWriter` also accepts an arbitrary `grid` for input that can be used to interpolate or subset the output fields. The output `grid` must be in the same architecture as the original grid.
+`NetCDFOutputWriter` can also be configured for `outputs` that are interpolated or regridded to a different grid than `model.grid`. To use this functionality, the `output_grid` must be passed explicitly when constructing `NetCDFOutputWriter` along with the regridded / interpolated `outputs`.
```
glwagner, line 161:
```suggestion
`NetCDFOutputWriter` can also be configured for `outputs` that are interpolated or regridded to a different grid than `model.grid`.
To use this functionality, the `output_grid` must be passed explicitly when constructing `NetCDFOutputWriter` along with the regridded / interpolated `outputs`.
```

iuryt:
Once the wording is finalized, I will edit the doctring.

docs/src/model_setup/output_writers.md
```diff
@@ -158,8 +158,46 @@ NetCDFOutputWriter scheduled on IterationInterval(1):
...
+coarse_u = Field{Face, Center, Center}(coarse_grid)
+
+using Oceananigans.Fields: interpolate!
+update_coarse_u(simulation) = interpolate!(coarse_u, simulation.model.velocities.u)
+simulation.callbacks[:update_interp] = Callback(update_coarse_u)
+
+simulation.output_writers[:coarsefields] = NetCDFOutputWriter(model, (; coarse_u,), coarse_grid;
+                                                              filename="coarsefields.nc",
+                                                              schedule=TimeInterval(5),
+                                                              overwrite_existing=true,)
```
glwagner, line 191: Nice work! Here's a few suggestions:

* don't include outputting the original fields. We have other docstrings for that.
* Make the example simpler. Don't set initial conditions, don't compute `dt`, don't put in a random stop time like "20". `stop_iteration=1` is fine.
* I recomend an output named tuple like `(; u = coarse_u)`. In other words, the user wil lbe able to tell from the filename and grid that the output is coarse. We don't need to additionally name the output "`coarse_u`".
* Make sure to use correct style with spaces after `=` etc

It will make the script shorter to write

```julia
interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
outputs = (; u = interpolate_u)
```

in other words, make the output a function itself. Then you don't have to use a callback to compute the output.

iuryt: Thanks! about the spaces. I usually do that, but got confused because there are no spaces after `=` for the other examples in the docs. Should I take the time and correct the other examples?

iuryt: @glwagner 
If I try what you suggested and use the function directly in the `outputs`, it returns
```julia
Custom output u needs dimensions!
```

glwagner: I'll fix that.

glwagner: Okay put in a feature that helps with user functions that return fields (the same feature already existed in JLD2OutputWriter after a fact, and I didn't realize it was missing for netcdf)

glwagner: see if you get the same error now

iuryt: I still got the same error. Is that anything I could share to help u debug it?

glwagner: Yes, a MWE would help.

iuryt: It is exactly the same code from the example.
Let me know if I am doing something wrong. What I am doing is, in a test directory, I am running

```bash
rm -f *.toml && touch Project.toml && julia --project -e 'using Pkg; Pkg.add(url = "https://github.com/CliMA/Oceananigans.jl/#tc/grid-to-netcdfwriter")'
```
(I also tried to run with `Pkg.add(path = "../Oceananigans.jl")` , which is the directory I have the code at the current branch)

Then 

```bash
julia --project code.jl
```

`code.jl` contains the following code

```julia
using Oceananigans

grid = RectilinearGrid(size = (1, 1, 8), extent = (1, 1, 1));
model = NonhydrostaticModel(; grid, closure = ScalarDiffusivity(ν = 1e-2))

set!(model, u = (x, y, z,) -> z)

simulation = Simulation(model, Δt = 1, stop_iteration = 1)

coarse_grid = RectilinearGrid(size = (grid.Nx, grid.Ny, grid.Nz÷2), extent = (grid.Lx, grid.Ly, grid.Lz))
coarse_u = Field{Face, Center, Center}(coarse_grid)

using Oceananigans.Fields: interpolate!

# # this works
# update_coarse_u(simulation) = interpolate!(coarse_u, simulation.model.velocities.u)
# simulation.callbacks[:update_interp] = Callback(update_coarse_u)

# outputs = (; u = coarse_u,)

# this doesn't work
interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
outputs = (; u = interpolate_u)

simulation.output_writers[:coarse_u] = NetCDFOutputWriter(model, outputs, coarse_grid;
                                                              filename = "coarse_u.nc",
                                                              schedule = TimeInterval(5),
                                                              overwrite_existing = true,)
```

which returns 

```julia
Custom output u needs dimensions!

Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] define_output_variable!(dataset::NCDatasets.NCDataset{Nothing, Missing}, output::Nothing, name::String, array_type::Type, deflatelevel::Int64, output_attributes::Dict{String, String}, dimensions::Dict{Any, Any})
   @ Oceananigans.OutputWriters ~/.julia/packages/Oceananigans/r2jWy/src/OutputWriters/netcdf_output_writer.jl:460
 [3] initialize_nc_file!(filepath::String, outputs::Dict{String, typeof(interpolate_u)}, .... .... 
```

glwagner: Ok fixed the issue (it was with `interpolate!`. Also I improved the readability of output writer code. I don't think custom output is really properly supported right now. There's some kind of assumption about the array type for custom output taking place which would prevent people from doing non-array output.

Here's a more minimal example that you can use in the docstring:

```julia
using Oceananigans
using Oceananigans.Fields: interpolate!

grid = RectilinearGrid(size=(1, 1, 8), extent=(1, 1, 1));
simulation = Simulation(model, Δt=1, stop_iteration=1)

coarse_grid = RectilinearGrid(size=(grid.Nx, grid.Ny, grid.Nz÷2), extent=(grid.Lx, grid.Ly, grid.Lz))
coarse_u = Field{Face, Center, Center}(coarse_grid)

interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
outputs = (; u = interpolate_u)

rm("coarse_u.nc", force=true)

simulation.output_writers[:coarse_u] = NetCDFOutputWriter(model, outputs, coarse_grid;
                                                          filename = "coarse_u.nc",
                                                          schedule = IterationInterval(1))
```

I got rid of everything that wasn't strictly necessary for the example. The problem with including "random" extra stuff like a closure to the model is that it can be confusing for a new user, whether or not those things are crucial to the example or not.

docs/src/model_setup/output_writers.md
```diff
@@ -163,38 +163,27 @@ To use this functionality, the `output_grid` must be passed explicitly when cons
...
+interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
+outputs = (; u = interpolate_u)
 
-simulation.output_writers[:coarsefields] = NetCDFOutputWriter(model, (; coarse_u,), coarse_grid;
-                                                              filename="coarsefields.nc",
-                                                              schedule=TimeInterval(5),
-                                                              overwrite_existing=true,)
+simulation.output_writers[:coarse_u] = NetCDFOutputWriter(model, outputs, coarse_grid;
+                                                          filename = "coarse_u.nc",
+                                                          schedule = IterationInterval(1))
```
glwagner, line 180: We don't need `Simulation` for this example right?

iuryt: hm... yes, but all other previous examples for the same docs is defining like this.

glwagner: Sometimes, we must leave the past behind for a better, brighter future.

iuryt: Agreed! But we also need to be consistent, right? Or not really for that case?

iuryt: What about this?

```julia
using Oceananigans
using Oceananigans.Fields: interpolate!

grid = RectilinearGrid(size=(1, 1, 8), extent=(1, 1, 1));

coarse_grid = RectilinearGrid(size=(grid.Nx, grid.Ny, grid.Nz÷2), extent=(grid.Lx, grid.Ly, grid.Lz))
coarse_u = Field{Face, Center, Center}(coarse_grid)

interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
outputs = (; u = interpolate_u)

output_writer = NetCDFOutputWriter(model, outputs, coarse_grid;
                                    filename = "coarse_u.nc",
                                    schedule = IterationInterval(1))
```

glwagner: > Agreed! But we also need to be consistent, right? Or not really for that case?

I think its more important to innovate, we can change the other examples if we come up with a better / more concise pattern.

iuryt: Yes, I think that there is a balance between conciseness and usability that we need to be at. I kind of see the value of having a simulation because this is inevitably most of the cases when you need to create a writer, but at the same time, we want the documentation to be simple and not being used as simply as copy and paste. What do you think about that?

iuryt: I simplified this example, but not the other examples as they explicitly mention the file saving.

docs/src/model_setup/output_writers.md
```diff
@@ -167,17 +167,16 @@ using Oceananigans.Fields: interpolate!
...
 
 interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
 outputs = (; u = interpolate_u)
 
-simulation.output_writers[:coarse_u] = NetCDFOutputWriter(model, outputs, coarse_grid;
-                                                          filename = "coarse_u.nc",
-                                                          schedule = IterationInterval(1))
+output_writer = NetCDFOutputWriter(model, outputs, coarse_grid;
+                                    filename = "coarse_u.nc",
+                                    schedule = IterationInterval(1))
```
glwagner, line 179:
```suggestion
output_writer = NetCDFOutputWriter(model, outputs, coarse_grid;
                                   filename = "coarse_u.nc",
                                   schedule = IterationInterval(1))
```

docs/src/model_setup/output_writers.md
glwagner: sounds good

docs/src/model_setup/output_writers.md
```diff
@@ -158,6 +158,36 @@ NetCDFOutputWriter scheduled on IterationInterval(1):
 └── file size: 17.8 KiB
 ```
 
+`NetCDFOutputWriter` can also be configured for `outputs` that are interpolated or regridded to a different grid than `model.grid`.
+To use this functionality, the `output_grid` must be passed explicitly when constructing `NetCDFOutputWriter` along with the regridded / interpolated `outputs`.
```
tomchor, line 162: Should we mention that once `NetCDFWriter` has `output_grid` passed it can _only_ write outputs on that grid. In other words, you can't have outputs that are the original/full grid and the interpolated/coarse grid in the same `NetCDFWriter` file (for now at least). I feel this can be a source of confusion for users in the future.

docs/src/model_setup/output_writers.md
```diff
@@ -158,6 +158,36 @@ NetCDFOutputWriter scheduled on IterationInterval(1):
...
+
+coarse_grid = RectilinearGrid(size=(grid.Nx, grid.Ny, grid.Nz÷2), extent=(grid.Lx, grid.Ly, grid.Lz))
+coarse_u = Field{Face, Center, Center}(coarse_grid)
+
+interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
+outputs = (; u = interpolate_u)
+
+output_writer = NetCDFOutputWriter(model, outputs, coarse_grid;
+                                   filename = "coarse_u.nc",
+                                   schedule = IterationInterval(1))
```
tomchor, line 179: Here `u = interpolate_u` is a function. Isn't this failing without passing dimensions for the output? That's why my original example output a `Field` directly and updated it with a `Callback`.

tomchor: Ah, nvmd. I just saw that @glwagner added new functionality that infers dimensions from functions that output `Field`s. Nice!

docs/src/model_setup/output_writers.md
tomchor: Also we don't pass `output_grid`. We pass the keyword `grid`, correct?

glwagner: > Also we don't pass output_grid. We pass the keyword grid, correct?

We made it a positional argument. But we can change that and make a keyword argument for sure. What do you think?

Also, I agree 100% that making that more explicit is a good idea. The `grid` and outputs _must_ be consistent, otherwise output writing will fail.

Also, you often write `NetCDFWriter`. I like that name because it's a little more concise ("output" is obvious). Should we change this and `JLD2`?

iuryt: I agree that `NetCDFWriter` is better and that we should also change that for JLD.

glwagner: Ok, after this is merged we will get on that.

tomchor: > > Also we don't pass output_grid. We pass the keyword grid, correct?
> 
> We made it a positional argument. But we can change that and make a keyword argument for sure. What do you think?

Ah, true. I have a bit of a preference for keyword arguments, but I don't feel strongly about it.

> Also, I agree 100% that making that more explicit is a good idea. The `grid` and outputs _must_ be consistent, otherwise output writing will fail.
> 
> Also, you often write `NetCDFWriter`. I like that name because it's a little more concise ("output" is obvious). Should we change this and `JLD2`?

Ah, yeah, sorry. I write that for the exact reason you mentioned, and the fact that `NetCDFOutputWriter` is kinda long lol

I often thought abut opening an issue to suggest that change myself, but haven't gotten around to it. So I'm very much in favor of making that change!

glwagner: Hmm ok I agree it should be a kwarg.

glwagner: I noticed that the docs examples are identical to the doc string examples. Ideally these would be different... there's no point in copy pasting here. It's more efficient to use Documenter to insert the docstring.

iuryt:
Let me know what else is missing or if we just need a third review to merge.

tomchor:
> Let me know what else is missing or if we just need a third review to merge.

Tests still aren't passing.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -163,28 +164,34 @@ end
 
 """
     NetCDFOutputWriter(model, outputs; filename, schedule
```
tomchor, line 166:
```suggestion
    NetCDFOutputWriter(model, outputs; filename, schedule,
```

glwagner:
Some commits got added that overwrote changes that I made

iuryt:
> Hmm ok somehow changes I made were lost, but I can't figure out how. Anyways I have tried to put them back in.

I was wondering the same. I noticed the error was the same we had before.

iuryt:
Not sure why this is failing now.

glwagner:
Might just need a retry -- gave it one. Thanks for the help!

glwagner:
heck ya

PR #3580, opened 2024-05-05
(0.92.0) Make RungeKutta3 the default time-stepper for NonhydrostaticModel

glwagner:
Who's with me?


test/test_boundary_conditions_integration.jl
```diff
@@ -13,7 +13,7 @@ function test_boundary_condition(arch, FT, topo, side, field_name, boundary_cond
                                 buoyancy=SeawaterBuoyancy(), tracers=(:T, :S))
 
     success = try
-        time_step!(model, 1e-16, euler=true)
```
navidcy, line 16: why `euler=true` is gone? or why was it there?

navidcy (approved):
lgtm!

I'm with you!

test/test_boundary_conditions_integration.jl
glwagner: AB2 has memory (one reason why I don't like it). If you don't know the status of the tendencies (ie if you cannot guarantee that they were computed consistently at the previous time step), then you need to write `euler=true` to use `time_step!(model)`. For most users running a simulation, this is not an issue because `run!` handles it. In typical circumstances, we can tell when somebody is taking the first time step (we check the previous time step size now; previously we used `clock.iteration`).

But this PR changes the default to RK3. RK3 doesn't have an `euler` option for `time_step`.

navidcy:
simulation tests take forever? I'll restart them...

glwagner:
> simulation tests take forever? I'll restart them...

Huh, I wonder if that's a downside of this change but not sure...

navidcy:
No, on main all tests are done in O(1h)… :(
But it’s strange.

glwagner:
> No, on main all tests are done in O(1h)… :( But it’s strange.

Simulation tests take 14 min normally:

https://buildkite.com/clima/oceananigans/builds/15704#018f5a18-1775-4a66-b63d-d908d95f9713

something fishy

navidcy:
hm... the distributed tests are actually failing; they were passing after [1936d62](https://github.com/CliMA/Oceananigans.jl/pull/3580/commits/1936d629224951de2c7845b0eeaffbbba58dc8ff) and I don't see anything that would justify this in the commits that followed..

glwagner:
I believe there's some issue with initialization. It doesn't help to rebuild the jobs, because the initialization only happens once and the results are cached are re-used for later builds.

navidcy:
They passed now :)

navidcy:
This is failing now:
https://buildkite.com/clima/oceananigans/builds/17582#01922b77-d047-4e99-a497-5facd7241c56/19-1087

glwagner:
exit status -1 means the agent was lost :-(

glwagner:
> This is failing now: https://buildkite.com/clima/oceananigans/builds/17582#01922b77-d047-4e99-a497-5facd7241c56/19-1087

oh yea this is something to fix

navidcy:
Merge?

navidcy:
I think we should before conflicts arise again ;)

glwagner:
Ahh, we keep prioritizing other PRs!

glwagner:
Let's try to prioritize this one and get it in

Issue #3568, opened 2024-04-29
Boundary conditions and diffusion with background fields


liuchihl:
I have modified the [simple 1D diffusion example](https://clima.github.io/OceananigansDocumentation/v0.90.13/literated/one_dimensional_diffusion/) to test how to correct the boundary conditions by using the gradient boundary condition, the script can be found [here](https://github.com/liuchihl/Simple-tests/blob/85f34ec2443a6ff9860f5adfbc68ff39995f00bb/1Ddiffusion_testBCflux.jl). 

In the experiment, I have specified both the initial temperature (perturbation) and a constant background temperature < T >, so that the total temperature T_total = < T > + T.
 
The movie below shows T_total and presents two simple experiments with an existing initial temperature. In the blue case, the default no-flux boundary condition is applied (i.e., no gradient in the wall-normal direction), but the gradient at the boundaries is nonzero because the boundary condition does not account for the background temperature. In contrast, the red case includes a boundary condition that forces the gradient of T_total to be 0. 
With this corrected boundary condition (red curve), the flux at the boundaries is 0, which is physical. However, it remains unclear how to incorporate these corrected fluxes for more complicated configurations.

https://github.com/CliMA/Oceananigans.jl/assets/68127124/69e64cf9-6248-4274-9b6d-5f763827e768

The movie below shows two additional cases without an initial temperature, indicating that the background temperature, < T > defines the entire field. 
The constant blue line throughout the simulation implies that the background scalar does not diffuse either within the domain or at the boundaries.  However, in the red case, despite the absence of diffusion affecting the background temperature, the nonzero flux at the boundaries causes the curve to become smoothed. 
If diffusion does not affect the background field, would it still make sense if the stratification is not a constant, such as in a Kelvin-Helmholtz instability configuration, e.g., < b > = tanh(z)?

https://github.com/CliMA/Oceananigans.jl/assets/68127124/d350c5ad-2e14-4992-8c5d-f947a1ddf7bb

This brings up the question: what should the default behavior for background fields be? Should they really only appear in the advective terms, or should they also be diffused? And what should the boundary conditions on perturbations be when there are background fields?

glwagner:
The reason that the advection term is the only term account for is because this is typically what one wants.

[...]

PR #3581, opened 2024-05-05
Enforce total buoyancy flux BC in tilted geometry example

hdrake:
This example includes no explicit modification to the BCs on buoyancy, meaning that it defaults to a no-flux BC (or `FluxBoundaryCondition()`) on buoyancy.

Following equation (6) of [Wenegrat and Thomas (2020)](https://journals.ametsoc.org/view/journals/phoc/50/6/JPO-D-20-0027.1.xml), we instead enforce a no-normal diffusive flux boundary condition on the *total* buoyancy, i.e. the perturbation plus the constantly-stratified `BackgroundField`. Because Oceananigans.jl does not allow diffusion to act on `BackgroundField`s, the background part of the no-flux BC is missing. However, we here can enforce it by specifying a perturbation flux BC that matches the implied background flux.

Other minor changes:
- Added equally-spaced buoyancy surfaces to movie panels
- Guess timestep by minimum of advective and diffusive timescales, instead of just the advective timescale

This addresses the problem with the `tilted_bottom_boundary_layer.jl` example identified by @liuchihl and noted in https://github.com/CliMA/Oceananigans.jl/issues/3568. Recommend closing that issue but perhaps continuing some of the discussion in a "Discussion" post.

### DIFF
--- a/examples/tilted_bottom_boundary_layer.jl
+++ b/examples/tilted_bottom_boundary_layer.jl
@@ -103,9 +103,7 @@ B_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = N²)
 # ```math
 # ∂_z B = ∂_z b + N^{2} \cos{\theta} = 0.
 # ```
-# The default `FluxBoundaryCondition()` (which imposes ``∂_z b = 0``) is inappropriate;
-# instead, we must rearrange this expression to override the boundary condition on the
-# perturbation flux:
+# This shows that to impose a no-flux boundary condition on the total buoyancy field ``B``, we must apply a boundary condition to the perturbation buoyancy ``b``,
 # ```math
 # ∂_z b = - N^{2} \cos{\theta}.
 #```
--- a/examples/tilted_bottom_boundary_layer.jl
+++ b/examples/tilted_bottom_boundary_layer.jl
@@ -150,7 +150,7 @@ model = NonhydrostaticModel(; grid, buoyancy, coriolis, closure,
                             advection = UpwindBiasedFifthOrder(),
                             tracers = :b,
                             boundary_conditions = (u=u_bcs, v=v_bcs, b=b_bcs),
-                            background_fields = (; b=B_field))
+                            background_fields = (; b=B∞_field))
 
 # Let's introduce a bit of random noise at the bottom of the domain to speed up the onset of
 # turbulence:
--- a/examples/tilted_bottom_boundary_layer.jl
+++ b/examples/tilted_bottom_boundary_layer.jl
@@ -95,9 +95,18 @@ coriolis = ConstantCartesianCoriolis(f = 1e-4, rotation_axis = ĝ)
 # _perturbations_ away from the constant density stratification by imposing
 # a constant stratification as a `BackgroundField`,
 
-B_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = 1e-5))
+N² = 1e-5 # s⁻¹ # background stratification
+B_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = N²))
 
 # where ``N^2 = 10^{-5} \rm{s}^{-2}`` is the background buoyancy gradient.
+
+# Because the bottom boundary condition is that there must be zero *total* diffusive
+# flux across the seafloor, the default `FluxBoundaryCondition()` is insufficient;
+# instead, the boundary condition on the perturbation flux must take the background
+# into account,
+
+negative_background_diffusive_flux = GradientBoundaryCondition(-N²*ĝ[3])
+b_bcs = FieldBoundaryConditions(bottom = negative_background_diffusive_flux)
 
 # ## Bottom drag
 #
@@ -128,13 +137,15 @@ v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
 # and a constant viscosity and diffusivity. Here we use a smallish value
 # of ``10^{-4} \, \rm{m}^2\, \rm{s}^{-1}``.
 
-closure = ScalarDiffusivity(ν=1e-4, κ=1e-4)
+ν=1e-4
+κ=1e-4
+closure = ScalarDiffusivity(ν=ν, κ=κ)
 
 model = NonhydrostaticModel(; grid, buoyancy, coriolis, closure,
                             timestepper = :RungeKutta3,
                             advection = UpwindBiasedFifthOrder(),
                             tracers = :b,
-                            boundary_conditions = (u=u_bcs, v=v_bcs),
+                            boundary_conditions = (u=u_bcs, v=v_bcs, b=b_bcs),
                             background_fields = (; b=B_field))
 
 # Let's introduce a bit of random noise at the bottom of the domain to speed up the onset of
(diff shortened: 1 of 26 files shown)


examples/tilted_bottom_boundary_layer.jl
```diff
@@ -128,13 +137,15 @@ v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
 # and a constant viscosity and diffusivity. Here we use a smallish value
 # of ``10^{-4} \, \rm{m}^2\, \rm{s}^{-1}``.
 
-closure = ScalarDiffusivity(ν=1e-4, κ=1e-4)
+ν=1e-4
+κ=1e-4
```
glwagner, line 141:
```suggestion
ν = 1e-4
κ = 1e-4
```

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -95,10 +95,19 @@ coriolis = ConstantCartesianCoriolis(f = 1e-4, rotation_axis = ĝ)
 # _perturbations_ away from the constant density stratification by imposing
 # a constant stratification as a `BackgroundField`,
 
-B_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = 1e-5))
+N² = 1e-5 # s⁻¹ # background stratification
+B_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = N²))
 
 # where ``N^2 = 10^{-5} \rm{s}^{-2}`` is the background buoyancy gradient.
```
glwagner, line 101:
```suggestion
```

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -95,10 +95,19 @@ coriolis = ConstantCartesianCoriolis(f = 1e-4, rotation_axis = ĝ)
...
 
+# Because the bottom boundary condition is that there must be zero *total* diffusive
+# flux across the seafloor, the default `FluxBoundaryCondition()` is insufficient;
+# instead, the boundary condition on the perturbation flux must take the background
+# into account,
```
glwagner, line 106: For this I think we should write down the boundary condition that we want to prescribe (I think it's important to make clear that the boundary condition on the total buoyancy is a _choice_ --- once that is decided, then one has to do the math to _compute_ what the boundary condition should be on the perturbation field). This could be important for users that want to use a different boundary condition than zero diffusive flux across the seafloor, and also just pedagogical...

hdrake: How's this: https://github.com/CliMA/Oceananigans.jl/pull/3581/commits/d3dec7f4c0849940d183eebc58a70b4cf5bb997b#diff-a8b1666e910f9c11760ffa5cd0c5b1b0964c32a97b4908aeeb096067a990fac9R101-R111

examples/tilted_bottom_boundary_layer.jl
hdrake: Addressed in https://github.com/CliMA/Oceananigans.jl/pull/3581/commits/727601d7c984d2eaf16031c2226c56cff37ee235

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -98,10 +98,17 @@ coriolis = ConstantCartesianCoriolis(f = 1e-4, rotation_axis = ĝ)
...
 B_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = N²))
 
-# Because the bottom boundary condition is that there must be zero *total* diffusive
-# flux across the seafloor, the default `FluxBoundaryCondition()` is insufficient;
-# instead, the boundary condition on the perturbation flux must take the background
-# into account,
+# We choose to impose a bottom boundary condition of zero *total* diffusive buoyancy
+# flux across the seafloor,
+# ```math
+# ∂_z B = ∂_z b + N^{2} \cos{\theta} = 0.
```
glwagner, line 104: Beautiful! Sorry to be a nag (and correct me if I am wrong) but do we not use `B` to denote the background field in this example? Let's just make sure to use consistent notation in the code and the math. But please feel free to make the notation whatever you like of course!

glwagner: and to clarify this we could write something like

```
where ``∂_z B = N^{2} \cos{\theta}`` is the vertical buoyancy gradient of the background buoyancy field.
```

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -98,10 +98,17 @@ coriolis = ConstantCartesianCoriolis(f = 1e-4, rotation_axis = ĝ)
...
+# ```
+# The default `FluxBoundaryCondition()` (which imposes ``∂_z b = 0``) is inappropriate;
+# instead, we must rearrange this expression to override the boundary condition on the
+# perturbation flux:
+# ```math
+# ∂_z b = - N^{2} \cos{\theta}.
+#```
```
glwagner, line 111:
```suggestion
# This shows that to impose a no-flux boundary condition on the total buoyancy field ``B``, we must apply a boundary condition to the perturbation buoyancy ``b``,
# ```math
# ∂_z b = - N^{2} \cos{\theta}.
#```
```

glwagner: I'm just suggesting a small change in tone here --- setting boundary conditions should always be an intentional act, even if using the defaults (this is one the arguments against having a default at all...) I don't think we want to think of setting boundary conditions as an "override".

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -98,10 +98,17 @@ coriolis = ConstantCartesianCoriolis(f = 1e-4, rotation_axis = ĝ)
...
 
 negative_background_diffusive_flux = GradientBoundaryCondition(-N²*ĝ[3])
```
glwagner, line 113:
```suggestion
∂z_b_bottom = - N² * cosd(θ)
negative_background_diffusive_flux = GradientBoundaryCondition(∂z_b_bottom)
```

I'm ok with using g[3] too, but we would need a comment to explain that it's computed above. So maybe it's clearer just to recompute `cosd` here? Up to you.

glwagner (approved):
Great contribution, thanks for clarifying the boundary conditions and helping users set up experiments more intentionally.

hdrake:
LGTM

examples/tilted_bottom_boundary_layer.jl
glwagner: Ah, but it's called `B_field` above. That should be `B∞_field`?

glwagner:
I'll try to figure out what's going on with the distributed tests

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -95,9 +95,22 @@ coriolis = ConstantCartesianCoriolis(f = 1e-4, rotation_axis = ĝ)
 # _perturbations_ away from the constant density stratification by imposing
 # a constant stratification as a `BackgroundField`,
 
-B_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = 1e-5))
+N² = 1e-5 # s⁻² # background vertical buoyancy gradient
+B_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = N²))
```
glwagner, line 99:
```suggestion
B∞_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = N²))
```

I realized there's an issue in our UI right now where `BackgroundField` is not actually a field, making things a little confusing (we have to redefine `B∞` below to get the actual field). I'll open a separate issue for this...

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -128,13 +141,15 @@ v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
...
+κ = 1e-4
+closure = ScalarDiffusivity(ν=ν, κ=κ)
 
 model = NonhydrostaticModel(; grid, buoyancy, coriolis, closure,
                             timestepper = :RungeKutta3,
                             advection = UpwindBiasedFifthOrder(),
                             tracers = :b,
-                            boundary_conditions = (u=u_bcs, v=v_bcs),
+                            boundary_conditions = (u=u_bcs, v=v_bcs, b=b_bcs),
                             background_fields = (; b=B_field))
```
glwagner, line 153:
```suggestion
                            background_fields = (; b=B∞_field))
```

glwagner:
Nice work @hdrake !

PR #3585, opened 2024-05-06
(0.91.1) CATKE numerics: discretely correct shear production, split-explicit TKE substepping

glwagner:
This PR implements two advancements for CATKE. First is a reformulation of the shear production term. Previously this was computed using the velocity field at time-step `n` in the TKE tendency, like all other terms in the velocity and tracer explicit tendencies. However, [Burchard (2002)](https://www.sciencedirect.com/science/article/pii/S1463500302000094?casa_token=3cKiqlICvN8AAAAA:Hg3iGP-1q_UYh2We7maQRb9z9F5blbp8lHiH61T_Gep7y0DG8VYVRWwGvzvCDlbD2iANNHbLp4g) argue that the shear production term should be formulated to _exactly_ conserve total kinetic energy, which means that it's form actually depends on the time-stepping scheme being used. In our case, temporally-discrete shear production requires using the velocity field at time step `n` and `n+1`.

To implement this scheme we have to add some features to the time stepping routine. In particular we have to allow closures to tell `HydrostaticFreeSurfaceModel` not to step forward certain tracers. Next, we step forward the TKE within `compute_diffusivities!`.

Second, this PR also implements split-explicit substepping for CATKE's TKE. It seems that this is required because even though the new shear production discretization dramatically stabilizes CATKE (allowing long time-steps stably), it does not render CATKE's solutions completely insensitive to the time-step. So, for _accuracy_ (and performance) purposes, we are motivated to also add the capability to advance the TKE on a short time-step while the velocities and tracers advance on a slower time-step. This only matters for high vertical resolutions and very strong forcing. But when you do global simulations, there is strong forcing somewhere on the Earth, and we probably want to represented it accurately...

More documentation to come.


src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_ab2_step.jl
```diff
@@ -11,8 +11,9 @@ import Oceananigans.TimeSteppers: ab2_step!
 
 setup_free_surface!(model, free_surface, χ) = nothing
 
-function ab2_step!(model::HydrostaticFreeSurfaceModel, Δt, χ)
+function ab2_step!(model::HydrostaticFreeSurfaceModel, Δt)
```
navidcy, line 14: nice

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -156,7 +156,7 @@ function HydrostaticFreeSurfaceModel(; grid,
     boundary_conditions = regularize_field_boundary_conditions(boundary_conditions, grid, prognostic_field_names)
 
     # Finally, we ensure that closure-specific boundary conditions, such as
-    # those required by TKEBasedVerticalDiffusivity, are enforced:
+    # those required by CATKEVerticalDiffusivity, are enforced:
```
navidcy, line 159: wow, the `TKEBasedVerticalDiffusivity` is quite old, right? :)

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
```diff
@@ -0,0 +1,206 @@
...
+
+    diffusivity_fields = model.diffusivity_fields
+    κe = diffusivity_fields.κe
+    Le = diffusivity_fields.Le
+    previous_velocities = diffusivity_fields.previous_velocities
+    tracer_index = findfirst(k -> k == :e, keys(model.tracers))
+    implicit_solver = model.timestepper.implicit_solver
+
+    Δt = model.clock.last_Δt
+    Δτ = tke_time_step(closure)
```
navidcy, line 37: is Δτ the tilmestep for `e`? and is it smaller than Δt?
is this is the first place it appears perhaps a comment?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
```diff
@@ -0,0 +1,206 @@
...
+    #          = - (Cᵂϵ * √e) e
+    #
+    # Then the contribution of Jᵉ to the implicit flux is
+    #
+    #       Lᵂ = - Cᵂϵ * √e / Δz.
+    #
+    on_bottom = !inactive_cell(i, j, k, grid) & inactive_cell(i, j, k-1, grid)
+    Δz = Δzᶜᶜᶜ(i, j, k, grid)
+    Cᵂϵ = closure_ij.turbulent_kinetic_energy_equation.Cᵂϵ
+    e⁺ = clip(eⁱʲᵏ)
```
navidcy, line 125: what does `clip` do? where is `clip` defined?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
```diff
@@ -0,0 +1,206 @@
...
+end
+
+@inline function top_tracer_boundary_conditions(grid, tracers)
+    names = propertynames(tracers)
+    values = Tuple(tracers[c].boundary_conditions.top for c in names)
+
+    # Some shenanigans for type stability?
+    return NamedTuple{tuple(names...)}(tuple(values...))
+end
+=#
```
navidcy, line 205: do we need these?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
```diff
@@ -227,16 +179,16 @@ end
 """ Computes the friction velocity u★ based on fluxes of u and v. """
 @inline function friction_velocity(i, j, grid, clock, fields, velocity_bcs)
     FT = eltype(grid)
```
navidcy, line 181: what's FT used for?

validation/vertical_mixing_closures/column_windy_convection.jl
```diff
@@ -30,7 +30,7 @@ coriolis = FPlane(f=f₀)
 b_bcs = FieldBoundaryConditions(top = FluxBoundaryCondition(Jᵇ))
 u_bcs = FieldBoundaryConditions(top = FluxBoundaryCondition(τˣ))
 #closures_to_run = [catke, ri_based, convective_adjustment]
-closures_to_run = [catke] #, ri_based, convective_adjustment]
+closures_to_run = [catke] #, ri_based] #, convective_adjustment]
```
navidcy, line 33: ?

validation/vertical_mixing_closures/column_windy_convection.jl
```diff
@@ -42,17 +42,18 @@ for closure in closures_to_run
     bᵢ(z) = N² * z
     set!(model, b=bᵢ, e=1e-6)
 
-    simulation = Simulation(model; Δt=2minutes, stop_time)
+    simulation = Simulation(model; Δt=20minutes, stop_time)
```
navidcy, line 45: wow

navidcy (approved):

navidcy:
I suggest  a patch release?

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
glwagner: in another file in this module

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
glwagner: ah these are for tuples

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
glwagner: is it not used int his function?

validation/vertical_mixing_closures/column_windy_convection.jl
```diff
@@ -42,17 +42,18 @@ for closure in closures_to_run
...
-    diffusivities = (κᵘ = model.diffusivity_fields.κᵘ,
-                     κᶜ = model.diffusivity_fields.κᶜ)
+    diffusivities = (κu = model.diffusivity_fields.κu,
+                     κc = model.diffusivity_fields.κc)
 
     outputs = merge(model.velocities, model.tracers, diffusivities)
 
     simulation.output_writers[:fields] = JLD2OutputWriter(model, outputs,
                                                           schedule = TimeInterval(20minutes),
+                                                          #schedule = IterationInterval(1),
```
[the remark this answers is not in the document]
glwagner: never "need" of course, but if your closure is messed up you probably should run just a few time steps and output every time step...

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
glwagner: > is Δτ the tilmestep for e?

yes. But isn't `Δτ = tke_time_step(closure)` self explanatory? (tke is e)

> and is it smaller than Δt?

It is either `nothing`, or some number... when taking the tke time-step, it is limited by the large-scale time-step. So it is either `nothing` (in which case we take the same value as `Δt`), or it is either equal to or smaller than `Δt` if explicitly prescribed.

src/TimeSteppers/quasi_adams_bashforth_2.jl
```diff
@@ -69,41 +67,59 @@ Step forward `model` one time step `Δt` with a 2nd-order Adams-Bashforth method
...
+    # Be paranoid and update state at iteration 0
+    model.clock.iteration == 0 && update_state!(model, callbacks)
+
     Δt == 0 && @warn "Δt == 0 may cause model blowup!"
 
-    # Shenanigans for properly starting the AB2 loop with an Euler step
-    euler = euler || (Δt != model.timestepper.previous_Δt)
+    ab2_timestepper = model.timestepper
+
+    # Change the default χ if necessary, which occurs i
```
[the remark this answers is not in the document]
glwagner: true

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
glwagner: Indeed

glwagner:
> I suggest a patch release?

yes that will be useful

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
glwagner: we _will_ need them if we want to run with closures additional to CATKE. If we run only with CATKE we don't need them. I think we should probably keep this code because we want to restore this capability. It's just pointless to support until we can actually verify that CATKE is ready to go in a 3D setting

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
glwagner: `clip` limits `e` to positive values, eg `max(zero(e), e)` to be type stable. Maybe we don't need it...

navidcy: It’s ok. But where does it come from? It’s not in Julia base..

validation/vertical_mixing_closures/column_windy_convection.jl
navidcy: Why this line change with extra comment in the comment..? But not that important tbh

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
navidcy: ok!

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/turbulent_kinetic_energy_equation.jl
navidcy: no is not, am I missing something?

navidcy: deleted it

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
glwagner: I believe it's defined within this same module! I put a comment in

validation/vertical_mixing_closures/column_windy_convection.jl
glwagner: it's two separate changes, one to compare only with catke + ri_based, and then a second to run just catke. I removed it

src/TurbulenceClosures/turbulence_closure_implementations/CATKEVerticalDiffusivities/time_step_turbulent_kinetic_energy.jl
navidcy: all good, I found it... no need to comment

PR #3596, opened 2024-05-09
Enhance `MultiregionGrid` with Conditional Halo Extensions and Debugging Capabilities

siddharthabishnu:
This PR serves as an important step toward integrating the split-explicit free surface with `MultiRegionGrid`. It introduces an optional parameter, `extended_halos`, which allows users to control the behavior of halo filling during each substep. Setting `extended_halos` to `false` maintains the halo size to be the same as the original grid, diverging from the previous default behavior where `extended_halos = true` would automatically extend the halo to cover all split-explicit substeps, significantly enhancing computational performance.

Although setting `extended_halos` to `false` may be less efficient for long simulations on high-resolution`MultiRegionGrids` grids, this feature is helpful for debugging purposes.

Additionally, this PR addresses and resolves a couple of bugs related to the interaction between `ImmersedBoundaryGrid` and `MultiRegionGrid`. These modifications ensure that `ImmersedBoundaryGrid` now correctly wraps over `MultiRegionGrid`, reversing the prior implementation approach.

With @simone-silvestri and @jm-c.


src/MultiRegion/multi_region_split_explicit_free_surface.jl
```diff
@@ -53,20 +88,81 @@ function materialize_free_surface(free_surface::SplitExplicitFreeSurface, veloci
...
+    U_args = (grid, Δτᴮ, η, ηᵐ, ηᵐ⁻¹, ηᵐ⁻², 
+              U, Uᵐ⁻¹, Uᵐ⁻², V,  Vᵐ⁻¹, Vᵐ⁻²,
+              η̅, U̅, V̅, Gᵁ, Gⱽ, Hᶠᶜ, Hᶜᶠ, g, 
+              timestepper)
+
+    @unroll for substep in 1:Nsubsteps
+        averaging_weight = weights[substep]
+
+        fill_halo_regions!(U)
+        fill_halo_regions!(V)
```
simone-silvestri: actually, this would work for multiregion, but I think it would not work for a cubed sphere where we need to fill `U` and `V` together

simone-silvestri: @siddharthabishnu have you seen this? I think you need to correct this for your simulation

simone-silvestri:
How is this PR going along? Is it ready to be merged?

glwagner:
is this stale @siddharthabishnu ?

glwagner:
@siddharthabishnu please re-open this if its not stale.

siddharthabishnu:
> @siddharthabishnu please re-open this if its not stale.

@glwagner yes, this has been stale for a while. Glad you closed it. I believe the changes in this PR are already covered by PR #4538.

PR #3599, opened 2024-05-13
Bitrounding + Lossless compression

milankl:
**TL;DR: We can compress 18GB of Oceananigans simulation checkpoints into 350MB with bitrounding and lossless compression.**

#### Problem
Output is currently uncompressed in Float64 which contains
- redundancies: zeros for immersed boundaries, halos; similar/identical exponent bits
- false information: tailing mantissa bits with no mutual information to neighbouring grid points

#### Proposed solution
Bitrounding to remove false information (replaced with zero bits -> redundancies) then lossless compression to remove redundancies.

I've looked into the bitwise real information content for a single checkpoint in Simone's OMIP simulations and I got this with the orange line denothing the 99.9% of real information 

image

So 
- u, v have 0-2 mantissa bits of information (=keepbits) with more information in the surface layer (k=60)
- w has 0 keepbits (exponent bits though!)
- tempreture T (in ˚C) has 7 keepbits (that's 3-4 digits) relatively independent of depth
- salinity S has 12 at the surface which however increases to 16 in the deep ocean
- sea surface height $\eta$ is at 6 keepbits
- tendencies are generally lower but maybe then shouldn't be stored anyway (use single Euler forward instead)

The checkpoint file Simone provided had
- 18GB total file size, single time step
- including 7 halo points in all directions
- 400MB are grid
- u,v,w,T,S,$\eta$ variables and 2x tendencies (AB2) for all but w, all in Float64

#### Compression options

The 18GB can be compressed into

- Only lossless: 6.9GB (2.6x), removes redundancies from halo and immersed boundaries
- Only Float32: 9GB (2x), removes only some false information in tailing bits
- Float32 then lossless: 3.25GB (5.5x)
- Bitrounded then lossless: 1GB (18x)
- Bitrounded, zero tendencies, then lossless: 350MB (51x), with lossy compression saving the tendencies becomes eventually pointless as restarting with a single Euler forward step might just do the job anyway

This currently uses Zstd (https://github.com/facebook/zstd), a modern yet already widely available lossless compressor through its commandline interface `zstd`. With JLD2 at the moment `compress=true` uses `ZlibCompressor` from https://github.com/JuliaIO/CodecZlib.jl which is similarly good but 2-3x slower. I'm working on getting CodecZstd supported in JLD2: https://github.com/JuliaIO/JLD2.jl/pull/560

While this PR is still a draft I'm proposing the new defaults
- lossless compression with `compress=true` for JLD2, `deflatelevel=3` for netCDF
- bitrounding to keepbits ~20 (single precision-ish) whether you output in Float32/64 (doesn't matter when lossless compression is on)
- a default `bitrounder` that rounds to the keepbits as suggested above that can be used instead of `bitrounder=nothing` (default)

We can then independently tweak the precision (how many keepbits, ideally as a function of the vertical, see salinity) and the lossless compressor (Zlib -> Zstandard)


milankl:
Just added BitInformation to the Project.toml, due to dependency on `StatsBase` and `Distributions` this also adds
```
    Updating `~/git/Oceananigans.jl/Project.toml`
  [de688a37] + BitInformation v0.6.1
    Updating `~/git/Oceananigans.jl/Manifest.toml`
  [66dad0bd] + AliasTables v1.1.2
  [de688a37] + BitInformation v0.6.1
  [49dc2e85] + Calculus v0.5.1
  [31c24e10] + Distributions v0.25.108
  [fa6b7ba4] + DualNumbers v0.6.8
  [1a297f60] + FillArrays v1.11.0
  [34004b35] + HypergeometricFunctions v0.3.23
  [77ba4419] + NaNMath v1.0.2
  [90014a1f] + PDMats v0.11.31
  [1fd47b50] + QuadGK v2.9.4
  [79098fc4] + Rmath v0.7.1
  [2913bbd2] + StatsBase v0.34.3
  [4c63d2b9] + StatsFuns v1.3.1
  [f50d1b31] + Rmath_jll v0.4.0+0
```
also why is the Manifest.toml committed?

glwagner:
> Just added BitInformation to the Project.toml, due to dependency on `StatsBase` and `Distributions` this also adds
> 
> ```
>     Updating `~/git/Oceananigans.jl/Project.toml`
>   [de688a37] + BitInformation v0.6.1
>     Updating `~/git/Oceananigans.jl/Manifest.toml`
>   [66dad0bd] + AliasTables v1.1.2
>   [de688a37] + BitInformation v0.6.1
>   [49dc2e85] + Calculus v0.5.1
>   [31c24e10] + Distributions v0.25.108
>   [fa6b7ba4] + DualNumbers v0.6.8
>   [1a297f60] + FillArrays v1.11.0
>   [34004b35] + HypergeometricFunctions v0.3.23
>   [77ba4419] + NaNMath v1.0.2
>   [90014a1f] + PDMats v0.11.31
>   [1fd47b50] + QuadGK v2.9.4
>   [79098fc4] + Rmath v0.7.1
>   [2913bbd2] + StatsBase v0.34.3
>   [4c63d2b9] + StatsFuns v1.3.1
>   [f50d1b31] + Rmath_jll v0.4.0+0
> ```
> 
> also why is the Manifest.toml committed?

Through past experience we found that we needed the Manifest committed to make sense of the errors we encounter during CI.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -326,7 +320,7 @@ simulation = Simulation(model, Δt=1.25, stop_iteration=3)
 
 f(model) = model.clock.time^2; # scalar output
 
-g(model) = model.clock.time .* exp.(znodes(grid, Center())) # vector/profile output
+g(model) = model.clock.time .* exp.(znodes(Center, grid)) # vector/profile output
```
glwagner, line 323: You may want to merge `main` because I think we need this change for the doctest to pass

src/OutputWriters/bit_rounding.jl
```diff
@@ -0,0 +1,52 @@
...
+    keepbits :: K
+end
+
+# number of keepbits (mantissa bits) for each variable
+default_bit_rounding(::Val{name}) where name = 23   # single precision default
+default_bit_rounding(::Val{:u}) = 2
+default_bit_rounding(::Val{:v}) = 2
+default_bit_rounding(::Val{:w}) = 2
+default_bit_rounding(::Val{:T}) = 7
+default_bit_rounding(::Val{:S}) = 16                # 12 at the surface, 16 deep ocean
```
glwagner, line 13: This is interesting. Why is there a difference between T, S? Is this specific to the simulation that this was tested on, or can we be sure this is valid for all simulations, past climates, future climates, idealized simulations at other resolutions, etc?

It seems we need to have default bit rounding for passive tracers.

src/OutputWriters/netcdf_output_writer.jl
milankl: I'm two commits ahead of main, none behind https://github.com/CliMA/Oceananigans.jl/compare/main...mk/compression I haven't actively changed these, but maybe @simone-silvestri and I started off from an outdated branch?

src/OutputWriters/bit_rounding.jl
milankl: Although relatively robust through time and space, this depends on a lot of things, also whether your unit carries some offset around (e.g. Kelvin vs ˚C, density vs density anomaly). So it's tricky to generalise. I suggest to have some reasonable defaults if someone uses bit rounding (default `nothing` or single precision as you like) but suggest to highlight that this should be checked similar to how I did it here with the bitinformation analysis above.

For global ocean simulations I expect these to be reasonable defaults. I believe for now this is mostly to reduce the filesizes for OMIP simulations

glwagner: True, I'm just not sure that OMIP is going to be the most common use case, so there's a question about what default is appropriate here

src/OutputWriters/netcdf_output_writer.jl
glwagner: yes not sure, but this change does walk back a recent PR

src/OutputWriters/bit_rounding.jl
```diff
@@ -0,0 +1,52 @@
...
+default_bit_rounding(::Val{:η}) = 6
+
+function BitRounding(outputs = nothing;
+                     user_rounding...)
```
glwagner, line 17: For the purpose of figuring out good defaults perhaps we should include `model` as an input here?

glwagner: Then `default_bit_rounding` can take `model` as an argument, and dispatch on various things, for example the equation of state (which should know the units of temperature), and perhaps the biogeochemistry model, which may know the units of some important tracers

src/OutputWriters/bit_rounding.jl
glwagner: The OMIP defaults might belong in the `ClimaOcean` setup, perhaps

milankl: We could set the defaults here as 23 mantissa bits (=Float32 precision, whether you use Float32 or 64) and then lower in ClimaOcean?

src/OutputWriters/netcdf_output_writer.jl
milankl: shoot then I might have created a non-consistent history, sorry, I'll try to resolve that.

glwagner:
@milankl I would like to revive this PR because I think it is very useful and powerful. Is that ok with you?

milankl:
Absolutely! Also feel free to close and start fresh if that's easier. In SpeedyWeather we do the bitrounding on the CPU as we have to transfer the array anyway and it's a very cheap operation that needs to be done on a copy of a field anyway to not affect the state. But happy to work towards a GPU version of bitrounding too for GPUArrays. It's easy to do but currently not done in BitInformation.jl. If you don't like dependencies therein we could also make a new package just for bitrounding if you prefer. Depends really just on how you need it

glwagner:
I think that one always has to transfer data to the CPU in order to save it. I guess we could avoid a dependency by using an extension?

PR #3509, opened 2024-03-13
`TKEDissipationVerticalDiffusivity` (aka k-epsilon) turbulence closure

glwagner:
This PR implements a two-equation turbulent closure, which evolve turbulent kinetic energy and an additional "scale variable". Here we choose the scale variable dissipation; another name for this two-equation model is "k-epsilon". In Oceananigans it will be called the `TKEDissipationVerticalDiffusivity`.

This PR is a WIP; we still have to implement the stability functions and, perhaps more importantly, figure out how to make time-stepping stable. 

Closes #2278

PR #3605, opened 2024-05-23
TKEDissipationVerticalDiffusivity (aka k-epsilon) turbulence closure (take 2)

glwagner:
This PR replaces #3509. Also, it builds off #3585 which hopefully we can merge soon so we don't enter merge conflict hell.

In short this PR implements k-epsilon. Although, the version that's implemented now is not by-the-book k-epsilon but rather a simplified version. I'd like to merge this PR basically as is and hopefully continue to work on it in the future.

There are some numerical considerations which should be reviewed and documented I think. First of all is the challenge of specifying a minimum tke, minimum dissipation rate, and also a maximum dissipation rate. These are useful for producing smooth diffusivities, which note are 

$$ \kappa \sim \frac{e^2}{\epsilon} $$

But then there are challenges with implementing the TKE and dissipation equations. Because depending on the term, one may either want to impose or not impose these constraints.

It could be useful to try to get some lessons from other codes too.

It's not ready for production use, but at least it's doing reasonable things...

image


glwagner:
It might be working...

image

with @legaya and @GabrielMouttapa

glwagner:
Here's a comparison between this model, LES, and some other closures implemented in GOTM.

image

it should be possible to also compare directly with the non-LT version of SMC in GOTM.

glwagner:
Some visualization of the stability functions:

image

navidcy:
OK! I'll review this!

so in Oceananigans terms is the `e-ϵ` parameterization, right?

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/TKEBasedVerticalDiffusivities.jl
```diff
@@ -0,0 +1,186 @@
...
+end
+
+@inline Riᶜᶜᶜ(i, j, k, grid, velocities, tracers, buoyancy) =
+    ℑbzᵃᵃᶜ(i, j, k, grid, Riᶜᶜᶠ, velocities, tracers, buoyancy)
+
+@inline function Riᶜᶜᶠ(i, j, k, grid, velocities, tracers, buoyancy)
+    ∂z_u² = ℑxᶜᵃᵃ(i, j, k, grid, ϕ², ∂zᶠᶜᶠ, velocities.u)
+    ∂z_v² = ℑyᵃᶜᵃ(i, j, k, grid, ϕ², ∂zᶜᶠᶠ, velocities.v)
+    N² = ∂z_b(i, j, k, grid, buoyancy, tracers)
+    S² = ∂z_u² + ∂z_v²
```
navidcy, line 76: isn't this repetitive? why not use `shearᶜᶜᶠ` here?

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
```diff
@@ -0,0 +1,337 @@
...
+- `FT`: Float type; default `Float64`.
+
+
+Keyword arguments
+=================
+
+- `maximum_diffusivity`: Maximum value for tracer, momentum, and TKE diffusivities.
+                        Used to clip the diffusivity when/if CATKE predicts
+                        diffusivities that are too large.
+                        Default: `Inf`.
```
navidcy, line 81:
```suggestion
                         Used to clip the diffusivity when/if CATKE predicts
                         diffusivities that are too large.
                         Default: `Inf`.
```

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
```diff
@@ -0,0 +1,356 @@
...
+- `FT`: Float type; default `Float64`.
+
+
+Keyword arguments
+=================
+
+- `maximum_diffusivity`: Maximum value for tracer, momentum, and TKE diffusivities.
+                        Used to clip the diffusivity when/if CATKE predicts
+                        diffusivities that are too large.
+                        Default: `Inf`.
```
navidcy, line 90: There is a remnant CATKE reference here!!

```suggestion
                         Used to clip the diffusivity when/if CATKE predicts
                         diffusivities that are too large.
                         Default: `Inf`.
```

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
```diff
@@ -0,0 +1,356 @@
...
+    κu = 𝕊u * e² / ϵ
+    κu_max = closure.maximum_viscosity
+    return min(κu, κu_max)
+end
+
+@inline function κcᶜᶜᶠ(i, j, k, grid, closure::TDVD, velocities, tracers, buoyancy)
+    e² = ℑzᵃᵃᶠ(i, j, k, grid, ϕ², turbulent_kinetic_energyᶜᶜᶜ, closure, tracers)
+    ϵ  = ℑzᵃᵃᶠ(i, j, k, grid, dissipationᶜᶜᶜ, closure, tracers, buoyancy)
+    𝕊c = tracer_stability_functionᶜᶜᶠ(i, j, k, grid, closure, velocities, tracers, buoyancy)
+    κc = 𝕊c * e² / ϵ + 1e-5
```
navidcy, line 303: why hardcode the 1e-5?

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
```diff
@@ -0,0 +1,356 @@
...
+    κc_max = closure.maximum_tracer_diffusivity
+    return min(κc, κc_max)
+end
+
+@inline function κeᶜᶜᶠ(i, j, k, grid, closure::TDVD, velocities, tracers, buoyancy)
+    e² = ℑzᵃᵃᶠ(i, j, k, grid, ϕ², turbulent_kinetic_energyᶜᶜᶜ, closure, tracers)
+    ϵ  = ℑzᵃᵃᶠ(i, j, k, grid, dissipationᶜᶜᶜ, closure, tracers, buoyancy)
+    𝕊e = tke_stability_functionᶜᶜᶠ(i, j, k, grid, closure, velocities, tracers, buoyancy)
+    κe = 𝕊e * e² / ϵ + 1e-4
```
navidcy, line 312: why hardcode the 1e-4

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_top_boundary_condition.jl
```diff
@@ -0,0 +1,91 @@
...
+Designed to be used with TKETopBoundaryConditionParameters in a FluxBoundaryCondition, eg:
+
+```
+top_tracer_bcs = top_tracer_boundary_conditions(grid, tracer_names, user_bcs)
+top_velocity_bcs = top_velocity_boundary_conditions(grid, user_bcs)
+parameters = TKETopBoundaryConditionParameters(top_tracer_bcs, top_velocity_bcs)
+top_tke_bc = FluxBoundaryCondition(top_tke_flux, discrete_form=true, parameters=parameters)
+```
+
+See the implementation in catke_equation.jl.
```
navidcy, line 38: > See the implementation in catke_equation.jl.

Is this file still the right place to look for implementation?

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_equations.jl
```diff
@@ -0,0 +1,305 @@
...
+using CUDA
+
+Base.@kwdef struct TKEDissipationEquations{FT}
+    Cᵋϵ :: FT = 1.92
+    Cᴾϵ :: FT = 1.44
+    Cᵇϵ :: FT = -0.65
+    Cᵂu★ :: FT = 1.0
+    CᵂwΔ :: FT = 1.0
+    Cᵂα  :: FT = 0.11
+    gravitational_acceleration  :: FT = 9.8065
```
navidcy, line 17: can we use the Oceananigans constant here?

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_equations.jl
```diff
@@ -0,0 +1,305 @@
...
+
+@inline function top_dissipation_flux(i, j, grid, clock, fields, parameters, closure::FlavorOfTD, buoyancy)
+    closure = getclosure(i, j, closure)
+
+    top_tracer_bcs = parameters.top_tracer_boundary_conditions
+    top_velocity_bcs = parameters.top_velocity_boundary_conditions
+    tke_dissipation_parameters = closure.tke_dissipation_equations
+
+    return _top_dissipation_flux(i, j, grid, clock, fields, tke_dissipation_parameters, closure,
+                         buoyancy, top_tracer_bcs, top_velocity_bcs)
```
navidcy, line 229:
```suggestion
                                 buoyancy, top_tracer_bcs, top_velocity_bcs)
```

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_equations.jl
```diff
@@ -0,0 +1,305 @@
...
+                               parameters::TKEDissipationEquations, closure::TDVD,
+                               buoyancy, top_tracer_bcs, top_velocity_bcs)
+
+    wΔ³ = top_convective_turbulent_velocity_cubed(i, j, grid, clock, fields, buoyancy, top_tracer_bcs)
+    u★ = friction_velocity(i, j, grid, clock, fields, top_velocity_bcs)
+
+    Cᵂu★ = parameters.Cᵂu★
+    CᵂwΔ = parameters.CᵂwΔ
+
+    return zero(grid) #- Cᵂu★ * u★^3 #- CᵂwΔ * wΔ³
```
navidcy, line 218: is this `zeros(grid)` + commented out code a left over from debugging or on purpose?

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_stability_functions.jl
```diff
@@ -0,0 +1,289 @@
...
+                                    Cd₃ = 0.005154,
+                                    Cd₄ = 0.006930,
+                                    Cd₅ = -0.0003372,
+                                    𝕊u₀ = nothing)
+
+    if isnothing(𝕊u₀)
+        a = Cd₅ - Cu₂
+        b = Cd₂ - Cu₀
+        c = Cd₀
+        𝕊u₀ = (2a / (-b - sqrt(b^2 - 4a * c)))^(1/4)
```
navidcy, line 79: where does this come from? reads like the solution of a quadratic polynomial somehow...

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
```diff
@@ -0,0 +1,356 @@
...
+                                      tke_dissipation_time_step = nothing)
+
+Return the `TKEDissipationVerticalDiffusivity` turbulence closure for vertical mixing by
+small-scale ocean turbulence based on the prognostic evolution of subgrid
+Turbulent Kinetic Energy (TKE).
+
+!!! note "TKEDissipationVerticalDiffusivity"
+    `TKEDissipationVerticalDiffusivity` is a new turbulence closure diffusivity. The default
+    values for its free parameters are obtained from calibration against large eddy
+    simulations. For more details please refer to [Wagner23catke](@cite).
```
navidcy, line 71: We don't mention here that this is otherwise known as the "k-epsilon". I think we should. Perhaps a reference? Or at least mention that is inspired or a simplified version of...

Also, why do we claim that this is "a new turbulence closure diffusivity"? Perhaps you meant to say "unvalidated" or something?

navidcy (approved):
good job; sprinkled few questions/comments here and there

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
iuryt: Maybe "newly implemented" or "recently implemented feature"? I agree that the wording seems like it is being claimed that we developed the parameterization.

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/TKEBasedVerticalDiffusivities.jl
```diff
@@ -0,0 +1,186 @@
...
+    p⁺ = peripheral_node(i, j, k⁺, grid, c, c, f)
+    p⁻ = peripheral_node(i, j, k⁻, grid, c, c, f)
+
+    f⁺ = ifelse(p⁺, f⁻, f⁺)
+    f⁻ = ifelse(p⁻, f⁺, f⁻)
+
+    return (f⁺ + f⁻) / 2
+end
+
+# To reconstruct buoyancy flux "conservatively" (ie approximately correpsonding to production/destruction
```
simone-silvestri, line 100:
```suggestion
# To reconstruct buoyancy flux "conservatively" (ie approximately corresponding to production/destruction
```

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_stability_functions.jl
navidcy: (a comment next to the expression in the code might be helpful?)

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/TKEBasedVerticalDiffusivities.jl
glwagner: I think we could yes

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_stability_functions.jl
glwagner: yes that's right, that's what it is. This is a log layer solution where production balances dissipation, I will write a comment. There's no good reference in the literature surprisingly

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_equations.jl
glwagner: Hmm yes, a non-zero TKE flux is non-canonical but we should support this with the parameters just chosen to be 0 for now

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_equations.jl
glwagner: But wouldn't that mislead people into trying to change the constant in the code rather than specifically set this?

glwagner: I guess we still don't have a solid policy for constants and this has come up previously with no solution

glwagner: To clarify, say that someone wants to change `g` in their problem. If we take all constants from one place, peeople might think "oh, I can just fork the code and recompile Oceananigans with a different constant". 

But this is not what we want. We want people to specifically change the constant in their user code.

So then we also have to ask ourselves, why do we want to take constants from a central location?

If we want to do that, then we should probably use clima parameters instead.

glwagner: This isn't the first time it's come up so clearly we are missing something here

navidcy: ok, let's leave it as is for now ;)

glwagner: We may actually be able to excise this ... let me look into it a little more

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_equations.jl
navidcy: OK, does this deserve a comment?

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_equations.jl
glwagner: Maybe we can actually get rid of the Charnock length dependence here which I think is the best solution. Already there is the possibility of a mismatch between the surface flux model and this dissipation flux model which is not ideal

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_top_boundary_condition.jl
glwagner: yes

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
glwagner: mistake

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
glwagner: mistake

glwagner:
> OK! I'll review this!
> 
> so in Oceananigans terms is the `e-ϵ` parameterization, right?

I think we should call it "TKE dissipation vertical diffusivity", or k-epsilon in a nod to the literature. I don't think we shoudl call it "e-epsilon".

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
glwagner: I deleted it

navidcy:
> > OK! I'll review this!
> > so in Oceananigans terms is the `e-ϵ` parameterization, right?
> 
> I think we should call it "TKE dissipation vertical diffusivity", or k-epsilon in a nod to the literature. I don't think we shoudl call it "e-epsilon".

No, definitely not!! I wasn't implying that we call it "e-epsilon". I was just trying to see whether the `k` from `k-ε` is the same as our `e`... But after my question I soon figured it out -- sorry, I should have deleted the comment.

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
glwagner: Those words were leftover from CATKE (I copy pasted the whole file from the CATKE implementation)

glwagner:
> > I think we should call it "TKE dissipation vertical diffusivity", or k-epsilon in a nod to the literature. I don't think we shoudl call it "e-epsilon".
> 
> No, definitely not!! I wasn't implying that we call it "e-epsilon". I was just trying to see whether the `k` from `k-ε` is the same as our `e`... But after my question I soon figured it out -- sorry, I should have deleted the comment.
> 
> Let's call it "TKE dissipation vertical diffusivity" but let's also mention in the docstring or somewhere that this is "the otherwise known as k-epsilon"

Ah yes, "k" is turbulent kinetic energy in the name "k-epsilon".

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
glwagner: I put a few referencessin there and removed the warning about validation

Issue #3593, opened 2024-05-09
Pressure has extremely high gradients in random chunks of simulation using `NonhydrostaticModel` with `ImmersedBoundaryGrid` and `BuoyancyTracer`


tomchor:
In some simulations the pressure values (and more importantly the values of the pressure gradient) are larger than expected by several orders of magnitude. This happens in seemingly random (time) chunks of the simulations, and then everything comes back to normal before happening again in other chunks.

Here's a MWE where I create a grid with channel-like topology which has an immersed boundary at `x=400meters` that acts as an East wall. I initialize it with a uniform stratification of `6e-6/second` and zero velocities everywhere.

```julia
using Oceananigans

grid_base = RectilinearGrid(topology = (Bounded, Periodic, Bounded),
                            size = (16, 20, 4), extent = (800, 1000, 100),)

@inline east_wall(x, y, z) = x > 400
grid = ImmersedBoundaryGrid(grid_base, GridFittedBoundary(east_wall))

model = NonhydrostaticModel(grid = grid, timestepper = :RungeKutta3,
                            buoyancy = BuoyancyTracer(), tracers = :b,
                            )

N² = 6e-6
b∞(x, y, z) = N² * z
set!(model, b=b∞)

simulation = Simulation(model, Δt=25, stop_time=1e4,)
simulation.output_writers[:snaps] = NetCDFOutputWriter(model, (; model.pressures.pNHS,),
                                                       filename = "test_pressure.nc",
                                                       schedule = TimeInterval(100),
                                                       overwrite_existing = true,)
run!(simulation)
```

Here are some snapshots of the pressure (the blue line is where the immersed boundary is):

![image](https://github.com/CliMA/Oceananigans.jl/assets/13205162/8904fd81-70ee-465b-a664-3eb6042f94ca)

And here's a plot of the standard deviation of pressure over the whole domain as a function of time where we can visualize when this issue happens:

![image](https://github.com/CliMA/Oceananigans.jl/assets/13205162/86e5687e-0290-45ac-80d9-0cc2552f2b7d)

[Here](https://pastebin.com/XArt8Bik) is the code I used to generate these figures in case anyone's interested in playing around with it.

[...]

Issue #3056, opened 2023-04-09
`NetCDFOutputWriter` sometimes outputs twice at approximately the same time step


tomchor:
I noticed that occasionally the `NetCDFOutputWriter` will output twice at what is virtually the the same time.

As an example, I'm showing below the time interval between time steps of one of my netcdf files created with a nominal `TimeInterval` of approximately 1236. However, you see that (while most of the times are separate by the correct amount) there are some instances with outputs separated by something much smaller:

```julia
julia> diff(ds[:time])
255-element Vector{Float64}:
 1236.3673926631677
 1236.3673926631677
 1236.3673926631677
 1236.3673926631677
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.367392663169
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
 1236.3673926631673
    ⋮
 1236.3673926631454
 1236.3673926631454
 1236.3673926632036
    5.820766091346741e-11
 1236.3673926631454
 1236.3673926631454
 1236.3673926631454
 1236.3673926631454
 1236.3673926631454
 1236.3673926631454
 1236.3673926632036
    5.820766091346741e-11
 1236.3673926631454
 1236.3673926632036
   73.29669670568546
 1163.07069595746
 1236.3673926631454
 1236.3673926631454
 1236.3673926631454
 1236.3673926631454
 1236.3673926631454
```

Furthermore, In this specific case, if the time steps were done correctly we would have 251 time steps, but we have 255.

Unfortunately I haven't been able to reproduce this in a MWE. My guess is that it needs a `TimeStepWizard` (which I am using) and complex dynamics which make the wizard increase and decrease the time step relatively often (which also happens in my case).

Two things to note:

- At no point while making this NetCDF file I restarted the simulation. It has always been one go.
- This seems to happen more often with a Float32 type than with a Float64.

glwagner:
Sounds like a roundoff error. I've noticed similar problems.

glwagner:
Maybe related to #2321

tomchor:
You're right, it does sound the same problem. My "extra time steps" also tend to happen at the end of long runs.

mncrowe:
[...]

PR #3606, opened 2024-05-24
(0.95.3) Reset model clock and skip `time_step!`ing if next actuation time is tiny

tomchor:
Closes https://github.com/CliMA/Oceananigans.jl/issues/3593

This is somewhat of a hack to investigate (and eventually solve) https://github.com/CliMA/Oceananigans.jl/issues/3593. For now I just want to investigate if this breaks too much stuff in the code, but it does seem to solve the problem. Here's a MWE to demonstrate:

```julia
using Oceananigans

grid_base = RectilinearGrid(topology = (Bounded, Periodic, Bounded),
                            size = (16, 20, 4), extent = (800, 1000, 100),)
    
model = NonhydrostaticModel(grid = grid, timestepper = :RungeKutta3, buoyancy = BuoyancyTracer(), tracers = :b,)

N² = 6e-6
b∞(x, y, z) = N² * z
set!(model, b=b∞)
    
using Statistics: std
using Printf
add_callback!(simulation, progress_message, IterationInterval(1))

simulation.output_writers[:snaps] = NetCDFOutputWriter(model, (; model.pressures.pNHS,),
                                                       filename = "test_pressure.nc",
                                                       schedule = TimeInterval(100),
                                                       overwrite_existing = true,)
run!(simulation)
```

On main this produces stuff like:

```
Iteration: 0001, time: 25.0, iteration×Δt: 25.0, std(pNHS) = 6.02e-03
Iteration: 0002, time: 50.0, iteration×Δt: 50.0, std(pNHS) = 6.02e-03
Iteration: 0003, time: 75.0, iteration×Δt: 75.0, std(pNHS) = 6.02e-03
Iteration: 0004, time: 99.99999999999999, iteration×Δt: 100.0, std(pNHS) = 6.02e-03
Iteration: 0005, time: 100.0, iteration×Δt: 125.0, std(pNHS) = 2.72e+10
```

The last two lines are of note where we went from `time: 99.99999999999999` to `time: 100.0`, implying a very tiny time-step, which results in a weird pressure field, as quantified by the last output of the last line: `std(pNHS) = 2.72e+10`. Note that because of this, `time` and `iteration×Δt` don't match up anymore in the last line. Namely `time: 100.0, iteration×Δt: 125.0`. This "misstep" happens many times throughout the run on `main`.

On this branch this doesn't happen anymore, and even after many time-steps things remain aligned (albeit with a very small round-off error):

```
Iteration: 0396, time: 9900.0, iteration×Δt: 9900.0, std(pNHS) = 5.99e-03
Iteration: 0397, time: 9925.000000000002, iteration×Δt: 9925.0, std(pNHS) = 5.99e-03
Iteration: 0398, time: 9950.000000000004, iteration×Δt: 9950.0, std(pNHS) = 5.99e-03
Iteration: 0399, time: 9975.000000000005, iteration×Δt: 9975.0, std(pNHS) = 5.99e-03
```

Ideally the way to really fix this would be to figure out a way to avoid round-off errors, but I haven't been able to do that yet.


src/Simulations/run.jl
```diff
@@ -131,7 +140,12 @@ function time_step!(sim::Simulation)
 
     else # business as usual...
         Δt = aligned_time_step(sim, sim.Δt)
-        time_step!(sim.model, Δt, callbacks=model_callbacks)
+        if Δt > (sim.Δt / 1e10)
+            time_step!(sim.model, Δt, callbacks=model_callbacks)
+        else
+            println("Skipping aligned time step, which is of ", Δt)
+            sim.model.clock.time = next_actuation_time(sim)
```
glwagner, line 147: Could this also be

```suggestion
            sim.model.clock.time += Δt
```

?

glwagner:
Nice work! I'm curious about the criteria. Should it be something like 

```julia
dt = 10 * eps(dt) * sim.dt
```

? Or does it have to be larger than that (hence the factor 1e10).

It'd be nice not to have to define `next_actuation_time` for every schedule... it doesn't really make sense for `WallTimeInterval` either. Plus, we want users to be able to provide custom schedules (since they only need to be a function of `model` that returns true/false) so that people can trigger output / action using interesting custom criteria...

src/Simulations/run.jl
tomchor: I don't think this works when we're _almost_ at `next_actuation_time` since it would skip that actuation time, no?

tomchor:
> Nice work! I'm curious about the criteria. Should it be something like
> 
> ```julia
> dt = 10 * eps(dt) * sim.dt
> ```
> 
> ? Or does it have to be larger than that (hence the factor 1e10).

I actually don't know what the proper criterion should be. With the one you proposed, the error doesn't go away in this example since the tiny time-step is about `1e-12`, but `10 * eps(dt) * sim.dt` come out to be about `1e-13`. If we use `100 * eps(dt) * sim.dt` then it works. But I don't yet know how much of this will generalize to other, more complex simulations. I still have to test these on my own simulations to see what works.

> It'd be nice not to have to define `next_actuation_time` for every schedule... it doesn't really make sense for `WallTimeInterval` either. Plus, we want users to be able to provide custom schedules (since they only need to be a function of `model` that returns true/false) so that people can trigger output / action using interesting custom criteria...

Yeah, agree. I'm not sure of a good workaround here though. Do you have suggestions?

For the time being we can just set a fallback method as `next_actuation_time(scheduke) = Inf` I guess? (Similar to what I did for `IterationInterval`.

Also, nice to see that tests pass and nothing is breaking :)

tomchor:
I did a few tests with some criteria for timestep-skipping with a couple of my own simulations in addition to the MWE included here. In summary:

1. Criterion `sim.Δt / 1e10`: successfully gets rids of the problem in both the MWE and in my simulations
2. Criterion `10 * eps(sim.Δt) * sim.Δt`: doesn't get rid of the problem in any simulation
3. `100 * eps(sim.Δt) * sim.Δt`: fixes the problem in the MWE but not in my simulations, although it does decrease its frequency of occurrence a good amount.
4. `1000 * eps(sim.Δt) * sim.Δt`: fixes everything in all simulations I've tried.

So only options 1 and 4 fully fix the problem (at least in the simulations I've tried so far). For me both those options rely on pretty arbitrary numbers though, so I'm not very happy with neither. From the point of view seeing the timestep-skipping as an approximation ($u^{n+1} \approx u^n$), then maybe criterion 1 makes more sense, although I'm not sure how it'd behave for Float32 simulations.

I see three possible ways to go about it right now:

1. Do what this PR is doing, and manually set the criterion to either option 1 or 4 above. If it turns out that some simulations still have issues, we revisit.
2. We add `min_Δt` as a property of  `NonhydrostaticModel` (or maybe `Simulation`?). I think the minimum `Δt` for which time skipping will be necessary will vary significantly between simulations, so this solution deals with that by leaving the decision up to the user if they are interested in the pressure output.
3. We try something that actually prevents these round-off errors instead of dealing with them. @glwagner suggested an `Integer`-based model clock, but there might be other options.

src/Simulations/run.jl
glwagner: What is

```julia
sim.model.clock.time + Δt
```

vs 

```julia
next_actuation_time(sim)
```

?

glwagner:
> I did a few tests with some criteria for timestep-skipping with a couple of my own simulations in addition to the MWE included here. In summary:
> 
> 1. Criterion `sim.Δt / 1e10`: successfully gets rids of the problem in both the MWE and in my simulations
> 2. Criterion `10 * eps(sim.Δt) * sim.Δt`: doesn't get rid of the problem in any simulation
> 3. `100 * eps(sim.Δt) * sim.Δt`: fixes the problem in the MWE but not in my simulations, although it does decrease its frequency of occurrence a good amount.
> 4. `1000 * eps(sim.Δt) * sim.Δt`: fixes everything in all simulations I've tried.
> 
> So only options 1 and 4 fully fix the problem (at least in the simulations I've tried so far). For me both those options rely on pretty arbitrary numbers though, so I'm not very happy with neither. From the point of view seeing the timestep-skipping as an approximation (un+1≈un), then maybe criterion 1 makes more sense, although I'm not sure how it'd behave for Float32 simulations.
> 
> I see three possible ways to go about it right now:
> 
> 1. Do what this PR is doing, and manually set the criterion to either option 1 or 4 above. If it turns out that some simulations still have issues, we revisit.
> 2. We add `min_Δt` as a property of  `NonhydrostaticModel` (or maybe `Simulation`?). I think the minimum `Δt` for which time skipping will be necessary will vary significantly between simulations, so this solution deals with that by leaving the decision up to the user if they are interested in the pressure output.
> 3. We try something that actually prevents these round-off errors instead of dealing with them. @glwagner suggested an `Integer`-based model clock, but there might be other options.

Note that `eps(sim.Δt)` is just `sim.Δt * eps(typeof(Δt))`. So `Δt / 1e10` is pretty similar to `100000 * eps(sim.Δt)`. The only point of using `eps` is to avoid hard coding `Float64`.

tomchor:
@glwagner are you okay if I just add `min_Δt` as a property of `NonhydrostaticModel` and maintain the strategy of skipping the timestep is Δt is smaller than that? I think that's a reasonable and simple way to fix this.

glwagner:
This isn't a problem of the nonhydrostatic model specifically. Why would we add it as a property there?

tomchor:
> This isn't a problem of the nonhydrostatic model specifically. Why would we add it as a property there?

This PR is supposed to fix the output of the pressure, which can be unreasonably large if the time step is really small due to, I believe, the nonhydrostatic pressure correction, which is specific to the `NonhydrostaticModel`. Unless I'm missing something.

This wouldn't fix https://github.com/CliMA/Oceananigans.jl/issues/3056, which, agreed, is more general. I can also add `min_Δt` to `Simulation`, so that this can be used with other models.

glwagner:
I think we should fix the problem once. Otherwise we'll end up with unnecessary code somewhere that has to be deleted.

tomchor:
> I think we should fix the problem once. Otherwise we'll end up with unnecessary code somewhere that has to be deleted.

@glwagner Can you please be clearer? Does that mean adding `min_Δt` to `Simulation` is an acceptable solution? Or should we try to avoid these round-off errors to even happen in the first place?

glwagner:
I think the solution discussed here, where the time-step _change_ associated with `TimeInterval` schedules is restricted by a sort of relative tolerance criteria, is acceptable if we can't tease out the underlying issue (or its unsolvable).

If we could indeed solve the problem simply by eliminating round off error, then this would almost certainly be preferred since it might be much simpler (eg just fixing an floating-point-unstable arithmetic operation by rearranging terms). That's be really easy...

@Sbozzolo might be able to help because I believe they do something special to avoid round off issues in `ClimaAtmos`.

glwagner:
I think #3616 may solve one of the issues here, which is that the alignment of time steps with `TimeInterval` caused an accumulation of round off error even when using a constant time-step and output that should have been aligned with the time-step. So that's good.

However, this issue also aims to solve an issue with the pressure solver that occurs when the time step is too small. I believe this issue remains --- despite that it is perhaps? less likely after fixing `TimeInterval`. In other words the pressure solver simply has a problem for small time steps, regardless of what produces those time steps. Should we re-orient this PR into solving that issue?

glwagner:
I'll try to get going with the MWE here. Is the immersed boundary acually part of the MWE (ie the error does not occur with it?) And buoyancy?

glwagner:
It seems that without the immersed boundary, there is still a problem of super small time-steps, but the pressure does not blow up.

glwagner:
Ok getting closer maybe...

I think this problem is generic and cannot be solved in general for arbitrary time steps. Here's a few thoughts:

- Reading about [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) makes it clear that we simply cannot avoid errors if we would like to add a small floating point number (the time step) to a very large number (the model time). 

- I think the issue with the time-step is whether or not we can compute the RHS of the pressure Poisson equation accurately --- which is `div(u') / Δt`, where `u' = u + Δt * Gu` is the predictor velocity and `div` is the divergence`. This is interesting, because I could not figure out why we would ever find large `div(u')` with small `Δt` even in this MWE. But now I realize that because of the status of the immersed Poisson solver, the velocity along the boundary is divergent, strongly so. So, `div(u')` is large along the boundary. And when we divide by `Δt` we get something huge. The magnitude of `div(u')` also somehow seems to depend on the time step (as does the magnitude of the spurious circulation). The correct solution to this case remains at rest of course. (An aside is that this problem _could_ be avoided by separately computing the hydrostatic pressure, and then using a special horizontal gradient operators that avoid computing a hydrostatic pressure gradient across an immersed boundary. However, this would only be correct for no-flux boundary conditions on buoyancy on side walls). Anyways, apparently because of this issue with the immersed pressure solver, it seems that `div(u')` is large (because `div(u)` is large) even when `Δt = O(1e-14)`...

- As a result of all of this I am confused about whether this MWE is actually reliable for debugging the issue. I guess we should expect to see problems simply when `Δt = O(eps)` because this is when `div(u') / Δt` cannot be reliably computed, I think. This leads to a fairly simple criteria for the time step that's compatible with the pressure correction. But as noted in this PR, this is not enough to fix issues with the immersed boundary MWE... but whether or not that is because of problems with the setup itself, I'm not sure...

- All of that said, taking @tomchor suggestion to be more careful in updating the clock for RK3 actually does solve the MWE here. Obviously, this is again addressing the (in principle not entirely solvable) issue of error accumulation in `clock.time`, rather than addressing the other issue with very small time-steps producing an ill-posed pressure correction. I think we should fix RK3 separately, basically because we cannot completely avoid accumulating error in `clock.time`, every little thing we do to make it more accurate is a good idea.

glwagner:
PR #3617 helps with this MWE

Sbozzolo:
> @Sbozzolo might be able to help because I believe they do something special to avoid round off issues in `ClimaAtmos`.

With regards to floating point instabilities due to arithmetic with time and time intervals, ultimately, we will be solving this issue (and others) by moving away from a floating point time in favor of an integer one (e.g., milliseconds from a start date). As a fun fact, if you are using Float32 and keep track of your time in seconds, `t + dt` is going to have numerical error after approximately one year of simulated time (which is not that much).

src/Simulations/run.jl
glwagner: @tomchor still have this question

src/Simulations/run.jl
```diff
@@ -131,7 +140,12 @@ function time_step!(sim::Simulation)
 
     else # business as usual...
         Δt = aligned_time_step(sim, sim.Δt)
-        time_step!(sim.model, Δt, callbacks=model_callbacks)
+        if Δt > (sim.Δt / 1e10)
```
glwagner, line 143: Seeing as this is a `Simulation` thing, we could add this as a property to `Simulation` with a default 0 (or `nothing` if we want...)

```julia
Simulation.minimum_relative_step = 0
```

I would write the logic as

```julia
if Δt < sim.minimum_relative_step * sim.Δt
    next_time = next_actuation_time(sim)
    @warn "Reseting clock to $next_time and skipping aligned time step Δt = ", Δt)
    sim.model.clock.time = next_time
else
    time_step!(sim.model, Δt, callbacks=model_callbacks)
end
```

This makes is clear that the information to pay attention to is when the time-step is too small I think.

src/Simulations/run.jl
tomchor: Ah, sorry I missed this. It's been so long that I forgot the details, but looking at the code for both, it's possible they end up calculating the same thing since `Δt` is calculated with `aligned_time_step()`.

src/Simulations/run.jl
tomchor: Assuming the logic also works with `next_actuation_time(sim)`, I agree with this. Is this the way we want to solve this issue though? I think in the past we were considering skipping a step a rather "hacky" solution.

glwagner: I do think its cleaner to use a digitized / discretized time variable with a minimum increment and @Sbozzolo has actually been developing such a thing. But also if you _need_ this then we should consider it. Although, nobody else is complaining (but I guess people are shy)

tomchor: For the record, I do know of at least two people that have reported this issue to me through personal communication, but never reported it online.

glwagner: ok

src/Simulations/simulation.jl
```diff
@@ -100,6 +103,7 @@ function Base.show(io::IO, s::Simulation)
                      "├── Stop time: $(prettytime(s.stop_time))", "\n",
                      "├── Stop iteration: $(s.stop_iteration)", "\n",
                      "├── Wall time limit: $(s.wall_time_limit)", "\n",
+                     "├── Minimum relative step: $(s.minimum_relative_step)", "\n",
```
glwagner, line 106:
```suggestion
                     "├── Minimum relative step: ", prettysummary(s.minimum_relative_step), "\n",
```

Looks nicer for round numbers

glwagner (approved):

glwagner:
Looks good, maybe bump the minor version

tomchor:
> Looks good, maybe bump the minor version

Done! I wanna test this code in my actual simulation before sending it off for reviews though.

Also, I'm thinking of using the MWE at the top comment as a test. It's the most reliable way I could reproduce this issue. Thoughts?

glwagner:
> > Looks good, maybe bump the minor version
> 
> Done! I wanna test this code in my actual simulation before sending it off for reviews though.
> 
> Also, I'm thinking of using the MWE at the top comment as a test. It's the most reliable way I could reproduce this issue. Thoughts?

I would just put in a very simple test that time-stepping is skipped correctly, by manually taking two time-steps and changing `dt` in between.

I think the kind of test you are describing may not be appropriate for CI; it's more a validation test. I think if you can put a unit test to show the feature works correctly, you can later show that using the feature solves pressure solver issues and be happy that the unit test ensures it will continue to work as in the validation test.

tomchor:
I tested this branch on my simulation and things work fine. However, when trying to include a test I came across some weird behavior. Namely the snippet below fails:

```julia
using Oceananigans
using Test

grid  = RectilinearGrid(size=(1, 1, 1), extent=(1, 1, 1))

simulation = Simulation(model, Δt=3, stop_iteration=1)
run!(simulation)

stop_time = 1.0
simulation = Simulation(model, Δt=1; stop_time)
run!(simulation)

@test simulation.model.clock.time == stop_time
```

with

```
Test Failed at REPL[11]:1
  Expression: simulation.model.clock.time == stop_time
   Evaluated: 4.0 == 1.0
```

If I re-build `model` before re-creating a simulation, then it works:

```julia
using Oceananigans
using Test

grid  = RectilinearGrid(size=(1, 1, 1), extent=(1, 1, 1))

simulation = Simulation(model, Δt=3, stop_iteration=1)
run!(simulation)

stop_time = 1.0
simulation = Simulation(model, Δt=1; stop_time)
run!(simulation)

@test simulation.model.clock.time == stop_time
```

So there seems to be some attribute of `model` (presumably `model.clock`?) that's leading to different `Simulation` behavior. Is this expected? I couldn't quite figure out what was happening.

src/Simulations/run.jl
```diff
@@ -102,11 +115,24 @@ end
 
 const ModelCallsite = Union{TendencyCallsite, UpdateStateCallsite}
 
+
+function time_step_or_skip!(sim)
+    model_callbacks = Tuple(cb for cb in values(sim.callbacks) if cb.callsite isa ModelCallsite)
+    Δt = aligned_time_step(sim, sim.Δt)
+    if Δt < sim.minimum_relative_step * sim.Δt
+        next_time = next_actuation_time(sim)
```
tomchor, line 123: FYI, `next_actuation_time(sim)` and `sim.model.clock.time + Δt` are equivalent. I chose to use the former here to avoid errors related to the addition operation. But let me know if I should change it to just use the latter.

tomchor:
> > So there seems to be some attribute of `model` (presumably `model.clock`?) that's leading to different `Simulation` behavior. Is this expected? I couldn't quite figure out what was happening.
> 
> Indeed, the clock is not reset to zero when the simulation is built, so you are restarting from `model.clock == 3.0`.

In this case shouldn't the simulation just not iterate and keep the clock at 3?

src/Simulations/simulation.jl
```diff
@@ -44,12 +45,16 @@ Keyword arguments
...
+                           This avoids extremely high values when writing the pressure to disk.
+                           Default value is 0. See github.com/CliMA/Oceananigans.jl/issues/3593 for details.
 """
 function Simulation(model; Δt,
                     verbose = true,
                     stop_iteration = Inf,
                     stop_time = Inf,
-                    wall_time_limit = Inf)
+                    wall_time_limit = Inf,
+                    minimum_relative_step = 0)
```
navidcy, line 57: we should update the method signature in the docstring

src/Simulations/simulation.jl
```diff
@@ -88,7 +93,8 @@ function Simulation(model; Δt,
                      0.0,
                      false,
                      false,
-                     verbose)
+                     verbose,
+                     Float64(minimum_relative_step))
```
navidcy, line 97: why Float64?

src/Simulations/simulation.jl
tomchor: Good catch. Done!

src/Simulations/simulation.jl
tomchor: I was following the other floats in the `Simulation` constructor, which are also converted to `Float64`. I can't remember the PR where this was decided, but it minimizes errors in time-step alignment. The error that this PR solves is an example of this type of error .

src/Simulations/run.jl
glwagner: Doesn't it complicate the code to define `next_actuation_time` in addition to `aligned_time_step`? Things go wrong if you change one but not the other. It's usually best to have one "source" of reality / truth

src/Simulations/simulation.jl
simone-silvestri: Probably we should make it optional. 
This will not work if we want to support Metal architectures that do not want Float64.

tomchor: That sounds good. Although I think we should do that on another PR, since there are other things that follow this convention. For this PR I think we should follow what's already there.

src/Simulations/run.jl
tomchor: Fair enough. Changed it to the simpler `sim.model.clock.time + Δt`.

src/Simulations/run.jl
```diff
@@ -102,11 +102,24 @@ end
 
 const ModelCallsite = Union{TendencyCallsite, UpdateStateCallsite}
 
+
+function time_step_or_skip!(sim)
+    model_callbacks = Tuple(cb for cb in values(sim.callbacks) if cb.callsite isa ModelCallsite)
+    Δt = aligned_time_step(sim, sim.Δt)
+    if Δt < sim.minimum_relative_step * sim.Δt
+        next_time = sim.model.clock.time + Δt
+        @warn "Reseting clock to $next_time and skipping aligned time step Δt = $Δt"
```
simone-silvestri, line 111:
```suggestion
        @warn "Resetting clock to $next_time and skipping time step of size Δt = $Δt"
```

src/Simulations/simulation.jl
simone-silvestri: Well now that I see it better, this never goes in a kernel, so it should be fine

simone-silvestri (approved):

src/Simulations/run.jl
```diff
@@ -102,11 +102,24 @@ end
...
+        time_step!(sim.model, Δt, callbacks=model_callbacks)
+    end
+end
+
+
 """ Step `sim`ulation forward by one time step. """
 function time_step!(sim::Simulation)
 
     start_time_step = time_ns()
-    model_callbacks = Tuple(cb for cb in values(sim.callbacks) if cb.callsite isa ModelCallsite)
```
simone-silvestri, line 109: are we removing these callbacks now?

src/Simulations/simulation.jl
glwagner: should be easy to change these if we need to though as stated I don't know if we do for Metal since they aren't kernel variables

tomchor:
@wenegrat @zhazorken This should fix the pressure issues when writing to disk. All you need to do is add the keyword argument `minimum_relative_step=1e-10` (or some other very small positive number) when constructing a `Simulation`. Let me know if you have any issues.

PR #3616, opened 2024-06-06
Change `TimeInterval` to avoid roundoff error issue plus some cleanup

glwagner:
This PR modifies `TimeInterval` so that, rather than keeping track of the `previous_actuation_time`, it keeps track of the `first_actuation_time` plus the number of `actuations` (including the initial actuation). The next actuation time is then computed by 

```julia
function next_actuation_time(schedule::TimeInterval)
    t₀ = schedule.first_actuation_time
    N = schedule.actuations
    T = schedule.interval
    return = t₀ + N * T
end
```

Note we use `N` instead of `N+1` because the first actuation occurs at `t₀`. So for example, after this first actuation the next actuation time is `t₀ + T`.

This may address #3056.

### DIFF
--- a/src/OutputWriters/OutputWriters.jl
+++ b/src/OutputWriters/OutputWriters.jl
@@ -19,7 +19,7 @@ using Oceananigans.Utils: pretty_filesize
 
 using OffsetArrays
 
-import Oceananigans: write_output!
+import Oceananigans: write_output!, initialize!
 
 Base.open(ow::AbstractOutputWriter) = nothing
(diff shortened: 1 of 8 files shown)


tomchor (approved):
Nice! This looks an improvement regardless, so I'm approving it, but does it actually solve https://github.com/CliMA/Oceananigans.jl/issues/3056?

glwagner:
Let's keep #3056 open in case it doesn't help

PR #3619, opened 2024-06-13
`Adapt`ing all fields of the `OrthogonalSphericalShellGrid`

simone-silvestri:
This PR improves the `adapt_structure` method of the `OrthogonalSphericalShellGrid` to allow a more complicated `conformal_mapping` to be transferred to the GPU

### DIFF
--- a/src/Grids/orthogonal_spherical_shell_grid.jl
+++ b/src/Grids/orthogonal_spherical_shell_grid.jl
@@ -969,8 +969,8 @@ function Adapt.adapt_structure(to, grid::OrthogonalSphericalShellGrid)
                                                     adapt(to, grid.Azᶠᶜᵃ),
                                                     adapt(to, grid.Azᶜᶠᵃ),
                                                     adapt(to, grid.Azᶠᶠᵃ),
-                                                    grid.radius,
-                                                    grid.conformal_mapping)
+                                                    adapt(to, grid.radius),
+                                                    adapt(to, grid.conformal_mapping))
 end
 
 function Base.summary(grid::OrthogonalSphericalShellGrid)
--- a/src/Grids/orthogonal_spherical_shell_grid.jl
+++ b/src/Grids/orthogonal_spherical_shell_grid.jl
@@ -969,7 +969,7 @@ function Adapt.adapt_structure(to, grid::OrthogonalSphericalShellGrid)
                                                     adapt(to, grid.Azᶠᶜᵃ),
                                                     adapt(to, grid.Azᶜᶠᵃ),
                                                     adapt(to, grid.Azᶠᶠᵃ),
-                                                    adapt(to, grid.radius),
+                                                    grid.radius,
                                                     adapt(to, grid.conformal_mapping))
 end
 


glwagner (approved):

navidcy:
But question: floats need adaptation?

simone-silvestri:
Not really, it will call the fallback. But it's no cost to just have an adapt there

glwagner:
> But question: floats need adaptation?

The answer is no.

For any object that doesn't have an explicitly defined `adapt` method, then `adapt(to, x) = x`.

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -969,8 +969,8 @@ function Adapt.adapt_structure(to, grid::OrthogonalSphericalShellGrid)
                                                     adapt(to, grid.Azᶠᶜᵃ),
                                                     adapt(to, grid.Azᶜᶠᵃ),
                                                     adapt(to, grid.Azᶠᶠᵃ),
-                                                    grid.radius,
-                                                    grid.conformal_mapping)
+                                                    adapt(to, grid.radius),
```
glwagner, line 972: don't need to adapt this I think @navidcy is saying, unless we want to generalize the concept of `radius`?

glwagner: I agree there is no downside in terms of functionality though.

But making the code readable is actually more important than making it work.

Consider the absurd limit that the code _cannot_ be read (for example, it's closed source), but works.

In that case Oceananigans dies, because it cannot be developed...

This shows that making the code readable and understandable takes priority over any feature.

simone-silvestri: I don't agree that it makes the code less understandable though. I wouldn't look at the adapt function to understand what radius is. 
Also if it has no downsides (again, I don't agree that it hampers readability) I would prefer to not bake in assumptions in the code.

simone-silvestri: I have no problem removing that. Just, I don't think there is any benefit in specializing adapt functions with underlying assumptions

simone-silvestri: unless they are necessary (for example the adapt for a `field`)

glwagner: but @navidcy asked a question. that's a proof that it raises questions, isn't it?

glwagner: It looks like it must be `FT`: https://github.com/CliMA/Oceananigans.jl/blob/664fe17fdafb237617b5a487cb657e17f44ffd40/src/Grids/orthogonal_spherical_shell_grid.jl#L45

navidcy: I was sincerely confused tbh

simone-silvestri: ok I can remove the adapt from there.

PR #3620, opened 2024-06-13
Make it easier to build simple HydrostaticFreeSurfaceModels

glwagner:
This PR changes the default `buoyancy` and `tracers` for `HydrostaticFreeSurfaceModel` and also makes constructs like

```julia
velocities = PrescribedVelocityFields(u=1)
```

possible


glwagner:
Also should we call it `PrescribedVelocities`, rather than `PrescribedVelocityFields`?

glwagner:
> Also should we call it `PrescribedVelocities`, rather than `PrescribedVelocityFields`?

We'll save this for a future PR

simone-silvestri (approved):

PR #3630, opened 2024-06-19
Generalize `DiscreteBoundaryFunction` to functions of any arguments

glwagner:
This makes it so that discrete-form function boundary conditions

```
bc(i, j, grid, a, b, c, d)
```

will work provided that `fill_halo_regions!` is called with

```julia
fill_halo_regions!(field, a, b, c, d)
```

The trade-off here is between code clarity and generality. I added a comment to make it clear that at least for Oceananigans models, we use `clock, model_fields`.

@simone-silvestri wants to advertise Oceananigans as more generally applicable to any model using a finite volume grid, so this will help achieve that goal...

### DIFF
--- a/src/BoundaryConditions/discrete_boundary_function.jl
+++ b/src/BoundaryConditions/discrete_boundary_function.jl
@@ -35,18 +35,19 @@ const UnparameterizedDBF = DiscreteBoundaryFunction{<:Nothing}
 const UnparameterizedDBFBC = BoundaryCondition{<:Any, <:UnparameterizedDBF}
 const DBFBC = BoundaryCondition{<:Any, <:DiscreteBoundaryFunction}
 
-@inline getbc(bc::UnparameterizedDBFBC, i::Integer, j::Integer, grid::AbstractGrid, clock, model_fields, args...) =
-    bc.condition.func(i, j, grid, clock, model_fields)
+# In the functions below, we often have args = (clock, model_fields)
+@inline getbc(bc::UnparameterizedDBFBC, i::Integer, j::Integer, grid::AbstractGrid, args...) =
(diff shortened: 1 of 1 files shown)


glwagner:
Curious if anything breaks with this change

glwagner:
Ok, this change doesn't work because we actually pass the `closure` and `buoyancy` to `apply_flux_bcs!`, right now:

https://github.com/CliMA/Oceananigans.jl/blob/7f2d512ceaa2ce3fec3132cf0b08ab39344270d9/src/Models/HydrostaticFreeSurfaceModels/compute_hydrostatic_free_surface_tendencies.jl#L42-L43

This is for models like CATKE that have boundary conditions which depend on the closure... which extend `getbc`...

https://github.com/CliMA/Oceananigans.jl/blob/7f2d512ceaa2ce3fec3132cf0b08ab39344270d9/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_top_boundary_condition.jl#L19-L20

So I think to make this change we also need to generalize the implementation of `DiscreteBoundaryFunction`. We would need to have a type that indicates whether or not we should somehow truncate the `args` that are passed to `fill_halo_regions!` (in that generalization, the pattern on `main` would be the special case that all the args after the first 2 are discarded).

## Nota bene

A different solution one might invent would _copy_ `model.closure` and `model.buoyancy` into CATKE's `TKETopBoundaryConditionParameters`. The reason this doesn't work is that we support _changing_ the parameters of `model.closure` after creating the model. (This is a nifty feature that makes calibration a lot more efficient.) This means that we always need to get the closure parameters from `model.closure` to be consistent.

Issue #3625, opened 2024-06-18
Vector rotation for grids where the "intrinsic" and "extrinsic" coordinate system are different


glwagner:
Both `RectilinearGrid` and `LatitudeLongitudeGrid` have the property that the "extrinsic" coordinate system --- Cartesian for `RectilinearGrid`, and lat-lon for `LatitudeLongitudeGrid` --- coincide with the "intrinsic" coordinate system in which vector fields (like velocities) are expressed.

However, for more general grids, like the tripolar grid, the intrinsic coordinate system of the grid differs from the extrinsic coordinate system.

The extrinsic coordinate system is prescribed by us. We've decided to use lat-lon on the sphere and Cartesian otherwise. But the intrinsic coordinate system is a property of each grid independently.

The main challenge with grids like the tripolar grid is providing an interface to `set!` the velocity fields. I think we should define `set!` so that `u, v, w` are assumed to be provided in the extrinsic coordinate system. This means that the provided velocities need to be rotated under the hood into the intrinsic coordinate system.

One of the basic features we need is a function that rotates a vector field:

```julia
ue, ve, we = extrinsic_vector(i, j, k, grid, ui, vi, wi)
```

and

```julia
ui, vi, wi = intrinsic_vector(i, j, k, grid, ue, ve, we)
```

We may also want something like `intrinsic_vector_from_centered` and `intrinsic_vector_from_staggered` etc...

Let's split the way we add this functionality into a few PRs, eg first develop `extrinsic_vector` and `intrinsic_vector`, and then in a new PR work on `set!`. 

cc @simone-silvestri @siddharthabishnu

navidcy:
OK, I think I understand the issue mentioned here but the terminology intrinsic and extrinsic is not self-explanatory (for my intuition at least). 

I'm not trying to argue against the names, I just want to understand exactly what's been discussed.

I think something in this spirit @siddharthabishnu bumped into in trying to set initial conditions or surface forcing on a ConformalCubedSphereGrid..?

glwagner:
Extrinsic is what we decide. When we use the cubed sphere we want to use geographic coordinates.

But vector fields on the c-grid have their own coordinate system dictated by the grid geometry. Except for `LatitudeLongitudeGrid`, grid vectors are not in the geographic coordinate system.

[...]

PR #3631, opened 2024-06-19
Conversion between "Intrinsic" and "Extrinsic" reference frame

simone-silvestri:
This PR starts the implementation of the conversion between frames described in #3625

cc @glwagner @siddharthabishnu


src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,127 @@
...
+Convert the three-dimensional vector with components `uₑ, vₑ, wₑ` defined in an _extrinsic_ 
+reference frame associated with the domain, to the reference frame _intrinsic_ to the grid. 
+
+_extrinsic_ reference frames are:
+
+- Cartesian for box domains 
+- Latitude - Longitude for spherical domains
+
+Therefore, for the `RectilinearGrid` and the `LatitudeLongitudeGrid`, the _extrinsic_ and the 
+_intrinsic_ reference frames are equivalent
```
navidcy, line 20:
```suggestion
Therefore, for the [`RectilinearGrid`](@ref) and the [`LatitudeLongitudeGrid`](@ref), the _extrinsic_ and the 
_intrinsic_ reference frames are equivalent. However, for other grids (e.g., for the
 [`ConformalCubedSphereGrid`](@ref)) that might not be the case.
```

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,127 @@
...
+"""
+@inline intrinsic_vector(i, j, k, grid::AbstractGrid, uₑ, vₑ, wₑ) = 
+    getvalue(uₑ, i, j, k, grid), getvalue(vₑ, i, j, k, grid), getvalue(wₑ, i, j, k, grid)
+
+"""
+    extrinsic_vector(i, j, k, grid::AbstractGrid, uᵢ, vᵢ, wᵢ)
+
+Convert the three-dimensional vector with components `uₑ, vₑ, wₑ` defined on the reference 
+reference frame _intrinsic_ to the grid, to the _extrinsic_ reference frame associated with the domain.
```
navidcy, line 29:
```suggestion
Convert the three-dimensional vector with components `uᵢ, vᵢ, wᵢ ` defined on the _intrinsic_ reference 
frame of the grid, to the _extrinsic_ reference frame associated with the domain.
```

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,127 @@
...
+
+_extrinsic_ reference frames are:
+
+- Cartesian for box domains 
+- Latitude - Longitude for spherical domains
+
+Therefore, for the `RectilinearGrid` and the `LatitudeLongitudeGrid`, the _extrinsic_ and the 
+_intrinsic_ reference frames are equivalent
```
navidcy, line 37:
```suggestion
Therefore, for the [`RectilinearGrid`](@ref) and the [`LatitudeLongitudeGrid`](@ref), the _extrinsic_ and the 
_intrinsic_ reference frames are equivalent. However, for other grids (e.g., for the
 [`ConformalCubedSphereGrid`](@ref)) that might not be the case.
```

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,127 @@
...
+
+    u  = getvalue(uₑ, i, j, k, grid)
+    v  = getvalue(vₑ, i, j, k, grid)
+    wᵢ = getvalue(wₑ, i, j, k, grid)
+
+    d₁ = ũ / 𝒰
+    d₂ = ṽ / 𝒰
+
+    uᵢ = u * d₁ + v * d₂
+    vᵢ = u * d₂ - v * d₁
```
navidcy, line 73: I'm a bit confused to what's happening here.
If I understand correctly, there is a local rotation to get uᵢ, vᵢ from u, v. Right?
So d₁ and d₂ are the sin and cos of an angle? Could we rename them so that's clear?
What is ũ and ṽ supposed to be? They seem to be unrelated to u and v but their name suggests they are.

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,127 @@
...
+
+    u  = getvalue(uᵢ, i, j, k, grid)
+    v  = getvalue(vᵢ, i, j, k, grid)
+    wₑ = getvalue(wᵢ, i, j, k, grid)
+
+    d₁ = ũ / 𝒰
+    d₂ = ṽ / 𝒰
+
+    uₑ = u * d₁ - v * d₂
+    vₑ = u * d₂ + v * d₁
```
navidcy, line 102: same remark as `intrinsic_vector` method

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,127 @@
...
+    @inbounds intrinsic_vector(i, j, k, grid, uₑ, vₑ, wₑ)[3]
+
+@inline extrinsic_vector_x_component(i, j, k, grid::AbstractGrid, uₑ, vₑ, wₑ) =
+    @inbounds intrinsic_vector(i, j, k, grid, uₑ, vₑ, wₑ)[1]
+    
+@inline extrinsic_vector_y_component(i, j, k, grid::AbstractGrid, uₑ, vₑ, wₑ) =
+    @inbounds intrinsic_vector(i, j, k, grid, uₑ, vₑ, wₑ)[2]
+
+@inline extrinsic_vector_z_component(i, j, k, grid::AbstractGrid, uₑ, vₑ, wₑ) =
+    @inbounds intrinsic_vector(i, j, k, grid, uₑ, vₑ, wₑ)[3]
```
navidcy, line 127:
```suggestion
    @inbounds intrinsic_vector(i, j, k, grid, uₑ, vₑ, wₑ)[3]

```

src/Operators/vector_rotation_operators.jl
glwagner: Are they unit vectors? Agree that more care with the names will go a long way towards interpreting this code. `x̂, ŷ` are nice names for unit vectors.

glwagner: We also need to come up with notation for the intrinsic / local unit vectors vs extrinsic (in this case, geographic) unit vectors. I propose something like `ξ̂, η̂` for the intrinsic vectors.

simone-silvestri: I have clarified the names

src/Operators/vector_rotation_operators.jl
simone-silvestri: changed names as the remark above

simone-silvestri:
This should be ready to merge

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,157 @@
...
+# This would be a generalization of `getbc` that could be used everywhere we need it
+@inline getvalue(::Nothing,        i, j, k, grid, args...) = nothing
+@inline getvalue(a::Number,        i, j, k, grid, args...) = a
+@inline getvalue(a::AbstractArray, i, j, k, grid, args...) = @inbounds a[i, j, k]
+
+"""
+    intrinsic_vector(i, j, k, grid::AbstractGrid, uₑ, vₑ, wₑ)
+
+Convert the three-dimensional vector with components `uₑ, vₑ, wₑ` defined in an _extrinsic_ 
+reference frame associated with the domain, to the reference frame _intrinsic_ to the grid.
```
glwagner, line 12: reference frame or coordinate system?

typically I think "reference frame" refers to relative motion, ie we can be in a reference frame that rotates with the Earth, or is stationary or inertial. 

pretty sure the issue here is the coordinate system, not the reference frame.

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,157 @@
...
+
+"""
+    intrinsic_vector(i, j, k, grid::AbstractGrid, uₑ, vₑ, wₑ)
+
+Convert the three-dimensional vector with components `uₑ, vₑ, wₑ` defined in an _extrinsic_ 
+reference frame associated with the domain, to the reference frame _intrinsic_ to the grid. 
+
+_extrinsic_ reference frames are:
+
+- Cartesian for box domains
```
glwagner, line 16:
```suggestion
- Cartesian for RectilinearGrid
```

nowhere do we use the term "box domain"?

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,157 @@
...
+"""
+    intrinsic_vector(i, j, k, grid::AbstractGrid, uₑ, vₑ, wₑ)
+
+Convert the three-dimensional vector with components `uₑ, vₑ, wₑ` defined in an _extrinsic_ 
+reference frame associated with the domain, to the reference frame _intrinsic_ to the grid. 
+
+_extrinsic_ reference frames are:
+
+- Cartesian for box domains 
+- Latitude - Longitude for spherical domains
```
glwagner, line 17:
```suggestion
- Geographic coordinates for AbstractCurvilinearGrid
```

right?

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,157 @@
...
+ [`ConformalCubedSphereGrid`](@ref)) that might not be the case.
+"""
+@inline intrinsic_vector(i, j, k, grid::AbstractGrid, uₑ, vₑ, wₑ) = 
+    getvalue(uₑ, i, j, k, grid), getvalue(vₑ, i, j, k, grid), getvalue(wₑ, i, j, k, grid)
+
+"""
+    extrinsic_vector(i, j, k, grid::AbstractGrid, uᵢ, vᵢ, wᵢ)
+
+Convert the three-dimensional vector with components `uᵢ, vᵢ, wᵢ ` defined on the _intrinsic_ reference 
+frame of the grid, to the _extrinsic_ reference frame associated with the domain.
```
glwagner, line 30: same comment as above regarding "reference frame" vs "coordinate system"

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,157 @@
...
+
+_extrinsic_ reference frames are:
+
+- Cartesian for box domains 
+- Latitude - Longitude for spherical domains
```
glwagner, line 35: same comment as above

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,157 @@
...
+ [`ConformalCubedSphereGrid`](@ref)) that might not be the case.
+"""
+@inline extrinsic_vector(i, j, k, grid::AbstractGrid, uᵢ, vᵢ, wᵢ) =
+    getvalue(uᵢ, i, j, k, grid), getvalue(vᵢ, i, j, k, grid), getvalue(wᵢ, i, j, k, grid)
+
+# 2D vectors
+@inline intrinsic_vector(i, j, k, grid::AbstractGrid, uₑ, vₑ) = 
+    getvalue(uₑ, i, j, k, grid), getvalue(vₑ, i, j, k, grid)
+
+@inline entrinsic_vector(i, j, k, grid::AbstractGrid, uᵢ, vᵢ) =
```
glwagner, line 48:
```suggestion
@inline extrinsic_vector(i, j, k, grid::AbstractGrid, uᵢ, vᵢ) = 
```

or is it `intrinsic`? Is this tested?

src/Operators/vector_rotation_operators.jl
simone-silvestri: How I see it is that the _extrinsic_ coordinate system is related to the Domain (in this case rectangular, cartesian domains), while the _intrinsic_ coordinate system is related to the grid (aka the discretization). 
For this reason it would be

simone-silvestri: This distinction does not matter at the moment because we only have one `RectlinearGrid`, but it matters conceptually (for example if we have an unstructured grid on a box domain it would still have a cartesian extrinsic reference frame).
I tried to make the explanation more

simone-silvestri:
This should be ready to merge

src/Operators/vector_rotation_operators.jl
```diff
@@ -0,0 +1,157 @@
...
+
+    return uₑ, vₑ, wₑ
+end
+
+#####
+##### Component-wise conversion between reference frames
+#####
+
+@inline intrinsic_vector_x_component(i, j, k, grid::AbstractGrid, uₑ, vₑ, wₑ) = 
+    @inbounds intrinsic_vector(i, j, k, grid, uₑ, vₑ, wₑ)[1]
```
glwagner, line 142: where are these used? Why did you decide on `intrinsic_vector_x_component` rather than `x_intrinsic_vector` or `intrinsic_vector_x` both of which are more common in the Oceananigans code base? 

Why do we not implement `intrinsic_vector_x_component` and then construct `intrinsic_vector` from that? It seems more natural than to write `intrinsic_vector` for all three components and then break it down like this.

simone-silvestri: these are just used in the tests, we probably it is better to remove them from the src and define them where they are used. I though `intrinsic_vector_x_component` makes more sense because `x_vector` or `vector_x` does not really tell me that it is the x-component (sounds more like a vector that has to do with the x direction).

Regarding the second question, the vector transformation requires all the components and naturally returns all the components in case a transformation is needed (if we use the thin shell approximation then only the horizonal ones are needed). So going component-wise will double (or triple) the amount of computation.

glwagner: Ok, but then what is the point of the function. I think `intrinsic_vector(i, j, k, grid, uₑ, vₑ, wₑ)[1]` is actually more clear than `intrinsic_vector_x_component`

simone-silvestri: you are right, I needed a function with a `Number` return value to be used in the tests in combination with `KernelFunctionOperation`. I have moved this function outside the `src` and inside the tests

glwagner (approved):

Issue #3637, opened 2024-06-24
More / more flexible Smagorinsky models


tomchor:
I have been thinking that we should expand a bit our Smagorinsky models.
Currently the only Smagorinsky-type model we have is the Smagorinsky-Lilly, which works well in some cases, notably the canonical horizontally-periodic turbulence, but has some downsides for more complex flows. For example when turbulence is not heterogeneous, or when there are persistent mean gradients in the flow. AMD is a good option in some cases, but often there are numerical artifacts in stratified, laminar regions in the flow (I don't think it's clear exactly what causes this, but it happens in many of my simulations, and I've herd similar things from others). So I think having at least one option to run a dynamic Smag would be good.

One possibility is that, as a start, we allow for user-defined Smagorinsky coefficients that could depend on the velocities. Something like

```julia
cₛ(i, j, k, grid, C::Function) = C(i, j, k, grid, u, v, w)
```
where `C(i, j, k, grid, u, v, w)` is provided by the user. That would allow the user a bit more control, and they could implement local dynamic Smag models, such as the [dynamic gradient Smagorinsky](https://journals.aps.org/prfluids/abstract/10.1103/PhysRevFluids.7.074604). Something similar is currently done for AMD, although I think it's a bit more restrictive than what I'm proposing.

Soon, if there's interest, I'm planning on implementing a scale-independent Smagorinsky that's averaged in the symmetric flow directions (generally this is referred to as Planar-Averaged Scale-Independent Smag, but we wouldn't be limited to a planar average in our case). From then on, the extension to a Langrangian-Averaged Scale-Independent Smag would be straightforward.

Eventually it would be nice to extend these to a Scale-_dependent_ model, but I think it's wise to start with the simpler cases.

How do people feel about this?

glwagner:
You can start by allowing the coefficient to be an array as in AMD:

https://github.com/CliMA/Oceananigans.jl/blob/7f2d512ceaa2ce3fec3132cf0b08ab39344270d9/src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl#L143

Then with a callback you can implement any model, including dynamic Smagorinsky. So this provides unlimited flexibility in developing a new closure.

I think its possible you will find that its more efficient to precompute (eg in a callback) the coefficient rather than compute it on the fly with a function, especially because the computation of the different filters is a bit non-local, eg you have to average over a few grid points, which could involve relatively expensive memory fetches.

[...]

PR #3638, opened 2024-06-24
Implementation of the scale-dependent Lagrangian-averaged dynamic Smagorinsky

xkykai:
Closes https://github.com/CliMA/Oceananigans.jl/issues/3637

This is an implementation of the scale-dependent Lagrangian-averaged dynamic Smagorinsky model as outlined in section C III. in [Bou-Zeid et al. 2005](https://dx.doi.org/10.1063/1.1839152).

@simone-silvestri @tomchor


src/TurbulenceClosures/turbulence_closure_implementations/lagrangian_dynamic_smagorinsky.jl
```diff
@@ -225,6 +225,8 @@ end
     u̅w̅ = ℑxyzᶜᶜᶜ(i, j, k, grid, ℑxyzᶠᶠᶠ, ℑxyzᶜᶜᶜ, ℑxyzᶠᶠᶠ, uᵢuⱼ, ℑxᶜᵃᵃ, ℑzᵃᵃᶜ, u, w)
     v̅w̅ = ℑxyzᶜᶜᶜ(i, j, k, grid, ℑxyzᶠᶠᶠ, ℑxyzᶜᶜᶜ, ℑxyzᶠᶠᶠ, uᵢuⱼ, ℑyᵃᶜᵃ, ℑzᵃᵃᶜ, v, w)
 
+    S̅ₘ = ℑxyzᶜᶜᶜ(i, j, k, grid, ℑxyzᶠᶠᶠ, ℑxyzᶜᶜᶜ, ℑxyzᶠᶠᶠ, ϕ⁰⁵, ΣᵢⱼΣᵢⱼᶜᶜᶜ, u, v, w)
```
xkykai, line 228: Currently the code goes into `StackOverflowError`, likely due to the way the filter averaging is done as we might have function compositions which are too deep.

tomchor:
We should "Close https://github.com/CliMA/Oceananigans.jl/issues/3637" with this one I think

src/TurbulenceClosures/turbulence_closure_implementations/lagrangian_dynamic_smagorinsky.jl
```diff
@@ -0,0 +1,319 @@
...
+    # is one because the model is assumed to be scale - invariant
+    α = 4
+    β = 1
+
+    M₁₁ = 2Δᶠ * (S̅S̅₁₁ - α^2 * β^2 * S̅ₘ * S̅₁₁)
+    M₂₂ = 2Δᶠ * (S̅S̅₂₂ - α^2 * β^2 * S̅ₘ * S̅₂₂)
+    M₃₃ = 2Δᶠ * (S̅S̅₃₃ - α^2 * β^2 * S̅ₘ * S̅₃₃)
+    M₁₂ = 2Δᶠ * (S̅S̅₁₂ - α^2 * β^2 * S̅ₘ * S̅₁₂)
+    M₁₃ = 2Δᶠ * (S̅S̅₁₃ - α^2 * β^2 * S̅ₘ * S̅₁₃)
+    M₂₃ = 2Δᶠ * (S̅S̅₂₃ - α^2 * β^2 * S̅ₘ * S̅₂₃)
```
tomchor, line 267: Should this be

```suggestion
    M₁₁ = 2Δᶠ * (S̅S̅₁₁ - α^2 * β * S̅ₘ * S̅₁₁)
    M₂₂ = 2Δᶠ * (S̅S̅₂₂ - α^2 * β * S̅ₘ * S̅₂₂)
    M₃₃ = 2Δᶠ * (S̅S̅₃₃ - α^2 * β * S̅ₘ * S̅₃₃)
    M₁₂ = 2Δᶠ * (S̅S̅₁₂ - α^2 * β * S̅ₘ * S̅₁₂)
    M₁₃ = 2Δᶠ * (S̅S̅₁₃ - α^2 * β * S̅ₘ * S̅₁₃)
    M₂₃ = 2Δᶠ * (S̅S̅₂₃ - α^2 * β * S̅ₘ * S̅₂₃)
```
?

src/TurbulenceClosures/turbulence_closure_implementations/lagrangian_dynamic_smagorinsky.jl
```diff
@@ -0,0 +1,319 @@
...
```
tomchor, line 314:
```suggestion
    M₁₁ = 2Δᶠ * (S̅S̅₁₁ - α^2 * β * S̅ₘ * S̅₁₁)
    M₂₂ = 2Δᶠ * (S̅S̅₂₂ - α^2 * β * S̅ₘ * S̅₂₂)
    M₃₃ = 2Δᶠ * (S̅S̅₃₃ - α^2 * β * S̅ₘ * S̅₃₃)
    M₁₂ = 2Δᶠ * (S̅S̅₁₂ - α^2 * β * S̅ₘ * S̅₁₂)
    M₁₃ = 2Δᶠ * (S̅S̅₁₃ - α^2 * β * S̅ₘ * S̅₁₃)
    M₂₃ = 2Δᶠ * (S̅S̅₂₃ - α^2 * β * S̅ₘ * S̅₂₃)
```

tomchor:
I'm curious as to why you picked `α=4`. It makes for a pretty big stencil and `α=2` works pretty well already. Does it have to do with the filtering being two consecutive interpolations?

simone-silvestri:
This is also not super clean because the forcing functions for the additional tracers are added manually to the model. 
It would be better to have a way to include them directly when prescribing this closure.

xkykai:
I believe what we are doing here is using two filter scales `α=2` and `α=2` to compute `β`, which is ratio of coefficients at different filter scales `Δ`, `αΔ`, and `α²Δ`, which assumes that `β` is scale-invariant.

tomchor:
> I believe what we are doing here is using two filter scales `α=2` and `α=2` to compute `β`, which is ratio of coefficients at different filter scales `Δ`, `αΔ`, and `α²Δ`, assuming that `β` is scale-invariant.

Hmm, then I'm confused. When you assume `β` to be scale-invariant you only need two filters: the grid (`Δ`) and a test filter (`αΔ`). afaik the second test filter is only needed when assuming that `β` can vary between scales.

src/TurbulenceClosures/turbulence_closure_implementations/lagrangian_dynamic_smagorinsky.jl
simone-silvestri: we can maybe use a box filter here instead of all this combination of interpolations. 
We can add a filter function to the `struct`

tomchor:
> > afaik the second test filter is only needed when assuming that `β` can vary between scales.
> 
> I think you're correct here! Sorry for the confusion here, but I suppose by scale-invariance I mean `β` has a power law dependence on the scale size. And since we were trying to implement the scale-dependent version, `β` would vary across different scales? Hence we need the `α=4` filter as well. Am I understanding Equation (26) in [Bou-Zeid et al. 2005](https://dx.doi.org/10.1063/1.1839152) correctly here?

Ah, I see, I was under the impression you were trying to implement the scale-_invariant_ version, but it does make sense to if you're implementing the scale-dependent version.

src/TurbulenceClosures/turbulence_closure_implementations/lagrangian_dynamic_smagorinsky.jl
glwagner: Hmm. But this should not be an overflow error, right? Are we hitting some heuristic?

I would try combining steps to see what happens. For example you could define

```julia
ϕ⁰⁵_ΣᵢⱼΣᵢⱼᶜᶜᶜ(i, j, k, grid, u, v, w) = ϕ⁰⁵(i, j, k, grid, ΣᵢⱼΣᵢⱼᶜᶜᶜ, u, v, w)
```

You could also simply write the whole thing manually. With composition this deep, it might also make this code easier to understand.

glwagner: > Does it work if we use a simpler filter?

What is complicated about this?

validation/stratified_couette_flow/stratified_couette_flow.jl
```diff
@@ -16,6 +16,18 @@ using Oceananigans.TurbulenceClosures:
...
+Nxy=64
+Nz=32
+arch=CPU()
+h=1
+U_wall=1
+Re=4250
+Pr=0.7
+Ri = 0.5
+Ni=10
+end_time=1000
```
glwagner, line 29: What is going on here? Isn't there already a script for invoking the utilities in this script?

validation/stratified_couette_flow/stratified_couette_flow.jl
```diff
@@ -113,7 +125,7 @@ function simulate_stratified_couette_flow(; Nxy, Nz, arch=GPU(), h=1, U_wall=1,
     ##### Impose boundary conditions
     #####
 
-    grid = RectilinearGrid(arch, size = (Nxy, Nxy, Nz), extent = (4π*h, 2π*h, 2h))
+    grid = RectilinearGrid(arch, size = (Nxy, Nxy, Nz), extent = (4π*h, 2π*h, 2h), halo = (6, 6, 6))
```
glwagner, line 128:
```suggestion
    grid = RectilinearGrid(arch; size = (Nxy, Nxy, Nz), extent = (4π*h, 2π*h, 2h), halo)
```

make this cleaner by adding `halo` as an argument to `simulate_stratified_couette_flow`.

Also, it doesn't look like the `closure` is an argument to `simulate_stratified_couette_flow`. You'll need that too. (You can also infer the halo size.)

Just as a general tip, I think you will save time if you generalize these functions first. Now, you are in the position of having to 1) walk back the chances you made and then 2) implement an interface that allows different closures. This is actually _more_ work then if you had simply done (2) to begin with.

src/TurbulenceClosures/turbulence_closure_implementations/lagrangian_dynamic_smagorinsky.jl
glwagner: Can you paste in the reference that shows this?

tomchor: Taking a second look at things I now think there may be more typos than this one. So I suggest @xkykai and @simone-silvestri double-check this, but if I understand correctly, what's called `𝒥ᴿᴺ` in this PR is called $\mathcal J^{QN}$in [Bou-Zeid et al. (2005)](https://dx.doi.org/10.1063/1.1839152), and here's its advection equation:

image

so using the language of this PR the forcing here should be
```julia
1 / T4Δ * (QᵢⱼNᵢⱼ - 𝒥ᴿᴺ)
```
instead of what's written below: `1 / TΔ * (LᵢⱼMᵢⱼ - 𝒥ᴿᴺ)`.

And the functions $Q_{ij}$ and $N_{ij}$ are:

image

Similar comments are valid for the function below (`𝒥ᴺᴺ_forcing_function()`).

tomchor: For the record, my original comment had to do with $M_{ij}$ being implemented as `M₁₁ = 2Δᶠ * (S̅S̅₁₁ - α^2 * β^2 * S̅ₘ * S̅₁₁)` when it should be

image

but now (per my previous comment) I realize that $M_{ij}$ isn't supposed to be in this forcing function at all.

simone-silvestri: The difference between Q and L (and M and N) is the filter size. The stresses for the `𝒥ᴿᴺ` forcings are averaged twice as much as the `𝒥ᴸᴹ` forcings (if you look at how the averages are calculated). However, I would say that to push this PR forward we need to:
- define an averaging operators (to be held in the closure struct) for the `M - L` fluxes and for the `R - N` fluxes
- remove all this combination of interpolations in favor of those averaging operators

tomchor: > The difference between Q and L (and M and N) is the filter size.

Agreed. The $N$ and $M$ nomenclature difference makes it clearer though. That said, the combination $\alpha^2 \beta^2$ shouldn't appear anywhere, right? It's either $\alpha^4 \beta^2$ or $\alpha^2 \beta$ (unless you hard-coded some sutff I guess)

glwagner: What do the tilde, hat, and bar mean?

glwagner:
> This is also not super clean because the forcing functions for the additional tracers are added manually to the model. It would be better to have a way to include them directly when prescribing this closure.

I suggest implementing the tracer-specific forcing in a similar manner that we do for CATKE and k-epsilon

src/TurbulenceClosures/turbulence_closure_implementations/lagrangian_dynamic_smagorinsky.jl
tomchor: - Tilde: LES-filtered velocity. In our case it's simply the velocities that come out of `model.velocities`
- Bar: First test filter
- Hat: Second test filter (at a larger spatial scale than the first)

glwagner:
should we close this since we have a dynamic smag implementation? Maybe this PR could be refactored, but it seems like it might be better to start fresh

tomchor:
Agreed!

PR #3639, opened 2024-06-25
`CuArray` `times` for `Cyclical` field time series

simone-silvestri:
this PR allows using a `CuArray` times in conjunction with a `Cyclical{Nothing}` time_indexing


glwagner:
We aren't testing this I guess? Might be good to add since we do try to convert arrays to `CuArray` right?

simone-silvestri:
should be ready to merge

src/OutputReaders/field_time_series_indexing.jl
```diff
@@ -227,6 +229,18 @@ end
 ##### FieldTimeSeries updating
 #####
 
+# Let's make sure `times` is available on the CPU. This is always the case 
+# for ranges. if `times` is a vector that resides on the GPU, it has to be moved to the CPU  for safe indexing. 
+# Copying the whole array is a bit unclean but considering `update_field_time_series!` is called quite rarely 
+# it should not impact too much memory / performance
```
glwagner, line 235: This is an assumption. We have no idea how often `update_field_time_series!` is called.

glwagner: You mean to say "for the simulations that I personally want to run".

I think you should rephrase this comment to make it clear that there is a penalty involved here, and we choose a method based on a certain assumption, and perhaps propose a TODO to fix it if use cases emerge that require it.

src/OutputReaders/field_time_series_indexing.jl
```diff
@@ -227,6 +229,18 @@ end
...
+cpu_safe_interpolating_time_indices(time_indexing, times, t, arch)  = interpolating_time_indices(time_indexing, times, t)
```
glwagner, line 236:
```suggestion
cpu_safe_interpolating_time_indices(arch, times, time_indexing, times, t)  = interpolating_time_indices(time_indexing, times, t)
```

I think this is a better ordering indicating the relative importance of dispatch on the different types.

Also isn't this function "gpu safe?" The issue is when `times` is on GPU. When it's on the CPU there's no need for a special method. You may want to rename it to `gpu_safe_interpolating_time_indices`

simone-silvestri: the issue is when you are interpolating on the CPU indices on the GPU. So the interpolation should be CPU-available

glwagner:
Still confused. Do we simply need to delete `times = on_architecture(architecture(grid), times)` in the constructor?

src/OutputReaders/field_time_series_indexing.jl
glwagner: Still confused. Do we simply need to delete times = on_architecture(architecture(grid), times) in the constructor?

src/OutputReaders/field_time_series_indexing.jl
```diff
@@ -227,6 +229,18 @@ end
 ##### FieldTimeSeries updating
 #####
 
+# Let's make sure `times` is available on the CPU. This is always the case 
+# for ranges. if `times` is a vector that resides on the GPU, it has to be moved to the CPU for safe indexing. 
+# TODO: Copying the whole array is a bit unclean, maybe find a way that avoids the penalty of allocating and copying memory.
+# This would require refactoring `FieldTimeSeries` to include a cpu-allocated times array
+cpu_available_interpolating_time_indices(::CPU, times, time_indexing, t, arch) = interpolating_time_indices(time_indexing, times, t)
```
glwagner, line 236: Maybe just `cpu_interpolating_time_indices` ("cpu available" doesn't totally compute for me)

src/OutputReaders/field_time_series_indexing.jl
```diff
@@ -227,6 +229,18 @@ end
...
+cpu_available_interpolating_time_indices(::CPU, times::AbstractVector, time_indexing, t) = interpolating_time_indices(time_indexing, times, t)
+
+function cpu_available_interpolating_time_indices(::GPU, times::AbstractVector, time_indexing, t) 
+    cpu_times = on_architecture(CPU(), times)
+    return interpolating_time_indices(time_indexing, cpu_times, t)
```
glwagner, line 241: Would it be better to launch a small kernel, eg something like

```julia
ñ = Ref(1.0) # ?
n₁ = Ref(1)
n₂ = Ref(1)

# launch kernel that stores indices in `ñ, n₁, n₂`
```

src/OutputReaders/field_time_series_indexing.jl
simone-silvestri: When `times` is not a range, we need it as a `CuArray` in the kernels (somewhat equivalently to a coordinate like `λᶠᵃᵃ` for grids). What breaks the code is trying to index `times` on the CPU while interpolating

src/OutputReaders/field_time_series_indexing.jl
simone-silvestri: I had tried that, but it seems to not work with `Ref`, allocating a `CuArray` to hold an index seems a bit much. I am not sure if there is a type that allows exchanging data from CPU to GPU inside a kernel

glwagner: I think that's what is done when doing a global reduction over all dimensions, eg

https://github.com/JuliaGPU/GPUArrays.jl/blob/cd1f59ae2fd17763ca73505f966b980bf7135354/src/host/mapreduce.jl#L71

glwagner (approved):

PR #3642, opened 2024-06-28
Implement scale-invariant dynamic Smagorinsky

tomchor:
Closes https://github.com/CliMA/Oceananigans.jl/issues/3637

This PR implements the scale-invariant dynamic Smagorinsky. My main reference for this was [Bou-Zeid et al. (2005)](https://dx.doi.org/10.1063/1.1839152), just because it's nicely summarized there.

As a refresher, the coefficient in this case is calculated as

```math
c_s^2 = \frac{\langle L_{ij} M_{ij}\rangle}{\langle M_{ij} M_{ij} \rangle},
```
where $M_{ij}$ and $L_{ij}$ are obtained by applying the Germano identity and minimizing the error. $\langle \cdot \rangle$ is an average which usually is implemented as planar averaging. I'm implementing it here with an arbitrary `DirectionalAveraging` procedure that can average in whatever direction the user wants. Following a suggestion by @glwagner in #3637, I'm also starting to put the pieces in place for a more general averaging procedure so that eventually we can add Lagrangian averaging as an option for this model.

Note that this PR is related to https://github.com/CliMA/Oceananigans.jl/pull/3638. The rationale for having this PR in addition to #3638 is that for most LES are doubly periodic, in which case there's not much advantage in having the Lagrangian averaging that's implemented in #3638. In these cases I'd argue the model implemented here is more efficient, so having both types of models is probably desirable.

Another note is that, once this PR is merged, it should be very straightforward to implement a scale-_dependent_ version, which works better close to boundaries.

The model seems to be working. Here's an animation of an unforced 3D turbulence simulation on a 32³ grid. Left is vorticity and right is the strain rate modulus squared:

https://github.com/CliMA/Oceananigans.jl/assets/13205162/7131a99d-df6c-4883-850d-d4a87988cdb7

Note that the value of the calculated Smag coefficient $c_s$ (which I'm showing at the top of the plots above) is about 0.17, which is very close to the theoretical value obtained by Lilly of 0.16.

I'm opening this as a draft PR for now because there are some things that need doing:

- [x] Generalize the filter for stretched grids. For now it assumes a regular grid for simplicity, but it's trivial to generalize.
- [x] Optimize the calculation of the coefficient. At the moment I'm creating four extra fields in order to calculate the Smag coefficient: `LM`, `MM` are 3D fields; and `LM_avg` and `MM_avg` are 1D or 2D. What I'm doing is to first calculate $L_{ij} M_{ij}$ and $M_{ij} M_{ij}$ pointwise, and then `LM_avg` and `MM_avg` receive their averages. We should be able to calculate everything without needing `LM`, `MM` are 3D fields, I just couldn't figure out how yet :)
- [ ] Write docs
- [x] Write tests
- [x] Validate that model is working as intended

CC @glwagner @simone-silvestri @xkykai @whitleyv 

Feel free to add more things to the to-do list that I may have forgotten.


src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -0,0 +1,331 @@
...
+    @inbounds νₑ[i, j, k] = cₛ² * (Δᶠ(i, j, k, grid))^2 * sqrt(2Σ²)
+end
+
+@inline directional_average!(averaged_field, field,   ::XDirectionalAveraging) = averaged_field .= compute!(Field(Average(field, dims=1)))
+@inline directional_average!(averaged_field, field,   ::YDirectionalAveraging) = averaged_field .= compute!(Field(Average(field, dims=2)))
+@inline directional_average!(averaged_field, field,   ::ZDirectionalAveraging) = averaged_field .= compute!(Field(Average(field, dims=3)))
+@inline directional_average!(averaged_field, field,  ::XYDirectionalAveraging) = averaged_field .= compute!(Field(Average(field, dims=(1, 2))))
+@inline directional_average!(averaged_field, field,  ::XZDirectionalAveraging) = averaged_field .= compute!(Field(Average(field, dims=(1, 3))))
+@inline directional_average!(averaged_field, field,  ::YZDirectionalAveraging) = averaged_field .= compute!(Field(Average(field, dims=(2, 3))))
+@inline directional_average!(averaged_field, field, ::XYZDirectionalAveraging) = averaged_field .= compute!(Field(Average(field, dims=(1, 2, 3))))
```
tomchor, line 87: This part definitely needs to be changed. I just haven't been able to come up an appropriate kernel that works.

glwagner: What kernel? If you want to compute an average then you shouldn't write your own kernel. You probably want to preallocate the result though. The code will be a lot simpler since then you just call `compute!`.

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -0,0 +1,330 @@
+#####
+##### The turbulence closure proposed by Smagorinsky and Lilly.
+##### We also call this 'Constant Smagorinsky'.
+#####
+using Oceananigans.Operators: volume
+using Oceananigans.Grids: XBoundedGrid, YBoundedGrid, ZBoundedGrid, XYBoundedGrid, XZBoundedGrid, YZBoundedGrid, XYZBoundedGrid
+
+abstract type AveragingProcedure end
```
glwagner, line 8: abstract types are always written with the prefix `Abstract`, eg `AbstractAveragingProcedure`

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -0,0 +1,330 @@
...
+    arch = model.architecture
+    grid = model.grid
+    buoyancy = model.buoyancy
+    velocities = model.velocities
+    tracers = model.tracers
+
+    launch!(arch, grid, parameters, compute_LM_MM!,
+            diffusivity_fields.LM, diffusivity_fields.MM, grid, closure, velocities)
+
+    # TODO: optimize this
```
glwagner, line 98: optimize in what way?

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -0,0 +1,330 @@
...
+
+# TODO: Generalize filter to stretched directions
+AG = AbstractGrid
+@inline ℱx²ᵟ(i, j, k, grid::AG{FT}, ϕ) where FT = @inbounds FT(0.5) * ϕ[i, j, k] + FT(0.25) * (ϕ[i-1, j, k] + ϕ[i+1, j,  k])
+@inline ℱy²ᵟ(i, j, k, grid::AG{FT}, ϕ) where FT = @inbounds FT(0.5) * ϕ[i, j, k] + FT(0.25) * (ϕ[i, j-1, k] + ϕ[i,  j+1, k])
+@inline ℱz²ᵟ(i, j, k, grid::AG{FT}, ϕ) where FT = @inbounds FT(0.5) * ϕ[i, j, k] + FT(0.25) * (ϕ[i, j, k-1] + ϕ[i,  j, k+1])
+
+@inline ℱx²ᵟ(i, j, k, grid::AG{FT}, f::F, args...) where {FT, F<:Function} = FT(0.5) * f(i, j, k, grid, args...) + FT(0.25) * (f(i-1, j, k, grid, args...) + f(i+1, j, k, grid, args...))
+@inline ℱy²ᵟ(i, j, k, grid::AG{FT}, f::F, args...) where {FT, F<:Function} = FT(0.5) * f(i, j, k, grid, args...) + FT(0.25) * (f(i, j-1, k, grid, args...) + f(i, j+1, k, grid, args...))
+@inline ℱz²ᵟ(i, j, k, grid::AG{FT}, f::F, args...) where {FT, F<:Function} = FT(0.5) * f(i, j, k, grid, args...) + FT(0.25) * (f(i, j, k-1, grid, args...) + f(i, j, k+1, grid, args...))
```
glwagner, line 125: are these all just double interpolations?

It might compile faster if you re-use interpolation operators. Not sure.

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -0,0 +1,330 @@
...
+@inline M₂₂ᶜᶜᶜ(i, j, k, grid, u, v, w, α, β) = 2*Δᶠ(i, j, k, grid)^2 * (var"⟨|S|S₂₂⟩ᶜᶜᶜ"(i, j, k, grid, u, v, w) - α^2*β * var"|S̄|S̄₂₂ᶜᶜᶜ"(i, j, k, grid, u, v, w))
+@inline M₃₃ᶜᶜᶜ(i, j, k, grid, u, v, w, α, β) = 2*Δᶠ(i, j, k, grid)^2 * (var"⟨|S|S₃₃⟩ᶜᶜᶜ"(i, j, k, grid, u, v, w) - α^2*β * var"|S̄|S̄₃₃ᶜᶜᶜ"(i, j, k, grid, u, v, w))
+
+@inline M₁₂ᶜᶜᶜ(i, j, k, grid, u, v, w, α, β) = 2*Δᶠ(i, j, k, grid)^2 * (var"⟨|S|S₁₂⟩ᶜᶜᶜ"(i, j, k, grid, u, v, w) - α^2*β * var"|S̄|S̄₁₂ᶜᶜᶜ"(i, j, k, grid, u, v, w))
+@inline M₁₃ᶜᶜᶜ(i, j, k, grid, u, v, w, α, β) = 2*Δᶠ(i, j, k, grid)^2 * (var"⟨|S|S₁₃⟩ᶜᶜᶜ"(i, j, k, grid, u, v, w) - α^2*β * var"|S̄|S̄₁₃ᶜᶜᶜ"(i, j, k, grid, u, v, w))
+@inline M₂₃ᶜᶜᶜ(i, j, k, grid, u, v, w, α, β) = 2*Δᶠ(i, j, k, grid)^2 * (var"⟨|S|S₂₃⟩ᶜᶜᶜ"(i, j, k, grid, u, v, w) - α^2*β * var"|S̄|S̄₂₃ᶜᶜᶜ"(i, j, k, grid, u, v, w))
+
+
+
+@inline ϕψ(i, j, k, grid, ϕ, ψ) = ϕ[i, j, k] * ψ[i, j, k]
```
glwagner, line 278: missing @inbounds

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -0,0 +1,330 @@
...
+            + 2 * M₁₂ᶜᶜᶜ(i, j, k, grid, u, v, w, 2, 1)^2
+            + 2 * M₁₃ᶜᶜᶜ(i, j, k, grid, u, v, w, 2, 1)^2
+            + 2 * M₂₃ᶜᶜᶜ(i, j, k, grid, u, v, w, 2, 1)^2)
+end
+
+
+@kernel function compute_LM_MM!(LM, MM, grid, closure, velocities)
+    i, j, k = @index(Global, NTuple)
+    @inbounds LM[i, j, k] = LᵢⱼMᵢⱼ_ccc(i, j, k, grid, velocities...)
+    @inbounds MM[i, j, k] = MᵢⱼMᵢⱼ_ccc(i, j, k, grid, velocities...)
```
glwagner, line 67: Just use a `KernelFunctionOperator`?

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -0,0 +1,330 @@
...
+
+abstract type AveragingProcedure end
+abstract type DirectionalAveraging <: AveragingProcedure end
+struct       XDirectionalAveraging <: DirectionalAveraging end
+struct       YDirectionalAveraging <: DirectionalAveraging end
+struct       ZDirectionalAveraging <: DirectionalAveraging end
+struct      XYDirectionalAveraging <: DirectionalAveraging end
+struct      XZDirectionalAveraging <: DirectionalAveraging end
+struct      YZDirectionalAveraging <: DirectionalAveraging end
+struct     XYZDirectionalAveraging <: DirectionalAveraging end
```
glwagner, line 16: You may not need all these types. They're just `dims` so you only need

```julia
struct DimensionalAveraging{D}
    dims :: D
end
```

glwagner:
Not sure if I am missing something but first I think the user can just write

```julia
averaging=(1, 2)
```

which would mean average over xy. Or if you want to be more verbose then

```julia
averaging=DimensionAveraging(dims=(1, 2))
```

or something.

Next for the average itself it seems you need

```julia
LM_op = KernelFunctionOperation{Center, Center, Center}(LᵢⱼMᵢⱼ_ccc, grid, u, v, w)
LM_avg = Field(Average(LM_op, averaging.dims))
```

In the constructor. Then you can just call `compute!(LM_avg)` without needing lots of additional dispatch.

tomchor:
> Not sure if I am missing something but first I think the user can just write
> 
> ```julia
> averaging=(1, 2)
> ```

I considered taking that approach but I felt it made it a bit awkward to implement Lagrangian averaging in the future. That said, I do agree that my currently implementation is probably unnecessarily verbose. I think your suggestion below of `DimensionAveraging(dims=(1, 2))` is a good compromise.

> 
> which would mean average over xy. Or if you want to be more verbose then
> 
> ```julia
> averaging=DimensionAveraging(dims=(1, 2))
> ```
> 
> or something.
> 
> Next for the average itself it seems you need
> 
> ```julia
> LM_op = KernelFunctionOperation{Center, Center, Center}(LᵢⱼMᵢⱼ_ccc, grid, u, v, w)
> LM_avg = Field(Average(LM_op, averaging.dims))
> ```
> 
> In the constructor. Then you can just call `compute!(LM_avg)` without needing lots of additional dispatch.

Ah, that's so simple! I'll implement that.

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
tomchor: I think you already cracked it by suggesting using `KernelFunctionOperators`!

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -77,30 +72,21 @@ end
...
-    launch!(arch, grid, parameters, compute_LM_MM!,
-            diffusivity_fields.LM, diffusivity_fields.MM, grid, closure, velocities)
+    LM_op = KernelFunctionOperation{Center, Center, Center}(LᵢⱼMᵢⱼ_ccc, grid, velocities...)
+    MM_op = KernelFunctionOperation{Center, Center, Center}(MᵢⱼMᵢⱼ_ccc, grid, velocities...)
 
-    # TODO: optimize this
-    directional_average!(diffusivity_fields.LM_avg, diffusivity_fields.LM, closure.averaging)
-    directional_average!(diffusivity_fields.MM_avg, diffusivity_fields.MM, closure.averaging)
+    LM_avg = Field(Average(LM_op, dims=closure.averaging.dims))
+    MM_avg = Field(Average(MM_op, dims=closure.averaging.dims))
```
glwagner, line 89: you want to preallocate these in `DiffusivityFields`, and just `compute!` them in `compute_diffusivities!`.

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -32,7 +27,7 @@ end
 """
     ScaleInvariantSmagorinsky([time_discretization::TD = ExplicitTimeDiscretization(), FT=Float64;] averaging=1.0, Pr=1.0)
 """
-ScaleInvariantSmagorinsky(time_discretization::TD = ExplicitTimeDiscretization(), FT=Float64; averaging=XYZDirectionalAveraging(), Pr=1.0) where TD =
+ScaleInvariantSmagorinsky(time_discretization::TD = ExplicitTimeDiscretization(), FT=Float64; averaging=DirectionalAveraging((1,2,3)), Pr=1.0) where TD =
```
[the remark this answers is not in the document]
tomchor: Thanks! yeah, sorry, I have a tendency to forget those. I'll pay more attention.

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
tomchor: Yes, thanks. I realized yesterday that preallocating them in `DiffusivityFields` would require creating some new methods for `DiffusivityFields` in which we can pass `model`. So I was trying to see if it made sense to not modify the methods and instead do this. Surprisingly, I didn't see much of a performance decrease in this current version, although I've only tried for small models on CPUs.

I'll preallocate them in `DiffusivityFields` as you suggested though!

glwagner:
> > Not sure if I am missing something but first I think the user can just write
> > ```julia
> > averaging=(1, 2)
> > ```
> 
> I considered taking that approach but I felt it made it a bit awkward to implement Lagrangian averaging in the future. That said, I do agree that my currently implementation is probably unnecessarily verbose. I think your suggestion below of `DimensionAveraging(dims=(1, 2))` is a good compromise.

My suggestion is an API thing. If the user writes `averaging = (1, 2)` then you can interpret this as `DimensionAveraging((1, 2))`. Saves some characters.

Implementing this requires writing a function like

```julia
averaging(dims::Tuple) = DimenisionAveraging(dims)
averaging(other_averaging) = other_averaging
```

Also just brainstorming... maybe a constructor

```
DimensionAveraging(dims...) = DimensionAveraging(tuple(dims))
DimensionAveraging(dims::Colon) = DimensionAveraging(:)
```

will save some characters because arguably `DimensionAveraging(1, 2)` is easier to read than `DimensionAveraging((1, 2))`.

I think we are a little keyword-happy in the API, sometimes its nice to user positional arguments when meanings are obvious...

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
glwagner: > would require creating some new methods for DiffusivityFields in which we can pass model.

Why is that?

glwagner: Ah because they depend on velocities. That is a little puzzle...

One option is to pass `velocities` into `DiffusivityFields`. This is a fairly large change that might affect a lot of code, but could be a positive change perhaps. Maybe, we want to save that for after this PR.

You can also preallocate the field you're averaging into, but not the kernel function operation. Then you can compute the average with something like

```julia
mean!(field, kernel_function_operation)
```

(instead of `compute!`)

That could be a stop gap that will still let you do high performance simulation while we figure out the right pattern for this

tomchor: I'm not sure I follow your last suggestion. What I understood you're suggesting is something like the following, but which doesn't seem to work:

```julia
julia> op = Average(model.velocities.u^2, dims=(1, 2)) # Stand-in for the Lij and Mij functions
Average of BinaryOperation at (Face, Center, Center) over dims (1, 2)
└── operand: BinaryOperation at (Face, Center, Center)
    └── grid: 32×32×32 RectilinearGrid{Float64, Periodic, Periodic, Periodic} on CPU with 3×3×3 halo

julia> f = Field{Nothing, Nothing, Center}(grid)
1×1×32 Field{Nothing, Nothing, Center} reduced over dims = (1, 2) on RectilinearGrid on CPU
├── grid: 32×32×32 RectilinearGrid{Float64, Periodic, Periodic, Periodic} on CPU with 3×3×3 halo
│   └── west: Nothing, east: Nothing, south: Nothing, north: Nothing, bottom: Periodic, top: Periodic, immersed: ZeroFlux
└── data: 1×1×38 OffsetArray(::Array{Float64, 3}, 1:1, 1:1, -2:35) with eltype Float64 with indices 1:1×1:1×-2:35

julia> mean!(f, op)
ERROR: MethodError: no method matching mean!(::Field{…}, ::Reduction{…})

Closest candidates are:
  mean!(::Union{Oceananigans.Fields.AbstractField{Nothing}, Oceananigans.Fields.AbstractField{<:Any, Nothing}, Oceananigans.Fields.AbstractField{<:Any, <:Any, Nothing}}, ::AbstractArray; kwargs...)
   @ Oceananigans ~/repos/Oceananigans.jl2/src/Fields/field.jl:744
  mean!(::AbstractArray, ::AbstractArray)
   @ Statistics ~/bin/julia-1.10.2/share/julia/stdlib/v1.10/Statistics/src/Statistics.jl:140
```

Or maybe I did understand correctly, but we'd need to implement the relevant method for reductions?

glwagner: I guess you need `mean!(f, op.operand)`. Like here:

https://github.com/CliMA/Oceananigans.jl/blob/e7bbb1b77066f49a06c9f20942c70c2950de158b/src/Fields/scans.jl#L69

tomchor: Thanks! I think it works now

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -76,14 +76,14 @@ function compute_diffusivities!(diffusivity_fields, closure::ScaleInvariantSmago
...
 
-    LM_op = KernelFunctionOperation{Center, Center, Center}(LᵢⱼMᵢⱼ_ccc, grid, velocities...)
-    MM_op = KernelFunctionOperation{Center, Center, Center}(MᵢⱼMᵢⱼ_ccc, grid, velocities...)
+    LM_op = Average(KernelFunctionOperation{Center, Center, Center}(LᵢⱼMᵢⱼ_ccc, grid, model.velocities...))
+    MM_op = Average(KernelFunctionOperation{Center, Center, Center}(MᵢⱼMᵢⱼ_ccc, grid, model.velocities...))
 
-    LM_avg = Field(Average(LM_op, dims=closure.averaging.dims))
-    MM_avg = Field(Average(MM_op, dims=closure.averaging.dims))
+    mean!(diffusivity_fields.LM_avg, LM_op.operand)
+    mean!(diffusivity_fields.MM_avg, MM_op.operand)
```
glwagner, line 83: Note this only works on uniform grids, but hopefully that's good enough for preliminary testing

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -292,11 +292,25 @@ Base.show(io::IO, closure::ScaleInvariantSmagorinsky) = print(io, summary(closur
...
 function DiffusivityFields(grid, tracer_names, bcs, closure::ScaleInvariantSmagorinsky)
 
     default_eddy_viscosity_bcs = (; νₑ = FieldBoundaryConditions(grid, (Center, Center, Center)))
     bcs = merge(default_eddy_viscosity_bcs, bcs)
     νₑ = CenterField(grid, boundary_conditions=bcs.νₑ)
-    #return (; νₑ, LM_avg, MM_avg)
-    return (; νₑ)
+
+    LM_avg = directionally_averaged_field(grid, Val(closure.averaging))
+    MM_avg = directionally_averaged_field(grid, Val(closure.averaging))
```
glwagner, line 313: I would use something like this to make things a little simpler:

```julia
# TODO: add `velocities` as argument to `DiffusivityFields` to simplify this
MM_avg = Field(Average(νₑ, dims=closure.averaging.dims))
```

glwagner: but it doesn't matter

tomchor: Yeah I see your point. Since the important part is just to allocate memory, that already does the trick. That does kind of imply that `MM_avg` will later get the average of `νₑ`, which may confuse readers of the code.

How about 

```julia
MM_avg = Field(Average(Field{Center, Center, Center}(grid), dims=closure.averaging.dims))
```
?

glwagner: That's fine, my suggestion merely aimed to avoid allocating the additional 3D temporary array. 

Hopefully this code is temporary. It's just a vehicle for you to validate this closure so it can be used as soon as possible.

glwagner: You can also use an empty / placeholder operation. But don't worry too much let's move on to validation...

tomchor:
Here's some validation. The animation below is from a triply-periodic 3D simulation with an initial reasonably-resolved initial condition without any forcings. Note that the IC looks a bit weird right now (a couple of straight lines in strain) because I'm taking a shortcut in defining a well-resolved random IC, but I can easily change it later if necessary.

Left panel shows the strain rate squared for the `SmagorinskyLilly` closure, while the one on the right shows results for the `ScaleInvarianteSmagorinsky` closure, implemented in this PR. The bottom panel shows the evolution of the dynamically-calculated Smagorinsky coefficient for the latter, in comparison with the constant value of 0.16 imposed on the former.

Important here is that the value of 0.16 was analytically derived by Lilly by assuming an isotropic 3D turbulence, a Kolmogorovian energy cascade, and further assuming that the cut-off filter is in the inertial range. I think all these assumptions are valid in this simulation, so we expect the dynamically-calculated value of the Smagorinsky coefficient (the black line) to approach the value 0.16 as time goes on.

https://github.com/CliMA/Oceananigans.jl/assets/13205162/4049e7cf-452e-4883-a709-a675cf12277c

Although the match is not exact (the value it approaches is ~0.17), I think this is close enough. That said, I'm planning on also implementing a boundary layer validation along similar lines, which we can use to validate the model in the same fashion as [Bou-Zeid et al. (2005)](https://dx.doi.org/10.1063/1.1839152).

One thing to note is that the current implementation appears to be very slow. While the simulation with the `SmagorinskyLilly` closure runs on my laptop in 10 seconds, it takes 4 minutes fort the simulation with the `ScaleInvariantSmagorinsky`, so I'm hoping something can be changed here to improve performance.

tomchor:
It's also worth noting that right now many calculations are done more than once in each timestep. For example for each component of $M_{ij}$ I'm calculating the whole strain rate tensor modulus in addition to the strain rate tensor component needed:

https://github.com/CliMA/Oceananigans.jl/blob/25cc34e6c395e210e0aecf8181919c25435d7919/src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl#L241-L257

This is done for legibility of the code, but it may be necessary to forfeit that in favor of doing fewer calculations. (Also note that I'm using a weird way to define function names here, so lmk if you guys think I should change it.)

Another thing to note that it's common to update dynamic Smagorinsky coefficients once every 5 or so time-steps only, since they can be pretty expensive. afaik this is generally done for the scale-dependent versions, which have two test filters instead of the one needed in this PR, but I wouldn't be surprised if it's occasionally necessary for the scale-invariant versions as well.

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -0,0 +1,316 @@
...
+@inline var"⟨|S|S₁₃⟩ᶜᶜᶜ"(i, j, k, grid, u, v, w) = ℑxzᶜᵃᶜ(i, j, k, grid, ℱ²ᵟ, var"|S|S₁₃ᶜᶜᶜ", u, v, w)
+@inline var"⟨|S|S₂₃⟩ᶜᶜᶜ"(i, j, k, grid, u, v, w) = ℑyzᵃᶜᶜ(i, j, k, grid, ℱ²ᵟ, var"|S|S₂₃ᶜᶜᶜ", u, v, w)
+
+@inline var"|S̄|S̄₁₁ᶜᶜᶜ"(i, j, k, grid, u, v, w) = √(Σ̄ᵢⱼΣ̄ᵢⱼᶜᶜᶜ(i, j, k, grid, u, v, w)) * Σ̄₁₁(i, j, k, grid, u, v, w) # ccc
+@inline var"|S̄|S̄₂₂ᶜᶜᶜ"(i, j, k, grid, u, v, w) = √(Σ̄ᵢⱼΣ̄ᵢⱼᶜᶜᶜ(i, j, k, grid, u, v, w)) * Σ̄₂₂(i, j, k, grid, u, v, w) # ccc
+@inline var"|S̄|S̄₃₃ᶜᶜᶜ"(i, j, k, grid, u, v, w) = √(Σ̄ᵢⱼΣ̄ᵢⱼᶜᶜᶜ(i, j, k, grid, u, v, w)) * Σ̄₃₃(i, j, k, grid, u, v, w) # ccc
+
+@inline var"|S̄|S̄₁₂ᶜᶜᶜ"(i, j, k, grid, u, v, w) = √(Σ̄ᵢⱼΣ̄ᵢⱼᶠᶠᶜ(i, j, k, grid, u, v, w)) * Σ̄₁₂(i, j, k, grid, u, v, w) # ffc
+@inline var"|S̄|S̄₁₃ᶜᶜᶜ"(i, j, k, grid, u, v, w) = √(Σ̄ᵢⱼΣ̄ᵢⱼᶠᶜᶠ(i, j, k, grid, u, v, w)) * Σ̄₁₃(i, j, k, grid, u, v, w) # fcf
+@inline var"|S̄|S̄₂₃ᶜᶜᶜ"(i, j, k, grid, u, v, w) = √(Σ̄ᵢⱼΣ̄ᵢⱼᶜᶠᶠ(i, j, k, grid, u, v, w)) * Σ̄₂₃(i, j, k, grid, u, v, w) # cff
```
tomchor, line 247: Just noticed that the terms here are calculated at different locations and need to be fixed...

tomchor: Also possibly caught an error in the strain-rate calculation that should be fixed here as well: https://github.com/CliMA/Oceananigans.jl/pull/3648

tomchor:
> One thing to note is that the current implementation appears to be very slow. While the simulation with the `SmagorinskyLilly` closure runs on my laptop in 10 seconds, it takes 4 minutes for the simulation with the `ScaleInvariantSmagorinsky`. I know the dynamic model will be slower given the extra computations, but such a difference seems large to me, so I'm hoping something can be changed here to improve performance:

Avoided recomputation of the strain rate at `ccc` and sped things up a bit more. Now it runs in 2.9 minutes. A lot more to go...

tomchor:
I have some validation scripts already that work on the CPU and indicate that the new closure is working. However, I can't make this work on the GPU. I keep getting this error:

```
ERROR: LoadError: GPU compilation of MethodInstance for Oceananigans.TurbulenceClosures.gpu__compute_scale_invariant_smagorinsky_viscosity!(::KernelAbstractions.CompilerMetadata{…}, ::OffsetArrays.OffsetArray{…}, ::Field{…}, ::Field{…}, ::RectilinearGrid{…}, ::ScaleInvariantSmagorinsky{…}, ::Nothing, ::@NamedTuple{…}, ::@NamedTuple{}) failed
KernelError: passing and using non-bitstype argument

Argument 7 to your kernel function is of type ScaleInvariantSmagorinsky{Oceananigans.TurbulenceClosures.ExplicitTimeDiscretization, Oceananigans.TurbulenceClosures.DirectionalAveraging{Tuple{Int64, Int64}}, Float64, @NamedTuple{}, Integer}, which is not isbits:
  .update_frequency is of type Integer which is not isbits.
```

Reading up on the CUDA.jl docs I _think_ I understand where this error comes from (although I thought `update_frequency`, which is an `Integer`, should work, but it throws an error). Still couldn't figure out how to fix it in this case here. I assume it's not hard to fix though, so I was wondering if someone (I'm assuming @simone-silvestri  or @glwagner) can please give me a hand or at least point me in the right direction?

ali-ramadhan:
@tomchor Not sure if this is the issue but have you tried adapting `ScaleInvariantSmagorinsky` for the GPU?

Looks like `ScalarBiharmonicDiffusivity` needed it here: https://github.com/CliMA/Oceananigans.jl/blob/b3be950ef6a1f2139c2f65e083f20f1f7958ea55/src/TurbulenceClosures/turbulence_closure_implementations/scalar_biharmonic_diffusivity.jl#L109-L113

Not sure why `update_frequency` is coming up as an `Integer` (which is an abstract supertype) instead of `Int` (which is a concrete type and an alias for `Int64`) but maybe adapting will fix this?

tomchor:
> @tomchor Not sure if this is the issue but have you tried adapting `ScaleInvariantSmagorinsky` for the GPU?
> 
> Looks like `ScalarBiharmonicDiffusivity` needed it here:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/b3be950ef6a1f2139c2f65e083f20f1f7958ea55/src/TurbulenceClosures/turbulence_closure_implementations/scalar_biharmonic_diffusivity.jl#L109-L113
> 
> Not sure why `update_frequency` is coming up as an `Integer` (which is an abstract supertype) instead of `Int` (which is a concrete type and an alias for `Int64`) but maybe adapting will fix this?

I tried that and got a different error. But now looking at `ScalarBiharmonicDiffusivity` I think I probably did something wrong. I'll try again soon!

src/TurbulenceClosures/turbulence_closure_implementations/scale_invariant_smagorinsky.jl
```diff
@@ -22,11 +22,18 @@ struct ScaleInvariantSmagorinsky{TD, AP, FT, P, UF} <: AbstractScalarDiffusivity
     function ScaleInvariantSmagorinsky{TD, AP, FT}(averaging, Pr; update_frequency = 5) where {TD, AP, FT}
         Pr = convert_diffusivity(FT, Pr; discrete_form=false)
         P = typeof(Pr)
-        update_frequency = Integer(update_frequency)
-        return new{TD, AP, FT, P, Integer}(averaging, Pr, update_frequency)
+        update_frequency = Int(update_frequency)
```
glwagner, line 25: this is probably an interval (I'm guessing, in iterations) if its an integer. The frequency is 1 / interval

tomchor:
I wrote some validation scripts and I think the model is working, but I'd appreciate some feedback. First, this animation shows two 3D decaying turbulence simulations, one with `SmagorinskyLilly` and another with the `ScaleInvSmag`. I'm also plotting the value of the Smagorinsky constant from the dynamic model. For similar conditions (3D homogeneous tubrulence), Lilly assumed a sharp spectral filter  in the inertial subrange, matched the SGS dissipation rate to the TKE dissipation rate and obtained the theoretical value of the Smag constant of $c_s$ ≈ 0.165. I think these assumptions should be roughly valid in this validation script, and indeed the value that we obtain for $c_s$ is kinda close:

https://github.com/user-attachments/assets/6a5826b7-f952-4f35-a37b-21b6375abdc7

Although to be honest I expected the value to be closer. That said, I believe Lilly also assumed a _stationary_ turbulent flow (although I'd have to double check that), which is not the case here and may be affecting the coefficient value. Btw, I had posted a similar video before but I made some modifications to the model since then.

I also reproduced the first test simulation in the original Bou-Zeid (2005) paper: namely a channel flow forced by a pressure gradient, with a Monin-Obukhov-compliant bottom drag. I couldn't reproduce the simulation perfectly (for example, they use a pseudo-spectral scheme in the horizontal directions, force their drag with a horizontal average, and modify the `SmagLilly` constant slightly) so the results are a bit different from what's in their paper, but I think that's okay. Let me know if anyone thinks otherwise. Importantly, the `ScaleInvSmag` shows an improvement over the `SmagLilly`, with the later being expectedly overly diffusive and pretty much killing any turbulence at this resolution:

https://github.com/user-attachments/assets/6440450d-52f4-43ef-9ad5-0f8fcd9993d5

For reference, this is what similar plots from the paper look like (the equivalent for us here would be SMAG and PASI):

image

Some quick notes:

- Many tests are failing become I made an ad-hoc modification for now which passes the velocities to `DiffusivityFields()`. I did that because otherwise it was hard to make the model performant and simulations were taking way too long. We can (and should) review this and either come up with a better way to construct the `diffusivity_fields` or make this change separately in another PR, which will require changing the other models too.
- I'm updating the dynamic model once every 5 time steps only (this is user-defined). This is common practice for dynamic models since their cost is significantly higher than that of constant Smag or even AMD. With that practice, the dynamic model is taking about 3 times longer to run than the constant Smagorinsky. It does take significantly longer to compile (I haven't timed it). I believe there might be some optimizations still on the table though.

tomchor:
With the latest changes the compilation time is much shorter (around 10x shorter on the GPU) and simulations have been running in about twice as long as the simulations with constant Smagorinsky.

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/scale_invariant_operators.jl
```diff
@@ -0,0 +1,143 @@
...
+@inline M₂₃ᶜᶜᶜ(i, j, k, grid, u, v, w, α, β, Σᶜᶜᶜ, Σ̄ᶜᶜᶜ) = 2*Δᶠ(i, j, k, grid)^2 * (var"⟨ΣΣ₂₃⟩ᶜᶜᶜ"(i, j, k, grid, u, v, w, Σᶜᶜᶜ) - α^2*β * Σ̄Σ̄₂₃ᶜᶜᶜ(i, j, k, grid, u, v, w, Σ̄ᶜᶜᶜ))
+
+@inline uᵢuⱼ(i, j, k, grid, uᵢ, uⱼ) = @inbounds uᵢ[i, j, k] * uⱼ[i, j, k]
+@inline u₁u₁ᶜᶜᶜ(i, j, k, grid, u, v, w) = ℑxᶜᵃᵃ(i, j, k, grid, uᵢuⱼ, u, u)
+@inline u₂u₂ᶜᶜᶜ(i, j, k, grid, u, v, w) = ℑyᵃᶜᵃ(i, j, k, grid, uᵢuⱼ, v, v)
+@inline u₃u₃ᶜᶜᶜ(i, j, k, grid, u, v, w) = ℑzᵃᵃᶜ(i, j, k, grid, uᵢuⱼ, w, w)
+
+@inline ūᵢūⱼ(i, j, k, grid, uᵢ, uⱼ) = ℱ²ᵟ(i, j, k, grid, uᵢ) * ℱ²ᵟ(i, j, k, grid, uⱼ)
+@inline ū₁ū₁ᶜᶜᶜ(i, j, k, grid, u, v, w) = ℑxᶜᵃᵃ(i, j, k, grid, ūᵢūⱼ, u, u)
+@inline ū₂ū₂ᶜᶜᶜ(i, j, k, grid, u, v, w) = ℑxᶜᵃᵃ(i, j, k, grid, ūᵢūⱼ, v, v)
```
simone-silvestri, line 125:
```suggestion
@inline ū₂ū₂ᶜᶜᶜ(i, j, k, grid, u, v, w) = ℑyᵃᶜᵃ(i, j, k, grid, ūᵢūⱼ, v, v)
```

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/scale_invariant_operators.jl
```diff
@@ -0,0 +1,143 @@
...
+@inline ū₃ū₃ᶜᶜᶜ(i, j, k, grid, u, v, w) = ℑxᶜᵃᵃ(i, j, k, grid, ūᵢūⱼ, w, w)
```
simone-silvestri, line 126:
```suggestion
@inline ū₃ū₃ᶜᶜᶜ(i, j, k, grid, u, v, w) = ℑzᵃᵃᶜ(i, j, k, grid, ūᵢūⱼ, w, w)
```

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/scale_invariant_operators.jl
```diff
@@ -0,0 +1,143 @@
...
+#####
+
+const AG{FT} = AbstractGrid{FT} where FT
+
+# For a regularly-spaced grids, the double interpolations below
+@inline ℱ²ᵟ(i, j, k, grid::AG{FT}, f::Field{Face,   Center, Center}) where FT = ℑxyzᶠᶜᶜ(i, j, k, grid, ℑxyzᶜᶠᶠ, f)
+@inline ℱ²ᵟ(i, j, k, grid::AG{FT}, f::Field{Center, Face,   Center}) where FT = ℑxyzᶜᶠᶜ(i, j, k, grid, ℑxyzᶠᶜᶠ, f)
+@inline ℱ²ᵟ(i, j, k, grid::AG{FT}, f::Field{Center, Center, Face  }) where FT = ℑxyzᶜᶜᶠ(i, j, k, grid, ℑxyzᶠᶠᶜ, f)
+@inline ℱ²ᵟ(i, j, k, grid::AG{FT}, f::Field{Center, Center, Center}) where FT = ℑxyzᶜᶜᶜ(i, j, k, grid, ℑxyzᶠᶠᶠ, f)
+@inline ℱ²ᵟ(i, j, k, grid::AG{FT}, f::F, args...) where {FT, F <:Function} = ℑxyzᶜᶜᶜ(i, j, k, grid, ℑxyzᶠᶠᶠ, f, args...)
```
simone-silvestri, line 16:
```suggestion
@inline ℱ²ᵟ(i, j, k, grid::AG, f::Field{Face,   Center, Center}) = ℑxyzᶠᶜᶜ(i, j, k, grid, ℑxyzᶜᶠᶠ, f)
@inline ℱ²ᵟ(i, j, k, grid::AG, f::Field{Center, Face,   Center}) = ℑxyzᶜᶠᶜ(i, j, k, grid, ℑxyzᶠᶜᶠ, f)
@inline ℱ²ᵟ(i, j, k, grid::AG, f::Field{Center, Center, Face  }) = ℑxyzᶜᶜᶠ(i, j, k, grid, ℑxyzᶠᶠᶜ, f)
@inline ℱ²ᵟ(i, j, k, grid::AG, f::Field{Center, Center, Center}) = ℑxyzᶜᶜᶜ(i, j, k, grid, ℑxyzᶠᶠᶠ, f)
@inline ℱ²ᵟ(i, j, k, grid::AG, f::Function, args...) = ℑxyzᶜᶜᶜ(i, j, k, grid, ℑxyzᶠᶠᶠ, f, args...)
```
why for the function it is assumed the ccc location?

glwagner: Hmm, it won't work anyways to use `Field` in a kernel because it is adapted for GPU.

glwagner: I fixed it

glwagner:
Some results from the "closure comparison" validation test (rotating wind-driven mixing in a stratified fluid):

# dz = 1 meter, dx = dy = 2 meter

smag_comparison_10
closure_comparison_10

# dz = 2 meter, dx = dy = 4 meter

smag_comparison_20
closure_comparison_20

Seems like dynamic smag is doing what it should which is great. I'm surprised to see that the Lilly coefficient doesn't do much at coarse resolution at least.

On cost: it is a bit interesting. I think as kernel saturate, WENO(order=9) will start to win. For small kernels however, AMD seems ideal. We'll have to do a proper benchmark.

This PR also contains an experimental implementation of Lagrangian-averaged (scale-invariant) dynamic Smagorinsky.

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/scale_invariant_operators.jl
glwagner: i fixed this

tomchor:
I'm surprised at how much `SmagLilly` kills turbulence at these resolutions. To the point that $w^2$ is pretty much zero. A visualization of that effect for a wall flow can be seen [here](https://private-user-images.githubusercontent.com/13205162/377179071-6440450d-52f4-43ef-9ad5-0f8fcd9993d5.mp4?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzAyMDY0MDIsIm5iZiI6MTczMDIwNjEwMiwicGF0aCI6Ii8xMzIwNTE2Mi8zNzcxNzkwNzEtNjQ0MDQ1MGQtNTJmNC00M2VmLTlhZDUtMGY4ZmNkOTk5M2Q1Lm1wND9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEwMjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMDI5VDEyNDgyMlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTNhMjJjNmY0N2JiYmMyODEyNjFhODVjYmVlYTYwOTE2NTE2MmI1MjI4MmU2M2NmNTdiYzNhYjA2YzE0YjhiZDYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.WNd9vmwc5C1DihzFsCjihaAjJbVCs6aQ-j4LjxI7W-8). Crazy stuff.

glwagner:
> I'm surprised at how much `SmagLilly` kills turbulence at these resolutions. To the point that w 2 is pretty much zero. A visualization of that effect for a wall flow can be seen [here](https://private-user-images.githubusercontent.com/13205162/377179071-6440450d-52f4-43ef-9ad5-0f8fcd9993d5.mp4?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzAyMDY0MDIsIm5iZiI6MTczMDIwNjEwMiwicGF0aCI6Ii8xMzIwNTE2Mi8zNzcxNzkwNzEtNjQ0MDQ1MGQtNTJmNC00M2VmLTlhZDUtMGY4ZmNkOTk5M2Q1Lm1wND9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEwMjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMDI5VDEyNDgyMlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTNhMjJjNmY0N2JiYmMyODEyNjFhODVjYmVlYTYwOTE2NTE2MmI1MjI4MmU2M2NmNTdiYzNhYjA2YzE0YjhiZDYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.WNd9vmwc5C1DihzFsCjihaAjJbVCs6aQ-j4LjxI7W-8). Crazy stuff.

Right, it requires high resolution to do something remotely physical. I think that's a big limitation.

tomchor:
> I can't see the image because it seems to be protected.

Weird. It's the second video in [this comment](https://github.com/CliMA/Oceananigans.jl/pull/3642#issuecomment-2417879667) in this PR.

glwagner:
# dz = 0.5 meters, dx = dy = 1 meter

closure_comparison_0

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -0,0 +1,248 @@
...
+        X⁻ = (x⁻, y⁻, z⁻)
+
+        itp_𝒥ᴹᴹ = interpolate(X⁻, 𝒥ᴹᴹ⁻, (c, c, c), grid)
+        itp_𝒥ᴸᴹ = interpolate(X⁻, 𝒥ᴸᴹ⁻, (c, c, c), grid)
+
+        # Take time-step
+        𝒥ᴹᴹ[i, j, k] = ϵ * MMⁿ + (1 - ϵ) * itp_𝒥ᴹᴹ
+
+        𝒥ᴸᴹ★ = ϵ * LMⁿ + (1 - ϵ) * max(itp_𝒥ᴸᴹ, 𝒥ᴸᴹ_min)
+        𝒥ᴸᴹ[i, j, k] = max(𝒥ᴸᴹ★, 𝒥ᴸᴹ_min)
```
tomchor, line 191: @glwagner is the first `max()` needed?

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -0,0 +1,248 @@
...
+    @inbounds begin
+        𝒥ᴸᴹ⁻ᵢ = max(𝒥ᴸᴹ⁻[i, j, k], 𝒥ᴸᴹ_min)
+        𝒥ᴹᴹ⁻ᵢ = 𝒥ᴹᴹ⁻[i, j, k]
+
+        # Compute time scale
+        𝒥ᴸᴹ𝒥ᴹᴹ = 𝒥ᴸᴹ⁻ᵢ * 𝒥ᴹᴹ⁻ᵢ
+
+        if 𝒥ᴸᴹ𝒥ᴹᴹ < 0
+            @show 𝒥ᴸᴹ⁻ᵢ 𝒥ᴹᴹ⁻ᵢ 
+        end
```
tomchor, line 152: @glwagner I'm assuming I can remove this?

tomchor:
Here are results including the Lagrangian-averaged version of the dynamic Smagorinsky with $\Delta z=2$ m:

closure_comparison_20

As expected, for this simulation the results are pretty similar to the directionally-averaged version.

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
glwagner: I believe so yes

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
glwagner: yes!

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
glwagner: The first is the important one. The second matters at initialization.

glwagner:
> Here are results including the Lagrangian-averaged version of the dynamic Smagorinsky with Δ z = 2 m:
> 
> closure_comparison_20
> 
> As expected, for this simulation the results are pretty similar to the directionally-averaged version.

Lagrangian-averaged should be slightly less diffusive right?

tomchor:
> Lagrangian-averaged should be slightly less diffusive right?

Hmm, I'm not aware of that. afaik it's about the same in a doubly periodic domain. Looking at Bou-Zeid's Figure 1 I also don't see any obvious trend that way. What is generally true is that scale-dependent formulations (not implemented here) tend to be less diffusive then scale-invariant ones (which is the one implemented here).

glwagner:
> > Lagrangian-averaged should be slightly less diffusive right?
> 
> Hmm, I'm not aware of that. afaik it's about the same in a doubly periodic domain. Looking at Bou-Zeid's Figure 1 I also don't see any obvious trend that way. What is generally true is that scale-dependent formulations (not implemented here) tend to be less diffusive then scale-invariant ones (which is the one implemented here).

I simply mean that with more averaging there will be more diffusion due to the distribution of strain events.

glwagner:
> > Lagrangian-averaged should be slightly less diffusive right?
> 
> Hmm, I'm not aware of that. afaik it's about the same in a doubly periodic domain. Looking at Bou-Zeid's Figure 1 I also don't see any obvious trend that way. What is generally true is that scale-dependent formulations (not implemented here) tend to be less diffusive then scale-invariant ones (which is the one implemented here).

Figure 1 in Bou-Zeid is below:

image


It shows that away from the wall, the Lagrangian-averaged scheme has a lower eddy viscosity. Adding scale dependence decreases the eddy viscosity further.

This is more or less consistent with the idea that "better" schemes have less diffusion, in a bulk and coarse sense. Of course, good schemes also need to produce _more_ diffusion "when they need it" which is why a constant Smagorinsky coefficient does not work. That's LES in a nutshell... enough diffusion when needed, and minimal otherwise...

glwagner:
Interesting. I disagree that fig 1 does not have an obvious trend, the trend is quite clear away from the wall:

image

Both Lagrangian schemes have larger coefficients.

But still, I was referring to the eddy viscosity, not the coefficient. It would be interesting to me if the Lagrangian scheme was consistently more diffusive. But that might actually also be consistent with convergence at lower resolution, perhaps.

tomchor:
> I disagree that fig 1 does not have an obvious trend, the trend is quite clear away from the wall:
>
> Both Lagrangian schemes have larger coefficients than their planar-averaged counter parts. Also scale dependence increases the Lagrangian coefficient.

I agree that far from the wall there is a clear trend. But I'm not sure that's a fair way to look at it.

> But still, I was referring to the eddy viscosity, not the coefficient. It would be interesting to me if the Lagrangian scheme was consistently more diffusive. But that might actually also be consistent with convergence at lower resolution, perhaps.

That said, good point that the coefficient doesn't necessarily translate into the eddy viscosity. I'll check the eddy viscosity later and see what comes up as I am now curious :)

tomchor:
I've been getting the following error right at the end of docs on buildkite that I can't reproduce locally (and therefore have not been able to solve):

```julia
┌ Error: 4 docstrings not included in the manual:
│ 
│     Oceananigans.TurbulenceClosures.Smagorinskys.DynamicCoefficient :: Union{Tuple{}, Tuple{Any}}
│     Oceananigans.TurbulenceClosures.Smagorinskys.SmagorinskyLilly :: Union{Tuple{}, Tuple{Any}, Tuple{Any, Any}}
│     Oceananigans.TurbulenceClosures.Smagorinskys.LillyCoefficient :: Union{Tuple{}, Tuple{Any}}
│     Oceananigans.TurbulenceClosures.Smagorinskys.Smagorinsky :: Union{Tuple{}, Tuple{TD}, Tuple{TD, Any}} where TD
│ 
│ These are docstrings in the checked modules (configured with the modules keyword)
│ that are not included in canonical @docs or @autodocs blocks.
└ @ Documenter /storage5/buildkite-agent/.julia-18496/packages/Documenter/C1XEF/src/utilities/utilities.jl:44
┌ Debug: checking footnote links.
└ @ Documenter /storage5/buildkite-agent/.julia-18496/packages/Documenter/C1XEF/src/docchecks.jl:114
[ Info: Populate: populating indices.
ERROR: LoadError: `makedocs` encountered an error [:missing_docs] -- terminating build before rendering.
Stacktrace:
 [1] error(s::String)
```

Anyone know what that's about? Those are the 4 new docstrings that I added, all in the `Smagorinskys` module.

tomchor:
FYI the tests are failing only because somehow tests are getting a kill signal for external causes. As far as I can tell, all tests are passing and this PR is ready for review.

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl
```diff
@@ -0,0 +1,147 @@
...
+    DiffusivityFields,
+    tracer_diffusivities
+
+#####
+##### The turbulence closure proposed by Smagorinsky and Lilly.
+#####
+
+# struct LagrangianAveragedCoefficient end
+
+struct Smagorinsky{TD, C, P} <: AbstractScalarDiffusivity{TD, ThreeDimensionalFormulation, 2}
```
simone-silvestri, line 31:
```suggestion
struct Smagorinsky{C, P} <: AbstractScalarDiffusivity{<:ExplicitTimeDisretization, ThreeDimensionalFormulation, 2}
```
I don't think there is an implicit version set up

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl
```diff
@@ -0,0 +1,147 @@
...
+        P = typeof(Pr)
+        C = typeof(coefficient)
+        return new{TD, C, P}(coefficient, Pr)
+    end
+end
+
+@inline viscosity(::Smagorinsky, K) = K.νₑ
+@inline diffusivity(closure::Smagorinsky, K, ::Val{id}) where id = K.νₑ / closure.Pr[id]
+
+const ConstantSmagorinsky = Smagorinsky{<:Any, <:Number}
```
simone-silvestri, line 45:
```suggestion
const ConstantSmagorinsky = Smagorinsky{<:Number}
```

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl
```diff
@@ -0,0 +1,147 @@
...
+
+`Cˢ` is the Smagorinsky coefficient and the default value is 0.16, according
+to the analysis by [Lilly66](@citet). For other options, see `LillyCoefficient`
+and `DynamicCoefficient`.
+"""
+function Smagorinsky(time_discretization::TD = ExplicitTimeDiscretization(), FT=Float64;
+                     coefficient = 0.16, Pr = 1.0) where TD
+    Pr = convert_diffusivity(FT, Pr; discrete_form=false)
+    return Smagorinsky{TD}(coefficient, Pr)
+end
```
simone-silvestri, line 79:
```suggestion
function Smagorinsky(; coefficient = 0.16, Pr = 1.0) 
    Pr = convert_diffusivity(FT, Pr; discrete_form=false)
    return Smagorinsky(coefficient, Pr)
end
```

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl
glwagner: It can use vertically implicit for the diffusion coefficient (like any abstract scalar diffusivity)

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl
```diff
@@ -0,0 +1,147 @@
...
+
+Smagorinsky(FT::DataType; kwargs...) = Smagorinsky(ExplicitTimeDiscretization(), FT; kwargs...)
```
simone-silvestri, line 81:
```suggestion
```

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl
```diff
@@ -0,0 +1,147 @@
...
+
+function with_tracers(tracers, closure::Smagorinsky{TD}) where TD
+    Pr = tracer_diffusivities(tracers, closure.Pr)
+    return Smagorinsky{TD}(closure.coefficient, Pr)
+end
```
simone-silvestri, line 86:
```suggestion
function with_tracers(tracers, closure::Smagorinsky)
    Pr = tracer_diffusivities(tracers, closure.Pr)
    return Smagorinsky(closure.coefficient, Pr)
end
```

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl
simone-silvestri: Ok, I guess we can keep that time-stepping option

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -0,0 +1,257 @@
...
+
+const DynamicSmagorinsky = Smagorinsky{<:Any, <:DynamicCoefficient}
+Adapt.adapt_structure(to, dc::DynamicCoefficient) = DynamicCoefficient(dc.averaging, dc.minimum_numerator, nothing)
+
+# Directional averaging with averaging=(1, 2), etc
+
+const DirectionallyAveragedCoefficient{N} = DynamicCoefficient{<:Union{NTuple{N, Int}, Int, Colon}} where N
+const DirectionallyAveragedDynamicSmagorinsky{N} = Smagorinsky{<:Any, <:DirectionallyAveragedCoefficient{N}} where N
+
+# Lagrangian averaging with averaging=(1, 2), etc
```
simone-silvestri, line 19: Is this comment correct?

tomchor: No it is not. Thanks for catching that!

tomchor: Fixed!

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl
glwagner: I think we want to keep support for vertically implicit for wall resolved LES etc

glwagner:
> Hmm, now I'm getting a "no space on device error":
> 
> image

gotta delete somem stuff

glwagner (approved):

tomchor:
@glwagner could you please give me a hand with the tests? The turbulence tests are getting stuck when building a `NonhydrostaticModel` with a dynamic Smagorinsky (with both kinds of averaging):

image

The thing is that I have not been able to reproduce this locally, so it's been hard for me to fix it. Do you have any ideas of how to proceed? All other tests seems to pass.

tomchor:
It seems that whenever the model has a `DynamicSmag`, `update_state!()` gets stuck here: https://github.com/CliMA/Oceananigans.jl/blob/5dd01ddfe73dd03515b52e6499b27c1b61c71364/src/Models/NonhydrostaticModels/update_nonhydrostatic_model_state.jl#L43

which ends up calling `compute_diffusivities!()` here: https://github.com/CliMA/Oceananigans.jl/blob/5dd01ddfe73dd03515b52e6499b27c1b61c71364/src/Models/NonhydrostaticModels/update_nonhydrostatic_model_state.jl#L64-L67

In this branch, `compute_diffusivities!()` looks like https://github.com/CliMA/Oceananigans.jl/blob/d00a7e6802ca73aaeb7e9c247f02775af719648f/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl#L108-L121

and is general for all Smagorinsky models, so I'm guessing the culprit is in `compute_coefficient_fields!()`, which for the `DirectionallyAveragedDynamicSmagorinsky` is: https://github.com/CliMA/Oceananigans.jl/blob/d00a7e6802ca73aaeb7e9c247f02775af719648f/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl#L117-L139

And something in there makes the compiler get stuck on buildkite. I have no idea what that something is. I was not able to reproduce this on any local machine that I've tried, which is making this hard for me to pinpoint the cause.

glwagner:
Is the problem for GPU or CPU

tomchor:
I haven't been able to run the GPU tests for a while, but it's happening on CPUs for sure

glwagner:
Which test / which file causes the problem?

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -0,0 +1,272 @@
...
+        𝒥ᴸᴹ⁻ = diffusivity_fields.𝒥ᴸᴹ⁻
+        𝒥ᴹᴹ⁻ = diffusivity_fields.𝒥ᴹᴹ⁻
+        𝒥ᴸᴹ  = diffusivity_fields.𝒥ᴸᴹ
+        𝒥ᴹᴹ  = diffusivity_fields.𝒥ᴹᴹ
+        𝒥ᴸᴹ_min = cˢ.minimum_numerator
+
+        if !isfinite(clock.last_Δt) || Δt == 0 # first time-step
+            @info "               Launching _compute_LM_MM! at t=0"
+            launch!(arch, grid, :xyz, _compute_LM_MM!, 𝒥ᴸᴹ, 𝒥ᴹᴹ, Σ, Σ̄, grid, u, v, w)
+            @info "               Finished _compute_LM_MM!"
```
tomchor, line 242: The buildkite tests are getting stuck here, compiling `_compute_LM_MM!()`. I found that if I comment out some stuff either in `_compute_LM_MM!()` or in `LM_and_MM()` (the former calls the latter) then they compile and tests pass.

I have tried this with different local machines, and tests pass in all of them. So I'm really not sure what's different about the buildkite set-up that's causing this.

glwagner: on GPU or CPU?

glwagner: eg on tartarus (CPU) or sverdrup (GPU)?

If one of them isn't up to the task, let's just mark the test as `@test_broken`

tomchor: I've been focusing on CPU tests, but I can also investigate this on the GPU.

What do you mean "up to the task" though? I thought for CPU these things should compile regardless of the machine or kernel complexity, no?

glwagner: Compilation still has a computational cost!

tomchor: Hmm, it seems neither the CPU nor GPU turbulence tests compile `_compute_LM_MM!()`. I just re-started the time-stepping tests so that we see if it compiles there. I'm reluctant to merge this PR without any sort of working tests for the dynamic closures.

glwagner: This PR does much more than just implement the dynamic closures. In terms of going out of date, the outcome is the same whether we merge the PR or not. It's definitely better to merge than to let it sit around growing stale.

glwagner: Focus on productivity, development, and moving forward.

test/test_turbulence_closures.jl
```diff
@@ -212,9 +259,10 @@ end
     @testset "Closure instantiation" begin
         @info "  Testing closure instantiation..."
         for closurename in closures
-            closure = getproperty(TurbulenceClosures, closurename)()
+            closure = @eval $closurename()
             @test closure isa TurbulenceClosures.AbstractTurbulenceClosure
 
+            closure isa DynamicSmagorinsky && continue # `DynamicSmagorinsky`s `_compute_LM_MM!()` kernel isn't compiling on buildkite
             grid = RectilinearGrid(CPU(), size=(2, 2, 2), extent=(1, 2, 3))
             model = NonhydrostaticModel(grid=grid, closure=closure, tracers=:c)
```
glwagner, line 267:
```suggestion
            model = NonhydrostaticModel(; grid, closure, tracers=:c)
```

test/test_time_stepping.jl
```diff
@@ -238,13 +239,22 @@ end
 
 Planes = (FPlane, ConstantCartesianCoriolis, BetaPlane, NonTraditionalBetaPlane)
 
-BuoyancyModifiedAnisotropicMinimumDissipation(FT) = AnisotropicMinimumDissipation(FT, Cb=1.0)
+BuoyancyModifiedAnisotropicMinimumDissipation(FT=Float64) = AnisotropicMinimumDissipation(FT, Cb=1.0)
+
+ConstantSmagorinsky(FT=Float64) = Smagorinsky(FT, coefficient=0.16)
+DirectionallyAveragedDynamicSmagorinsky(FT=Float64) =
+    Smagorinsky(FT, coefficient=DynamicCoefficient(averaging=(1,2)))
```
glwagner, line 246:
```suggestion
    Smagorinsky(FT, coefficient=DynamicCoefficient(averaging=(1, 2)))
```

test/test_time_stepping.jl
```diff
@@ -238,13 +239,22 @@ end
 
 Planes = (FPlane, ConstantCartesianCoriolis, BetaPlane, NonTraditionalBetaPlane)
 
-BuoyancyModifiedAnisotropicMinimumDissipation(FT) = AnisotropicMinimumDissipation(FT, Cb=1.0)
+BuoyancyModifiedAnisotropicMinimumDissipation(FT=Float64) = AnisotropicMinimumDissipation(FT, Cb=1.0)
+
+ConstantSmagorinsky(FT=Float64) = Smagorinsky(FT, coefficient=0.16)
```
glwagner, line 244:
```suggestion
ConstantSmagorinsky(FT=Float64) = Smagorinsky(FT)
```

test/test_turbulence_closures.jl
```diff
@@ -148,6 +156,45 @@ function time_step_with_variable_discrete_diffusivity(arch)
...
+    @test size(model.diffusivity_fields.𝒥ᴹᴹ) == (grid.Nx, 1, 1)
+
+    closure = DynamicSmagorinsky(averaging=LagrangianAveraging())
+    model = NonhydrostaticModel(; grid, closure)
+    @test size(model.diffusivity_fields.𝒥ᴸᴹ)  == size(grid)
+    @test size(model.diffusivity_fields.𝒥ᴹᴹ)  == size(grid)
+    @test size(model.diffusivity_fields.𝒥ᴸᴹ⁻) == size(grid)
+    @test size(model.diffusivity_fields.𝒥ᴹᴹ⁻) == size(grid)
+    @test size(model.diffusivity_fields.Σ)    == size(grid)
+    @test size(model.diffusivity_fields.Σ̄)    == size(grid)
```
glwagner, line 193: What's the purpose of all of these tests? Are they looking for a specific error that cropped up?

tomchor: No, I just thought it was helpful to test that the diffusivity fields were the appropriate size given that the directionally-averaged versions have sizes that are different from the underlying `grid`.

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
tomchor: > Add a hook that avoids the computation of LM_MM, and use it in the test. Then there is some kind of unit test which is only missing the LM_MM computation (which is very unlikely to break). Then we can use the closure.

Ah, I missed this suggestion! I'll open another PR later to implement this.

Can you clarify what you mean by hook though? Do you mean an extra flag for the dynamic coefficient that skips that computation?

PR #3643, opened 2024-06-28
Add test of enzyme flux bc

jlk9:
This adds an enzyme test for differentiating flux boundary conditions with Enzyme, which is of interest for solving inverse problems and other applications. @wsmoses


test/test_enzyme.jl
```diff
@@ -267,3 +267,83 @@ end
...
+    U = 1
+    u₀(x, y, z) = - U * cos(x + L/8) * sin(y) * (z + L/2)
+    v₀(x, y, z) = + U * sin(x + L/8) * cos(y) * (z + L/2)
+
+    set!(u, u₀)
+    set!(v, v₀)
+    fill_halo_regions!(u)
+    fill_halo_regions!(v)
+
+    @inline function tracer_flux(x, y, t, c, p)
```
glwagner, line 298: suggest using the discrete form because most real cases will use that.

wsmoses:
@jlk9 and @glwagner I just rebased this. If it passes, is there anything else needed?

test/test_enzyme.jl
jlk9: I switched this test to discrete form and it now passes.

jlk9:
Greg's main issue seemed to be that we didn't use discrete form for the BC function, which was fixed. So if we can get tests to pass it's probably good to merge.

glwagner:
Is this PR still active?

jlk9:
Yes it is, we would like to get a flux BC incorporated into the testing of the advection-diffusion problem. This allows for some more interesting applications than just inverting for initial conditions (for example).

jlk9:
@glwagner Ok to merge?

test/test_enzyme.jl
```diff
@@ -183,16 +183,84 @@ end
...
+    set!(u, u₀)
+    set!(v, v₀)
+    fill_halo_regions!(u)
+    fill_halo_regions!(v)
+
+    @inline function tracer_flux(i, j, grid, clock, model_fields, p)
         c₀ = p.surface_tracer_concentration
         u★ = p.piston_velocity
-        return - u★ * (c₀ - c)
+        return - u★ * (c₀ - model_fields.c[i,j,end])
```
glwagner, line 257: `end` usually fails because it goes into a halo region. Also spaces after commas...

test/test_enzyme.jl
```diff
@@ -183,16 +183,84 @@ end
...
+
+    Lx = Ly = L = 2π
+    Lz = 1
+
+    x = y = (-L/2, L/2)
+    z = (-Lz/2, Lz/2)
+    topology = (Periodic, Periodic, Bounded)
+
+    grid = RectilinearGrid(size=(Nx, Ny, Nz); x, y, z, topology)
+    diffusion = VerticalScalarDiffusivity(κ=0.1)
```
glwagner, line 240: This looks like it copy-pastes a lot of code from above. Can we somehow combine the two so we don't copy / paste? It will make it easier to maintain in the future.

jlk9: I just combined the tests into one that reuses most of the code.

test/test_enzyme.jl
jlk9: Is there better way to get the last non-halo element along an axis? Or should it be hardcoded or set as a parameter?

test/test_enzyme.jl
```diff
@@ -268,39 +201,51 @@ end
     # 2. Add surface fluxes
     # 3. Do a problem where we invert for the tracer fluxes (maybe with CATKE)
 
-    model = HydrostaticFreeSurfaceModel(; grid,
+    model_no_bc = HydrostaticFreeSurfaceModel(; grid,
                                         tracer_advection = WENO(),
                                         tracers = :c,
                                         velocities = PrescribedVelocityFields(; u, v),
-                                        boundary_conditions = (; c=c_bcs),
                                         closure = diffusion)
```
glwagner, line 208: fix alignment

test/test_enzyme.jl
```diff
@@ -268,39 +201,51 @@ end
...
-    κ₁, κ₂ = 0.9, 1.1
-    c²₁ = stable_diffusion!(model, 1, κ₁)
-    c²₂ = stable_diffusion!(model, 1, κ₂)
-    dc²_dκ_fd = (c²₂ - c²₁) / (κ₂ - κ₁)
+    model_bc = HydrostaticFreeSurfaceModel(; grid,
+                                        tracer_advection = WENO(),
+                                        tracers = :c,
+                                        velocities = PrescribedVelocityFields(; u, v),
+                                        boundary_conditions = (; c=c_bcs),
+                                        closure = diffusion)
```
glwagner, line 216:
```suggestion
    model_bc = HydrostaticFreeSurfaceModel(; grid,
                                           tracer_advection = WENO(),
                                           tracers = :c,
                                           velocities = PrescribedVelocityFields(; u, v),
                                           boundary_conditions = (; c=c_bcs),
                                           closure = diffusion)

```

glwagner (approved):

glwagner:
Tidy up the formatting then you're good to go if tests pass

test/test_enzyme.jl
jlk9: Just replaced `end` with `p.level`, which is set to `Nz`

test/test_enzyme.jl
```diff
@@ -183,54 +184,69 @@ end
     fill_halo_regions!(u)
     fill_halo_regions!(v)
 
-    @inline function tracer_flux(x, y, t, c, p)
+    @inline function tracer_flux(i, j, grid, clock, model_fields, p)
         c₀ = p.surface_tracer_concentration
         u★ = p.piston_velocity
-        return - u★ * (c₀ - c)
+        return - u★ * (c₀ - model_fields.c[i, j, p.level])
```
glwagner, line 190:
```suggestion
        return - u★ * (c₀ - model_fields.c[i, j, grid.Nz])
```

will work too if this is a surface flux.

test/test_enzyme.jl
```diff
@@ -183,54 +184,69 @@ end
...
     end
 
     parameters = (surface_tracer_concentration = 1,
-                  piston_velocity = 0.1)
+                  piston_velocity = 0.1,
+                  level = Nz)
 
-    top_c_bc = FluxBoundaryCondition(tracer_flux, field_dependencies=:c; parameters)
+    top_c_bc = FluxBoundaryCondition(tracer_flux; discrete_form=true, parameters=parameters)
```
glwagner, line 197:
```suggestion
    top_c_bc = FluxBoundaryCondition(tracer_flux; discrete_form=true, parameters)
```

test/test_enzyme.jl
```diff
@@ -183,54 +184,69 @@ end
...
+        amplitude = 1.0
+        κ = 1.0
+        dmodel = Enzyme.make_zero(models[i])
+        set_diffusivity!(dmodel, 0)
+
+        dc²_dκ = autodiff(Enzyme.set_runtime_activity(Enzyme.Reverse),
+                        stable_diffusion!,
+                        Duplicated(models[i], dmodel),
+                        Const(amplitude),
+                        Active(κ))
```
glwagner, line 239:
```suggestion
        dc²_dκ = autodiff(Enzyme.set_runtime_activity(Enzyme.Reverse),
                          stable_diffusion!,
                          Duplicated(models[i], dmodel),
                          Const(amplitude),
                          Active(κ))
```

test/test_enzyme.jl
```diff
@@ -157,7 +157,8 @@ end
     end
 end
 
-@testset "Enzyme on advection and diffusion" begin
+
+@testset "Enzyme on advection and diffusion WITHOUT then WITH flux boundary condition" begin
```
glwagner, line 161:
```suggestion
@testset "Enzyme for advection and diffusion with various boundary conditions" begin
```

PR #3645, opened 2024-07-01
Add new constructor for KernelParameters

glwagner:
This PR adds a new constructor for `KernelParameters` that takes ranges instead of sizes and offsets. I believe this leads to code that is easier to read for some applications.

I tried changing the source code where it seemed appropriate. However, in some files I was unable to understand what the code was doing. I wanted to improve those files, but couldn't. So I left a comment instead.

@simone-silvestri


simone-silvestri (approved):

src/Models/interleave_communication_and_computation.jl
```diff
@@ -38,7 +38,8 @@ function interior_tendency_kernel_parameters(grid, arch)
     Hx, Hy, _ = halo_size(grid)
     Tx, Ty, _ = topology(grid)
     Nx, Ny, Nz = size(grid)
-    
+
+    # TODO: add a comment explaining what is going on here
     Sx = Rx == 1 ? Nx : (Tx == RightConnected || Tx == LeftConnected ? Nx - Hx : Nx - 2Hx)
     Sy = Ry == 1 ? Ny : (Ty == RightConnected || Ty == LeftConnected ? Ny - Hy : Ny - 2Hy)
```
simone-silvestri, line 44:
```suggestion
    # Kernel parameters to compute the tendencies in all the interior if the direction is local (`R == 1`) and only in 
    # the part of the domain that does not depend on the halo cells if the direction is partitioned. 
    local_x = Rx == 1
    local_y = Ry == 1
    one_sided_connection_x = Tx == RightConnected || Tx == LeftConnected
    one_sided_connection_y = Ty == RightConnected || Ty == LeftConnected 
    Sx = local_x ? Nx : ( one_sided_connection_x ? Nx - Hx : Nx - 2Hx)
    Sy = local_y ? Ny : ( one_sided_connection_y ? Ny - Hy : Ny - 2Hy)
```

PR #3648, opened 2024-07-08
(0.91.4) Fix interpolations in off-diagonal components of strain-rate tensor

tomchor:
I think the interpolation here for the w-terms of the strain-rate tensor hasn't being correctly done. In particular it was assuming  that `w` was a `CenterField`.

Can someone please double-check?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.91.3"
+version = "0.91.4"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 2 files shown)


glwagner (approved):
I think this is right

tomchor:
If people are okay with it, I'll merge as soon as tests pass and register a new patch version, since this is a bugfix.

tomchor:
These tests aren't passing, and I have no clue why. The error is:

```
ERROR: `Enzyme=7da242da-08ed-463a-9acd-ee780be4f1d9` depends on `ChainRulesCore=d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4`, but no such entry exists in the manifest.
```

But I never changed anything related to that. Does anyone know what's going on?

navidcy:
Stochastic CI variability? I restarted them...

tomchor:
Okay finally tests passed!

@navidcy I tried restarting the tests over and over (honestly more than 10 times over the past few days) and they always failed. I could only make them pass this morning when I restarted them one at a time. That is, starting one, and only restarting the next failed test when the previous one had fully run.

So maybe the variability we see in tests has to do with different processes trying to access the same resources... ?

navidcy:
What you observe about one at a time etc I have observed it other times but I can’t say for sure…

glwagner:
> Okay finally tests passed!
> 
> @navidcy I tried restarting the tests over and over (honestly more than 10 times over the past few days) and they always failed. I could only make them pass this morning when I restarted them one at a time. That is, starting one, and only restarting the next failed test when the previous one had fully run.
> 
> So maybe the variability we see in tests has to do with different processes trying to access the same resources... ?

Yes it is a race condition

PR #3652, opened 2024-07-11
Add Downloads badges on README.md

navidcy:
Is something like this useful? See how they look at https://github.com/CliMA/Oceananigans.jl/tree/ncc/readme-downloads


glwagner:
Seems nice but what exactly is a "download"? A fork, `Pkg.add("Oceananigans")` or what?

navidcy:
> Seems nice but what exactly is a "download"? A fork, `Pkg.add("Oceananigans")` or what?

I’m not quite sure.

glwagner:
It could be based on this post https://discourse.julialang.org/t/announcing-package-download-stats/69073

which means that the number represents the number of unique IP addresses that have downloaded Oceananigans from the julia registry (I think), excluding IP addresses that are suspected to originate from bots / CI. that would mostly be `Pkg.add` I think.

glwagner (approved):
Might be good just to add a comment somewhere / somehow that explains what it means

PR #3653, opened 2024-07-15
Add inlines for general interpolation

simone-silvestri:
some inline were missing for the general interpolation function. 
This PR fixes it

edit: also corrects a bug in the documentation


glwagner (approved):

navidcy:
> edit: also corrects a bug in the documentation 

I'm slightly puzzled about this fix because it's within a doctest... so how was the doctest passing? But the doctests also pass now so it's fine I guess.

PR #3659, opened 2024-07-23
Removing some advection schemes

simone-silvestri:
These schemes are dominated (in terms of both accuracy and speed) by the default methods and, since they should not be used because they are _always_ a worst option, it makes sense to remove them not to continue maintaining them.

These are:
- JS WENO formulation for smoothness indicators (dominated by the ZWENO formulation)
- Velocity-based upwinding for divergence and Bernoulli term in the vector invariant formulation (dominated by the other two options: `OnySelfUpwinding` and `CrossandSelfUpwinding`)

Thus, this PR removes these methods


glwagner (approved):

simone-silvestri:
There is some problem with Enzyme here, which is weird because this PR only removes code does not add anything new

glwagner:
That's good to find. Try updating the manifest?

PR #3660, opened 2024-07-23
(0.91.5) Tweak initialization procedure so that callback schedules work with checkpointed simulations

### LINKED ISSUE #3657: Post-Checkpoint Output Saving Interval Issue
Hi all,
I have an issue with post-checkpoint runs. I've set the output saving interval to 30 minutes, but for the first few hours after picking up the checkpoint, the saving interval is actually equal to Δt (approximately 30 seconds in my case). Additionally, the progress message is generated every iteration (30 seconds) instead of every 30 minutes as intended. This strange behavior continues until iteration 13398, after which the interval and progress messages behave correctly (see the progress message below).
It is unclear to me what went wrong. Any ideas would be much appreciated!
(I can provide the code about how I set the `outputwriter` and progress message if needed; I am using Oceananigans v0.91.3, and Julia/1.10.2)

image

glwagner:
This PR tweaks the initialization procedure for `run!` so that schedules are initialized separately from the execution of callbacks and output writers at `iteration == 0`. This allows simulations that start from iterations other than `0` to execute correctly.

It also re-defines the meaning of "actuations" for `TimeInterval` to be more sensible. Previously the actual number of actuations was `actuations - 1` which is confusing.

There are still some issues with initialization. However I think this PR helps to clarify the distinction between "initialization", which occurs whenever we call `run!`, vs the thing we do at "iteration 0", like writing output.

Resolves #3657


simone-silvestri (approved):

test/test_checkpointer.jl
```diff
@@ -191,5 +203,23 @@ for arch in archs
...
+        N = iteration(simulation)
+        checkpoint = "initialization_test_iteration$N.jld2"
+        simulation = initialization_test_simulation(arch, 8)
+        run!(simulation, pickup=checkpoint)
+
+        progress_cb = simulation.callbacks[:progress]
+        progress_cb.schedule.first_actuation_time
+        @test progress_cb.schedule.first_actuation_time == 4
+
+        rm("initialization_test_iteration*.jld2", force=true)
```
navidcy, line 222: fyi, no need to do that since these lines deal with that:

https://github.com/CliMA/Oceananigans.jl/blob/abb66e32d333562dd9aaeb7dd2ed8fac5e781368/docs/make.jl#L185-L202

test/test_simulations.jl
```diff
@@ -51,7 +51,6 @@ function wall_time_step_wizard_tests(arch)
     Δt = new_time_step(Δt, wizard, model)
     @test Δt ≈ diff_CFL * Δx^2 / model.closure.ν
 
-
     grid_stretched = RectilinearGrid(arch, 
                                     size = (1, 1, 1),
```
navidcy, line 55:
```suggestion
                                     size = (1, 1, 1),
```

navidcy (approved):

test/test_checkpointer.jl
glwagner: oh good we can delete then

src/Utils/schedules.jl
```diff
@@ -67,9 +69,8 @@ function (schedule::TimeInterval)(model)
     if t >= t★
         if schedule.actuations < typemax(Int)
             schedule.actuations += 1
-        else
-            schedule.first_actuation_time = t★
-            schedule.actuations = 1
+        else # re-initialize the schedule to t★
+            initialize!(schedule, t★)
```
hdrake, line 73: @glwagner, does this automatically propagate through to `AveragedTimeInterval`?

We're still finding issues with our `AveragedTimeInterval` diagnostics after picking up from a checkpoint. Does something need to be modified here:
https://github.com/CliMA/Oceananigans.jl/blob/abb66e32d333562dd9aaeb7dd2ed8fac5e781368/src/OutputWriters/windowed_time_average.jl#L89-L96

hdrake (changes requested):

src/Utils/schedules.jl
glwagner: More work is needed for `AveragedTimeInterval`. We have to design an interface for checkpointing aspects of the simulation itself, not just the `model`. Basically a time-average is a callback that has a state (like the model) and that state has to be checkpointed and restored in a similar manner to the model state.

glwagner: I suggest opening a new issue to discuss this.

glwagner: that might work! I would test it with a simple example though first!

hdrake: Yep, we're working on that and will try to add a MWE to the issue.

PR #3662, opened 2024-07-25
Prevents race condition from multiple builds attempting to use/delete the same julia install

jagoosw:
See #3661

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,15 +1,15 @@
 env:
   JULIA_VERSION: "1.10.2"
   JULIA_MINOR_VERSION: "1.10"
-  SVERDRUP_HOME: "/data5/glwagner"
-  TARTARUS_HOME: "/storage5/buildkite-agent"
+  SVERDRUP_HOME: "/data5/glwagner/$BUILDKITE_BUILD_NUMBER"
+  TARTARUS_HOME: "/storage5/buildkite-agent/$BUILDKITE_BUILD_NUMBER"
   JULIA_PKG_SERVER_REGISTRY_PREFERENCE: eager
 
 steps:
   - label: "🎪 initialize gpu environment"
     key: "init_gpu"
     env:
-      JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
+      JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia"
       TEST_GROUP: "init"
     commands:
       # Download julia binaries
@@ -34,7 +34,7 @@ steps:
   - label: "🏕️ initialize cpu environment"
     key: "init_cpu"
     env:
-      JULIA_DEPOT_PATH: "$TARTARUS_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
+      JULIA_DEPOT_PATH: "$TARTARUS_HOME/.julia"
       TEST_GROUP: "init"
       CUDA_VISIBLE_DEVICES: "-1"
       JULIA_BINDIR: "$TARTARUS_HOME/julia-$JULIA_VERSION/bin"
@@ -62,7 +62,7 @@ steps:
 
   - label: "🐿️ gpu unit tests"
     env:
-      JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
+      JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia"
       TEST_GROUP: "unit"
     commands:
       - "$SVERDRUP_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
@@ -73,7 +73,7 @@ steps:
 
   - label: "🐇 cpu unit tests"
     env:
-      JULIA_DEPOT_PATH: "$TARTARUS_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
+      JULIA_DEPOT_PATH: "$TARTARUS_HOME/.julia"
       TEST_GROUP: "unit"
       CUDA_VISIBLE_DEVICES: "-1"
     commands:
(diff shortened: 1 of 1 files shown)


jagoosw:
I'm not sure how this works for distributed tests but I don't think anything needs to change there

jagoosw:
I think everything got cancelled rather than failing but not sure why...

glwagner:
There's one segfault: https://buildkite.com/clima/oceananigans/builds/16495#0190e987-f1f0-4773-a8bd-e5f12aebebf9

.buildkite/pipeline.yml
```diff
@@ -1,15 +1,15 @@
 env:
   JULIA_VERSION: "1.10.2"
   JULIA_MINOR_VERSION: "1.10"
-  SVERDRUP_HOME: "/data5/glwagner"
-  TARTARUS_HOME: "/storage5/buildkite-agent"
+  SVERDRUP_HOME: "/data5/glwagner/$BUILDKITE_BUILD_NUMBER"
```
glwagner, line 4:
```suggestion
  SVERDRUP_HOME: "/data5/glwagner/oceananigans-buildkite-$BUILDKITE_BUILD_NUMBER"
```

.buildkite/pipeline.yml
```diff
@@ -1,15 +1,15 @@
...
+  TARTARUS_HOME: "/storage5/buildkite-agent/$BUILDKITE_BUILD_NUMBER"
```
glwagner, line 5:
```suggestion
  TARTARUS_HOME: "/storage5/buildkite-agent/oceananigans-$BUILDKITE_BUILD_NUMBER"
```

glwagner:
suggested some changes (otherwise we get a bunch of folders like "16574", "16344" with no other indication so I added a label)

glwagner:
This doesn't seem to fix the problem: https://buildkite.com/clima/oceananigans/builds/16499#0190eab5-ee7c-47da-8bb6-f0026a31e110

Although I was hopeful. But perhaps having separate depots is enough...

One issue could be that `init` is not doing enough somehow. So additional actions are taken after `init` that conflict with each other.

glwagner:
I wonder if we are not instantiating the test environment correctly.

We do instantiate here:

https://github.com/CliMA/Oceananigans.jl/blob/abb66e32d333562dd9aaeb7dd2ed8fac5e781368/.buildkite/pipeline.yml#L20

but the segmentation fault is coming from `Pkg`:

```
Stacktrace:
--
  | [1] pipeline_error
  | @ ./process.jl:565 [inlined]
  | [2] read(cmd::Cmd)
  | @ Base ./process.jl:449
  | [3] collect_artifacts(pkg_root::String; platform::Base.BinaryPlatforms.Platform)
  | @ Pkg.Operations /net/ocean/home/data44/data5/glwagner/oceananigans-buildkite-16499/julia-1.10.2/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:720
  | [4] collect_artifacts
  | @ /net/ocean/home/data44/data5/glwagner/oceananigans-buildkite-16499/julia-1.10.2/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:706 [inlined]
  | [5] download_artifacts(env::Pkg.Types.EnvCache; platform::Base.BinaryPlatforms.Platform, julia_version::VersionNumber, verbose::Bool, io::Base.DevNull)
  | @ Pkg.Operations /net/ocean/home/data44/data5/glwagner/oceananigans-buildkite-16499/julia-1.10.2/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:752
  | [6] up(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}, level::Pkg.Types.UpgradeLevel; skip_writing_project::Bool, preserve::Nothing)
  | @ Pkg.Operations /net/ocean/home/data44/data5/glwagner/oceananigans-buildkite-16499/julia-1.10.2/share/julia/stdlib/v1.10/Pkg/src/Operations.jl:1542
  | [7] up(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}; level::Pkg.Types.UpgradeLevel, mode::Pkg.Types.PackageMode, preserve::Nothing, update_registry::Bool, skip_writing_project::Bool, kwargs::@Kwargs{io::Base.DevNull})
  | @ Pkg.API /net/ocean/home/data44/data5/glwagner/oceananigans-buildkite-16499/julia-1.10.2/share/julia/stdlib/v1.10/Pkg/src/API.jl:351
  | [8] up
  | @ /net/ocean/home/data44/data5/glwagner/oceananigans-buildkite-16499/julia-1.10.2/share/julia/stdlib/v1.10/Pkg/src/API.jl:326 [inlined]
  | [9] up
  | @ /net/ocean/home/data44/data5/glwagner/oceananigans-buildkite-16499/julia-1.10.2/share/julia/stdlib/v1.10/Pkg/src/API.jl:164 [inlined]
  | [10] #resolve#143
  | @ /net/ocean/home/data44/data5/glwagner/oceananigans-buildkite-16499/julia-1.10.2/share/julia/stdlib/v1.10/Pkg/src/API.jl:357 [inlined]
  | [11] resolve
  | @ /net/ocean/home/data44/data5/glwagner/oceananigans-buildkite-16499/julia-1.10.2/share/julia/stdlib/v1.10/Pkg/src/API.jl:356 [inlined]
```

There may be some clues here: https://discourse.julialang.org/t/how-do-i-activate-the-test-environment-of-a-package/86740/3

but we may also want to reach out on the Julia slack

simone-silvestri:
I think to do this cleanly we might want to follow up with what @vchuravy was doing here #3042

simone-silvestri:
or at least we could try following that route (it seems like there were some problems also with that implementation)

glwagner:
If the issue is downloading julia, I think this is something that we only do if we update the tests to use a new julia version right?

jagoosw:
That might be another solution.

As it stands each time the pipeline is deployed it downloads julia:
https://github.com/CliMA/Oceananigans.jl/blob/eade289dc857396d6d73abde0a4037c1a089d980/.buildkite/pipeline.yml#L20-L24
And then deletes it when its done:
https://github.com/CliMA/Oceananigans.jl/blob/eade289dc857396d6d73abde0a4037c1a089d980/.buildkite/pipeline.yml#L862-L866

I guess this was originally done to make it easy to change between versions instead of having to manually install on the agents, but it might be more straight forward to do that.

glwagner:
> That might be another solution.
> 
> As it stands each time the pipeline is deployed it downloads julia:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/eade289dc857396d6d73abde0a4037c1a089d980/.buildkite/pipeline.yml#L20-L24
> 
> 
> And then deletes it when its done:
> https://github.com/CliMA/Oceananigans.jl/blob/eade289dc857396d6d73abde0a4037c1a089d980/.buildkite/pipeline.yml#L862-L866
> 
> I guess this was originally done to make it easy to change between versions instead of having to manually install on the agents, but it might be more straight forward to do that.

`wget -N` doesn't download a file unless it has a newer time-stamp, so I believe that should avoid re-downloading.

glwagner:
closing this as stale but re-open if needed

PR #3666, opened 2024-07-26
Bug in NetCDF output for curvilinear immersed grids

simone-silvestri:
`NetCDFOutputwriter` has a bug for Immersed curvilinear grids, in practice for an immersed boundary the default 
`native_dimensions_for_netcdf_output` is called instead of the one for curvilinear grids.

This PR should address the issue. 
Also it seems to not be tested so I ll add a test


src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -49,21 +49,16 @@ zdim(::Nothing) = tuple()
...
-         "yC" => parent(φnodes(grid, Center(); with_halos=true))[parent_index_range(indices["yC"][2], Center(), TY(), Hy)],
-         "yF" => parent(φnodes(grid, Face();   with_halos=true))[parent_index_range(indices["yF"][2],   Face(), TY(), Hy)],
-         "zC" => parent(znodes(grid, Center(); with_halos=true))[parent_index_range(indices["zC"][3], Center(), TZ(), Hz)],
-         "zF" => parent(znodes(grid, Face();   with_halos=true))[parent_index_range(indices["zF"][3],   Face(), TZ(), Hz)])
+    Dict("xC" => nodes(grid, (c, c, c))[1],
+         "xF" => nodes(grid, (f, c, c))[1],
+         "yC" => nodes(grid, (c, c, c))[2],
+         "yF" => nodes(grid, (c, f, c))[2],
+         "zC" => nodes(grid, (c, c, c))[3],
+         "zF" => nodes(grid, (c, c, f))[3])
```
glwagner, line 61: using `nodes[1]` is a little hacky.

in `nodes_and_spacings` we introduce `ξnode`, `ηnode`, `rnode` to denote the 1, 2, 3 extrinsic nodes. Probably you just want to add the plural version of this function and then define it appropriately for the three grids.

glwagner: https://github.com/CliMA/Oceananigans.jl/blob/abb66e32d333562dd9aaeb7dd2ed8fac5e781368/src/Grids/nodes_and_spacings.jl#L31-L33

ali-ramadhan:
Is this PR looking for an adopter? Not sure what the original error was but happy to work on this PR since it might be the cause of the error I reported in issue https://github.com/CliMA/Oceananigans.jl/issues/3775.

simone-silvestri:
Hey @ali-ramadhan, if you want to chaperone this PR to conclusion that would be great! I guess you just need to add 
`ξnode`, `ηnode`, and `rnode` for an immersed boundary grid and the corresponding plural version to be used instead of `nodes(...)[i]` and the PR should be ready.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -49,16 +49,59 @@ zdim(::Nothing) = tuple()
...
-const f = Face()
-
-native_dimensions_for_netcdf_output(grid, indices, TX, TY, TZ, Hx, Hy, Hz) =
-    Dict("xC" => nodes(grid, (c, c, c))[1],
-         "xF" => nodes(grid, (f, c, c))[1],
-         "yC" => nodes(grid, (c, c, c))[2],
-         "yF" => nodes(grid, (c, f, c))[2],
-         "zC" => nodes(grid, (c, c, c))[3],
-         "zF" => nodes(grid, (c, c, f))[3])
+function native_dimensions_for_netcdf_output(grid, indices, TX, TY, TZ, Hx, Hy, Hz)
```
simone-silvestri, line 52: I wonder if, as @glwagner suggested, it wouldn't be better to have `ξnodes`, `ηnodes` and `rnodes` that return the native coordinates of the grid (as we have now `ξnode`, `ηnode`, and `rnode`). That would avoid needing to specialize this function

ali-ramadhan: Agreed and already done!

ali-ramadhan:
Tests pass locally on CPU and GPU so PR should be ready for review! Sorry my VS Code trimmed whitespace.

@simone-silvestri I can't request your review since you opened the PR, but if you approve via text I can click the approve button and merge lol.

simone-silvestri:
Looks good!

PR #3674, opened 2024-07-30
Update `time_step!` docstring for AB2 

### LINKED ISSUE #3565: What's the point of neglecting the computation of tendencies when taking a time-step?
The docstring for `time_step!` says that "Setting `compute_tendencies=false` will not calculate new tendencies ". But why would anybody want to do that? Since tendencies are computed at the end of a time-step rather than before, this would prevent one from taking a subsequent time-step. I think this should be clarified in the docstring. (It's also missing a period.)

https://github.com/CliMA/Oceananigans.jl/blob/1ee317b9b6986436f18dd12d7fcdd7067d9234b7/src/TimeSteppers/quasi_adams_bashforth_2.jl#L70

cc @simone-silvestri

simone-silvestri:
closes #3565

### DIFF
--- a/src/TimeSteppers/quasi_adams_bashforth_2.jl
+++ b/src/TimeSteppers/quasi_adams_bashforth_2.jl
@@ -61,19 +61,21 @@ reset!(timestepper::QuasiAdamsBashforth2TimeStepper) = nothing
 
 Step forward `model` one time step `Δt` with a 2nd-order Adams-Bashforth method and
 pressure-correction substep. Setting `euler=true` will take a forward Euler time step.
-The new tendencies are calculated in the `update_step!` at the end of the `time_step!` function.
+The tendencies are calculated by the `update_step!` at the end of the `time_step!` function.
 
-The steps of the AB2 algorithm are:
+The steps of the Quasi-Adams-Bashforth second-order (AB2) algorithm are:
 
-1. If this the first time step `(model.clock.iteration == 0)`, then `update_state!` and calculate the tendencies
-2. advance tracers in time and compute predictor velocities (including implicit vertical diffusion)
-3. solve the elliptic equation for pressure (three dimensional for the non-hydrostatic model and two-dimensional for the hydrostatic model)
-4. correct the velocities based on the results of step 3.
-5. Store the old tendencies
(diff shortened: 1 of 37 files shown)


glwagner (approved):

src/TimeSteppers/quasi_adams_bashforth_2.jl
```diff
@@ -61,7 +61,19 @@ reset!(timestepper::QuasiAdamsBashforth2TimeStepper) = nothing
...
+
+1. If this the first time step `(model.clock.iteration == 0)`, then `update_state!` and calculate the tendencies
+2. advance tracers in time and compute predictor velocities (including implicit vertical diffusion)
+3. solve the elliptic equation for pressure (three dimensional for the non-hydrostatic model and two-dimensional for the hydrostatic model)
+4. correct the velocities based on the results of step 3.
+5. Store the old tendencies
+6. update the model state
+7. If `compute_tendencies == true`, compute the tendencies for the next time step. NOTE: if `compute_tendencies == false` the new tendencies will not be calculated! 
+It is never suggested to set `compute_tendencies == false` except for debugging purposes.
+
```
navidcy, line 76:
```suggestion
The tendencies are calculated by the `update_step!` at the end of the `time_step!` function.

The steps of the Quasi-Adams-Bashforth second-order (AB2) algorithm are:

1. If this the first time step (`model.clock.iteration == 0`), then call `update_state!` and calculate the tendencies.
2. Advance tracers in time and compute predictor velocities (including implicit vertical diffusion).
3. Solve the elliptic equation for pressure (three dimensional for the non-hydrostatic model, two-dimensional for the hydrostatic model).
4. Correct the velocities based on the results of step 3.
5. Store the old tendencies.
6. Update the model state.
7. If `compute_tendencies == true`, compute the tendencies for the next time step.

!!! warning "`compute_tendencies` kwarg"
    If `compute_tendencies == false` then the new tendencies at the last step are _not_ calculated!
    Setting `compute_tendencies == false` is not recommended _except_ for debugging purposes.
```

navidcy (approved):
lgtm; definitely improved!

PR #3676, opened 2024-08-02
Introduces `SumOfFields`

jagoosw:
This PR adds a method to `mask_immersed_field!` for `SumOfArrays`. I was trying to make a model which has an auxiliary field which is a `SumOfArrays` but ran into the error that all of the prognostic and auxiliary fields in the `HydrostaticFreeSurfaceModel` get masked during the `update_state!` step.

(Ran into the issue here: https://github.com/OceanBioME/OceanBioME.jl/pull/195)

### DIFF
--- a/src/Fields/sum_of_fields.jl
+++ b/src/Fields/sum_of_fields.jl
@@ -1,0 +1,79 @@
+using Base: @propagate_inbounds
+using Oceananigans.Fields: AbstractField, location
+
+import Adapt: adapt_structure
+import Base: getindex, show, summary
+import Oceananigans.BoundaryConditions: fill_halo_regions!
+import Oceananigans.ImmersedBoundaries: mask_immersed_field!, mask_immersed_field_xy!
+
+"""
+    SumOfFields{N, F}
+
+`SumOfFields` objects hold `N` field/fields and return their sum when indexed.
+"""
+struct SumOfFields{N, LX, LY, LZ, G, T, F} <: AbstractField{LX, LY, LZ, G, T, 3}
+    fields :: F
+      grid :: G
+    
+    function SumOfFields{N}(fields...) where N
+        grid = first(fields).grid
+        loc = location(first(fields))
+
+        all(f.grid == grid for f in fields) || 
+            throw(ArgumentError("All `fields` must be on the same grid"))
+
+        all(location(f) == loc for f in fields) || 
+            throw(ArgumentError("All `fields` must be on the same location"))
+
+        T = eltype(first(fields).data)
+        F = typeof(fields)
+        G = typeof(grid)
+
+        return new{N, loc..., G, T, F}(fields, grid)
+    end
+end
+
+adapt_structure(to, sum::SumOfFields{N}) where N = SumOfFields{N}((adapt_structure(to, field) for field in sum.fields)...)
+
+summary(::SumOfFields{N, LX, LY, LZ}) where {N, LX, LY, LZ}  = "Sum of $N fields at ($LX, $LY, $LZ)"
+
+show(io::IO, sof::SumOfFields{N, LX, LY, LZ}) where {N, LX, LY, LZ} = print(io, summary(sof))
+
+@propagate_inbounds function getindex(s::SumOfFields{N, LX, LY, LZ, G, T, F}, i...) where {N, LX, LY, LZ, G, T, F}
+    first = getindex(SumOfFields{3, F, LX, LY, LZ, G, T}((s.fields[1], s.fields[2], s.fields[3]), s.grid), i...)
+    last = getindex(SumOfFields{N - 3, F, LX, LY, LZ, G, T}(s.fields[4:N], s.grid), i...)
+    return first + last
+end
+
+@propagate_inbounds getindex(s::SumOfFields{1}, i...) = getindex(s.fields[1], i...)
+@propagate_inbounds getindex(s::SumOfFields{2}, i...) = getindex(s.fields[1], i...) + getindex(s.fields[2], i...)
+
+@propagate_inbounds getindex(s::SumOfFields{3}, i...) = 
+    getindex(s.fields[1], i...) + getindex(s.fields[2], i...) + getindex(s.fields[3], i...)
+    
+@propagate_inbounds getindex(s::SumOfFields{4}, i...) = 
+    getindex(s.fields[1], i...) + getindex(s.fields[2], i...) + getindex(s.fields[3], i...) + getindex(s.fields[4], i...)
+
+function mask_immersed_field!(sof::SumOfFields, value=zero(eltype(sof.grid)))
+    for field in sof.fields
+        mask_immersed_field!(field, value)
+    end
+    
+    return nothing
+end
+
+function mask_immersed_field_xy!(sof::SumOfFields, value=zero(eltype(sof.grid)))
+    for field in sof.fields
+        mask_immersed_field_xy!(field, value)
+    end
+    
+    return nothing
+end
+
+function fill_halo_regions!(sof::SumOfFields)
+    for f in sof.fields
+        fill_halo_regions!(f)
+    end
+    
+    return nothing
+end
--- a/src/Fields/sum_of_fields.jl
+++ b/src/Fields/sum_of_fields.jl
@@ -20,10 +20,10 @@ struct SumOfFields{N, LX, LY, LZ, G, T, F} <: AbstractField{LX, LY, LZ, G, T, 3}
         loc = location(first(fields))
 
         all(f.grid == grid for f in fields) || 
-            throw(ArgumentError("All `fields` must be on the same grid"))
+            throw(ArgumentError("All `fields` in `SumOfFields` must be on the same grid"))
 
         all(location(f) == loc for f in fields) || 
-            throw(ArgumentError("All `fields` must be on the same location"))
+            throw(ArgumentError("All `fields` in `SumOfFields` must be on the same location"))
 
         T = eltype(first(fields).data)
         F = typeof(fields)
--- a/src/ImmersedBoundaries/mask_immersed_field.jl
+++ b/src/ImmersedBoundaries/mask_immersed_field.jl
@@ -10,6 +10,13 @@ mask_immersed_field!(field, grid, loc, value) = nothing
 mask_immersed_field!(field::Field, value=zero(eltype(field.grid))) =
     mask_immersed_field!(field, field.grid, location(field), value)
 
+function mask_immersed_field!(sumofarrays::SumOfArrays, value=zero(eltype(sumofarrays.arrays[1])))
+    loc = @inbounds location(sumofarrays.arrays[1])
+    grid = @inbounds sumofarrays.arrays[1].grid
+
+   return mask_immersed_field!(sumofarrays, grid, loc, value)
+end
+
 """
     mask_immersed_field!(field::Field, grid::ImmersedBoundaryGrid, loc, value)
 
@@ -23,7 +30,7 @@ function mask_immersed_field!(field::Field, grid::ImmersedBoundaryGrid, loc, val
 end
 
 function mask_immersed_field!(sumofarrays::SumOfArrays, grid::ImmersedBoundaryGrid, loc, value; k, mask)
-    arch = architecture(sumofarrays.arrays[1])
+    arch = @inbounds architecture(sumofarrays.arrays[1])
     loc = instantiate.(loc)
 
     for field in sumofarrays.arrays
@@ -62,7 +69,7 @@ end
 end
 
 function mask_immersed_field_xy!(sumofarrays::SumOfArrays, grid::ImmersedBoundaryGrid, loc, value; k, mask)
-    arch = architecture(sumofarrays.arrays[1])
+    arch = @inbounds architecture(sumofarrays.arrays[1])
     loc = instantiate.(loc)
 
     for field in sumofarrays.arrays
--- a/src/ImmersedBoundaries/mask_immersed_field.jl
+++ b/src/ImmersedBoundaries/mask_immersed_field.jl
@@ -11,10 +11,11 @@ mask_immersed_field!(field::Field, value=zero(eltype(field.grid))) =
     mask_immersed_field!(field, field.grid, location(field), value)
 
 function mask_immersed_field!(sumofarrays::SumOfArrays, value=zero(eltype(sumofarrays.arrays[1])))
-    loc = @inbounds location(sumofarrays.arrays[1])
-    grid = @inbounds sumofarrays.arrays[1].grid
-
-   return mask_immersed_field!(sumofarrays, grid, loc, value)
+    for field in sumofarrays.arrays
+        mask_immersed_field!(field, value)
+    end
+    
+    return nothing
 end
 
 """
@@ -24,20 +25,8 @@ masks `field` defined on `grid` with a value `val` at locations where `periphera
 """
 function mask_immersed_field!(field::Field, grid::ImmersedBoundaryGrid, loc, value)
     arch = architecture(field)
-    loc = instantiate.(loc)
+    loc = map(instantiate, loc)
     launch!(arch, grid, :xyz, _mask_immersed_field!, field, loc, grid, value)
-    return nothing
-end
-
-function mask_immersed_field!(sumofarrays::SumOfArrays, grid::ImmersedBoundaryGrid, loc, value; k, mask)
-    arch = @inbounds architecture(sumofarrays.arrays[1])
-    loc = instantiate.(loc)
-
-    for field in sumofarrays.arrays
-        launch!(arch, grid, :xyz,
-                _mask_immersed_field!, field, loc, grid, value, k, mask)
-    end
-
     return nothing
 end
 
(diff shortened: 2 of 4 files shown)


jagoosw:
Ran into the issue here: https://github.com/OceanBioME/OceanBioME.jl/pull/195

src/ImmersedBoundaries/mask_immersed_field.jl
```diff
@@ -9,6 +10,13 @@ mask_immersed_field!(field, grid, loc, value) = nothing
 mask_immersed_field!(field::Field, value=zero(eltype(field.grid))) =
     mask_immersed_field!(field, field.grid, location(field), value)
 
+function mask_immersed_field!(sumofarrays::SumOfArrays, value=zero(eltype(sumofarrays.arrays[1])))
+    loc = @inbounds location(sumofarrays.arrays[1])
```
glwagner, line 14:
```suggestion
    loc = @inbounds location(first(sumofarrays))
```

src/ImmersedBoundaries/mask_immersed_field.jl
```diff
@@ -49,6 +68,18 @@ end
     @inbounds field[i, j, k] = scalar_mask(i, j, k, grid, grid.immersed_boundary, loc..., value, field, mask)
 end
 
+function mask_immersed_field_xy!(sumofarrays::SumOfArrays, grid::ImmersedBoundaryGrid, loc, value; k, mask)
+    arch = @inbounds architecture(sumofarrays.arrays[1])
+    loc = instantiate.(loc)
```
glwagner, line 73:
```suggestion
    loc = map(instantiate, loc)
```

broadcasting takes longer to compile and can lead to type inference issues

src/ImmersedBoundaries/mask_immersed_field.jl
```diff
@@ -49,6 +68,18 @@ end
     @inbounds field[i, j, k] = scalar_mask(i, j, k, grid, grid.immersed_boundary, loc..., value, field, mask)
 end
 
+function mask_immersed_field_xy!(sumofarrays::SumOfArrays, grid::ImmersedBoundaryGrid, loc, value; k, mask)
+    arch = @inbounds architecture(sumofarrays.arrays[1])
```
glwagner, line 72:
```suggestion
    arch = architecture(first(sumofarrays))
```

glwagner:
It might make sense to take a step back and ask what you're trying to accomplish because I'm not quite sure this makes sense. It might work for you, but seems to rely on hidden assumptions about `SumOfArrays` which wasn't designed for this

jagoosw:
Hmm yeah, that is true, we never make any guarantees about what is in a `SumOfArrays`, but I think even at this point we're trusting developers that use it to make sure that the sum is valid (i.e. on the same grid and location), and to only put fields in it where the rest of the code would expect fields right?

In the same way there is no check on what e.g. biogeochemical_auxiliary_fields` is returning so it could return just a normal array which would also break at this step. 

I just rearranged it a bit so it just runs `mask_immersed_field!(field, value)` on each array element rather than `mask_immersed_field!(field, loc, grid, value)` which is maybe better.

glwagner:
> but I think even at this point we're trusting developers that use it to make sure that the sum is valid (i.e. on the same grid and location), and to only put fields in it where the rest of the code would expect fields right?

Well no --- we don't trust developers to do that. We only use it inside a kernel and in a case we know it is correctly. This PR does something different, it introduces "support" for `SumOfArray` that suggests it should be used to represent a sum of fields, but doesn't implement any other facilities that correspond to such support. We want users to be able to trust the code we write so I feel this isn't the most helpful thing to do.

It's interesting to consider an abstraction that represents a sum of fields (though note we already have this through `MultiaryOperation`, but maybe there is a reason you would like something different?). I would put a bit more effort into that than we put into `SumOfArrays`.

glwagner:
Also with `SumOfFields` we can put in proper support for other things. It can be an `AbstractField` and be used in operations (perhaps the biggest advantages because users will want to do diagnostics on the sum probably), we can give it a location and expose its grid, and we can support `fill_halo_regions` and `mask_immersed_field`. It's a decent idea provided that the normal way of summing fields that produces a `BinaryOperation` or `MultiaryOperation` won't work.

jagoosw:
Okay, I understand your point. Since it's not part of the public API I wasn't thinking that this suggests to a user that it should be used, but more made it possible for developers to do things like what my other PR does.

What I meant about trusting users for auxiliary fields is that you can do e.g. this:
```julia
model = NonhydrostaticModel(; grid, auxiliary_fields=(; A = randn(1, 2))
```
and it doesn't error, and the only reason the same for a `HydrostaticFreeSurfaceModel` errors is from `mask_immersed_field!` because the `mask_immersed_field!(field::Field, value=zero(eltype(field.grid)))` method doesn't have a fallback defined.

So this change wouldn't affect how the model failed because putting arrays etc. into the sum would do the same thing.

But yeah, having a `SumOfFields` makes sense and I can try and do that. The main reason I wanted to use this vs and operation is the speed because when I checked indexing into a sum of arrays is >5x faster than into the equivalent operation (for a 3 field sum).

src/Fields/sum_of_fields.jl
```diff
@@ -0,0 +1,79 @@
...
+struct SumOfFields{N, LX, LY, LZ, G, T, F} <: AbstractField{LX, LY, LZ, G, T, 3}
+    fields :: F
+      grid :: G
+    
+    function SumOfFields{N}(fields...) where N
+        grid = first(fields).grid
+        loc = location(first(fields))
+
+        all(f.grid == grid for f in fields) || 
+            throw(ArgumentError("All `fields` must be on the same grid"))
```
glwagner, line 23:
```suggestion
            throw(ArgumentError("All `fields` in `SumOfFields` must be on the same grid"))
```

src/Fields/sum_of_fields.jl
```diff
@@ -0,0 +1,79 @@
...
+
+        all(location(f) == loc for f in fields) || 
+            throw(ArgumentError("All `fields` must be on the same location"))
```
glwagner, line 26:
```suggestion
            throw(ArgumentError("All `fields` in `SumOfFields` must be on the same location"))
```

src/Fields/sum_of_fields.jl
```diff
@@ -0,0 +1,79 @@
...
+        F = typeof(fields)
+        G = typeof(grid)
+
+        return new{N, loc..., G, T, F}(fields, grid)
+    end
+end
+
+adapt_structure(to, sum::SumOfFields{N}) where N = SumOfFields{N}((adapt_structure(to, field) for field in sum.fields)...)
+
+summary(::SumOfFields{N, LX, LY, LZ}) where {N, LX, LY, LZ}  = "Sum of $N fields at ($LX, $LY, $LZ)"
```
glwagner, line 38: This summary does not state the name of the type (`SumOfFields`). The purpose of `summary` and `show` is to give information about the object so I think it's important to state the name.

Also the `summary` should be more uniform with other field summaries

src/Fields/sum_of_fields.jl
```diff
@@ -0,0 +1,79 @@
...
+
+show(io::IO, sof::SumOfFields{N, LX, LY, LZ}) where {N, LX, LY, LZ} = print(io, summary(sof))
+
+@propagate_inbounds function getindex(s::SumOfFields{N, LX, LY, LZ, G, T, F}, i...) where {N, LX, LY, LZ, G, T, F}
+    first = getindex(SumOfFields{3, F, LX, LY, LZ, G, T}((s.fields[1], s.fields[2], s.fields[3]), s.grid), i...)
+    last = getindex(SumOfFields{N - 3, F, LX, LY, LZ, G, T}(s.fields[4:N], s.grid), i...)
+    return first + last
```
glwagner, line 45: write this on more lines please.

as a rule, try not to nest function calls within one another. Sometimes one-argument functions can be ok. Here though, the inner call has 7 type parameters and two arguments while the outer call has two arguments.

src/Fields/sum_of_fields.jl
```diff
@@ -0,0 +1,79 @@
...
+    return nothing
+end
+
+function mask_immersed_field_xy!(sof::SumOfFields, value=zero(eltype(sof.grid)))
+    for field in sof.fields
+        mask_immersed_field_xy!(field, value)
+    end
+    
+    return nothing
+end
```
glwagner, line 71: All `mask_immersed_field` belong in the immersed boundaries module

glwagner:
It might have been better to have a new PR for this because the conversation will be hard to follow, but we can continue.

We need a mission statement for `SumOfFields`. Why doesn't `MultiaryOperation` work?

glwagner:
> when I checked indexing into a sum of arrays is >5x faster than into the equivalent operation (for a 3 field sum).

Doesn't this suggest there is a problem with `MultiaryOperation`? It might be nice to document this.

glwagner:
> What I meant about trusting users for auxiliary fields is that you can do e.g. this: `model = NonhydrostaticModel(; grid, auxiliary_fields=(; A = randn(1, 2))` and it doesn't error

Why do we want to support arrays in `auxiliary_fields`?

The main usage for `auxiliary_fields` is that they can be invoked in continuous-form forcing functions. In that usage they have to be `AbstractField`, otherwise we can't figure out how to interpolate them into a forcing function. `auxiliary_fields` isn't meant to be a place to store just any information needed for a model (what would the purpose of that be?) We can adapt `auxiliary_fields` but it would be good to understand why. Separately, perhaps we should check that all `auxiliary_fields` are `AbstractField`.

src/Fields/sum_of_fields.jl
glwagner: I suggest this

```julia
function Base.summary(field::SumOfFields{N}) where N
    LX, LY, LZ = location(field)
    prefix = string(size_summary(size(field)), " SumOfFields{$LX, $LY, $LZ} with $N fields")

    reduced_dims = reduced_dimensions(field)

    suffix = reduced_dims === () ?
        string(" on ", grid_name(field), " on ", summary(architecture(field))) :
        string(" reduced over dims = ", reduced_dims,
               " on ", grid_name(field), " on ", summary(architecture(field)))

    return string(prefix, suffix)
end
```

This is plagiarized from `summary` for `Fields`.

glwagner: There's also a nice `show` method for NamedTuple of Fields that could inspire a `show` method for `SumOfFields`

src/Fields/sum_of_fields.jl
```diff
@@ -0,0 +1,79 @@
...
+"""
+struct SumOfFields{N, LX, LY, LZ, G, T, F} <: AbstractField{LX, LY, LZ, G, T, 3}
+    fields :: F
+      grid :: G
+    
+    function SumOfFields{N}(fields...) where N
+        grid = first(fields).grid
+        loc = location(first(fields))
+
+        all(f.grid == grid for f in fields) ||
```
glwagner, line 22:
```suggestion
        all(f.grid === grid for f in fields) || 
```

I think object equality is appropriate here. And below. (This will be faster because we do not have to perform numeric comparison)

src/Fields/sum_of_fields.jl
```diff
@@ -0,0 +1,79 @@
...
+
+        T = eltype(first(fields).data)
+        F = typeof(fields)
+        G = typeof(grid)
+
+        return new{N, loc..., G, T, F}(fields, grid)
+    end
+end
+
+adapt_structure(to, sum::SumOfFields{N}) where N = SumOfFields{N}((adapt_structure(to, field) for field in sum.fields)...)
```
glwagner, line 36: please write this on more lines

jagoosw:
So I started making the suggested changes but then decided to do more benchmarking and it appears that when I interpolate the sums into the benchmark properly there isn't a difference (i.e. https://juliaci.github.io/BenchmarkTools.jl/stable/manual/#Interpolating-values-into-benchmark-expressions):

```julia
julia> @btime sof[1, 2, 3]
  45.412 ns (1 allocation: 16 bytes)
-1.0065959327459124

julia> @btime bo[1, 2, 3]
  44.571 ns (1 allocation: 16 bytes)
-1.0065959327459124

julia> @btime fbo[1, 2, 3]
  238.209 ns (1 allocation: 16 bytes)
-1.0065959327459124

julia> @btime $(sof)[1, 2, 3]
  3.958 ns (0 allocations: 0 bytes)
-1.0065959327459124

julia> @btime $(bo)[1, 2, 3]
  2.708 ns (0 allocations: 0 bytes)
-1.0065959327459124

julia> @btime $(fbo)[1, 2, 3]
  2.666 ns (0 allocations: 0 bytes)
-1.0065959327459124
```
where 
```julia
julia> sof = SumOfFields{3}(c[1:3]...)
200×200×200 SumOfFields{3, Center, Center, Center} on RectilinearGrid on CPU
└── grid: 200×200×200 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo

julia> bo = sum(c[1:3])
BinaryOperation at (Center, Center, Center)
├── grid: 200×200×200 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
└── tree: 
    + at (Center, Center, Center)
    ├── + at (Center, Center, Center)
    │   ├── 200×200×200 Field{Center, Center, Center} on RectilinearGrid on CPU
    │   └── 200×200×200 Field{Center, Center, Center} on RectilinearGrid on CPU
    └── 200×200×200 Field{Center, Center, Center} on RectilinearGrid on CPU

julia> fbo = c[1] + c[2] + c[3]
MultiaryOperation at (Center, Center, Center)
├── grid: 200×200×200 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
└── tree: 
    + at (Center, Center, Center)
    ├── 200×200×200 Field{Center, Center, Center} on RectilinearGrid on CPU
    ├── 200×200×200 Field{Center, Center, Center} on RectilinearGrid on CPU
    └── 200×200×200 Field{Center, Center, Center} on RectilinearGrid on CPU
```

To check that the interpolation into the benchmark wasn't important I also tried benchmarking a kernel that fills an array with the sum and got these results:
```julia
julia> @btime launch!(grid.architecture, grid, :xyz, _fill_with_sum!, A, sof)
  20.074 ms (20 allocations: 13.67 KiB)

julia> @btime launch!(grid.architecture, grid, :xyz, _fill_with_sum!, A, bo)
  20.167 ms (20 allocations: 16.77 KiB)

julia> @btime launch!(grid.architecture, grid, :xyz, _fill_with_sum!, A, fbo)
  1.289 s (20 allocations: 13.67 KiB)

```

So in conclusion `SumOfFields` isn't needed (but that is a weird result with `fbo`.

glwagner:
Ok. The case where I think `SumOfFields` could be needed is if one runs into parameter space issues. This is a compilation issue that is very hard to fix (performance problems, on the other hand, should be fixable).

jagoosw:
I'm now having the problem that if I return a `BinaryOperation` in `biogeochemical_auxiliary_fields` then it gets stuck on `mask_immersed_field!` again, would it be reasonable to define a fallback like `mask_immersed_field!(field) = nothing`?

This fallback:
https://github.com/CliMA/Oceananigans.jl/blob/bf767af3c40c049ebc9499b7553065a9ef350178/src/ImmersedBoundaries/mask_immersed_field.jl#L8
is already there but since we get there via:
https://github.com/CliMA/Oceananigans.jl/blob/bf767af3c40c049ebc9499b7553065a9ef350178/src/ImmersedBoundaries/mask_immersed_field.jl#L9-L10
it fails

glwagner:
Should we extend `mask_immersed_field` for `BinaryOperation`? And for `MultiaryOperation`? This might be useful also for post-processing

glwagner:
We don't want a fallback I don't think --- the failure is intentional. We want to know when we need to define a new method

simone-silvestri:
Masking an operation might be a bit tough because we do not know "a priori" what is the neutral value and most likely there are many ways to "mask" an operation by acting on the memory in the operands. How would this look like practically?

glwagner:
There are a few ways it could be implemented. One simple way is

```julia
function mask_immersed_field(bop::BinaryOperation, value=zero(b.grid))
    mask_immersed_field(bop.a, value)
    mask_immersed_field(bop.b, value)
    return nothing
end
```

One could also be more specific, eg

```julia
const FieldBinaryOperation = BinaryOperation{<:Any, <:Any, <:Any, <:Any, <:Field, <:Field}
```

Then we are sure its a binary operation between two fields. And the operation can be restricted to addition as well.

Just some ideas...

jagoosw:
Okay, that all makes sense.

For `BinaryOperation` I guess we might want to allow the an entry in the operation to be a number so we could define:
```
mask_immersed_field(::Number, value) = nothing
```
but otherwise, I guess would work.

Perhaps if we only do this for `BinaryOperations` we could do:

```julia
#fallback for `+` and `-`, I guess for `*` too since `a` could just be a number so if we only masked `a` it wouldn't be correct
function mask_immersed_field(bop::BinaryOperation, value=zero(b.grid))
    mask_immersed_field(bop.a, value)
    mask_immersed_field(bop.b, value)
    return nothing
end


function mask_immersed_field(bop::BinaryOperation{<:Any, <:Any, <:Any, typeof(/)}, value=zero(b.grid))
    mask_immersed_field(bop.a, value)
    return nothing
end

function mask_immersed_field(bop::BinaryOperation{<:Any, <:Any, <:Any, typeof(^)}, value=zero(b.grid))
    mask_immersed_field(bop.a, value)
    mask_immersed_field(bop.a, one(b.grid))
    return nothing
end
```
Which would make the values correct for all binary operations I think?

glwagner:
I'm ok to try this out, hopefully we won't run into issues

simone-silvestri:
For `*` you can just mask one of the so you save computation time (you can use the same function as for `/`)

PR #3683, opened 2024-08-06
Adds methods to `mask_immersed_field!` for `BinaryOperations`

jagoosw:
As discussed in #3676, this PR attempts to add methods to `mask_immersed_field!` for `BinaryOperations` so they can be passed as auxiliary fields.

This works for binary operations which combine fields or a field and a number. To make + and - work I have to set the value to -/+ the number value for the field.

mask_immersed


glwagner (approved):
Well done.

src/ImmersedBoundaries/mask_immersed_field.jl
```diff
@@ -12,9 +12,15 @@ mask_immersed_field!(field::Field, value=zero(eltype(field.grid))) =
...
-    mask_immersed_field!(bop.b, value)
+function mask_immersed_field!(bop::BinaryOperation{<:Any, <:Any, <:Any, typeof(+)}, value=zero(eltype(bop)))
+    mask_immersed_field!(bop.a, ifelse(bop.b isa Number, -bop.b, value))
+    mask_immersed_field!(bop.b, ifelse(bop.a isa Number, -bop.a, value))
+    return nothing
+end
+
+function mask_immersed_field!(bop::BinaryOperation{<:Any, <:Any, <:Any, typeof(-)}, value=zero(eltype(bop)))
+    mask_immersed_field!(bop.a, ifelse(bop.b isa Number, bop.b, value))
+    mask_immersed_field!(bop.b, ifelse(bop.a isa Number, bop.a, value))
```
glwagner, line 23: please try to avoid nesting function calls within one another

jagoosw: Fixed now

jagoosw:
I think the only place this will fail is for number ^ field since we can never make this zero. We could do:
```julia
b_value = ifelse(a < 0, Inf, -Inf)
```
but this also fails in the case of `a==1`, although that would be a bit of a strange operation to have written?

src/ImmersedBoundaries/mask_immersed_field.jl
```diff
@@ -13,14 +13,20 @@ mask_immersed_field!(field::Field, value=zero(eltype(field.grid))) =
 mask_immersed_field!(::Number, args...) = nothing
 
 function mask_immersed_field!(bop::BinaryOperation{<:Any, <:Any, <:Any, typeof(+)}, value=zero(eltype(bop)))
-    mask_immersed_field!(bop.a, ifelse(bop.b isa Number, -bop.b, value))
-    mask_immersed_field!(bop.b, ifelse(bop.a isa Number, -bop.a, value))
+    a_value = ifelse(bop.b isa Number, -bop.b, value)
+    mask_immersed_field!(bop.a, )
```
glwagner, line 17:
```suggestion
    mask_immersed_field!(bop.a, a_value)
```

glwagner:
Is it actually important to support the operation `^`?

glwagner:
You don't need to use `ifelse` here by the way, `if` is ok. This is not inside a kernel.

glwagner:
Note that the masking operation is really supposed to set the field to some prescribed _value_. Not always "0". So it's going to be challenge to cover all cases it would seem.

Do we have to support operations that involve numbers and fields? We could just support fields with fields. Things are easier.

Also there is a problem if a field appears "on its own", and also in a binary operation. Then one masking operation could cancel the other for operations like `/` and `^`. Seems like a bit of a rabbit hole.

jagoosw:
I think it would be a lot more straight forward to support just + and - and we can guarantee the value is correct then too

glwagner:
> I think it would be a lot more straight forward to support just + and - and we can guarantee the value is correct then too

Ok, let's do that

PR #3627, opened 2024-06-18
Support non-trivial horizontal coordinates for single column grids

glwagner:
This PR makes it so we can build single column grids with non-trivial horizontal coordinates and nice syntax.

Resolves #3626 

@simone-silvestri

PR #3685, opened 2024-08-06
Fixs fallback for `AbstractContinuousFormBiogeochemistry`

jagoosw:
Now that `xnode` etc. return `nothing` for flat dimensions this fallback breaks when `x` is flat


glwagner (approved):

jagoosw:
I think this was introduced in #3627 because flat dimensions used to return 1 but now return nothing

PR #3689, opened 2024-08-07
Distributed tridiagonal Fourier solver

simone-silvestri:
This PR introduces a distributed pressure solver for grids that are stretched in one direction. The algorithm implemented here is the same as the one described in #2538 but it uses the FFT and transposes built in #3279


src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -200,8 +200,8 @@ function validate_global_grid(global_grid::RectilinearGrid)
                Please rotate the domain to obtain the required topology")
     end
     
-    if !(global_grid isa XYZRegularRG) 
-        throw("Stretched directions are not supported with distributed grids at the moment.")
+    if !(global_grid isa YZRegularRG) && !(global_grid isa XYRegularRG) && !(global_grid isa XZRegularRG) 
+        throw(ArgumentError("Only stretching in one direction is supported with distributed grids at the moment."))
```
glwagner, line 204: This error doesn't belong here. 

The error message should say something like

> X type does not support `RectilinearGrid` on `Distributed` architectures that have variably-spaced cells in more than one direction. Cells may be variably-spaced in at most one direction.

In other words you have to explain 1) where the problem arises specifically (`ArgumentError` should arise from a constructor ideally to drive this home) and 2) specifically what the issue is. "stretching in one direction" is too vague.

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,331 @@
...
+        backward = (x! = plan.backward.x!, z! = plan.backward.z!)
+    elseif irreg_dim == 3
+        forward  = (x! = plan.forward.x!,  y! = plan.forward.y!)
+        backward = (x! = plan.backward.x!, y! = plan.backward.y!)
+    end
+
+    plan = (; forward, backward)
+
+    # Storage space for right hand side of Poisson equation
+    source_term = on_architecture(arch, zeros(complex(eltype(grid)), size(grid)...))
```
glwagner, line 231:
```suggestion
    T = complex(eltype(grid))
    source_term = zeros(T, size(grid)...)
    source_term = on_architecture(arch, source_term)
```

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
simone-silvestri: Should be resolved

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -196,12 +197,13 @@ function validate_global_grid(global_grid::RectilinearGrid)
...
-               TZ Periodic requires also TY and TX to be Periodic, while TY Periodic requires also TX to be Periodic.
-               Please rotate the domain to obtain the required topology")
+               A Periodic z-direction requires also the y- and and x-directions to be Periodic, while a Periodic y-direction requires also 
+               the x-direction to be Periodic.")
     end
     
     if !(global_grid isa YZRegularRG) && !(global_grid isa XYRegularRG) && !(global_grid isa XZRegularRG) 
-        throw(ArgumentError("Only stretching in one direction is supported with distributed grids at the moment."))
+        throw("The provided grid is stretched in directions $(stretched_dimensions(global_grid)). 
+               Distributed architectures support only RectilinearGrid that have variably-spaced cells in at most one direction.")
```
glwagner, line 206: still confused about this, because isn't this point specific to Poisson solvers for the nonhydrostatic model? There's no problem with grids at all. Why don't we want people to build grids?

glwagner: I myself was confused just reading the code because I don't understand why this code would end up in "validate global grid"... why can't we have global grids with these topologies?!??

simone-silvestri: with regards to the stretching it is the same as for the serial algorithm. With regards to the topologies it has to do with the order of the transforms, you don't want to perform a r2r after a r2c transform. In the end, we can relax this requirement later because it requires a lot of additional code and multiple additional transposes so I think it is better just to not support it for the moment and add support incrementally.

The explanation should be quite clear now

src/DistributedComputations/distributed_fft_based_poisson_solver.jl
```diff
@@ -189,39 +189,41 @@ end
...
+validate_poisson_solver_distributed_grid(global_grid) = 
         throw("Grids other than the RectilinearGrid are not supported in the Distributed NonhydrostaticModels")
 
-function validate_global_grid(global_grid::RectilinearGrid) 
+function validate_poisson_solver_distributed_grid(global_grid::RectilinearGrid) 
     TX, TY, TZ = topology(global_grid)
 
     if (TY == Bounded && TZ == Periodic) || (TX == Bounded && TY == Periodic) || (TX == Bounded && TZ == Periodic)
-        throw("NonhydrostaticModels on Distributed grids do not support topology ($TX, $TY, $TZ) at the moment.
+        throw("A distributed Poisson solver does not support grids with topology ($TX, $TY, $TZ) at the moment.
```
glwagner, line 199:
```suggestion
        throw("Distributed Poisson solvers do not support grids with topology ($TX, $TY, $TZ) at the moment.
```

Much better and clearer overall.

simone-silvestri:
Tests pass; this should be ready for review.

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
...
+    _Any_ configuration decomposed in ``z`` direction is _not_ supported.
+    Furthermore, any ``(x, y)`` decompositions other than the configurations mentioned above are also _not_ supported.
+    
+Algorithm for pencil decompositions
+============================================
+
+For pencil decompositions (useful for three-dimensional problems),
+there are two forward transforms, two backward transforms, one tri-diagonal matrix inversion
+and a variable number of transpositions that require MPI communication, dependent on the 
+strethed direction:
```
glwagner, line 61:
```suggestion
stretched direction:
```

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
...
+
+In our implementation we require `Nz ≥ Ry` and `Nx ≥ Ry` with the additional constraint 
+that `Nz % Ry = 0` and `Ny % Rx = 0`.
+
+X - stretched algorithm
+========================
+
+1. `storage.zfield`, partitioned over ``(x, y)`` is initialized with the `rhs`.
+2. Transform along ``z``.
+3. Transpose + communicate to `storage.yfield` partitioned into `(Rx, Ry)` processes in ``(x, z)``.
```
francispoulin, line 80: Could you state explicitly right after the word transpose as to what fields are being transposed?  I think it would make it easier to read.

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
...
+For pencil decompositions (useful for three-dimensional problems),
+there are two forward transforms, two backward transforms, one tri-diagonal matrix inversion
+and a variable number of transpositions that require MPI communication, dependent on the 
+strethed direction:
+
+- a stretching in the x-direction requires four transpositions
+- a stretching in the y-direction requires six transpositions
+- a stretching in the z-direction requires eight transpositions
+
+!!! warning "Computational cost"
```
glwagner, line 67:
```suggestion
!!! note "Computational cost"
```

a warning is like a picture of a diseased lung on tobacco: it could make sense, if it might kill you. Expensive simulations are not such a case.

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
...
+    Because of the additional transpositions, a stretching in the x-direction
+    is computationally cheaper than a stretching in the y-direction, and the latter
+    is cheaper than a stretching in the z-direction
+
+In our implementation we require `Nz ≥ Ry` and `Nx ≥ Ry` with the additional constraint 
+that `Nz % Ry = 0` and `Ny % Rx = 0`.
+
+X - stretched algorithm
```
glwagner, line 75: Is it `x` or `X`? This whole docstring uses ``x`` but the title is `X`.

glwagner:
It looks like https://github.com/CliMA/Oceananigans.jl/pull/2538/ implemented just one solver, with an optional tridiagonal component, is that right. Why does this PR take a different approach? Wouldn't using a single solver result in less code / duplication of transform logic?

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
...
+                          YZRegularRG
+
+struct DistributedFourierTridiagonalPoissonSolver{G, L, P, B, R, S, β} 
+                          plan :: P              
+                   global_grid :: G
+                    local_grid :: L
+    batched_tridiagonal_solver :: B
+                   source_term :: R
+                       storage :: S
+                        buffer :: β
```
glwagner, line 17: can you align these in the normal way

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
...
+end
+
+architecture(solver::DistributedFourierTridiagonalPoissonSolver) =
+    architecture(solver.global_grid)
+
+"""
+    DistributedFourierTridiagonalPoissonSolver(global_grid, local_grid)
+
+Return an FFT-based solver for the Poisson equation evaluated on a `local_grid`
+with one stretched direction
```
glwagner, line 27: explain what you mean by "one stretched direction"

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
...
+============
+
+1. Pencil decompositions:
+    - `Ny ≥ Rx` and `Ny % Rx = 0`
+    - `Nz ≥ Ry` and `Nz % Ry = 0`
+    - If the ``z`` direction is `Periodic`, also the ``y`` and the ``x`` directions must be `Periodic`
+    - If the ``y`` direction is `Periodic`, also the ``x`` direction must be `Periodic`
+
+2. Slab decomposition:
+    - same as for two-dimensional decompositions with `Rx` (or `Ry`) equal to one
```
glwagner, line 134:
```suggestion
    - Same as for two-dimensional decompositions with `Rx` (or `Ry`) equal to one
```

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
+using CUDA: @allowscalar
+using Oceananigans.Solvers: BatchedTridiagonalSolver
+using Oceananigans.Solvers: stretched_dimensions, stretched_direction
```
glwagner, line 3: why are these in `Solvers`. Is this a grid property?

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
...
+
+"""
+function DistributedFourierTridiagonalPoissonSolver(global_grid, local_grid, planner_flag=FFTW.PATIENT)
+    
+    validate_poisson_solver_distributed_grid(global_grid)
+    validate_poisson_solver_configuration(global_grid, local_grid)
+ 
+    irreg_dim = stretched_dimensions(local_grid)[1]
+
+    topology(global_grid, irreg_dim) != Bounded && error("`DistributedFourierTridiagonalPoissonSolver` can only be used when the stretched direction's topology is `Bounded`.")
```
glwagner, line 144:
```suggestion
    topology(global_grid, irreg_dim) != Bounded &&
        error("`DistributedFourierTridiagonalPoissonSolver` requires that the stretched direction (dimension $irreg_dim) is `Bounded`.")
```

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,333 @@
...
+2. Slab decomposition:
+    - same as for two-dimensional decompositions with `Rx` (or `Ry`) equal to one
+
+"""
+function DistributedFourierTridiagonalPoissonSolver(global_grid, local_grid, planner_flag=FFTW.PATIENT)
+    
+    validate_poisson_solver_distributed_grid(global_grid)
+    validate_poisson_solver_configuration(global_grid, local_grid)
+ 
+    irreg_dim = stretched_dimensions(local_grid)[1]
```
glwagner, line 142: err... irreg? stretched?

```suggestion
    stretched_dim = stretched_dimensions(local_grid)[1]
```

glwagner (commented):
Just a question about implementation:

1. https://github.com/CliMA/Oceananigans.jl/pull/2538 implements just one solver for all cases. Why do we take a different approach here?
2. It might make sense to split the constructor into two parts, so that we can build a tridiagonal solver even when all three directions are regular. That could be useful for testing, for example. Also note that in terms of operation count the tridiagonal solve is cheaper than FFT...

glwagner:
Supercedes https://github.com/CliMA/Oceananigans.jl/pull/2538

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
simone-silvestri: changed to `x`

simone-silvestri:
> Just a question about implementation:
> 
> It looks like #2538 implemented just one solver, with an optional tridiagonal component, is that right. Why does this PR take a different approach? Wouldn't using a single solver result in less code / duplication of transform logic?

Here, the approach follows the serial implementation quite closely, where we have an FFT solver and a FourierTridiagonal solver. Additionally, the fields of the solvers are tailored to their task; for example, there is no tridiagonal solver in the pure FFT solver. You might argue that we just put nothing if we don't need it, but then we have more "ambiguous" fields like eigenvalues that are not needed in the tridiagonal solver and source_term that is not needed in the FFT. I am not convinced that a single solver would lead to less (or cleaner) code. Mostly because the underlying code (constructor and `solve!` functions) is compact and slender enough to justify writing individual functions for different grids (the number of lines in the docstring for the `DistributedFourierTridiagonalPoissonSolver` is comparable to the code related to the solver). I think this improves the interpretability of the algorithm. Maybe an improvement would be writing a unified constructor that would spit out the different solvers.

> 2. It might make sense to split the constructor into two parts, so that we can build a tridiagonal solver even when all three directions are regular. That could be useful for testing, for example. 

I have added this capability by passing the `stretched_direction` kwarg.

> Also note that in terms of operation count the tridiagonal solve is cheaper than FFT...

I think, all things considered, the mixed FFT / tridiagonal solve will have basically the same computational cost as the pure FFT solve only for a stretched x direction. The additional transposes required for a y or stretched z direction will completely dominate the cost of the actual operations.

As an example, this is a slab decomposition with a fairly big grid (512 x 256^2) split on 2 GPUs on Tartarus
311333172-43dba752-a91f-4b33-8ade-5a6ec57c982b
The AlltoAllv is the dominant cost, while the FFT (in between the two transposes) is quite irrelevant. In the near future, I'll perform scaling tests on Perlmutter, which has a much better network, so it might be that (even if I think it's unlikely) the cost is not all communication after all.

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -134,15 +139,20 @@ Restrictions
     - Same as for two-dimensional decompositions with `Rx` (or `Ry`) equal to one
 
 """
-function DistributedFourierTridiagonalPoissonSolver(global_grid, local_grid, planner_flag=FFTW.PATIENT)
+function DistributedFourierTridiagonalPoissonSolver(global_grid, local_grid, planner_flag=FFTW.PATIENT; stretched_direction = nothing)
```
glwagner, line 143:
```suggestion
function DistributedFourierTridiagonalPoissonSolver(global_grid, local_grid, planner_flag=FFTW.PATIENT; tridiagonal_direction = nothing)
```

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -1,5 +1,6 @@
 using CUDA: @allowscalar
-using Oceananigans.Grids: stretched_dimensions, stretched_direction
+using Oceananigans.Grids: tridiagonal_dimensions, stretched_direction
```
glwagner: may not have meant to change this one

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,350 @@
...
+```math
+∇²φ = b
+```
+
+for `Distributed` architectures.
+
+Supported configurations
+========================
+
+In the following, `Nx`, `Ny`, and `Nz` are the number of grid points of the **global** grid,
```
tomchor, line 49:
```suggestion
In the following, `Nx`, `Ny`, and `Nz` are the number of grid points of the **global** grid
```

src/DistributedComputations/distributed_fft_tridiagonal_solver.jl
```diff
@@ -0,0 +1,350 @@
...
+5. Transform along ``x``.
+6. Transpose `storage.xfield` + communicate to `storage.yfield` partitioned into `(Rx, Ry)` processes in ``(x, z)``.
+7. Transpose `storage.yfield` + communicate to `storage.zfield` partitioned into `(Rx, Ry)` processes in ``(x, y)``.
+8. Solve the tri-diagonal linear system in the ``z`` direction.
+
+Steps 7 -> 1 are reversed to obtain the result in physical space stored in `storage.zfield` 
+partitioned over ``(x, y)``.
+
+Algorithm for slab decompositions
+============================================
```
tomchor, line 129:
```suggestion
Algorithm for slab decompositions
=============================
```

src/DistributedComputations/distributed_grids.jl
```diff
@@ -84,7 +84,7 @@ function RectilinearGrid(arch::Distributed,
     xl = Rx == 1 ? x : partition_coordinate(x, nx, arch, 1)
     yl = Ry == 1 ? y : partition_coordinate(y, ny, arch, 2)
     zl = Rz == 1 ? z : partition_coordinate(z, nz, arch, 3)
-
+
```
tomchor, line 87:
```suggestion

```

src/DistributedComputations/partition_assemble.jl
```diff
@@ -43,11 +43,12 @@ function partition_coordinate(c::AbstractVector, n, arch, idx)
     r  = arch.local_index[idx]
 
     start_idx = sum(nl[1:r-1]) + 1 # sum of all previous rank's dimension + 1
-    end_idx   = if r == arch.ranks[idx]
-        length(c) 
-    else
+    end_idx   = if r == ranks(arch)[idx] 
+        length(c)
+    else
```
tomchor, line 48:
```suggestion
    else
```

src/DistributedComputations/partition_assemble.jl
```diff
@@ -43,11 +43,12 @@ function partition_coordinate(c::AbstractVector, n, arch, idx)
     r  = arch.local_index[idx]
 
     start_idx = sum(nl[1:r-1]) + 1 # sum of all previous rank's dimension + 1
-    end_idx   = if r == arch.ranks[idx]
-        length(c) 
-    else
+    end_idx   = if r == ranks(arch)[idx]
```
tomchor, line 46:
```suggestion
    end_idx   = if r == ranks(arch)[idx]
```

glwagner:
> > Just a question about implementation:
> > It looks like #2538 implemented just one solver, with an optional tridiagonal component, is that right. Why does this PR take a different approach? Wouldn't using a single solver result in less code / duplication of transform logic?
> 
> Here, the approach follows the serial implementation quite closely, where we have an FFT solver and a FourierTridiagonal solver. Additionally, the fields of the solvers are tailored to their task; for example, there is no tridiagonal solver in the pure FFT solver. You might argue that we just put nothing if we don't need it, but then we have more "ambiguous" fields like eigenvalues that are not needed in the tridiagonal solver and source_term that is not needed in the FFT. I am not convinced that a single solver would lead to less (or cleaner) code. Mostly because the underlying code (constructor and `solve!` functions) is compact and slender enough to justify writing individual functions for different grids (the number of lines in the docstring for the `DistributedFourierTridiagonalPoissonSolver` is comparable to the code related to the solver). I think this improves the interpretability of the algorithm. Maybe an improvement would be writing a unified constructor that would spit out the different solvers.
> 
> > 2. It might make sense to split the constructor into two parts, so that we can build a tridiagonal solver even when all three directions are regular. That could be useful for testing, for example.
> 
> I have added this capability by passing the `stretched_direction` kwarg.
> 
> > Also note that in terms of operation count the tridiagonal solve is cheaper than FFT...
> 
> I think, all things considered, the mixed FFT / tridiagonal solve will have basically the same computational cost as the pure FFT solve only for a stretched x direction. The additional transposes required for a y or stretched z direction will completely dominate the cost of the actual operations.
> 
> As an example, this is a slab decomposition with a fairly big grid (512 x 256^2) split on 2 GPUs on Tartarus 311333172-43dba752-a91f-4b33-8ade-5a6ec57c982b The AlltoAllv is the dominant cost, while the FFT (in between the two transposes) is quite irrelevant. In the near future, I'll perform scaling tests on Perlmutter, which has a much better network, so it might be that (even if I think it's unlikely) the cost is not all communication after all.

Just to clarify --- the mixed tridiagonal + FFT solver also needs eigenvalues, doesn't it?

glwagner (approved):

simone-silvestri:
> Just to clarify --- the mixed tridiagonal + FFT solver also needs eigenvalues, doesn't it?

yep, it does but they are "embedded" in the diagonal terms of the tridiagonal solver. So to clarify, there is no need for an additional eigenvalue field because these are already included in bathced_tridiagonal_solver

tomchor:
I meant to ask this on the other PR but forgot: how are the scalings for this algorithm? Any recommendations when using GPU partitioning on nonhydrostatic models?

simone-silvestri:
A long time ago, I did some scaling tests of the pure FFT algorithm. These were the results

Screenshot 2024-08-10 at 2 57 20 PM

I will probably redo the scaling test later on after the summer. 
In general, always use slab partitioning if you can because you avoid one transposition. 
This might lead to larger halo to domain ratio, but the fill halo, contrarily to the transpose, is hidden so it should still be better to have a slab partitioning then a pencil one.

tomchor:
Thanks that's super useful info!

PR #3690, opened 2024-08-07
Remove Immersed map from `Adapt`ing the `ImmersedBoundaryGrid`

simone-silvestri:
This PR removes the adapting of the immersed map from the immersed boundary grid by shuffling some definitions around and making sure that the active cells map is always passed as an argument instead of being embedded in the grid.

This can possibly help with parameter space issues in complex kernels (see [ClimaOcean, Issue#116](https://github.com/CliMA/ClimaOcean.jl/issues/116#issuecomment-2273878888)). 

This PR is still a draft because I would like to take the opportunity to add some docstring for the immersed map


src/ImmersedBoundaries/active_cells_map.jl
```diff
@@ -26,23 +26,16 @@ A constant representing an immersed boundary grid, where active columns in the Z
...
-
-@inline active_interior_map(::ActiveCellsIBG)            = InteriorMap()
-@inline active_interior_map(::DistributedActiveCellsIBG) = InteriorMap()
+@inline active_interior_map(grid::ArrayActiveCellsIBG,      ::Val{:interior}) = grid.interior_active_cells
+@inline active_interior_map(grid::NamedTupleActiveCellsIBG, ::Val{:interior}) = grid.interior_active_cells.interior
+@inline active_interior_map(grid::NamedTupleActiveCellsIBG, ::Val{:west})     = grid.interior_active_cells.west
+@inline active_interior_map(grid::NamedTupleActiveCellsIBG, ::Val{:east})     = grid.interior_active_cells.east
+@inline active_interior_map(grid::NamedTupleActiveCellsIBG, ::Val{:south})    = grid.interior_active_cells.south
+@inline active_interior_map(grid::NamedTupleActiveCellsIBG, ::Val{:north})    = grid.interior_active_cells.north
+@inline active_interior_map(grid::ActiveZColumnsIBG,        ::Val{:surface})  = grid.active_z_columns
```
glwagner, line 38: I don't understand the purpose of this system of functions or what a "named tuple active cells ibg" means, for example.

Also the function is called "map", but the object is called "cells". Which is it -- a map, or cells?

glwagner:
I think this is a good change. It might be nice to change some of the names of the functions as well, because I'm having a hard time understand the logic of them

src/ImmersedBoundaries/active_cells_map.jl
```diff
@@ -187,12 +189,22 @@ function convert_interior_indices(interior_indices, k, IndicesType)
     return interior_indices
 end
 
-@inline add_3rd_index(t::Tuple, k) = (t[1], t[2], k) 
+@inline add_3rd_index(ij::Tuple, k) = (ij[1], ij[2], k) 
 
+# In case of a serial grid, the interior computations are performed over the whole three-dimensional
+# domain. Therefore, the `interior_active_cells` field contains the indices of all the active cells in 
+# the active domain in the range 1:Nx, 1:Ny and 1:Nz (i.e., we construct the map with parameters :xyz)
```
simone-silvestri, line 196:
```suggestion
# the range 1:Nx, 1:Ny and 1:Nz (i.e., we construct the map with parameters :xyz)
```

src/ImmersedBoundaries/active_cells_map.jl
```diff
@@ -187,12 +189,22 @@ function convert_interior_indices(interior_indices, k, IndicesType)
...
+# the range 1:Nx, 1:Ny and 1:Nz (i.e., we construct the map with parameters :xyz)
 map_interior_active_cells(ibg) = interior_active_indices(ibg; parameters = :xyz)
 
-# In case of a `DistributedGrid` we want to have different maps depending on the 
-# partitioning of the domain
+# In case of a `DistributedGrid` we want to have different maps depending on the partitioning of the domain:
+#
+# If we partition the domain in the x-direction, we typically want to have the option to split three-dimensional 
+# kernels in a `halo-independent` part in the range Hx+1:Nx-Hx, 1:Ny, 1:Nz and two `halo-dependent` computations:
+# a west one spanning 1:Hx, 1:Ny, 1:Nz and a east one spanning Nx-Hx+1:Nx, 1:Ny, 1:Nz.
```
simone-silvestri, line 203:
```suggestion
# a west one spanning 1:Hx, 1:Ny, 1:Nz and an east one spanning Nx-Hx+1:Nx, 1:Ny, 1:Nz. 
```

glwagner (approved):

simone-silvestri:
This PR should be ready, but I would like to add a test before merging it. I will plan to add it after next week

simone-silvestri:
I have added some tests, and everything is passing, including some tests for distributed immersed boundary grids that compare `active_cells_map = true` solutions with `active_cells_map = false` solutions in the serial version so I am quite confident the implementation works.

PR #3694, opened 2024-08-09
Restrict LatitudeLongitueGrid topologies

glwagner:
I believe @simone-silvestri argued for this. I may have been against it at some time. But now I think it is actually useful for documenting the behavior of `LatitudeLongitudeGrid`, so I'm adding it...

cc @navidcy


jagoosw:
I might be missing something but why can't they be periodic in z?

simone-silvestri:
If I have to guess, it is not physical to have something that moves radially outwards and re-enters at a smaller radius.

Since we have a thin shell approximation, we can "technically" allow it, but a periodic domain is meant to represent an infinite domain, so the think shell approximation would not hold with a periodic z-direction

navidcy (approved):

glwagner:
We can allow `Periodic` in z if people want that

navidcy:
@glwagner merge?

PR #3482, opened 2024-02-25
(0.91.7) Open boundary conditions for `NonhydrostaticModel`

jagoosw:
This PR implements the infrastructure for open boundary conditions in the `NonhydrostaticModel` as well as a few simple methods. 

This PR:

- [x] Adds a `matching_scheme` property to `Open` boundary classifications to allow different `fill_X_halo!` to be dispatched
- [x] Introduces `update_boundary_condition!` to be called before halo fills allowing the `mathcing_scheme` to have properties which can evolve with the model
- [x] Make the existing tests pass
- [x] and implements a zero gradient matching scheme

(Others please feel free to update this comment as necessary.)

------------------------------------------------------------------------
Hi all,

Following discussion with @glwagner, @simone-silvestri, and @jm-c this is a first attempt at implementing open boundary conditions. First I will try to get it working for the non-hydrostatic model which seems to be relatively straightforward.

As a first step, I have implemented east/west boundaries which allow a flow to be prescribed or to travel out of the domain (i.e. if you set `OpenBoundaryCondition(nothing)` then my code is assuming the flow will be travelling into that boundary). The outflow condition is equivalent to having a nondimensional phase speed of 1 (sec 3.1 of https://doi.org/10.1016/S0924-7963(97)00023-7) which seems to work fine.

When I vary the inflow velocity I do see waves in the velocity field reflecting from the downstream boundary. I gather that we expect this with any outflowing boundary and would remedy this with a sponge layer, but maybe this is where we need to add something to the pressure solver.

With a constant inflow (0.1 m/s, 0.625m resolution, AMD and SeawaterBuoyancy), and a turbulent outflow, it seems to work okay:

https://github.com/CliMA/Oceananigans.jl/assets/26657828/1509c8fa-ea24-40a2-84d5-33da424567b1

I am going to try and implement the test cases in the paper mentioned below next.

I have a few things I think we should think about:
- Does it make more sense for `Open` to be a topology because otherwise, users have to manually specify boundary conditions on everything? Then we can automatically set `OpenBoundaryCondition(nothing)` unless a user sets something else.
- Should we have an automatic way to setup a sponge layer?
- Do we think it is correct not to add a term to the pressure for time-varying inflow? I think we do not (see below).

<details>
<summary>How I understand the maths:</summary>
Starting from the momentum equation:

$\partial_t\vec u = \vec{G_u} - \nabla P$

We split the time step by defining:

[...]

PR #3695, opened 2024-08-09
Go back to support `BoundaryCondition(Classification, args...)`

simone-silvestri:
It looks like this constructor was indeed removed by mistake in #3482.
This is a change that breaks several other codes / packages. I reintroduce it here.

On the other hand, if we want to keep it out of the codebase, I will switch this PR to a version bump. 
I vote for reintroducing this convenience constructor.


jagoosw:
Sorry I thought I had kept this in!

glwagner (approved):

glwagner:
Can you add a test?

PR #3700, opened 2024-08-13
Leverage Makie extension in quick start

glwagner:
Makes the code a little cleaner. Plus change the aspect ratio to 1.


navidcy (approved):
excellent!

navidcy:
[09dddb3](https://github.com/CliMA/Oceananigans.jl/pull/3700/commits/09dddb33d2609aaf5d0c5a9e1198bf7c59538746) is another place where we'd be way well off with a doctest filter!

glwagner:
yeah that is horrible

PR #3707, opened 2024-08-14
Change log level for doctesting with warnings

### LINKED ISSUE #3705: Convert code blocks documenting `ImmersedBoundaryCondition` to doctests
lgtm!

@glwagner, could you open an issue pointing to this PR and saying that some doctests were removed and we need to attend to them with proper filtering etc?

_Originally posted by @navidcy in https://github.com/CliMA/Oceananigans.jl/pull/3692#pullrequestreview-2236731923_

glwagner:
Closes #3705 

*sigh*

### DIFF
--- a/docs/src/model_setup/boundary_conditions.md
+++ b/docs/src/model_setup/boundary_conditions.md
@@ -436,11 +436,13 @@ top and bottom of both `model.velocities.u` and `model.tracers.c`.
 Immersed boundary conditions are supported experimentally. A no-slip boundary condition is specified
 with
 
-```@meta
-DocTestFilters = r"┌ Warning:[\s\S]*\.jl:[0-9]*"
+```@setup immersed_bc
+using Logging
+logger = ConsoleLogger(2000) # ignore warnings
+global_logger(logger)
 ```
 
-```julia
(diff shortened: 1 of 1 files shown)


navidcy (approved):
lgtm

thanks for dealing with the annoying doctests....

navidcy:
hey you need to use 

```@meta
DocTestSetup = quote
    function foo(x)
        return x^2
    end
end
```

glwagner:
> hey you need to use
> 
> ```
> DocTestSetup = quote
>     function foo(x)
>         return x^2
>     end
> end
> ```

I wanted to avoid activating it globally

glwagner:
Feel free to try to get it to work

navidcy:
OK I'll try later!

navidcy:
but FYI:

```@setup example1

```
 is used to configure `@examples`, not doctests; that's why it didn't work.

On a different note, using `# hide` notation won't work in the sense the code will run but it won't be hidden -- that's syntax from Literate.jl package and we are not liberating the docs; only examples.jl

glwagner:
Didn't work

PR #3713, opened 2024-08-16
Use the word "grid" in the grid tutorial

glwagner:
Little improvement for the grid tutorial.


glwagner:
@navidcy I don't think we should write any examples that don't use the extension (eg we should try to eliminate `interior` from plotting code. If the plotting extension is missing functionality, let's add it first rather than trying to work around it. This will help keep us honest and ensure the extension is useful not just for quick plots but for production (so that it is used in production code, helping to make production code more readable).

navidcy (approved):

PR #3714, opened 2024-08-16
Bugfix in distributed fill halos

simone-silvestri:
The code now uses an incremental tag counter to track how many mpi requests are live. 
This tag is reset when communication is synchronized.

As it is defined right now, the tag is always incremented at the end of a `fill_halo_regions!` on a distributed grid, irrespective of what happened in the `fill_halo_regions!`, with the assumption that all cores participate in the `fill_halo_regions!` so the tags are correctly synchronized. 
https://github.com/CliMA/Oceananigans.jl/blob/315e66bb330b44acc2a0daf74ae357ee66e801d1/src/DistributedComputations/halo_communication.jl#L117-L121

Unfortunately, I experienced a situation where this was not the case. 
In this case, I wanted to do different things on different cores, which is allowed when using the `only_local_halos = true` keyword argument (a very rare occurrence, but a possibility nonetheless).

For example, if we execute this code on the main branch
```julia
arch = Distributed(CPU())
grid = RectilinearGrid(size = (2, 2, 1), extent = (1, 1, 1))
c = Field(grid)

if rank == 0
    fill_halo_regions!(c; only_local_halos = true)
end
```
The mpi_tag will be `1` on rank 0 and `0` on other ranks. This means that in subsequent halo passes, MPI will stall because it cannot match the tag cores communicating with rank 0.

This PR fixes this issue by incrementing the tag _only_ if we have actually launched a mpi send or receive operation, that happens when at least one of the `bcs` is a distributed boundary condition _and_ `only_local_halos == false`

### DIFF
--- a/src/DistributedComputations/halo_communication.jl
+++ b/src/DistributedComputations/halo_communication.jl
@@ -98,25 +98,30 @@ function fill_halo_regions!(field::DistributedField, args...; kwargs...)
                               kwargs...)
 end
 
-function fill_halo_regions!(c::OffsetArray, bcs, indices, loc, grid::DistributedGrid, buffers, args...; fill_boundary_normal_velocities = true, kwargs...)
+function fill_halo_regions!(c::OffsetArray, bcs, indices, loc, grid::DistributedGrid, buffers, args...; only_local_halos = false, fill_boundary_normal_velocities = true, kwargs...)
     if fill_boundary_normal_velocities
         fill_open_boundary_regions!(c, bcs, indices, loc, grid, args...; kwargs...)
     end
     
-    arch             = architecture(grid)
+    arch = architecture(grid)
+    
     fill_halos!, bcs = permute_boundary_conditions(bcs) 
-
-    number_of_tasks = length(fill_halos!)
+    number_of_tasks  = length(fill_halos!)
 
     for task = 1:number_of_tasks
-        fill_halo_event!(c, fill_halos![task], bcs[task], indices, loc, arch, grid, buffers, args...; kwargs...)
+        fill_halo_event!(c, fill_halos![task], bcs[task], indices, loc, arch, grid, buffers, args...; only_local_halos, kwargs...)
     end
 
-    fill_corners!(c, arch.connectivity, indices, loc, arch, grid, buffers, args...; kwargs...)
+    fill_corners!(c, arch.connectivity, indices, loc, arch, grid, buffers, args...; only_local_halos, kwargs...)
     
-    # Switch to the next field to send
-    arch.mpi_tag[] += 1
-
+    # We increment the tag counter only if we have actually initiated the MPI communication.
+    # This is the case only if any of the boundary conditions is a distributed communication 
+    # boundary condition (DCBCT) _and_ the `only_local_halos` keyword argument is false.
+    increment_tag = any(isa.(bcs, DCBCT)) && !only_local_halos
+    if increment_tag 
+        arch.mpi_tag[] += 1
+    end
+    
     return nothing
 end
 
@@ -309,7 +314,8 @@ for side in sides
             send_buffer = $get_side_send_buffer(c, grid, buffers, arch)
             send_tag = $side_send_tag(arch, location)
 
-            @debug "Sending " * $side_str * " halo: local_rank=$local_rank, rank_to_send_to=$rank_to_send_to, send_tag=$send_tag"
+            @show typeof(send_buffer), size(send_buffer), arch.mpi_tag[]
+            @info "Sending " * $side_str * " halo: local_rank=$local_rank, rank_to_send_to=$rank_to_send_to, send_tag=$send_tag"
             
             send_req = MPI.Isend(send_buffer, rank_to_send_to, send_tag, arch.communicator)
 
@@ -336,7 +342,8 @@ for side in sides
             recv_buffer = $get_side_recv_buffer(c, grid, buffers, arch)
             recv_tag = $side_recv_tag(arch, location)
 
-            @debug "Receiving " * $side_str * " halo: local_rank=$local_rank, rank_to_recv_from=$rank_to_recv_from, recv_tag=$recv_tag"
+            @show typeof(recv_buffer), size(recv_buffer), arch.mpi_tag[]
+            @info "Receiving " * $side_str * " halo: local_rank=$local_rank, rank_to_recv_from=$rank_to_recv_from, recv_tag=$recv_tag"
             recv_req = MPI.Irecv!(recv_buffer, rank_to_recv_from, recv_tag, arch.communicator)
 
             return recv_req
--- a/src/DistributedComputations/halo_communication.jl
+++ b/src/DistributedComputations/halo_communication.jl
@@ -104,7 +104,7 @@ function fill_halo_regions!(c::OffsetArray, bcs, indices, loc, grid::Distributed
     end
     
     arch = architecture(grid)
-    
+
     fill_halos!, bcs = permute_boundary_conditions(bcs) 
     number_of_tasks  = length(fill_halos!)
 
@@ -118,6 +118,7 @@ function fill_halo_regions!(c::OffsetArray, bcs, indices, loc, grid::Distributed
     # This is the case only if any of the boundary conditions is a distributed communication 
     # boundary condition (DCBCT) _and_ the `only_local_halos` keyword argument is false.
     increment_tag = any(isa.(bcs, DCBCT)) && !only_local_halos
+    
     if increment_tag 
         arch.mpi_tag[] += 1
     end
--- a/src/DistributedComputations/distributed_fields.jl
+++ b/src/DistributedComputations/distributed_fields.jl
@@ -68,7 +68,7 @@ function synchronize_communication!(field)
         cooperative_waitall!(arch.mpi_requests)
 
         # Reset MPI tag
-        arch.mpi_tag[] -= arch.mpi_tag[]
+        arch.mpi_tag[] = 0
 
         # Reset MPI requests
         empty!(arch.mpi_requests)


glwagner:
Why is a counter the best solution? This seems implicit and indirect. Can we write the code to be more obvious and direct? Where is `mpi_tag` used?

glwagner:
Maybe more specifically I don't understand the motivation for this:

https://github.com/CliMA/Oceananigans.jl/blob/315e66bb330b44acc2a0daf74ae357ee66e801d1/src/DistributedComputations/halo_communication.jl#L44-L48

This is a kind of abstraction for a send/recv event, but its super implicit just consisting of integers, rather than simply recording this information as symbols or strings (is a number the only way to generate a unique ID for a field?). I can't figure out why we would record the Z location, this seems random. Why not record the 3D location directly (not use a digit)? Otherwise this code is seemingly more specific than it needs to be. It would be annoying to debug this too since you'd have to be constantly computing these codes.

simone-silvestri:
Tthe MPI tag must be an integer. The maximum value is vendor dependent but it is quite strict. 
The smaller maximum value is [32767](https://www.intel.com/content/www/us/en/developer/articles/technical/large-mpi-tags-with-the-intel-mpi.html#:~:text=For%20the%20InfiniBand*%20support%20via,be%20queried%20in%20the%20application.)

We could probably record the whole location without incurring in integer dimension issue. We cannot do 
`field_d + loc_x + loc_y + loc_z + from_side + to_side` because the tag could be larger than 32767. 
However, we could combine the three locations in a dictionary
```julia
(Center, Center, Center) -> 0
(Center, Center, Face) -> 1
(Center, Face, Center) -> 2
(Face, Center, Center) -> 3
(Face, Face, Center) -> 4
(Center, Face, Face) -> 5
(Face, Center, Face) -> 6
(Center, Center, Nothing) -> 7
(Face, Center, Nothing) -> 8
(Center, Face, Nothing) -> 9
Everything else: -> 10
```
thus reducing the "location counter" to two indices which would fit.

glwagner:
> Tthe MPI tag must be an integer.

But why does the ID for the process have to be an integer? You can distinguish the problem of creating an abstraction for a process, and the problem of generating an ID from that abstraction. We skip the intermediate step of a human-understandable abstraction here.

glwagner:
> ```julia
> (Center, Center, Center) -> 0
> (Center, Center, Face) -> 1
> (Center, Face, Center) -> 2
> (Face, Center, Center) -> 3
> (Face, Face, Center) -> 4
> (Center, Face, Face) -> 5
> (Face, Center, Face) -> 6
> (Center, Center, Nothing) -> 7
> (Face, Center, Nothing) -> 8
> (Center, Face, Nothing) -> 9
> ...
> ```
> 
> If all the permutations fit into 99 values, we consume only 2 digits which would probably fit within the limits.
> 
> We can also compress the `from_side` and `to_side` into only one digit because we have
> 
> ```julia
> west to east -> 0
> east to west -> 1
> south to north -> 2
> north to south -> 3
> top to bottom -> 4
> bottom to top -> 5
> ```

I think this is fine and could be encapsulated in some function `generate_MPI_tag_from_fill_halo_ID`.

But first we need to create the concept of the "ID" for a fill halo event. The rest is detail.

src/DistributedComputations/halo_communication.jl
```diff
@@ -41,33 +41,40 @@ opposite_side = Dict(
...
-@inline loc_id(LX, LY, LZ) = loc_id(LZ)
+location_counter = 0
+for LX in (:Face, :Center, :Nothing)
+    for LY in (:Face, :Center, :Nothing)
+        for LZ in (:Face, :Center, :Nothing)
+            @eval loc_id(::$LX, ::$LY, ::$LZ) = $location_counter
+            location_counter += 1
+        end
+    end
+end
```
glwagner, line 54:
```suggestion
loc_id(lx, ly, lz) = loc_id(lx) + loc_id(ly) + loc_id(lz)
```

glwagner:
You're gonna want to implement a struct that's someting like

```julia
struct HaloFillingEvent
    location
    z_indices
    side
    field_id
end
```

and then a function

```julia
mpi_tag(hfe::HaloFillingEvent) = # number
```

src/DistributedComputations/halo_communication.jl
glwagner: But why don't you reserve a digit for each location in xyz rather than adding them together? like

```
Nothing, Nothing, Nothing = "000"
Center, Nothing, Nothing = "100"
Center, Nothing, Face = "102"
```

etc

glwagner: Oh I see, the integer has to be small

simone-silvestri:
> You're gonna want to implement a struct that's someting like
> 
> ```julia
> struct HaloFillingEvent
>     location
>     z_indices
>     from_side
>     to_side
>     field_id
> end
> ```
> 
> and then a function
> 
> ```julia
> mpi_tag(hfe::HaloFillingEvent) = # number
> ```

I am not sure about this solution. The tag is used immediately (and only) where created, not recorded, and automatically destroyed by MPI after the communication is complete, so I do not immediately see the immediate utility of extra steps, or to save something in memory. A function that, given architecture, location, and side, spits out a unique tag seems sufficient for interpretability without having to record the output somewhere (it's a bit like a hash function, if you have function and inputs you have everything you need).

glwagner:
> > You're gonna want to implement a struct that's someting like
> > ```julia
> > struct HaloFillingEvent
> >     location
> >     z_indices
> >     from_side
> >     to_side
> >     field_id
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then a function
> > ```julia
> > mpi_tag(hfe::HaloFillingEvent) = # number
> > ```
> 
> I am not sure about this solution. The tag is used immediately (and only) where created, not recorded, and automatically destroyed by MPI after the communication is complete, so I do not immediately see the immediate utility of extra steps, or to save something in memory. A function that, given architecture, location, and side, spits out a unique tag seems sufficient for interpretability without having to record the output somewhere (it's a bit like a hash function, if you have function and inputs you have everything you need).

There are two purposes:

- Make it easier for a future developer to understand the code by using words rather than a digit code
- Make debugging possible
- Replace this:

```
arch.active_requests[] += 1
```

with an actual list of the active requests (rather than simply counting them)

The objective is not to write the minimal code that will work, but to create a system that is human understandable. While a minimal functionality can be debugged and made to work once, it will be very brittle because if it breaks, it could shut down the whole system and prevent future development.

simone-silvestri:
From how I understand it you are suggesting this
```julia
struct HaloFillingEvent
    location
    z_indices
    from_side
    to_side
    field_id
end
```
and then a function
```julia
mpi_tag(hfe::HaloFillingEvent) = # number
```
at the moment, the step is just this
```julia
mpi_side_tag(arch, location) = # number
```
The number is going to be the same, so the interpretability of that tag is not going to improve with the above solution. 
The `mpi_tag` was indeed a misnomer, and I was about to write here that we should change its name because it can create confusion (let me revert it to what it was so we can decide how to call it). That `mpi_tag` is what you called `field_id` or an active counter that keeps track of how many communications are active, which is reset after communication is complete.

glwagner:
If you can demonstrate that this code can be implemented in a way that encourages debugging, inspection at the REPL, etc, without the abstraction, then that can be accepted. I just think that creating the abstraction is not only helpful for future people but will also help _you_ organize your ideas. I would start with that and in the end if you find its not useful, eliminate it. But I wouldn't start by designing code without it and "seeing what happens". The objective here is not simply to "make things work".

glwagner:
> From how I understand it you are suggesting this
> 
> ```julia
> struct HaloFillingEvent
>     location
>     z_indices
>     from_side
>     to_side
>     field_id
> end
> ```
> 
> and then a function
> 
> ```julia
> mpi_tag(hfe::HaloFillingEvent) = # number
> ```
> 
> at the moment, the step is just this
> 
> ```julia
> mpi_side_tag(arch, location) = # number
> ```
> 
> The number is going to be the same, so the interpretability of that tag is not going to improve with the above solution. The `mpi_tag` was indeed a misnomer, and I was about to write here that we should change its name because it can create confusion (let me revert it to what it was so we can decide how to call it). That `mpi_tag` is what you called `field_id` or an active counter that keeps track of how many communications are active, which is reset after communication is complete.

The fact of the matter is the code doesn't make sense right now, so you have not demonstrated / proven that you shouldn't spend a bit more time on the abstractions.

glwagner:
> an active counter that keeps track of how many communications are active

Why would you not simply keep track of the communications themselves, rather than just the _number_? There's no price to keeping a list of the events versus the number of them active so I don't understand why you throw away that info.

src/DistributedComputations/halo_communication.jl
glwagner: Oh, maybe it's somethiing like

```julia
loc_id(lx, ly, lz) = loc_id(lx) + 3 * loc_id(ly) + 6 * loc_id(lz)
```

either way metaprogramming is definitely not needed here...

simone-silvestri:
We could do that, in fact the requests are stored in the `arch.mpi_requests` vector. The problem with using the length of that vector for inferring a `field_id` is that is not unique for different ranks. For example, a slab decomposition on a bounded domain will lead to 2 requests for the rank adjacent to boundaries (RightConnected or LeftConnected) and 4 for ranks connected on both sides. In this way, the counter is incremented globally when one field is sent, the side is inferred from the `fill_halo_side!` routine so this, combined to a location, leads to a unique tag. 

What about changing the `mpi_tag` name to `active_comm_counter` and add an inverse function to extract the inputs from the tag? Then it is quite easy to inspect the tag and understand what communication we are looking at.

glwagner:
> We could do that, in fact the requests are stored in the `arch.mpi_requests` vector. The problem with using the length of that vector for inferring a `field_id` is that is not unique for different ranks. For example, a slab decomposition on a bounded domain will lead to 2 requests for the rank adjacent to boundaries (RightConnected or LeftConnected) and 4 for ranks connected on both sides. In this way, the counter is incremented globally when one field is sent, the side is inferred from the `fill_halo_side!` routine so this, combined to a location, leads to a unique tag.
> 
> What about changing the `mpi_tag` name to `active_comm_counter` and add an inverse function to extract the inputs from the tag? Then it is quite easy to inspect the tag and understand what communication we are looking at.

I don't follow. You have all of that information if you simply store the `HaloFillEvents` in a vector.

src/DistributedComputations/halo_communication.jl
glwagner: Ok well then I think this is equivalent to a conversion from base-3 to base-10 so its

```
loc_id(lx, ly, lz) = loc_id(lx) + 3 * loc_id(ly) + 9 * loc_id(lz)
```

glwagner: Ah right, so the problem is equivalent to converting a number from base-3 to base-10, eg

```julia
loc_id(lx, ly, lz) = loc_id(lx) + 3 * loc_id(ly) + 9 * loc_id(lz)
```

This counts from 0 ("000") up to 26 ("222")

glwagner:
What's the status of this? Is this the cause of what @kburns saw?

glwagner (approved):
I still find the distributed algorithm hard to understand but we should merge this if its a bugfix

simone-silvestri:
> What's the status of this? Is this the cause of what @kburns saw?

This ensures that in fringe cases where only one rank updates the halos of a field like
```julia
if rank == 0
    fill_halo_regions!(f; only_local_halos=true)
end
```
rank 0 does not end up being out of sync with other ranks. This is not something that would happen in a simulation, but it is still a protection against these types of situations.

PR #3716, opened 2024-08-16
Add `child_architecture` methods for `Fields` and `FieldTimeSeries`

simone-silvestri:
In addition, this PR changes the input to `cpu_interpolating_time_indices` from `architecture(fts)` to `child_architecture(fts)` to make sure that the function works also for a distributed `FieldTimeSeries`

### DIFF
--- a/docs/src/grids.md
+++ b/docs/src/grids.md
@@ -257,8 +257,8 @@ true
 To use more than one CPU, we use the `Distributed` architecture,
 
 ```jldoctest grids
-child_architecture = CPU()
-architecture = Distributed(child_architecture)
+device_architecture = CPU()
+architecture = Distributed(device_architecture)
 
 # output
 [ Info: MPI has not been initialized, so we are calling MPI.Init().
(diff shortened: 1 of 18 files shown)


glwagner:
Seeing this I wonder two things:

1. We could always return `CPU` or `GPU` from `architecture` and then define a different function that would return `Distributed` for a distributed simulation. Probably we shouldn't do this because we want `architecture(grid)` to correspond to `grid.architecture`. But its a thought.

2. `child_architecture` seems weird and not super intuitive. Maybe `device_architecture` or something would make more sense?

simone-silvestri:
I agree that maybe it would be more convenient that `architecture` returns the child_architecture.

glwagner:
I don't think "child" is used in standard way here. Usually "child" refers to a processes that is dependent on another process. That's not the relationship between `Distributed` and `CPU`...

https://www.computerhope.com/jargon/p/parechil.htm

glwagner:
This PR is very stale, so I'm closing it.

PR #3717, opened 2024-08-17
Fix `AverageTimeInterval`

liuchihl:
This PR addresses issue  https://github.com/CliMA/Oceananigans.jl/issues/3670, focusing on a problem likely caused by floating-point errors
https://github.com/CliMA/Oceananigans.jl/blob/088f9738877f0aeef7aa4038cf788f69ee102019/src/OutputWriters/windowed_time_average.jl#L91
Similar to https://github.com/CliMA/Oceananigans.jl/blob/a18b81d4c8776987d3cf619521941ba85000a1cb/src/Utils/schedules.jl#L58-L63
Implementing actuation time helps prevent floating-point errors, ensuring the correct execution of statements in
`advance_time_average!` https://github.com/CliMA/Oceananigans.jl/blob/56dbae31bb54c1293e12b6ccc88f77976cc1103f/src/OutputWriters/windowed_time_average.jl#L271-L335.

There is still an issue for post-checkpoint when checkpoint interval is _not_ an integer multiple of the `AveragedTimeInterval`


hdrake:
Some notes while trying to sort out the output writer logic flow.

The key function is `schedule(model)`, which is called every `timestep!`. When `schedule(model) == true`, the `write_output!(writer, sim.model)` function gets triggered:

https://github.com/CliMA/Oceananigans.jl/blob/088f9738877f0aeef7aa4038cf788f69ee102019/src/Simulations/run.jl#L146-L148

The trick for `AveragedTimeInterval` is that, for every timestep in an averaging window, we increment the average:

https://github.com/CliMA/Oceananigans.jl/blob/088f9738877f0aeef7aa4038cf788f69ee102019/src/OutputWriters/windowed_time_average.jl#L220

The difficult part is handling the logic of when to stop collecting increments for a given window:

https://github.com/CliMA/Oceananigans.jl/blob/088f9738877f0aeef7aa4038cf788f69ee102019/src/OutputWriters/windowed_time_average.jl#L270-L280

and to reset the `WindowedTimeAverage.result` back to zero to begin accumulating over the next window:

https://github.com/CliMA/Oceananigans.jl/blob/088f9738877f0aeef7aa4038cf788f69ee102019/src/OutputWriters/windowed_time_average.jl#L255-L268

src/OutputWriters/windowed_time_average.jl
```diff
@@ -83,18 +84,59 @@ JLD2OutputWriter scheduled on TimeInterval(4 days):
...
+        if sch.actuations < typemax(Int)
+            sch.actuations += 1
+        else # re-initialize the schedule to t★
+            initialize!(sch, t★)
+        end
+        return true
+    else
+        return false
+    end
+end
```
hdrake, line 120: I don't think this function can be carried over from `TimeInterval` like this. 

We need this function to return `true` every time the model should be incrementing the time-average. If `window == interval`, that should mean every single timestep! Right now, I think it is only returning `true` at the very end of a window.

That said, we also want the `actuations+=1` to happen once per interval (at the end of the collecting window).

src/OutputWriters/windowed_time_average.jl
```diff
@@ -83,18 +84,59 @@ JLD2OutputWriter scheduled on TimeInterval(4 days):
...
+
+# initialize_schedule!(sch::AveragedTimeInterval, model) = initialize_schedule!(sch, model.clock.time)
+
+
+function next_actuation_time(sch::AveragedTimeInterval)
+    t₀ = sch.first_actuation_time
+    N = sch.actuations
+    T = sch.interval
+    return t₀ + (N + 1) * T
+end
```
hdrake, line 104: This look good, but requires that `sch.actuations` are being incremented (by `sch.actuations +=1`) at the right times. I think that is the problem right now.

src/OutputWriters/windowed_time_average.jl
```diff
@@ -83,18 +84,59 @@ JLD2OutputWriter scheduled on TimeInterval(4 days):
...
+# end
+
 # Schedule actuation
-(sch::AveragedTimeInterval)(model) = sch.collecting || model.clock.time >= sch.previous_interval_stop_time + sch.interval - sch.window
-initialize_schedule!(sch::AveragedTimeInterval, clock) = sch.previous_interval_stop_time = clock.time - rem(clock.time, sch.interval)
-outside_window(sch::AveragedTimeInterval, clock) = clock.time <  sch.previous_interval_stop_time + sch.interval - sch.window   
-end_of_window(sch::AveragedTimeInterval, clock) = clock.time >= sch.previous_interval_stop_time + sch.interval
+# (sch::AveragedTimeInterval)(model) = sch.collecting || model.clock.time >= next_actuation_time(sch) - sch.window
+# initialize_schedule!(sch::AveragedTimeInterval, clock) = sch.previous_interval_stop_time = next_actuation_time(sch) - sch.interval
+initialize_schedule!(sch::AveragedTimeInterval, clock) = next_actuation_time(sch) - sch.interval
```
hdrake, line 131: I'm not 100% sure this is right, since it depends on `next_actuation_time` always being right.

src/OutputWriters/windowed_time_average.jl
hdrake: I think the `sch.actuations += 1` should be removed from here and moved into the appropriate point in `advance_time_average!`

hdrake: I think this should actually be more similar to the previous implementation, something like:

```
(sch::AveragedTimeInterval)(model) = sch.collecting || model.clock.time >= next_actuation_time(sch) - sch.window
```

This comparison operation is basically just asking if we are in the current collecting window (between `next_actuation_time(sch) - sch.window` and `next_actuation_time(sch)`) or in the non-collecting part of the interval outside of the window (between `next_actuation_time(sch) - sch.interval` and `next_actuation_time(sch) - sch.window`).

src/OutputWriters/windowed_time_average.jl
```diff
@@ -83,18 +84,59 @@ JLD2OutputWriter scheduled on TimeInterval(4 days):
...
+outside_window(sch::AveragedTimeInterval, clock) = clock.time <  next_actuation_time(sch) - sch.window
```
hdrake, line 132: Looks good.

src/OutputWriters/windowed_time_average.jl
```diff
@@ -83,18 +84,59 @@ JLD2OutputWriter scheduled on TimeInterval(4 days):
...
 # Schedule actuation
-(sch::AveragedTimeInterval)(model) = sch.collecting || model.clock.time >= sch.previous_interval_stop_time + sch.interval - sch.window
-initialize_schedule!(sch::AveragedTimeInterval, clock) = sch.previous_interval_stop_time = clock.time - rem(clock.time, sch.interval)
-outside_window(sch::AveragedTimeInterval, clock) = clock.time <  sch.previous_interval_stop_time + sch.interval - sch.window   
-end_of_window(sch::AveragedTimeInterval, clock) = clock.time >= sch.previous_interval_stop_time + sch.interval
+# (sch::AveragedTimeInterval)(model) = sch.collecting || model.clock.time >= next_actuation_time(sch) - sch.window
+# initialize_schedule!(sch::AveragedTimeInterval, clock) = sch.previous_interval_stop_time = next_actuation_time(sch) - sch.interval
+initialize_schedule!(sch::AveragedTimeInterval, clock) = next_actuation_time(sch) - sch.interval
+outside_window(sch::AveragedTimeInterval, clock) = clock.time <  next_actuation_time(sch) - sch.window   
+end_of_window(sch::AveragedTimeInterval, clock) = clock.time >= next_actuation_time(sch)
```
hdrake, line 133: Looks good.

src/OutputWriters/windowed_time_average.jl
```diff
@@ -272,16 +315,19 @@ function advance_time_average!(wta::WindowedTimeAverage, model)
         # Note that this may induce data collecting more frequently than proscribed
         # by `wta.schedule.stride`.
         accumulate_result!(wta, model)
```
hdrake, line 317: Add the actuation incrementation here:
```
sch.actuations += 1
```

hdrake (changes requested):

src/OutputWriters/windowed_time_average.jl
```diff
@@ -266,22 +310,25 @@ function advance_time_average!(wta::WindowedTimeAverage, model)
         wta.window_start_time = model.clock.time
         wta.window_start_iteration = model.clock.iteration
         wta.previous_collection_time = model.clock.time
+        # print("elseif end_of_window has been triggered! Setting collecting = false")
```
hdrake, line 313: @liuchihl mentioned that the logic breaks down if the interval is equal to a single timestep, because the logic flow only goes through `elseif` statement and never has to chance to accumulate.

Can we address this corner case by include the following in here:

```
if end_of_window(wta.schedule, model.clock)
  accumulate_result!(wta, model)
```

hdrake (changes requested):

src/OutputWriters/windowed_time_average.jl
```diff
@@ -101,23 +101,27 @@ function next_actuation_time(sch::AveragedTimeInterval)
...
-            initialize!(sch, t★)
-        end
-        return true
-    else
-        return false
-    end
+function actuations(sch::AveragedTimeInterval)
+    N = sch.actuations
+    return N
 end
```
hdrake: What is the purpose of this function? It doesn't seem to be used anywhere.

liuchihl: yeah, there is no use for that, I have already deleted it I believe.

src/OutputWriters/windowed_time_average.jl
```diff
@@ -321,13 +326,12 @@ function advance_time_average!(wta::WindowedTimeAverage, model)
 
         # Reset the "previous" interval time, subtracting a sliver that presents overshoot from accumulating.
         initialize_schedule!(wta.schedule, model.clock)
```
hdrake, line 325: This doesn't do anything anymore, right?

liuchihl: Yes, you are correct!

src/OutputWriters/windowed_time_average.jl
```diff
@@ -266,22 +310,25 @@ function advance_time_average!(wta::WindowedTimeAverage, model)
         wta.window_start_time = model.clock.time
         wta.window_start_iteration = model.clock.iteration
         wta.previous_collection_time = model.clock.time
+        # print("elseif end_of_window has been triggered! Setting collecting = false")
 
     elseif end_of_window(wta.schedule, model.clock)
```
hdrake, line 315: I don't understand why this is an `elseif` statement. 

Relatedly, I don't understand this comment about not accumulating during the initialization:

https://github.com/CliMA/Oceananigans.jl/blob/56dbae31bb54c1293e12b6ccc88f77976cc1103f/src/OutputWriters/windowed_time_average.jl#L293-L295

Shouldn't this just be
```
if collecting
  if end_of_window
    accumulate_result!(wta, model)
    wta.schedule.collecting = false
  elseif mod(model.clock.iteration - wta.window_start_iteration, stride(wta)) == 0
    accumulate_result!(wta, model)
  end
end
```

liuchihl:
Here is the MWE related to this issue:
```julia
# this is a MWE for reproducing AveragedTimeInterval errors
using Oceananigans
using Printf
using Plots

""" Set up a simple simulation to test picking up from a checkpoint. """


function test_simulation(stop_time, Δt, δt, overwrite)
    grid = RectilinearGrid(size=(2), z = (-1,1), topology=(Oceananigans.Flat, Oceananigans.Flat, Oceananigans.Bounded))
    uᵢ(z) = 0
    T=2
    u_forcing(z, t) = 10*sin(2*pi/T*t)

                                  forcing = (u = u_forcing,),
                                  timestepper = :RungeKutta3)

   
    set!(model, u=uᵢ)

    u = model.velocities.u

    simulation.output_writers[:timeavg] = NetCDFOutputWriter(model, (u=u,),
                                        filename = "timeavg.nc",
                                        schedule = AveragedTimeInterval(1δt, window=δt, stride = 1),
                                        # schedule = TimeInterval(δt),
                                 
                                prefix = "test",
                                cleanup = true)

function progress_message(sim) 
    wta = sim.output_writers[:timeavg].outputs["u"]
    wd = wta.window_start_time
    ws = wta.window_start_iteration
    pc = wta.previous_collection_time
    fo = wta.fetch_operand
    res = sum(wta.result)
    mct = model.clock.time
    interval = wta.schedule.interval
    window = wta.schedule.window
    timestep = model.clock.last_stage_Δt
    # pis = wta.schedule.previous_interval_stop_time
    cll = wta.schedule.collecting
    nat = next_actuation_time(wta.schedule)
    fat = wta.schedule.first_actuation_time
    actuation =  wta.schedule.actuations
    @info string("Iter: ", iteration(sim), ", time: ", prettytime(sim), ", model clock time:", mct,
     ", window: ", window, ", u-avg: ", res, ", window_start_time: ", wd, ", window_start_iteration: ", ws, ", previous_collection_time:", pc, ", fetch_operand: ", fo, ", interval: ", interval, ", collecting: ", cll,
     ", next_actuation_time: ", nat, ", first_actuation_time: ", fat, ", actuation: ", actuation, ", timestep: ", timestep)

    # @info string("Iter: ", iteration(sim), ", time: ", prettytime(sim), ", model clock time:", mct,
    #  ", window: ", window, ", u-avg: ", res, ", window_start_time: ", wd, ", window_start_iteration: ", ws, ", previous_collection_time:", pc, ", fetch_operand: ", fo, ", interval: ", interval, ", collecting: ", cll)
end
# progress_message(sim) = @info string("Iter: ", iteration(sim), ", time: ", prettytime(sim))
    simulation.callbacks[:progress] = Callback(progress_message, TimeInterval(δt))

    return simulation
end

function next_actuation_time(sch::AveragedTimeInterval)
    t₀ = sch.first_actuation_time
    N = sch.actuations
    T = sch.interval
    return t₀ + (N + 1) * T
end

if isfile("timeavg.nc")
    rm("timeavg.nc")
    println("File 'timeavg.nc' was successfully removed.")
else
    println("File 'timeavg.nc' does not exist.")
end
# run(`sh -c "rm test_iteration*.jld2"`)

Δt = .01    # timestep (s)
T1 = 4      # first simulation stop time (s)

δt = .01   # progress message interval and output saving interval

simulation = test_simulation(T1, Δt, δt, true)
run!(simulation)

# N = iteration(simulation)
# checkpoint = "test_iteration$N.jld2"
# simulation = test_simulation(T2, Δt, δt, false)
# run!(simulation, pickup=checkpoint)

using NCDatasets
file = "timeavg.nc"
ds = Dataset(file)
t = ds["time"][:]
z = ds["zC"][:]
u = dropdims(ds["u"][:,:,:,:],dims = (1,2))
ln = Plots.plot(t[:],u[end,:],xlabel="time", ylabel="u",legend=nothing)
# Plots.savefig("ln.png")
ln
close(ds)
```

src/OutputWriters/windowed_time_average.jl
liuchihl: Afer testing, separating the elseif statements would create NaNs
```julia
 elseif end_of_window(wta.schedule, model.clock)
        # Output is imminent. Finalize averages and cease data collection.
        # Note that this may induce data collecting more frequently than proscribed
        # by `wta.schedule.stride`.
        accumulate_result!(wta, model)
        
        # Averaging period is complete.
        wta.schedule.collecting = false

        # Reset the "previous" interval time, subtracting a sliver that presents overshoot from accumulating.
        initialize_schedule!(wta.schedule, model.clock)
        wta.schedule.actuations += 1

    elseif mod(model.clock.iteration - wta.window_start_iteration, stride(wta)) == 0
        # Collect data as usual
        accumulate_result!(wta, model)
        # wta.window_start_time = model.clock.time - stride(wta) * model.clock.last_Δt
    end
```

liuchihl:
I think we are close to sorting this issue out. I will make the update once we figure out the remaining checkpoint issue.

src/OutputWriters/windowed_time_average.jl
glwagner: We only need `initialize_schedule!` for this purpose right? Should we simply delete the method?

hdrake: Yes, I wasn't sure if it served some other purpose, but we don't use it in our rewrite.

glwagner: Delete it to see if anything fails, it can be added back if needed. Apply this logic also to other situations.

liuchihl: After testing, nothing has failed after deleting `initialize_schedule!`.

PR #3723, opened 2024-08-22
Bugfix in `interpolate!` for `FieldTimeSeries`

simone-silvestri:
`target_times` is not passed to the correct architecture before interpolation, leading to errors like [this one](https://buildkite.com/clima/climaocean-ci/builds/1251#0191774b-12c6-4f07-9b7f-35119ef4e639).

This happens when `target_times` is a range, then `map` creates a vector which is not isbits.
This PR corrects this bug.


glwagner (approved):
This seems better than trying to convert `map(Time, times)` to `on_architecture` or something

PR #3724, opened 2024-08-22
Make the quickstart a bit more fun

glwagner:
Increases the resolution of the GPU example and adds a plot.


docs/src/quick_start.md
```diff
@@ -47,20 +51,42 @@ heatmap(ζ, axis=(; aspect=1))
 
 Fine, we'll re-run this code on the GPU then:
```
navidcy, line 52: seems you added a tracer? perhaps mention that?

```suggestion
Fine, we'll re-run this code on the GPU and add a tracer initially concentrated at the center of the domain.
```

navidcy (approved):

docs/src/quick_start.md
navidcy: Oh I see you added these new details at the bottom. I think if reads more natural if they are spelled out at first before people see the code. Otherwise they might think (initially) "oh it's more involved to run on GPU"

navidcy (approved):

docs/src/quick_start.md
glwagner: hmm yes, we wouldn't want them to think that!

glwagner: i'll push an update once the docs build and we can see that things worked (this is the first time we use `@example` with GPU on tartarus I think)

navidcy:
We should disable the previews btw...

The docs seem to hang?

navidcy:
Btw, the docs seem to hang?

glwagner:
Looks like you fixed it! @navidcy merge when ready. Yes we should disable previews. Also I'm glad you didn't label the axes because that would ruin the joke:

image

PR #3741, opened 2024-08-27
Remove tendency computation at model construction

simone-silvestri:
At the moment, the tendencies were updated at each call of `update_state!`. This PR makes it so that the tendencies are computed only inside the time-stepping where we need them.

see [ClimaOcean#164](https://github.com/CliMA/ClimaOcean.jl/issues/164)

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/single_column_model_mode.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/single_column_model_mode.jl
@@ -61,7 +61,7 @@ compute_free_surface_tendency!(::SingleColumnGrid, ::SplitExplicitFreeSurfaceHFS
 
 # Fast state update and halo filling
 
-function update_state!(model::HydrostaticFreeSurfaceModel, grid::SingleColumnGrid, callbacks)
+function update_state!(model::HydrostaticFreeSurfaceModel, grid::SingleColumnGrid, callbacks; compute_tendencies = true)
 
     fill_halo_regions!(prognostic_fields(model), model.clock, fields(model))
 
@@ -78,6 +78,9 @@ function update_state!(model::HydrostaticFreeSurfaceModel, grid::SingleColumnGri
     end
 
     update_biogeochemical_state!(model.biogeochemistry, model)
+    
+    compute_tendencies && 
+        @apply_regionally compute_tendencies!(model, callbacks)
 
     return nothing
 end
--- a/src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
@@ -25,10 +25,10 @@ Update peripheral aspects of the model (auxiliary fields, halo regions, diffusiv
 hydrostatic pressure) to the current model state. If `callbacks` are provided (in an array),
 they are called in the end.
 """
-update_state!(model::HydrostaticFreeSurfaceModel, callbacks=[]; compute_tendencies = true) =
+update_state!(model::HydrostaticFreeSurfaceModel, callbacks=[]; compute_tendencies = false) =
     update_state!(model, model.grid, callbacks; compute_tendencies)
 
-function update_state!(model::HydrostaticFreeSurfaceModel, grid, callbacks; compute_tendencies = true)
+function update_state!(model::HydrostaticFreeSurfaceModel, grid, callbacks; compute_tendencies = false)
     @apply_regionally mask_immersed_model_fields!(model, grid)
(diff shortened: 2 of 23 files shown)


src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
```diff
@@ -25,10 +25,10 @@ Update peripheral aspects of the model (auxiliary fields, halo regions, diffusiv
 hydrostatic pressure) to the current model state. If `callbacks` are provided (in an array),
 they are called in the end.
 """
-update_state!(model::HydrostaticFreeSurfaceModel, callbacks=[]; compute_tendencies = true) =
+update_state!(model::HydrostaticFreeSurfaceModel, callbacks=[]; compute_tendencies = false) =
```
glwagner, line 28:
```suggestion
update_state!(model::HydrostaticFreeSurfaceModel, callbacks=[]; compute_tendencies = true) =
```

I prefer the default to be "safe"

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
```diff
@@ -25,10 +25,10 @@ Update peripheral aspects of the model (auxiliary fields, halo regions, diffusiv
...
     update_state!(model, model.grid, callbacks; compute_tendencies)
 
-function update_state!(model::HydrostaticFreeSurfaceModel, grid, callbacks; compute_tendencies = true)
+function update_state!(model::HydrostaticFreeSurfaceModel, grid, callbacks; compute_tendencies = false)
```
glwagner, line 31:
```suggestion
function update_state!(model::HydrostaticFreeSurfaceModel, grid, callbacks; compute_tendencies = true)
```

src/Models/NonhydrostaticModels/update_nonhydrostatic_model_state.jl
```diff
@@ -17,7 +17,7 @@ Update peripheral aspects of the model (halo regions, diffusivities, hydrostatic
 pressure) to the current model state. If `callbacks` are provided (in an array),
 they are called in the end.
 """
-function update_state!(model::NonhydrostaticModel, callbacks=[]; compute_tendencies = true)
+function update_state!(model::NonhydrostaticModel, callbacks=[]; compute_tendencies = false)
```
glwagner, line 20:
```suggestion
function update_state!(model::NonhydrostaticModel, callbacks=[]; compute_tendencies = true)
```

src/Models/ShallowWaterModels/update_shallow_water_state.jl
```diff
@@ -18,7 +18,7 @@ Next, `callbacks` are executed.
 
 Finally, tendencies are computed if `compute_tendencies=true`.
 """
-function update_state!(model::ShallowWaterModel, callbacks=[]; compute_tendencies=true)
+function update_state!(model::ShallowWaterModel, callbacks=[]; compute_tendencies=false)
```
glwagner, line 21:
```suggestion
function update_state!(model::ShallowWaterModel, callbacks=[]; compute_tendencies=true)
```

glwagner:
The logic should be reversed, we should have to include "compute_tendencies=false" as an _optimization_. Rather than what's implemented here, which does the unsafe thing by default

src/Models/HydrostaticFreeSurfaceModels/single_column_model_mode.jl
```diff
@@ -78,6 +78,9 @@ function update_state!(model::HydrostaticFreeSurfaceModel, grid::SingleColumnGri
     end
 
     update_biogeochemical_state!(model.biogeochemistry, model)
+    
+    compute_tendencies && 
+        @apply_regionally compute_tendencies!(model, callbacks)
```
simone-silvestri, line 83: I just noticed that the `SingleColumnGrid` was never computing the tendencies. Is this a bug? Was this mode working? I can add a test for it here

simone-silvestri: Found out the issue, because of this line
https://github.com/CliMA/Oceananigans.jl/blob/5a3d3c92f4e737a199e41be2c95da483f266a9f9/src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl#L28-L30

the `update_state!` in this file was neglected because of the lack of the `compute_tendencies` keyword argument.

simone-silvestri:
This should be ready to merge

glwagner (approved):

PR #3744, opened 2024-08-28
Bug in the vector rotation operator

simone-silvestri:
The meridional velocity had the wrong sign in the vector rotation operator for switching from extrinsic to intrinsic coordinates in an `OrthogonalSphericalShellGrid`.
The test was also poorly designed (my bad) and conspired to hide the bug. This PR should fix the rotation and update the test to make sure everything is correct.

### DIFF
--- a/test/test_cubed_spheres.jl
+++ b/test/test_cubed_spheres.jl
@@ -27,7 +27,7 @@ function kinetic_energy(u, v)
     return compute!(ke)
 end
 
-function test_vector_rotation(grid, T=Float64)
+function test_vector_rotation(grid)
     u = XFaceField(grid)
     v = YFaceField(grid)
     
--- a/test/test_cubed_spheres.jl
+++ b/test/test_cubed_spheres.jl
@@ -61,8 +61,8 @@ function test_vector_rotation(grid, T=Float64)
 
     # Make sure that the flow was converted back to a 
     # purely zonal flow in the extrensic frame (v ≈ 0)
-    @test all(Array(interior(vₑ)) .≈ 0)
-    @test all(Array(interior(uₑ)) .≈ 1)
+    @test all(on_architecture(CPU(), interior(vₑ)) .≈ 0)
+    @test all(on_architecture(CPU(), interior(uₑ)) .≈ 1)
 
     # Purely meridional flow.
     set!(u, 0)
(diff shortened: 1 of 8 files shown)


glwagner:
Can we add a test that uses `OrthogonalSphericalShellGrids`?

test/test_cubed_spheres.jl
```diff
@@ -20,6 +22,116 @@ using Oceananigans.Models.HydrostaticFreeSurfaceModels: VerticalVorticityField
...
+function kinetic_energy(u, v)
+    ke = Field(0.5 * (u * u + v * v))
+    return compute!(ke)
+end
+
+function test_vector_rotation(grid)
+    u = XFaceField(grid)
+    v = YFaceField(grid)
+    
+    # Purely zonal flow.
```
glwagner, line 34: not for all grids right?

glwagner:
Perhaps we don't have this, but can we rotate a `LatitudeLongitudeGrid` by a certain angle, and then test that the vectors are rotated exactly by that angle? This would provide a simple deterministic test. This could work for a very small grid on the equator, for example...

glwagner (approved):
I see that the challenge of writing tests is that we don't have a good way of generating general grids...

test/test_ensemble_hydrostatic_free_surface_models.jl
```diff
@@ -4,6 +4,41 @@ using Oceananigans.Models.HydrostaticFreeSurfaceModels: ColumnEnsembleSize, Slic
...
+    run!(sic_simulation)
+    run!(per_simulation)
+    
+    @info "Testing Single column grid results..."
+    
+    @test all(sic_model.velocities.u.data[i, j, :] .≈ per_model[i, j].velocities.u.data[1, 1, :])
+    @test all(sic_model.velocities.v.data[i, j, :] .≈ per_model[i, j].velocities.v.data[1, 1, :])
+    @test all(sic_model.tracers.c.data[i, j, :]    .≈ per_model[i, j].tracers.c.data[1, 1, :])
+end
+
```
glwagner: was this sent to the wrong PR?

simone-silvestri: yeah this was meant to be in #3741. I have moved it there

test/test_cubed_spheres.jl
simone-silvestri: right, it's in the extrinsic frame. I have changed the comment

PR #3754, opened 2024-08-30
Supporting indexing a `FieldDataset` with a `Symbol`

ali-ramadhan:
100% bikeshedding. I feel like `FieldDataset` should just use `Symbols` but might be nice to support both for now.

### DIFF
--- a/src/OutputReaders/field_dataset.jl
+++ b/src/OutputReaders/field_dataset.jl
@@ -58,3 +58,15 @@ function Base.getproperty(fds::FieldDataset, name::Symbol)
     end
 end
 
+function Base.show(io::IO, fds::FieldDataset)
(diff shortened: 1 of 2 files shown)


glwagner:
Why not a `getproperty` while you're at it :-D

glwagner (approved):

navidcy (approved):

ali-ramadhan:
We now have `getproperty` for `FieldDataset`! I also improved the `Base.show` function so it summarizes what fields are in the dataset.

ali-ramadhan:
Hmmm, the `@test ds isa FieldDataset` test is passing but all the other tests are failing with `UndefVarError: `ds` not defined` even though they come right after and within the scope of the same `@testset`...

Reproduces locally too. Will see if I can fix it locally.

PR #3756, opened 2024-09-03
Single column grid update_state! is missing the `update_model_field_time_series!`

simone-silvestri:
required to force the model with a field time series.
This PR adds `update_model_field_time_series!`  to the `update_state!` for single column grids and a test


simone-silvestri:
Bonus, also the `mask_immersed_field` was missing. I don't expect it will be ever used (why using an immersed boundary grid in a single column?), however I have added a test for it. We can remove it if not needed.

glwagner:
> Bonus, also the `mask_immersed_field` was missing. I don't expect it will be ever used (why using an immersed boundary grid in a single column?), however I have added a test for it. We can remove it if not needed.

Good to have because it might be useful for testing and also, its definitely best if the single column mode is identical (except for performance) to 3D mode.

glwagner (approved):

glwagner (approved):

test/test_ensemble_hydrostatic_free_surface_models.jl
```diff
@@ -37,6 +39,62 @@ const CAVD = ConvectiveAdjustmentVerticalDiffusivity
     @test all(sic_model.velocities.u.data[1, 1, :] .≈ per_model.velocities.u.data[1, 1, :])
     @test all(sic_model.velocities.v.data[1, 1, :] .≈ per_model.velocities.v.data[1, 1, :])
     @test all(sic_model.tracers.c.data[1, 1, :]    .≈ per_model.tracers.c.data[1, 1, :])
+
+    @info "Testing forcing a single column grid model with a FieldTimeSeries..."
```
glwagner, line 43: it looks like this tests both single column and periodic

glwagner: Another way to do this test would be to add a single column grid test alongside wherever `update_model_field_time_series` is already tested. That might be less test code / easier to maintain.

glwagner (approved):

test/test_ensemble_hydrostatic_free_surface_models.jl
```diff
@@ -37,6 +39,62 @@ const CAVD = ConvectiveAdjustmentVerticalDiffusivity
...
+    per_simulation = Simulation(per_model; simulation_kwargs...)
+    
+    run!(sic_simulation)
+    run!(per_simulation)
+    
+    @info "Testing Single column grid results..."
+    
+    @test all(sic_model.velocities.u.data[1, 1, :] .≈ per_model.velocities.u.data[1, 1, :])
+    @test all(sic_model.velocities.v.data[1, 1, :] .≈ per_model.velocities.v.data[1, 1, :])
+    @test all(sic_model.tracers.c.data[1, 1, :]    .≈ per_model.tracers.c.data[1, 1, :])
```
glwagner, line 97:
```suggestion
    @test all(sic_model.velocities.u[1, 1, :] .≈ per_model.velocities.u[1, 1, :])
    @test all(sic_model.velocities.v[1, 1, :] .≈ per_model.velocities.v[1, 1, :])
    @test all(sic_model.tracers.c[1, 1, :]    .≈ per_model.tracers.c[1, 1, :])
```

src/Models/HydrostaticFreeSurfaceModels/single_column_model_mode.jl
```diff
@@ -63,6 +63,12 @@ compute_free_surface_tendency!(::SingleColumnGrid, ::SplitExplicitFreeSurfaceHFS
 
 function update_state!(model::HydrostaticFreeSurfaceModel, grid::SingleColumnGrid, callbacks; compute_tendencies = true)
 
+    # Mask possibly immersed fields?
```
glwagner:
```suggestion
    # Paranoid masking of single column immersed regions
```

test/test_ensemble_hydrostatic_free_surface_models.jl
```diff
@@ -18,14 +20,14 @@ const CAVD = ConvectiveAdjustmentVerticalDiffusivity
 
     Δt = 0.01
 
-    model_kwargs = (; tracers=:c, buoyancy=nothing, closure, coriolis)
+    model_kwargs = (tracers=:c, momentum_advection=nothing, tracer_advection=nothing; closure, coriolis)
```
glwagner:
```suggestion
    model_kwargs = (; tracers=:c, momentum_advection=nothing, tracer_advection=nothing, closure, coriolis)
```

PR #3760, opened 2024-09-04
Enable direct `FieldTimeSeries` forcing

simone-silvestri:
this PR enables using `FieldTimeSeries` for direct forcing as
```julia
u_forcing = FieldTimeSeries{}(...)
forcing = Forcing(u_forcing)
model = NonhydrostaticModel(; grid, forcing = (; u = u_forcing))
```


glwagner:
What's the point of wrapping it in `Forcing`?

I think a cleaner syntax is 

```julia
u_forcing = FieldTimeSeries{}(...)
model = NonhydrostaticModel(; grid, forcing = (; u = u_forcing))
```

simone-silvestri:
yeah sorry, I meant that. I don't know why I slipped that extra `forcing = Forcing(u_forcing)`

src/Forcings/model_forcing.jl
```diff
@@ -27,6 +27,8 @@ regularize_forcing(::Nothing, field::AbstractField, field_name, model_field_name
 
 # TODO: some checking that `array` is validly-sized could be done here
 regularize_forcing(array::AbstractArray, field::AbstractField, field_name, model_field_names) = Forcing(array)
+regularize_forcing(fts::FlavorOfFTS, field::AbstractField, field_name, model_field_names) = Forcing(fts)
```
glwagner, line 30: the same TODO above applies to FTS too

glwagner: not just size but for FTS we can also check location and even the grid

glwagner (approved):

PR #3763, opened 2024-09-05
Bump compat entry for CubedSphere

navidcy:


Sbozzolo (approved):

ali-ramadhan (commented):
Does Oceananigans.jl actually work with CubedSphere.jl v0.1? I'm just assuming stuff has been fixed/updated since then, but it's still in the `[compat]`.

navidcy:
not sure tbh

navidcy:
> Does Oceananigans.jl actually work with CubedSphere.jl v0.1? I'm just assuming stuff has been fixed/updated since then, but it's still in the `[compat]`.

Tried it and `test_multiregion_cubedsphere.jl` was failing with CubedSphere v0.1 so I removed the compat entry! Good catch @ali-ramadhan

simone-silvestri (approved):

ali-ramadhan (approved):
Thanks for checking @navidcy!

Issue #3761, opened 2024-09-05
`LagrangianParticle`s not working with `ImmersedGrid`s


tomchor:
I realize that https://github.com/CliMA/Oceananigans.jl/issues/3609 exists, but I think this is a different issue. The following MWE, which tries to advect `LagrangianParticle`s over an `ImmersedGrid` fails with this error:

```
ERROR: LoadError: MethodError: no method matching cpu__advect_particles!(::KernelAbstractions.CompilerMetadata{…}, ::StructArrays.StructVector{…}, ::Float64, ::ImmersedBoundaryGrid{…}, ::Int64, ::@NamedTuple{…})

Closest candidates are:
  cpu__advect_particles!(::Any, ::Any, ::Any, ::Oceananigans.Grids.AbstractUnderlyingGrid, ::Any, ::Any)
   @ Oceananigans none:0

Stacktrace:
  [1] __thread_run(tid::Int64, len::Int64, rem::Int64, obj::KernelAbstractions.Kernel{…}, ndrange::Nothing, iterspace::KernelAbstractions.NDIteration.NDRange{…}, args::Tuple{…}, dynamic::KernelAbstractions.NDIteration.NoDynamicCheck)
    @ KernelAbstractions ~/.julia/packages/KernelAbstractions/60cqT/src/cpu.jl:144
  [2] __run(obj::KernelAbstractions.Kernel{…}, ndrange::Nothing, iterspace::KernelAbstractions.NDIteration.NDRange{…}, args::Tuple{…}, dynamic::KernelAbstractions.NDIteration.NoDynamicCheck, static_threads::Bool)
    @ KernelAbstractions ~/.julia/packages/KernelAbstractions/60cqT/src/cpu.jl:111
  [3] (::KernelAbstractions.Kernel{…})(::StructArrays.StructVector{…}, ::Vararg{…}; ndrange::Nothing, workgroupsize::Nothing)
    @ KernelAbstractions ~/.julia/packages/KernelAbstractions/60cqT/src/cpu.jl:46
  [4] (::KernelAbstractions.Kernel{…})(::StructArrays.StructVector{…}, ::Vararg{…})
```

This is the MWE:

```julia
using Oceananigans

grid_base = RectilinearGrid(size = (4, 4, 4), extent = (1,1,1))
GFB = GridFittedBottom((x, y) -> -1/2)
grid = ImmersedBoundaryGrid(grid_base, GFB)

n_particles = 3
[...]

PR #3765, opened 2024-09-06
(0.91.14) Extend Lagrangian advection to immersed grids and add tests

tomchor:
Closes https://github.com/CliMA/Oceananigans.jl/issues/3761


src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -48,27 +48,26 @@ bouncing the particle off the immersed boundary with a coefficient or `restituti
...
-            y⁺ = enforce_boundary_conditions(Bounded(), y, yᴸ, yᴿ, Cʳ)
-            z⁺ = enforce_boundary_conditions(Bounded(), z, zᴸ, zᴿ, Cʳ)
-
-        end
+        Cʳ = restitution
+        x⁺ = enforce_boundary_conditions(Bounded(), x, xᴸ, xᴿ, Cʳ)
+        y⁺ = enforce_boundary_conditions(Bounded(), y, yᴸ, yᴿ, Cʳ)
+        z⁺ = enforce_boundary_conditions(Bounded(), z, zᴸ, zᴿ, Cʳ)
+    else
+        x⁺, y⁺, z⁺ = x, y, z
```
tomchor, line 67: This part of the algorithm was failing so I had to change it. However, I'm not familiar with the lagrangian advection algorithm so I'm sure if this is the right way to do things. Can someone double-check this part? From what I can tell, @Jamie-Hilditch and @jagoosw were the main authors of this algorithm, so maybe we should hear from them also.

jagoosw: It might be cleaner todo

```
if immersed_cell
    ...
    return ...
else
    return x, y, z
end
```

But otherwise I agree this looks like it is needed.

simone-silvestri: you can also use an `ifelse` syntax to avoid branching:
```julia
immersed = immersed_cell(i, j, k, ibg)  
x⁺ = ifelse(immersed, enforce_boundary_conditions(Bounded(), x, xᴸ, xᴿ, Cʳ), x)
y⁺ = ifelse(immersed, enforce_boundary_conditions(Bounded(), y, yᴸ, yᴿ, Cʳ), y)
z⁺ = ifelse(immersed, enforce_boundary_conditions(Bounded(), z, zᴸ, zᴿ, Cʳ), z)
```

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -54,14 +54,20 @@ bouncing the particle off the immersed boundary with a coefficient or `restituti
...
+    # Right bounds of the previous cell
+    xᴸ = xnode(i⁻, j,  k,  ibg, f, f, f)
+    yᴸ = ynode(i,  j⁻, k,  ibg, f, f, f)
+    zᴸ = znode(i,  j,  k⁻, ibg, f, f, f)
 
     Cʳ = restitution
-    xb⁺ = enforce_boundary_conditions(Bounded(), x, xᴸ, xᴿ, Cʳ)
-    yb⁺ = enforce_boundary_conditions(Bounded(), y, yᴸ, yᴿ, Cʳ)
-    zb⁺ = enforce_boundary_conditions(Bounded(), z, zᴸ, zᴿ, Cʳ)
+    xb⁺ = enforce_boundary_conditions(topology(ibg)[1] == Flat ? Flat() : Bounded(), x, xᴸ, xᴿ, Cʳ)
```
Jamie-Hilditch, line 68: This might be a little easier to read if you extracted the logic into its own function e.g.
```julia
immersed_boundary_topology(grid_topology) = ifelse(grid_topology == Flat, Flat(), Bounded()
```
and then
```julia
tx, ty, tz = map(immersed_boundary_topology, topology(ibg))
xb⁺ = enforce_boundary_conditions(tx, x, xᴸ, xᴿ, Cʳ)
```

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -38,6 +38,8 @@ Do nothing on Flat dimensions.
 const f = Face()
 const c = Center()
 
+immersed_boundary_topology(grid_topology) = ifelse(grid_topology == Flat, Flat(), Bounded())
+
```
Jamie-Hilditch, line 42: A quick bit of documentation.
```julia
"""
    immersed_boundary_topology(grid_topology)

Unless `Flat`, immersed boundaries are treated as `Bounded` regardless of underlying grid topology.
"""
immersed_boundary_topology(grid_topology) = ifelse(grid_topology == Flat, Flat(), Bounded())
```

Jamie-Hilditch (commented):
Looks good to me!

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
tomchor: Good point! Already added it

Jamie-Hilditch:
> Let me know if there's anything else needed. If there are no other code suggestions this should be ready for merging.

I'm happy but this is one for @simone-silvestri, @jagoosw or someone else with the power to write.

tomchor:
@simone-silvestri @jagoosw @glwagner could please I get a review from someone?

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -48,27 +55,34 @@ bouncing the particle off the immersed boundary with a coefficient or `restituti
...
+    zᴿ = znode(i, j, k⁻ + 1, ibg, f, f, f)
 
-            Cʳ = restitution
-            x⁺ = enforce_boundary_conditions(Bounded(), x, xᴸ, xᴿ, Cʳ)
-            y⁺ = enforce_boundary_conditions(Bounded(), y, yᴸ, yᴿ, Cʳ)
-            z⁺ = enforce_boundary_conditions(Bounded(), z, zᴸ, zᴿ, Cʳ)
+    # Right bounds of the previous cell
+    xᴸ = xnode(i⁻, j,  k,  ibg, f, f, f)
+    yᴸ = ynode(i,  j⁻, k,  ibg, f, f, f)
+    zᴸ = znode(i,  j,  k⁻, ibg, f, f, f)
```
simone-silvestri, line 72: This is quite different from how it was written before. Was the previous implementation  bug?

tomchor: Yes, there was a bug with the previous implementation and it needed re-writing. Specifically, it was failing for `Flat` dimensions iirc. I'm not familiar with how this is implemented in curvilinear coords, and I don't remember lagrangian tracking being done in them before. Can we keep the code as it currently is in the PR?

simone-silvestri: well, I understand why `node(i⁻+1, j⁻+1, k⁻+1, ibg, f, f, f)` would fail, but `node(i⁻, j, k, ibg, f, f, f)` should exist also for `Flat` topologies. Can we do this:
```julia
xᴸ, yᴸ, zᴸ = node(i⁻, j⁻, k⁻, ibg, f, f, f)
```
and keep the right boundaries as written in this PR?

test/test_lagrangian_particle_tracking.jl
```diff
@@ -269,6 +269,19 @@ lagrangian_particle_test_grid(arch, ::Periodic, z) =
 lagrangian_particle_test_grid(arch, ::Flat, z) =
     RectilinearGrid(arch; topology=(Periodic, Flat, Bounded), size=(5, 5), x=(-1, 1), z)
 
+lagrangian_particle_test_grid_expanded(arch, ::Periodic, z) =
+    RectilinearGrid(arch; topology=(Periodic, Periodic, Bounded), size=(5, 5, 5), x=(-1, 1), y=(-1, 1), z = 2 .*z)
+lagrangian_particle_test_grid_expanded(arch, ::Flat, z) =
+    RectilinearGrid(arch; topology=(Periodic, Flat, Bounded), size=(5, 5), x=(-1, 1), z = 2 .*z)
```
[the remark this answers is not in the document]
tomchor: I'm trying to add this but bumping against this error:

```julia
julia> grid = LatitudeLongitudeGrid(topology=(Bounded, Flat, Bounded), size = (5, 5), longitude=(-1, 1), z=(-1, 0))
5×1×5 LatitudeLongitudeGrid{Float64, Bounded, Flat, Bounded} on CPU with 3×0×3 halo and without precomputed metrics
├── longitude: Bounded  λ ∈ [-1.0, 1.0] regularly spaced with Δλ=0.4
├── latitude:  Flat φ                   
└── z:         Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=0.2

julia> NonhydrostaticModel(; grid)
ERROR: None of the implemented pressure solvers for NonhydrostaticModel currently support more than one stretched direction.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] PressureSolver(arch::CPU, grid::LatitudeLongitudeGrid{Float64, Bounded, Flat, Bounded, Nothing, Nothing, Float64, Float64, Float64, OffsetVector{…}, Nothing, OffsetVector{…}, CPU})
   @ Oceananigans.Models.NonhydrostaticModels ~/repos/Oceananigans.jl2/src/Models/NonhydrostaticModels/NonhydrostaticModels.jl:61
 [3] NonhydrostaticModel(; grid::LatitudeLongitudeGrid{…}, clock::Clock{…}, advection::Centered{…}, buoyancy::Nothing, coriolis::Nothing, stokes_drift::Nothing, forcing::@NamedTuple{}, closure::Nothing, boundary_conditions::@NamedTuple{}, tracers::Tuple{}, timestepper::Symbol, background_fields::@NamedTuple{}, particles::Nothing, biogeochemistry::Nothing, velocities::Nothing, hydrostatic_pressure_anomaly::Oceananigans.Models.NonhydrostaticModels.DefaultHydrostaticPressureAnomaly, nonhydrostatic_pressure::Field{…}, diffusivity_fields::Nothing, pressure_solver::Nothing, immersed_boundary::Nothing, auxiliary_fields::@NamedTuple{})
```

Looking at the methods for `PressureSolver()`, I understand why this error is being triggered. But the pressure should actually be able to deal with this case, no?

Just trying to avoid setting up a separate test using the hydrostatic model.

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
tomchor: The previous code fails not because of the `+1` index in the flat dimension, but because it doesn't return anything for the flat dimension:

```
julia> grid_base = RectilinearGrid(topology = (Periodic, Flat, Periodic), size=(5, 5), extent=(1, 1))
5×1×5 RectilinearGrid{Float64, Periodic, Flat, Periodic} on CPU with 3×0×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.2
├── Flat y                   
└── Periodic z ∈ [-1.0, 0.0) regularly spaced with Δz=0.2

julia> node(2, 2, 2, grid_base, Center(), Center(), Center())
(0.3, -0.7)
```

So in these cases the statement `xᴸ, yᴸ, zᴸ = node(i⁻, j⁻, k⁻, ibg, f, f, f)` fails with

```
  BoundsError: attempt to access Tuple{Float64, Float64} at index [3]
  Stacktrace:
    [1] indexed_iterate
      @ ./tuple.jl:92 [inlined]
    [2] bounce_immersed_particle
      @ ~/repos/Oceananigans.jl2/src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl:75 [inlined]
```

tomchor: So, if I understand your comment correctly, we can just replace

```julia
    xᴿ = xnode(i⁻ + 1, j, k, ibg, f, f, f)
    yᴿ = ynode(i, j⁻ + 1, k, ibg, f, f, f)
    zᴿ = znode(i, j, k⁻ + 1, ibg, f, f, f)
```

with 

```julia
    xᴿ = xnode(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
    yᴿ = ynode(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
    zᴿ = znode(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
```

I think that should be equivalent. Note that `node()` is just a shorthand for `(xnode, ynode, znode)`

https://github.com/CliMA/Oceananigans.jl/blob/0f52a6a00db0cca06785df7ac5658638dd85d5a9/src/Grids/nodes_and_spacings.jl#L35-L37

simone-silvestri: right, I am not completely sure this is needed for the `xᴿ`, `yᴿ` and `zᴿ` coordinates, but I think that the left ones need to be computed at the previous_particle_indices which is `i⁻, j⁻, k⁻`.

test/test_lagrangian_particle_tracking.jl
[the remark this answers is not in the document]
simone-silvestri: Ok, that makes sense; we do not have a pressure solver for curvilinear grids yet. Once the one for immersed boundaries is fully functional, it will work also on curvilinear grids (albeit with no preconditioner). Why don't you use the hydrostatic model with prescribed velocities for the test? It is quite simpler and faster because of this:

```julia

# If the model only tracks particles... do nothing but that!!!
const OnlyParticleTrackingModel = HydrostaticFreeSurfaceModel{TS, E, A, S, G, T, V, B, R, F, P, U, C} where
                 {TS, E, A, S, G, T, V, B, R, F, P<:AbstractLagrangianParticles, U<:PrescribedVelocityFields, C<:NamedTuple{(), Tuple{}}}

function time_step!(model::OnlyParticleTrackingModel, Δt; callbacks = [], kwargs...)
    tick!(model.clock, Δt)
    model.clock.last_Δt = Δt
    step_lagrangian_particles!(model, Δt)
    update_state!(model, callbacks)
end

update_state!(model::OnlyParticleTrackingModel, callbacks) =
```

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
tomchor: Yeah I see that. Check the current version of the code. I just made a commit that should have the correct indices. Thanks for catching that!

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -48,27 +55,34 @@ bouncing the particle off the immersed boundary with a coefficient or `restituti
...
+    i⁻, j⁻, k⁻ = previous_particle_indices
 
-        if !immersed_cell(i⁻, j⁻, k⁻, ibg)
-            # Left-right bounds of the previous, non-immersed cell
-            xᴿ, yᴿ, zᴿ = node(i⁻+1, j⁻+1, k⁻+1, ibg, f, f, f)
-            xᴸ, yᴸ, zᴸ = node(i⁻,   j⁻,   k⁻,   ibg, f, f, f)
+    # Left bounds of the previous cell
+    xᴿ = xnode(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
+    yᴿ = ynode(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
+    zᴿ = znode(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
```
simone-silvestri, line 67:
```suggestion
    xᴿ = xnode(i⁻ + 1, j⁻, k⁻, ibg, f, f, f)
    yᴿ = ynode(i⁻, j⁻ + 1, k⁻, ibg, f, f, f)
    zᴿ = znode(i⁻, j⁻, k⁻ + 1, ibg, f, f, f)
```
sorry if I am going back and forth, but I am now convinced this is the correct right bounds formulation, the left bounds look good

simone-silvestri (approved):

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
tomchor: Are you sure about this? The original algorithm had this: 

```julia
xᴿ, yᴿ, zᴿ = node(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
```

The `node()` call is equivalent to three independent calls to `xnode`, `ynode` and `znode` with the same arguments:

https://github.com/CliMA/Oceananigans.jl/blob/0f52a6a00db0cca06785df7ac5658638dd85d5a9/src/Grids/nodes_and_spacings.jl#L35-L37

So to me, the original algorithm should translate to:

```julia
xᴿ = xnode(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
yᴿ = ynode(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
zᴿ = znode(i⁻ + 1, j⁻ + 1, k⁻ + 1, ibg, f, f, f)
```

which is what this PR currently has implemented. Am I missing something?

Or maybe you're saying that the original algorithm was wrong to begin with?

test/test_lagrangian_particle_tracking.jl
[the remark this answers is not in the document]
tomchor: I wasn't aware of that, interesting! I'll try it out :)

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
glwagner: if you need either `x` or `λ` depending on the grid type you can use `ξnode` for example.

glwagner: I think it's better to use `ξnode` than `node[1]` which is a hack and not as easy to understand either.

tomchor: To clarify, I think there are at least 2 separate issues being discussed here.

1. The indices to be used when calling the location for the right indices. The code before this PR was `node(i⁻+1, j⁻+1, k⁻+1, ibg, f, f, f)`, which I believe should translate to `ξnode(i⁻+1, j⁻+1, k⁻+1, ibg, f, f, f)` and the same for `η` and `r`, with the same indices (that's how the PR actually is). @simone-silvestri seems to think that the correct call should be (as per a code suggestion below)

```julia
    xᴿ = xnode(i⁻ + 1, j⁻, k⁻, ibg, f, f, f)
    yᴿ = ynode(i⁻, j⁻ + 1, k⁻, ibg, f, f, f)
    zᴿ = znode(i⁻, j⁻, k⁻ + 1, ibg, f, f, f)
```
or perhaps with `ξ`, `η` and `r`, but importantly here is that the indices are different. @simone-silvestri, could you elaborate?

2. @glwagner is suggesting using `ξnode` (`η` and `r`) instead of `xnode` (and `y` and `z`). This is how it was in the code before this PR, but I think we actually want `xnode` and `ynode` since the advection is performed with velocities in `length/time`, no?

test/test_lagrangian_particle_tracking.jl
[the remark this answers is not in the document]
tomchor: Added LatLon tests with HydrostaticModel!

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
glwagner: I don't know what's correct. Seems like there is a missing test... but maybe @simone-silvestri can say. We did use this code on the sphere I thought.

test/test_lagrangian_particle_tracking.jl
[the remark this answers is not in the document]
simone-silvestri: nice

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
simone-silvestri: Ok, I am a bit confused which node should be the correct ``right'' node. I guess if there are tests now and the `i⁻+1, j⁻+1, k⁻+1, ibg, f, f, f` signature works, then the implementation is correct

tomchor: To be fair, tests works regardless of which of the two signatures we're considering here, so whatever the difference is, the tests don't pick up on it. I think I'll keep the algorithm as close to as it was before this PR as possible. This includes using the `i⁻+1, j⁻+1, k⁻+1, ibg, f, f, f` signature. As for the `xnode` versus `ξnode` debate, I actually realized the existing code used `xnode` for the non-immersed part of the algorithm and `ξnode` for the immersed grid. Likely the latter was done in error, so I'll keep everything as `xnode` for now. We can revisit this is the future if we find evidence that this isn't the proper way to do things.

jagoosw (approved):
LGTM! Sorry it took me so long to get to this

PR #3766, opened 2024-09-06
Fix FieldTimeSeries interpolation for single column grids with x, y locations

glwagner:
This is needed for interpolating to "located" single column grids, which we use in `ClimaOcean` for examples.


src/Grids/grid_utils.jl
```diff
@@ -175,7 +175,7 @@ constant grid spacing `Δ`, and interior extent `L`.
 
 # Grid domains
 @inline domain(topo, N, ξ) = CUDA.@allowscalar ξ[1], ξ[N+1]
-@inline domain(::Flat, N, ξ::AbstractArray) = CUDA.@allowscalar ξ[1], ξ[1]
+@inline domain(::Flat, N, ξ::AbstractArray) = ξ[1]
```
navidcy, line 178: wouldn't that fail without the `@allowscalar`?

glwagner: A one-element range is used to store the coordinate when we have a "located" `Flat` direction. Array / CuArray is only used for stretched coordinates. I can think of a use case but I'm inclined to say that we should cross that bridge when we get there...

navidcy: ok!

navidcy:
Is this ready to merge if tests pass?

navidcy:
I'm suggesting to wait to include this PR in the v0.91.13 release

navidcy (approved):

PR #3778, opened 2024-09-14
Replace IterativeSolvers.jl by Krylov.jl

amontoison:


src/Solvers/heptadiagonal_iterative_solver.jl
```diff
@@ -311,12 +310,13 @@ function solve!(x, solver::HeptadiagonalIterativeSolver, b, Δt)
...
     end
     
-    solver.iterative_solver(x, solver.matrix, b, 
-                            statevars = solver.state_vars,
-                            maxiter = solver.maximum_iterations, 
-                            reltol = solver.tolerance, 
-                            Pl = solver.preconditioner, 
-                            verbose = solver.verbose)
+    solver.iterative_solver(solver.state_vars, solver.matrix, b,
+                            itermax = solver.maximum_iterations,
```
amontoison, line 314:
```suggestion
                            itmax = solver.maximum_iterations,
```

src/Solvers/heptadiagonal_iterative_solver.jl
```diff
@@ -311,12 +310,13 @@ function solve!(x, solver::HeptadiagonalIterativeSolver, b, Δt)
...
-                            reltol = solver.tolerance, 
-                            Pl = solver.preconditioner, 
-                            verbose = solver.verbose)
+    solver.iterative_solver(solver.state_vars, solver.matrix, b,
+                            itmax = solver.maximum_iterations,
+                            atol = zero(eltype(b)),
+                            rtol = solver.tolerance,
+                            ldiv = true,
+                            M = solver.preconditioner,
+                            verbose = Int(solver.verbose))
```
amontoison, line 319:
```suggestion
                            verbose = Int(solver.verbose))
                            copyto!(x, solver.x)
```

src/Solvers/heptadiagonal_iterative_solver.jl
```diff
@@ -311,12 +310,14 @@ function solve!(x, solver::HeptadiagonalIterativeSolver, b, Δt)
...
+                            copyto!(x, solver.x)
```
amontoison, line 320:
```suggestion
                            copyto!(x, solver.state_vars.x)
```

amontoison:
The error with the testset `cpu hydrostatic free surface model` is probably because the preconditioner is not symmetric positive definite.

src/Solvers/sparse_preconditioners.jl
```diff
@@ -107,7 +107,7 @@ end
     asymptotic_diagonal_inverse_preconditioner(A::AbstractMatrix; asymptotic_order)
 
 Compute the diagonally-dominant inverse preconditioner is constructed with an asymptotic
-expansion of `A⁻¹` up to the second order. If `I` is the Identity matrix and `D` is the matrix
+expansion of `A⁻¹` up to the second order. With `I` the dentity matrix and `D` is the matrix
```
amontoison, line 110:
```suggestion
expansion of `A⁻¹` up to the second order. With `I` the identity matrix and `D` is the matrix
```

simone-silvestri:
Hello, @amontoison. Nice work. Do you see a performance improvement when switching to this package?
There should be some benchmarks in the `benchmark` folder that we can test (probably we need to update that folder a bit, let me know if you have problems with it)

amontoison:
> Hello, @amontoison. Nice work. Do you see a performance improvement when switching to this package? There should be some benchmarks in the `benchmark` folder that we can test (probably we need to update that folder a bit, let me know if you have problems with it)

Hi @simone-silvestri, `cg` is not too hard to implement, so I don't think we will see a significant performance improvement on CPU.  
However, for other methods like `gmres`, `Krylov.jl` easily outperforms `IterativeSolvers.jl`.
On GPU, though, we might see a difference because I try to dispatch to the BLAS/LAPACK routines of the GPU vendors as much as possible, whereas `IterativeSolvers.jl` relies on some broadcast.

Also, to the best of my knowledge, only `cg` works on (NVIDIA) GPUs for `IterativeSolvers.jl`, while all solvers in `Krylov.jl` work on the GPUs of any vendor.

I'll try to run some benchmarks before the end of the week.

glwagner:
@simone-silvestri can we remove the matrix solver? I don't think we use it anymore and it will help us move forward.

amontoison:
We can remove the dependency on `IterativeSolvers.jl` if we remove `HeptadiagonalIterativeSolver`.
For Krylov.jl, I developed a dedicated solver in #4041.

simone-silvestri:
> @simone-silvestri can we remove the matrix solver? I don't think we use it anymore and it will help us move forward.

I can remove it. I ll open a PR. I can merge  it next week, however, it is still our default implicit solver for non-rectilinear grids, so I suggest a bit of benchmarking before merging the PR

amontoison:
@simone-silvestri I close the PR.
Maybe you can keep the solver but switch to your own implementation of CG or the one of Krylov.jl.
I wanted to reduce the number of dependencies.

simone-silvestri:
Well, we can also try switching the implementation to Krylov over here and seeing the results.

amontoison:
I benchmarked a few times `Krylov.jl` versus the other Julia packages for Krylov methods and it was never slower.
One big mistake that you do in your implementation is that you compute `norm(r)` explicitly while you have this information for free.

amontoison:
-> https://github.com/CliMA/Oceananigans.jl/blob/main/src/Solvers/conjugate_gradient_solver.jl#L194
If you take the square root of this quantity you have your residual norm and don't need to compute anything at each iteration for your stopping conditions.

simone-silvestri:
that's a good idea! Thanks for the suggestion, let's try it out. I think there are still performance problems with our custom pcg solver, that is why I am a bit cautious in removing completely the other implementation. If we do so before fixing our pcg, we will not be able to compare our custom implementation anymore.

glwagner:
@simone-silvestri I think it is a bad infrastructure strategy to keep around lots of methods. Performance at the present is not that important, what matters far more is the direction we want to go with the code.

PR #3780, opened 2024-09-17
Add docs to caltech cluster

### LINKED ISSUE #3779: Out of memory error with Docs tests
For example:
https://buildkite.com/clima/oceananigans/builds/17468#0191fc2c-b421-4cf7-80b5-0429336b1d7f
https://buildkite.com/clima/oceananigans/builds/17473#0191fd88-d8b9-48d5-9c7f-18efc6747ea7
I believe this is because we are launching the docs from many different branches on a relatively small GPU. I think it would be best to move this test on the caltech cluster (I can open a PR to fix this)

simone-silvestri:
tries to fix #3779 

This PR also switches the parallelization of the docs from Distributed to MPI, with a couple of advantages:
- we can use more CPUs to run the docs which should speed up the computation
- we can run distributed Oceananigans example in the docs, both on CPUs and GPUs 

This last new feature probably warrants (in another PR) writing a distributed example


simone-silvestri:
we probably need to add secret tokens and deploy docs ssh keys to the cluster to make this work

glwagner:
I'm skeptical about this because our docs are super computationally intensive.

glwagner:
Tartarus is probably a lot faster than the caltech nodes?

simone-silvestri:
We decided not to do this.

PR #3783, opened 2024-09-19
Remove Manifest.toml

glwagner:
I noticed that some issues with buildkite have to do with needing to re-resolve the Manifest, so this is yet another attempt to fix the race condition in our CI...


glwagner:
@simone-silvestri @jagoosw I think this PR may solve the race condition issue. The essential reason is that the Manifest was being re-resolved after init because downstream tests imported additional packages that were not imported during the initial instantiation.

glwagner:
It may not be necessary to delete the Manifest here but on the other hand, since we are always re-resolving I think we may as well.

glwagner:
> Is it possible to get julia to ignore the manifest (i.e. specify that we want to ignore it in the CI), so that when a user downloads it they can still use the manifest?
> 
> Having asked that I don't actually know what the manifest does that isn't already taken care of by the Project?

The Project compat is supposed to take care of everything. It's just that subtle bugs can creep into packages that might cause things to fail for us (for example something subtle can break type inference, causing everything to fail on the GPU). So using the Manifest is a bit more conservative, because we fix all of the packages that don't _need_ to be upgraded. On the other hand, we are better community members by deleting the Manifest because catching those bugs is important for everyone.

glwagner:
> > Is it possible to get julia to ignore the manifest (i.e. specify that we want to ignore it in the CI), so that when a user downloads it they can still use the manifest?
> > Having asked that I don't actually know what the manifest does that isn't already taken care of by the Project?
> 
> The Project compat is supposed to take care of everything. It's just that subtle bugs can creep into packages that might cause things to fail for us (for example something subtle can break type inference, causing everything to fail on the GPU). So using the Manifest is a bit more conservative, because we fix all of the packages that don't _need_ to be upgraded. On the other hand, we are better community members by deleting the Manifest because catching those bugs is important for everyone.

To be more specific, with Project.toml we specify compat on packages we depend on, but we do not have compat on packages that our dependencies depend on. The Manifest tracks all packages including dependencies of dependencies (and dependencies of those). We've caught issues in the past by comparing the committed Manifest to another Manifest with different packages. Some issues include, a version of MPI_jll that doesn't work on Mac, or some weird bug in ChainRulesCore... things that we don't have explicit compat for...

jagoosw:
Ah I understand, that makes sense!

glwagner:
Well, it looks like this PR does not completely solve the problem, and the jury is out perhaps on whether it helps or not. I did learn a few things. My current hypothesis is that the race condition stems from when the manifest is "re-resolved" during the test. This can have the effect of trying to access whether certain binaries (like `CUDA_Runtime_jll`) are downloaded or not. And this seems to be where the race condition occurs:

https://buildkite.com/clima/oceananigans/builds/17516#01920c32-4e89-4654-a4a8-9b077e51e87c/39-273

glwagner:
Here's a tutorial for automatically retrying buildkite builds: https://buildkite.com/blog/job-retries

glwagner:
Interesting. I noticed that there were some unnecessary test dependencies and removed those. That seems to have solved the problem.

glwagner:
Well, the tests passed finally with no manual intervention so I declare my work complete @simone-silvestri

simone-silvestri (approved):
this is great!

PR #3792, opened 2024-09-25
Optimizes fill_halo_regions_open.jl

jagoosw:
I was doing some profiling on a model with no open boundaries and discovered that this function was causing a big slow down. I guess this is because the compiler isn't managing to work out its just a load of nothing operations but this change appears to make it completely go away.

### DIFF
--- a/src/BoundaryConditions/fill_halo_regions_open.jl
+++ b/src/BoundaryConditions/fill_halo_regions_open.jl
@@ -61,13 +61,13 @@ end
 # no open fills
 @inline get_open_halo_filling(args...) = nothing, nothing
 
-@inline get_open_halo_filling(field, indices, boundary_conditions, ::Tuple{Face, Center, Center}, grid) = 
+@inline get_open_halo_filling(field, indices, boundary_conditions, loc::Tuple{Face, Center, Center}, grid) = 
     _fill_west_and_east_open_halo!, fill_halo_size(field, fill_west_and_east_halo!, indices, boundary_conditions, loc, grid)
 
-@inline get_open_halo_filling(field, indices, boundary_conditions, ::Tuple{Center, Face, Center}, loc) = 
+@inline get_open_halo_filling(field, indices, boundary_conditions, loc::Tuple{Center, Face, Center}, grid) = 
     _fill_south_and_north_open_halo!, fill_halo_size(field, fill_south_and_north_halo!, indices, boundary_conditions, loc, grid)
 
-@inline get_open_halo_filling(field, indices, boundary_conditions, ::Tuple{Center, Center, Face}, loc) = 
+@inline get_open_halo_filling(field, indices, boundary_conditions, loc::Tuple{Center, Center, Face}, grid) = 
     _fill_bottom_and_top_open_halo!, fill_halo_size(field, fill_bottom_and_top_halo!, indices, boundary_conditions, loc, grid)
 
 @kernel function _fill_west_and_east_open_halo!(c, west_bc, east_bc, loc, grid, args) 
--- a/src/BoundaryConditions/fill_halo_regions_open.jl
+++ b/src/BoundaryConditions/fill_halo_regions_open.jl
@@ -17,14 +17,19 @@ function fill_open_boundary_regions!(field, boundary_conditions, indices, loc, g
     open_fill, regular_fill = get_open_halo_filling_functions(loc) 
     fill_size = fill_halo_size(field, regular_fill, indices, boundary_conditions, loc, grid)
 
-    launch!(arch, grid, fill_size, open_fill, field, left_bc, right_bc, loc, grid, args)
+    fill_open_halo_event!(open_fill, field, left_bc, right_bc, fill_size, loc, arch, grid, args) 
 
     return nothing
 end
 
+@inline fill_open_halo_event!(open_fill, field, left_bc, right_bc, fill_size, loc, arch, grid, args) =
+    launch!(arch, grid, fill_size, open_fill, field, left_bc, right_bc, loc, grid, args)
 
-fill_open_boundary_regions!(fields::NTuple, boundary_conditions, indices, loc, grid, args...; kwargs...) = 
-    ntuple(n->fill_open_boundary_regions!(fields[n], boundary_conditions[n], indices, loc[n], grid, args...; kwargs...), Val(length(fields)))
+@inline fill_open_halo_event!(::Nothing, field, left_bc, right_bc, fill_size, loc, arch, grid, args) = nothing
+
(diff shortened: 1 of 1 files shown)


simone-silvestri:
what was the slowdown?

glwagner:
Might be good to profile with `fill_open_boundary_regions!(fields::NTuple, boundary_conditions, indices, loc, grid, args...; kwargs...) = nothing` also to check this

src/BoundaryConditions/fill_halo_regions_open.jl
```diff
@@ -22,8 +22,9 @@ function fill_open_boundary_regions!(field, boundary_conditions, indices, loc, g
     return nothing
 end
 
-fill_open_boundary_regions!(fields::NTuple, boundary_conditions, indices, loc, grid, args...; kwargs...) =
-    [fill_open_boundary_regions!(field, boundary_conditions[n], indices, loc[n], grid, args...; kwargs...) for (n, field) in enumerate(fields)]
+
+fill_open_boundary_regions!(fields::NTuple, boundary_conditions, indices, loc, grid, args...; kwargs...) = 
+    ntuple(n->fill_open_boundary_regions!(fields[n], boundary_conditions[n], indices, loc[n], grid, args...; kwargs...), Val(length(fields)))
```
glwagner, line 27:
```suggestion
function fill_open_boundary_regions!(fields::NTuple, bcs, indices, loc, args...; kw...) 
    N = length(fields)
    return ntuple(n -> fill_open_boundary_regions!(fields[n], bcs[n], indices, loc[n], args...; kw...), Val(N))
end
```

glwagner: If we can't read the code, we've little to show for all our hard work

glwagner:
The profile should be documented here. I'll tentatively approach but the documentation is important

glwagner (approved):

jagoosw:
Ah so I've realised this isn't the fix we needed, and I was just hiding it from myself in the profile because I replaced the function by writing it in the REPL. I made an MWE:

```jula
using Oceananigans

grid = RectilinearGrid(GPU(), topology = (Flat, Flat, Bounded), size = (100, ), extent = (400, ))

model = HydrostaticFreeSurfaceModel(; grid, velocities = PrescribedVelocityFields(), momentum_advection=nothing, buoyancy=nothing, tracers = ntuple(n->Symbol(:T, n), Val(30)))
```
Screenshot 2024-09-26 at 12 00 29
You can see from this profile that `fill_open_boundary_regions!` takes a lot longer than `fill_halo_event!`, even though there are no velocity open boundaries. This is because it is launching a load of zero size kernels where as `fill_halo_event!` just returns nothing instead.

I've fixed this now and get this from the profile instead:
Screenshot 2024-09-26 at 12 02 11

In numbers, the original version benchmarks `time_step!` at around 4.074 ms ± 581.472 μs and the new version 2.438 ms ± 501.642 μs

src/BoundaryConditions/fill_halo_regions_open.jl
```diff
@@ -17,13 +17,19 @@ function fill_open_boundary_regions!(field, boundary_conditions, indices, loc, g
...
-    launch!(arch, grid, fill_size, open_fill, field, left_bc, right_bc, loc, grid, args)
+    fill_open_halo_event!(open_fill, field, left_bc, right_bc, fill_size, loc, arch, grid, args) 
 
     return nothing
 end
 
-fill_open_boundary_regions!(fields::NTuple, boundary_conditions, indices, loc, grid, args...; kwargs...) =
-    [fill_open_boundary_regions!(field, boundary_conditions[n], indices, loc[n], grid, args...; kwargs...) for (n, field) in enumerate(fields)]
+@inline fill_open_halo_event!(open_fill, field, left_bc, right_bc, fill_size, loc, arch, grid, args) =
+    launch!(arch, grid, fill_size, open_fill, field, left_bc, right_bc, loc, grid, args)
```
glwagner, line 26:
```suggestion
@inline fill_open_halo_event!(fill_open_halo, field, left_bc, right_bc, fill_size, loc, arch, grid, args) =
    launch!(arch, grid, fill_size, fill_open_halo, field, left_bc, right_bc, loc, grid, args)
```

src/BoundaryConditions/fill_halo_regions_open.jl
```diff
@@ -17,13 +17,19 @@ function fill_open_boundary_regions!(field, boundary_conditions, indices, loc, g
     open_fill, regular_fill = get_open_halo_filling_functions(loc)
```
glwagner, line 17: This is a little convoluted. Why do you need a function to determine the size? Why can't we just use `loc`?

src/BoundaryConditions/fill_halo_regions_open.jl
```diff
@@ -17,13 +17,19 @@ function fill_open_boundary_regions!(field, boundary_conditions, indices, loc, g
...
-fill_open_boundary_regions!(fields::NTuple, boundary_conditions, indices, loc, grid, args...; kwargs...) =
-    [fill_open_boundary_regions!(field, boundary_conditions[n], indices, loc[n], grid, args...; kwargs...) for (n, field) in enumerate(fields)]
+@inline fill_open_halo_event!(open_fill, field, left_bc, right_bc, fill_size, loc, arch, grid, args) =
+    launch!(arch, grid, fill_size, open_fill, field, left_bc, right_bc, loc, grid, args)
+
+@inline fill_open_halo_event!(::Nothing, field, left_bc, right_bc, fill_size, loc, arch, grid, args) = nothing
+
+@inline fill_open_boundary_regions!(fields::NTuple{N}, boundary_conditions, indices, loc, grid, args...; kwargs...) where N =
+    ntuple(n->fill_open_boundary_regions!(fields[n], boundary_conditions[n], indices, loc[n], grid, args...; kwargs...), 
+           Val(N))
```
glwagner, line 32: more lines

src/BoundaryConditions/fill_halo_regions_open.jl
jagoosw: This comes from normal halo filling and I think its main purpose is for windowed fields:
https://github.com/CliMA/Oceananigans.jl/blob/4dbec269c7fa88a611121a78924589945b535d2d/src/BoundaryConditions/fill_halo_regions.jl#L349-L433

jagoosw:
I'll stream line those two line before I merge

simone-silvestri:
Is this PR close to merging?

tomchor:
cc @ali-ramadhan

simone-silvestri:
I think this has been superseded by #3834. @jagoosw let us know if it is the case or not.

jagoosw:
> I think this has been superseded by #3834. @jagoosw let us know if it is the case or not.

Yeah definitely superseded!

Issue #3543, opened 2024-04-06
Doing better than `overwriting_existing=true` with output?


glwagner:
Both JLD2 and NetCDF output writers currently support a keyword argument `overwrite_existing`, which indicates, if a file is found that has the same name as the currently requested output, whether that file should be deleted or "overwritten". For example,

https://github.com/CliMA/Oceananigans.jl/blob/1775f2ba9cc2e00bf53f1f36d93f72541d868287/examples/two_dimensional_turbulence.jl#L110-L113

Basically, users probably almost always want to set this `true` for convenience (even though the default is `false`). And that's why we do it in the example too. Though it seems like a pointless always-required line, we've nevertheless reasoned that users should be aware that data might be lost... sort of like signing a waiver...

In a recent conversation @Sbozzolo suggested that this is a bit silly and I have to agree it is sort of pointless red tape in some ways, rather like "accepting cookies" every time we have to visit a website. We don't really think about that anymore, we just click as fast as we can so we can move on with out lives...

I think @Sbozzolo might have some idea for how to do this better. More or less I think the gist is that, rather than having a system where an output writer might have to delete a file, we instead create a directory system where new output is always saved in a unique directory. In other words, rather than saving user output at

```julia
filepath = joinpath(dir, filename)
```

we would save output at the path

```julia
filepath = joinpath(dir, unique_simulation_id, filename)
```

The upside of this system is that the output writers are relieved of any potential need to delete data. That onus is passed to the user instead, where the responsibility belongs.

The downside is that we have to generate the directory name `unique_simulation_id`. No matter what we choose, its going to require effort from users to interpret and learn. It also has the major downside of "hiding" information from users: they'll run a script, and then hunt around for the data that was saved. No matter what naming system we choose for `unique_simulation_id`, I think it makes it harder for users to find their data.

Finally, we should note of course that there's no reason why users can't do this themselves in their own scripts. We don't have to make directories for them, they can simply generate ID's themselves and `mkdir`. If the user is sophisticated enough to be running lots of experiments with highly valuable data, they can probably figure out how to create directories...

We're also entering into dangerous territory I think --- trying to manage user's workflows. Workflow management tools are good in general, of course, but I just think its sort of hard to do well and in a general enough way that is uniformly useful to everyone. So by wading into this area, we risk doing a crap job and interfering / hindering at least some people, rather than helping.

[...]

Issue #3750, opened 2024-08-30
Reductions with `FieldTimeSeries` on an `ImmersedBoundaryGrid` are very slow


ali-ramadhan:
This also makes `data_summary` and `Base.show` very slow since showing a `FieldTimeSeries` prints its min, mean, and max. So it's harder to work with `FieldTimeSeries` interactively. Seems fine when not on a `ImmersedBoundaryGrid`.

I'm guessing it's slower because it's masking out the immersed values but I don't know if we expect it to be ~2000x slower than without an immersed boundary. It's those memory allocations...

A quick quality-of-life fix could be to not call `data_summary` when showing a `FieldTimeSeries`.

## MWE

```julia
using Oceananigans

arch = CPU()

L = 1
H = 1

underlying_grid = LatitudeLongitudeGrid(
    arch;
    topology = (Bounded, Bounded, Bounded),
    size = (512, 512, 64),
    latitude = (-L/2, L/2),
    longitude = (-L/2, L/2),
    z = (-H, 0),
    halo = (4, 4, 4)
)

h = L/2
w = L/5 
mount(x, y) = h * exp(-x^2 / 2w^2) * exp(-y^2 / 2w^2)
bottom(x, y) = -H + mount(x, y)

grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom))

model = HydrostaticFreeSurfaceModel(; grid)

simulation = Simulation(model, Δt=1, stop_iteration=1)

simulation.output_writers[:fields] =
    JLD2OutputWriter(
        model,
        model.velocities;
        filename = "test.jld2",
        schedule = IterationInterval(1),
        overwrite_existing = true
    )

run!(simulation)

u = FieldTimeSeries("test.jld2", "u")
u2 = u[2]
```

Reduction over the `FieldTimeSeries`:

```julia
julia> @time minimum(u2)
 20.954897 seconds (118.72 M allocations: 130.792 GiB, 25.74% gc time)
0.0
```

Reduction over the underlying data:

```julia
julia> @time minimum(u2.data)
  0.011304 seconds (3 allocations: 1.562 KiB)
0.0
```

or almost 2000x faster.

## Environment

Oceananigans.jl `main` branch with

```
Julia Version 1.10.5
Commit 6f3fdf7b362 (2024-08-27 14:19 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
[...]

Issue #3144, opened 2023-06-14
Various `FieldTimeSeries` bugs


jagoosw:
I've been using `FieldTimeSeries` a lot over the last few weeks and have noticed a couple of issues calculating means and constructing operations on them.

First, `mean` with `dims` only calculates it on the first time index (I think). For example, if I do `mean(f, dims = (1, 2, 3))` it returns a  size `[1, 1, 1]` array vs calculating `mean(interior(f), dims = (1, 2, 3))` which returns a `[1, 1, 1, Nt]` array. I suspect it is also calculating means with no dimensions specified incorrectly.

Second, constructing operators on field time series fails and returns operators with dimensions `[Nx, Ny, Nz]` which if you then try and index into fails as it tries to index into the underlying fields which have a time dimension. For example, `speed = √(u^2 + v^2 + w^2)` gives `speed` with size `[Nx, Ny, Nz]`, if I then try and index at `[1, 1, 1]` it throws a bounds error trying to access `Nx x Ny x Nz x Nt` array at `1, 1, 1`.

Neither of these is particularly important but thought I would document in case anyone else has issues, and because I will try and fix them at some point.

For reference, I am currently getting around this by calculating means on the interior (which I think would produce the wrong results with immersed boundaries because it wouldn't have the masking step), and by just calculating arrays like `speed = √(interior(u)^2 + interior(v)^2 + interior(w)^2)` which doesn't lose me too much performance since I end up indexing into the whole array anyway.

glwagner:
Thanks for bringing these up, I've also noticed a number of issues. Can you reformat your issue into a bulleted list? Maybe we can add more items.

glwagner:
The implementation of `mean` is naive. Even if it does work for `dims=:` it is slow and should be improved:

https://github.com/CliMA/Oceananigans.jl/blob/19dac0b8f6b621057e250ecf0510b3c4c91915e9/src/OutputReaders/field_time_series.jl#L276-L295

ali-ramadhan:
I know it's documented in the code but also: https://github.com/CliMA/Oceananigans.jl/blob/bf04295910ef02ca3d4105bee801070deef99175/src/OutputReaders/field_time_series_reductions.jl#L17

Would something simple like this work?

```julia
function reduce_4d(spatial_reduction::Function, temporal_reduction::Function, f::Function, fts::FTS4D; dims, kw...)
    reduced_snapshots = [
[...]

Issue #1171, opened 2020-11-12
Reducing boilerplate in OutputWriter specification


glwagner:
Building up OutputWriters can be a major source of boilerplate in "production" scripts that often involve multiple types of output (slices at different locations, different types of averages, etc). An example is:

https://github.com/glwagner/EadyTurbulence/blob/master/initial_value_problem/eady_initial_value_problem.jl

I think the point is more or less that output writers are often related to one another. For example, we typically want all our output in a single directory, and on the same schedule. I also pretty much always use `force=true` (I guess the equivalent for NetCDF is "clobber" mode).

Is there a way to reduce this boilerplate or is it a fact of Oceananigans life? Can we specify some output parameters in `Simulation` rather than in each `OutputWriter` individually? Resolving #1163 might help but I'm not sure. It does seem to make sense to coordinate the output directory through the `Simulation` rather than each output writer individually. But this also contradicts some of what was discussed on #963, such as using the keyword `filepath` rather than `prefix` and `dir`.

ali-ramadhan:
Yeah definitely agree with output writer construction being boilerplate-y...

Taking the Eady turbulence IVP example, it seems like there are two kinds of output writers:
1. Slices (`xy_surface`, `xy_subsurface`, `xy_middepth`, `xy_bottom`, `xz`, `yz`).
2. Sets of computed fields (`horizontal_averages`, `volume_averages`).

# Saving collections of slices

For number 1, I can imagine an abstraction, perhaps a collection of output writers like `NetCDFOutputWriters`, that allows you to specify a number of slice outputs for all fields:

```julia
slices = (all           = FieldSlicer(),
          xy_surface    = FieldSlicer(k=grid.Nz),
          xy_subsurface = FieldSlicer(k=k_subsurface),
          xy_middepth   = FieldSlicer(k=round(Int, grid.Nz/2)),
          xy_bottom     = FieldSlicer(k=1),
          xz            = FieldSlicer(j=1),
          yz            = FieldSlicer(i=1)
)

simulation.output_writers[:slices] = NetCDFOutputWriters(filepath_prefix="eady_turbulence", slices=slices, ...)
```

This would save all fields in `merge(model.velocities, model.tracers)` to `eady_turbulence_all.nc`, `eady_turbulence_xy_surface.nc`, etc. with appropriate slicing. You could pass in one schedule for all slices, or a named tuple of schedules if you want different schedules for different output writers.

[...]

PR #3793, opened 2024-09-26
A new interface for output

glwagner:
I've long been unsatisfied with how we build output. It requires a lot of typing --- that is _boilerplate_. I often feel a sense of dread when I have to go beyond "visualizing the final iteration" of a prototype to defining an output writer. _So much typing_.

This PR is an attempt to make output easier and more fun. I have focused for the moment on JLD2 but if there is some consensus then I think this PR should extend the same to NetCDF.

The main thrust of this PR is a new function called `output!`. It works like this:

```julia
output!(simulation, outputs; schedule=TimeInterval(1), filename="low_hanging_fruit")
```

The function adds an output writer to `simulation`, choosing a "generic name", and defaulting to `JLD2Format()`. Does this enable one line output writing?

I'd love to hear feedback about this design.

There are two more things. First, we need to default `with_halos=true` for JLD2OutputWriter. The time has come because `FieldTimeSeries` is mature. We do, in fact, want halos.

The other conundrum is `overwrite_existing` which is discussed on #3543. In short I am wondering whether the best course of action is simply to make default `overwrite_existing=true` and solve so much boilerplate. Simulations are cheap, but life is short!

PS I also want to change `add_callback!` to just `callback!`.

### DIFF
--- a/src/Simulations/callback.jl
+++ b/src/Simulations/callback.jl
@@ -71,27 +71,8 @@ Callback(wta::WindowedTimeAverage, schedule; kw...) =
     throw(ArgumentError("Schedule must be inferred from WindowedTimeAverage. 
                         Use Callback(windowed_time_average)"))
 
-struct GenericName end
-
-function unique_callback_name(name, existing_names)
-    if name ∈ existing_names
-        return Symbol(:another_, name)
-    else
-        return name
-    end
-end
-
-function unique_callback_name(::GenericName, existing_names)
-    prefix = :callback # yeah, that's generic
-
-    # Find a unique one
-    n = 1
-    while Symbol(prefix, n) ∈ existing_names
-        n += 1
-    end
-
-    return Symbol(prefix, n)
-end
+struct GenericCallbackName end
+unique_name(::GenericCallbackName, existing) = unique_name(:callback, existing)
 
 """
     add_callback!(simulation, callback::Callback; name = GenericName(), callback_kw...)
@@ -109,8 +90,8 @@ already exists.
 
 The `callback` (which contains a schedule) can also be supplied directly.
 """
-function add_callback!(simulation, callback::Callback; name = GenericName())
-    name = unique_callback_name(name, keys(simulation.callbacks))
+function add_callback!(simulation, callback::Callback; name = GenericCallbackName())
+    name = unique_name(GenericCallbackName(), keys(simulation.callbacks))
     simulation.callbacks[name] = callback
     return nothing
 end
--- a/examples/two_dimensional_turbulence.jl
+++ b/examples/two_dimensional_turbulence.jl
@@ -106,11 +106,7 @@ s = sqrt(u^2 + v^2)
 
 # We pass these operations to an output writer below to calculate and output them during the simulation.
 filename = "two_dimensional_turbulence"
-
-simulation.output_writers[:fields] = JLD2OutputWriter(model, (; ω, s),
-                                                      schedule = TimeInterval(0.6),
-                                                      filename = filename * ".jld2",
-                                                      overwrite_existing = true)
+output!(simulation, (; ω, s); schedule=TimeInterval(0.6), filename)
 
 # ## Running the simulation
 #
--- a/src/OutputWriters/jld2_output_writer.jl
+++ b/src/OutputWriters/jld2_output_writer.jl
@@ -28,17 +28,17 @@ ext(::Type{JLD2OutputWriter}) = ".jld2"
 
 """
     JLD2OutputWriter(model, outputs; filename, schedule,
-                              dir = ".",
-                          indices = (:, :, :),
-                       with_halos = false,
-                       array_type = Array{Float64},
-                   file_splitting = NoFileSplitting(),
-               overwrite_existing = false,
-                             init = noinit,
-                        including = [:grid, :coriolis, :buoyancy, :closure],
-                          verbose = false,
-                             part = 1,
-                          jld2_kw = Dict{Symbol, Any}())
+                     dir = ".",
+                     indices = (:, :, :),
+                     with_halos = false,
+                     array_type = Array{Float64},
+                     file_splitting = NoFileSplitting(),
+                     overwrite_existing = false,
+                     init = noinit,
+                     including = [:grid, :coriolis, :buoyancy, :closure],
+                     verbose = false,
+                     part = 1,
+                     jld2_kw = Dict{Symbol, Any}())
 
 Construct a `JLD2OutputWriter` for an Oceananigans `model` that writes `label, output` pairs
 in `outputs` to a JLD2 file.
@@ -163,17 +163,17 @@ JLD2OutputWriter scheduled on TimeInterval(20 minutes):
 ```
 """
 function JLD2OutputWriter(model, outputs; filename, schedule,
-                                   dir = ".",
-                               indices = (:, :, :),
-                            with_halos = false,
-                            array_type = Array{Float64},
-                        file_splitting = NoFileSplitting(),
-                    overwrite_existing = false,
-                                  init = noinit,
-                             including = default_included_properties(model),
-                               verbose = false,
-                                  part = 1,
-                               jld2_kw = Dict{Symbol, Any}())
+                          dir = ".",
+                          indices = (:, :, :),
+                          with_halos = true,
+                          array_type = Array{Float64},
+                          file_splitting = NoFileSplitting(),
+                          overwrite_existing = true,
+                          init = noinit,
+                          including = default_included_properties(model),
+                          verbose = false,
+                          part = 1,
+                          jld2_kw = Dict{Symbol, Any}())
 
     mkpath(dir)
     filename = auto_extension(filename, ".jld2")
@@ -283,7 +283,7 @@ function write_output!(writer::JLD2OutputWriter, model)
         verbose && @info "Writing JLD2 output $(keys(writer.outputs)) to $path..."
 
         start_time, old_filesize = time_ns(), filesize(writer.filepath)
-        jld2output!(writer.filepath, model.clock.iteration, model.clock.time, data, writer.jld2_kw)
+        write_jld2_output!(writer.filepath, model.clock.iteration, model.clock.time, data, writer.jld2_kw)
         end_time, new_filesize = time_ns(), filesize(writer.filepath)
 
         verbose && @info @sprintf("Writing done: time=%s, size=%s, Δsize=%s",
@@ -296,14 +296,14 @@ function write_output!(writer::JLD2OutputWriter, model)
 end
 
 """
-    jld2output!(path, iter, time, data, kwargs)
+    write_jld2_output!(path, iter, time, data, kwargs)
 
 Write the (name, value) pairs in `data`, including the simulation
 `time`, to the JLD2 file at `path` in the `timeseries` group,
 stamping them with `iter` and using `kwargs` when opening
 the JLD2 file.
 """
-function jld2output!(path, iter, time, data, kwargs)
+function write_jld2_output!(path, iter, time, data, kwargs)
     jldopen(path, "r+"; kwargs...) do file
         file["timeseries/t/$iter"] = time
         for name in keys(data)
--- a/src/Simulations/Simulations.jl
+++ b/src/Simulations/Simulations.jl
@@ -6,6 +6,7 @@ export run!
 export Callback, add_callback!
 export iteration
 export stopwatch
+export JLD2_output!
 
 using Oceananigans.Models
 using Oceananigans.Diagnostics
@@ -20,9 +21,23 @@ using OrderedCollections: OrderedDict
 
 import Base: show
 
+function unique_name(prefix::Symbol, existing_names)
+    if !(prefix ∈ existing_names)
+        name = prefix
+    else # make it unique
+        n = 1
+        while Symbol(prefix, n) ∈ existing_names
+            n += 1
+        end
+        name = Symbol(prefix, n)
+    end
+    return name
+end
+
 include("callback.jl")
 include("simulation.jl")
 include("run.jl")
 include("time_step_wizard.jl")
+include("output_helpers.jl")
 
 end # module
--- a/src/Simulations/output_helpers.jl
+++ b/src/Simulations/output_helpers.jl
@@ -1,0 +1,25 @@
+#####
+##### Lend a helping hand
+#####
+
+struct GenericJLD2Name end
+unique_name(::GenericJLD2Name, existing) = unique_name(:jld2, existing)
+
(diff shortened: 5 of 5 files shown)


glwagner:
There's an intriguing side benefit of this "unified" interface for output. It means that it is possible (though we don't have it now) for users to specify an "output preference" in a `Preferences.toml`, which would then determine the default behavior of `output!`. I think it also legitimately makes it easier to switching between formats.

glwagner:
Another abstraction I think would be useful is a utility for building multiple outputs. Imagine this:

```julia
indices = (xy=(:, :, k), xz=(:, 1, :), yz=(1, :, :))
sliced_outputs!(simulation, outputs, indices; schedule=TimeInterval(1), filename="sliced")
```

this would append `filename` with the keys of `indices`, eg we would get 3 outputs titled `"sliced_xy"`, `"sliced_xz"`, and `"sliced_yz"`.

I personally find myself using `indices` the most for this kind of thing but others might have additional input.

ali-ramadhan:
> I'd love to hear feedback about this design.
 
Maybe I'm personally quite picky about specifying outputs and output file names, so I might always end up with verbose boilerplate for output writing (and I'm personally fine with that). But I'd support reducing boilerplate and maybe just a bit of flexibility would work even for picky people! I remember having this conversation in #1171 actually!

One-line flexible output writing would be especially great for examples, new user friendliness, and quick iteration.

Some thoughts:
1. I think the name `output!` is a bit vague in what it does. Does it just output the current state of the simulation? Would `add_output_writers!` be clearer and align more closely with existing Oceananigans nomenclature? For the same reason, I'd suggest keeping `add_callback!` over renaming to `callback!`.
2. I frequently output both JLD2 and NetCDF versions of the exact same data. So ideally subsequent calls to `add_output_writers!` with `JLD2Format()` and `NetCDFFormat()` would add both.
3. Would definitely support extending this to NetCDF as well!

> First, we need to default `with_halos=true` for JLD2OutputWriter. The time has come because `FieldTimeSeries` is mature. We do, in fact, want halos.

Sounds good! Will be nice for derivatives to work out of the box. Although `FieldTimeSeries` could do with a bit more maturing, e.g. issues #3144 and #3750.

> The other conundrum is `overwrite_existing` which is discussed on #3543. In short I am wondering whether the best course of action is simply to make default `overwrite_existing=true` and solve so much boilerplate.

Is the "so much boilerplate" just the extra one line `overwrite_existing = true` for each output writer? Maybe I'm too conservative here but I think the default should be `overwrite_existing = false` just because the cost of overwriting and losing data can be very high.

But I really like the suggestions in #3543 of having the option to save output in unique directories be easily specifiable.

If we want an easy default, then maybe it could do some version of the unique directories? Or maybe have `overwrite_existing = false` (the default) just rename existing files?

I'm not sure of the best approach but as someone who's conservative about overwriting by default I'm tempted to err on the side of caution.

> Another abstraction I think would be useful is a utility for building multiple outputs. Imagine this:
> 
> ```julia
> indices = (xy=(:, :, k), xz=(:, 1, :), yz=(1, :, :))
> sliced_outputs!(simulation, outputs, indices; schedule=TimeInterval(1), filename="sliced")
> ```
> 
> this would append `filename` with the keys of `indices`, eg we would get 3 outputs titled `"sliced_xy"`, `"sliced_xz"`, and `"sliced_yz"`.

Love this idea! Hoping that you can also pass e.g. `(surface=(:, :, k), zonal=(:, 1, :), meridional=(1, :, :))` to get `sliced_surface`, `sliced_zonal`, and `sliced_meridional`.

glwagner:
> Maybe I'm personally quite picky about specifying outputs and output file names, so I might always end up with verbose boilerplate for output writing

Totally and to be clear, when we think about the economy of an interface, we are thinking about prototyping, illustrating, testing, not necessarily "production". I think "production" places fewer demands on the user interface and what we have now is ok for production. This PR mainly improves the small stuff. Also arguably it's more helpful for experienced than new users.

> I think the name output! is a bit vague in what it does. Does it just output the current state of the simulation? Would add_output_writers! be clearer and align more closely with existing Oceananigans nomenclature? For the same reason, I'd suggest keeping add_callback! over renaming to callback!.

I agree that with "add" and "writer" the meaning is cemented. I think it's important to recognize trade-offs though, because there is a limit to the benefit of being explicit (when things become hard to read or understand). I think in this case I accept that `output_writer!` is probably better than `ouput!`. I think prepending `add_` has a more marginal benefit (and is a little ugly) and that context is really what drives understanding of `callback!` / `add_callback!` (eg a schedule, etc). But this is certainly open for discussion.

> Love this idea! Hoping that you can also pass e.g. (surface=(:, :, k), zonal=(:, 1, :), meridional=(1, :, :)) to get sliced_surface, sliced_zonal, and sliced_meridional.

Yes for sure! In that example the keys "xy", "xz", etc would be names appended to the filename prefix. 

> I think the default should be overwrite_existing = false just because the cost of overwriting and losing data can be very high.

Do you run with this option? Curious because I never use it. I think the cost of losing data is actually usually very small, it's only in a small 1% of cases that the data is valuable. I think that's actually the key insight behind the default, that expensive simulations are rare so it doesn't make sense to default it.

glwagner:
> Is the "so much boilerplate" just the extra one line overwrite_existing = true for each output writer?

It's one line --- but it's in every script, sometimes many times! Add all that up and you get to a huge amount...

glwagner:
@navidcy any thoughts?

josuemtzmo:
I agree with @ali-ramadhan,  keeping a `overwrite_existing` seems quite important to me too, potentially with a default value of `overwrite_existing = false`. 

Another issue to add into this discussion is the fact of how to handle output (e.g. `file.nc`) with checkpoints and `file_splitting`. Currently, when a simulation is pickup from a checkpoint having the flag `overwrite_existing = true` rewrites to empty all the pre-existing files (i.e. file_part1.nc, file_partN.nc). With `overwrite_existing = false`, the simulation crashes because it doesn't find the file `file.nc`. I think it will be useful to handle automatic concatenation to splitted files, to allow a more flexible output particularly in the context of HPC computing with manageable file sizes, chunks, and wall times.

glwagner:
> potentially with a default value of overwrite_existing = false.

Do you run with `overwrite_existing=false`? (Outside the context of restoring from a checkpoint.)

glwagner:
> Another issue to add into this discussion is the fact of how to handle output (e.g. file.nc) with checkpoints and file_splitting.

I propose handling this by initializing output files within `run!` rather than during instantiation of the output writer. This is separate from the interface discussion here though.

josuemtzmo:
> > potentially with a default value of overwrite_existing = false.
> 
> Do you run with `overwrite_existing=false`? (Outside the context of restoring from a checkpoint.)

I agree, that is not a common use case scenario. I have only used `overwrite_existing=false` without a checkpoint for short tests within the same Julia instance.

glwagner:
> I have only used overwrite_existing=false without a checkpoint for short tests within the same Julia instance to extend the model output.

Thank you for pointing out this use case. I think this is another situation that could be solved by waiting until `run!` for initialization. We can analyze an existing file and determine whether or not any data within the file will be overwritten.

Another idea by the way would be to move the concept of "overwriting" to `run!`, as well. Then the single keyword can apply to all output, or not, which presumably more aligned with what a user would want (rather than toggling `overwrite_existing` for each writer individually).

src/Simulations/callback.jl
josuemtzmo, line 1: Line 100 should also have the new GenericCallbackName() instead of GenericName().

glwagner: Ok! (Note you can insert your comment at the line you're referring to! Check out (4) here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request)

glwagner: You can also "suggest" a change, which makes things super easy for me since then all I have to do is click a button to push the suggested change up to the PR

josuemtzmo: Ok, I will push some changes in a bit with include that comment. That's why I didn't suggested it.

src/Simulations/callback.jl
```diff
@@ -109,8 +90,8 @@ already exists.
 
 The `callback` (which contains a schedule) can also be supplied directly.
 """
-function add_callback!(simulation, callback::Callback; name = GenericName())
```
glwagner, line 112: here right @josuemtzmo ?

src/Simulations/callback.jl
```diff
@@ -71,27 +71,8 @@ Callback(wta::WindowedTimeAverage, schedule; kw...) =
...
-        n += 1
-    end
-
-    return Symbol(prefix, n)
-end
+struct GenericCallbackName end
+unique_name(::GenericCallbackName, existing) = unique_name(:callback, existing)
 
 """
     add_callback!(simulation, callback::Callback; name = GenericName(), callback_kw...)
```
glwagner, line 78: I think this is the line you're talking about right @josuemtzmo ?

src/Simulations/callback.jl
glwagner: No you don't have to push the changes, that's not what I'm suggesting. See (7) here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request

image

glwagner:
Another idea after talking with @josuemtzmo: add another function called `checkpoint!(simulation)` that adds a checkpointer. (More to come on this idea)

glwagner:
Another utility that I believe is needed is a function that displays the information in an output file. For example something like

```julia
julia> outputinfo(filename)
```

which displays things like

* the names of the fields
* information about the fields? location, indices, size
* the grid
* the number of output snapshots
* the output times

anything else?

glwagner:
> I'm not too picky about `add_output_writer!` vs. `output_writer!` although `output!` is a bit too vague. I might lean too utilitarian though 🤖 Curious what other people think.

I'm starting to feel that it's better with "add"... following the precept that functions are verbs and objects are nouns.

> I do actually haha. But what I also do is set different output directories for each run so I can always go back and compare different runs as I'm iteratively modifying stuff. So I guess setting different directories is guaranteeing that I never overwrite existing files, but then `overwrite_existing = false` is only a guardrail for me.

Ah nice. That sounds like something one would do for an expensive simulation, ie if they take more than a few minutes, right? I think it makes sense to set `overwrite_existing = false` as a guardrail when working with expensive simulations.

The main thing for me is, if we _always_ start by writing a script with `overwrite_existing=true`.

> We might almost already have this with the `show` function of `FieldDataset` if it's called with `backend=OnDisk()`. Maybe it could be shared/re-purposed?

For sure that makes a lot of sense.

@navidcy @simone-silvestri it'd be nice to hear your thoughts.

navidcy:
I think `overwrite_existing = false` could be the default. Seems to me that users could easily add the `overwrite_existing = true` in their script if they know they are experimenting with something and don't wanna be deleting the output every time. Or they can also use `dir = "exp1"`, etc to separate their experimentations.

navidcy:
> I'm starting to feel that it's better with "add"... following the precept that functions are verbs and objects are nouns.

I find `add_output!(simulation)` to be more intuitive in terms of creating the output and `output!(simulation)` actually outputs. (Thus, in the latter "output" is the verb, not the noun.)

glwagner:
> I think overwrite_existing = false could be the default. Seems to me that users could easily add the overwrite_existing = true in their script if they know they are experimenting with something and don't wanna be deleting the output every time.

So, `overwrite_existing=false` is the current default.

I don't dispute that it's "easy" to add `overwrite_existing=true`, but I don't think we should design the interface only up to the point where "easy" changes close the gap between the default and what users want to do, most of the time. I think the default should legitimately be useful.

I think that `overwrite_existing=false` is rarely desirable -- users only want this when they run big, expensive simulations. I would say that it's almost impossible to set up a big, expensive simulation without going through some prototyping phase first. Moreover, prototyping consumes much more human time than big, expensive simulations. Therefore `overwrite_existing=true` is the correct default.

glwagner:
Apart from `overwrite_existing` default, it seems like we have consensus that we should use `add_output_writer!` for this PR. I'll make that change and switch over the examples as well.

I'll open an issue to discuss the `outputinfo` utility.

I also think that an `output!` function would be useful, which simply writes the current state. Recently I have been found myself wanting only the final state of the simulation. It's a little convoluted to have to set up an output writer with a schedule for that task, it'd be easier to write

```julia
run!(simulation)
output!("cool_stuff.jld2", simulation)
```

in the above, the filename goes first because that's the thing being "modified" (similar to how `save` works)

tomchor:
I vote to for an `add_` prefix in the function name. I don't think it's ugly and it clarifies things a fair bit.

Also, rather than creating new structs for the format, why not use the actual writer constructors in the user interface? i.e. instead of 

```julia
add_output!(simulation, outputs, NetCDFFormat(); kw...)
```

use

```julia
add_output!(simulation, outputs; writer=NetCDFOutputWriter, kw...)
```

tomchor:
> > Also, rather than creating new structs for the format, why not use the actual writer constructors in the user interface?
> 
> In a future where the "writer" is subsurface and users only interact with output through this interface, which do you think would be preferred? Why would we want to keep the "outputwriter" syntax around?

I don't think it makes much difference for the user as long as the argument makes it clear what kind of file will be written. So in that sense I think `NetCDFWriter` does the job as well as `NetCDFFormat` without creating another `struct`. I also don't feel strongly about it so feel free to ignore.

glwagner:
> > > Also, rather than creating new structs for the format, why not use the actual writer constructors in the user interface?
> > 
> > 
> > In a future where the "writer" is subsurface and users only interact with output through this interface, which do you think would be preferred? Why would we want to keep the "outputwriter" syntax around?
> 
> I don't think it makes much difference for the user as long as the argument makes it clear what kind of file will be written. So in that sense I think `NetCDFWriter` does the job as well as `NetCDFFormat` without creating another `struct`. I also don't feel strongly about it so feel free to ignore.

Hmm yes, the difference mainly then is that the new `NetCDFFormat()` is instantiated whereas we would not do the same with `NetCDFWriter`. 

If we use symbols then we can be potentially even more clear, eg

```julia
add_writer!(simulation, outputs, :netcdf; kw...)
```

glwagner:
I think using a new type is more future-proof. We can change the type `NetCDFOutputWriter` without changing the UI.

glwagner:
Update on this: given we changed `JLD2OutputWriter` to `JLD2Writer`, I think a good API for this is `add_writer!`. That's concise enough, and similar to `add_callback!` so easy to remember I think.

glwagner:
this didn't gain much traction and doesn't really seem necessary, so I am closing to focus on other priorities. Please re-open if you want to pursue it.

PR #3799, opened 2024-09-28
Enzyme (performance) mark some functions as non-differentiable

wsmoses:


wsmoses:
@glwagner the CI error seems unrelated?

glwagner:
@navidcy In the context of Enzyme (a tool for automatically differentiating LLVM), to "mark" a function as non-differentiable means to use `inactive_noinl` to tell Enzyme that it should not try to apply the chain rule to that function. Marking functions as non-differentiable has various reasons but here the reason is given as "performance". The code changed is also quite short so I think the description is complete.

glwagner (approved):

PR #3794, opened 2024-09-26
Massive clean up for conditional operations

glwagner:
This PR greatly simplifies the logic behind `conditional_operation` and how its extended for immersd boundary grid. The end result besides source code clean up seems to be an improvement in type inference.

Resolves #3750 I think. But @ali-ramadhan please test.

EDIT: I decided not to work on #3791 (here) because this requires fixing a bunch of tests and is a bigger effort.

PR #3801, opened 2024-09-30
New Base.axes for AbstractField

glwagner:
Resolves https://github.com/CliMA/Oceananigans.jl/issues/3750

Also supercedes #3794 where some tests are failing, not sure why. But that PR goes a bit further to clean up conditional ops. I think we want that too but let's start with this.

cc @ali-ramadhan


ali-ramadhan:
Confirming that this PR also resolves #3750 based on the MWE there!

```julia
julia> @time minimum(u2)
  0.061013 seconds (659 allocations: 49.336 KiB)
0.0

julia> @time minimum(u2.data)
  0.010983 seconds (3 allocations: 1.562 KiB)
0.0
```

PR #3812, opened 2024-10-02
Try using Krylov to solve the Poisson equation

### LINKED ISSUE #3803: Implement KrylovPoissonSolver
With @amontoison

We'd like to implement a Poisson solver that uses [`Krylov`](https://github.com/JuliaSmoothOptimizers/Krylov.jl) under the hood instead of our custom (preconditioned) conjugate gradient solver. This will open the door to [more solvers (that may be more appropriate for our pressure Poisson equation than conjugate gradient)](https://arxiv.org/abs/2310.01757) like conjugate residual, etc.

To make this work we need to overload some of Krylov's operators for Oceananigans `Field`:

* [`kaxpby!(n, s, x, dx, y, dy)`](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/200a9cd01c3bba906b82eaa87103aa2881983e5f/src/krylov_utils.jl#L326C1-L326C8) which performs `y = y + s * x` where `n` is the total length (eg `Nx * Ny * Nz`), `s` is the output, `dx` and `dy` are strides (irrelevant for us)
* [`kaxpby!(n, s, x, dx, t, y, dy)`](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/200a9cd01c3bba906b82eaa87103aa2881983e5f/src/krylov_utils.jl#L326C1-L326C8) which performs `y = t * y + s * x` where `n` is the total length (eg `Nx * Ny * Nz`), `s` is the output, `dx` and `dy` are strides (irrelevant for us)
* [`kdot`](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/200a9cd01c3bba906b82eaa87103aa2881983e5f/src/krylov_utils.jl#L310C1-L310C5)
* [`knrm2`](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/200a9cd01c3bba906b82eaa87103aa2881983e5f/src/krylov_utils.jl#L316C1-L316C6)
* Either `kcopy!` or `copyto!`

glwagner:
This PR adds support for Krylov.jl and aims to test it out for solving the Poisson equation.

cc @xkykai @amontoison

Closes #3803


glwagner:
To do:

- Implement `mul!` for the Laplacian operator
- Implement a wrapper type for `Field` that subtypes `AbstractVector`?
- Implement `ldiv!` for the FFT-based preconditioner (after the first two are done), planning to use [`ldiv=true` with `cg`](https://jso.dev/Krylov.jl/dev/preconditioners/#Examples)

This example illustrates implementing a custom operator:

https://jso.dev/Krylov.jl/dev/matrix_free/

Here is the documentation for implementing preconditioners in Krylov:

https://jso.dev/Krylov.jl/dev/preconditioners/

In Krylov, we want to remove broadcasting. 

To make `Field` look like a vector we just need

```julia
struct QuasiVector{FT, F} <: AbstractVector{FT}
    field :: F
    function QuasiVector(field)
        FT = eltype(field)
        F = typeof(field)
        return new{FT, FT}(field)
    end
end
```

amontoison:
@glwagner I remember now why the right-hand side `b` is an `AbstractVector{FC}`: I use the type `FC` for type stability with the keyword arguments (tolerances, etc.)

glwagner:
> @glwagner I remember now why the right-hand side `b` is an `AbstractVector{FC}`: I use the type `FC` for type stability with the keyword arguments (tolerances, etc.) as well as default values (`sqrt(eps(real(FC)))`).

True but then you could use `AbstractArray{FT}` similarly right?

amontoison:
Yes, I can use `AbstractArray{FT}`, but it should represent a vector.
For common array types, it always leads to an `AbstractVector`, so I never found the need to remove it.
I was always hesitant to relax this constraint because the type of `b` is used for allocating the "vector" in the workspace of a Krylov method.
If a user makes a mistake and provides a matrix with one column instead of a vector, it will fail internally with an unfriendly error message.

glwagner:
> If a user makes a mistake and provides a matrix with one column instead of a vector, it will fail internally with an unfriendly error message.

Ah for sure. I was just remarking that if you need the eltype, you can use `AbstractArray{FT}` (or you can simply call `eltype`, which is preferred by [YASGuide](https://github.com/jrevels/YASGuide) --- type parameters should be used for dispatch only).

I think it's ok if you keep `AbstractVector`. We can create a wrapper for Krylov.

amontoison:
@glwagner @xkykai 
I added a example in the documentation:
https://jso.dev/Krylov.jl/previews/PR927/custom_workspaces

glwagner:
> @glwagner @xkykai I added a example in the documentation for you: https://jso.dev/Krylov.jl/previews/PR927/custom_workspaces

Just a comment --- the Poisson operator is usually denoted $\nabla^2$ (rather than just $\nabla$ which denotes a gradient and yields a vector):

image

amontoison:
Thanks!
Good catch for typo with `\nabla`, I will fix it.

glwagner:
@amontoison this should be up to date and happy if you re-open this PR from a fork!

PR #3813, opened 2024-10-02
Fix formatting and adds back interior velocity in tilted bottom boundary layer example

tomchor:
The formatting was off in this example:

image

### DIFF
--- a/examples/tilted_bottom_boundary_layer.jl
+++ b/examples/tilted_bottom_boundary_layer.jl
@@ -112,11 +112,9 @@ B∞_field = BackgroundField(constant_stratification, parameters=(; ĝ, N² = N
 negative_background_diffusive_flux = GradientBoundaryCondition(∂z_b_bottom)
 b_bcs = FieldBoundaryConditions(bottom = negative_background_diffusive_flux)
 
-# ## Bottom drag
+# ## Bottom drag and along-slope interior velicity
 #
-# We impose bottom drag that follows Monin--Obukhov theory.
-# We include the background flow in the drag calculation,
-# which is the only effect the background flow enters the problem,
+# We impose bottom drag that follows Monin--Obukhov theory:
 
 V∞ = 0.1 # m s⁻¹
 z₀ = 0.1 # m (roughness length)
@@ -134,6 +132,14 @@ drag_bc_v = FluxBoundaryCondition(drag_v, field_dependencies=(:u, :v), parameter
 u_bcs = FieldBoundaryConditions(bottom = drag_bc_u)
 v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
 
+# Note that, similar to the buoyancy boundary conditions, we had to
+# include the background flow in the drag calculation.
+#
+# Let us also create `BackgroundField` for the along-slope interior velocity:
+
+@inline constant_velocity(x, z, t, p) = p.V∞
+V∞_field = BackgroundField(constant_velocity, parameters=(; V∞))
+
 # ## Create the `NonhydrostaticModel`
 #
 # We are now ready to create the model. We create a `NonhydrostaticModel` with an
@@ -150,7 +156,7 @@ model = NonhydrostaticModel(; grid, buoyancy, coriolis, closure,
                             advection = UpwindBiasedFifthOrder(),
                             tracers = :b,
                             boundary_conditions = (u=u_bcs, v=v_bcs, b=b_bcs),
-                            background_fields = (; b=B∞_field))
+                            background_fields = (; b=B∞_field, v=V∞_field))
 
 # Let's introduce a bit of random noise at the bottom of the domain to speed up the onset of
 # turbulence:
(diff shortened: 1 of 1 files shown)


tomchor:
I can create a separate issue to address this, but perhaps more importantly, I don't see the interior velocity `V∞` being added anywhere there except in the BCs. Usually we add as a background field to avoid inertial oscillations, and indeed it used to be done like that in [older versions of the docs](https://clima.github.io/OceananigansDocumentation/v0.73.8/generated/tilted_bottom_boundary_layer/),  but at some point this was changed and I don't really understand why.

I think it's okay to run the example without adding it as a background field, but then we're solving the problem in a reference frame that's moving with the interior flow and that should be made explicit and I don't see that explanation there. Personally, I'm in favor of including `V∞` as a background field because it's simpler.

CC @hdrake @glwagner who were in https://github.com/CliMA/Oceananigans.jl/pull/3581

glwagner (approved):

glwagner:
> I can create a separate issue to address this, but perhaps more importantly, I don't see the interior velocity `V∞` being added anywhere there except in the BCs. Usually we add it as a background field to avoid inertial oscillations, and indeed it used to be done like that in [older versions of the docs](https://clima.github.io/OceananigansDocumentation/v0.73.8/generated/tilted_bottom_boundary_layer/), but at some point this was changed and I don't really understand why (nor can I pinpoint where).
> 
> I think it's okay to run the example without adding it as a background field, but then we're solving the problem in a reference frame that's moving with the interior flow and that should be made explicit, and I don't see that explanation there. Personally, I'm in favor of including `V∞` as a background field because it's simpler.
> 
> CC @hdrake @glwagner who were in #3581

What's the problem with adding it back?

tomchor:
> > I can create a separate issue to address this, but perhaps more importantly, I don't see the interior velocity `V∞` being added anywhere there except in the BCs. Usually we add it as a background field to avoid inertial oscillations, and indeed it used to be done like that in [older versions of the docs](https://clima.github.io/OceananigansDocumentation/v0.73.8/generated/tilted_bottom_boundary_layer/), but at some point this was changed and I don't really understand why (nor can I pinpoint where).
> > I think it's okay to run the example without adding it as a background field, but then we're solving the problem in a reference frame that's moving with the interior flow and that should be made explicit, and I don't see that explanation there. Personally, I'm in favor of including `V∞` as a background field because it's simpler.
> > CC @hdrake @glwagner who were in #3581
> 
> What's the problem with adding it back?

No problem at all, and in my opinion that's what we should do. I'm just checking with the community first since it might have been removed for some reason I'm not aware of. Are you okay if I add it back?

glwagner:
Definitely ok. The problem makes more physical sense, plus it illustrates using a velocity in `BackgroundFields`. Also given that the drag is computed with this now-phantom velocity, the example doesn't make sense right now...

examples/tilted_bottom_boundary_layer.jl
```diff
@@ -134,6 +132,14 @@ drag_bc_v = FluxBoundaryCondition(drag_v, field_dependencies=(:u, :v), parameter
 u_bcs = FieldBoundaryConditions(bottom = drag_bc_u)
 v_bcs = FieldBoundaryConditions(bottom = drag_bc_v)
 
+# Note that, similar to the buoyancy boundary conditions, we had to
+# include the background flow in the drag calculation.
+#
+# Let us also create `BackgroundField` for the along-slope interior velocity:
+
+@inline constant_velocity(x, z, t, p) = p.V∞
+V∞_field = BackgroundField(constant_velocity, parameters=(; V∞))
```
glwagner, line 141:
```suggestion
V∞_field = BackgroundField(ConstantField(V∞))
```

It's also easy to make `BackgroundField(V∞)` work. I'm surprised we haven't added that!

glwagner: It's just a matter of adding

```julia
regularize_background_field(LX, LY, LZ, c::Number, grid, clock) = ConstantField(c)
```

https://github.com/CliMA/Oceananigans.jl/blob/main/src/Fields/background_fields.jl#L63

then you can write

```julia
V∞_field = BackgroundField(V∞)
```

glwagner: Definitely don't want to encourage the rather convoluted approach of writing a function to return a constant! We must be a force for good

tomchor: > It's also easy to make BackgroundField(V∞) work. I'm surprised we haven't added that!

Yeah I added that at first and was also surprised that there was no method for that.

I think I'll add it in another single-purpose PR and then do it more cleanly here :)

glwagner: I can add it if you want, it'll be less work

glwagner: The CI takes annoyingly long to finish so its nice to combine steps like this I think

tomchor: Ah, sorry I may have jumped the gun!

glwagner: no worries.

I usually feel swamped so try to do things like this quick. But it's good to take care.

PR #3814, opened 2024-10-02
Simplify kernel launching and add `exclude_periphery` option

glwagner:
cc @jagoosw


glwagner:
@jagoosw can you review this?

jagoosw:
I'm not sure I exactly understand whats changed here because of all the renaming. The important changes are here right?

https://github.com/CliMA/Oceananigans.jl/blob/7cc97812d42982b2c8bd5635b40f807e800b2472/src/Utils/kernel_launching.jl#L120-L158

And this makes it so that instead of being able to optionally fill the right boundary point, you can set it to `true` and it will instead compute 2:N for face fields (and presumably still 1:N for center fields)?

glwagner:
@wsmoses any idea what's causing this?

https://buildkite.com/clima/oceananigans/builds/17774#019258de-aa27-4f1d-845a-4c36f52d778a/18-568

navidcy:
Distributed tests actually fail with errors :(

jagoosw (approved):
Okay yeah that all makes sense. I agree the renaming makes sense it was just hard to see in the file changes which was new code vs renamed things. LGTM!

src/Utils/kernel_launching.jl
```diff
@@ -193,43 +215,66 @@ If the worksize is 0 after adjusting for active cells, the function returns `not
...
 """
+@inline function configure_kernel(arch, grid, workspec, kernel!;
+                                  exclude_periphery = false,
+                                  reduced_dimensions = (),
+                                  location = nothing,
+                                  active_cells_map = nothing)
+
+    if !isnothing(active_cells_map) # everything else is irrelevant
+        workgroup = min(length(active_cells_map), 256)
+        worksize = length(active_cells_map)
```
simone-silvestri, line 228: we probably need to keep this here for non-active domains
```suggestion
        if worksize == 0
            return nothing
        end
```

simone-silvestri: ok nothing, I see it has moved in the `launch!`

simone-silvestri:
I am not sure about the purpose of this PR. What is the reason we do not want to calculate the periphery in the nonhydrostatic model tendencies?
 If there is a specific reason, does this solution also apply to immersed boundaries with active cell map? It seems like nothing changed for that case.

simone-silvestri:
Also I am not sure about renaming to `buffer`. What is called `buffer` here is not really a buffer but a `halo-dependent` region of the active domain. What about calling it `compute_halo_dependent_tendencies!`?

glwagner:
> I am not sure about the purpose of this PR. What is the reason we do not want to calculate the periphery in the nonhydrostatic model tendencies?
> 
>  If there is a specific reason, does this solution also apply to immersed boundaries with active cell map? It seems like nothing changed for that case.

Because some open boundary conditions integrate an alternative PDE on the boundaries (eg for radiation conditions).

In other words the algorithm we are using is only valid for simple open boundary conditions (we designed it for impenetrable condition). More generally, the algorithm is wrong.

There are two other benefits. First, we do not use the tendencies on the periphery. So this increases the clarity of the code. Previously, it might not be obvious that even though we compute tendencies on the boundary, we overwrite the field values in `fill_halo_regions!`.

It saves a little bit of computation as well...

glwagner:
> Also I am not sure about renaming to `buffer`. What is called `buffer` here is not really a buffer but a `halo-dependent` region of the active domain. What about calling it `compute_halo_dependent_tendencies!`?

This PR _defines_ the "buffer" as the halo dependent region of the domain.

glwagner:
This should be done already (I'm not sure why the tests keep failing, its annoying).

I made the change because I seriously could not understand the code. I kept getting confused between "compute boundary tendencies" (which adds fluxes to the tendencies, coming in from the boundaries), and "compute tendencies_boundaries", which did something completely different (compute the halo dependent tendencies). So I started by changing the names so I could reason about the code without my head spinning.

glwagner:
@simone-silvestri can you resolve the conflicts?

Issue #3586, opened 2024-05-06
`BackgroundField` is a confusing name


glwagner:
It's not a `Field`. The docstring for the struct admits as much:

https://github.com/CliMA/Oceananigans.jl/blob/7a4b3f04e402be70d45fcb775a4dedef087f3bb0/src/Fields/background_fields.jl#L33-L41

and this temporary struct is always thrown away to build a "real" field:

https://github.com/CliMA/Oceananigans.jl/blob/7a4b3f04e402be70d45fcb775a4dedef087f3bb0/src/Fields/background_fields.jl#L63-L64

The problem is that the name leads to confusing patterns like

https://github.com/CliMA/Oceananigans.jl/blob/7a4b3f04e402be70d45fcb775a4dedef087f3bb0/examples/tilted_bottom_boundary_layer.jl#L98

where `B_field` is not a field. Indeed, later on when we need a "real" field we have to write

https://github.com/CliMA/Oceananigans.jl/blob/7a4b3f04e402be70d45fcb775a4dedef087f3bb0/examples/tilted_bottom_boundary_layer.jl#L178

Can we think of a better name? For example `BackgroundFieldInfo` is one. That's sorta clumsy though. Maybe we can come up with something more concise and elegant.

cc @tomchor @hdrake @liuchihl because you are using this feature

glwagner:
Not bad...

PR #3815, opened 2024-10-02
Add method to permit creating `BackgroundField`s out of `Number`s directly

tomchor:
See https://github.com/CliMA/Oceananigans.jl/pull/3813

### DIFF
--- a/src/Fields/background_fields.jl
+++ b/src/Fields/background_fields.jl
@@ -60,7 +60,7 @@ func(x, y, z, t, parameters)
 """
 BackgroundField(func; parameters=nothing) = BackgroundField(func, parameters)
 
-regularize_background_field(LX, LY, LZ, c::Number, grid, clock) = ConstantField(c)
+regularize_background_field(LX, LY, LZ, bf::BackgroundField{<:Number}, grid, clock) = ConstantField(bf.func)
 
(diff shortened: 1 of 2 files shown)


glwagner:
Nice. It's not tested until we use that example though. We could also go the extra step of testing it directly if we are going to add it separately>?

glwagner:
happy to merge this as is and then test it with the example

glwagner (approved):

tomchor:
I think we should test this independently of an example. Examples change but it'd be nice to make sure this functionality still works. I'll add something simple and then we can iterate.

src/Fields/background_fields.jl
```diff
@@ -60,6 +60,8 @@ func(x, y, z, t, parameters)
 """
 BackgroundField(func; parameters=nothing) = BackgroundField(func, parameters)
 
+regularize_background_field(LX, LY, LZ, c::Number, grid, clock) = ConstantField(c)
```
glwagner, line 63:
```suggestion
regularize_background_field(LX, LY, LZ, bg::BackgroundField{<:Number}, grid, clock) = ConstantField(bg.func)
```

My b, this is what it should look like

I guess we assumed the background fields would be functions and called that property `func`...

I'd chalk up that cleanup to also resolving #3586 not worth tackling here

PR #3819, opened 2024-10-03
Neural network parameterization of salty turbulent mixing in the upper ocean

xkykai:
This PR aims to incorporate trained neural networks that represent the entrainment process due to turbulent convection in the upper ocean into Oceananigans.jl. This parameterization is designed to work using the TEOS-10 nonlinear equation of state, hence it works only in typical oceanographic conditions on Earth.


xin_kai_vertical_diffusivity.jl
```diff
@@ -0,0 +1,255 @@
...
+    # Convective adjustment diffusivity
+    ν_local = ifelse(convecting, - (νᶜⁿ - νˢʰ) / 2 * tanh(Ri / δRi) + νˢʰ, clamp(Riᶜ * Ri + νˢʰ + ν₀, ν₀, νˢʰ))
+
+    # Entrainment diffusivity
+    x = Qᵇ / (N² + 1e-11)
+    ν_nonlocal = ifelse(entraining,  Cᵉⁿ * νᶜⁿ * 0.5 * (tanh((x - Q₀) / δQ) + 1), 0)
+
+    # Update by averaging in time
+    @inbounds diffusivities.κᵘ[i, j, k] = ifelse(k <= 1 || k >= grid.Nz+1, 0, ν_local + ν_nonlocal)
+    @inbounds diffusivities.κᶜ[i, j, k] = ifelse(k <= 1 || k >= grid.Nz+1, 0, (ν_local + ν_nonlocal) / Prₜ)
```
glwagner, line 252:
```suggestion
    @inbounds diffusivities.κᵘ[i, j, k] = ifelse((k <= 1) | (k >= grid.Nz+1), 0, ν_local + ν_nonlocal)
    @inbounds diffusivities.κᶜ[i, j, k] = ifelse((k <= 1) | (k >= grid.Nz+1), 0, (ν_local + ν_nonlocal) / Prₜ)
```

xin_kai_vertical_diffusivity.jl
```diff
@@ -0,0 +1,255 @@
...
+    # Conditions
+    convecting = Ri < 0 # applies regardless of Qᵇ
+    entraining = (Ri > 0) & (Ri_above < 0) & (Qᵇ > 0)
+
+    # Convective adjustment diffusivity
+    ν_local = ifelse(convecting, - (νᶜⁿ - νˢʰ) / 2 * tanh(Ri / δRi) + νˢʰ, clamp(Riᶜ * Ri + νˢʰ + ν₀, ν₀, νˢʰ))
+
+    # Entrainment diffusivity
+    x = Qᵇ / (N² + 1e-11)
+    ν_nonlocal = ifelse(entraining,  Cᵉⁿ * νᶜⁿ * 0.5 * (tanh((x - Q₀) / δQ) + 1), 0)
```
glwagner, line 248: careful with number type

xin_kai_vertical_diffusivity.jl
glwagner, line 1: if this works better than ri-based, I'm ok to upgrade ri-based to this. I suppose you've tested that?

NN_closure_global_nof_BBLintegral.jl
```diff
@@ -0,0 +1,224 @@
...
+@inline diffusive_flux_x(i, j, k, grid, clo::NNC, K, ::Val{tracer_index}, c, clock, fields, buoyancy) where tracer_index = zero(grid)
+@inline diffusive_flux_y(i, j, k, grid, clo::NNC, K, ::Val{tracer_index}, c, clock, fields, buoyancy) where tracer_index = zero(grid)
+
+# Viscous fluxes are zero (for now)
+@inline viscous_flux_uz(i, j, k, grid, clo::NNC, K, clk, fields, b) = zero(grid)
+@inline viscous_flux_vz(i, j, k, grid, clo::NNC, K, clk, fields, b) = zero(grid)
+
+# The only function extended by NNFluxClosure
+@inline diffusive_flux_z(i, j, k, grid, clo::NNC, K, ::Val{1}, c, clock, fields, buoyancy) = @inbounds K.wT[i, j, k]
+@inline diffusive_flux_z(i, j, k, grid, clo::NNC, K, ::Val{2}, c, clock, fields, buoyancy) = @inbounds K.wS[i, j, k]
```
glwagner, line 224: seems like it would be better to have the tracer names here since `tracers=(:S, :T)` would make this fail right?

It's possible to refactor diffusive fluxes for that, eg we already use names for the bgc:

https://github.com/CliMA/Oceananigans.jl/blob/d367f112e0ae5c279fe3a63f1726469540d4eb5b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_tendency_kernel_functions.jl#L136

NN_closure_global_nof_BBL.jl
```diff
@@ -0,0 +1,224 @@
...
+                  Adapt.adapt(to, nn.scaling))
+
+Adapt.adapt_structure(to, nn :: NN) = 
+    NN(Adapt.adapt(to, nn.model), 
+       Adapt.adapt(to, nn.ps), 
+       Adapt.adapt(to, nn.st))
+
+function NNFluxClosure(arch)
+    dev = ifelse(arch == GPU(), gpu_device(), cpu_device())
+    nn_path = "./NDE_FC_Qb_nof_BBL_trainFC24new_scalingtrain54new_2layer_64_relu_2Pr_model.jld2"
```
glwagner, line 71: how big is this file?

2D_model_LES_sin.jl
```diff
@@ -0,0 +1,225 @@
...
+using Statistics
+
+using Oceananigans
+using Oceananigans.Units
+using Oceananigans.OutputReaders: FieldTimeSeries
+using Oceananigans.Grids: xnode, ynode, znode
+using SeawaterPolynomials
+using SeawaterPolynomials:TEOS10
+import SeawaterPolynomials.TEOS10: s, ΔS, Sₐᵤ
+s(Sᴬ::Number) = Sᴬ + ΔS >= 0 ? √((Sᴬ + ΔS) / Sₐᵤ) : NaN
```
glwagner, line 13: I think this is an interesting idea. Just need to use non-short-circuiting logic and be mindful of number type and it could potentially go in `SeawaterPolynomials`

xin_kai_vertical_diffusivity.jl
xkykai: out of curiosity, why do we prefer bitwise OR over the short circuiting version?

xkykai: out of curiosity, why do we prefer bitwise OR over the short circuiting version?

xin_kai_vertical_diffusivity.jl
xkykai: you mean change it to
```julia
x = Qᵇ / (N² + 10^-11)
```
right?

xin_kai_vertical_diffusivity.jl
xkykai: yes, this works quite well.

2D_model_LES_sin.jl
xkykai: I was writing this in order to address the issue of 
```julia
ERROR: DomainError with -2.0:
sqrt was called with a negative real argument but will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
```
when I was training my model, in order to continue the training with `NaN`. 
Wouldn't doing non-short-circuiting logic mean that we are forced into a `DomainError` if `Sᴬ + ΔS` becomes negative?

xin_kai_vertical_diffusivity.jl
glwagner: What's your strategy re: having this tied to this PR, rather than implemented in a separate PR?

NN_closure_global_nof_BBL.jl
xkykai: less that 1MB, maximally perhaps 2MB. Do you think we should host this file in this repository? Or do it somewhere else using something like `DataDeps.jl`

xin_kai_vertical_diffusivity.jl
xkykai: if you think it's okay I can open a separate PR to replace `RiBasedVerticalDiffusivity` with this

2D_model_LES_sin.jl
glwagner: Hmm, we could clip to avoid it, for example:

```julia
function s(Sᴬ)
    not_positive = Sᴬ + ΔS >= 0
    d = max(zero(Sᴬ), Sᴬ + ΔS)
    return ifelse(not_positive, NaN, √(d / Sₐᵤ))
end
```

but after reflecting on this a bit more, I think it's better to receive a `DomainError` from this point, than to get a NaN and not know why.

xkykai: I agree. I believe I started doing this to throw `NaN`s when using EKI, which was necessary to continue training, but apart from training purposes there shouldn't be a need to do this

glwagner: Ahh no but I see your point now.

The thing is, we really do want to support training / automatic calibration and it shouldn't require hacks like this, I feel this really impacts reproducibility and understandability (technically this is type piracy...)

Maybe clipping+NaN should be an option of the equation of state then.

glwagner: Another option would be to wrap `run!(simulation)` inside a `try/catch` so that, if a simulation errors, we can replace the output with NaN / mark the simulation as failed for the purpose of estimating parameters. This would work too right?

xkykai: I suspect what one could do is throw `NaN` and a warning message that this is due to the model having salinity values that are beyond the regime where TEOS-10 is correct.

glwagner: We can't warn from inside a kernel though.

xkykai: ah, so try-catch it is then

xin_kai_vertical_diffusivity.jl
navidcy: No, `1e-11` is fine!

I think @glwagner means that you what you wrote above might imply that 

```julia
ν_nonlocal = 0
```

which is an Int. Right?

But also, if `Qᵇ` and `N²` are Float32, then by adding `1e-11` which is Float64 you implicitly convert the number type...

glwagner: two things.

first `1e-11` should be a parameter anyways. So users can change it. Basic rule: never hard code. Maybe one run with a hard coded parameter. If it works, implement it.

Second is that number types have to be converted, unless you can use an integer which is promoted. So

```julia
small_number = convert(FT, 1e-11)
x = Qᵇ / (N² + small_number)
```

remember that `1e-11` is Float64 and `x` will be promoted to `Float64` even if the other members of the computation are Float32

glwagner:
@xkykai this PR seems to have gone hopelessly stale. I think it would be nice to support NORi but perhaps it has to be done in another PR, or another package (if it adds dependencies). An extension could also be used.

xkykai:
Let's close this for now.

PR #3820, opened 2024-10-04
Support for different file formats & creation of files in runtime

josuemtzmo:
This PR completes the output_helpers.jl to support different output_writers: 
- NetCDFOutputWriter
- Checkpointer

Additionally, it ensures that the output of `NetCDFOutputWriter` and `JLD2OutputWriter` are created during the simulation `run!` instead of when the writer is instantiated. 

I don't particularly like the fact that the `NetCDFOutputWriter` struct requires to parametric value for the dataset, which requires to open a dataset to instantiate it. I solved this by opening a temporal netCDF the first time it's called ("a hack"), i.e. `Dataset(Base.Filesystem.tempname(),"c")`.

### DIFF
--- a/src/OutputWriters/netcdf_output_writer.jl
+++ b/src/OutputWriters/netcdf_output_writer.jl
@@ -483,19 +483,8 @@ function NetCDFOutputWriter(model, outputs;
     # Ensure we can add any kind of metadata to the global attributes later by converting to Dict{Any, Any}.
     global_attributes = Dict{Any, Any}(global_attributes)
 
-    dataset, outputs, schedule = initialize_nc_file!(filepath,
-                                                     outputs,
-                                                     schedule,
-                                                     array_type,
-                                                     indices,
-                                                     with_halos,
-                                                     global_attributes,
-                                                     output_attributes,
-                                                     dimensions,
-                                                     overwrite_existing,
-                                                     deflatelevel,
-                                                     grid,
-                                                     model)
+    # Create empty temporal dataset in the first call.
+    dataset = Dataset(Base.Filesystem.tempname(),"c")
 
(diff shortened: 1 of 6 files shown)


src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -483,19 +483,8 @@ function NetCDFOutputWriter(model, outputs;
...
-                                                     with_halos,
-                                                     global_attributes,
-                                                     output_attributes,
-                                                     dimensions,
-                                                     overwrite_existing,
-                                                     deflatelevel,
-                                                     grid,
-                                                     model)
+    # Create empty temporal dataset in the first call.
+    dataset = Dataset(Base.Filesystem.tempname(),"c")
```
josuemtzmo, line 487: @glwagner This is the only way I came up to be able to use the current `NetCDFOutputWriter` struct. I don't see an easy work around, but I also think it's not great to create an empty netCDF file each time we create a new instance of the `NetCDFOutputWriter`

glwagner:
Can we keep these features in separate PRs? The output interface really is distinct from the lower-level features needed for the NetCDFOutputWriter.

josuemtzmo:
Ok

glwagner:
Sorry maybe there was a miscommunication. The `output!` interface does not delay creating files until `run!`. 

Delaying output initialization to `run!` does not exist. I would like to implement it, however. (In another PR.)

glwagner:
Could be good to chat, can you get in touch on the Julia slack?

josuemtzmo:
I'm closing this PR after the discussion with @glwagner, to split it into different and smaller PRs.

PR #3822, opened 2024-10-04
Added momentum equation test

jlk9:
This adds a test for Enzyme AD applied to a hydrostatic free surface model with `ExplicitFreeSurface`  momentum advection.

### DIFF
--- a/test/test_enzyme.jl
+++ b/test/test_enzyme.jl
@@ -105,6 +105,37 @@ function set_initial_condition_via_launch!(model_tracer, amplitude)
     return nothing
 end
 
+function momentum_equation!(model)
+        
+    # Do time-stepping
+    Nx, Ny, Nz = size(model.grid)
+    Δz = 1 / Nz
+    Δt = 1e-1 * Δz^2
+
+    model.clock.time = 0
+    model.clock.iteration = 0
+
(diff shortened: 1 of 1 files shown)


test/test_enzyme.jl
```diff
@@ -236,3 +267,73 @@ end
...
+    # 3. Do a problem where we invert for the tracer fluxes (maybe with CATKE)
+
+    buoyancy = SeawaterBuoyancy(equation_of_state=LinearEquationOfState())
+
+    model = HydrostaticFreeSurfaceModel(; grid,
+                                        momentum_advection = WENO(),
+                                        free_surface = ExplicitFreeSurface(gravitational_acceleration=1),
+                                        #buoyancy = buoyancy,
+                                        #tracers = (:T, :S),
+                                        #velocities = PrescribedVelocityFields(; u, v),
```
glwagner, line 311:
```suggestion
```

glwagner (approved):
I suggest you set up a system that allows you to loop through model properties to generate a bunch of tests automatically. For example, we can run the current test with a few advection schemes. If you set the models up with a passive tracer `c`, you can use derivatives of that for all configurations including for prescribed velocities and prognostic velocities.

PR #3823, opened 2024-10-04
Support AbstractOperations with MakieExtension

glwagner:
Kinda cool to be able to do

```julia
heatmap(c^2)
```

and stuff

### DIFF
--- a/ext/OceananigansMakieExt.jl
+++ b/ext/OceananigansMakieExt.jl
@@ -1,6 +1,7 @@
 module OceananigansMakieExt
 
 using Oceananigans
+using Oceananigans.AbstractOperations: AbstractOperation
 using Oceananigans.Architectures: on_architecture
 using Oceananigans.ImmersedBoundaries: mask_immersed_field!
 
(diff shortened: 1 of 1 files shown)


ext/OceananigansMakieExt.jl
```diff
@@ -41,7 +66,7 @@ grids) with NaNs;
 - transferring data from GPU to CPU if necessary.
 """
 function make_plottable_array(f)
-
+    compute!(f)
```
navidcy, line 69: so is `compute!(::Field) = nothing`?

glwagner: yes https://github.com/CliMA/Oceananigans.jl/blob/840f6026c3025872e57c66e9b134ffbeb24bdc8c/src/Fields/field.jl#L440

navidcy (approved):
lgtm!

ext/OceananigansMakieExt.jl
navidcy: great! thanks! mostly my question was a sanity check for my understanding

PR #3824, opened 2024-10-05
Temporary support for OrthogonalSphericalShellGrid for MakieExt

glwagner:
This is a placeholder until we can use `mesh` to reinterpret `heatmap` in geographic coordinates, like ImaginOcean does. Otherwise, plotting fields on `OrthogonalSphericalShellGrid` fails because the coordinate vectors are not 1D.


glwagner:
Here's how Imaginocean does it:

https://github.com/navidcy/Imaginocean.jl/blob/f5cc5f27dd2e99e0af490e8dca5a53daf6837ead/src/Imaginocean.jl#L259

I think it could be fun to make that code run under the hood for any `heatmap`

navidcy (approved):

PR #3825, opened 2024-10-05
Don't return special length for PartlyInMemory time series

glwagner:
I'm not sure why this was added, so we will see if something breaks. But through usage I've realized that its inconvenient (and unexpected) that the length of `FieldTimeSeries` depends on the backend. My intuition is that the length stays the same regardless of _where_ the data is (in memory, or on disk, or a combination of the two). It's also helpful that behavior is the same between backends, which makes switching backends for performance reasons painless.


simone-silvestri (approved):
seems like all the tests pass.

PR #3827, opened 2024-10-07
Bugfix in `work_layout`

simone-silvestri:
proabably because of previous refactoring


glwagner:
This would be superceded by #3814

glwagner:
I believe this PR has been superceded

PR #3830, opened 2024-10-07
Fix `ConjugateGradientPoissonSolver` construction with default pre-conditioner

### LINKED ISSUE #3829: Cannot construct a `ConjugateGradientPoissonSolver` with a default preconditioner
I'll open a PR to add a test that covers this.

---

MWE:

```julia
using Oceananigans

grid = RectilinearGrid(
    CPU(),
    Float64,
    size = (16, 16, 16),
    x = (0, 1),
    y = (0, 1),
    z = (-1, 0)
)

ConjugateGradientPoissonSolver(grid)
```

Error:

```
ERROR: UndefVarError: `ImmersedBoundaryGrid` not defined
Stacktrace:
 [1] ConjugateGradientPoissonSolver(grid::RectilinearGrid{…}; preconditioner::Oceananigans.Solvers.DefaultPreconditioner, reltol::Float64, abstol::Float64, kw::@Kwargs{})
   @ Oceananigans.Solvers ~/atdepth/Oceananigans.jl/src/Solvers/conjugate_gradient_poisson_solver.jl:54
 [2] ConjugateGradientPoissonSolver(grid::RectilinearGrid{Float64, Bounded, Bounded, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{…}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{…}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{…}}, CPU})
   @ Oceananigans.Solvers ~/atdepth/Oceananigans.jl/src/Solvers/conjugate_gradient_poisson_solver.jl:47
 [3] top-level scope
   @ REPL[6]:1
```

ali-ramadhan:
I added a test that fails due to #3829. I should also test the non-FFT case and test that the proper pre-conditioner was initialized. So the test could be better but it does fail as it should.

What we need is a `using Oceananigans.ImmersedBoundaries` but the solvers module is defined well before the immersed boundaries module.

So for a solver to depend on the immersed boundaries module, and really just the `ImmersedBoundaryGrid` type then the immersed boundaries module needs to be included first.

Based on these comments maybe it's desirable to change the order of inclusion? But maybe it'll take some work. So otherwise we probably need to define another abstract type in `src/Oceananigans.jl` but this solution isn't ideal.

https://github.com/CliMA/Oceananigans.jl/blob/13bf409616af8c155b72d8869b7b8f97ae0e844b/src/Oceananigans.jl#L214-L228

Resolves #3829

### DIFF
--- a/test/test_preconditioned_conjugate_gradient_solver.jl
+++ b/test/test_preconditioned_conjugate_gradient_solver.jl
@@ -1,6 +1,6 @@
 include("dependencies_for_runtests.jl")
 
-using Oceananigans.Solvers: solve!
+using Oceananigans.Solvers: solve!, ConjugateGradientPoissonSolver
 using Statistics
 
 function identity_operator!(b, x)
@@ -61,11 +61,17 @@ function run_poisson_equation_test(grid)
     return nothing
 end
 
+function construct_conjugate_gradient_poisson_solver_with_default_preconditioner(grid)
+    solver = ConjugateGradientPoissonSolver(grid)
+    return true
+end
+
 @testset "ConjugateGradientSolver" begin
     for arch in archs
         @info "Testing ConjugateGradientSolver [$(typeof(arch))]..."
         grid = RectilinearGrid(arch, size=(4, 8, 4), extent=(1, 3, 1))
         run_identity_operator_test(grid)
         run_poisson_equation_test(grid)
+        construct_conjugate_gradient_poisson_solver_with_default_preconditioner(grid)
     end
 end
--- a/src/Oceananigans.jl
+++ b/src/Oceananigans.jl
@@ -6,16 +6,16 @@ module Oceananigans
 
 export
     # Architectures
-    CPU, GPU, 
+    CPU, GPU,
 
     # Logging
     OceananigansLogger,
 
     # Grids
     Center, Face,
-    Periodic, Bounded, Flat, 
+    Periodic, Bounded, Flat,
     FullyConnected, LeftConnected, RightConnected,
-    RectilinearGrid, 
+    RectilinearGrid,
     LatitudeLongitudeGrid,
     OrthogonalSphericalShellGrid,
     xnodes, ynodes, znodes, nodes,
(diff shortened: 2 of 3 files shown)


test/test_preconditioned_conjugate_gradient_solver.jl
```diff
@@ -61,11 +61,17 @@ function run_poisson_equation_test(grid)
...
+    return true
+end
+
 @testset "ConjugateGradientSolver" begin
     for arch in archs
         @info "Testing ConjugateGradientSolver [$(typeof(arch))]..."
         grid = RectilinearGrid(arch, size=(4, 8, 4), extent=(1, 3, 1))
         run_identity_operator_test(grid)
         run_poisson_equation_test(grid)
+        construct_conjugate_gradient_poisson_solver_with_default_preconditioner(grid)
```
glwagner, line 75:
```suggestion
    solver = ConjugateGradientPoissonSolver(grid)
    @test solver isa ConjugateGradientPoissonSolver
```

test/test_preconditioned_conjugate_gradient_solver.jl
```diff
@@ -61,11 +61,17 @@ function run_poisson_equation_test(grid)
     return nothing
 end
 
+function construct_conjugate_gradient_poisson_solver_with_default_preconditioner(grid)
+    solver = ConjugateGradientPoissonSolver(grid)
+    return true
+end
```
glwagner, line 67:
```suggestion
```

glwagner:
Oof, good point.

As a temporary fix, we could add a special constructor for `ImmersedBoundaryGrid` (ie use dispatch rather than the if-statement)

simone-silvestri:
To do this, we need to shift the definitions of
- immersed tridiagonal solver (primarily in `abstract_grid_fitted_boundary.jl`) in the Solver or TurbulenceClosures module
- distributed immersed boundary definitions  (the `distributed_immersed_boundaries.jl` file mostly) in the DistributedComputations module
- the viscous fluxes (in `conditional_fluxes.jl`) in the TurbulenceClosure module 
- probably better at this point to include ImmersedBoundaries before the Advection module and shift the advective fluxes definition  (in `conditional_fluxes.jl`) in Advection

This will make the `ImmersedBoundaries` module very light weight and the others a bit more complicated but it something we discussed already and I think it's a positive change now that immersed boundaries are less "experimental". @ali-ramadhan  I can help on this PR if you want.

glwagner:
It'd be easier to understand the suggestions if you put the links to the source code in your comment.

ali-ramadhan:
Thanks for the suggestions @simone-silvestri! And I noticed you started doing some refactoring in PR #3847 so I'll close this in favor of your PR.

PR #3834, opened 2024-10-08
Open halo filling on Windowed fields

### LINKED ISSUE #3833: Open boundary conditions do not work for windowed fields
There is some problem with open boundary conditions and windowed fields
MWE:
```julia
grid = RectilinearGrid(size = (5, 5, 5), extent = (1, 1, 1), topology  = (Bounded, Bounded, Bounded))
u, v, _ = Oceananigans.Fields.VelocityFields(grid)
V = Field(v, indices = (:, :, 5))
fill!(V, 1)
Oceananigans.BoundaryConditions.fill_halo_regions!(V)
```
I would expect that V[:, 1, :] = 0 but
```julia
julia> V[:, 1, :]
11×1 OffsetArray(::Matrix{Float64}, -2:8, 5:5) with eltype Float64 with indices -2:8×5:5:
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
```

simone-silvestri:
Should close #3833. Probably, we can wait to merge #3792 before this one and add some tests for fill halo regions here


simone-silvestri:
There was a missing offset that was preventing filling the open halos for windowed fields correctly

glwagner:
> There was a missing offset that was preventing filling the open halos for windowed fields correctly

Put a comment in the code on the key lines to prevent the problem from happening again. This PR only fixes the code but does not address the underlying issue that it's easy to miss the need for that.

src/BoundaryConditions/fill_halo_regions_open.jl
```diff
@@ -15,9 +15,11 @@ function fill_open_boundary_regions!(field, boundary_conditions, indices, loc, g
...
     # which is the function which fills non-open boundaries at `loc` which informs `fill_halo_size` 
     open_fill, regular_fill = get_open_halo_filling_functions(loc) 
-    fill_size = fill_halo_size(field, regular_fill, indices, boundary_conditions, loc, grid)
-
-    launch!(arch, grid, fill_size, open_fill, field, left_bc, right_bc, loc, grid, args)
+    size = fill_halo_size(field, regular_fill, indices, boundary_conditions, loc, grid)
+    offset = fill_halo_offset(size, regular_fill, indices)
+    params = KernelParameters(size, offset)
+    
+    launch!(arch, grid, params, open_fill, field, left_bc, right_bc, loc, grid, args)
```
glwagner, line 22: It's hard for me to tell how what these changes do so a comment would help.

glwagner (approved):

src/BoundaryConditions/fill_halo_regions_open.jl
```diff
@@ -15,9 +15,15 @@ function fill_open_boundary_regions!(field, boundary_conditions, indices, loc, g
     # gets `open_fill`, the function which fills open boundaries at `loc`, as well as `regular_fill`
     # which is the function which fills non-open boundaries at `loc` which informs `fill_halo_size` 
     open_fill, regular_fill = get_open_halo_filling_functions(loc) 
-    fill_size = fill_halo_size(field, regular_fill, indices, boundary_conditions, loc, grid)
 
-    launch!(arch, grid, fill_size, open_fill, field, left_bc, right_bc, loc, grid, args)
+    # Computing the size of the fill halo kernel. We use the `regular_fill` function to
+    # calculate the index range. Remember: windowed fields require the halos
+    # to be computed only on a part of the boundary so we need to offset the kernel indices
```
glwagner, line 21: I don't understand how we "use a function" to calculate an index range. This doesn't make sense to me. The index range depends on indices, loc, boundary conditions, grid? Why don't we use those arguments to compute the index range?

jagoosw: This has been resolved in https://github.com/CliMA/Oceananigans.jl/pull/3792

glwagner: Touche, I thought it was just me that has a weird dream where I was a broken record going round and round...

simone-silvestri:
@jagoosw I have changed the PR a bit to reuse the implementation in `fill_halo_regions.jl` because I think we do not need to define other functions; we just need to set to `nothing` the BCs that are not open in the signature of the `fill_halo_event!`

Let me know if something like this would work or if you see glaring errors in the implementation.

jagoosw:
> @jagoosw I have changed the PR a bit to reuse the implementation in `fill_halo_regions.jl` because I think we do not need to define other functions; we just need to set to `nothing` the BCs that are not open in the signature of the `fill_halo_event!`
> 
> Let me know if something like this would work or if you see glaring errors in the implementation.

That looks good! I can't see anything wrong.

src/BoundaryConditions/fill_halo_regions_open.jl
```diff
@@ -14,17 +14,19 @@ function fill_open_boundary_regions!(field, boundary_conditions, indices, loc, g
...
+    if !isnothing(fill_halo_function)
+        # Computing the parameters of the fill halo kernel. 
+        # Remember: windowed fields require the halos to be computed only on a part of the boundary 
+        # so we need to offset the kernel indices 
+        size   = fill_halo_size(field, fill_halo_function, indices, boundary_conditions, loc, grid)
+        offset = fill_halo_offset(size, fill_halo_function, indices)
+        params = KernelParameters(size, offset)
+        
+        launch!(arch, grid, params, fill_halo_function, field, left_bc, right_bc, loc, grid, args)
+    end
```
glwagner: Why not replace this with a function that fills the halos?

simone-silvestri: Yeah, I changed this PR to reuse the implementation we already have in `fill_halo_regions.jl`. I found nothing special for open boundary conditions to warrant writing a new function.

PR #3836, opened 2024-10-08
CI with Julia v1.11

navidcy:


Sbozzolo:
> Note the the distributed CI still does not have Julia v1.11 (right @Sbozzolo?) so there Julia v1.10 is used. This is possible because there are two Manifests.

I installed julia 1.11 on the Caltech cluster, but we haven't made a module yet (https://github.com/CliMA/ClimaModules/pulls)

Sbozzolo (approved):

glwagner:
> This is possible because there are two Manifests.

There's only one manifest?

ali-ramadhan:
> > This is possible because there are two Manifests.
> 
> There's only one manifest?

I assume he's referring to this new feature of Julia 1.11: https://julialang.org/blog/2024/10/julia-1.11-highlights/#manifest_versioning

navidcy:
> > This is possible because there are two Manifests.
> 
> There's only one manifest?

The PR adds another Manifest. Or we can just have the Manifest for v1.11 and drop the one for v1.10

glwagner:
> keeps the older Manifest that works for v1.10

The Manifest was deleted in #3783

navidcy:
oh great! I missed that!
I wanted to delete it long now but I got resistance doing that before... Great! So I'll delete the new Manifest then as well!

glwagner:
> oh great! I missed that! I wanted to delete it long now but I got resistance doing that before... Great! So I'll delete the new Manifest then as well!

Deleting it seemed to help increase the likelihood that CI passed. Although, it did not fully solve the problem (and note a few other changes were also made on #3783).

simone-silvestri (approved):

navidcy:
Noting that internal_tide.jl gives NaN with Julia v1.11 while all is OK with Julia v1.10; something with immersed boundaries....? I'm looking into it.

simone-silvestri:
I think it's a plotting issue. We are filling up the immersed boundaries with NaN and, apparently, we cannot plot NaNs anymore? The error says:
```julia
ERROR: LoadError: On worker 2:
  | Looking up a non-finite or NaN value in a colormap is undefined.
```

glwagner:
> I ran the script and the actual simulation NaN-ed.

That means Oceanangians isn't compatible with julia 1.11.

Do any other tests catch the issue? We can use this opportunity to add more tests.

navidcy:
I’m trying to make an mwe

ali-ramadhan:
Weird that this one example is behaving quite differently.

In Julia 1.10.6, the example produces flows that generally remain at a max CFL of ~0.13 or lower.

In Julia 1.11.1, the flow reaches a max CFL of ~0.2 after ~60 iterations and blows up within ~100 iterations.

Took a quick look but couldn't find anything obviously different.

Are there any tests that time step a hydrostatic model with an immersed boundary for 100+ time steps?

glwagner:
> Are there any tests that time step a hydrostatic model with an immersed boundary for 100+ time steps?

There are regression tests.

simone-silvestri:
> Are there any tests that time step a hydrostatic model with an immersed boundary for 100+ time steps?

There are some tests but quite fragmented, not very exhaustive and that run only for 10 time steps, [here](https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_dynamics.jl) for diffusion, [here](https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_immersed_advection.jl) for advection and [here](https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_distributed_hydrostatic_model.jl) for distributed vs serial. 

There is scope to add more tests for the immersed boundary grid, especially for a partial cell bottom, which I believe is still not tested.
I can start adding some tests for it.

glwagner:
> Are there any tests that time step a hydrostatic model with an immersed boundary for 100+ time steps?

Good to add tests though I don't know why 100+ time steps matters specifically?

ali-ramadhan:
I can confirm that Julia 1.11 is changing how the immersed boundary (and free surface?) works. Here's the difference (including halos) in the internal tide u field after 1 time step (difference is Julia 1.10 - Julia 1.11 with + = red). Color range is ±0.003 m/s so it's a pretty significant for just 1 time step.

Probably time to pull out Debugger.jl and step through a time step to find what causes the difference!

display

ali-ramadhan:
Well, Debugger.jl doesn't work in Julia 1.11 either lol.

X-Ref: https://github.com/JuliaDebug/Debugger.jl/issues/361

glwagner:
That is weird and alarming.

glwagner:
@wsmoses @vchuravy do you have any guesses how we could possibly observe a _numerical_ difference between julia 1.10 and 1.11? This does not seem to be a syntax issue but rather a numerical calculation issue. Unless there is a latent syntax issue that has changed functionality which I suppose is also possible...

glwagner:
@ali-ramadhan it might be worth testing the nonhydrostatic model and also problems without immersed boundaries to pinpoint the problem.

wsmoses:
The random number generator doesn’t guarantee the same results even with a seed across versions

glwagner:
One possibility is that syntax changed for something in a subtle way, so the code still runs but some function is being called incorrectly. Not sure what that could be though

glwagner:
Here a list of changes:

https://docs.julialang.org/en/v1/NEWS/#:~:text=New%20language%20features,-New%20Memory%20type&text=Most%20of%20the%20Array%20type,public%20are%20considered%20public%20API.

the change to `zero` stands out the most to me since that is called extensively; then again I don't think the described feature applies to us.

We should check if there are changes only on CPU, or on both CPU and GPU. @ali-ramadhan I'm assuming your test was on CPU.

navidcy:
Good find. Is there a way to redefine/import `zero(::AbstractArray)` to be how it was in Julia v1.10 to check if problem goes away? That would give us a very good hint!

glwagner:
> Good find. Is there a way to redefine/import `zero(::AbstractArray)` to be how it was in Julia v1.10 to check if problem goes away? That would give us a very good hint!

I don't think we use `zero(::AbstractArray)`. I was more wondering if changes related to that update might have affected us (which are not written in the notes).

glwagner:
> I can confirm that Julia 1.11 is changing how the immersed boundary (and free surface?) works. Here's the difference (including halos) in the internal tide u field after 1 time step (difference is Julia 1.10 - Julia 1.11 with + = red). Color range is ±0.003 m/s so it's a pretty significant for just 1 time step.
> 
> Probably time to pull out Debugger.jl and step through a time step to find what causes the difference!
> 
> display

I can't reproduce this

ali-ramadhan:
@glwagner Does the `internal_tide.jl` example still blow up for you with Julia 1.11? And does the Julia 1.10 - Julia 1.11 difference show a different pattern or is there no difference?

I can try to reproduce on a different machine to confirm.

glwagner:
In the test I did, the difference was identically 0. However, I just noticed that I was not using the latest `main`. I was on 0.95.5.

On 0.95.5, it looks like `internal_tide.jl` runs fine. This is what I get

https://github.com/user-attachments/assets/d17e83da-659a-48ec-9bd5-0b647c1ab79d

I have an Mac M1 Max and using 1.11.2:

```julia
julia> versioninfo()
Julia Version 1.11.2
Commit 5e9a32e7af2 (2024-12-01 20:02 UTC)
Build Info:
Platform Info:
  OS: macOS (arm64-apple-darwin24.0.0)
  CPU: 10 × Apple M1 Max
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, apple-m1)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
Environment:
  JULIA_NUM_PRECOMPILE_TASKS = 6
  LD_LIBRARY_PATH = /Users/gregorywagner/Software/hdf5-1.14.2/lib:
  JULIA_EDITOR = vim
```

I'll redo these tests on the current `main` and also try to reproduce on tartarus...

giordano:
> This does not seem to be a syntax issue but rather a numerical calculation issue.

Is there a small-ish reproducer to look at?

simone-silvestri:
Also on my side, on my mac laptop, the `internal_tide.jl` example runs on julia 1.11.2 without crashing. The results are actually identical on julia 1.10.7 and 1.11.2 (on the current main branch, aka version 0.95.7)


https://github.com/user-attachments/assets/21e9ee09-e12d-4df9-a8b8-0504a4b3c706

navidcy:
That sounds good! Let's see if the docs built now!

glwagner:
> > This does not seem to be a syntax issue but rather a numerical calculation issue.
> 
> Is there a small-ish reproducer to look at?

We need to make one and something relatively small should be possible. It looks like its not an issue on Mac so we have to try to reproduce the docs failure on the machine we use for CI. I can try in a few days.

glwagner:
current error @wsmoses https://buildkite.com/clima/oceananigans/builds/20587#0194fc6e-8851-4789-9631-e77fd4c34693/18-817

giordano:
That's fixed by #4093.

glwagner:
Ok, looks like the error has progressed. Here's the CPU enzyme error now:

[oceananigans_build_20621_cpu-enzyme-extension-tests.log](https://github.com/user-attachments/files/18813476/oceananigans_build_20621_cpu-enzyme-extension-tests.log)

navidcy:
I suggest we put a compat entry for Julia 1.10. Otherwise people use Julia v1.11 and run into problems!

navidcy:
@glwagner, note that the ocean mixing example NaNs....

ali-ramadhan:
Since Debugger.jl seems to have started working with Julia 1.11 (https://github.com/JuliaDebug/Debugger.jl/issues/361) I thought I should take another look here.

But with Julia 1.11.6 all the examples seem to run just fine. I built the docs and ran all the examples, and the outputs all look good. So maybe no examples are blowing up anymore?

Looking at a recent-ish test build, if the docs will build now then it's just an Enzyme test failure plus a few test failures (see https://github.com/CliMA/Oceananigans.jl/issues/4663#issuecomment-3234871455) that might be fixed by #4734 (tentatively)?

I'll run the full test suite locally with Julia 1.11 (with and without this potential fix: https://github.com/CliMA/Oceananigans.jl/issues/4663#issuecomment-3234544386) and report back.

glwagner:
> Since Debugger.jl seems to have started working with Julia 1.11 ([JuliaDebug/Debugger.jl#361](https://github.com/JuliaDebug/Debugger.jl/issues/361)) I thought I should take another look here.
> 
> But with Julia 1.11.6 all the examples seem to run just fine. I built the docs and ran all the examples, and the outputs all look good. So maybe no examples are blowing up anymore?
> 
> Looking at a recent-ish test build, if the docs will build now then it's just an Enzyme test failure plus a few test failures (see [#4663 (comment)](https://github.com/CliMA/Oceananigans.jl/issues/4663#issuecomment-3234871455)) that might be fixed by #4734 (tentatively)?
> 
> I'll run the full test suite locally with Julia 1.11 (with and without this potential fix: [#4663 (comment)](https://github.com/CliMA/Oceananigans.jl/issues/4663#issuecomment-3234544386)) and report back.

Are you on ubuntu?  I believe the errors only manifest on ubuntu (possibly only on CPU). I'd expect them to get fixed eventually though.

ali-ramadhan:
> Are you on ubuntu? I believe the errors only manifest on ubuntu (possibly only on CPU). I'd expect them to get fixed eventually though.

I ran the examples locally on Arch Linux / EndeavourOS (where the examples have NaNed before). The tests are currently running on Debian.

Curious why it would be specific to a Linux distro.

glwagner:
> Curious why it would be specific to a Linux distro.

Yes, this is the question of the hour!

simone-silvestri:
Not sure if it is something relevant, but this was giving me a bounds error in julia 1.11
https://github.com/CliMA/Oceananigans.jl/pull/3836/commits/d7393267799beb4e1f6f0d8b4daa5da4f4450be5

ali-ramadhan:
Got pretty far in before `test_distributed_model.jl` complained about the `instantiate` function for some reason which we might need to fix for Julia 1.11. But so far so good I think. Then I realized I only ran the CPU tests lol.

Maybe it will be easier to run tests for Julia 1.10 + 1.11 (CPU + GPU) once CI is back online (hopefully next week!).

```
Oceananigans: Error During Test at /home/alir/atdepth/Oceananigans.jl/test/runtests.jl:20
  LoadError: invalid method definition in Main: function Main.instantiate must be explicitly imported to be extended
  Stacktrace:
    [1] top-level scope
      @ none:0
    [2] top-level scope
      @ ~/atdepth/Oceananigans.jl/test/test_distributed_models.jl:38
    [3] include
      @ ./sysimg.jl:38 [inlined]
    [4] macro expansion
      @ ~/atdepth/Oceananigans.jl/test/runtests.jl:178 [inlined]
    [5] macro expansion
      @ ~/.julia/juliaup/julia-1.11.6+0.x64.linux.gnu/share/julia/stdlib/v1.11/Test/src/Test.jl:1709 [inlined]
    [6] (::var"#11#12")()
      @ Main ~/atdepth/Oceananigans.jl/test/runtests.jl:22
    [7] task_local_storage(body::var"#11#12", key::Symbol, val::GPUArraysCore.ScalarIndexing)
      @ Base ./task.jl:315
    [8] allowscalar(f::Function)
      @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:179
    [9] top-level scope
      @ ~/atdepth/Oceananigans.jl/test/runtests.jl:18
   [10] include(fname::String)
      @ Main ./sysimg.jl:38
   [11] top-level scope
      @ none:6
  in expression starting at /home/alir/atdepth/Oceananigans.jl/test/test_distributed_models.jl:38
Test Summary:                               |    Pass  Error  Broken    Total       Time
Oceananigans                                | 4062298      1      10  4062309  210m22.2s
  Unit tests                                | 4006064                 4006064    5m10.9s
  AbstractOperations and broadcasting tests |     933                     933    2m44.5s
  TripolarGrid tests                        |      46                      46      26.9s
  Poisson Solvers 1                         |     286                     286    4m04.7s
  Poisson Solvers 2                         |     660                     660   20m38.6s
  Matrix Poisson Solvers                    |      31                      31      20.2s
  General Solvers                           |      70                      70       6.2s
  Simulation tests                          |   19101                   19101   10m54.3s
  Lagrangian particle tracking tests        |    1771                    1771    4m36.3s
  Model and time stepping tests (part 1)    |    2992              2     2994   11m28.1s
  Model and time stepping tests (part 2)    |   28606                   28606   22m27.3s
  Model and time stepping tests (part 3)    |     633              1      634   58m05.0s
  Turbulence closures tests                 |     218                     218    3m22.7s
  Shallow Water Models                      |      42                      42    1m27.6s
  HydrostaticFreeSurfaceModel tests         |     254              7      261    8m23.1s
  Multi Region tests                        |     591                     591   55m34.4s
ERROR: LoadError: Some tests did not pass: 4062298 passed, 0 failed, 1 errored, 10 broken
```

ali-ramadhan:
This is worth trying again now that CUDA.jl v5.8.4 is out which should fix some of the GPU failures on Buildkite.

giordano:
Superseded by #4836!

PR #3841, opened 2024-10-09
`static_column_depth` interface and numerical bottom height in `AbstractGridFittedBottom`

simone-silvestri:
This PR introduces an interface for `static_column_depth ` that returns the height of the water column (not including the free surface). 
This is simply `grid.Lz` in regular grids.

For `AbstractGridFittedBottom` immersed grids, the column height is readily read from the immersed boundary, provided that the immersed boundary represents the _numerical_ bottom rather than the _physical_ bottom. 

Therefore, this PR changes the constructor of the  `ImmersedBoundaryGrid` to store the z-coordinate of the last immersed cell. In this way, the bottom height is also uniquely defined for `GridFittedBottom`, removing the necessity of having a`CenterImmersedCondition` and a `InterfaceImmersedCondition`. 

An immediate application of this interface is in the `SplitExplicitFreeSurface` where the bottom_height was previously stored.


simone-silvestri:
Correction: `InterfaceImmersedCondition` and `CenterImmersedCondition` might still useful during construction because they inform where the bottom height will be.

glwagner:
Is `column_height` a good name?

src/ImmersedBoundaries/grid_fitted_bottom.jl
```diff
@@ -17,46 +17,43 @@ abstract type AbstractGridFittedBottom{H} <: AbstractGridFittedBoundary end
...
-struct CenterImmersedCondition end
 struct InterfaceImmersedCondition end
-
-Base.summary(::CenterImmersedCondition) = "CenterImmersedCondition"
-Base.summary(::InterfaceImmersedCondition) = "InterfaceImmersedCondition"
+struct CenterImmersedCondition end
 
 struct GridFittedBottom{H, I} <: AbstractGridFittedBottom{H}
-    bottom_height :: H
+    z_bottom :: H
```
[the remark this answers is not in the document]
simone-silvestri: I though we want to refer to `z` as the coordinate and height as a positive value. 
At least, this is the direction we are taking in ClimaOcean where we are distinguishing very carefully between `z` and `h` or `height`. 
In this case, this is technically a coordinate, not a height, for example we have now
https://github.com/CliMA/Oceananigans.jl/blob/989108a6205bd39f7b215b2145ff5111015610df/src/ImmersedBoundaries/grid_fitted_bottom.jl#L109 which conflates the meaning of z with height. I thought we had decided this over at ClimaOcean and I wanted to update this code to reflect the decision we had made.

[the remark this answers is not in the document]
glwagner: > I though we want to refer to z as the coordinate and height as a positive value.

Usually "z" is called "geopotential height". What would you call "z" instead? Altitude?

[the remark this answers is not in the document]
glwagner: > this is technically a coordinate, not a height

Another name for z is "height coordinate".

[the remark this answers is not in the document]
jagoosw: As someone who hasn't used immersed boundaries that much, I always find it a bit confusing that it's called bottom height because it makes me think I'm meant to put in a positive "height" from the bottom of the domain, but I would know that "z" is meant to be from the top

simone-silvestri:
Shall we merge this as is without changing the name of the `bottom_height` then open an issue were we can discuss the name change?

simone-silvestri:
There was another `total_depth` in the `TurbulenceClosures` module. I think the name `domain_depth` is appropriate (the domain is static) so I unified the two. 

There are also a couple of `depth` functions defined in the `TurbulenceClosures` module that are 3-dimensional. To distinguish from those, I believe `domain_depth` should accept two integers (i, j) instead of 3.

src/MultiRegion/multi_region_field.jl
```diff
@@ -70,6 +70,8 @@ Base.size(f::GriddedMultiRegionField) = size(getregion(f.grid, 1))
 Reconstruct a global field from `mrf::MultiRegionField` on the `CPU`.
 """
 function reconstruct_global_field(mrf::MultiRegionField)
+
+    # TODO: Is this correct? Shall we reconstruct a global field on the architecture of the grid?
```
glwagner, line 74: `arch` should be an argument to `reconstruct_global_field` if we need to do this on CPU for some reason

simone-silvestri: agreed

glwagner:
I see `domain_depth`. What is the proposed interface?

simone-silvestri:
I proposed `domain_depth` to distinguish it from the already existing `depth` interface that tracks the `depth` as a function of `k` which is used in the turbulence closure module. I think `domain_depth` communicates the idea of a total static depth.

glwagner:
Ok we will call it `static_column_depth`

Project.toml
```diff
@@ -79,10 +79,9 @@ julia = "1.9"
...
 DataDeps = "124859b0-ceae-595e-8997-d05f6a7a8dfe"
 Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
-OrthogonalSphericalShellGrids = "c2be9673-fb75-4747-82dc-aa2bb9f4aed0"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 TimesDates = "bdfc003b-8df8-5c39-adcd-3a9087f5df4a"
 
 [targets]
-test = ["DataDeps", "Enzyme", "OrthogonalSphericalShellGrids", "SafeTestsets", "Test", "TimesDates"]
+test = ["DataDeps", "Enzyme", "SafeTestsets", "Test", "TimesDates"]
```
glwagner, line 87: Put this in another PR!!!!!

src/Grids/grid_utils.jl
```diff
@@ -314,6 +314,15 @@ coordinate_summary(topo, Δ::Union{AbstractVector, AbstractMatrix}, name) =
              name, prettysummary(minimum(parent(Δ))),
              name, prettysummary(maximum(parent(Δ))))
 
+#####
+##### Bottom height
```
glwagner, line 318:
```suggestion
##### Static column depth
```

src/ImmersedBoundaries/grid_fitted_bottom.jl
```diff
@@ -37,19 +39,13 @@ Keyword Arguments
...
-                        `immersed_condition` to allow `GridFittedBottom` and
-                        `PartialCellBottom` to have the same behavior when the
-                        minimum fractional cell height for partial cells is set
-                        to 0.
+              bottom in absolute ``z`` coordinates.
+
+* `immersed_condition`: Determine whether the part of the domain that is immersed are all the cell centers that lie below
+        `bottom_height` (`CenterImmersedCondition()`; default) or all the cell faces that lie below `bottom_height` (`InterfaceImmersedCondition()`). 
+        The only purpose of `immersed_condition` to allow `GridFittedBottom` and `PartialCellBottom` to have the same behavior when the
+        minimum fractional cell height for partial cells is set to 0.
```
glwagner, line 47: indenting is messed up. Also why does this change?

src/ImmersedBoundaries/grid_fitted_bottom.jl
```diff
@@ -72,50 +68,79 @@ Base.summary(ib::GridFittedBottom{<:Function}) = @sprintf("GridFittedBottom(%s)"
...
 
-Computes `ib.bottom_height` and wraps it in a Field.
+Computes `ib.bottom_height` and wraps it in a Field. `ib.bottom_height` is the z-coordinate of top-most interface
+of the last ``immersed`` cell in the column.
 """
 function ImmersedBoundaryGrid(grid, ib::GridFittedBottom)
     bottom_field = Field{Center, Center, Nothing}(grid)
     set!(bottom_field, ib.bottom_height)
-    @apply_regionally clamp_bottom_height!(bottom_field, grid)
+    @apply_regionally correct_bottom_height!(bottom_field, grid, ib)
```
glwagner, line 97: I think "correct" is too vague, can we go back to clamp or something more specific?

validation/multi_region/multi_region_near_global_quarter_degree.jl
```diff
@@ -147,7 +147,7 @@ biharmonic_viscosity   = HorizontalScalarBiharmonicDiffusivity(ν=νhb, discrete
 @inline cyclic_interpolate(u₁::Number, u₂, time) = u₁ + mod(time / thirty_days, 1) * (u₂ - u₁)
 
 Δz_top    = @allowscalar Δzᵃᵃᶜ(1, 1, grid.Nz, grid.underlying_grid)
-Δz_bottom = @allowscalar Δzᵃᵃᶜ(1, 1, 1, grid.underlying_grid)
+Δbottom_height = @allowscalar Δzᵃᵃᶜ(1, 1, 1, grid.underlying_grid)
```
glwagner, line 150: I think this was a mistake

src/MultiRegion/multi_region_grid.jl
```diff
@@ -184,6 +184,7 @@ end
 function reconstruct_global_grid(mrg::ImmersedMultiRegionGrid) 
     global_grid     = reconstruct_global_grid(mrg.underlying_grid)
     global_boundary = reconstruct_global_boundary(mrg.immersed_boundary)
+    global_boundary = on_architecture(architecture(mrg), global_boundary)
```
glwagner, line 187:
```suggestion
    global_ib = reconstruct_global_boundary(mrg.immersed_boundary)
    global_ib = on_architecture(architecture(mrg), global_ib)
```

when we leave the word "immersed" from "immersed boundary" things quickly get confusing

src/ImmersedBoundaries/partial_cell_bottom.jl
```diff
@@ -129,15 +138,14 @@ end
...
+    # Get bottom z-coordinate and fractional Δz parameter
+    zb = @inbounds ib.bottom_height[i, j, 1]
 
     # Are we in a bottom cell?
     at_the_bottom = bottom_cell(i, j, k, ibg)
 
-    full_Δz = Δzᶜᶜᶜ(i, j, k, ibg.underlying_grid)
-    partial_Δz = max(ϵ * full_Δz, z - h)
+    full_Δz    = Δzᶜᶜᶜ(i, j, k, ibg.underlying_grid)
+    partial_Δz = z - zb
```
glwagner, line 148: This should go in a different PR.

src/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -55,6 +55,7 @@ using Oceananigans.ImmersedBoundaries: z_bottom
 import Oceananigans.Grids: required_halo_size_x, required_halo_size_y, required_halo_size_z
 import Oceananigans.Architectures: on_architecture
 
+import Oceananigans.ImmersedBoundaries: z_bottom
```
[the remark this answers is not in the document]
simone-silvestri: it's the new rearrangement of the modules.

src/ImmersedBoundaries/partial_cell_bottom.jl
simone-silvestri: The new numerical bottom height is located at by convention at `z⁻ + Δz * (1 - ϵ)` so `ϵ * full_Δz` == `z - h` removing the necessity of the max. I thought this change can clean up the partial cell code

src/TurbulenceClosures/TurbulenceClosures.jl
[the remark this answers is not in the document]
simone-silvestri: I have changed z_bottom to be only in `TurbulenceClosures` as it is the only module in which is needed / used and all similar methods (`z_top`, `depth`, `height_above_bottom`...) are defined only there

simone-silvestri:
The dependency on OrthogonalSphericalShellGrids here is blocking the progress (the `SplitExplicitAuxiliaryFields` type has changed in this PR). How do we deal with this? We could remove the dependency in the tests of Oceananigans and move those tests over at OrthogonalSphericalShellGrids

src/ImmersedBoundaries/partial_cell_bottom.jl
simone-silvestri: actually, there is a bug here with how the immersed bottom height is computed. I ll correct it

simone-silvestri:
> The dependency on OrthogonalSphericalShellGrids here is blocking the progress (the `SplitExplicitAuxiliaryFields` type has changed in this PR). How do we deal with this? We could remove the dependency in the tests of Oceananigans and move those tests over at OrthogonalSphericalShellGrids

Shall we do this? In case we want to do it I have prepared https://github.com/CliMA/OrthogonalSphericalShellGrids.jl/pull/46.
If yes, we can merge that one then this one, then update OrthogonalSphericalShellGrids with the new interface

simone-silvestri:
This PR should be ready to merge

simone-silvestri:
This PR requires an approval

src/ImmersedBoundaries/ImmersedBoundaries.jl
```diff
@@ -92,10 +95,6 @@ with_halo(halo, ibg::ImmersedBoundaryGrid) =
 inflate_halo_size_one_dimension(req_H, old_H, _, ::IBG)            = max(req_H + 1, old_H)
 inflate_halo_size_one_dimension(req_H, old_H, ::Type{Flat}, ::IBG) = 0
 
-# Defining the bottom
-@inline z_bottom(i, j, grid) = znode(i, j, 1, grid, c, c, f)
-@inline z_bottom(i, j, ibg::IBG) = error("The function `bottom` has not been defined for $(summary(ibg))!")
```
glwagner, line 97: isn't this still useful? Why delete it?

src/ImmersedBoundaries/grid_fitted_bottom.jl
```diff
@@ -72,50 +73,77 @@ Base.summary(ib::GridFittedBottom{<:Function}) = @sprintf("GridFittedBottom(%s)"
...
+@inline static_column_depthᶠᶜᵃ(i, j, ibg::AGFBIBG) = min(static_column_depthᶜᶜᵃ(i-1, j, ibg), static_column_depthᶜᶜᵃ(i, j, ibg))
+@inline static_column_depthᶠᶠᵃ(i, j, ibg::AGFBIBG) = min(static_column_depthᶠᶜᵃ(i, j-1, ibg), static_column_depthᶠᶜᵃ(i, j, ibg))
+
+# Make sure column_height works for horizontally-Flat topologies.
+XFlatAGFIBG = ImmersedBoundaryGrid{<:Any, <:Flat, <:Any, <:Any, <:Any, <:AbstractGridFittedBottom}
+YFlatAGFIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Flat, <:Any, <:Any, <:AbstractGridFittedBottom}
 
-Adapt.adapt_structure(to, ib::GridFittedBottom) = GridFittedBottom(adapt(to, ib.bottom_height),
-                                                                             ib.immersed_condition)
+@inline static_column_depthᶠᶜᵃ(i, j, ibg::XFlatAGFIBG) = static_column_depthᶜᶜᵃ(i, j, ibg)
```
glwagner, line 146: when this happens it could suggest that this should be more like `node` with

```julia
static_column_depth(i, j, grid, lx, ly)
```

src/ImmersedBoundaries/partial_cell_bottom.jl
glwagner: i just dont understand how this is related to a PR that implements a new function called `static_column_depth`

glwagner:
I'm confused about this PR. There are all sorts of changes in different places. There is even a change to the Makie extension. What am I reviewing?

src/ImmersedBoundaries/ImmersedBoundaries.jl
simone-silvestri: I have moved it in the `TurbulenceClosures` module where also `z_top` and `height_above_bottom` are defined

simone-silvestri:
I have merged in #3880 because I wanted to see if the GPU tests pass. I am planning to merge this PR after that one. The rest is all from this PR. The OceananigansMakieExt change is a typo fix, but I see you have added a PR for it.

src/ImmersedBoundaries/ImmersedBoundaries.jl
glwagner: Ok so its unrelated to the static column depth? I feel z_bottom belongs in the Grids module. What do you think?

glwagner: Consider that it has to be extended for different immersed boundary implementation so it depends on partial cell bottom, etc. 

The original implementation may not have been optimal, but since we are moving things around here, I think we should try to improve on the original organization.

src/TurbulenceClosures/TurbulenceClosures.jl
```diff
@@ -120,14 +121,21 @@ end
...
+
 @inline z_top(i, j, grid) = znode(i, j, grid.Nz+1, grid, c, c, f)
+@inline z_bottom(i, j, grid) = znode(i, j, 1, grid, c, c, f)
+@inline z_bottom(i, j, ibg::AGFBIBG) = @inbounds ibg.immersed_boundary.bottom_height[i, j, 1]
 
-@inline depthᶜᶜᶠ(i, j, k, grid)    = clip(z_top(i, j, grid) - znode(i, j, k, grid, c, c, f))
-@inline depthᶜᶜᶜ(i, j, k, grid)    = clip(z_top(i, j, grid) - znode(i, j, k, grid, c, c, c))
-@inline total_depthᶜᶜᵃ(i, j, grid) = clip(z_top(i, j, grid) - z_bottom(i, j, grid))
+@inline depthᶜᶜᶠ(i, j, k, grid) = clip(z_top(i, j, grid) - znode(i, j, k, grid, c, c, f))
+@inline depthᶜᶜᶜ(i, j, k, grid) = clip(z_top(i, j, grid) - znode(i, j, k, grid, c, c, c))
```
glwagner, line 138: is this used?

simone-silvestri: seems to be used in CATKE here for example https://github.com/CliMA/Oceananigans.jl/blob/6c40d7e225c2127051b2703b9c62a8b18260e3a5/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_mixing_length.jl#L62

src/ImmersedBoundaries/ImmersedBoundaries.jl
simone-silvestri: I am ok with moving all these functions in the Grids module, shall we do it in this PR?

src/ImmersedBoundaries/partial_cell_bottom.jl
```diff
@@ -63,13 +63,40 @@ end
...
+    for k in 1:grid.Nz
+        z⁻ = znode(i, j, k,   grid, c, c, f)
+        z⁺ = znode(i, j, k+1, grid, c, c, f)
+        Δz = Δzᶜᶜᶜ(i, j, k, grid)
+        bottom_cell = (z⁻ ≤ zb) & (z⁺ ≥ zb)
+
+        # If the size of the bottom cell is less than ϵ Δz, use 
+        # a simple `GridFittedBottom` approach where the bottom 
+        # height is the top interface of cell k.
+        capped_zb   = ifelse(zb < z⁻ + Δz * (1 - ϵ), zb, z⁺)
```
glwagner, line 94:
```suggestion
        partial_Δz = z⁺ - zb
        too_thin = partial_Δz < ϵ * Δz
        numerical_zb  = ifelse(too_thin, zb, z⁺)
```

src/ImmersedBoundaries/partial_cell_bottom.jl
```diff
@@ -63,13 +63,40 @@ end
...
+
+        @inbounds bottom_field[i, j, 1] = ifelse(bottom_cell, capped_zb, bottom_field[i, j, 1])
+    end
```
glwagner, line 97:
```suggestion
    end
    @inbounds bottom_field[i, j, 1] = numerical_zb
```

glwagner: should be outside the `k`-loop

src/ImmersedBoundaries/partial_cell_bottom.jl
```diff
@@ -63,13 +63,40 @@ end
...
+    ϵ  = ib.minimum_fractional_cell_height
+    for k in 1:grid.Nz
+        z⁻ = znode(i, j, k,   grid, c, c, f)
+        z⁺ = znode(i, j, k+1, grid, c, c, f)
+        Δz = Δzᶜᶜᶜ(i, j, k, grid)
+        bottom_cell = (z⁻ ≤ zb) & (z⁺ ≥ zb)
+
+        # If the size of the bottom cell is less than ϵ Δz, use 
+        # a simple `GridFittedBottom` approach where the bottom 
+        # height is the top interface of cell k.
```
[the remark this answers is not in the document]
glwagner: Which is a fine proposal but I think you should change the title of the PR. Imagine us looking for what is causing this kind of different in the future, combing through PRs. It's pretty hidden in this one

src/ImmersedBoundaries/partial_cell_bottom.jl
glwagner: this just makes it easier to understand. I was having trouble with the math. Also I think it's convoluted to evaluate the criterion by using the interface _below_ and then adding `Δz` to both sides, isn't it?

src/ImmersedBoundaries/partial_cell_bottom.jl
glwagner: and there's no need for an if statement

src/ImmersedBoundaries/partial_cell_bottom.jl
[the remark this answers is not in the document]
simone-silvestri: Hmmm, I thought there was supposed to be no difference. 
Here, the `bottom_height` is set to the actual bottom height if it is smaller than `z⁻ + ϵ Δz`, otherwise the cell is closed. 
So, the only difference with main will be for bottom heights that are in between `z⁻ + ϵ Δz` and `z⁺` (where `z⁻` and `z⁺` are the interfaces of the bottom cell.
I will check the difference in the internal_tide example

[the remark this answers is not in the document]
glwagner: But I thought before we would keep the cell open (rather than closing it).

Actually I think that MITgcm uses some criteria to decide which one to do --- either snapping up or down depending on which change is least.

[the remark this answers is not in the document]
simone-silvestri: ah, ok, I understand; open till `ϵ Δz` instead of snapping up.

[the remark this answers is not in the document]
glwagner: Actually I think we used to use that criteria too. It's something like 

```julia
partial_Δz = z⁺ - zb
too_thin = partial_Δz < ϵ * Δz
snap_up = partial_Δz < ϵ * Δz / 2
snapped_zb = ifelse(snap_up, z⁺, z⁻ + ϵ * Δz)
numerical_zb  = ifelse(too_thin, snapped_zb, zb)
```

simone-silvestri:
In the new commit, I reverted to the previous implementation, so now there is no difference

https://github.com/user-attachments/assets/761fd5b0-e54c-40f6-b924-ab23ed69c17f

glwagner (approved):

src/ImmersedBoundaries/partial_cell_bottom.jl
simone-silvestri: right, I have modified it

Issue #3831, opened 2024-10-07
Immediate NaNs running a non-hydrostatic model with an immersed boundary and open boundary conditions and immersed pressure solver


ali-ramadhan:
I'm trying to set up a small test where a zonal velocity forced by open boundary conditions goes around an immersed sea mount, using the new immersed pressure solver.

Unfortunately I get immediate NaNs on iteration 1. Maybe there's a bug somewhere? I know I'm using a bunch of _experimental_ features together so perhaps this is not surprising. My setup could also be bad though.

Curious if anyone has any insights on what's going wrong here.

Does the pressure solver need to be modified to account for non-zero velocities at the boundaries? I guess the FFT pressure solver assumes either periodic or no-penetration at the boundaries, but then shouldn't the conjugate gradient solver converge on the correct pressure with enough iterations? Or maybe not if the pre-conditioner is very wrong?

"MWE" setup:

```julia
using Printf
using Oceananigans
using Oceananigans.Solvers: ConjugateGradientPoissonSolver, fft_poisson_solver

L = 100
H = 100

underlying_grid = RectilinearGrid(
    CPU(),
    Float64,
    topology = (Bounded, Bounded, Bounded),
    size = (16, 16, 16),
    x = (0, L),
    y = (0, L),
    z = (-H, 0)
)

h = H/2
w = L/5
mount(x, y) = h * exp(-x^2 / 2w^2) * exp(-y^2 / 2w^2)
bottom(x, y) = -H + mount(x, y)

grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom))

@inline u_inflow(y, z, t) = 0.01

u_bcs = FieldBoundaryConditions(
    west = OpenBoundaryCondition(u_inflow),
    east = OpenBoundaryCondition(u_inflow)
)

boundary_conditions = (; u=u_bcs)

model = NonhydrostaticModel(;
    grid,
    boundary_conditions,
    timestepper = :RungeKutta3,
    pressure_solver = ConjugateGradientPoissonSolver(
        grid;
        preconditioner = fft_poisson_solver(grid.underlying_grid)
    )
)

simulation = Simulation(model; Δt=0.1, stop_time=60)

progress(sim) = @printf(
    "iteration: %d, time: %.4f, U_max=(%.2e, %.2e, %.2e)\n",
    iteration(simulation),
    time(simulation),
    maximum(abs, model.velocities.u),
    maximum(abs, model.velocities.v),
[...]

PR #3846, opened 2024-10-10
Solves bug when calling `ConjugateGradientPoissonSolver` without specifying preconditioner

tomchor:
See https://github.com/CliMA/Oceananigans.jl/issues/3831#issuecomment-2406235013

Solves https://github.com/CliMA/Oceananigans.jl/issues/3829


src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -1,4 +1,5 @@
 using Oceananigans.Operators: divᶜᶜᶜ, ∇²ᶜᶜᶜ 
+using Oceananigans.Grids: ImmersedBoundaryGrid
```
glwagner, line 2: Hmm, I think `ImmersedBoundaryGrid` is defined by `ImmersedBoundaries`, which isn't imported until after `Solvers`:

https://github.com/CliMA/Oceananigans.jl/blob/99d3c1fcb8cb7951c0b82f82d9d61d2fe9d541e1/src/Oceananigans.jl#L210-L227

A bit of refactoring is needed to import `ImmersedBoundaries` before `Solvers`. So I think if we want to resolve this without all that refactoring we should define the constructor that depends on `ImmersedBoundaryGrid` separately (either in `ImmersedBoundaries` or `Model.NonhydrostaticModels`)

glwagner:
I can fix this if you want

src/Solvers/conjugate_gradient_poisson_solver.jl
tomchor: yeah I think you're right!

tomchor:
> I can fix this if you want

Thanks, yeah feel free to jump in!

ali-ramadhan:
This might be a duplicate of PR #3830 but looks like both will be superseded by PR #3847 which is doing the refactoring necessary for this fix.

ali-ramadhan (approved):
Thanks @tomchor!

PS: You can mark this PR as resolving https://github.com/CliMA/Oceananigans.jl/issues/3829

tomchor:
> Thanks @tomchor!

I mean, the real hero was @simone-silvestri with https://github.com/CliMA/Oceananigans.jl/pull/3847 :)

> PS: You can mark this PR as resolving #3829

Just did!

PR #3646, opened 2024-07-01
Include background tracer and velocities in closure tendency terms

glwagner:
cc @hdrake @liuchihl

PR #3862, opened 2024-10-23
Correct accessing velocities from the background field

liuchihl:
https://github.com/CliMA/Oceananigans.jl/pull/3646#issuecomment-2433267716 @glwagner implementing the fix suggested by Greg


hdrake:
@glwagner, is that what you were intending? We're still learning the ropes on collaborative development via git here...

glwagner (approved):

glwagner:
Yes! I was ok to put this in though. Just wanted to see the test first, it's like eating vegetables before dessert!

glwagner:
I also forgot that the branch was mine and that you can only change it through a PR

hdrake:
@glwagner, I think there was a misunderstanding about the test failing. For the background-flux-divergence PR, we were just confused as to why it wasn't passing the existing test suite.

The separate issue of adding a new test is for the [unrelated PR](https://github.com/CliMA/Oceananigans.jl/pull/3721) related to refactoring `AveragedTimeInterval` to fix the existing bug. @liuchihl has some local branches where he'd merged both of those and was testing them in combination, so we got our signals crossed.

PR #3864, opened 2024-10-23
Add more advection scheme tests; fix order adapting with tracer-specific schemes; fix WENOVectorInvariant

glwagner:
Closes #3863 (+ fixes FluxFormAdvection docstring)
Closes #3865

PR #3866, opened 2024-10-24
Fix docs for `VectorInvariant` and `WENOVectorInvariant`

navidcy:
x-ref: https://github.com/CliMA/Oceananigans.jl/pull/3864#issuecomment-2434274960


src/Advection/vector_invariant_advection.jl
```diff
@@ -175,10 +175,19 @@ Base.show(io::IO, a::VectorInvariant{N, FT}) where {N, FT} =
...
+                        order = nothing,
+                        vorticity_order = nothing,
+                        vertical_order = nothing,
+                        divergence_order = nothing,
+                        kinetic_energy_gradient_order = nothing, 
+                        multi_dimensional_stencil = false,
+                        weno_kw...)
+
+Return a vector-invariant weighted essentially non-oscillatory (WENO) scheme.
+See [`VectorInvariant`](@ref) and [`WENO`](@ref) for kwargs definitions.
```
glwagner, line 190: @simone-silvestri what does `multi_dimensional_stencil` do?

ali-ramadhan: When I asked about it in https://github.com/CliMA/Oceananigans.jl/discussions/3856 he said that
>  multi-dimensional stencils implement a 2D horizontal stencil instead of a 1D stencil that performs a centered 5th-order WENO reconstruction of vorticity, divergence and kinetic energy in the tangential direction. In theory, it is more accurate, but it is very computationally expensive and doesn't seem to improve the results much, so I haven't found a good use for this reconstruction.

navidcy:
OK. Should we merge this, it's definitely an improvement. But I'm happy to hear more suggestions or if anyone else wants to edit please go ahead!

src/Advection/vector_invariant_advection.jl
```diff
@@ -175,9 +175,24 @@ Base.show(io::IO, a::VectorInvariant{N, FT}) where {N, FT} =
...
+
+If `multi_dimensional_stencil = true` is selected, then a 2D horizontal stencil
+is implemented for the WENO scheme (instead of a 1D stencil). This 2D horizontal
+stencil performs a centered 5th-order WENO reconstruction of vorticity,
+divergence and kinetic energy in the tangential direction.
```
simone-silvestri, line 195:
```suggestion
divergence and kinetic energy in the horizontal direction tangential to the upwind direction.
```

ali-ramadhan (approved):
Looks good! I know @simone-silvestri has this to say about the `upwinding` argument

>  The upwinding keyword argument is a bit of a vestigial argument because it allows two different forms of unwinding, one that ensures energy dissipation locally and one that might lead to anti-dissipative behavior locally (and thus might be unstable). There might be arguments to remove that option since I haven't really found a practical use for it.

but not sure what's worth including.

simone-silvestri:
Right, while we're at it. The docstring of `VectorInvariant` is wrong. 
https://github.com/CliMA/Oceananigans.jl/blob/6c40d7e225c2127051b2703b9c62a8b18260e3a5/src/Advection/vector_invariant_advection.jl#L65-L68
The `VelocityUpwinding` option was removed in #3659

navidcy:
Done.

navidcy:
@wsmoses any idea why this appeared?

https://buildkite.com/clima/oceananigans/builds/18550#0192eaa0-74f7-424b-a987-f65a73e6b3cc/16-591

an update of a package?

wsmoses:
@navidcy I think https://github.com/EnzymeAD/Enzyme.jl/pull/2049 ought to fix it

wsmoses:
Okay hopefully this should be resolved by last nights patch release

PR #3143, opened 2023-06-13
(0.94.0) Overhaul spacings functions to return `KernelFunctionOperation`s

tomchor:
The spacings functions aren't working for `ImmersedBoundaryGrid`s on `main`. This PR fixes that.

On main:

```julia
julia> using Oceananigans

julia> grid_base = RectilinearGrid(size=(4, 5, 6), extent=(1, 1, 1));

julia> ibg = ImmersedBoundaryGrid(grid_base, GridFittedBottom((x, y) -> 1))
4×5×6 ImmersedBoundaryGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo:
├── immersed_boundary: GridFittedBottom(min(h)=1.00e+00, max(h)=1.00e+00)
├── underlying_grid: 4×5×6 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.25
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=0.2
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=0.166667

julia> xspacings(ibg, Center())
ERROR: MethodError: no method matching xspacings(::ImmersedBoundaryGrid{Float64, Periodic, Periodic, Bounded, RectilinearGrid{Float64, Periodic, Periodic, Bounded, Float64, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, CPU}, GridFittedBottom{OffsetArrays.OffsetMatrix{Float64, Matrix{Float64}}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, CPU}, ::Center)
Closest candidates are:
  xspacings(::Any, ::Any, ::Any, ::Any; with_halos) at ~/repos/Oceananigans.jl/src/Grids/grid_utils.jl:363
[...]

PR #3872, opened 2024-10-26
(0.94.0) Refactor exported names

glwagner:
This PR refactors the names we export publically. I think that many names have been added that are poorly motivated --- for example, MultiRegionGrid and cubed sphere grid are _very_ experimental yet still exported.

This PR mostly removes exports. Most of the removals are motivated by one of two reasons: (i) the feature is experimental or (ii) the feature is not commonly used. For (i) we would like to avoid export because as an experimental feature we would like to fix bugs or change the API without incurring an "officially breaking change". For (ii), the purpose is merely to avoid over-polluting the namespace. For example, `XPartition` for `MultiRegionGrid` is very easily confused with `Partition` for `DistributedComputations` (the API here should probably be reconsidered, cc @simone-silvestri). More than once I have lost cluster time because I confused `YPartition(N)` with `Partition(y=N)`. But `YPartition` isn't really useful so there's little benefit to exporting this.

Of course these changes are up for debate so I welcome comments.

More detail:

* Remove `OceananigansLogger`: rarely used
* Add `PartialCellBottom`: we are using it in an example, indicating that it is ready for use (not necessarily completely bug-free but we would like to bring it into regular / default use now)
* Remove `WENOThirdOrder`, `WENOFifthOrder`: the syntax `WENO(order=5)` is preferred (same for `Centered` and `UpwindBiased`)
* Remove `location`: this is a low-level function that does not need to be part of the API
* Add `CATKEVerticalDiffusivity`: this is ready for use, a publication is imminent.
* Remove `StateChecker`: not used as far as I know.
* Remove multi-region and cubed sphere functionality: these are experimental.
* Remove `apply_regionally!, construct_regionally, @apply_regionally, MultiRegionObject`: experimental.
* Remove `Time`: too low-level.


ali-ramadhan:
Good idea! Does this mean tagging v0.94.0 (and one stop closer to v0.100) as it's a change to the public interface?

> Remove `OceananigansLogger`: rarely used

I use it in almost every script I write haha, but that's because I wrote it to my liking. There's a strong argument to be made that a logger configuration does not belong in Oceananigans.jl. I feel it's similar to Oceananigans' decision to not provide a default progress callback seeing as each user will probably want a different one.

> Remove `StateChecker`: not used as far as I know.

Yeah I think I used that to debug cubed sphere stuff. Could easily be defined in a validation script though.

navidcy (approved):
I like the cleanup.
Probably a breaking change tho.

glwagner:
> Good idea! Does this mean tagging v0.94.0 (and one stop closer to v0.100) as it's a change to the public interface?

It's a very breaking change!

src/Fields/Fields.jl
```diff
@@ -15,6 +15,12 @@ using Oceananigans.BoundaryConditions
 using Oceananigans.Utils
 
 import Oceananigans.Architectures: on_architecture
+import Oceananigans: location, instantiated_location
+
+"Returns the location `(LX, LY, LZ)` of an `AbstractField{LX, LY, LZ}`."
```
navidcy, line 20:
```suggestion
"Return the location `(LX, LY, LZ)` of an `AbstractField{LX, LY, LZ}`."
```

glwagner:
PS when this is merged we can probably combined with [#immersedmetrics](https://github.com/CliMA/Oceananigans.jl/pull/3143) before registering 0.94

src/MultiRegion/multi_region_grid.jl
```diff
@@ -93,11 +95,11 @@ julia> grid = RectilinearGrid(size=(12, 12), extent=(1, 1), topology=(Bounded, B
...
+├── connectivity: Oceananigans.Utils.MultiRegionObject{Tuple{@NamedTuple{west::Nothing, east::Oceananigans.MultiRegion.RegionalConnectivity{Oceananigans.MultiRegion.East, Oceananigans.MultiRegion.West}, north::Nothing, south::Nothing}, @NamedTuple{west::Oceananigans.MultiRegion.RegionalConnectivity{Oceananigans.MultiRegion.West, Oceananigans.MultiRegion.East}, east::Oceananigans.MultiRegion.RegionalConnectivity{Oceananigans.MultiRegion.East, Oceananigans.MultiRegion.West}, north::Nothing, south::Nothing}, @NamedTuple{west::Oceananigans.MultiRegion.RegionalConnectivity{Oceananigans.MultiRegion.West, Oceananigans.MultiRegion.East}, east::Oceananigans.MultiRegion.RegionalConnectivity{Oceananigans.MultiRegion.East, Oceananigans.MultiRegion.West}, north::Nothing, south::Nothing}, @NamedTuple{west::Oceananigans.MultiRegion.RegionalConnectivity{Oceananigans.MultiRegion.West, Oceananigans.MultiRegion.East}, east::Nothing, north::Nothing, south::Nothing}}, NTuple{4, CPU}}
```
glwagner, line 102: I don't think showing this is useful at all

glwagner: I can't read it. Can we just hide it? This is not important either.

navidcy: True, OK! I added it via the `doctest(Oceananigans, fix=true)`.

navidcy:
merging this, but we should wait for #3143 before tagging v0.94

PR #3876, opened 2024-10-28
(0.93.2) Update Adapt.jl compat and fix `Float32` CATKE on GPU

ali-ramadhan:
Just opening this PR with what is currently a band-aid fix for `Float32` CATKE on GPUs. Hoping I can figure out what's actually wrong. But for now it'll be useful to have a working branch.

Resolves https://github.com/CliMA/Oceananigans.jl/issues/3870 (eventually, hopefully)


src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
```diff
@@ -258,33 +259,33 @@ end
...
 end
 
-@inline function κuᶜᶜᶠ(i, j, k, grid, closure, velocities, tracers, buoyancy, surface_buoyancy_flux)
+@inline function κuᶜᶜᶠ(i, j, k, grid::AbstractGrid{FT}, closure, velocities, tracers, buoyancy, surface_buoyancy_flux) where FT
     w★ = ℑzᵃᵃᶠ(i, j, k, grid, turbulent_velocityᶜᶜᶜ, closure, tracers.e)
     ℓu = momentum_mixing_lengthᶜᶜᶠ(i, j, k, grid, closure, velocities, tracers, buoyancy, surface_buoyancy_flux)
     κu = ℓu * w★
     κu_max = closure.maximum_viscosity
-    return min(κu, κu_max)
+    return convert(FT, min(κu, κu_max))
```
glwagner, line 267:
```suggestion
    FT = eltype(grid)
    return convert(FT, min(κu, κu_max))
```

(following YASGuide --- type parameters to be used only for dispatch)

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
```diff
@@ -258,33 +259,33 @@ end
     end
 end
 
-@inline function κuᶜᶜᶠ(i, j, k, grid, closure, velocities, tracers, buoyancy, surface_buoyancy_flux)
+@inline function κuᶜᶜᶠ(i, j, k, grid::AbstractGrid{FT}, closure, velocities, tracers, buoyancy, surface_buoyancy_flux) where FT
```
glwagner, line 262:
```suggestion
@inline function κuᶜᶜᶠ(i, j, k, grid, closure, velocities, tracers, buoyancy, surface_buoyancy_flux)
```

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
glwagner: Aslo better to split this into two lines rather than convert and perform `min` on the same line.

glwagner:
Just some minor comments:

1. Use `eltype(grid)` instead of a type parameter, following YASGuide
2. Does the annotation `::FT` work?

ali-ramadhan:
Thanks for the review @glwagner. I made the changes and also added a test that fails (well CUDA crashes) without this PR, and passes with this PR.

> Does the annotation `::FT` work?

Unfortunately not. I ended up getting GPU exceptions instead: https://github.com/CliMA/Oceananigans.jl/issues/3870#issuecomment-2441915837

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
```diff
@@ -228,7 +227,7 @@ end
     Jᵇᵋ = closure.minimum_convective_buoyancy_flux
     Jᵇᵢⱼ = @inbounds Jᵇ[i, j, 1]
     Jᵇ⁺ = max(Jᵇᵋ, Jᵇᵢⱼ, Jᵇ★) # selects fastest (dominant) time-scale
-    t★ = (ℓᴰ^2 / Jᵇ⁺)^(1/3)
+    t★ = cbrt(ℓᴰ^2 / Jᵇ⁺)
```
glwagner, line 230: Can you check if this fixes the issue without the need for extra `convert`?

It will be good to avoid "over converting", because this could cause us to fail to catch spurious promotion which will hurt performance (eg removing some of the benefit of using `Float32`)

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
```diff
@@ -263,28 +262,34 @@ end
     ℓu = momentum_mixing_lengthᶜᶜᶠ(i, j, k, grid, closure, velocities, tracers, buoyancy, surface_buoyancy_flux)
     κu = ℓu * w★
     κu_max = closure.maximum_viscosity
-    return min(κu, κu_max)
+    κu★ = min(κu, κu_max)
+    FT = eltype(grid)
+    return convert(FT, κu★)
 end
```
glwagner, line 268: For this PR I suggest changing this to a a type annotation (`κu★::FT`) since this is what we will want in the long run

glwagner:
I think to preserve the work in this PR, we should add a `Float32` test which will fail if a spurious promotion undermines performance

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
ali-ramadhan: I [checked](https://github.com/CliMA/Oceananigans.jl/issues/3870#issuecomment-2444956478) but it was not enough unfortunately.

Agree that it would be good to not over convert.

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
ali-ramadhan: Unfortunately that does not work, but maybe I can pinpoint where the conversion is happening then we won't need any conversion here.

ali-ramadhan:
> I think to preserve the work in this PR, we should add a `Float32` test which will fail if a spurious promotion undermines performance

Agreed. I'll revisit this PR later to see if I can find where the conversion happens. The test I added only checks to see if we can take a time step. But I should be able to also add a test to ensure no spurious promotion occurred.

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
glwagner: Sorry I am not suggesting this as a solution, but rather as a way to catch a bug in the future.

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
glwagner: The point is just that you may lose much of the advantage of using Float32 in the first place if you throw `convert` around

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
ali-ramadhan: Ah I don't think I actually know how type annotations work. So if we say
```julia
return κu★::FT
```
and `κu★` is not of type `FT` then an error/exception will be thrown?

glwagner: I think so, isn't that what you found?

glwagner: It's a bug checking mechanism which may be more broadly useful as we try to prevent spurious promotion

ali-ramadhan: No idea why I thought it was a different mechanism to convert types lol. But yes that is what I found.

glwagner:
> > I think to preserve the work in this PR, we should add a `Float32` test which will fail if a spurious promotion undermines performance
> 
> Agreed. I'll revisit this PR later to see if I can find where the conversion happens. The test I added only checks to see if we can take a time step. But I should be able to also add a test to ensure no spurious promotion occurred.

Ah, that will work as a test if we remove the `convert`

ali-ramadhan:
Following https://github.com/CliMA/Oceananigans.jl/issues/3870#issuecomment-2445532635 this PR now just changes how grid coordinate ranges are constructed. Curious to see if any tests fail. But locally it fixed CATKE + `Float32`.

@glwagner I ended up doing this if it looks okay:

```julia
    κu★ = min(κu, κu_max)
    FT = eltype(grid)
    return κu★::FT
```

I assume there's a small cost associated with the type annotation `::FT`?

src/Grids/grid_generation.jl
```diff
@@ -113,6 +113,10 @@ function generate_coordinate(FT, topo::AT, N, H, node_interval::Tuple{<:Number,
     F = range(FT(F₋), FT(F₊), length = TF)
     C = range(FT(C₋), FT(C₊), length = TC)
 
+    # See: https://github.com/CliMA/Oceananigans.jl/issues/3870#issuecomment-2445532635
+    F = StepRangeLen{FT, FT, FT, Int}(F)
+    C = StepRangeLen{FT, FT, FT, Int}(C)
```
glwagner, line 118: I don't think it is correct to describe this change as "making the grid coordinate ranges type-safe".

Perhaps more conservatively, we only require the _first_ type parameter of `StepRangeLen` to be `FT`. The second two should be twice precision.

Now, if we want to support non-standard ranges, we can perhaps consider that. But I suspect there is something going on that isn't completely understood here.

glwagner: This is discussed on #3870, but just to express the problem here: we _expect_ the ranges to have the type `StepRangeLen{Float32, Float64, Float64, Int}` for a `Float32` grid. Therefore, this PR does some damage by producing an unexpected range type.

It is unexpected or a bug that a range of type `StepRangeLen{Float32, Float64, Float64, Int}` would produce `Float64`. Therefore the first course of action is to _verify_ that `StepRangeLen{Float32, Float64, Float64, Int}` is producing `Float64` --- on either CPU or GPU.

I have a hunch that `StepRangeLen` is somehow adapted for GPU incorrectly. Or, if it is intentional, then I think we should fix the output of `xnode`, `ynode`, `znode`...

If there's good motivation for reducing the precision of ranges (either in addition or alternatively to the above suggestion) then I think we can entertain it. I'm not sure we want to hardcode this change though, it might be better to provide it as an option. The imprecision of ranges is already a bit annoying.

ali-ramadhan: Since this makes the test pass and the MWE in #3870 not error, maybe the hint is that `StepRangeLen` just needs to be adapted for the GPU?

But yeah I'm not sure why `StepRangeLen{Float32, Float64, Float64, Int}` seems to produce `Float64` numbers in GPU kernels. Perhaps this is worth opening an issue on CUDA.jl with a MWE?

glwagner: Yes, though it is not for us to adapt (that would be type piracy) because we own neither `Adapt` nor `StepRangeLen`. Plus it seems to be adapted, so what is happening...

https://github.com/JuliaGPU/Adapt.jl/blob/5ef7c5329609df7ffb5b19942d6747b3dcc162c2/src/base.jl#L79-L80

glwagner:
This may help https://github.com/JuliaGPU/Adapt.jl/pull/88

ali-ramadhan:
I just changed the Adapt.jl compat entry to make use of the new version with the `StepRangeLen` fix. The MWE from #3870 does not error with Adapt.jl v4.1.1 locally.

Project.toml
```diff
@@ -43,7 +43,7 @@ OceananigansEnzymeExt = "Enzyme"
 OceananigansMakieExt = ["MakieCore", "Makie"]
 
 [compat]
-Adapt = "3, 4"
+Adapt = "^4.1.1"
```
navidcy, line 46:
```suggestion
Adapt = "4.1.1"
```

navidcy (approved):
this is heroic debugging!!

navidcy:
you can forget my minor suggestion to remove `^`! let's merge this!

ali-ramadhan:
Haha it did take a while but with a satisfying ending!

And thanks for the suggestion! I didn't realize that `4.1.1` and `^4.1.1` would be the same here. But since it's okay with you, I'll merge to avoid waiting on another round of tests to pass :upside_down_face:

navidcy:
> to avoid waiting on another round of tests to pass 🙃

Exactly! Takes for ever...!

PR #3877, opened 2024-10-28
Variance dissipation computation

simone-silvestri:
This PR introduces a metric to compute the implicit dissipation introduced by flux form advection schemes.
This metric has been devised by @jm-c and is still subject of some tests + writeup of the method, so this PR is still very much a work in progress.


glwagner:
How about TracerVarianceDissipation, omitting Computation?

tomchor:
@simone-silvestri this is great! Looking forward to trying this out!

One quick question: since this is a diagnostic quantity, would it make sense to include this on Oceanostics instead of Oceananigans?

src/Models/dissipation_computation.jl
```diff
@@ -0,0 +1,207 @@
...
+
+    previous_state = merge(cⁿ⁻¹, (; Uⁿ⁻¹, Uⁿ))
+    advective_fluxes = (; Fⁿ, Fⁿ⁻¹)
+
+    return TracerVarianceDissipation(P, advective_fluxes, previous_state)
+end
+
+# Function to call in a callback
+# Note: This works only if the callback is called with an IterationInterval(1), if not the
+# previous fluxes and velocities will not be correct
```
glwagner, line 55: You can actually enforce this by extending the `Callback` constructor

src/Models/Models.jl
```diff
@@ -189,6 +189,8 @@ default_nan_checker(::OnlyParticleTrackingModel) = nothing
 include("seawater_density.jl")
 
 # Implementation of the diagnostic for computing the dissipation rate
-include("dissipation_computation.jl")
+include("VarianceDissipationComputation/VarianceDissipationComputation.jl")
+
+using .VarianceDissipationComputation
```
glwagner, line 194:
```suggestion
include("VarianceDissipations/VarianceDissipations.jl")

using .VarianceDissipations
```

glwagner: This is explicitly the julia convention for modules that introduce a type

src/Models/VarianceDissipationComputation/get_dissipation_fields.jl
```diff
@@ -0,0 +1,27 @@
...
+function get_dissipation_fields(t::TracerVarianceDissipation, tracer_name) 
+    A = t.advective_production[tracer_name]
+    D = t.diffusive_production[tracer_name]
+    G = t.gradient_squared[tracer_name]
+
+    dirs = tracer_name == :ζ ? (:x, :y) : (:x, :y, :z)
+
+    prod_names = Tuple(Symbol(:A, tracer_name, dir) for dir in dirs)
+    diff_names = Tuple(Symbol(:D, tracer_name, dir) for dir in dirs)
+    grad_names = Tuple(Symbol(:G, tracer_name, dir) for dir in dirs)
```
glwagner, line 20: wut

src/Models/VarianceDissipationComputation/get_dissipation_fields.jl
```diff
@@ -0,0 +1,27 @@
...
+
+    advective_prod = Tuple(getproperty(A, dir) for dir in dirs)
+    diffusive_prod = Tuple(getproperty(D, dir) for dir in dirs)
+    grad = Tuple(getproperty(G, dir) for dir in dirs)
+
+    return NamedTuple{tuple(prod_names..., diff_names..., grad_names...)}(tuple(advective_prod..., diffusive_prod..., grad...))
```
glwagner, line 26: ummm

you may have meant to write something like

`names = tuple(prod_names..., diff_names..., grad_names...)`

src/Models/VarianceDissipationComputation/dissipation_utils.jl
```diff
@@ -0,0 +1,48 @@
...
+    return KernelParameters(sz, of)
+end
+
+function tracer_fluxes(grid)
+    x = XFaceField(grid)
+    y = YFaceField(grid)
+    z = ZFaceField(grid)
+
+    return (; x, y, z)
+end
```
glwagner, line 41: does this help?

src/Models/VarianceDissipationComputation/diffusive_fluxes.jl
```diff
@@ -0,0 +1,40 @@
...
+end
+
+@inline function compute_diffusive_tracer_fluxes!(Vⁿ, Vⁿ⁻¹, i, j, k, grid, clo, K, b, c, c_id, clk, fields)
+    Vⁿ⁻¹.x[i, j, k] = Vⁿ.x[i, j, k]
+    Vⁿ⁻¹.y[i, j, k] = Vⁿ.y[i, j, k]
+    Vⁿ⁻¹.z[i, j, k] = Vⁿ.z[i, j, k]
+
+    Vⁿ.x[i, j, k] = _diffusive_tracer_flux_x(i, j, k, grid, clo, K, Val(c_id), c, clk, fields, b) * Axᶠᶜᶜ(i, j, k, grid)
+    Vⁿ.y[i, j, k] = _diffusive_tracer_flux_y(i, j, k, grid, clo, K, Val(c_id), c, clk, fields, b) * Ayᶜᶠᶜ(i, j, k, grid)
+    Vⁿ.z[i, j, k] = _diffusive_tracer_flux_z(i, j, k, grid, clo, K, Val(c_id), c, clk, fields, b) * Azᶜᶜᶠ(i, j, k, grid)
```
glwagner, line 20: missing @inbounds

simone-silvestri:
The metric seems to work in a simple 1D setting, I need to add docstings and tests

src/Models/VarianceDissipationComputation/dissipation_utils.jl
simone-silvestri: we could write it manually but it is needed in many places, for example:
```julia
K    = NamedTuple{tracers}(tracer_fluxes(grid) for tracer in tracers)
Fⁿ   = NamedTuple{tracers}(tracer_fluxes(grid) for tracer in tracers)
Fⁿ⁻¹ = NamedTuple{tracers}(tracer_fluxes(grid) for tracer in tracers)
```

simone-silvestri:
> One quick question: since this is a diagnostic quantity, would it make sense to include this on Oceanostics instead of Oceananigans?

I thought to have it here because (as I see it) this metric is not only a diagnostic but a complement to our implicit advection, a necessary piece of the puzzle for users that want to use WENO schemes. I am also envisioning parameterizations depending on this metric, i.e., backscattering, where you need to know the numerical dissipation to reinject the lost energy onto the resolved scales (it will probably require a tad of algebra to expand the tracer variance formulation to kinetic energy, but nonetheless the development starts from here). 

However, if the consensus is to add it to Oceanostics, I am also fine, since at the moment, the burden of maintenance might be lower in another package.

src/Models/VarianceDissipationComputation/get_dissipation_fields.jl
simone-silvestri: I have corrected the function, it is now called `flatten_dissipation_fields` and flattens all the fields computed from the diagnostic in one easy namedtuple that can be passed to the output writer

simone-silvestri:
This PR is probably ready to merge, an example of how to use this is shown in `validation/implicit_dissipation/variance_dissipation.jl`

src/Models/dissipation_computation.jl
simone-silvestri: I have made it into a diagnostic with a hardcoded `IterationInterval(1)` schedule

src/Diagnostics/VarianceDissipationComputations/VarianceDissipationComputations.jl
```diff
@@ -0,0 +1,126 @@
...
+Argument
+=========
+
+- `model`: The model for which variance dissipation is to be computed. 
+  The model must use a `QuasiAdamsBashforth2TimeStepper` for its time-stepping scheme.
+
+Keyword Argument
+================
+
+- `tracers`: An optional argument specifying the tracers for which variance dissipation
```
tomchor, line 59: How hard is it to generalize for momentum?

src/Diagnostics/VarianceDissipationComputations/assemble_dissipation.jl
```diff
@@ -0,0 +1,67 @@
...
+#
+# Where ``F'' is the flux associated with the particular process,``U'' is the adecting velocity,
+#  while ``c★'' and ``c²'' are functions defined above.
+# Note that ``F'' and ``U'' need to be numerically accurate for the budgets to close,
+# i.e. for and AB2 scheme:
+#
+# F = 1.5 Fⁿ - 0.5 Fⁿ⁻¹
+# U = 1.5 Uⁿ - 0.5 Uⁿ⁻¹
+#
+# For an RK3 method (not implemented at the moment), the whole substepping procedure needs to be accounted for.
```
tomchor, line 27: Can we make this into a docstring?

src/Diagnostics/VarianceDissipationComputations/assemble_dissipation.jl
```diff
@@ -0,0 +1,67 @@
...
+end
+
+@inline c★(i, j, k, grid, cⁿ⁺¹, cⁿ) = @inbounds (cⁿ⁺¹[i, j, k] + cⁿ[i, j, k]) / 2
+@inline c²(i, j, k, grid, cⁿ⁺¹, cⁿ) = @inbounds (cⁿ⁺¹[i, j, k] * cⁿ[i, j, k])
+
+# Assemble the dissipation from the previously calculated advective and 
+# diffusive fluxes, the formulation is:
+#
+# A = 2 * δc★ * F - U δc²  # For advective dissipation
+# D = 2 * δc★ * F           # For diffusive dissipation
```
tomchor, line 17:
```suggestion
# A = 2 * δc★ * F - U δc²  # For advective dissipation
# D = 2 * δc★ * F          # For diffusive dissipation
```

src/Diagnostics/VarianceDissipationComputations/assemble_dissipation.jl
tomchor: Also, it's not clear to me what "assemble" means. Could you please clarify?

src/Diagnostics/VarianceDissipationComputations/dissipation_utils.jl
```diff
@@ -0,0 +1,21 @@
+import Oceananigans.Utils: KernelParameters
```
tomchor, line 1: This doesn't seem to be used in this file

src/Diagnostics/VarianceDissipationComputations/VarianceDissipationComputations.jl
simone-silvestri: For flux form momentum transport is rather trivial (the formulation is exaclty the same, we need to change only the `update_transport!` to cache interpolated advecting velocities.

For vector invariant momentum transport it is rather difficult, we are trying to figure out a way to have a numerical solution that closes the budgets, also in that case.

src/Diagnostics/VarianceDissipationComputations/assemble_dissipation.jl
simone-silvestri: I have changed the name of the function to compute

src/Diagnostics/VarianceDissipationComputations/VarianceDissipationComputations.jl
```diff
@@ -0,0 +1,126 @@
...
+    diffusivities = model.diffusivity_fields
+    closure       = model.closure
+    grid          = model.grid
+
+    P    = NamedTuple{tracers}(tracer_fluxes(grid) for tracer in tracers)
+    K    = NamedTuple{tracers}(tracer_closure_dissipation(grid, diffusivities, closure, id) for id in eachindex(tracers))
+    Vⁿ   = NamedTuple{tracers}(tracer_closure_dissipation(grid, diffusivities, closure, id) for id in eachindex(tracers))
+    Vⁿ⁻¹ = NamedTuple{tracers}(tracer_closure_dissipation(grid, diffusivities, closure, id) for id in eachindex(tracers))    
+
+    K    = NamedTuple{tracers}(tracer_fluxes(grid) for tracer in tracers)
```
glwagner, line 82: this is created twice

glwagner:
> @simone-silvestri this is great! Looking forward to trying this out!
> 
> One quick question: since this is a diagnostic quantity, would it make sense to include this on Oceanostics instead of Oceananigans?

This should probably be in Oceananigans because it does not use any abstract interfaces and invokes a lot of custom kernels. I guess perhaps the main point is that it needs to be tested on GPU. But also I think we should think of Oceanostics as providing implementations that use Oceananigans interfaces rather than implementing independent interfaces. The reason to use this logic is to ensure that Oceanostics does not stray from / introduce new abstractions that properly should be defined in a single package to remain coherent.

simone-silvestri:
I will merge this when the tests pass, to have the metric in but there are a couple of changes that I am planning immediately after merging (in this order):
1. changing the variance dissipation metric to be per-field instead of per-model
2. making the metric callable in a `Callback` rather than a diagnostic so that it is possible to call it in `update_state!` rather than at the end of the time step (needed for adaptation to RK3)
3. generalization to flux-form advection
4. support for RK3

PR #3880, opened 2024-10-29
Bugfix in distributed GPU tests and Distributed `set!`

### LINKED ISSUE #3897: The MPI we use in the distributed tests is not CUDA-aware
Somewhere between this commit 
https://buildkite.com/clima/oceananigans-distributed/builds/3113#01917ace-fe81-401d-ba21-467037e6aead
and main, we switched from using `libmpitrampoline.so` in the distributed tests to `libmpi.so` downloaded from the artifacts. 

Previously, the mpi trampoline was loading a CUDA-aware implementation of Open MPI, while the libmpi.so we use now is a 
MPICH implementation non CUDA-aware:
https://buildkite.com/clima/oceananigans-distributed/builds/4227#0192f70a-b947-4d38-bd1c-c2497a964de9

This makes our GPU distributed tests fail. 
I am wondering where this switch happened because I couldn't trace any changes to the code. @Sbozzolo, do you know if something changed in the `LocalPreferences.toml` in the Caltech cluster?

simone-silvestri:
This PR modifies the configuration of the distributed pipeline that runs on the Caltech cluster to allow using CUDA-aware MPI.

closes #3897


test/utils_for_runtests.jl
```diff
@@ -3,21 +3,26 @@ using Oceananigans.DistributedComputations: Distributed, Partition, child_archit
 
 import Oceananigans.Fields: interior
 
-test_child_arch() = CUDA.has_cuda() ? GPU() : CPU()
+test_child_arch() = parse(Bool, get(ENV, "GPU_TEST", "false")) ? GPU() : CPU()
+mpi_test() = parse(Bool, get(ENV, "MPI_TEST", "false"))
```
glwagner, line 7: Why does this have to be a function?

glwagner, line 7:
```suggestion
mpi_test = get(ENV, "MPI_TEST", nothing) == "true"
```

test/utils_for_runtests.jl
```diff
@@ -3,21 +3,26 @@ using Oceananigans.DistributedComputations: Distributed, Partition, child_archit
 
 import Oceananigans.Fields: interior
 
-test_child_arch() = CUDA.has_cuda() ? GPU() : CPU()
+test_child_arch() = parse(Bool, get(ENV, "GPU_TEST", "false")) ? GPU() : CPU()
```
glwagner, line 6:
```suggestion
const gpu_test = get(ENV, "GPU_TEST", nothing) == "true"
test_child_arch() = gpu_test ? GPU() : CPU()
```

glwagner: Probably could just get rid of `test_child_arch()` though, seems convoluted

glwagner:
Adding the Manifest won't help for the tests because the test environment generates a new manifest every time, I learned

simone-silvestri:
Ok, good to know. I am trying different approaches, but everything seems to fail; at least I can remove the manifest.

simone-silvestri:
Finally, this works. It would need approval to get the tests back online.

ali-ramadhan (approved):
Looks good to me!

.buildkite/distributed/pipeline.yml
```diff
@@ -16,60 +16,83 @@ steps:
...
   - label: "🐉 cpu distributed unit tests"
     key: "distributed_cpu"
     env:
       TEST_GROUP: "distributed"
+      MPI_TEST: "true"
     commands:
       - "srun julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
     agents:
-      slurm_mem: 120G
+      slurm_mem: 8G
```
glwagner, line 45: why?

Project.toml
```diff
@@ -77,10 +77,11 @@ julia = "1.9"
...
-Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 TimesDates = "bdfc003b-8df8-5c39-adcd-3a9087f5df4a"
+Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
+MPIPreferences = "3da0fdf6-3ccc-4f1b-acd9-58baa6c99267"
 
 [targets]
-test = ["DataDeps", "Enzyme", "SafeTestsets", "Test", "TimesDates"]
+test = ["DataDeps", "SafeTestsets", "Test", "Enzyme", "MPIPreferences", "TimesDates"]
```
glwagner, line 87: Was this the crucial part?

glwagner (approved):
This looks good. For future generations, can you please write a little bit about what you tried and what ended up working? I can't tell if all the changes are necessary, though the end result is fairly clean. Mostly I am wondering about `slurm_mem`. I'm also curious why we cannot call `precompile_runtime` inside `runtests.jl` and it is necessary to call it before `Pkg.test()`. This has implications for the CI of other packages.

.buildkite/distributed/pipeline.yml
simone-silvestri: 120G is much more than we need for those tests. I noticed that the agent starts much quicker by requesting a smaller memory amount. So I am deducing that the tests run on shared nodes instead of exclusive ones, and specifying lower resources allows us to squeeze in also when the cluster is busy.

simone-silvestri:
I think it is equivalent. I am trying to precompile inside the runtests.

Btw, having again access to GPU distributed tests highlighted a couple of bugs related to distributed architectures specifically for the `set!` and the `constructor_arguments` function, which I am fixing in this PR.

simone-silvestri:
there are two distinct issues with the GPU tests
1) MPI was not CUDA-aware
2) CUDA_runtime was not found in the tests

Issue number (1) is solved, but, unfortunately, some tests fail stochastically because CUDA_runtomel is not found. 
It seems to be linked to how late they are started after the `init` step, i.e. tests that start much later will fail with CUDA_runtime not found error. I am still investigating

simone-silvestri:
Ok, with some fiddling CUDA seems to be found correctly now. I think this https://github.com/CliMA/Oceananigans.jl/blob/908b31a67d1e8d4011d790746777fee3b63b170a/Project.toml#L81 is the change that made it work.

However, I have added a failsafe option following this suggestion (which was the error we were encountering)
https://github.com/JuliaGPU/CUDA.jl/blob/a085bbb3d7856dfa929e6cdae04a146a259a2044/src/initialization.jl#L96
which reloads and compiles CUDA_Runtime_jll and restarts the julia session.

.buildkite/distributed/pipeline.yml
glwagner: good reason. might warrant a comment

.buildkite/distributed/pipeline.yml
```diff
@@ -51,7 +51,7 @@ steps:
     commands:
       - "srun julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
     agents:
-      slurm_mem: 8G
+      slurm_mem: 50G
```
glwagner: 👀

glwagner: we can probably reduce the memory usage of the tests right? I think often a bigger grid is used than needed

simone-silvestri: Right, unit tests do not require too much memory. I have seen that 32G was not enough for the regression tests on the GPU.

navidcy:
merge?

PR #3891, opened 2024-10-31
Use launch! for particle advection

glwagner:
Trying to support particle vectors of length 0, see #3884


glwagner:
ok @aleksok this should be ready to go

jagoosw (approved):

PR #3900, opened 2024-11-05
Mask `ReducedFields` accounting the immersed column

### LINKED ISSUE #3899: Plotting a reduced field on an `ImmersedBoundaryGrid` fills in NaNs incorrectly
MWE:
```julia
using Oceananigans
using GLMakie

grid = RectilinearGrid(size = (10, 10, 10), extent = (10, 10, 10))
bottom(x, y) = - rand() * 5 # Between -5 and 0

f1 = Field{Center, Center, Nothing}(grid)
f2 = Field{Center, Center, Nothing}(grid)
set!(f1, (x, y) -> rand())
set!(f2, (x, y) -> rand())

fig = Figure(); ax1 = Axis(fig[1, 1]); ax2 = Axis(fig[1, 2])
heatmap!(ax1, f1)
heatmap!(ax2, interior(f2, :, :, 1))
```
Produces:

Screenshot 2024-11-05 at 12 21 29 PM

It seems the field has been masked by checking the immersed condition at `k = 1`, which is not true for `Reduced` fields that need to check the whole direction to see if they are immersed. To mask the field to `NaN,` we should check that all the `k` indices at `i, j` are immersed. 

This is not a bug in the visualization but in the masking procedure that should do something different if a direction is nothing. I can create a PR for this; some computational efficiency considerations might be needed if this `mask_immersed_field!` is included in time-stepping loops.

simone-silvestri:
From a MWE similar to the one in #3899 (adding some more reduced directions):
```julia
using Oceananigans
using GLMakie

grid = RectilinearGrid(size = (10, 10, 10), extent = (10, 10, 10))
bottom(x, y) = - rand() * 5 # Between -5 and 0

f1 = Field{Center, Center, Nothing}(grid)
f2 = Field{Center, Nothing, Center}(grid)
f3 = Field{Nothing, Center, Center}(grid)
set!(f1, (x, y) -> rand())
set!(f2, (x, y) -> rand())
set!(f3, (x, y) -> rand())

fig = Figure(); ax1 = Axis(fig[1, 1]); ax2 = Axis(fig[1, 2]); ax3 = Axis(fig[1, 3])
heatmap!(ax1, f1)
heatmap!(ax2, f2)
heatmap!(ax3, f3)
```
Screenshot 2024-11-05 at 12 26 38 PM

Should we add some tests?

closes #3899

### DIFF
--- a/src/ImmersedBoundaries/mask_immersed_field.jl
+++ b/src/ImmersedBoundaries/mask_immersed_field.jl
@@ -96,44 +96,32 @@ end
 # For a `ReducedField` we mask if the entire direction is immerded in the reduced direction.
 # This requires a sweep over the reduced direction
 
-function mask_immersed_field!(field::ReducedField, grid, loc, value)
-    loc = instantiate.(loc)
-    launch!(architecture(field), grid, size(field), _mask_immersed_reduced_field!, field, loc, grid, value)
+function mask_immersed_field!(field::ReducedField, grid::ImmersedBoundaryGrid, loc, value)
+    loc  = instantiate.(loc)
+    dims = reduced_dimensions(field)
+    launch!(architecture(field), grid, size(field), _mask_immersed_reduced_field!, field, dims, loc, grid, value)
     return nothing
 end
 
-@kernel function _mask_immersed_reduced_field!(field, loc, grid, value)
+@kernel function _mask_immersed_reduced_field!(field, dims, loc, grid, value)
     i, j, k = @index(Global, NTuple)
-
-    mask = mask_reduced_x_direction!(field, i, j, k, grid, loc, true)
-    mask = mask_reduced_y_direction!(field, i, j, k, grid, loc, mask)
-    mask = mask_reduced_z_direction!(field, i, j, k, grid, loc, mask)
-
+    mask = mask_reduced_direction(i, j, k, grid, dims, loc)
     @inbounds field[i, j, k] = ifelse(mask, value, field[i, j, k]) 
 end
 
-@inline mask_reduced_x_direction!(field, i, j, k, grid, loc, mask = true) = mask & peripheral_node(i, j, k, grid, loc...)
-@inline mask_reduced_y_direction!(field, i, j, k, grid, loc, mask = true) = mask & peripheral_node(i, j, k, grid, loc...)
-@inline mask_reduced_z_direction!(field, i, j, k, grid, loc, mask = true) = mask & peripheral_node(i, j, k, grid, loc...)
+@inline masking_range(i, grid, dim, dims) = ifelse(dim ∈ dims, 1:size(grid, dim), i:1)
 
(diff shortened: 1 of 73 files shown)


src/ImmersedBoundaries/mask_immersed_field.jl
```diff
@@ -89,6 +89,55 @@ end
     @inbounds field[i, j, k] = scalar_mask(i, j, k, grid, grid.immersed_boundary, loc..., value, field, mask)
 end
 
+#####
+##### Masking a `ReducedField`
+#####
+
+# We mask a `ReducedField` if the entire reduced direction is immersed.
+# This requires a sweep over the reduced direction
```
glwagner, line 97: But fields only have 1 grid point in reduced directions?

src/ImmersedBoundaries/mask_immersed_field.jl
```diff
@@ -89,6 +89,55 @@ end
...
+
+@kernel function _mask_immersed_reduced_field!(field, dims, loc, grid, value)
+    i, j, k = @index(Global, NTuple)
+    mask = mask_reduced_direction(i, j, k, grid, dims, loc)
+    @inbounds field[i, j, k] = ifelse(mask, value, field[i, j, k]) 
+end
+
+@inline masking_range(i, grid, dim, dims) = ifelse(dim ∈ dims, 1:size(grid, dim), i:i)
+
+@inline function mask_reduced_direction(i₀, j₀, k₀, grid, dims, loc)
```
glwagner, line 114:
```suggestion
@inline function inactive_dimensions(i₀, j₀, k₀, grid, dims, loc)
```

This function returns `true` if the entire `dims` are inactive. Thus it is like `inactive`, but applied to a whole dimension.

src/ImmersedBoundaries/mask_immersed_field.jl
```diff
@@ -89,6 +89,55 @@ end
...
+    return nothing
+end
+
+@kernel function _mask_immersed_reduced_field!(field, dims, loc, grid, value)
+    i, j, k = @index(Global, NTuple)
+    mask = mask_reduced_direction(i, j, k, grid, dims, loc)
+    @inbounds field[i, j, k] = ifelse(mask, value, field[i, j, k]) 
+end
+
+@inline masking_range(i, grid, dim, dims) = ifelse(dim ∈ dims, 1:size(grid, dim), i:i)
```
glwagner, line 112:
```suggestion
@inline inactive_search_range(i, grid, dim, dims) = ifelse(dim ∈ dims, 1:size(grid, dim), i:i)
```

glwagner: or something. This is not specific to masking. This is about looking for inactive cells.

src/ImmersedBoundaries/mask_immersed_field.jl
glwagner: also the argument can contain multiple dimensions so you need plural

glwagner (approved):

PR #3902, opened 2024-11-05
For output readers, make `reader_kw` default to an empty `NamedTuple`

### LINKED ISSUE #3898: Use default `reader_kw=nothing` to make `FieldTimeSeries` concretely typed
The default is not a concrete type which may cause performance issues when `FieldTimeSeries` is used as boundary condition or forcing in a kernel:

https://github.com/CliMA/Oceananigans.jl/blob/6c40d7e225c2127051b2703b9c62a8b18260e3a5/src/OutputReaders/field_time_series.jl#L373

I suggest `nothing` as the default. Users can still change it if they want.

We could also use a `NamedTuple` instead of `Dict`.

@ali-ramadhan I think you maybe were not thinking that we use `FieldTimeSeries` in kernels and models now (not just for reading output)

ali-ramadhan:
Resolves #3898

### DIFF
--- a/test/test_output_readers.jl
+++ b/test/test_output_readers.jl
@@ -390,7 +390,7 @@ end
         @testset "FieldDataset{$Backend} parallel reading" begin
             @info "  Testing FieldDataset{$Backend} parallel reading..."
 
-            reader_kw = Dict(:parallel_read => true)
+            reader_kw = (parallel_read = true,)
             ds = FieldDataset(filepath3d; backend=Backend(), reader_kw)
 
             @test ds isa FieldDataset
--- a/test/test_output_readers.jl
+++ b/test/test_output_readers.jl
@@ -390,7 +390,7 @@ end
         @testset "FieldDataset{$Backend} parallel reading" begin
             @info "  Testing FieldDataset{$Backend} parallel reading..."
 
-            reader_kw = (parallel_read = true,)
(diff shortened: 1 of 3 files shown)


test/test_output_readers.jl
```diff
@@ -390,7 +390,7 @@ end
         @testset "FieldDataset{$Backend} parallel reading" begin
             @info "  Testing FieldDataset{$Backend} parallel reading..."
 
-            reader_kw = Dict(:parallel_read => true)
+            reader_kw = (parallel_read = true,)
```
glwagner, line 393:
```suggestion
            reader_kw = (; parallel_read = true)
```

preferred syntax (less prone to typos)

glwagner (approved):

ali-ramadhan:
Hmmm, a GPU Enzyme test is consistently failing and distributed tests are segfaulting. Lemme re-run the builds and hope things magically change haha :crossed_fingers:

ali-ramadhan:
Seems like all GPU Buildkite agents are stuck here: https://buildkite.com/clima/oceananigans/builds/18654#_ and no CPU Buildkite agents are free?

simone-silvestri:
Both the problems with the tests should be solved by #3880; btw it would need a review / approval if you can 😄

PR #3907, opened 2024-11-06
Update examples to use new fluxes notation

glwagner:
This PR updates the examples to use the notation for fluxes which is used by [Wagner et al 2024](https://glwagner.github.io/assets/pdf/CATKE.pdf) and which we are planning to use for subsequent work as well. In short this is:

* $Q$ for heat flux
* $\tau_x$ and $\tau_y$ for the zonal and meridional _kinematic_ momentum flux. The momentum flux is therefore $\rho_o \tau_x$, where $\rho_o$ is the ocean surface density. This is also consistent with the notation used for Clima's atmosphere model
* $J^c$ for the flux of a tracer $c$.

### DIFF
--- a/examples/langmuir_turbulence.jl
+++ b/examples/langmuir_turbulence.jl
@@ -90,18 +90,18 @@ uˢ(z) = Uˢ * exp(z / vertical_scale)
 #
 # At the surface ``z = 0``, Wagner et al. (2021) impose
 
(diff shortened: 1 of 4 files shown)


navidcy (approved):
Lgtm

glwagner:
does anyone ever say "lbtm"

navidcy:
Haha

You hope that it that’s the case they are more explicit to what doesn’t good that good

simone-silvestri (approved):

Issue #3910, opened 2024-11-08
GPU enzyme tests are running on the CPU


simone-silvestri:
I guess it is intended because there is no architecture argument passed to the grids:
https://github.com/CliMA/Oceananigans.jl/blob/4e224320c3d10f0558aef78d19a4905f0ec95b72/test/test_enzyme.jl#L117
https://github.com/CliMA/Oceananigans.jl/blob/4e224320c3d10f0558aef78d19a4905f0ec95b72/test/test_enzyme.jl#L172

and
https://github.com/CliMA/Oceananigans.jl/blob/4e224320c3d10f0558aef78d19a4905f0ec95b72/test/test_enzyme.jl#L71

So, I guess this is a placeholder for when the GPU support will be online.
However, Sverdrup has quite a slow CPU so having a duplicate test is not very efficient for development at the moment

PR #3911, opened 2024-11-08
Run Enzyme tests on the GPU on Sverdrup

simone-silvestri:
Solves #3910. 

However, if it was just a typo and GPU support is in fact enabled, we can use this PR to pass the correct architecture to the grid and enable GPU tests

Let me know @jlk9 @wsmoses @glwagner


glwagner:
What's the problem? Why disable the tests?

simone-silvestri:
It's running on the CPU so it is a duplicate of the Tartarus one, but it takes quite some time on Sverdrup because the CPU there is not that great.

glwagner:
the fix is to change the test to GPU

glwagner:
I'll do it

simone-silvestri:
Yeah I mentioned it in the PR

simone-silvestri:
I changed the name of the PR then

glwagner:
If they take too long to run / compile, we will have to come up with a differerent strategy

glwagner:
How long do they take right now?

simone-silvestri:
It was taking upwards of an hour. But on the GPU I imagine it takes less.

simone-silvestri:
About an hour, but it depends if there is space on the cluster or not.

glwagner:
Ok, so if it takes an hour we are alright I think

glwagner:
@jlk9 you may want to push this forward so the enzyme tests run on GPU

glwagner:
@jlk9 do you want the enzyme tests to run on GPU?

PR #3894, opened 2024-11-01
(0.94.1) Refactor `SplitExplicitFreeSurface`

simone-silvestri:
refactoring of the `SplitExplicitFreeSurface` to comply with issue #3873 

Additional changes included in this PR not described in #3873:
- the grid halos are extended automatically upon passing a grid with a `Connected` topology (`FullyConnected`, `RightConnected` or `LeftConnected`) so there is no need to extend the constructor for grids (such as distributed grids) that have connected topologies.
- All the `ExplicitFreeSurface` relevant code has been moved in the `explicit_free_surface.jl` file

PR #3912, opened 2024-11-10
Remove redundant advection aliases

simone-silvestri:
Since they have been removed from the exported methods, I think we do not have a need for them anymore.

### DIFF
--- a/benchmark/benchmark_advection_schemes.jl
+++ b/benchmark/benchmark_advection_schemes.jl
@@ -44,7 +44,7 @@ if GPU in Architectures
 end
 
 for Arch in Architectures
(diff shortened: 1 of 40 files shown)


glwagner (approved):
Agree. They aren't useful

tomchor:
Wasn't this a breaking PR?

simone-silvestri:
Hmmm, you are right. We probably should bump the version

tomchor:
I think so. Although maybe we wanna for https://github.com/CliMA/Oceananigans.jl/pull/3143 before registering a new version in order to avoid two breaking releases in a row. That PR seems ready to merge.

simone-silvestri:
sounds good! I want to merge also #3894 but I can merge now when the tests pass

PR #3913, opened 2024-11-11
Tripolar grid in Oceananigans (only serial)

simone-silvestri:
Our implementation of the tripolar grid is in https://github.com/CliMA/OrthogonalSphericalShellGrids.jl at the moment. 

We have figured out that the implementation of the tripolar grid is very much entangled with Oceananigans because it requires extending a lot of time-stepping specific code. 

This leads to delays in development when we have to develop the two packages in tandem.
Therefore, this PR migrates the implementation of the Tripolar grid and its specific methods to Oceananigans.

Edit: This PR implements only the serial version. The distributed TripolarGrid requires a little more work and will come in subsequent PRs


navidcy (approved):
nice

we should update

https://github.com/CliMA/Oceananigans.jl/blob/dc4a327fbc149970190dbc077bff84d32b615141/docs/src/grids.md?plain=1#L86-L93

navidcy:
@simone-silvestri 

Which `device` is meant to be used here?

https://github.com/CliMA/Oceananigans.jl/blob/db49f4db9c7b8aa6e9b4d7ebd4658cf4b0674dd3/src/Grids/tripolar_grid.jl#L108

```julia
WARNING: both Architectures and CUDA export "device"; uses of it in module Grids must be qualified
Unit tests...: Error During Test at /Users/navid/Research/OC11.jl/test/test_tripolar_grid.jl:31
  UndefVarError: `device` not defined
  Stacktrace:
    [1] (TripolarGrid)(arch::CPU, FT::DataType; size::Tuple{Int64, Int64, Int64}, southernmost_latitude::Int64, halo::Tuple{Int64, Int64, Int64}, radius::Float64, z::Tuple{Int64, Int64}, north_poles_latitude::Int64, first_pole_longitude::Int64)
      @ Oceananigans.Grids ~/Research/OC11.jl/src/Grids/tripolar_grid.jl:108
```

glwagner:
Why draft?

navidcy:
> Why draft?

I guess because it needs work... Still most test fail.

glwagner:
Huh true. Just thought it would have been a copy/paste job but maybe there's more to it

navidcy:
It shouldn't be hard to fix... I can give it a go later!

simone-silvestri:
the main problem is that we use Fields and `fill_halo_regions!` in the construction of the grid. I ll search a way around it

navidcy:
Oh I just saw your comment @simone-silvestri!

simone-silvestri:
We could move the zipper-filling halo kernels in grids and import them in boundary conditions.
It is a bit convoluted though.

navidcy:
```Julia
using Oceananigans.Grids: R_Earth, lat_lon_to_cartesian
using Distances

# Is this the same as in Oceananigans? 
# TODO: check it out
function haversine_simone(a, b, radius)
    λ₁, φ₁ = a
    λ₂, φ₂ = b

    x₁, y₁, z₁ = lat_lon_to_cartesian(φ₁, λ₁, radius)
    x₂, y₂, z₂ = lat_lon_to_cartesian(φ₂, λ₂, radius)

    return radius * acos(max(-1.0, min((x₁ * x₂ + y₁ * y₂ + z₁ * z₂) / radius^2, 1.0)))
end

using Test

for i in 1:100
    (λ1, φ1) = rand()*360, (2rand()-1)*90
    (λ2, φ2) = rand()*360, (2rand()-1)*90

    radius = rand() * 6300e3

    @test haversine_simone((λ1, φ1), (λ2, φ2), radius) ≈ Distances.haversine((λ1, φ1), (λ2, φ2), radius)
end
```

navidcy:
@simone, what's the paper mentioned here:

> Note: There is probably a better way to do this, in Murray (2016) they give analytical expressions for the metric terms.

simone-silvestri:
The date was wrong 😅 
The paper is this one https://www.sciencedirect.com/science/article/abs/pii/S0021999196901369

simone-silvestri:
Let's see if it works with the current workaround

glwagner:
Should we change how the code is organized? Can we put `Fields` before `Grids`?

simone-silvestri:
Hmmm, I think Grids should be before Fields. 
I was thinking to implement the rules to fold the north boundary in the `tripolar_grid.jl` file 
https://github.com/CliMA/Oceananigans.jl/blob/ce8c49e92a831d12dafbbc09c91be8f141a7bccc/src/Grids/tripolar_grid.jl#L234-L260

and then use it in boundary conditions:
https://github.com/CliMA/Oceananigans.jl/blob/ce8c49e92a831d12dafbbc09c91be8f141a7bccc/src/BoundaryConditions/fill_halo_regions_zipper.jl#L47-L56

glwagner:
> Hmmm, I think Grids should be before Fields. I was thinking to implement the rules to fold the north boundary in the `tripolar_grid.jl` file
> 
> https://github.com/CliMA/Oceananigans.jl/blob/ce8c49e92a831d12dafbbc09c91be8f141a7bccc/src/Grids/tripolar_grid.jl#L234-L260
> 
> and then use it in boundary conditions:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/ce8c49e92a831d12dafbbc09c91be8f141a7bccc/src/BoundaryConditions/fill_halo_regions_zipper.jl#L47-L56

what about having an `OrthogonalSphericalShellGrids` file after `Fields`?

It makes sense to me that for non-trivial grids, we actually want to use grids and fields to build the grid. For example, the cubed sphere grid is a conformal map of a rectilinear grid.

Just a thought to save some work but if you think it will ultimate accelerate progress (for some reason) to put everything in `Grids` then we should do it.

simone-silvestri:
Well, I guess if we move everything in a module below Fields and before AbstractOperations, it is just basically a copy and paste of `OrthogonalSphericalShellGrids` in `Oceananigans` which indeed, will be very fast. 

I am ok with this solution if you guys think we can do this now. This would simplify all the construction of non-trivial grids in the future (probably the tripolar grid will not be the last)

glwagner:
> Well, I guess if we move all the tripolar grid construction in a `OrthogonalSphericalShellGrids` in a module below Fields and before AbstractOperations, it is just basically a copy and paste of `OrthogonalSphericalShellGrids` in `Oceananigans` which indeed, will be very fast. We could also add the cubed sphere panel there.
> 
> I am ok with this solution if you guys think we can do this at this moment. This would simplify all the construction of non-trivial grids in the future (probably the tripolar grid will not be the last).
> 
> We could reassess this implementation in the future maybe?

I think that's a nice plan for sure. It's nice to have the basic utilities available when building new grids via conformal maps, etc. We can also implement the rotated lat-lon grid there?

simone-silvestri:
The tests seem to pass

src/BoundaryConditions/boundary_condition.jl
```diff
@@ -82,20 +82,23 @@ const GBC  = BoundaryCondition{<:Gradient}
...
 const DCBC = BoundaryCondition{<:DistributedCommunication}
+const ZBC  = BoundaryCondition{<:Zipper}
 
 # More readable BC constructors for the public API.
                 PeriodicBoundaryCondition() = BoundaryCondition(Periodic(),                 nothing)
                   NoFluxBoundaryCondition() = BoundaryCondition(Flux(),                     nothing)
             ImpenetrableBoundaryCondition() = BoundaryCondition(Open(), nothing)
 MultiRegionCommunicationBoundaryCondition() = BoundaryCondition(MultiRegionCommunication(), nothing)
 DistributedCommunicationBoundaryCondition() = BoundaryCondition(DistributedCommunication(), nothing)
+                  ZipperBoundaryCondition() = BoundaryCondition(Zipper(), 1)
```
glwagner, line 93: "1"?

glwagner:
```suggestion
                  ZipperBoundaryCondition(sign) = BoundaryCondition(Zipper(), sign)
```

src/OrthogonalSphericalShellGrids/generate_tripolar_coordinates.jl
```diff
@@ -0,0 +1,135 @@
...
+"""
+@kernel function _compute_tripolar_coordinates!(λFF, φFF, λFC, φFC, λCF, φCF, λCC, φCC,
+                                                λᶠᵃᵃ, λᶜᵃᵃ, φᵃᶠᵃ, φᵃᶜᵃ,
+                                                first_pole_longitude,
+                                                focal_distance, Nλ)
+
+    i, j = @index(Global, NTuple)
+
+    λ2Ds = (λFF,  λFC,  λCF,  λCC)
+    φ2Ds = (φFF,  φFC,  φCF,  φCC)
```
glwagner, line 61: why different notation than usual

glwagner (approved):

test/test_tripolar_grid.jl
```diff
@@ -0,0 +1,193 @@
...
+
+using Statistics
+using Oceananigans.Utils: get_cartesian_nodes_and_vertices
+using Oceananigans.ImmersedBoundaries: immersed_cell
+using Oceananigans.BoundaryConditions: Zipper
+
+using Oceananigans.Utils: KernelParameters, contiguousrange
+
+
+@kernel function compute_nonorthogonality_angle!(angle, grid, xF, yF, zF)
```
glwagner, line 11: What's a "nonorthogonality_angle"? Is this the deviation from orthogonality? (So it should be small)

simone-silvestri: yeah, it should be small. I am not sure what is acceptable, so I am comparing it here with a cubes sphere

glwagner (approved):

Issue #3918, opened 2024-11-12
Improve active cells map kernels


simone-silvestri:
Currently, if we build an immersed grid with `active_cells_map = true`, specific kernels compute only on the map provided by the grid. This is very beneficial, but the implementation is a bit clunky, requiring to pass the `active_cells_map` and define two different kernels, for example:
https://github.com/CliMA/Oceananigans.jl/blob/9111a8f3aa3742c57422079d78692f6114e33ea1/src/Models/NonhydrostaticModels/compute_nonhydrostatic_tendencies.jl#L145-L155

Since the implementation is a bit clunky, we restrict this behavior to the largest performance-critical kernels that are seldom called during time stepping. Otherwise, the code base would explode.

However, launching kernels only on the active cells is a low-hanging fruit for really speeding up the code, and most kernels in Oceananigans could benefit from it. This calls for a more implicit implementation that does not require defining two different kernels. There might be some way to pass a vector of indices to the kernel configuration and launch a one-dimensional kernel that iterates over that map.

That way, we could automatically compute everything on active cells if `active_cells_map = true` without bloating the code base. We could even avoid masking fields that live on a grid with an active map.

I think @vchuravy mentioned launching kernels on a map some time ago.

This is probably more of an issue for KernelAbstraction than for Oceananigans, but I just want to see what people think.

glwagner:
Can you implement such a feature like you implemented offsets, eg defining an `IndexMap` (rather than `KernelOffsets`) and `MappedNDRange` instead of `OffsetNDRange`?

https://github.com/CliMA/Oceananigans.jl/blob/9111a8f3aa3742c57422079d78692f6114e33ea1/src/Utils/kernel_launching.jl#L357

glwagner:
> We could even avoid masking fields that live on a grid with an active map.

I'm not even sure we need masking as is, but agree, it would be nice to consider masking as a convenience for output rather than something we need to do while the simulation runs.

simone-silvestri:
> Can you implement such a feature like you implemented offsets, eg defining an `IndexMap` (rather than `KernelOffsets`) and `MappedNDRange` instead of `OffsetNDRange`?

We could try, but the map is an `AbstractArray`, so I suspect it has to be stored in the `Kernel` and passed as an additional argument to the kernel.

simone-silvestri:
The offsets is passed as a type parameter in the NDRange argument of the kernel here
[...]

PR #3920, opened 2024-11-12
support for active_cells_map in kernels

simone-silvestri:
I am not sure how hacky this is or if it will work throughout the board, but this is supposed to achieve what was discussed in #3918:

for example, the current script:
```
using Oceananigans
grid = RectilinearGrid(size = (3, 3, 3), extent = (1, 1, 1))

using Oceananigans.Utils

@kernel function _test_indices(a)
    @show a[i, j, k]
end

a = [(i, j, k) for i in 1:3, j in 1:3, k in 1:3]
```

returns

```
julia> launch!(CPU(), grid, :xyz, _test_indices, a)
a[i, j, k] = (1, 1, 1)
a[i, j, k] = (2, 1, 1)
a[i, j, k] = (3, 1, 1)
a[i, j, k] = (1, 2, 1)
a[i, j, k] = (2, 2, 1)
a[i, j, k] = (3, 2, 1)
a[i, j, k] = (1, 3, 1)
a[i, j, k] = (2, 3, 1)
a[i, j, k] = (3, 3, 1)
a[i, j, k] = (1, 1, 2)
a[i, j, k] = (2, 1, 2)
a[i, j, k] = (3, 1, 2)
a[i, j, k] = (1, 2, 2)
a[i, j, k] = (2, 2, 2)
a[i, j, k] = (3, 2, 2)
a[i, j, k] = (1, 3, 2)
a[i, j, k] = (2, 3, 2)
a[i, j, k] = (3, 3, 2)
a[i, j, k] = (1, 1, 3)
a[i, j, k] = (2, 1, 3)
a[i, j, k] = (3, 1, 3)
a[i, j, k] = (1, 2, 3)
a[i, j, k] = (2, 2, 3)
a[i, j, k] = (3, 2, 3)
a[i, j, k] = (1, 3, 3)
a[i, j, k] = (2, 3, 3)
a[i, j, k] = (3, 3, 3)

julia> launch!(CPU(), grid, :xyz, _test_indices, a; active_cells_map = [(1, 2, 3), (3, 2, 1)])
a[i, j, k] = (1, 2, 3)
a[i, j, k] = (3, 2, 1)
```


src/Utils/kernel_launching.jl
```diff
@@ -413,3 +445,58 @@ function partition(kernel::OffsetKernel, inrange, ingroupsize)
     return iterspace, dynamic
 end
 
+#####
+##### Utilities for Mapped kernels
+#####
+
+struct IndexMap{M}
+    imap :: M
```
glwagner, line 453:
```suggestion
    index_map :: M
```

src/Utils/kernel_launching.jl
```diff
@@ -80,6 +82,28 @@ end
 contiguousrange(range::NTuple{N, Int}, offset::NTuple{N, Int}) where N = Tuple(1+o:r+o for (r, o) in zip(range, offset))
 flatten_reduced_dimensions(worksize, dims) = Tuple(d ∈ dims ? 1 : worksize[d] for d = 1:3)
 
+#### 
+#### Internal utility to launch a function mapped on an index_map
+####
+
+struct MappedFunction{F, M} <: Function
+    f::F
+    imap::M
```
glwagner, line 91: index_map?

src/Utils/kernel_launching.jl
```diff
@@ -80,6 +82,28 @@ end
 contiguousrange(range::NTuple{N, Int}, offset::NTuple{N, Int}) where N = Tuple(1+o:r+o for (r, o) in zip(range, offset))
 flatten_reduced_dimensions(worksize, dims) = Tuple(d ∈ dims ? 1 : worksize[d] for d = 1:3)
 
+#### 
+#### Internal utility to launch a function mapped on an index_map
+####
+
+struct MappedFunction{F, M} <: Function
+    f::F
```
glwagner, line 90:
```suggestion
    func::F
```

src/Utils/kernel_launching.jl
```diff
@@ -80,6 +82,28 @@ end
...
+    imap::M
+end
+
+Adapt.adapt_structure(to, m::MappedFunction) = 
+    MappedFunction(Adapt.adapt(to, m.f), Adapt.adapt(to, m.imap))
+
+@inline function (m::MappedFunction)(_ctx_)  
+    m.f(_ctx_)
+    return nothing
+end
```
glwagner, line 100: isn't this case covered by the case with `args...`?

src/Utils/kernel_launching.jl
```diff
@@ -238,9 +262,20 @@ the architecture `arch`.
...
     loop = kernel!(dev, workgroup, worksize)
+
+    # Map out the function to use active_cells_map
+    # as an index map
+    if !isnothing(active_cells_map)
+        func  = MappedFunction(loop.f, active_cells_map)
+        param = get_kernel_parameters(loop)
+        M     = typeof(func)
+        loop  = Kernel{param..., M}(dev, func)
+    end
```
glwagner, line 273: doesn't it make more sense to put this in the first if-statement in this function?

test.jl
```diff
@@ -0,0 +1,18 @@
...
+arch = CPU()
+
+@kernel function _test_indices(a)
+    i, j, k = @index(Global, NTuple)
+    a[i, j, k] = i + j + k
+end
+
+grid  = RectilinearGrid(arch, size = (3, 3, 3), extent = (1, 1, 1))
+array = zeros(arch, 3, 3, 3)
+imap  = on_architecture(arch, [(i, j, k) for i in 1:3, j in 1:3, k in 1:2])
```
glwagner, line 16:
```suggestion
imap  = on_architecture(arch, [(i, j, k) for i in 1:3, j in 1:3, k in 1:3])
```

src/Utils/kernel_launching.jl
```diff
@@ -440,54 +430,85 @@ end
...
\ No newline at end of file
+end
+
+using KernelAbstractions: CompilerMetadata, NoDynamicCheck
+using CUDA: CUDABackend
+
+import KernelAbstractions: mkcontext, __dynamic_checkbounds, __validindex
+
+# Very dangerous override of mkcontext which will not work if we are not 
+# carefull with making sure that indices are correct when launching a `MappedKernel`
```
glwagner, line 495:
```suggestion
# careful with making sure that indices are correct when launching a `MappedKernel`
```

src/Utils/kernel_launching.jl
simone-silvestri: we need to build the kernel first to be able to retrieve the function

simone-silvestri:
This solution seems to work.

PROS: makes the codebase much simpler and allows to use of `active_cells_map` in many more kernels with very little change
CONS: The solution is a bit convoluted in the `kernel_launching.jl` file

I am in favor of merging, it will make future low-hanging fruit optimization (using active cells map in `compute_diffusivities`, the catke TKE time stepping, and the split explicit free surface) much more straightforward.

src/DistributedComputations/distributed_immersed_boundaries.jl
```diff
@@ -121,23 +121,26 @@ function map_interior_active_cells(ibg::ImmersedBoundaryGrid{<:Any, <:Any, <:Any
...
     include_north = !isa(ibg, YFlatGrid) && (Ry != 1) && !(Ty == LeftConnected)
 
-    west_halo_dependent_cells  = include_west  ? interior_active_indices(ibg; parameters = KernelParameters(x_boundary, left_offsets))    : nothing
-    east_halo_dependent_cells  = include_east  ? interior_active_indices(ibg; parameters = KernelParameters(x_boundary, right_x_offsets)) : nothing
-    south_halo_dependent_cells = include_south ? interior_active_indices(ibg; parameters = KernelParameters(y_boundary, left_offsets))    : nothing
-    north_halo_dependent_cells = include_north ? interior_active_indices(ibg; parameters = KernelParameters(y_boundary, right_y_offsets)) : nothing
+    west_halo_dependent_cells  = interior_active_indices(ibg; parameters = KernelParameters(west_boundary...))
+    east_halo_dependent_cells  = interior_active_indices(ibg; parameters = KernelParameters(east_boundary...))
+    south_halo_dependent_cells = interior_active_indices(ibg; parameters = KernelParameters(south_boundary...))
+    north_halo_dependent_cells = interior_active_indices(ibg; parameters = KernelParameters(north_boundary...))
```
glwagner, line 137: you can also add a constructor for tuples

glwagner (approved):
It's much better! Possibly, @vchuravy may want to look at the implementation

simone-silvestri:
I don't think this is correct, or at least, it works but it is very ad-hoc because it is adding a CUDA-specific method (with the cuda-specific `@device_override` macro which would be, for example, different for `Metal` and `AMDGPU`). This might hamper our possibility of making the code compatible with other backends so I would prefer not to merge it.

@vchuravy, what would be the best way to add something like this to KA? Would adding a new field to the `Kernel` struct (for example `index_map`) work?

simone-silvestri:
Or, maybe I don't completely understand the `@device_override` macro. Would specializing the functions with 
```julia
CUDA.@device_override function
    cuda specific body...
end

Metal.@device_override function
    metal specific body...
end
```
work to specialize the function for different backends?

glwagner:
lots of conflicts here unfortunately. Also it might be nice to get @vchuravy's thoughts here. Since a better implementation here is all the more useful someday when KA is ready

simone-silvestri:
This PR might be able to be merged, I am just not sure why the `OceananigansAMDGPUExt` extension is not precompiling

ext/OceananigansAMDGPUExt.jl
```diff
@@ -2,12 +2,15 @@ module OceananigansAMDGPUExt
...
 using AMDGPU
 using Oceananigans
+using KernelAbstractions: __dynamic_checkbounds, __iterspace
 
 import Oceananigans.Architectures:
     architecture,
     convert_to_device,
     on_architecture
 
+using Ocenanigans.Utils: linear_expand, __linear_ndrange, MappedCompilerMetadata
```
glwagner, line 12:
```suggestion
using Oceananigans.Utils: linear_expand, __linear_ndrange, MappedCompilerMetadata
```

src/Utils/kernel_launching.jl
```diff
@@ -83,6 +86,12 @@ KernelParameters(args::Tuple) = KernelParameters(args...)
 contiguousrange(range::NTuple{N, Int}, offset::NTuple{N, Int}) where N = Tuple(1+o:r+o for (r, o) in zip(range, offset))
 flatten_reduced_dimensions(worksize, dims) = Tuple(d ∈ dims ? 1 : worksize[d] for d = 1:3)
 
+# Internal utility to launch a function mapped on an index_map
+struct MappedFunction{F, M} <: Function
+    func::F
+    index_map::M
+end
```
glwagner, line 93:
```suggestion
    func :: F
    index_map :: M
end
```

src/Utils/kernel_launching.jl
```diff
@@ -83,6 +86,12 @@ KernelParameters(args::Tuple) = KernelParameters(args...)
 contiguousrange(range::NTuple{N, Int}, offset::NTuple{N, Int}) where N = Tuple(1+o:r+o for (r, o) in zip(range, offset))
 flatten_reduced_dimensions(worksize, dims) = Tuple(d ∈ dims ? 1 : worksize[d] for d = 1:3)
 
+# Internal utility to launch a function mapped on an index_map
```
glwagner, line 89: Might make sense to write a longer comment here. `index_map` is the key feature that is utilized by active_cells_map, right?

src/Utils/kernel_launching.jl
```diff
@@ -241,10 +251,18 @@ the architecture `arch`.
...
+    if !isnothing(active_cells_map)
+        func = MappedFunction(loop.f, active_cells_map)
+        loop = Kernel{get_kernel_parameters(loop)..., typeof(func)}(dev, func)
+    end
+
     return loop, worksize
 end
 
-
+@inline get_kernel_parameters(::Kernel{Dev, B, W}) where {Dev, B, W} = Dev, B, W
```
glwagner, line 264: This is poorly named because this file includes a type called `KernelParameters`.

Can you define a constructor for `Kernel` that takes in `loop`, `dev`, `func`?

ext/OceananigansAMDGPUExt.jl
simone-silvestri: that's why 😅

src/Utils/kernel_launching.jl
simone-silvestri: We can do it, It would have to take in `{Dev, B, W}` however, because these are parametric types that do not correspond to a field of `Kernel`.

simone-silvestri:
I guess we can merge this because it seems to work. It complicates `kernel_launching.jl`, but it simplifies the rest of the code

glwagner:
how about benchmarks

simone-silvestri:
Should not play a role in performance, we can check the immersed benchmarks, which are the only ones affected.

simone-silvestri:
In the `diff_immersed_output.txt` there seems to be something wrong in the data scraped by the main build, it looks like it is taking the previous main before the performance improvements of using the reciprocals, maybe updating the branch might fix this issue. However, it seems that this PR does not affect performance looking at the results on this branch and (the correct ones) on main. I ll update the branch to confirm

simone-silvestri:
In the `diff_immersed_output.txt` there seems to be something wrong in the data scraped by the main build, it looks like it is taking the previous main before the performance improvements of using the reciprocals, maybe updating the branch might fix this issue. However, it seems that this PR does not affect performance looking at the results on this branch and (the correct ones) on main. I ll update the branch to confirm

simone-silvestri:
In the `diff_immersed_output.txt` there seems to be something wrong in the data scraped by the main build, it looks like it is taking the previous main before the performance improvements of using the reciprocals, maybe updating the branch might fix this issue. However, it seems that this PR does not affect performance looking at the results on this branch and (the correct ones) on main. I ll update the branch to confirm

PR #3922, opened 2024-11-13
Remove `fill_halo_regions!` specific kwargs form `launch!`

simone-silvestri:
The `only_local_halos` and `async` kwargs are specific to fill halos and in particular they are active only for distributed grids. So in this PR I remove them from launch and move them in the BoundaryConditions module

### DIFF
--- a/src/BoundaryConditions/fill_halo_regions.jl
+++ b/src/BoundaryConditions/fill_halo_regions.jl
@@ -64,7 +64,9 @@ function fill_halo_regions!(c::MaybeTupledData, boundary_conditions, indices, lo
     return nothing
 end
 
-function fill_halo_event!(c, fill_halos!, bcs, indices, loc, arch, grid, args...; kwargs...)
+function fill_halo_event!(c, fill_halos!, bcs, indices, loc, arch, grid, args...; 
+                          async = false, # This kwargs is specific to DistributedGrids, here is does nothing
+                          kwargs...)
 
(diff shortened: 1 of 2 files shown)


src/BoundaryConditions/fill_halo_regions_periodic.jl
```diff
@@ -15,19 +15,19 @@ end
...
     launch!(arch, grid, KernelParameters(yz_size, offset), fill_periodic_west_and_east_halo!, c_parent, Val(grid.Hx), grid.Nx; kw...)
     return nothing
 end
 
-function fill_south_and_north_halo!(c, ::PBCT, ::PBCT, size, offset, loc, arch, grid, args...; kw...)
+function fill_south_and_north_halo!(c, ::PBCT, ::PBCT, size, offset, loc, arch, grid, args...; only_local_halos = false, kw...)
     c_parent, xz_size, offset = parent_size_and_offset(c, 1, 3, size, offset)
     launch!(arch, grid, KernelParameters(xz_size, offset), fill_periodic_south_and_north_halo!, c_parent, Val(grid.Hy), grid.Ny;  kw...)
     return nothing
 end
```
glwagner, line 28: `only_local_halos` isn't used in the function

glwagner: Oh I see it's only for distributed

glwagner (approved):

Issue #3917, opened 2024-11-12
Some Lagrangian particles like to travel Southwest on an immersed lat-lon grid + hydrostatic model


ali-ramadhan:
So I'm trying to advect some particles in a `HydrostaticFreeSurfaceModel` with an immersed `LatitudeLongitudeGrid`.

I noticed that some particles travel really fast in the southwest direction and get stuck at the southwestern corner. Depending on the dynamics and the immersed boundary, tons of particles move this way or very few do. The particles behave fine once you remove the immersed boundary.

I was able to reduce the issue down to a somewhat lengthy MWE. I add an immersed seamount and some particles (with a zero coefficient of restitution) and impose a surface momentum flux to get them moving. I use WENO to stabilize the simulation. It takes a few days for the issue to show up then I plot the particles at high temporal resolution so their trajectories are clear.

I'm digging into the code to see what the issue might, but if anyone has any ideas I'd appreciate it!

I suspect it has something to do with the particles bouncing off bathymetry incorrectly. Will try running on the CPU and with a small coefficient of restitution to see what happens.

---

MWE:

```julia
using Printf
using Distributions
using JLD2
using CairoMakie

using Oceananigans
using Oceananigans.Units

using Oceananigans.Architectures: on_architecture
using Oceananigans.OutputReaders: iterations_from_file

# Grid setup

arch = GPU()

Nλ, Nφ, Nz = 100, 100, 32

Lλ = Lφ = 1
H = 100

underlying_grid = LatitudeLongitudeGrid(
    arch;
    topology = (Bounded, Bounded, Bounded),
    size = (Nλ, Nφ, Nz),
    longitude = (-Lλ, Lλ),
    latitude = (-Lφ, Lφ),
    z = (-H, 0),
    halo = (5, 5, 5)
)

height = H/2
width = Lλ / 3
mount(λ, φ) = height * exp(-λ^2 / 2width^2) * exp(-φ^2 / 2width^2)
bottom(λ, φ) = -H + mount(λ, φ)

grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom))

# Boundary conditions

v_bcs = FieldBoundaryConditions(
      top = FluxBoundaryCondition(-1e-5)
)

boundary_conditions = (; v = v_bcs)

# Particle setup

Np = 100000 # lots of particles so the issue is clear

ε = 0.1
λ_particles = 0.0
φ_particles = 0.5

[...]

PR #3923, opened 2024-11-13
Fix bug in `interpolate!` for regular grids

simone-silvestri:
I have also added a very simple unit test. Probably we need some more,

### DIFF
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -68,14 +68,14 @@ end
     x₀ = xnode(1, 1, 1, grid, locs...)
     Δx = xspacings(grid, locs...)
     FT = eltype(grid)
-    return convert(FT, (x - x₀) / Δx)
+    return convert(FT, (x - x₀) / Δx) + 1 # 0 - based indexing so we sum 1
 end
 
 @inline function fractional_x_index(λ, locs, grid::XRegularLLG)
     λ₀ = λnode(1, 1, 1, grid, locs...)
     Δλ = λspacings(grid, locs...)
     FT = eltype(grid)
-    return convert(FT, (λ - λ₀) / Δλ)
+    return convert(FT, (λ - λ₀) / Δλ) + 1 # 0 - based indexing so we sum 1
 end
 
 @inline function fractional_x_index(x, locs, grid::RectilinearGrid)
@@ -100,14 +100,14 @@ end
     y₀ = ynode(1, 1, 1, grid, locs...)
     Δy = yspacings(grid, locs...)
     FT = eltype(grid)
-    return convert(FT, (y - y₀) / Δy)
+    return convert(FT, (y - y₀) / Δy) + 1 # 0 - based indexing so we sum 1
 end
 
 @inline function fractional_y_index(φ, locs, grid::YRegularLLG)
     φ₀ = φnode(1, 1, 1, grid, locs...)
     Δφ = φspacings(grid, locs...)
     FT = eltype(grid)
-    return convert(FT, (φ - φ₀) / Δφ)
+    return convert(FT, (φ - φ₀) / Δφ) + 1 # 0 - based indexing so we sum 1
 end
 
 @inline function fractional_y_index(y, locs, grid::RectilinearGrid)
@@ -115,7 +115,7 @@ end
      Ty = topology(grid, 2)()
      Ny = length(loc, Ty, grid.Ny)
      yn = ynodes(grid, locs...)
-    return fractional_index(y, yn, Ny) - 1
+    return fractional_index(y, yn, Ny) - 1 
 end
 
 @inline function fractional_y_index(y, locs, grid::LatitudeLongitudeGrid)
@@ -133,7 +133,7 @@ ZRegGrid = Union{ZRegularRG, ZRegularLLG, ZRegOrthogonalSphericalShellGrid}
 @inline function fractional_z_index(z::FT, locs, grid::ZRegGrid) where FT
     z₀ = znode(1, 1, 1, grid, locs...)
     Δz = zspacings(grid, locs...)
-    return convert(FT, (z - z₀) / Δz)
+    return convert(FT, (z - z₀) / Δz) + 1 # 0 - based indexing so we sum 1
 end
 
 @inline function fractional_z_index(z, locs, grid)
@@ -144,6 +144,11 @@ end
     return fractional_index(z, zn, Nz) - 1
 end
 
+flip(t::Tuple) = map(flip, t)
+flip(::Nothing) = nothing
+flip(::Center)  = Face()
+flip(::Face)    = Center() 
+
 """
     fractional_indices(x, y, z, grid, loc...)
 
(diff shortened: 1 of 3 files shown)


src/Fields/interpolate.jl
```diff
@@ -68,30 +68,30 @@ end
     x₀ = xnode(1, 1, 1, grid, locs...)
     Δx = xspacings(grid, locs...)
     FT = eltype(grid)
-    return convert(FT, (x - x₀) / Δx)
+    return convert(FT, (x - x₀) / Δx) # 0 - based indexing!
```
glwagner, line 71:
```suggestion
    return convert(FT, (x - x₀) / Δx) + 1
```

why isn't this the fix?

glwagner:
Why don't we change `fractional_indices` to always use 1-based indexing? Isn't that the straightforward thing to do?

simone-silvestri:
I am trying option 2. I also think it might be the cleaner option. Let's see if something breaks

src/Fields/interpolate.jl
```diff
@@ -68,30 +68,30 @@ end
     x₀ = xnode(1, 1, 1, grid, locs...)
     Δx = xspacings(grid, locs...)
     FT = eltype(grid)
-    return convert(FT, (x - x₀) / Δx) # 0 - based indexing!
+    return convert(FT, (x - x₀) / Δx) + 1 # 0 - based indexing so we add 1
```
glwagner, line 71:
```suggestion
    return convert(FT, (x - x₀) / Δx) + 1 # 1-based indexing
```

Just a nitpick, but I think it's better to simply state that this is 1-based

glwagner:
> However, as it stands now, we probably support extrapolation (because fractional_index can be negative), so we would lose that behavior.

Why does 0-based indexing support extrapolation when 1-based does not?

ali-ramadhan:
Thanks so much for the fix @simone-silvestri! I'm guessing the PR is not ready for testing? Happy to help here if I can.

I tried running the MWE from https://github.com/CliMA/Oceananigans.jl/issues/3917#issuecomment-2470956042 which now throws a GPU exception at the first time step, caught with `julia -g2`:

```
iteration: 0, time: 0 seconds, Δt: 1.100 seconds, U_max=(4.66e-03, 4.59e-03, 5.99e-05)
[ Info:     ... simulation initialization complete (15.507 seconds)
[ Info:     ... initial time step complete (2.446 seconds).
ERROR: a undefined variable error was thrown during kernel execution on thread (65, 1, 1) in block (45, 1, 1).
Stacktrace:
 [1] fractional_z_index at /home/alir/atdepth/Oceananigans.jl/src/Fields/interpolate.jl:137
 [2] fractional_z_index at /home/alir/atdepth/Oceananigans.jl/src/ImmersedBoundaries/immersed_boundary_nodes.jl:65
 [3] _fractional_indices at /home/alir/atdepth/Oceananigans.jl/src/Fields/interpolate.jl:169
 [4] fractional_indices at /home/alir/atdepth/Oceananigans.jl/src/Fields/interpolate.jl:156
 [5] advect_particle at /home/alir/atdepth/Oceananigans.jl/src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl:113
 [6] macro expansion at /home/alir/atdepth/Oceananigans.jl/src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl:187
 [7] gpu__advect_particles! at /home/alir/.julia/packages/KernelAbstractions/iW1Rw/src/macros.jl:97
 [8] gpu__advect_particles! at ./none:0

ERROR: LoadError: KernelException: exception thrown during kernel execution on device NVIDIA GeForce RTX 4090
Stacktrace:
  [1] check_exceptions()
    @ CUDA ~/.julia/packages/CUDA/2kjXI/src/compiler/exceptions.jl:39
  [2] synchronize(stream::CUDA.CuStream; blocking::Bool, spin::Bool)
    @ CUDA ~/.julia/packages/CUDA/2kjXI/lib/cudadrv/synchronization.jl:207
  [3] synchronize (repeats 2 times)
    @ ~/.julia/packages/CUDA/2kjXI/lib/cudadrv/synchronization.jl:194 [inlined]
  [4] (::CUDA.var"#1127#1128"{Float64, Vector{Float64}, Int64, CUDA.CuArray{Float64, 3, CUDA.DeviceMemory}, Int64, Int64})()
    @ CUDA ~/.julia/packages/CUDA/2kjXI/src/array.jl:554
  [5] #context!#990
    @ ~/.julia/packages/CUDA/2kjXI/lib/cudadrv/state.jl:168 [inlined]
  [6] context!
    @ ~/.julia/packages/CUDA/2kjXI/lib/cudadrv/state.jl:163 [inlined]
  [7] unsafe_copyto!(dest::Vector{Float64}, doffs::Int64, src::CUDA.CuArray{Float64, 3, CUDA.DeviceMemory}, soffs::Int64, n::Int64)
    @ CUDA ~/.julia/packages/CUDA/2kjXI/src/array.jl:550
  [8] copyto!
    @ ~/.julia/packages/CUDA/2kjXI/src/array.jl:503 [inlined]
  [9] getindex
    @ ~/.julia/packages/GPUArrays/qt4ax/src/host/indexing.jl:52 [inlined]
 [10] getindex
    @ ~/.julia/packages/OffsetArrays/hwmnB/src/OffsetArrays.jl:440 [inlined]
 [11] _getindex
    @ ./abstractarray.jl:1323 [inlined]
 [12] getindex
    @ ./abstractarray.jl:1290 [inlined]
 [13] getindex
    @ ~/atdepth/Oceananigans.jl/src/Fields/field.jl:401 [inlined]
 [14] first(a::Field{Nothing, Nothing, Nothing, Nothing, ImmersedBoundaryGrid{…}, Tuple{…}, OffsetArrays.OffsetArray{…}, Float64, FieldBoundaryConditions{…}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{…}})
    @ Base ./abstractarray.jl:452
 [15] macro expansion
    @ ~/.julia/packages/GPUArraysCore/GMsgk/src/GPUArraysCore.jl:210 [inlined]
 [16] maximum(f::Function, c::Field{Face, Center, Center, Nothing, ImmersedBoundaryGrid{…}, Tuple{…}, OffsetArrays.OffsetArray{…}, Float64, FieldBoundaryConditions{…}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{…}}; condition::Nothing, mask::Float64, dims::Function)
    @ Oceananigans.Fields ~/atdepth/Oceananigans.jl/src/Fields/field.jl:703
 [17] maximum(f::Function, c::Field{Face, Center, Center, Nothing, ImmersedBoundaryGrid{…}, Tuple{…}, OffsetArrays.OffsetArray{…}, Float64, FieldBoundaryConditions{…}, Nothing, Oceananigans.Fields.FieldBoundaryBuffers{…}})
    @ Oceananigans.Fields ~/atdepth/Oceananigans.jl/src/Fields/field.jl:689
 [18] progress(sim::Simulation{HydrostaticFreeSurfaceModel{…}, Float64, Float64, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}})
    @ Main ~/atdepth/Oceananigans.jl/3917.jl:84
 [19] (::Callback{Nothing, typeof(progress), IterationInterval, Oceananigans.TimeStepCallsite})(sim::Simulation{HydrostaticFreeSurfaceModel{…}, Float64, Float64, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}})
    @ Oceananigans.Simulations ~/atdepth/Oceananigans.jl/src/Simulations/callback.jl:15
 [20] time_step!(sim::Simulation{HydrostaticFreeSurfaceModel{…}, Float64, Float64, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}})
    @ Oceananigans.Simulations ~/atdepth/Oceananigans.jl/src/Simulations/run.jl:143
 [21] run!(sim::Simulation{HydrostaticFreeSurfaceModel{…}, Float64, Float64, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}}; pickup::Bool)
    @ Oceananigans.Simulations ~/atdepth/Oceananigans.jl/src/Simulations/run.jl:97
 [22] run!(sim::Simulation{HydrostaticFreeSurfaceModel{…}, Float64, Float64, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}, OrderedCollections.OrderedDict{…}})
    @ Oceananigans.Simulations ~/atdepth/Oceananigans.jl/src/Simulations/run.jl:85
 [23] top-level scope
    @ ~/atdepth/Oceananigans.jl/3917.jl:105
 [24] include(fname::String)
    @ Base.MainInclude ./client.jl:494
 [25] top-level scope
    @ REPL[1]:1
in expression starting at /home/alir/atdepth/Oceananigans.jl/3917.jl:105
```

src/Fields/interpolate.jl
```diff
@@ -132,16 +134,16 @@ ZRegGrid = Union{ZRegularRG, ZRegularLLG, ZRegOrthogonalSphericalShellGrid}
 
 @inline function fractional_z_index(z::FT, locs, grid::ZRegGrid) where FT
     z₀ = znode(1, 1, 1, grid, locs...)
-    Δz = @inbounds first(zspacings(grid, locs...))
-    return convert(FT, (z - z₀) / Δz)
+    Δz = Δz(1, 1, 1, grid, locs...)
```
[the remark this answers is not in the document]
simone-silvestri: I think we should use the pointwise operators defined in the operator module instead of the `zspacing` function and so on. (there is a conflict in the name here though, I ll correct it)

simone-silvestri:
> Looking at the latest Buildkite build, I think one or some of the tests are failing because `Δxᶠᵃᶠ` is not defined in `Operators/spacings_and_areas_and_volumes.jl` following PR #3143.
> 
> https://buildkite.com/clima/oceananigans/builds/19175#01934dbf-8dcf-471b-a3dc-ffecd42006df/18-582

Hmmm, shall we define it here in this PR?

simone-silvestri:
The problem is that `Δxᶠᵃᶠ` does not exist for a curvilinear grid, so it is difficult to define. A solution is to use just three-dimensional spacings always with just `ᶠ` and `ᶜ`. A `nothing` location can point to `ᶜ`.

simone-silvestri:
> In #3143 we also decided that a `nothing` location would point to `ᵃ`, at least for the purposes of spacings. If we want to stick to this then the better long-term solution might be to try and always use spacings with just `ᶠ` and `ᶜ`.

I see. I think the problem is that `ᵃ` is well-defined only for `Flat` directions and is almost always ambiguous for three-dimensional grids. There is also a problem with immersed boundary grids (for example, partial cells where dz is always three-dimensional). 

In my opinion `ᵃ` should be a placeholder for when the location does not matter (as for rectilinear grids for example).

glwagner:
We should never use ^a in numerics. Only for grid-specific operators.

simone-silvestri:
I understand now the role of nothing pointing to `a`. If it points to `a` it will spit an error for those grids where the `a` is ambiguous. It makes sense because it is an extra check that we don't do anything incorrect.

I defined the operators that were not defined and the tests now pass

glwagner (approved):

glwagner:
solid

PR #3924, opened 2024-11-14
Reduce parameter space for velocity tendency kernels

simone-silvestri:
If we use complicated forcings in the model, for example the [ECCORestoring](https://github.com/CliMA/ClimaOcean.jl/blob/05e6de31ba58594775f739cca47e6ae0ffbf3385/src/DataWrangling/ECCO/ECCO_restoring.jl#L235) type we use in ClimaOcean, we could incur in parameter space by passing all the forcings in the kernels, see:
https://buildkite.com/clima/climaocean-examples/builds/942#01932566-3781-41ab-9464-031b0bfcf562
and
https://github.com/CliMA/ClimaOcean.jl/pull/244

This PR should reduce the parameter space in those cases by passing only the u and v specific forcings.


glwagner:
This is nice, but can you implement this uniformally for all models? It'll be faster to do it now, rather than run into this issue in the future and be scratching our heads because we forgot about this

simone-silvestri:
I suspect that this PR is slowing down the code. I ll investigate

glwagner (approved):

src/Models/HydrostaticFreeSurfaceModels/compute_hydrostatic_free_surface_tendencies.jl
```diff
@@ -137,20 +137,19 @@ function compute_hydrostatic_momentum_tendencies!(model, velocities, kernel_para
...
-                                model.forcing,
                                 model.clock)
 
     u_kernel_args = tuple(start_momentum_kernel_args..., u_immersed_bc, end_momentum_kernel_args...)
     v_kernel_args = tuple(start_momentum_kernel_args..., v_immersed_bc, end_momentum_kernel_args...)
 
     launch!(arch, grid, kernel_parameters,
             compute_hydrostatic_free_surface_Gu!, model.timestepper.Gⁿ.u, grid, 
-            active_cells_map, u_kernel_args;
+            active_cells_map, u_kernel_args, model.forcing.u;
```
glwagner, line 147:
```suggestion
            active_cells_map, tuple(u_kernel_args..., model.forcing.u);
```

glwagner: but it doesnt matter

PR #3928, opened 2024-11-15
Support interpolation from triple-Flat grids

glwagner:
This is helpful for very simple simulations in ClimaOcean where the PrescribedAtmosphere is represented by a 0D `FieldTimeSeries`.


src/OutputReaders/field_time_series_indexing.jl
```diff
@@ -168,13 +168,17 @@ end
...
 
-    # Build space interpolators
-    to_node = flatten_node(to_node...)
-    ii, jj, kk = fractional_indices(to_node, from_grid, from_loc...)
-
-    ix = interpolator(ii)
-    iy = interpolator(jj)
-    iz = interpolator(kk)
+    if topology(from_grid) === (Flat, Flat, Flat)
+        ix = iy = iz = (1, 1, 0)
```
simone-silvestri, line 172:
```suggestion
        ix = iy = iz = (1, 1, 0)
```

why is `iz = 0`?

glwagner: `iz` is not 0, `iz = (1, 1, 0)`. And `ix = iy = iz`.

glwagner: The third element is the "fractional index displacement", its between 0 and 1 generally, but for this trivial situation it is irrelevant and we set it to 0.

PR #3930, opened 2024-11-15
RK3 for the HydrostaticFreeSurfaceModel version 2

simone-silvestri:
This PR is a cleaner working remake of PR #3874

I copy here the description of that PR:

RK3 paired with a splitting of the barotropic- baroclinic mode described in 
https://www.sciencedirect.com/science/article/abs/pii/S0021999122001127?fr=RR-2&ref=pdf_download&rr=8d81dd3bab0dba9a which seems to ensure consistency in the forcing of the barotropic and baroclinic mode. 

The algorithm follows:

***substep 1***
- $u^{*, n+1/2} = u^{n} + \Delta t G(u^{n})$
- $U^{n+1/2} = \text{barotropic-step}\left( u^n, \eta^n, G(u^{n}) \right)$
- $u^{n+1/2} = \text{barotropic-correction}\left(u^{*, n+1/2}, U^{n+1/2}  \right)$

***substep 2***
- $u^{*, n+1/3} = \frac{3}{4}u^{n} + \frac{1}{4}\left(u^{n+1/2} + \Delta t G(u^{n+1/2}) \right)$
- $U^{*, n+1/3} = \text{barotropic-step}\left(u^{n+1/2}, \eta^{n+1/2}, G(u^{n+1/2})  \right)$
- $U^{n+1/3} = \frac{3}{4}U^{n} + \frac{1}{4}U^{*, n+1/3}$
- $u^{n+1/3} = \text{barotropic-correction}\left(u^{*, n+1/3}, U^{n+1/3} \right)$

***substep 3***
- $u^{*, n+1} = \frac{1}{3}u^{n} + \frac{2}{3}\left(u^{n+1/3} + \Delta t G(u^{n+1/3})\right)$
- $\overline{G}^{n+1} = \frac{1}{6}G(u^{n}) + \frac{1}{6}G(u^{n+1/2}) + \frac{2}{3}G(u^{n+1/3})$
- $U^{n+1} = \text{barotropic-step}\left(u^{n}, \eta^{n}, \overline{G}^{n+1} \right)$ 
- $u^{n+1} = \text{barotropic-correction}\left(u^{*, n+1}, U^{n+1} \right)$

The proof of concept in this PR is working and allowing a much larger CFL than the AB2 code we use (with the added benefit of reducing the implicit diffusion of the AB2 scheme and being self-starting from time-step `n`).

The cost of RK3 vs AB2 is slightly more than 3 times because the barotropic solve is done 3 times on the entire 
baroclinic time step. So in general, speed-wise RK3 ≈ AB2.


glwagner:
> The cost of RK3 vs AB2 is slightly more than 3 times because the barotropic solve is done 3 times on the entire
baroclinic time step. So in general, speed-wise RK3 ≈ AB2.

Do you have any thoughts about speeding it up? For example, doing one barotropic solve over the whole time-step a la Le and Moin 1991 (for example)

simone-silvestri:
We could study what they do in NEMO. I know they do only one barotropic solve instead of 3 and interpolate the free surface in between the substeps.

glwagner (approved):

src/TimeSteppers/split_hydrostatic_runge_kutta_3.jl
```diff
@@ -0,0 +1,175 @@
+using Oceananigans.Architectures: architecture
+using Oceananigans: fields
+
+"""
+    SplitRungeKutta3TimeStepper{FT, TG} <: AbstractTimeStepper
+
+Holds parameters and tendency fields for a low storage, third-order Runge-Kutta-Wray
+time-stepping scheme described by [LeMoin1991](@citet).
```
glwagner, line 8: LeMoin1991?

src/TimeSteppers/split_hydrostatic_runge_kutta_3.jl
```diff
@@ -0,0 +1,175 @@
...
+end
+
+"""
+    SplitRungeKutta3TimeStepper(grid, tracers;
+                                implicit_solver = nothing,
+                                Gⁿ = TendencyFields(grid, tracers),
+                                G⁻ = TendencyFields(grid, tracers),
+                                S⁻ = TendencyFields(grid, tracers))
+
+Return a 3rd-order SplitRungeKutta3 timestepper (`RungeKutta3TimeStepper`) on `grid` and with `tracers`.
```
glwagner, line 28:
```suggestion
Return a 3rd-order `SplitRungeKutta3TimeStepper` on `grid` and with `tracers`.
```

ali-ramadhan:
Super excited to try this out! @simone-silvestri Do you think hydrostatic RK3 needs more testing? If so, I'm happy to open an issue to track and implement a few tests. I know the test suite build time is also a concern.

PR #3950, opened 2024-11-22
Simplify active cells kernels

simone-silvestri:
I think #3920 is a bit complicated to achieve and probably should be done in KA. In the meantime, another way to write just one kernel is to define a new `@active_index` macro that accepts the map as an argument.


simone-silvestri:
Ok I'll close this so that we figure out how to do it correctly

PR #3952, opened 2024-11-22
Fix positivity preserving WENO

simone-silvestri:
It was broken so this PR fixes it. This method is untested and unvalidated so adding tests is probably not required right now.

### DIFF
--- a/src/Advection/weno_reconstruction.jl
+++ b/src/Advection/weno_reconstruction.jl
@@ -104,6 +104,10 @@ function WENO(FT::DataType=Float64;
 
     mod(order, 2) == 0 && throw(ArgumentError("WENO reconstruction scheme is defined only for odd orders"))
 
+    if !isnothing(bounds)
+        @warn "Bounds preserving WENO is experimental and not tested. Use with caution."
+    end
+
     if order < 3
         # WENO(order=1) is equivalent to UpwindBiased(order=1)
         return UpwindBiased(FT; order=1)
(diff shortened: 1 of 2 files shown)


src/Advection/weno_reconstruction.jl
```diff
@@ -104,6 +104,10 @@ function WENO(FT::DataType=Float64;
 
     mod(order, 2) == 0 && throw(ArgumentError("WENO reconstruction scheme is defined only for odd orders"))
 
+    if !isnothing(bounds)
+        @warn "Bounds preserving WENO is experimental and not tested. Use with caution."
```
glwagner, line 108:
```suggestion
        @warn "Bounds preserving WENO is experimental."
```

glwagner (approved):

glwagner:
@masonrogers14 try to run your case with this branch and see what happens!

masonrogers14:
Thanks for flagging me--so far on my trivial test case (Gaussian patch of tracer in constant flow), negative values do appear, but they do not seem to grow nearly as quickly as before. I'll keep you posted as I scale up to actual fluid dynamics

glwagner:
_actual_ fluid dynamics 🔥

glwagner:
Hummm I wonder if we can also use a nudging term for negative values in the tracer equation. Also I am not sure where this ends up being a problem but one can try to clip negative values where they are used. For example, CATKE does this for the TKE.

PR #3953, opened 2024-11-22
(0.94.4) `PolarBoundaryCondition` for tracer + views for AbstractOperations

simone-silvestri:
A `FluxBoundaryCondition` is not correct on a latitude-longitude grid that covers the whole earth, i.e. when `latitude = (-90, 90)`. 
 
The actual boundary condition (for tracers) should be a value boundary condition that holds the average of the last / first row.

Vectors are a little more complicated because we need to account for the frame of reference.
(We can probably do that in a new PR)

This is required mainly for input data define on a `LatitudeLongitudeGrid` that needs interpolation on full spherical grids for global ocean simulations.

As part of this PR, to allow reducing on the last / first row of the `LatitudeLongitudeGrid` I needed to implement reductions on windowed reduced fields, which required extending `views` for `AbstractOperations`

The implementation is not final since it does not account for the case of an array (non-field) being part of an `AbstractOperation`. But I guess we can tackle that case when we encounter it.

In a sense this is a bugfix, since boundary conditions for `LatitudeLongitudeGrid`s were incorrect at the north and south for grids that reach to pole, so I will bump the minor version

### DIFF
--- a/src/BoundaryConditions/polar_boundary_condition.jl
+++ b/src/BoundaryConditions/polar_boundary_condition.jl
@@ -1,10 +1,10 @@
-const PolarBoundaryCondition{V} = BoundaryCondition{<:Value, <:PolarValue}
-
 condition_operand(data, grid, Loc, condition, mask) = data
 
 struct PolarValue{C}
     c :: C
 end
+
+const PolarBoundaryCondition{V} = BoundaryCondition{<:Value, <:PolarValue}
 
 PolarBoundaryCondition(field) = 
     ValueBoundaryCondition(PolarValue(field))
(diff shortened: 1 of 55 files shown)


glwagner (approved):

src/BoundaryConditions/polar_boundary_condition.jl
```diff
@@ -0,0 +1,76 @@
...
+    end
+end
+
+# North - South flux boundary conditions are not valid on a Latitude-Longitude grid if the last / first rows represent the poles
+function regularize_south_boundary_condition(bc::DefaultBoundaryCondition, grid::LatitudeLongitudeGrid, loc, dim, args...) 
+    φmin = φnode(1, grid, Face()) 
+
+    if φmin == - 90 
+        _, LY, LZ = loc
+        field = Field{Nothing, LY, LZ}(grid; indices = (:, 1, :))
```
glwagner, line 34: This does seem right

glwagner: But should `LY` actually be flipped? Eg this is Face for a Center location. (It may not matter)

simone-silvestri: Probably you are right. I have found out unfortunately that reductions on windowed fields do not work. I ll try to make it work so here we can do
```julia
mean!(windowed_field, full_field`
```

simone-silvestri:
In the end, using fields in boundary conditions is a bit too complicated because fields are defined after boundary conditions making it a bit of a mess. For this reason, I just wrote a very simple averaging kernel in the polar boundary conditions file.

The improvements for reductions on windowed fields still stand though, so I have added some tests also to make sure that reductions on windowed fields work.

PR #3955, opened 2024-11-23
Refactor of spacings and correction of spacings-connected bugs

### LINKED ISSUE #3701: `xspacings` is broken for `ImmersedBoundaryGrid`
MWE

```julia
using Oceananigans

grid = RectilinearGrid(size = (128, 128, 1),
                       x = (-5, 5),
                       y = (-5, 5),
                       z = (0, 1),

island(x, y) = (x^2 + y^2) < 1
grid = ImmersedBoundaryGrid(grid, GridFittedBottom(island))

Δx = xspacings(grid, Center(), Center(), Center())
```

produces

```julia
julia> include("hydrostatic_turbulence.jl")
ERROR: LoadError: MethodError: no method matching xspacings(::ImmersedBoundaryGrid{…}, ::Center; with_halos::Bool)

Closest candidates are:
  xspacings(::Any, ::Any, ::Any, ::Any; with_halos)
   @ Oceananigans ~/.julia/packages/Oceananigans/OMBY0/src/Grids/nodes_and_spacings.jl:200
  xspacings(::LatitudeLongitudeGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:Number, <:Number}, ::Center, ::Center; with_halos)
   @ Oceananigans ~/.julia/packages/Oceananigans/OMBY0/src/Grids/latitude_longitude_grid.jl:674
  xspacings(::LatitudeLongitudeGrid, ::Center, ::Center; with_halos)
   @ Oceananigans ~/.julia/packages/Oceananigans/OMBY0/src/Grids/latitude_longitude_grid.jl:658
  ...

Stacktrace:
 [1] xspacings(grid::ImmersedBoundaryGrid{…}, ℓx::Center, ℓy::Center, ℓz::Center)
   @ Oceananigans.Grids ~/.julia/packages/Oceananigans/OMBY0/src/Grids/nodes_and_spacings.jl:200
```

This is probably also a problem for `yspacings` and `zspacings`, but I'm not sure.

For `GridFittedBottom`, the spacings should just return spacings for the underlying grid (right now only `PartialCellBottom` immersed boundaries have different metrics than the underlying grid).

### LINKED ISSUE #3954: `minimum_xspacing` does not account for the immersed boundary
I think it's because of these definitions

https://github.com/CliMA/Oceananigans.jl/blob/504ab5da6c1cac49f0b5c364e58c117563bde0e6/src/ImmersedBoundaries/immersed_boundary_nodes.jl#L67-L69

This causes the CFL to be limited by possibly immersed nodes.
It is quite relevant on the `TripolarGrid` where nodes near the north poles are immersed with the spacings tending to zero.

I can open a PR to fix this.

### LINKED ISSUE #3957: Duplicate `ψspacing` and `Δψ`
At the moment we have a duplication of spacing functions: we have 
```julia
xspacing
yspacing
zspacing
```
and 
```julia
Δx
Δy
Δz
```

that appear to be doing the same thing. I think this is because the `Δ` versions are required for timestepping and the `spacing` functions are not used but defined because we have other spacings (such as `λspacing`) that we define but are not needed for timestepping.

I propose to remove all the `ψspacing` functions and keep the pointwise versions defined in the Operators module. We can add `Δλ` and `Δφ` for completeness in the Operators module if we want

simone-silvestri:
Correct spacings on an `ImmersedBoundaryGrid`
closes #3954
closes #3701

~~Should correct also partial cell bottom~~

Unification of general grid metrics
closes #3957


glwagner (approved):

simone-silvestri:
I was thinking of using this PR to tackle also issue #3957. What do people think about it?
@ali-ramadhan @navidcy @glwagner

simone-silvestri:
The PartialCellBottom example crashes with this bugfix. I suggest I just revert this PR not to close #3958 and we can tackle that in a new PR if that's ok.

simone-silvestri:
In this PR there is a suggestion to improve the internals following #3957. 
The exported functions are only the verbose ones, while the "mathy" ones are left for internal use.
Let me know what you guys think @ali-ramadhan @glwagner @tomchor

tomchor:
> The PartialCellBottom example crashes with this bugfix. I suggest I just revert this PR not to close #3958 and we can tackle that in a new PR if that's ok.

Sounds good to me!

> In this PR there is a suggestion to improve the internals following #3957. The exported functions are only the verbose ones, while the "mathy" ones are left for internal use. Let me know what you guys think @ali-ramadhan @glwagner @tomchor

I'm also okay with this. Is this a breaking change?

simone-silvestri:
> I'm also okay with this. Is this a breaking change?

I think it's not breaking. The metrics that were defined in `AbstractOperations` to be part of abstract operations are removed in this PR so what was
```julia
using Oceananigans.AbstractOperations: Ax
c = CenterField(grid)
cAx = c * Ax
```
cane be done in this PR with 
```julia
using Oceananigans.Operators: Ax
c = CenterField(grid)
cAx = c * Ax
```
However, since `AbstractOperations` imports the metric functions, the first version is still valid, despite the actual objects being different.

src/AbstractOperations/grid_metrics.jl
```diff
@@ -55,94 +73,15 @@ BinaryOperation at (Center, Center, Center)
 ├── grid: 1×1×1 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 1×1×1 halo
 └── tree:
     * at (Center, Center, Center)
-    ├── 1×1×1 Field{Center, Center, Center} on RectilinearGrid on CPU
-    └── Δzᶜᶜᶜ at (Center, Center, Center)
+   ├── 1×1×1 Field{Center, Center, Center} on RectilinearGrid on CPU
+   └── Δzᶜᶜᶜ at (Center, Center, Center)
```
glwagner, line 77: isn't this missing a space

glwagner:
What's the advantage of getting rid of `xspacing`? I guess the idea was that users should use this in forcing functions etc to generate more readable code. Did the original logic behind defining the verbose versions change?

glwagner:
> I was thinking of using this PR to tackle also issue #3957. What do people think about it? @ali-ramadhan @navidcy @glwagner @tomchor

Better to have different issues in different PRs, as usual.

simone-silvestri:
> What's the advantage of getting rid of `xspacing`? I guess the idea was that users should use this in forcing functions etc to generate more readable code. Did the original logic behind defining the verbose versions change?

In the end, this PR does not change the user interface, except for adding `xarea`, `yarea`, and `zarea` to be consistent with the verbose spacings. Most of the work is refactoring the internals of the code to make it better organized and extendable which includes:
- use only the `Δ*` functions in the internal functions and export the `*spacing` functions for use in scripts where the `*spacing` function is defined only once in the `Operators` module as `*spacing(args...) = Δ*(args...)`.
- remove the conflicting definition of `Δ*` in `AbstractOperations` since we can use the `Operators` version
- remove extra redundant definitions here and there, for example, `Δz` does not depend on the grid type so define it only once
- Treat `Δλ` and `Δφ` as the other spacings (which involves defining the `Δλᵃᵇᶜ` functions and then the general `Δλ(i, j, k, grid, ℓx, ℓy, ℓz)` functions)
- move all the spacings definitions to the `spacings_and_areas.jl` file in the `Operators` module.

PR #3956, opened 2024-11-25
(0.95.7) Yet another ZStar implementation

simone-silvestri:
## Refactoring

Following #3411 it is clear that the grids require some refactor to allow ZStar. 
This PR implements another proposal for ZStar which hinges on changing the grids to have just one `z` field.
The case of a static z (the only allowed in Oceananigans at the moment) is covered with a `StaticVerticalCoordinate` type that contains all the specifications that were in the grid before:
https://github.com/CliMA/Oceananigans.jl/blob/9e05e21921affdf3c8bdeb5a1f58da926550e907/src/Grids/vertical_coordinate.jl#L14-L19
This refactor should change nothing in the user interface nor the internals of the code, since the whole API should depend on `znode` and `Δz`, but just reorganizes the vertical coordinate in a compact type that can be modified.

## ZStar Implementation

With this refactor, a ZStar coordinate is implemented through a `ZStarVerticalCoordinate` type
https://github.com/CliMA/Oceananigans.jl/blob/9e05e21921affdf3c8bdeb5a1f58da926550e907/src/Grids/vertical_coordinate.jl#L36-L48
`e₃` is the vertical scaling of the grid spacing defined as 
```math
\frac{\partial z}{\partial r} = \frac{H + \eta}{H}
```
with $z$ the actual spatial coordinate (moving with the free surface) and $r$ is the reference vertical coordinate also called $z^\star$ (equivalent to a static coordinate, or $z$ when $\eta = 0$).
`ηⁿ` is the value of the free surface at the current time step, needed to calculate `znodes` 
```math
z = r \cdot e_3 + \eta \ .
```
Finally $H$ is the output of `static_column_depth`.
As a consequence of the definitions, all $z$ properties (`znodes`, `zspacings`, `Δz`, etc...) are replaced with $r$ (`rnodes`, `rspacings`,`Δr`, etc...) and the $z$ properties for a zstar vertical coordinate are calculated appropriately in the Operators module (https://github.com/CliMA/Oceananigans.jl/blob/ss/new-zstar/src/Operators/variable_grid_operators.jl)

An example of the proposed user interface is shown in the `validation/z_star_coordinate` examples

## Changes in the model timestepping

Another requirement is changing the internals of the `HydrostaticFreeSurfaceModel` to solve the correct equations, in particular, the notable changes are:

(1) Updating the grid at each timestep in the `update_state!` step

(2) including the gradient of the grid in the momentum equations, calculated as:
https://github.com/CliMA/Oceananigans.jl/blob/9e05e21921affdf3c8bdeb5a1f58da926550e907/src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl#L85-L100

(3) Changing the computation of the vertical velocity to include the movement of the grid
https://github.com/CliMA/Oceananigans.jl/blob/9e05e21921affdf3c8bdeb5a1f58da926550e907/src/Models/HydrostaticFreeSurfaceModels/compute_w_from_continuity.jl#L22-L35

(4) Advancing $e_3\theta$ instead of $\theta$ in the tracer equations and subsequently unscale it back after the grid scaling factor at the new time step $e_3^{n+1}$ is known
https://github.com/CliMA/Oceananigans.jl/blob/9e05e21921affdf3c8bdeb5a1f58da926550e907/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_ab2_step.jl#L111-L128

(5) Adding a non-linear free surface by changing the `static_column_depth` in the split explicit free surface implementation to a `dynamic_column_depth` defined as
https://github.com/CliMA/Oceananigans.jl/blob/9e05e21921affdf3c8bdeb5a1f58da926550e907/src/ImmersedBoundaries/zstar_immersed_grid.jl#L14-L17

## OutputWriters?

The last piece of the puzzle (not implemented in this PR) would be to change the `OutputWriters` to include by default the variable grid properties in the `timeseries` field of the jld2 / netcdf writer.

## Possibility for improvements

I got a bit carried away and completed a working implementation to make sure it could be done this way, but I am happy to change just about everything in here. I would like to know what people think about this implementation and what people suggest especially in terms of 
(1) Implementation
(2) Variable naming

I can also split this PR in a couple of ones, probably one that includes the refactor to the grids (implementation of a `StaticVerticalCoordinate`) and one that implements a working version of a `ZStarVerticalCoordinate`.


simone-silvestri:
Thanks for the feedback. I was indeed not super sure about the naming so we can decide how to make it more understandable.

> 2. Why use $r$ (`rnodes`, `rspacings`, etc.)? I ask because $r$ suggests radial coordinates on a sphere where the horizontal spacing of each grid cell increases as $r$ increases. But as far as I know from looking at Figure 1 of Adcroft & Campin (2004) [see below] the vertical grid lines are still straight in z* coordinates.

I have used `r` for reference. This coordinate stands for a generalized vertical coordinate equivalent to the `z` in a static geopotential grid (`r` in a zstar grid is equivalent to the z of a grid with a free surface equal to zero). Adcroft and Campin use `r` for a generalized coordinate in the appendix (zstar is a particular type of `r` in this case). But I am down to finding another name to avoid confusion with radius. In NEMO they call it `s` for example, which might be compact and unique enough.

> 3. More a question of notation: I know you've defined `e₃` is the vertical scaling of the grid spacing $\partial z / \partial r$. Is this the unit vector in the third dimension on the new $z^\star$ grid? I guess $e_n$ usually denotes the unit vector in the nth dimension (or maybe $\hat{e}_n$).

I have chosen `e₃` because it is used also in NEMO and it is compact enough. My previous version had `vertical_scaling` which I found out to be a little too verbose for an operator. We can decide on another symbol we want to use for this quantity. Maybe `∂z`? or it might be confused with the derivative in z? If we stick with the `r` notation `∂r_z` would be probably the best choice (or `∂s_z` if we go with `s`)

> Does this mean Oceananigans now supports a non-linear free surface?
 
Yep, with zstar and a splitexplicit free surface, we have a non-linear free surface implementation automatically.

simone-silvestri:
actually, the more I think about it, probably `s` is the correct coordinate to use and `e₃` can become `∂s_z`

validation/stratified_couette_flow/stratified_couette_flow.jl
```diff
@@ -11,7 +11,7 @@ using Oceananigans.Advection: cell_advection_timescale
 
 """ Friction velocity. See equation (16) of Vreugdenhil & Taylor (2018). """
 function uτ(model, Uavg, U_wall, n)
-    Nz, Hz, Δz = model.grid.Nz, model.grid.Hz, model.grid.Δzᵃᵃᶜ
+    Nz, Hz, Δz = model.grid.Nz, model.grid.Hz, model.grid.z.Δzᶜ
```
glwagner:
```suggestion
    Nz, Hz, Δz = model.grid.Nz, model.grid.Hz, model.grid.z.Δzᵃᵃᶜ
```

I'm not sure we want to change the notation. Dropping the "a" has produced confusion in the past.

simone-silvestri:
Luckily this PR slimmed down a lot thanks to #3974. A lot of lines that this PR adds are validation examples, we can probably remove a couple and leave only one.

simone-silvestri:
I have changed `e₃` to `σ` and, therefore, `∂t_e₃` to `∂t_σ`. If the tests pass, this PR should be ready to merge

simone-silvestri:
The TODOs after this PR will be:

- Implement non-linear free surfaces for the other free surface solvers
- Adapt OutputWriters to spit out the grid evolution
- Implement the correct grid update for the split RK3 scheme

src/Advection/vector_invariant_cross_upwinding.jl
```diff
@@ -17,20 +17,37 @@
...
+# which represents the static volume times the time derivative of the vertical grid scaling.
+@inline V_times_∂t_σ(i, j, k, grid) = Azᶜᶜᶜ(i, j, k, grid) * Δrᶜᶜᶜ(i, j, k, grid) * ∂t_σ(i, j, k, grid)
+
 @inline function upwinded_divergence_flux_Uᶠᶜᶜ(i, j, k, grid, scheme::VectorInvariantCrossVerticalUpwinding, u, v)
     @inbounds û = u[i, j, k]
     δ_stencil = scheme.upwinding.divergence_stencil
 
-    δᴿ = _biased_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, scheme.divergence_scheme, bias(û), flux_div_xyᶜᶜᶜ, δ_stencil, u, v) 
+    δᴿ  =    _biased_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, scheme.divergence_scheme, bias(û), flux_div_xyᶜᶜᶜ, δ_stencil, u, v) 
+    ∂ts = _symmetric_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, cross_scheme, V_times_∂t_σ)
```
glwagner, line 40:
```suggestion
    ∂t_σ = _symmetric_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, cross_scheme, V_times_∂t_σ)
```

Is this 0 by default?

src/Advection/vector_invariant_self_upwinding.jl
```diff
@@ -2,23 +2,26 @@
 ##### Self Upwinding of Divergence Flux, the best option!
 #####
 
-@inline δx_U(i, j, k, grid, u, v) =  δxᶜᵃᵃ(i, j, k, grid, Ax_qᶠᶜᶜ, u)
-@inline δy_V(i, j, k, grid, u, v) =  δyᵃᶜᵃ(i, j, k, grid, Ay_qᶜᶠᶜ, v)
+@inline δx_U(i, j, k, grid, u, v) = δxᶜᶜᶜ(i, j, k, grid, Ax_qᶠᶜᶜ, u)
+@inline δy_V(i, j, k, grid, u, v) = δyᶜᶜᶜ(i, j, k, grid, Ay_qᶜᶠᶜ, v)
+
+@inline δx_U_plus_metric(i, j, k, grid, u, v) = δxᶜᶜᶜ(i, j, k, grid, Ax_qᶠᶜᶜ, u) + V_times_∂t_σ(i, j, k, grid)
+@inline δy_V_plus_metric(i, j, k, grid, u, v) = δyᶜᶜᶜ(i, j, k, grid, Ay_qᶜᶠᶜ, v) + V_times_∂t_σ(i, j, k, grid)
```
glwagner, line 9: what does "times" mean? Like multiplicatin?

I think just `V_∂t_σ` is good enough. It's better to choose _either_ math notation or english, but not mix the two.

glwagner: Also how does `V` get in here if this function just depends on `i, j, k, grid`

src/Grids/grid_generation.jl
```diff
@@ -150,3 +150,48 @@ function generate_coordinate(FT, ::Flat, N, H, ::Nothing, coordinate_name, arch)
         return FT(1), nothing, nothing, FT(1), FT(1)
     end
 end    
+
+####
+#### ZStarVerticalCoordinate
+####
```
glwagner, line 156:
```suggestion
#####
##### ZStarVerticalCoordinate
#####
```

src/Grids/grid_generation.jl
```diff
@@ -150,3 +150,48 @@ function generate_coordinate(FT, ::Flat, N, H, ::Nothing, coordinate_name, arch)
...
+    Lr, rᵃᵃᶠ, rᵃᵃᶜ, Δrᵃᵃᶠ, Δrᵃᵃᶜ = generate_coordinate(FT, topo[3](), Nz, Hz, r_faces, :r, arch)
+
+    args = (topo, (Nx, Ny, Nz), (Hx, Hy, Hz))
+
+    σᶜᶜ⁻ = new_data(FT, arch, (Center, Center, Nothing), args...)
+    σᶜᶜⁿ = new_data(FT, arch, (Center, Center, Nothing), args...)
+    σᶠᶜⁿ = new_data(FT, arch, (Face,   Center, Nothing), args...)
+    σᶜᶠⁿ = new_data(FT, arch, (Center, Face,   Nothing), args...)
+    σᶠᶠⁿ = new_data(FT, arch, (Face,   Face,   Nothing), args...)
+    ηⁿ    = new_data(FT, arch, (Center, Center, Nothing), args...)
```
glwagner, line 188:
```suggestion
    ηⁿ   = new_data(FT, arch, (Center, Center, Nothing), args...)
```

glwagner:
I'm finding a lot of little typos, is this ready / has been proofread?

src/Advection/vector_invariant_self_upwinding.jl
simone-silvestri: V is the volume actually, probably volume is a better name

simone-silvestri:
Right, I forgot to change all the comments after the last naming change. I have given it a thorough read and correction, I think it is ready now.

NoraLoose:
@simone-silvestri thanks so much for working on the z-star implementation!

I just took the `ZStarVerticalCoordinate` for a spin and tested it for passive tracer conversation (I'm sure you have done the same!). I released a tracer at the surface in a global 1 degree setup (using `ClimaOcean`), and the tracer conservation is much improved over the z-coordinate implementation - fantastic!! For the z-coordinate, the relative error is order 10^-4 (blue line), but for the new z-star coordinate the relative error is order 10^-6 (orange line), making it appear like zero in the figure below. Nice work! :muscle:

Screenshot 2025-01-09 at 2 51 52 PM

NoraLoose:
Going through your PR, I am trying to figure out whether I can use exponential vertical spacing together with the z-star coordinate. (I have been using the exponential vertical spacing for the z coordinate.) Here, I simply did:
```
grid = LatitudeLongitudeGrid(arch;
                             size = (Nx, Ny, Nz),
                             halo = (7, 7, 7),
                             z = ZStarVerticalCoordinate(-6000meters, 0),
                             latitude  = (-75, 75),
                             longitude = (0, 360))
```

NoraLoose:
One more figure. Over the course of 1 year, there is a slight drift in tracer conservation (we are losing tracer), increasing the relative error from O(10^-16) to O(10^-13). But that's all within machine precision, I think, so pretty good!!

Screenshot 2025-01-09 at 3 11 09 PM

glwagner:
> One more figure. Over the course of 1 year, there is a slight drift in tracer conservation (we are losing tracer), increasing the relative error from O(10^-16) to O(10^-13). But that's all within machine precision, I think, so pretty good!!
> 
> Screenshot 2025-01-09 at 3 11 09 PM

Do we expect a machine precision error to manifest within this metric as something like $\epsilon N_x N_y N_z N_t$?

NoraLoose:
> Do we expect a machine precision error to manifest within this metric as something like 

To me it is not clear why there is a sign-definite drift. If it was only a machine precision error, I would expect it to be fluctuating and not grow in time. Maybe I'm wrong?

I'm not concerned about the O(10^-13) error, but I can run the simulation for longer, and see if it keeps growing.

NoraLoose:
Maybe it plays a role that I'm releasing the tracer into a surface cell, and over time it sinks, so it will hit more cells in the vertical. In other words, the *effective* Nz grows over time.

simone-silvestri:
Thanks @NoraLoose for testing this out with ClimaOcean. I am not incredibly concerned by the 1e-13 error, but in theory the error should be machine precision so it is definitely worth investigating why the error is sign positive.

With respect to the user interface, the `ZStarVerticalCoordinate` does not pose any restriction compared to what is possible now in Oceananigans / ClimaOcean (for the `HydortatocFreeSurfaceModel`). So anything you would pass to a grid under the `z` keyword argument, you can wrap with `ZStarVerticalCoordinate` and it will work.

simone-silvestri:
an example would be 
```julia
r_faces = [1,2,3,4,5,6,7] # (can be used directly as z=r_faces in the grid constructor)
z_faces = ZStarVerticalCoordinate(r_faces)
```

NoraLoose:
@simone-silvestri @glwagner thanks for clarifying the use of the wrapper functions to set up the z-star coordinate.

To finish the discussion point on the drift in tracer conservation, I ran the experiment not just for 1 year, but for 10 years. The conclusions are:
* Blue line: We will keep losing the tracer (in a sign-definite fashion), but not at the same rate. While the relative error increases from 1e-16 to 1e-13 over the course of the first year (this is the result we discussed above), it only further increases to 1e-12 over the course of the next 9 years.
> Maybe it plays a role that I'm releasing the tracer into a surface cell, and over time it sinks, so it will hit more cells in the vertical. In other words, the effective Nz grows over time.
* Orange line: To test the hypothesis that the magnitude of the error depends on the effective number of grid cells that the tracer takes up, I released a tracer in **all** surface cells (orange line),  not just one surface cell as shown by the blue line. The error is larger initially (which I guess confirms the hypothesis), but after ~1.5 years becomes smaller than the error implied by the single-cell release. Not sure what conclusions to draw from that...





Screenshot 2025-01-13 at 9 26 51 AM

TLDR: I think we can feel good about the tracer conservation of the z-star coordinate.

glwagner:
@NoraLoose another test could release the tracer at depth, then we can see if there is a specific issue with the surface or not. 

I sort of agree that this is completely negligible compared to a real signal so for all intents and purposes tracers are conserved. But we should still try to wrap our heads around this eventually I think.

src/Advection/vector_invariant_cross_upwinding.jl
```diff
@@ -17,20 +17,36 @@
...
+# The discrete continuity equation is calculated as:
+#
+#  wᵏ⁺¹ - wᵏ      δx(Ax u) + δy(Ay v)     Δrᶜᶜᶜ ∂t_σ
+# ---------- = - --------------------- - -------------
+#    Δzᶜᶜᶜ                Vᶜᶜᶜ              Δzᶜᶜᶜ
+#
+# We upwind the discrete divergence `δx(Ax u) + δy(Ay v)` and then divide by the volume, 
+# therefore, the correct term to be added to the divergence transport due to the moving grid is:
+#
+#  Azᶜᶜᶜ Δrᶜᶜᶜ ∂t_σ
```
glwagner, line 29: what happens if the grid is stationary

src/Advection/vector_invariant_cross_upwinding.jl
```diff
@@ -17,20 +17,36 @@
...
     @inbounds û = u[i, j, k]
     δ_stencil = scheme.upwinding.divergence_stencil
 
-    δᴿ = _biased_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, scheme.divergence_scheme, bias(û), flux_div_xyᶜᶜᶜ, δ_stencil, u, v) 
+    δᴿ   =    _biased_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, scheme.divergence_scheme, bias(û), flux_div_xyᶜᶜᶜ, δ_stencil, u, v) 
+    ∂t_σ = _symmetric_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, cross_scheme, Az_Δr_∂t_σ)
 
-    return û * δᴿ
+    return û * (δᴿ + ∂t_σ)
 end
```
glwagner, line 42: confused about this. It makes it look like we always have a moving grid

src/Advection/vector_invariant_self_upwinding.jl
```diff
@@ -2,23 +2,26 @@
...
+@inline divergence_smoothness(i, j, k, grid, u, v) = δx_U(i, j, k, grid, u, v) + δy_V(i, j, k, grid, u, v) + Az_Δr_∂t_σ(i, j, k, grid)
 
 @inline function upwinded_divergence_flux_Uᶠᶜᶜ(i, j, k, grid, scheme::VectorInvariantSelfVerticalUpwinding, u, v)
 
     δU_stencil   = scheme.upwinding.δU_stencil    
     cross_scheme = scheme.upwinding.cross_scheme
 
     @inbounds û = u[i, j, k]
-    δvˢ = _symmetric_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, cross_scheme, δy_V, u, v) 
+    δvˢ = _symmetric_interpolate_xᶠᵃᵃ(i, j, k, grid, scheme, cross_scheme, δy_V_plus_metric, u, v)
```
glwagner, line 24: what metric?

src/Grids/grid_generation.jl
```diff
@@ -150,3 +150,48 @@ function generate_coordinate(FT, ::Flat, N, H, ::Nothing, coordinate_name, arch)
...
 end    
+
+#####
+##### ZStarVerticalCoordinate
+#####
+
+generate_coordinate(FT, ::Periodic, N, H, ::ZStarVerticalCoordinate, coordinate_name, arch, args...) = 
+    throw(ArgumentError("Periodic domains are not supported for ZStarVerticalCoordinate"))
+
+# Generate a moving coordinate with evolving scaling (`σ`) for spacings and znodes
```
glwagner, line 161: I don't think this function does what the comment says. A moving coordinate is presumably implemented by a model. The comment should be updated to state exactly what this function does. PResumably, it makes it _possible_ to have a moving coordinate, somehow.

src/Grids/grid_generation.jl
```diff
@@ -150,3 +150,48 @@ function generate_coordinate(FT, ::Flat, N, H, ::Nothing, coordinate_name, arch)
...
+    σᶠᶠⁿ = new_data(FT, arch, (Face,   Face,   Nothing), args...)
+    ηⁿ   = new_data(FT, arch, (Center, Center, Nothing), args...)
+    ∂t_σ = new_data(FT, arch, (Center, Center, Nothing), args...)
+
+    # Fill all the scalings with one (at rest coordinate)
+    for σ in (σᶜᶜ⁻, σᶜᶜⁿ, σᶠᶜⁿ, σᶜᶠⁿ, σᶠᶠⁿ)
+        fill!(σ, 1)
+    end
+    
+    return Lr, ZStarVerticalCoordinate(rᵃᵃᶠ, rᵃᵃᶜ, Δrᵃᵃᶠ, Δrᵃᵃᶜ, ηⁿ, σᶜᶜⁿ, σᶠᶜⁿ, σᶜᶠⁿ, σᶠᶠⁿ, σᶜᶜ⁻, ∂t_σ)
```
glwagner, line 196: Why does this return two things

src/Grids/vertical_coordinates.jl
```diff
@@ -29,12 +29,41 @@ struct StaticVerticalCoordinate{C, D, E, F} <: AbstractVerticalCoordinate
...
+    Δᵃᵃᶠ :: E
+    Δᵃᵃᶜ :: F
+      ηⁿ :: H
+    σᶜᶜⁿ :: CC
+    σᶠᶜⁿ :: FC
+    σᶜᶠⁿ :: CF
+    σᶠᶠⁿ :: FF
+    σᶜᶜ⁻ :: CC
+    ∂t_σ :: CC
+end
```
glwagner, line 44: Isn't this more general than z*? It's just a vertical coordinate that _can_ evolve in time

src/Grids/vertical_coordinates.jl
```diff
@@ -99,11 +154,27 @@ z_domain(grid) = domain(topology(grid, 3)(), grid.Nz, grid.z.cᵃᵃᶠ)
...
+    return ZStarVerticalCoordinate(cᶠ, cᶜ, args[3:end]...)
+end
+
 # Summaries
-coordinate_summary(::Bounded, z::AbstractVerticalCoordinate, name) = 
-    @sprintf("Free-surface following with Δ%s=%s", name, prettysummary(z.Δᵃᵃᶜ))
+coordinate_summary(topo, z::StaticVerticalCoordinate, name) = coordinate_summary(topo, z.Δᵃᵃᶜ, name)
+
+coordinate_summary(::Bounded, z::RegularZStarVerticalCoordinate, name) = 
+    @sprintf("Free-surface following with Δr=%s", prettysummary(z.Δᵃᵃᶜ))
```
glwagner, line 175: I don't think this has to follow a free surface. That is just a particular choice that is made in another module (HydrostaticFreeSurfaceModels)

src/Grids/vertical_coordinates.jl
```diff
@@ -99,11 +154,27 @@ z_domain(grid) = domain(topology(grid, 3)(), grid.Nz, grid.z.cᵃᵃᶠ)
...
+
+coordinate_summary(::Bounded, z::ZStarVerticalCoordinate, name) = 
+    @sprintf("Free-surface following with min(Δr)=%s, max(Δr)=%s",
```
glwagner, line 178: same comment as above

src/ImmersedBoundaries/zstar_immersed_grid.jl
```diff
@@ -0,0 +1,58 @@
...
+                           dynamic_column_depthᶜᶠᵃ,
+                           dynamic_column_depthᶠᶜᵃ,
+                           dynamic_column_depthᶠᶠᵃ
+
+import Oceananigans.Operators: σⁿ, σ⁻, ∂t_σ
+
+const ZStarImmersedGrid   = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:AbstractZStarGrid}
+const ZStarGridOfSomeKind = Union{ZStarImmersedGrid, AbstractZStarGrid}
+
+@inline dynamic_column_depthᶜᶜᵃ(i, j, k, grid::ZStarGridOfSomeKind, η) = static_column_depthᶜᶜᵃ(i, j, grid) +  @inbounds η[i, j, k]
```
glwagner, line 15:
```suggestion
@inline column_depthᶜᶜᵃ(i, j, k, grid::ZStarGridOfSomeKind, η) = static_column_depthᶜᶜᵃ(i, j, grid) +  @inbounds η[i, j, k]
```

This is just the "column depth" isn't it? The concept of dynamic vs static doesn't make sense for a fixed grid. Better to refer to this as something that generalizes to all grids. Then we have "static column depth" for the specific case that it's needed.

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/barotropic_split_explicit_corrector.jl
```diff
@@ -1,39 +1,51 @@
...
+    k_top  = size(grid, 3) + 1
+
+    hᶠᶜ = static_column_depthᶠᶜᵃ(i, j, grid)
+    hᶜᶠ = static_column_depthᶜᶠᵃ(i, j, grid)
+
+    Hᶠᶜ = dynamic_column_depthᶠᶜᵃ(i, j, k_top, grid, η)
+    Hᶜᶠ = dynamic_column_depthᶜᶠᵃ(i, j, k_top, grid, η)
+
+    σᶠᶜ = ifelse(hᶠᶜ == 0, one(grid), Hᶠᶜ / hᶠᶜ)
+    σᶜᶠ = ifelse(hᶜᶠ == 0, one(grid), Hᶜᶠ / hᶜᶠ)
```
glwagner, line 23: why is this 1 for the case that hits the conditional?

src/Models/HydrostaticFreeSurfaceModels/compute_w_from_continuity.jl
```diff
@@ -18,12 +19,38 @@ compute_w_from_continuity!(model; kwargs...) =
 compute_w_from_continuity!(velocities, arch, grid; parameters = w_kernel_parameters(grid)) = 
     launch!(arch, grid, parameters, _compute_w_from_continuity!, velocities, grid)
 
+######################################################
+# The derivative of the moving grid is:
```
glwagner, line 23: This comment says something about a moving grid, but I think the code here is general and also works for static grids. This comment is very confusing, it will make people think that we always have a moving grid

src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
```diff
@@ -0,0 +1,160 @@
...
+
+    # Scalings and free surface
+    σᶜᶜ⁻  = grid.z.σᶜᶜ⁻
+    σᶜᶜⁿ  = grid.z.σᶜᶜⁿ
+    σᶠᶜⁿ  = grid.z.σᶠᶜⁿ
+    σᶜᶠⁿ  = grid.z.σᶜᶠⁿ
+    σᶠᶠⁿ  = grid.z.σᶠᶠⁿ
+    ∂t_σ  = grid.z.∂t_σ
+    ηⁿ    = grid.z.ηⁿ
+    η     = model.free_surface.η
```
glwagner, line 27: This code only works for a HydrostaticFreeSurfaceModel?

src/Operators/variable_grid_operators.jl
glwagner, line 1: should this be `time_variable_grid_operators.jl`?

glwagner: just "variable" grid is confusing because one might interpret this as referring to a stretched grid.

test/test_active_cells_map.jl
```diff
@@ -0,0 +1,124 @@
...
+
+@testset "Active cells map" begin
+    for arch in archs
+        underlying_grid = LatitudeLongitudeGrid(arch, size = (Nx, Ny, Nz),
+                                                halo = (4, 4, 4),
+                                                latitude = (-80, 80),
+                                                longitude = (-160, 160),
+                                                z = (-10, 0),
+                                                radius = 1,
+                                                topology=(Bounded, Bounded, Bounded))
```
glwagner, line 64:
```suggestion
                                                topology = (Bounded, Bounded, Bounded))
```

test/test_split_explicit_vertical_integrals.jl
```diff
@@ -70,38 +72,38 @@ using Oceananigans.Models.HydrostaticFreeSurfaceModels.SplitExplicitFreeSurfaces
...
         @testset "Vertical Integral " begin
             Δz = zeros(Nz)
             Δz .= grid.z.Δᵃᵃᶜ
 
-            u .= 0.0
-            U .= 1.0
-            compute_barotropic_mode!(U, V, grid, u, v)
-            @test all(Array(U.data.parent) .== 0.0)
+            set!(u, 0.0)
+            set!(U, 1.0)
```
glwagner, line 84:
```suggestion
            set!(u, 0)
            set!(U, 1)
```

test/test_split_explicit_vertical_integrals.jl
```diff
@@ -70,38 +72,38 @@ using Oceananigans.Models.HydrostaticFreeSurfaceModels.SplitExplicitFreeSurfaces
...
+            compute_barotropic_mode!(U, V, grid, u, v, η̅)
+            @test all(Array(interior(U)) .== 0.0)
 
-            u .= 1.0
-            U .= 1.0
-            compute_barotropic_mode!(U, V, grid, u, v)
+            set!(u, 1.0)
+            set!(U, 1.0)
```
glwagner, line 89:
```suggestion
            set!(u, 1)
            set!(U, 1)
```

validation/z_star_coordinate/lock_release.jl
glwagner, line 1: what are the results of this test?

glwagner (commented):
The names and comments need to be changed to reflect the notions that:

* grids can be static. Where time-depednent metrics enter, we need a comment / explanation for what happens when the grid is not changing in time
* The data structures for the grid are likely general for any time-dependent vertical coordinate
* The "z*" approximation is a physical / modeling approximation that is independent of the data structures. Specifically, this is a rule for updating the time-dependent grid, plus an approximation that neglects some slope terms, assuming the free surface displacement is small compared to the depth of the domain. However, the "z*" prescription and data structures are properly viewed as independent. This is important for understanding how the code works, plus also for grasping how one would implement other vertical coordinates.

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/barotropic_split_explicit_corrector.jl
simone-silvestri: I was thinking that a zero `dz` might create some NaNs were we don't want them. This assumes that if the height is zero (which basically means that we do not have active cells in the column) we revert to a z-coordinate representation

validation/z_star_coordinate/lock_release.jl
simone-silvestri: It calculates the drift in a lock-exchange simulation with an sloping immersed boundary.
The drift stays within 1e-16
```julia
julia> @show drift
drift = Any[0.0, -1.734723475976807e-16, -2.1510571102112408e-16, -3.469446951953614e-17, -8.326672684688674e-17, -1.0408340855860843e-16, -6.245004513516506e-17, -1.942890293094024e-16, -2.5673907444456745e-16, -1.734723475976807e-16, -9.020562075079397e-17, 4.163336342344337e-17, -2.498001805406602e-16, -1.1796119636642288e-16, -2.498001805406602e-16, -2.7755575615628914e-17, -1.5959455978986625e-16, -1.457167719820518e-16, -4.163336342344337e-17, -2.3592239273284576e-16, -2.0816681711721685e-16, -8.326672684688674e-17, -1.5959455978986625e-16, -1.942890293094024e-16, -1.734723475976807e-16, -1.8735013540549517e-16, 6.938893903907228e-18, -1.3877787807814457e-17, -3.608224830031759e-16, -1.8735013540549517e-16, -1.3183898417423734e-16, -2.498001805406602e-16, -1.3183898417423734e-16, -2.914335439641036e-16, -1.457167719820518e-16, -3.0531133177191805e-16, -9.020562075079397e-17, -3.0531133177191805e-16, -1.1102230246251565e-16, -1.5959455978986625e-16, -6.245004513516506e-17, -7.632783294297951e-17, -5.551115123125783e-17, -2.0122792321330962e-16, -1.942890293094024e-16, -7.632783294297951e-17, -1.942890293094024e-16, -6.245004513516506e-17, -1.1102230246251565e-16, -1.249000902703301e-16, -1.942890293094024e-16, -1.8735013540549517e-16, -1.5265566588595902e-16, -2.5673907444456745e-16, -1.0408340855860843e-16, -1.0408340855860843e-16, -1.5265566588595902e-16, -1.3183898417423734e-16, -9.71445146547012e-17, -2.2898349882893854e-16, -5.551115123125783e-17, -3.469446951953614e-17]
```

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/barotropic_split_explicit_corrector.jl
glwagner: I think it's good to define that behavior. Maybe we will learn more about the implications when we have tests. Should we make a comment about it?

simone-silvestri: Sure I am going to resee al the comments

src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
glwagner: It could make sense since it looks like it is supposed to be extended for other cases

src/Grids/vertical_coordinates.jl
```diff
@@ -29,12 +29,41 @@ struct StaticVerticalCoordinate{C, D, E, F} <: AbstractVerticalCoordinate
...
+    σᶜᶜ⁻ :: CC
+    ∂t_σ :: CC
+end
+
+"""
+    ZStarVerticalCoordinate(r_faces)
+
+Construct a `ZStarVerticalCoordinate` from `r_faces` that can be a `Tuple`, a function of an index `k`,
+or an `AbstractArray`. A `ZStarVerticalCoordinate` is a vertical coordinate that evolves in time 
+following the surface.
```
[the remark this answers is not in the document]
simone-silvestri: given that the scaling in 2D though.

simone-silvestri:
I have changed the PR to reflect the comments, in particular, now we have a 
`MutableVerticalCoordinate` (as opposed to the `StaticVerticalCoordinate`) that contains the data structures required for a possible time-evolution of the grid. This is controlled specifically by a new field of the HydrostaticFreeSurface called `vertical_coordinate`, that (at this moment) can only be either `ZStar()` or `ZCoordinate()` (default). Then the UI to enable a z coordinate model will be
```julia
r_faces = (-1, 0)
z_faces = MutableVerticalCoordinate(r_faces)
grid = RectilinearGrid(size = (10, 10, 10), x = (0, 1), y = (0, 1), z = z_faces)
model = HydrostaticFreeSurfaceModel(; grid, vertical_coordinate = ZStar())
```
without the last `ZStar()` argument in the model, despite the grid being an `AbstractMutableGrid`, the grid will not evolve and behave as a static grid

glwagner (approved):

glwagner:
Do we test that two identical models with `Static` or `Mutable` vertical coordinate produce the same result?

glwagner:
Tiny user interface change could be to choose the default `vertical_coordinate` based on the grid; eg if the grid has a `MutableVerticalCoordinate`, then we can choose ZStar. The downside of this is a loss of readability though so we may want to keep the current situation.

glwagner:
One small issue is that both the grid and model use the name "vertical coordinate", but to mean different things (data structures versus physics). I don't know if something like `MutableVerticalDirection` would help for the grid, or not

src/Oceananigans.jl
```diff
@@ -11,6 +11,7 @@ export
     # Grids
     Center, Face,
     Periodic, Bounded, Flat,
+    MutableVerticalCoordinate,
```
glwagner, line 14: may be better not to export this _immediately_, instead we can have a phase where it is experimental?

simone-silvestri:
> Tiny user interface change could be to choose the default `vertical_coordinate` based on the grid; eg if the grid has a `MutableVerticalCoordinate`, then we can choose ZStar. The downside of this is a loss of readability though so we may want to keep the current situation.

I was thinking about this in `ClimaOcean`, where probably this feature is mostly going to be used, since we already have a bunch of defaults, we can add a `default_vertical_coordinate(grid::MutableGridOfSomeKind) = ZStar()`

simone-silvestri:
> One small issue is that both the grid and model use the name "vertical coordinate", but to mean different things (data structures versus physics). I don't know if something like `MutableVerticalDirection` would help for the grid, or not

Sounds good, a `MutableVerticalDiscretization` is a good name, I will change also `StaticVerticalCoordinate` for uniformity to `StaticVerticalDiscretization`

PR #3959, opened 2024-11-25
Fix partial cell bottom areas and volumes

### LINKED ISSUE #3958: Issue with `PartialCellBottom`
I think I found a bug in the `PartialCellBottom` code. 
The definition of volumes for ImmersedBoundaryGrids
https://github.com/CliMA/Oceananigans.jl/blob/d289fed2d06e9530e23b60d8c421243cafb6a996/src/ImmersedBoundaries/immersed_grid_metrics.jl#L26
bypasses the calculation of the volumes done in 
https://github.com/CliMA/Oceananigans.jl/blob/d289fed2d06e9530e23b60d8c421243cafb6a996/src/Operators/spacings_and_areas_and_volumes.jl#L263

Therefore, not accounting for reduced spacing in volumes.

I will open a PR to fix it.

simone-silvestri:
closes #3958

### DIFF
--- a/docs/oceananigans.bib
+++ b/docs/oceananigans.bib
@@ -979,3 +979,13 @@
     year = {2005},
     doi = {10.1063/1.1839152},
 }
+
+@article{Lan2022,
+	author = {Lan, Rihui and Ju, Lili and Wanh, Zhu and Gunzburger, Max and Jones, Philip},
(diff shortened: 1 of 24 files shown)


simone-silvestri:
Actually, also this PR can be lumped in into #3955 since that one is already changing the immersed spacings

PR #3960, opened 2024-11-26
Allow Tuple of closures with CATKE

simone-silvestri:
in this way we don't timestep the energy equation twice within a time step when using a tuple of closures containing CATKE

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_ab2_step.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_ab2_step.jl
@@ -63,21 +63,26 @@ end
 
 const EmptyNamedTuple = NamedTuple{(),Tuple{}}
 
+hasclosure(closure, ClosureType) = closure isa ClosureType
+hasclosure(closure_tuple::Tuple, ClosureType) = any(hasclosure(c, ClosureType) for c in closure_tuple)
+
 ab2_step_tracers!(::EmptyNamedTuple, model, Δt, χ) = nothing
 
 function ab2_step_tracers!(tracers, model, Δt, χ)
 
     closure = model.closure
 
+    catke_closures = hasclosure(closure, FlavorOfCATKE)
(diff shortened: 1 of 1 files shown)


simone-silvestri:
I think it is a bug because the tracer is advanced in the `TurbulenceClosures` module here 
https://github.com/CliMA/Oceananigans.jl/blob/fe4123f92155f5086bce72e366843045606b4bf5/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl#L12

So with closure tuple it is advanced twice

simone-silvestri:
Should I add a test for it?

ali-ramadhan:
Always good to add a test! Especially since this issue might come up again in the future with refactors.

I'm just trying to think of a good test... Are there analytic solutions to the TKE where we can take one time step and make sure the TKE values are correct (and that only one, not two, time steps were taken)? 

As a side note: I've been running with CATKE in a tuple of closures and I encountered NaNs at the bottom-most grid cell but the simulations seems to continue time stepping just fine. Maybe it's related to this bug fix?

simone-silvestri:
We can also brutally just check that we do not step the kinetic energy twice: 
i.e that 
```julia
eⁿ⁺¹ == eⁿ + Δt * Gᵉ # we have to account for the fast parts that depend on uⁿ⁺¹ here
```
and not
```
eⁿ⁺¹ == eⁿ + 2 * Δt * Gᵉ 
```

glwagner:
> We can also brutally just check that we do not step the kinetic energy twice: i.e that
> 
> ```julia
> eⁿ⁺¹ == eⁿ + Δt * Gᵉ # we have to account for the fast parts that depend on uⁿ⁺¹ here
> ```
> 
> and not
> 
> ```
> eⁿ⁺¹ == eⁿ + 2 * Δt * Gᵉ 
> ```

We do this elsewhere, eg for `test_forcings.jl`, I think adding a forcing test with CATKE + `e` would be quick and easy.

PR #3961, opened 2024-11-26
No warning when passing `nothing` to the `immersed` kwarg of `FieldBoundaryConditions`

simone-silvestri:


glwagner (approved):
This seems a little dirty, why can't people use `DefaultBoundaryCondition`? In other words I don't see the point of having two names for this.

glwagner:
> I thought that `nothing` is not really a boundary condition, so it kind of seems awkward a warning that says that `nothing` immersed boundary condition is not supported on a non-immersed grid
> 
> A user might think that they need to provide a specific immersed boundary condition, but I am ok closing this without merging.

I guess the default is `nothing`, right?

The question is, why do users set the bc to `nothing`? What do we want users to do?

glwagner:
> The default now seems to be `NoFluxBoundaryCondition`. I think the default should be `nothing`. I can change the default here if people agree.

The default for `immersed`? That makes sense. The annoying thing is that technically, `nothing` is `NoFlux`.

bgroenks96:
@simone-silvestri I was able to make this go away by using `DefaultBoundaryCondition` but it would still be nice to see this get merged.

glwagner:
I think https://github.com/CliMA/Oceananigans.jl/pull/4722 might be a better / simpler solution.

Also I am wary about conflating `nothing` with `NoFlux`. These have a decidedly different meaning on non-immeresed boundaries --- `nothing` boundary condition would definitely incur a flux if used with a diffusive closure.

simone-silvestri:
Sounds good, I think the default should be `NoFlux` for immersed grids and `nothing` for non-immersed grids. I ll close this PR then so we can do this in #4722

PR #3972, opened 2024-12-03
Add helper function for building `buoyancy` `KernelFunctionOperation`

glwagner:
There is still a problem with the API, because we have a `Buoyancy` type, but then also a `buoyancy` constructor which returns a `KernelFunctionOperation`.

I think we should change `Buoyancy` to something else and reserve the colloquial "buoyancy" for the field. This still leaves an issue with the model property `model.buoyancy` but I feel that is a more minor issue. (We could also change the model property to something like `buoyancy_formulation` to distinguish between the concept of the "model" or "formulation" that is used to compute buoyancy, and the "buoyancy field" `b`, but I don't want to go so far yet.)


glwagner:
superceded by #3973

PR #3973, opened 2024-12-03
(0.95.0) `Buoyancy` -> `BuoyancyForce` and `BuoyancyModels` -> `BuoyancyFormulations`

glwagner:
I also changed the API for `BuoyancyForce` so that the "formulation" is a positional argument. I think this is nicer esp because there is no sensible default and it is central to `BuoyancyForce` and required.

I also changed the property of `BuoyancyForce.model` to `BuoyancyForce.formulation`. This avoids the name conflict with `NonhydrostaticModel` etc, which leads to very silly and confusing constructions like `model.buoyancy.model.gravity_unit_vector`. Now this will be `model.buoyancy.formulation.gravity_unit_vector`.

@tomchor you may want to take a look.


src/Oceananigans.jl
```diff
@@ -44,8 +44,8 @@ export
     # Coriolis forces
     FPlane, ConstantCartesianCoriolis, BetaPlane, NonTraditionalBetaPlane,
 
-    # BuoyancyModels and equations of state
-    Buoyancy, BuoyancyTracer, SeawaterBuoyancy,
+    # BuoyancyForces and equations of state
```
tomchor, line 47:
```suggestion
    # BuoyancyFormulations and equations of state
```
?

tomchor:
Good change!

src/Oceananigans.jl
glwagner: either could work here, though I think what you suggest is better!

simone-silvestri (approved):

PR #3974, opened 2024-12-04
(0.95.0) Refactor grids to allow generalized vertical coordinate

simone-silvestri:
Since the Zstar PR (#3956) is very long and requires quite some refactor, I am opening this PR just to perform the refactor that will allow us to add the Zstar seamlessly. 

This PR reworks the grids to have just one `z` field that (for the moment) is a `StaticVerticalCoordinate` containing the vertical grid information (one-dimensional nodes and the spacings).

To distinguish between the physical `z` coordinate and the reference generalized vertical coordinate, this PR introduces new spacings and nodes that refer to the generalized vertical coordinate. At the moment the vertical coordinate is `r` (we already had the notion of `rnodes` and `rnode` in the src) following the nomenclature of Adcroft & Campin 2004, but I am happy to hear everyone's thoughts on this. 

I think @ali-ramadhan was not convinced about `r` because people could think it might refer to a deep atmosphere with horizontal areas increasing with `r`.

Let me know what you think @ali-ramadhan @glwagner @tomchor @navidcy @jm-c 

I would like to merge #3955 before this because it also refactors a bit the spacings


glwagner:
@ali-ramadhan I think `r` is a fine choice, considering that it does indicate a radial coordinate and also is used by MOM6. But `s` could be another choice.

src/Operators/spacings_and_areas_and_volumes.jl
```diff
@@ -35,6 +35,17 @@ The operators in this file fall into three categories:
 #####
 #####
 
+const ZRG = Union{LLGZ, RGZ, OSSGZ}
+
+@inline getspacing(k, Δz::AbstractVector) = @inbounds Δz[k]
+@inline getspacing(k, Δz::Number)         = @inbounds Δz
+
+@inline Δrᵃᵃᶜ(i, j, k, grid::AbstractGrid) = getspacing(k, grid.z.Δᵃᵃᶜ)
+@inline Δrᵃᵃᶠ(i, j, k, grid::AbstractGrid) = getspacing(k, grid.z.Δᵃᵃᶠ)
```
glwagner, line 44: We don't strictly need the `r` stuff in this PR right? It's ok to keep it, just want to understand.

src/ImmersedBoundaries/partial_cell_bottom.jl
```diff
@@ -150,55 +153,55 @@ end
     return !immersed_cell(i, j, k, grid, ib) & immersed_cell(i, j, k-1, grid, ib)
 end
 
-@inline function Δzᶜᶜᶜ(i, j, k, ibg::PCBIBG)
+@inline function Δrᶜᶜᶜ(i, j, k, ibg::PCBIBG)
```
glwagner, line 156: did you mean to change these?

src/Operators/spacings_and_areas_and_volumes.jl
simone-silvestri: right, we don't need it here because at the moment `z = r`

src/ImmersedBoundaries/partial_cell_bottom.jl
simone-silvestri: yeah, I think we need to change these, then the `Δzᶜᶜᶜ` is calculated as `Δrᶜᶜᶜ * dzdr`

glwagner: interesting I see

glwagner (approved):

tomchor:
I think `r` is a more common nomenclature than `s` for the vertical coordinate, so I vote for that one.

tomchor:
@simone-silvestri could you please clarify what's breaking about this change from a user perspective.

simone-silvestri:
A user cannot do `grid.zᵃᵃᶠ` or `grid.Δzᵃᵃᶠ` anymore. Not that they should do it, but still I suspect many scripts access grid spacings or nodes directly from the grid

PR #3937, opened 2024-11-18
(0.95.3) Restores open boundary condition functionality

jagoosw:
The refactoring in https://github.com/CliMA/Oceananigans.jl/pull/3834 broke flat extrapolation open boundary conditions. This restores them.

Closes https://github.com/CliMA/Oceananigans.jl/issues/3975

PR #3977, opened 2024-12-06
Adds a pertubation advection open boundary matching scheme

jagoosw:
This PR adds a `PertubationAdvection` open boundary condition which assumes that there is some mean flow (i.e. prescribed in a channel or from a coarser resolution model) which is homogeneous on the boundary but can vary in time.

On the boundary, we approximate (for an x-boundary) to:
```math
\frac{\partial u}{\partial t} \approx -U\frac{\partial u\prime}{\partial x},
```
where $u=u\prime+U$
I have chosen to take an explicit/backwards Euler step:
```math
\frac{u\prime(t^{n+1}) - u\prime(t^n)}{\Delta t} + \frac{U\prime(t^{n+1}) - U\prime(t^n)}{\Delta t} \approx -U\frac{u\prime(x_b, t^{n+1}) - u\prime(x_{b-1}, t^{n+1})}{\Delta x},
```
because then we don't have to store any other information (i.e. if we did a forward step we would need $u\prime(x_{b-1}, t^n)$ ). This results in:

```math
u(t^{n+1}) = \frac{u^n+\tilde{U}u(x_{b-1}, t^{n+1})}{1+\tilde{U}},
```
where $\tilde{U}=\min\left(1, U\Delta t / \Delta x\right)$. I have also added restoring to $U$ (i.e. damping $u\prime\to0$) in some timescale $\tau$ which can be different for inflow and outflow to allow the scheme to work when there is inflow, which results in:
```math
u(t^{n+1}) = \frac{u^n+\tilde{U}u(x_{b-1}, t^{n+1}) + U\tilde{\tau}}{1+\tilde{U}+\tilde{\tau}},
```
where $\tilde{\tau}=\Delta t / \tau$.

This has the limitation that if we want to extend to have wall tangent mean velocity ($\partial_t u \approx -(\vec{U}\cdot\nabla) u$) then we either have to solve the whole boundary at once since every point is going to depend on its boundary neighbours, or we need to take a forward Euler step in which case we need the first interior point at the previous time step.

I think we probably will need to address this for real nesting cases at some point.

https://github.com/user-attachments/assets/e37fb984-fba2-47b0-a528-1b22d4c9e93d

For now, it seems to be working okay (left plot vorticity, right plot x-velocity):
<details>
<summary>(old plot)</summary>
https://github.com/user-attachments/assets/21ae3eb2-5d3b-4c08-86fb-bb9c222e8e6a
</details>
and the drag coefficient is ~1.4 at Re = 100 and I'm running a case at Re=1000 now.

I have also checked the [Strouhal number](https://en.wikipedia.org/wiki/Strouhal_number) which matches exactly to the expected values (~0.17).


jagoosw:
@tomchor, do you think something like this:
https://github.com/CliMA/Oceananigans.jl/blob/32cd354fdf6a1e8bcc27e774f7ef323f18c23703/validation/open_boundaries/cylinder.jl#L18-L76
could belong in Oceanostics?

jagoosw:
I have also run the same case with different domain lengths (6m, 12m - shown, 18m and 30m) and they all produce very similar Cd and St

tomchor:
> @tomchor, do you think something like this:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/32cd354fdf6a1e8bcc27e774f7ef323f18c23703/validation/open_boundaries/cylinder.jl#L18-L76
> 
> could belong in Oceanostics?

Sure! Why not? :)

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -0,0 +1,90 @@
...
+               bc.classification.matching_scheme.outflow_timescale, 
+               bc.classification.matching_scheme.inflow_timescale)
+
+
+    τ̃ = Δt / τ
+
+    uᵢⁿ⁺¹ = (uᵢⁿ + U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ + U)
+
+    @inbounds u[i, j, k] = uᵢⁿ⁺¹#ifelse(active_cell(i, j, k, grid), uᵢⁿ⁺¹, zero(grid))
+end
```
tomchor, line 64: I think we could implement everything using `getindex()` so that we implement it once (or at max twice; one for left and other for right BC) and can use it in three dimensions.

For example this

```julia
    uᵢⁿ     = @inbounds u[i, j, k]
    uᵢ₋₁ⁿ⁺¹ = @inbounds u[i - 1, j, k]
```

would become 

```julia
    uᵢⁿ     = @inbounds getindex(u, boundary_indices...)
    uᵢ₋₁ⁿ⁺¹ = @inbounds getindex(u, one_off_boundary_indices...)
```

This would avoid errors when transcribing to other dimensions (I remember catching a couple for the flat extrapolation matching scheme).

jagoosw: yeah, I was trying to think of a clean way to do this

jagoosw: Attempted this now (but I've still not written out the other directions in case we want to change things first)

validation/open_boundaries/cylinder.jl
```diff
@@ -1,127 +1,214 @@
...
 
-architecture = GPU()
+Returns the drag within the `bounding_box` computed by:
 
-# model parameters
-Re = 200
-U = 1
-D = 1.
-resolution = D / 40
+```math
```
jagoosw, line 23: Actually the pressure contribution shouldn't be included here since ``G^n`` doesn't include it right?

glwagner:
There's a typo in your top equation right? I think it should be $\partial_x$.

glwagner:
> A problem I think this scheme might have is that as the flow changes direction the assumption that the perturbation is small falls apart, which is maybe why it's been a bit unstable so far in an oscillating case.

What do you mean? What is the consequence of this approximation?

validation/open_boundaries/cylinder.jl
glwagner: I can't see the line you're referring to

jagoosw:
> What do you mean? What is the consequence of this approximation?

I was thinking about this more and really to remain valid then in :
```math
\partial_t (u\prime+U)\approx u\prime\partial_xu\prime + U\partial_xu\prime - u\prime/\tau
```
as $U\to0$ then we need $u\prime\partial_xu\prime\ll u\prime/\tau$ to still be able to throw away the first term so it might be okay actually.

I found a typo in the west boundary which might be why the oscillating case wasn't working, rerunning now

glwagner:
> > What do you mean? What is the consequence of this approximation?
> 
> I was thinking about this more and really to remain valid then in :
> 
> 𝜕
> 𝑡
> (
> 𝑢
> ′
> +
> 𝑈
> )
> ≈
> 𝑢
> ′
> 𝜕
> 𝑥
> 𝑢
> ′
> +
> 𝑈
> 𝜕
> 𝑥
> 𝑢
> ′
> +
> (
> 𝑈
> −
> 𝑢
> ′
> )
> /
> 𝜏
> as U → 0 then we need u ′ ∂ x u ′ ≪ u ′ / τ to still be able to throw away the first term, so I think its actually okay as long as you have the relaxation.
> 
> I found a typo in the west boundary which might be why the oscillating case wasn't working, rerunning now

Okay, I agree with you. You can also analyze the update formula / backward Euler step in the limit U -> 0. It doesn't look like it needs to be regularized in any way in that case though, it looks ok.

tomchor:
Is there a reference for this method? If there is, please include it on the methods docstring. If not, it'd be good to include a brief explanation like the one at the top comment in the dosctring. I don't think this radiation method is very well known.

jagoosw:
I started implementing some code to return the wall-normal velocity which can be put in as the U in this scheme, and I'm not exactly sure where to put it. I have it in the boundary conditions module but it would be very helpful to have it after fields and abstract operations. Could it just go in Utils?

jagoosw:
I've pushed the wall normal velocity thing in `src/` and we can discuss where is appropriate for it

validation/open_boundaries/cylinder.jl
```diff
@@ -1,127 +1,214 @@
...
+\\vec{F}_T=\\int_\\Omega\\left(\\frac{\\partial \\vec{u}}{\\partial t}\\right)dV + \\oint_{\\partial\\Omega}\\left(\\vec{u}(\\vec{u}\\cdot\\hat{n}) + P\\hat{n} - \\vec{\\tau}\\cdot\\hat{n}\\right)dS,\\newline
+\\F_u=\\int_\\Omega\\left(\\frac{\\partial u}{\\partial t}\\right)dV + \\oint_{\\partial\\Omega}\\left(u(\\vec{u}\\cdot\\hat{n}) - \\tau_{xx}\\right)dS + \\int_{\\partial\\Omega}P\\hat{x}\\cdot d\\vec{S},\\newline
+F_u=\\int_\\Omega\\left(\\frac{\\partial u}{\\partial t}\\right)dV - \\int_{\\partial\\Omega_1} \\left(u^2 - 2\\nu\\frac{\\partial u}{\\partial x} + P\\right)dS + \\int_{\\partial\\Omega_2}\\left(u^2 - 2\\nu\\frac{\\partial u}{\\partial x}+P\\right)dS -  \\int_{\\partial\\Omega_2} uvdS + \\int_{\\partial\\Omega_4} uvdS,
+```
+where the bounding box is ``\\Omega`` which is formed from the boundaries ``\\partial\\Omega_{1}``, ``\\partial\\Omega_{2}``, ``\\partial\\Omega_{3}``, and ``\\partial\\Omega_{4}`` 
+which have outward directed normals ``-\\hat{x}``, ``\\hat{x}``, ``-\\hat{y}``, and ``\\hat{y}``
```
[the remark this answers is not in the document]
jagoosw: I guess so, docs should render latex in docstrings but since this isn't going to ever be rendered by Documenter its probably makes more sense in unicode

jagoosw:
Ah there is a problem with the mean velocity approach that the step has to be forward not backward euler since the mean velocity is not in the future like when its prescribed (and I checked and it blows up on simple cases if we try todo it like that)

tomchor (commented):
One comment is that I think the name is a bit ambiguous. There are several open boundary methods that have been developed that (in addition to outflow) include some sort of added perturbation for inflows to allow for faster turbulence development. When I read `PerturbationAdvection` I think of those methods. But if I understand the code correctly the "perturbation" here refers only to perturbations in the outflow, no?

Maybe we could name it `PerturbationOutflow` to avoid confusion?

jagoosw:
The results of simple cases seem more correct doing a forward step + `BoundaryAdjacentMean`, so I'm going to try pushing that and testing it against the headland case. You can currently choose forward or backward but we might want to just go with the forward version

jagoosw:
I'm also pulling the changes from #3937 in but the diff with main will go away when that PR is merged

jagoosw:
> One comment is that I think the name is a bit ambiguous. There are several open boundary methods that have been developed that (in addition to outflow) include some sort of added perturbation for inflows to allow for faster turbulence development. When I read `PerturbationAdvection` I think of those methods. But if I understand the code correctly the "perturbation" here refers only to perturbations in the outflow, no?
> 
> Maybe we could name it `PerturbationOutflow` to avoid confusion?

I'm not exactly sure, if there was a point inflowing but the mean flow is out it still gets advected. Also when you add inflow pertubation its not advecting anything right?

tomchor:
> > One comment is that I think the name is a bit ambiguous. There are several open boundary methods that have been developed that (in addition to outflow) include some sort of added perturbation for inflows to allow for faster turbulence development. When I read `PerturbationAdvection` I think of those methods. But if I understand the code correctly the "perturbation" here refers only to perturbations in the outflow, no?
> > Maybe we could name it `PerturbationOutflow` to avoid confusion?
> 
> I'm not exactly sure, if there was a point inflowing but the mean flow is out it still gets advected. Also when you add inflow pertubation its not advecting anything right? You could also achieve that by setting e.g. `U(x, y, z, t) = U0(t) + rand()`

I thought the inflow was also advecting stuff into the domain. But it would advect a mostly-laminar flow. In any case, imo it would be nice to change the name to something that would differentiate from methods that add a perturbation to inflow velocities in order to speed up transition to turbulence. Do you have suggestions?

jagoosw:
> The results of simple cases seem more correct doing a forward step + `BoundaryAdjacentMean`, so I'm going to try pushing that and testing it against the headland case. You can currently choose forward or backward but we might want to just go with the forward version

I can't get it to work either way with a `BoundaryAdjacentMean` background velocity on both sides and the interior just stays near zero

jagoosw:
Also I can't convince myself that its valid to take a forward euler step because the interior nodes are approximately at the next time step already so I think I should revert those changes?

jagoosw:
> > > One comment is that I think the name is a bit ambiguous. There are several open boundary methods that have been developed that (in addition to outflow) include some sort of added perturbation for inflows to allow for faster turbulence development. When I read `PerturbationAdvection` I think of those methods. But if I understand the code correctly the "perturbation" here refers only to perturbations in the outflow, no?
> > > Maybe we could name it `PerturbationOutflow` to avoid confusion?
> > 
> > 
> > I'm not exactly sure, if there was a point inflowing but the mean flow is out it still gets advected. Also when you add inflow pertubation its not advecting anything right? You could also achieve that by setting e.g. `U(x, y, z, t) = U0(t) + rand()`
> 
> I thought the inflow was also advecting stuff into the domain. But it would advect a mostly-laminar flow. In any case, imo it would be nice to change the name to something that would differentiate from methods that add a perturbation to inflow velocities in order to speed up transition to turbulence. Do you have suggestions?

What kind of schemes do you mean?

jagoosw:
Also, on the naming. I think this scheme would traditionally be referred to as a radiation scheme but that doesn't really tell you what it does.

Maybe `Advective` and `AdvectiveOpenBoundaryCondition` for the helper would work?

On that note, should e.g. `OpenBoundaryCondition` etc. be camel_case since they're actually helper functions returning `BoundaryCondition{Open}` etc. or is this case allowed for aliases?

glwagner:
@tomchor is probably talking about methods along the lines of this paper: https://www.sciencedirect.com/science/article/pii/S0021999106001963?casa_token=bbQ2uunfOG0AAAAA:d6Sp2JC0ubNwDd7j7ExYwGXz1tnuN1qgnOOmmA0xhgDuXGFHwbsDR1rrQchs-qN-AAGLbRY-TFI

there might be references in there too. Basically open boundary conditions are important for CFD (eg numerical wind tunnels, etc) and so there is a big literature around these schemes in that context.

> On that note, should e.g. OpenBoundaryCondition etc. be camel_case since they're actually helper functions returning BoundaryCondition{Open} etc. or is this case allowed for aliases?

We use TitleCase for aliases.

tomchor:
> Is `Advective` and `AdvectiveOpenBoundaryCondition` okay then since it could be used for lots of different schemes?

Can you clarify what you mean by this?

tomchor:
By the way, I'm seeing compilation times (or more precisely times from constructing the model until _starting_ to initialize the simulation) go from around 30 seconds when using periodic BCs to 20 minutes when using `PerturbationAdvectionOBC`. Do you have any idea of what might be causing this?

glwagner:
My guess is that fill_halo_regions logic can be cleaned up. The code is pretty beastly in there

tomchor:
@jagoosw should we try to merge this by end of week? Do you mind if I merge main?

jagoosw:
happy for you to merge main

I think the main thing we need todo is add tests but I'm not exactly sure how

Are you still having the problem with the long start up time? I don't have that problem

tomchor:
> happy for you to merge main

:+1: 
 
> I think the main thing we need todo is add tests but I'm not exactly sure how

I think for now I'd be happy to test that we can build the BCs and that we can do a successful time-step with both a `Number` and `Function` as boundary conditions to relax to. We can also do whatever we're doing for `FlatExtrapolation`.

> Are you still having the problem with the long start up time? I don't have that problem

Not anymore! Compile times seem similar to `Periodic` boundary conditions. Thanks for fixing whatever you fixed :)

tomchor:
Also feel free to mark this as ready for review once you add tests.

jagoosw:
> Also feel free to mark this as ready for review once you add tests.

Should be ready now

jagoosw:
Boundary condition tests passed just some GPU tests failed because they lost Sverdrup

src/boundary_mean.jl
tomchor, line 1: @jagoosw you still need to remove this file from here, right?

Is this only used for the perturbation OBC? If so, it can go in the same directory.

src/boundary_mean.jl
```diff
@@ -0,0 +1,60 @@
...
+    (i, j, k), dims = boundary_adjacent_indices(val_side, grid, loc)
+    
+    total_area = CUDA.@allowscalar sum(An; dims)[i, j, k]
+
+    Ū = sum(u * An; dims) / total_area
+
+    bc.condition.value[] = CUDA.@allowscalar Ū[i, j, k]
+
+    return nothing
+end
```
tomchor, line 60:
```suggestion
function update_boundary_condition!(bc::MOPABC, val_side, u, model)
    grid = model.grid
    loc = instantiated_location(u)

    An = boundary_normal_area(val_side, grid)
    (i, j, k), dims = boundary_adjacent_indices(val_side, grid, loc)
    total_area = CUDA.@allowscalar sum(An; dims)[i, j, k]
    Ū = sum(u * An; dims) / total_area

    bc.condition.value[] = CUDA.@allowscalar Ū[i, j, k]
    return nothing
end
```

test/test_boundary_conditions_integration.jl
```diff
@@ -160,6 +164,67 @@ function test_flat_extrapolation_open_boundary_conditions(arch, FT)
...
+
+        obc = PerturbationAdvectionOpenBoundaryCondition(1, inflow_timescale = 0.1)
+
+        boundary_conditions = wall_normal_boundary_condition(Val(orientation), obc)
+
+        model = NonhydrostaticModel(; grid, boundary_conditions, timestepper = :QuasiAdamsBashforth2)
+
+        u = normal_velocity(Val(orientation), model)
+        
+        CUDA.@allowscalar u.data .= 1
```
tomchor, line 193: I think we're trying to avoid introducing new `@alloscalar` instances to testing. @glwagner can you confirm?

test/test_boundary_conditions_integration.jl
```diff
@@ -160,6 +164,67 @@ function test_flat_extrapolation_open_boundary_conditions(arch, FT)
...
+
+        @test all(u .== 1)
+
+        end_index = tuple(map(n -> ifelse(n == orientation, 5, 1), 1:3)...)
+        
+        CUDA.@allowscalar u[end_index...] = 2
+
+        time_step!(model, 1)
+        
+        CUDA.@allowscalar @test all(interior(u, end_index...) .== 1.5)
```
tomchor, line 205: I think it's great that we're testing values, and not just that things compile. However, I don't fully understand why these values are the appropriate ones (mostly referring to 1.5). Could you please add a comment here to explain the reasoning?

test/test_boundary_conditions_integration.jl
```diff
@@ -160,6 +164,67 @@ function test_flat_extrapolation_open_boundary_conditions(arch, FT)
...
+
+        obc = PerturbationAdvectionOpenBoundaryCondition((t) -> 0.1*t, inflow_timescale = 0.01, outflow_timescale = 0.5)
+
+        forcing = velocity_forcing(Val(orientation), Forcing((x, t) -> 0.1))
+
+        boundary_conditions = wall_normal_boundary_condition(Val(orientation), obc)
```
tomchor, line 211:
```suggestion
        obc = PerturbationAdvectionOpenBoundaryCondition((t) -> 0.1*t, inflow_timescale = 0.01, outflow_timescale = 0.5)
        forcing = velocity_forcing(Val(orientation), Forcing((x, t) -> 0.1))
        boundary_conditions = wall_normal_boundary_condition(Val(orientation), obc)
```

validation/open_boundaries/cylinder.jl
tomchor: This comment hasn't been resolved.

validation/open_boundaries/cylinder.jl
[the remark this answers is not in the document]
tomchor: This comment also hasn't been resolved.

validation/open_boundaries/cylinder.jl
```diff
@@ -1,127 +1,214 @@
...
+    Ny = 2^12
+    Nx = Ny
+elseif Re == 10^5
+    Ny = 2^13
+    Nx = Ny
+elseif Re == 10^6
+    Ny = 3/2 * 2^13 |> Int
+    Nx = Ny
+end
+=#
```
tomchor, line 96: There are several commented portions of this script. Maybe you forgot to clean up?

validation/open_boundaries/cylinder.jl
```diff
@@ -1,127 +1,214 @@
...
+using Oceananigans.Models.NonhydrostaticModels: ConjugateGradientPoissonSolver
+using Oceananigans.Solvers: DiagonallyDominantPreconditioner
+using Oceananigans.Operators: ℑxyᶠᶜᵃ, ℑxyᶜᶠᵃ
+using Oceananigans.Solvers: FFTBasedPoissonSolver
+using Printf
+using CUDA
 
-using Oceananigans, CairoMakie
-using Oceananigans.BoundaryConditions: FlatExtrapolationOpenBoundaryCondition
+using Oceananigans.BoundaryConditions: PerturbationAdvectionOpenBoundaryCondition
```
tomchor, line 9: Can we keep testing FlatExtrapolation also? Ideally we'd have a do loop and test both, and keep adding to it the more open boundary conditions we implement.

tomchor (changes requested):
Made a few comments, but it looks like there are old unresolved comments and some files here and there which do not seem to be ready for prime yet. let me know if anything needs clarification.

src/boundary_mean.jl
jagoosw: We never resolved this, the problem is that `AbstractOperations` are defined after `BoundaryConditions` so this can't go in the boundary module

test/test_boundary_conditions_integration.jl
jagoosw: I thought the problem was that tests currently have a global `@allowscalar`, but some tests need them or will be more complicated to write?

jagoosw:
@tomchor I think I've addressed your comments now

test/test_boundary_conditions_integration.jl
tomchor: I think you can write most of them simply without `allowscalar`, but I'm not an expert on this. For example you may be able to write something like `interior(u) .= 1` without it (I haven't tested it though), and it's a pretty simple change.

I say try to avoid it with simple statements like the one I just mentioned. If we can't do it simply and @glwagner doesn't manifest any opinions on it, we move forward. How does that sound?

validation/open_boundaries/cylinder.jl
```diff
@@ -1,127 +1,210 @@
...
+
+    simulation.output_writers[:jld2] = JLD2OutputWriter(model, outputs,
+                                                        schedule = IterationInterval(Int(2/Δt)),#TimeInterval(0.1),
+                                                        filename = prefix * "_fields.jld2",
+                                                        overwrite_existing = true,
+                                                        with_halos = true)
 
-ax = Axis(fig[1, 1], aspect = DataAspect())
+    simulation.output_writers[:drag] = JLD2OutputWriter(model, (; drag_force),
+                                                        schedule = IterationInterval(Int(0.1/Δt)),#TimeInterval(0.1),
```
tomchor, line 187:
```suggestion
                                                        schedule = IterationInterval(Int(0.1/Δt)),
```

validation/open_boundaries/cylinder.jl
```diff
@@ -1,127 +1,210 @@
...
+    ζ = ∂x(v) - ∂y(u)
 
-# plot the results
+    p = model.pressures.pNHS
 
-fig = Figure(size = (600, 600))
+    outputs = (; u, v, p, ζ)
+
+    simulation.output_writers[:jld2] = JLD2OutputWriter(model, outputs,
+                                                        schedule = IterationInterval(Int(2/Δt)),#TimeInterval(0.1),
```
tomchor, line 181:
```suggestion
                                                        schedule = IterationInterval(Int(2/Δt)),
```

validation/open_boundaries/cylinder.jl
```diff
@@ -1,127 +1,210 @@
...
+    vᵢ(x, y) = 1e-2 * randn()
+    set!(model, u=uᵢ, v=vᵢ)
 
-simulation.callbacks[:wizard] = Callback(wizard, IterationInterval(100))
+    #Δx = minimum_xspacing(grid)
+    Δt = max_Δt = 0.002#0.2 * Δx^2 * Re
 
-progress(sim) = @info "$(time(sim)) with Δt = $(prettytime(sim.Δt)) in $(prettytime(sim.run_wall_time))"
+    simulation = Simulation(model; Δt, stop_time)
+    #conjure_time_step_wizard!(simulation, cfl=1.0, IterationInterval(3); max_Δt)
```
tomchor, line 136:
```suggestion
```

src/boundary_mean.jl
tomchor: Hmm, I see. I don't have a good answer here, but I don't think it can be in `/src`. @glwagner @simone-silvestri  can you advise here?

Also, is there a simple way to define these functions without `AbstractOperation`s?

validation/open_boundaries/cylinder.jl
```diff
@@ -1,127 +1,210 @@
...
+using Oceananigans.Operators: ℑxyᶠᶜᵃ, ℑxyᶜᶠᵃ
+using Oceananigans.Solvers: FFTBasedPoissonSolver
+using Printf
+using CUDA
+
 using Oceananigans.BoundaryConditions: FlatExtrapolationOpenBoundaryCondition
+                                       PerturbationAdvectionOpenBoundaryCondition
+
+# there is some problem using ConjugateGradientPoissonSolver with TimeInterval because the timestep can go really small
+# so while I identify the issue I'm using IterationInterval and a fixed timestep
```
tomchor, line 13: Have you tried setting `minimum_relative_step`? (see https://github.com/CliMA/Oceananigans.jl/pull/3606)

jagoosw: This seems to fix the issue! I hadn't tried this since early november/december so I hadn't seen you'd made this fix so this is great, thank you

jagoosw: I'll update the validation script

src/boundary_mean.jl
jagoosw: It would be possible to write it without but it would not be trivial to do it for general grids and would end up reinventing the wheel to get the flux.

Perhaps we could move it to `Utils`?

validation/open_boundaries/cylinder.jl
```diff
@@ -18,10 +18,10 @@ using Oceananigans.BoundaryConditions: FlatExtrapolationOpenBoundaryCondition,
...
 
 ∂ₜu⃗ + (u⃗⋅∇)u⃗ = −∇P + ∇⋅τ⃗ + F⃗
-F⃗ₜ =∫ᵥF⃗dV = ∫ᵥ(∂ₜ u⃗ + (u⃗⋅∇)u⃗ + ∇P − ∇⋅τ⃗)dV
-F⃗ₜ=∫ᵥ(∂ₜu⃗)dV + ∮ₛ(u⃗(u⃗⋅n̂) + Pn̂ − τ⃗⋅n̂)dS
-Fᵤ=∫ᵥ ∂ₜu dV + ∮ₛ(u(u⃗⋅n̂) − τₓₓ)dS + ∮ₛPx̂⋅dS⃗
-Fᵤ=∫ᵥ ∂ₜ u dV − ∫ₛ₁(u² − 2ν∂ₓ u + P)dS + ∫ₛ₂(u² − 2ν∂ₓ u+P)dS − ∫ₛ₃uvdS + ∫ₛ₄ uvdS
+F⃗ₜ = ∫ᵥF⃗dV = ∫ᵥ(∂ₜ u⃗ + (u⃗⋅∇)u⃗ + ∇P − ∇⋅τ⃗)dV
+F⃗ₜ = ∫ᵥ(∂ₜu⃗)dV + ∮ₛ(u⃗(u⃗⋅n̂) + Pn̂ − τ⃗⋅n̂)dS
+Fᵤ = ∫ᵥ ∂ₜu dV + ∮ₛ(u(u⃗⋅n̂) − τₓₓ)dS + ∮ₛPx̂⋅dS⃗
+Fᵤ = ∫ᵥ ∂ₜ u dV − ∫ₛ₁(u² − 2ν∂ₓ u + P)dS + ∫ₛ₂(u² − 2ν∂ₓ u+P)dS − ∫ₛ₃uvdS + ∫ₛ₄ uvdS
```
glwagner, line 24: FYI what this looks like on my screen / browser

image

tomchor: I looks similarly messed up on my browser, but the actual document looks fine on my laptop:

image

Probably it's best to prioritize rendering on github, since a lot of people use it. I guess the easiest way to do that is to copy-paste stuff on a github text box and see what displays correctly and what doesn't. I just tested it out right now and it seems to work. The `\oint` character seems to not work for example.

src/boundary_mean.jl
tomchor: I'm okay with that.

validation/open_boundaries/cylinder.jl
jagoosw: > Probably it's best to prioritize rendering on github, since a lot of people use it. I guess the easiest way to do that is to copy-paste stuff on a github text box and see what displays correctly and what doesn't. I just tested it out right now and it seems to work. The `\oint` character seems to not work for example.

I tried this but then when I copy it over it renders wrong:

```math
\frac{\partial \vec{u}}{\partial t} + (\vec{u}\cdot\nabla)\vec{u}=-\nabla P + \nabla\cdot\vec{\tau} + \vec{F},
```
```math
\vec{F}_T=\int_\Omega\vec{F}dV = \int_\Omega\left(\frac{\partial \vec{u}}{\partial t} + (\vec{u}\cdot\nabla)\vec{u}+\nabla P - \nabla\cdot\vec{\tau}\right)dV,\newline
```
```math
\vec{F}_T=\int_\Omega\left(\frac{\partial \vec{u}}{\partial t}\right)dV + \oint_{\partial\Omega}\left(\vec{u}(\vec{u}\cdot\hat{n}) + P\hat{n} - \vec{\tau}\cdot\hat{n}\right)dS,
```

```math
F_u=\int_\Omega\left(\frac{\partial u}{\partial t}\right)dV + \oint_{\partial\Omega}\left(u(\vec{u}\cdot\hat{n}) - \tau_{xx}\right)dS + \int_{\partial\Omega}P\hat{x}\cdot d\vec{S},
```

```math
F_u=\int_\Omega\left(\frac{\partial u}{\partial t}\right)dV - \int_{\partial\Omega_1} \left(u^2 - 2\nu\frac{\partial u}{\partial x} + P\right)dS + \int_{\partial\Omega_2}\left(u^2 - 2\nu\frac{\partial u}{\partial x}+P\right)dS -  \int_{\partial\Omega_2} uvdS + \int_{\partial\Omega_4} uvdS
```

where the bounding box is $\Omega$ which is formed from the boundaries $\partial\Omega_1$, $\partial\Omega_2$, $\partial\Omega_3$, $\partial\Omega_4$ which have outward directed normals $-\hat{x}$, $\hat{x}$, $-\hat{y}$ and $\hat{y}$

Rendered in julia file:
∂ u → ∂ t + ( u → ⋅ ∇ ) u → = − ∇ P + ∇ ⋅ τ → + F → ,
F → T = ∫ Ω F → d V = ∫ Ω ( ∂ u → ∂ t + ( u → ⋅ ∇ ) u → + ∇ P − ∇ ⋅ τ → ) d V ,
F → T = ∫ Ω ( ∂ u → ∂ t ) d V + ∮ ∂ Ω ( u → ( u → ⋅ n ^ ) + P n ^ − τ → ⋅ n ^ ) d S ,
F u = ∫ Ω ( ∂ u ∂ t ) d V + ∮ ∂ Ω ( u ( u → ⋅ n ^ ) − τ x x ) d S + ∫ ∂ Ω P x ^ ⋅ d S → ,
F u = ∫ Ω ( ∂ u ∂ t ) d V − ∫ ∂ Ω 1 ( u 2 − 2 ν ∂ u ∂ x + P ) d S + ∫ ∂ Ω 2 ( u 2 − 2 ν ∂ u ∂ x + P ) d S − ∫ ∂ Ω 2 u v d S + ∫ ∂ Ω 4 u v d S

where the bounding box is Ω which is formed from the boundaries ∂ Ω 1 , ∂ Ω 2 , ∂ Ω 3 , ∂ Ω 4 which have outward directed normals − x ^ , x ^ , − y ^ and y

I'll play around and try and find an intermediary

jagoosw: How does this look:
```suggestion
∂ₜU + (U⋅∇)U = −∇P + ∇⋅τ + F⃗
F = ∫ᵥFdV = ∫ᵥ(∂U/∂t + (U⋅∇)U + ∇P − ∇⋅τ)dV
F = ∫ᵥ(∂U/∂t)dV + ∫ₛ(U(U⋅n̂) + Pn̂ − τ⋅n̂)dS
Fᵤ = ∫ᵥ ∂U/∂t dV + ∫ₛ(u(U⋅n̂) − τₓₓ)dS + ∮ₛPx̂⋅dS⃗
Fᵤ = ∫ᵥ ∂U/∂t dV − ∫ₛ₁(u² − 2ν∂ₓ u + P)dS + ∫ₛ₂(u² − 2ν∂ₓ u+P)dS − ∫ₛ₃uvdS + ∫ₛ₄ uvdS
```

I think I got all of the unrendered characters gone (vectors and oint?).

jagoosw: Strange that GH renders the original fine for me, maybe because I have a systemwide juliamono installation?

glwagner: not sure. But your update renders fine.

tomchor: It renders almost 100% fine for me, except for a couple of characters:

image

glwagner: oh yeah i missed that. I also see boxes for that term.

jagoosw: How does this look:
```suggestion
∂ₜU + (U⋅∇)U = −∇P + ∇⋅τ + F
F = ∫ᵥFdV = ∫ᵥ(∂U/∂t + (U⋅∇)U + ∇P − ∇⋅τ)dV
F = ∫ᵥ(∂U/∂t)dV + ∫ₛ(U(U⋅n̂) + Pn̂ − τ⋅n̂)dS
Fᵤ = ∫ᵥ ∂U/∂t dV + ∫ₛ(u(U⋅n̂) − τₓₓ)dS + ∫ₛPx̂⋅dS⃗
Fᵤ = ∫ᵥ ∂U/∂t dV − ∫ₛ₁(u² − 2ν∂ₓ u + P)dS + ∫ₛ₂(u² − 2ν∂ₓ u+P)dS − ∫ₛ₃uvdS + ∫ₛ₄ uvdS
```
how about now?

tomchor: Almost 
image

tomchor: Also for the record, I just set juliamono as my browser font and things display correctly. So this appears to be system-dependent. Given that, should we still try to tailor the characters to appear on github? Whether or not it displays is apparently dependent on the browser, rather than on github...

glwagner: fair point

jagoosw: How about now:
```suggestion
∂ₜU + (U⋅∇)U = −∇P + ∇⋅τ + F
F = ∫ᵥFdV = ∫ᵥ(∂U/∂t + (U⋅∇)U + ∇P − ∇⋅τ)dV
F = ∫ᵥ(∂U/∂t)dV + ∫ₛ(U(U⋅n̂) + Pn̂ − τ⋅n̂)dS
Fᵤ = ∫ᵥ ∂U/∂t dV + ∫ₛ(u(U⋅n̂) − τₓₓ)dS + ∫ₛPx̂⋅dS
Fᵤ = ∫ᵥ ∂U/∂t dV − ∫ₛ₁(u² − 2ν∂ₓ u + P)dS + ∫ₛ₂(u² − 2ν∂ₓ u+P)dS − ∫ₛ₃uvdS + ∫ₛ₄ uvdS
```

src/boundary_mean.jl
jagoosw: Ah utils is imported long before fields and boundary conditions too because it includes the kernel launching stuff

jagoosw: I would also not object to removing it from the source code

validation/open_boundaries/cylinder.jl
tomchor: Seems to work now (on Fira Code, which I think is the default for Chrome)! Thanks for the patience @jagoosw. Let's resolve this issue and move on :)

src/boundary_mean.jl
tomchor: Can you explain why that piece of code is necessary and what would be the impact of removing it? It's not 100% clear to me.

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -0,0 +1,242 @@
...
+    (Uⁿ⁺¹ - Uⁿ) / Δt + (u′ⁿ⁺¹ - u′ⁿ) / Δt = - Uⁿ⁺¹ (u′ⁿ⁺¹ᵢ - u′ⁿ⁺¹ᵢ₋₁) / Δx + Fᵤ.
+
+This can not be solved for general forcing, but if we assume the dominant forcing is
+relaxation to the mean velocity (i.e. u′→0) then Fᵤ = -u′ / τ then we can find u′ⁿ⁺¹:
+    u′ⁿ⁺¹ = (uⁿ + Ũu′ⁿ⁺¹ᵢ₋₁ - Uⁿ⁺¹) / (1 + Ũ + Δt/τ),
+
+where Ũ = U Δt / Δx, then uⁿ⁺¹ is:
+    uⁿ⁺¹ = (uᵢⁿ + Ũuᵢ₋₁ⁿ⁺¹ + Uⁿ⁺¹τ̃) / (1 + τ̃ + U)
+    
+where τ̃ = Δt/τ.
```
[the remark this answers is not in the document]
tomchor: If I use FiraCode on chrome then some of the stuff in the first line of code doesn't render correctly:

image

On firefox it all renders fine:

image

And when I manually switch to JuliaMono it all renders fine.

But after finding out that this is all browser/machine dependent, I do think we shouldn't spend too much time trying to "fix" this. I say we move on from tweaking the unicode and address the rest of the comments so that we can merge.

jagoosw:
I can no longer see the buildkite logs (its says I have to be part of the buildkite organisation) so could someone check what the problem with those tests was please?

tomchor:
A couple of them are failing with an `Out of GPU memory` error, but there's also this error:

image

tomchor:
We still have a comment that hasn't been addressed, right? About what to do with the `boundary_mean.jl` file. After that's dealt with, I can make tests pass if you want.

jagoosw:
> We still have a comment that hasn't been addressed, right? About what to do with the `boundary_mean.jl` file. After that's dealt with, I can make tests pass if you want.

yeah that's the only problem, whats wrong with the tests?

tomchor:
> > We still have a comment that hasn't been addressed, right? About what to do with the `boundary_mean.jl` file. After that's dealt with, I can make tests pass if you want.
> 
> yeah that's the only problem, whats wrong with the tests?

No idea what's up with GPU errors. @glwagner mentioned at the some point that that server is pretty full, so I'm assuming some cleanup will be done soon. (I don't have access to that server btw, so I only know what people tell me.)

jagoosw:
On the boundary mean issue, perhaps we should make a new module called something like `HelperFunctions` since I imagine there's other things that are easiest to define with all of the other infrastructure (I.e. abstract operators, etc.) so could be user defined, but are also nice to have in the source code?

glwagner:
> On the boundary mean issue, perhaps we should make a new module called something like `HelperFunctions` since I imagine there's other things that are easiest to define with all of the other infrastructure (I.e. abstract operators, etc.) so could be user defined, but are also nice to have in the source code?
> 
> @glwagner any thoughts?

Can you elaborate? What you're suggesting might make sense but I need more background, motivation, design proposal, etc.

jagoosw:
> Can you elaborate? What you're suggesting might make sense but I need more background, motivation, design proposal, etc.

Basically, the best and most general way to implement a boundary condition that uses the boundary mean velocity is by using AbstractOperations, but the problem is that they're defined after boundary conditions so it can't go in the boundary conditions module. 

I think there is kind of a grey area of things like this that aren't part of the core functionality of the code and _could_ be implemented by a user, but are important and useful for us to support and maintain.

glwagner:
> > Can you elaborate? What you're suggesting might make sense but I need more background, motivation, design proposal, etc.
> 
> Basically, the best and most general way to implement a boundary condition that uses the boundary mean velocity is by using AbstractOperations, but the problem is that they're defined after boundary conditions so it can't go in the boundary conditions module.
> 
> I think there is kind of a grey area of things like this that aren't part of the core functionality of the code and _could_ be implemented by a user, but are important and useful for us to support and maintain.

Possibly, this belongs in `Models` then.

jagoosw:
> Possibly, this belongs in `Models` then. Are we assuming something specific about the presence of a velocity field vs tracer field? In that case, this is not a "finite volume engine" capability, but rather a "fluid dynamics model" capability, and it does not belong in `BoundaryConditions` (which has primitive utilities that should be independent of fluid dynamics) so isn't fluid specific.

That's a good idea, it doesn't make any assumptions about what it is and is just computing the mean of whatever field it is associated with on a boundary:
https://github.com/CliMA/Oceananigans.jl/blob/045a6b3e315b45e63b17c714988c39bdf4ab0f79/src/boundary_mean.jl#L45-L56

glwagner:
shouldn't you define a kernel for `sum`? the way implemented seems like its going to be slow

glwagner:
Also this function takes in `model` which to me clearly says it belongs in `model`. This must only work with specific models right? It will not work as a generic boundary condition

jagoosw:
> Also this function takes in `model` which to me clearly says it belongs in `Models`. This must only work with specific models right? It will not work as a generic boundary condition for any Field / with any `fill_halo_regions!`.

Model doesn't actually do anything (if we wrote `u.grid` rather than `model.grid`) so maybe I should rewrite this like:

```julia
funciton (bam::BoundaryAdjacentMean)(val_side, u)
   ... the above function
   return bam.value
end

update_boundary_condition!(bc::MOPABC, val_side, u, model) = (bc.condition(val_side, u); nothing)
```

jagoosw:
> shouldn't you define a kernel for `sum`? the way implemented seems like its going to be slow

Maybe? Doesn't `sum` go back to this reduction: 
https://github.com/CliMA/Oceananigans.jl/blob/b67ad568a18f24cfbe983fac99bd1395d54b71da/src/Fields/field.jl#L695-L716

I guess its more efficient to preallocate and use the preallocating version? We could put a field like:
https://github.com/CliMA/Oceananigans.jl/blob/b67ad568a18f24cfbe983fac99bd1395d54b71da/src/Fields/field.jl#L703
in the boundary adjacent mean struct.

I guess summing is always inefficient since it has to be serialized though right?

jagoosw:
> > shouldn't you define a kernel for `sum`? the way implemented seems like its going to be slow
> 
> Maybe? Doesn't `sum` go back to this reduction:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/b67ad568a18f24cfbe983fac99bd1395d54b71da/src/Fields/field.jl#L695-L716
> 
> I guess its more efficient to preallocate and use the preallocating version? We could put a field like:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/b67ad568a18f24cfbe983fac99bd1395d54b71da/src/Fields/field.jl#L703
> 
> in the boundary adjacent mean struct.
> 
> I guess summing is always inefficient since it has to be serialized though right?

glwagner:
Where is `update_boundary_condition!` defined? In my opinion, `update_boundary_condition!` is defining a model-specific interface.

glwagner:
> > shouldn't you define a kernel for `sum`? the way implemented seems like its going to be slow
> 
> Maybe? Doesn't `sum` go back to this reduction:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/b67ad568a18f24cfbe983fac99bd1395d54b71da/src/Fields/field.jl#L695-L716
> 
> I guess its more efficient to preallocate and use the preallocating version? We could put a field like:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/b67ad568a18f24cfbe983fac99bd1395d54b71da/src/Fields/field.jl#L703
> 
> 
> in the boundary adjacent mean struct.
> I guess summing is always inefficient since it ~has to be serialized~ can't be straightforwardly parallelised without making a race condition though right?

I'm confused. Is this working inside a kernel? You cannot launch a kernel (to compute `sum`) from within a kernel. Does this work on GPU?

jagoosw:
> Where is `update_boundary_condition!` defined? In my opinion, `update_boundary_condition!` is defining a model-specific interface.

The fallback is defined in boundary conditions, but this is the first time we've actually implemented on in the source code. Happy for it to go in the model modules.

jagoosw:
> > > shouldn't you define a kernel for `sum`? the way implemented seems like its going to be slow
> > 
> > 
> > Maybe? Doesn't `sum` go back to this reduction:
> > https://github.com/CliMA/Oceananigans.jl/blob/b67ad568a18f24cfbe983fac99bd1395d54b71da/src/Fields/field.jl#L695-L716
> > 
> > I guess its more efficient to preallocate and use the preallocating version? We could put a field like:
> > https://github.com/CliMA/Oceananigans.jl/blob/b67ad568a18f24cfbe983fac99bd1395d54b71da/src/Fields/field.jl#L703
> > 
> > in the boundary adjacent mean struct.
> > I guess summing is always inefficient since it ~has to be serialized~ can't be straightforwardly parallelised without making a race condition though right?
> 
> I'm confused. Is this working inside a kernel? You cannot launch a kernel (to compute `sum`) from within a kernel. Does this work on GPU? Furthermore, `@allowscalar` should never be used to fetch a generic index (eg i, j, k). The only acceptable usage is to fetch a single known index, typically `first` or `1, 1, 1`.

Not from within a kernel, and yeah I used in on GPU. We only want a single value where 2 of i, j, k and k will be 1 but the other will either be 1 or the end. I don't know how else we can pull the single value out?

glwagner:
Maybe I was confused then. You're saying that `i, j, k` don't vary and take just a single value per boundary condition?

jagoosw:
> Maybe I was confused then. You're saying that `i, j, k` don't vary and take just a single value per boundary condition?

yes, it will always be the same and just depends on which boundary its acting on.

glwagner:
> > Maybe I was confused then. You're saying that `i, j, k` don't vary and take just a single value per boundary condition?
> 
> yes, it will always be the same and just depends on which boundary its acting on.

Can you potentially choose a name that conveys this? We use `i, j, k` inside kernels, to mean a local index that varies over the whole grid from 1:Nx, 1:Ny, 1:Nz. When we choose specific indices, we change the notation, like `iB` of `i0` for example.

src/Models/boundary_mean.jl
```diff
@@ -7,31 +7,36 @@ using Oceananigans.BoundaryConditions: BoundaryCondition, Open, PerturbationAdve
 import Adapt: adapt_structure
 import Oceananigans.BoundaryConditions: update_boundary_condition!
 
-struct BoundaryAdjacentMean{V}
-   value :: V
+struct BoundaryAdjacentMean{FF, BV}
+   flux_field :: FF
+        value :: BV
```
glwagner, line 12:
```suggestion
    flux_field :: FF
         value :: BV
```

src/Models/boundary_mean.jl
```diff
@@ -7,31 +7,36 @@ using Oceananigans.BoundaryConditions: BoundaryCondition, Open, PerturbationAdve
...
 
-   BoundaryAdjacentMean(; value::BV = Ref(0.0)) where BV = new{BV}(value)
+   BoundaryAdjacentMean(grid, side; 
+                        flux_field::FF = boundary_reduced_field(Val(side), grid),
+                        value::BV = Ref(0.0)) where {FF, BV} = new{FF, BV}(value)
```
glwagner, line 16: don't hard code Float64

src/Models/boundary_mean.jl
tomchor, line 1: Can you include some documentation in the file? Perhaps some words at the top to give it context and a docstring in the main function?

src/Models/boundary_mean.jl
```diff
@@ -0,0 +1,75 @@
...
+@inline first_interior_index(::Union{Val{:west}, Val{:east}}, ::Tuple{Center, <:Any, <:Any}) = 1
+@inline first_interior_index(::Union{Val{:west}, Val{:east}}, ::Tuple{Face, <:Any, <:Any}) = 2
+
+@inline first_interior_index(::Union{Val{:south}, Val{:north}}, ::Tuple{<:Any, Center, <:Any}) = 1
+@inline first_interior_index(::Union{Val{:south}, Val{:north}}, ::Tuple{<:Any, Face, <:Any}) = 2
+
+@inline first_interior_index(::Union{Val{:bottom}, Val{:top}}, ::Tuple{<:Any, <:Any, Center}) = 1
+@inline first_interior_index(::Union{Val{:bottom}, Val{:top}}, ::Tuple{<:Any, <:Any, Face}) = 2
+
+function compute_boundary_mean!(bam::BoundaryAdjacentMean, val_side, u)
```
tomchor, line 50: docstring for this function, please

src/Models/boundary_mean.jl
```diff
@@ -0,0 +1,125 @@
...
+using Oceananigans.BoundaryConditions: BoundaryCondition, Open, PerturbationAdvection
+
+import Adapt: adapt_structure
+import Base: summary, show
+import Oceananigans.BoundaryConditions: update_boundary_condition!
+
+"""
+    BoundaryAdjacentMean
+
+Stores the bounary mean `value` of a field. Updated by calling
```
tomchor, line 14:
```suggestion
Stores the boundary mean `value` of a `Field`. Updated by calling 
```

test/test_boundary_conditions_integration.jl
```diff
@@ -160,6 +164,69 @@ function test_flat_extrapolation_open_boundary_conditions(arch, FT)
...
+
+        model = NonhydrostaticModel(; grid, boundary_conditions, timestepper = :QuasiAdamsBashforth2)
+
+        u = normal_velocity(Val(orientation), model)
+        
+        view(parent(u), :, :, :) .= 1
+
+        time_step!(model, 1)
+
+        @test all(view(parent(u), :, :, :) .== 1)
```
tomchor, line 197: Just some formatting suggestions to aid on readability. Kinda like separating the code into paragraphs that are on the same theme. I don't feel strongly about them so take it or leave it.

```suggestion
        u = normal_velocity(Val(orientation), model)
        view(parent(u), :, :, :) .= 1
        time_step!(model, 1)
        @test all(view(parent(u), :, :, :) .== 1)
```

test/test_boundary_conditions_integration.jl
```diff
@@ -160,6 +164,69 @@ function test_flat_extrapolation_open_boundary_conditions(arch, FT)
...
+        
+        view(parent(u), absolute_end_index...) .= 2
+
+        time_step!(model, 1)
+
+        end_index = tuple(map(n -> ifelse(n == orientation, 5, 1), 1:3)...)
+        
+        # uⁿ⁺¹ = (uⁿ + Ūuⁿ⁺¹ᵢ₋₁) / (1 + Ū)
+        # Δx = Δt = U = 1 -> uⁿ⁺¹ = (uⁿ + uⁿ⁺¹ᵢ₋₁) / 2 = 1.5
+        @test all(interior(u, end_index...) .== 1.5)
```
tomchor, line 209:
```suggestion
        absolute_end_index = tuple(map(n -> ifelse(n == orientation, 8, 1), 1:3)...)
        view(parent(u), absolute_end_index...) .= 2
        time_step!(model, 1)
        end_index = tuple(map(n -> ifelse(n == orientation, 5, 1), 1:3)...)
        
        # uⁿ⁺¹ = (uⁿ + Ūuⁿ⁺¹ᵢ₋₁) / (1 + Ū)
        # Δx = Δt = U = 1 -> uⁿ⁺¹ = (uⁿ + uⁿ⁺¹ᵢ₋₁) / 2 = 1.5
        @test all(interior(u, end_index...) .== 1.5)
```

test/test_boundary_conditions_integration.jl
```diff
@@ -160,6 +164,69 @@ function test_flat_extrapolation_open_boundary_conditions(arch, FT)
...
+                              timestepper = :QuasiAdamsBashforth2,
+                              forcing)
+        
+        u = normal_velocity(Val(orientation), model)
+        
+        for _ in 1:100
+            time_step!(model, 0.1)
+        end
+
+        @test all(isapprox.(interior(u), 1, atol = 0.1))
```
tomchor, line 226:
```suggestion
        model = NonhydrostaticModel(; grid,
                                      boundary_conditions,
                                      timestepper = :QuasiAdamsBashforth2,
                                      forcing)

        u = normal_velocity(Val(orientation), model)        
        for _ in 1:100
            time_step!(model, 0.1)
        end
        @test all(isapprox.(interior(u), 1, atol = 0.1))
```

tomchor:
Thanks, @jagoosw. This is looking good and I think we're very close to merging. I do have a couple of questions though.

1. Looking at the code, I don't see `BoundaryAdjacentMean` used anywhere. Maybe I missed something, but can you clarify why (and where) it's needed?
2. Related to 1: I also don't see `BoundaryAdjacentMean` being tested anywhere. If it's necessary, we should probably include at least a simple test.

test/test_boundary_conditions_integration.jl
glwagner: I like this better, I find double spaced code harder to read

jagoosw:
> Thanks, @jagoosw. This is looking good and I think we're very close to merging. I do have a couple of questions though.
> 
>     1. Looking at the code, I don't see `BoundaryAdjacentMean` used anywhere. Maybe I missed something, but can you clarify why (and where) it's needed?
> 
>     2. Related to 1: I also don't see `BoundaryAdjacentMean` being tested anywhere. If it's necessary, we should probably include at least a simple test.

Yeah its not in use anywhere, it can be put in as the "condition" for the boundary condition like:
```julia
bam = BoundaryAdjacentMean(grid, :east)
obc = PerturbationAdvectionOpenBoundaryCondition(bam)
```

I've added a doctest which checks that it is functioning, do you think thats sufficient?

jagoosw:
I have hopefully fixed the tests now and made it so it tests both sides too

test/test_boundary_conditions_integration.jl
```diff
@@ -170,7 +266,7 @@ test_boundary_conditions(C, FT, ArrayType) = (integer_bc(C, FT, ArrayType),
                                               parameterized_field_dependent_function_bc(C, FT, ArrayType),
                                               discrete_function_bc(C, FT, ArrayType),
                                               parameterized_discrete_function_bc(C, FT, ArrayType))
-
+#=
```
tomchor, line 269:
```suggestion
```
?

test/test_boundary_conditions_integration.jl
```diff
@@ -336,6 +432,8 @@ test_boundary_conditions(C, FT, ArrayType) = (integer_bc(C, FT, ArrayType),
             A = typeof(arch)
             @info "  Testing open boundary conditions [$A, $FT]..."
             test_flat_extrapolation_open_boundary_conditions(arch, FT)
+            test_pertubation_advection_open_boundary_conditions(arch, FT)
         end
     end
 end
+=#
```
tomchor, line 439:
```suggestion
```

tomchor (approved):

tomchor:
@jagoosw I took the liberty to un-comment some tests that was left commented out by mistake and apparently there's one more error. We're very close though!

jagoosw:
> @jagoosw I took the liberty to un-comment some tests that was left commented out by mistake and apparently there's one more error. We're very close though!

Great, thank you. I think I have it sorted now

tomchor:
Hmm, for some reason I can't make the distributed tests pass. Does this error ring a bell to anyone?

image

cc @glwagner

glwagner:
we're looking into it cc @charleskawczynski

PR #3979, opened 2024-12-11
Fix horizontally `Flat` WENO advection

simone-silvestri:
I recently encountered a bounds error when trying `WENOVectorInvariant` on a horizontally flat grid. 

The problem was two-fold
- the `Flat` interpolators were not updated to the new syntax
- we were using a wrong signature for interpolations in the WENO stencil that does not account for the grid topology
- (bonus!) there are no tests for `WENOVectorInvariant`

This PR should resolve these issues

edit, I will also remove the "no-grid" interpolation operators to not allow something like this.


glwagner (approved):
😱

PR #3985, opened 2024-12-13
(0.95.2) Bugfix in the Polar BC

simone-silvestri:
Which now are tested also for `Flat` grids in the meridional direction. 

Required by https://github.com/CliMA/ClimaOcean.jl/pull/297
(see https://buildkite.com/clima/climaocean-ci/builds/2146#0193bf37-560c-4923-83d0-3a891fb843e7)


glwagner (approved):

glwagner:
When tests pass here, I will bump to 0.95.2 and then merge.

glwagner:
nooo that restarted the tests

glwagner:
We have a way to bypass the test restrictions now which is useful, since the update here was a trivial bump of Project.toml

simone-silvestri:
Ah sorry 😅  I forgot we can avoid updating now

PR #3991, opened 2024-12-14
Update the docstring for `HydrostaticFreeSurfaceModel`

simone-silvestri:


navidcy (approved):

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -93,6 +94,8 @@ Keyword arguments
                preallocated `CenterField`s.
   - `forcing`: `NamedTuple` of user-defined forcing functions that contribute to solution tendencies.
   - `closure`: The turbulence closure for `model`. See `Oceananigans.TurbulenceClosures`.
+  - `timestepper`: A symbol that specifies the time-stepping method. 
+                   Either `:QuasiAdamsBashforth2` (default) or `:SplitRungeKutta3`.
```
glwagner, line 98: Why isn't the AB2 scheme "split"? Weirded out that we use the same name for AB2 for nonhydrostatic/hydrostatic but a different name for RK3.

navidcy: Just noticed this. I suggest to remove the split from the timestepper name?

simone-silvestri: The scheme is called like that, as opposed to the `:RungeKutta3` scheme for the non-hydrostatic model. Are you suggesting a name change to the method or only in the constructor?

glwagner: Why is RK3 "split" but AB2 is not?

PR #3993, opened 2024-12-15
Remove grid requirements from a `Model` in a `Simulation`

simone-silvestri:
This PR allows us to remove the grid from the `OceanSeaIceModel` type defined in ClimaOcean.jl
(see https://github.com/CliMA/ClimaOcean.jl/blob/a193aee12de264600dda9ca196acdb52ee3a726d/src/OceanSeaIceModels/ocean_sea_ice_model.jl#L20)

### DIFF
--- a/src/Models/Models.jl
+++ b/src/Models/Models.jl
@@ -23,6 +23,8 @@ import Oceananigans.Architectures: architecture
 import Oceananigans.TimeSteppers: reset!
 import Oceananigans.Solvers: iteration
 
+import Base
+
 # A prototype interface for AbstractModel.
 #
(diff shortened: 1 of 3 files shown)


src/Models/Models.jl
```diff
@@ -23,6 +23,8 @@ import Oceananigans.Architectures: architecture
 import Oceananigans.TimeSteppers: reset!
 import Oceananigans.Solvers: iteration
 
+import Base
```
glwagner, line 26: you don't need this if you use `Base.eltype` below

glwagner (approved):

PR #4001, opened 2024-12-16
Try to add plot labels automagically within Makie extension

glwagner:
This PR adds plot labels to Makie plots. It only adds them when using the plotting function that creates an axis, eg `heatmap(c)`. It doesn't work for `heatmap!(ax, c)`.

Example:

```julia
using Oceananigans
using GLMakie

grid = RectilinearGrid(size=(4, 4, 1), extent=(1, 1, 1))
c = CenterField(grid)
set!(c, (x, y, z) -> randn())
heatmap(c)
```

gives 

image

but

```julia
heatmap(c; axis=(xlabel="yo", ylabel="hi"))
```

preserves the user labels,

image

and

```julia
fig = Figure()
ax = Axis(fig[1, 1])
heatmap!(ax, c)
```

doesn't add labels.

### DIFF
--- a/ext/OceananigansMakieExt.jl
+++ b/ext/OceananigansMakieExt.jl
@@ -25,13 +25,22 @@ function drop_singleton_indices(N)
     end
 end
 
+"""
+    deduce_dimensionality(f)
+
+Deduce the dimensionality of the field `f` and return a 3-tuple `d1, d2, D`, where
+`d1` is the first dimension along which `f` varies, `d2` is the second dimension (if any),
+and `D` is the total dimensionality of `f`.
+"""
 function deduce_dimensionality(f)
-    # Indices of the non-zero dimensions
+    # Find indices of the dimensions along which `f` varies
     d1 = findfirst(n -> n > 1, size(f))
     d2 =  findlast(n -> n > 1, size(f))
-    # Deduce dimensionality
+
+    # Deduce total dimensionality
     Nx, Ny, Nz = size(f)
     D = (Nx > 1) + (Ny > 1) + (Nz > 1)
+
     return d1, d2, D
 end
 
@@ -44,13 +53,17 @@ function _create_plot(F::Function, attributes::Dict, f::Field)
     if !(:axis ∈ keys(attributes)) # Let's try to add labels automatically
         d1, d2, D = deduce_dimensionality(f) 
         grid = f.grid
+
         if D === 1 # 1D plot
-            if d1 === 1 # horizontal
+
+            # See `convert_field_argument` for this horizontal/vertical plotting convention.
+            if d1 === 1 # This is a horizontal plot, so we add xlabel
                 axis = (; xlabel=axis_str(grid, 1))
-            else # vertical plot
+            else # vertical plot with a ylabel
                 axis = (; ylabel=axis_str(grid, d1))
             end
-        elseif D === 2
+
+        elseif D === 2 # it's a two-dimensional plot
             axis = (xlabel=axis_str(grid, d1), ylabel=axis_str(grid, d2))
         else
             throw(ArgumentError("Cannot create axis labels for a 3D field!"))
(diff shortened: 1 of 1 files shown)


glwagner:
@tomchor is this what you're looking for?

tomchor:
> @tomchor is this what you're looking for?

Yes! Thanks. I see it even works with 1D output and `lines`. Nice :)

tomchor (approved):

tomchor:
This is a parallel discussion, but at some point it would be nice to allow 3D outputs. Also maybe automatic colorbars for 2D plots?

glwagner:
Automatic colorbars can't be implemented in the scope of `convert_arguments` I don't think, since this is actually a modification to `Figure` (creating another Axis). We could create a new plotting function though that would add both a `heatmap` and a `Colorbar` to some figure, like `heatmap_colorbar!(fig)` for example.

My only concern is whether its really worth the complication since we have to design an API for it. I personally think it's easy enough to inspect the max/min values of a `Field` at the REPL that the colorbar is not that useful interactively. For presenting information the colorbar is useful, but in that case it's just an extra line in a script.

As for 3D, I would like to build something that can work either with 3D data or plotting slices in 3D. It's a little bit involved though, but well worth it I think. An isosurface utility is also possible, but I have less experience with those kinds of plots.

tomchor:
> As for 3D, I would like to build something that can work either with 3D data or plotting slices in 3D. It's a little bit involved though, but well worth it I think. An isosurface utility is also possible, but I have less experience with those kinds of plots.

Wouldn't extending `volume` and `volumeslices` to Oceananigans types be enough? (Except for the isosurfaces though)

glwagner:
@tomchor feel free to close the associated issue if you feel this meets the requirements, or change the issue to discuss `volume` and `volumeslices`.

tomchor:
> I don't know, I haven't used those. If you take a look at the plotting code, you'll see that we don't extend the plot functions directly. We extend `_create_plot` and `convert_arguments`. Possibly, 3D already works. You could give it a shot.

I already tried it. Got greeted with this:

```
julia> volumeslices(c)
ERROR: ArgumentError: Cannot convert_arguments for a 3D field!
```

> For 3D, I am not that interested in volume personally. I prefer to assemble slices of data along various planes to construct a 3D picture. It's more feasible to create animations this way, since one only has to time-resolve data on slices.

Is that different from what [`volumeslices`](https://docs.makie.org/dev/reference/plots/volumeslices) does?

glwagner:
You have to change the source code to remove the error

glwagner:
`volumeslices` looks like it requires 3D data?

https://docs.makie.org/dev/reference/plots/volumeslices

The idea is similar, however, and can possibly be achieved with multiple calls to `volumeslices`.

tomchor:
Yeah, `volumeslices` requires 3D data. I guess part of your point is that you can save space and write separate slices of 2D data if you manually put those slices together, which makes sense.

On the Oceananigans side, we could develop the functionality of taking different slices and assembling them automatically in 3D space instead of manually assembling them like we do in [this example](https://clima.github.io/OceananigansDocumentation/stable/literated/baroclinic_adjustment/). Shouldn't be too hard since we have all the information. And I think these plots are extremely useful (and while not super hard, they're not trivial to make).

tomchor:
I don't even think we need to specify `east`, `west`, etc. We can figure out place and orientation from any given slice by seeing it's associated dimensions, no?

glwagner:
how can you distinguish between east and west?

tomchor:
I don't think you need to, right? Plotting several slices side by side can be useful too. Like this picture as an example:

image

There's no `bottom` or `top` slice in this case. Unless I'm misunderstanding your point?

glwagner:
No I think you're right. With windowed fields, we can get the coordinates. I think that's true

PR #4003, opened 2024-12-17
Generalized Checkpointing plus interface for `set!(::Simulation, pickup)`

glwagner:
This will allow us to extend the interface for special `Simulation`s that require special checkpointing procedures.


simone-silvestri (approved):
nice the checkpoints become smaller for RK3.

PR #4008, opened 2024-12-18
(0.95.5) Correctly account for field names in `ContinuousBoundaryFunctions`

simone-silvestri:
continuous boundary functions with field dependencies are broken at the moment because of the recent changes with split explicit that changed the order of the fields outputted by `fields(model)`.

This PR corrects this bug and introduces 
(1) some tests 
(2) a `field_names` function to be able to change the field names in a `HydrostaticFreeSurfaceModel` depending on different inputs

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
@@ -75,9 +75,18 @@ Return a flattened `NamedTuple` of the fields in `model.velocities`, `model.free
 `model.tracers`, and any auxiliary fields for a `HydrostaticFreeSurfaceModel` model.
 """
 @inline fields(model::HydrostaticFreeSurfaceModel) = 
-    merge(hydrostatic_fields(model.velocities, model.free_surface, model.tracers),
-          model.auxiliary_fields,
-          biogeochemical_auxiliary_fields(model.biogeochemistry))
+    merge(hydrostatic_fields(model.velocities, model.free_surface, model.tracers), model.auxiliary_fields)
+
+"""
+    field_names(model)
+
+Return a tuple of the field names associated with `HydrostaticFreeSurfaceModel`, which is equivalent to doing `keys(fields(model))`.
+"""
+@inline field_names(model::HydrostaticFreeSurfaceModel) = field_names(model.free_surface, model.tracernames, model.auxiliary_fields)
+
+@inline field_names(free_surface, tracernames, auxiliary_fields) = (:u, :v, :w, tracernames..., :η, keys(auxiliary_fields)...)
+@inline field_names(::Nothing, tracernames, auxiliary_fields)    = (:u, :v, :w, tracernames..., keys(auxiliary_fields)...)
+@inline field_names(::SplitExplicitFreeSurface, tracernames, auxiliary_fields) = (:u, :v, :w, tracernames..., :η, :U, :V, keys(auxiliary_fields)...)
 
 """
     prognostic_fields(model::HydrostaticFreeSurfaceModel)
@@ -111,11 +120,11 @@ Return a flattened `NamedTuple` of the prognostic fields associated with `Hydros
 
 @inline hydrostatic_fields(velocities, free_surface::SplitExplicitFreeSurface, tracers) = merge((u = velocities.u,
                                                                                                  v = velocities.v,
-                                                                                                 w = velocities.w,
-                                                                                                 η = free_surface.η,
-                                                                                                 U = free_surface.barotropic_velocities.U,
-                                                                                                 V = free_surface.barotropic_velocities.V),
-                                                                                                 tracers)
+                                                                                                 w = velocities.w),
+                                                                                                 tracers,
+                                                                                                 (; η = free_surface.η,
+                                                                                                    U = free_surface.barotropic_velocities.U,
+                                                                                                    V = free_surface.barotropic_velocities.V))
 
 @inline hydrostatic_fields(velocities, ::Nothing, tracers) = merge((u = velocities.u,
                                                                     v = velocities.v,
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.95.2"
+version = "0.95.4"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/BoundaryConditions/continuous_boundary_function.jl
+++ b/src/BoundaryConditions/continuous_boundary_function.jl
@@ -72,7 +72,7 @@ The regularization of `bc.condition::ContinuousBoundaryFunction` requries
    of the boundary.
 """
 function regularize_boundary_condition(bc::BoundaryCondition{C, <:ContinuousBoundaryFunction},
-                                       grid, loc, dim, Side, prognostic_field_names) where C
+                                       grid, loc, dim, Side, field_names) where C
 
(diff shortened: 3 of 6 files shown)


src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
```diff
@@ -75,9 +75,19 @@ Return a flattened `NamedTuple` of the fields in `model.velocities`, `model.free
 `model.tracers`, and any auxiliary fields for a `HydrostaticFreeSurfaceModel` model.
 """
 @inline fields(model::HydrostaticFreeSurfaceModel) = 
-    merge(hydrostatic_fields(model.velocities, model.free_surface, model.tracers),
-          model.auxiliary_fields,
-          biogeochemical_auxiliary_fields(model.biogeochemistry))
```
simone-silvestri, line 80: I am a bit confused about this function. I think that the biogeochemical auxiliary fields have been added to the `auxiliary_fields` during model construction, so we shouldn't need an extra `biogeochemical_auxiliary_fields(model.biogeochemistry)` here right? @jagoosw is it true?

src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
```diff
@@ -75,9 +75,19 @@ Return a flattened `NamedTuple` of the fields in `model.velocities`, `model.free
...
+    field_names(model)
+
+Return a tuple of the field names associated with `HydrostaticFreeSurfaceModel`. 
+Equivalent to doing `keys(fields(model))`.
+"""
+@inline field_names(model::HydrostaticFreeSurfaceModel) = field_names(model.free_surface, tracernames(model.tracers), model.auxiliary_fields)
+
+@inline field_names(free_surface, tracernames, auxiliary_fields) = (:u, :v, :w, tracernames..., :η, keys(auxiliary_fields)...)
+@inline field_names(::Nothing, tracernames, auxiliary_fields)    = (:u, :v, :w, tracernames..., keys(auxiliary_fields)...)
+@inline field_names(::SplitExplicitFreeSurface, tracernames, auxiliary_fields) = (:u, :v, :w, tracernames..., :η, :U, :V, keys(auxiliary_fields)...)
```
glwagner, line 90: I think this is a bad idea. I think that `fields` and `prognostic_fields` should be the source of truth. Those functions return NamedTuple --- both names, and fields. Thus the names can be derived from `fields` or `prognostic_fields`.

glwagner:
> (2) a field_names function to be able to change the field names in a HydrostaticFreeSurfaceModel depending on different inputs

I think it is a bad design to have multiplicitous sources of truth. This PR does not solve the underlying issue, which is that there are two conflicting sources of truth. Instead, it creates a _third_ source of truth which is definitely not learning the lesson that the bug should have given.

glwagner (changes requested):
Get rid of `field_names`, and instead create a design where there is only one source of truth about the field names --- either from `fields` or `prognostic_fields`. We may not need `fields` independently from `prognostic_fields`.

src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
simone-silvestri: we need something to regularize the boundary conditions, so we need some names that correspond to `keys(fields(model))` before the model is built. I am open to a different implementation though.

glwagner: I see

glwagner:
Can you add a test that exposes the issue with `ContinuousBoundaryFunction` directly?

simone-silvestri:
done in a different way

src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
simone-silvestri: I have changed the design for the `fields(model)` to always output a predictable namedtuple with `nothing` in place of fields that don't exist. In this way we don't need a new function

src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
glwagner: no, the biogeochemistry struct is allowed to have extra auxiliary fields that are not attached to the model

glwagner:
I see the issue now that the source of truth was hardcoded int he constructor... hmm..

glwagner (approved):

test/test_biogeochemistry.jl
```diff
@@ -68,7 +68,7 @@ end
 const MB = Union{MinimalDiscreteBiogeochemistry, MinimalContinuousBiogeochemistry}
 
 @inline          required_biogeochemical_tracers(::MB) = tuple(:P)
-@inline required_biogeochemical_auxiliary_fields(::MB) = tuple(:Iᴾᴬᴿ)
+@inline required_biogeochemical_auxiliary_fields(::MB) = ()
```
glwagner, line 71: We want this for the test?

simone-silvestri: It seems that I^{PAR} is repeated in fields. So, indeed, it looks like the `bialogical_auxiliary_fields` are added to the model's auxiliary fields upon model construction here:
```julia
    biogeochemical_fields = merge(auxiliary_fields, biogeochemical_auxiliary_fields(biogeochemistry))
    tracers, auxiliary_fields = validate_biogeochemistry(tracers, biogeochemical_fields, biogeochemistry, grid, clock)
```
which leads to having repeated I^{PAR} here
```julia
@inline fields(model::HydrostaticFreeSurfaceModel) = 
    merge(hydrostatic_fields(model.velocities, model.free_surface, model.tracers), 
          model.auxiliary_fields,
          biogeochemical_auxiliary_fields(model.biogeochemistry))
```
So I think we have a decision to make
(1) leave the biogeochemical fields in the `model.biogeochemistry`
(2) copy them in `model.auxiliary_fields` but modify the `fields` function

src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
simone-silvestri: It actually looks like the fields are added to the `auxiliary_fields`, and, thus they are accounted for twice:
https://github.com/CliMA/Oceananigans.jl/blob/d8273e7a07e8f53f9772f87f67f6eca8d17cb98c/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl#L131-L133

test/test_biogeochemistry.jl
glwagner: `required_biogeochemical_auxiliary_fields` are the `model.auxiliary_fields`. Then there are also `biogeochemical_auxiliary_fields(model.biogeochemistry)`, which are distinct? So there are two lists of auxiliary fields.

We can change the design.

navidcy:
@simone-silvestri could you resolve the conflicts? and then should we merge this?

navidcy:
I [resoved][4853626](https://github.com/CliMA/Oceananigans.jl/pull/4008/commits/4853626c2725b4a05dfeddf743b9d4c1c617ce11)
 them.

simone-silvestri:
No please do not merge. I still need to verify that things work. I think there were some issues that I still need to check because we do not seem to have test coverage for some cases (which I intend to add)

glwagner (approved):

simone-silvestri:
I think this PR solves a bug that is there also with CPU boundary conditions, but it is blocked because I added an extra test on the GPU and because of #4165 the tests do not pass.

We should merge this one anyways, so I will mark the GPU test as test broken and merge

PR #4009, opened 2024-12-18
Lagrangian Smagorinsky for 2D flows

simone-silvestri:
At the moment, Lagrangian Smagorinsky does not work for 2D flows because nodes on flat topologies return `nothing`. 
It is a very easy correction to make it work.

However, I know that Lagrangian Smagorinsky is not really meant for 2D flows, so if we want to keep this incompatibility to prevent people from using it, it is also fine (I vote for correcting it anyways).

@tomchor please advise


tomchor:
I think we should keep it incompatible with anything but 3D flows. Not only it goes against LES assumptions, but it just doesn't work and the results look terrible (because they violate said assumptions). Personally, I don't see anything to gain from generalizing it.

simone-silvestri:
I would propose to fix the error and including a warning when the topology is Flat.
Users will try using it on flat domains regardless, so it is better to have an explicit warning (or error) instead of an obscure error deep in the weed of the kernel.

tomchor:
> I would propose to fix the error and including a warning (or error) when the topology is Flat. Users will try using it on flat domains regardless, so it is better to have an explicit warning (or error) instead of an obscure error deep in the weed of the kernel.

Agreed. Should we do the same for other LES closures?

simone-silvestri:
why not, what should we warn against? Probably `AnisotropicMinimumDissipation()` for 2D domains as well? I Think `TwoDimensionalLeith` is fine for both 2d and 2d domains

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -244,7 +247,7 @@ const c = Center()
         δx = u[i, j, k] * Δt
         δy = v[i, j, k] * Δt
         δz = w[i, j, k] * Δt
-
+
```
tomchor, line 250:
```suggestion

```

tomchor (approved):

PR #4011, opened 2024-12-18
Suppress "run forever warning" in Simulation constructor if verbose=false

glwagner:
cc @francispoulin


simone-silvestri (approved):

francispoulin (approved):
Looks good!  Thanks for doing this so quickly.

PR #4017, opened 2024-12-19
Correct `tupled_halo_filling` for distributed fields

simone-silvestri:
the code for `tupled_fill_halo_regions!` seems to have become a bit more complicated.

The distributed fill halo regions for CATKE is broken because non-fields are passed to the `tupled_fill_halo_regions!` function.
This PR fixes it and adds a test


glwagner (approved):
It's because our scheme for tupled halo filling was not type inferrable. There's still some work to do on this topic.

src/DistributedComputations/halo_communication.jl
```diff
@@ -80,7 +80,11 @@ end
 
 function tupled_fill_halo_regions!(fields, grid::DistributedGrid, args...; kwargs...)
     for field in fields
-        fill_halo_regions!(field, args...; kwargs...)
+        # Sometimes diffusivity_fields might not be Field objects
+        # and they are passed anyways to the fill halo regions routine
```
glwagner, line 84: What does mean: "they are passed anyways"

PR #4019, opened 2024-12-20
Make RK3 time-stepping compatible with CATKE

simone-silvestri:
This implementation may not be correct. 

The sub stepping scheme is in principle different than the baroclinic stepping scheme so I am not extremely sure this method works. I ll think a bit more about it.

We actually need to implement it like we did for the `SplitExplicitFreeSurface`, by saving down the slow tendencies, and averaging them in the last step. All stages should go from `n` to `n+1`.

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
@@ -194,13 +194,6 @@ function compute_diffusivities!(diffusivities, closure::FlavorOfCATKE, model; pa
     Δt = model.clock.time - diffusivities.previous_compute_time[]
     diffusivities.previous_compute_time[] = model.clock.time
 
-    if isfinite(model.clock.last_Δt) # Check that we have taken a valid time-step first.
-        # Compute e at the current time:
-        #   * update tendency Gⁿ using current and previous velocity field
-        #   * use tridiagonal solve to take an implicit step
-        time_step_catke_equation!(model)
-    end
-
     # Update "previous velocities"
     u, v, w = model.velocities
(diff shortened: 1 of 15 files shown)


simone-silvestri:
Sounds good. I have implemented this strategy for the moment (just revert to the standard trace time stepping if `tke_time_step == nothing), so we can test how CATKE performs with RK3. I suspect it will be better than AB2 and maybe we don't need substepping after all.

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
```diff
@@ -34,9 +34,14 @@ end
 CATKEVerticalDiffusivity(FT::DataType; kw...) =
     CATKEVerticalDiffusivity(VerticallyImplicitTimeDiscretization(), FT; kw...)
 
-const CATKEVD{TD} = CATKEVerticalDiffusivity{TD} where TD
-const CATKEVDArray{TD} = AbstractArray{<:CATKEVD{TD}} where TD
-const FlavorOfCATKE{TD} = Union{CATKEVD{TD}, CATKEVDArray{TD}} where TD
+const CATKEVD{TD} = CATKEVerticalDiffusivity{TD, <:Any, <:Any, DT} where {TD, DT}
+const CATKEVDArray{TD, DT} = AbstractArray{<:CATKEVD{TD, DT}} where {TD, DT}
+const FlavorOfCATKE{TD, DT} = Union{CATKEVD{TD, DT}, CATKEVDArray{TD, DT}} where TD, DT
```
glwagner: it's kind of a mess, is it worth it?

simone-silvestri:
Right. It seems a bit complicated to make it work like this. 
Probably we need a refactor and we need find an abstraction for the time stepping details of the TKE.

PR #4022, opened 2024-12-22
Add info for Julia version used for testing

navidcy:
PR stems from https://github.com/CliMA/Oceananigans.jl/commit/d8273e7a07e8f53f9772f87f67f6eca8d17cb98c#commitcomment-150654994


docs/src/index.md
```diff
@@ -24,8 +24,11 @@ julia> using Pkg
 julia> Pkg.add("Oceananigans")
 ```
 
-!!! compat "Julia 1.9 or later is required; Julia 1.10 is suggested"
-    Oceananigans requires Julia 1.9 or later. We strongly urge you to use Julia 1.10.
+!!! compat "Julia 1.9 or later is required"
+    Oceananigans requires Julia 1.9 or later.
+    
+!!! info "Continuous integration testing"
```
glwagner, line 30:
```suggestion
!!! info "Tested Julia versions"
```

docs/src/index.md
```diff
@@ -24,8 +24,11 @@ julia> using Pkg
...
+    Oceananigans is tested on Julia 1.10.
```
glwagner, line 31:
```suggestion
    Oceananigans is currently tested on Julia 1.10.
```

navidcy: What does "currently" add here? The docs are associated with package versions; later on for version 0.423.2 we may change this to Julia 1.12. Right?

glwagner: My intent is that if someone notices it is wrong it gives them a clue why and they can submit a PR to fix it

simone-silvestri (approved):

PR #4026, opened 2024-12-28
Enzyme: mark prettytime as non-differentiable

wsmoses:


glwagner (approved):

wsmoses:
@glwagner the failure seems unrelated if it can be rerun

PR #4027, opened 2025-01-03
Add unit tests for background field flux divergence

liuchihl:
This PR adds a unit test for https://github.com/CliMA/Oceananigans.jl/pull/3646 

The tests compare model results with and without background fields using a linear stratification profile.


glwagner:
Why does this PR change so many files? Do either the target branch or this source branch need updating?

liuchihl:
Yeah, I only meant to add the unit test. Since you have already updated the target branch, I will update the source branch.

PR #4029, opened 2025-01-05
Allows properties of `StaticVerticalCoordinate` to have different types

wsmoses:
Required for reactant on GPU


simone-silvestri:
in theory `coord.cᵃᵃᶠ` and `coord.cᵃᵃᶜ` should always have the same type and similarly for `coord.Δᵃᵃᶠ` and `coord.Δᵃᵃᶜ`. 
When is this not the case?

If this is not the case on the GPU wouldn't it be better to add other type parameters to the struct from the get go?

wsmoses:
Also happy to add type parameters if that's the desired resolution

simone-silvestri:
Hmmm, ok I see. 
We use this pattern also in grids though, for example here
https://github.com/CliMA/Oceananigans.jl/blob/aa5f2393b2bd82b8bc13e217e9dd62590c9663e4/src/Grids/rectilinear_grid.jl#L14-L21
So probably that should also be corrected?

glwagner:
I think it's fine to add type parameters

wsmoses:
okay added the type parameter version

glwagner (approved):

wsmoses:
oh man CI finally is happy, @glwagner can we merge?

simone-silvestri:
tests pass, looks good to merge.

Issue #2996, opened 2023-03-22
10-100x slowdown on CPU after upgrade to KernelAbstractions 0.8 (due to type inference failure?)


navidcy:
I observe a discontinuous doubling of docs building time from 3 hours on

https://buildkite.com/clima/oceananigans/builds/9920#01862a69-61ee-4e59-82b4-5c4b4684e7b6

to 6 hours after "Merge pull request #2899 from CliMA/vc/ka_upgrade"

https://buildkite.com/clima/oceananigans/builds/9935#01862d0c-e009-4a05-a8bb-023d1159a2ae

cc @simone-silvestri, @vchuravy

glwagner:
I'm detecting also a massive slow down in model execution and out of control allocations on the CPU. Type inference failure? Unfortunate to find it this late after the PRs are merged.

glwagner:
@navidcy what's the last version before the catastrophic performance loss? I'll do a benchmark to compare with `main`.

glwagner:
Kind of a random case, but here's some timings from a calibration problem I'm doing (I'm running 6 3D simulations, ranging from something like 33 to 117 time-steps, either size (6, 10, 32) or (6, 10, 64)). For these tests I just downgraded KernelAbstractions via Project.toml:

## With `KernelAbstractions` 0.7.2 (also downgraded CUDAKernels)

```
 23.405205 seconds (25.27 M allocations: 2.610 GiB, 1.78% gc time, 99.80% compilation time)
  5.019944 seconds (5.92 M allocations: 475.126 MiB, 0.86% gc time, 98.22% compilation time)
  0.067385 seconds (107.25 k allocations: 72.628 MiB)
  0.090308 seconds (107.25 k allocations: 72.628 MiB)
  0.139109 seconds (217.20 k allocations: 147.487 MiB)
  0.197798 seconds (217.20 k allocations: 147.487 MiB)
```

The two simulations are affected by compilation but things go fast after that.

## With `KernelAbstractions` 0.8.6

```
  4.914645 seconds (28.10 M allocations: 6.039 GiB, 15.62% gc time, 51.22% compilation time)
  5.011717 seconds (31.58 M allocations: 10.844 GiB, 19.87% gc time, 16.87% compilation time)
  4.236418 seconds (27.41 M allocations: 11.073 GiB, 21.75% gc time)
  8.501561 seconds (55.04 M allocations: 22.118 GiB, 22.03% gc time)
[...]

PR #4032, opened 2025-01-07
Inference problem on the CPU

simone-silvestri:
Working with @milankl I noticed that a stripped-down version of ClimaOcean (with only the `HydrostaticFreeSurfaceModel`) was terribly slow compared to a SpeedyWeather simulation with the same gridsize.

After some investigating I found out that the problem was the `forcing` added as an argument to the momentum kernels after the splatting of the other `args...`.
I am not sure what caused this loss of performance, but the benchmark of the code:
```julia
using ClimaOcean
using Oceananigans
using Oceananigans.Units
using BenchmarkTools

arch    = Oceananigans.CPU()
r_faces = ClimaOcean.exponential_z_faces(; Nz = 8, depth = 5000)

grid = LatitudeLongitudeGrid(arch, 
                             size=(360, 152, 8), 
                             latitude=(-75.5, 75.5), 
                             longitude=(0, 360),
                             halo = (6, 6, 2),
                             z = r_faces)

ocean = ClimaOcean.ocean_simulation(grid; 
                                    closure=nothing,
                                    momentum_advection=VectorInvariant(),
                                    tracer_advection=Centered(),
                                    free_surface=SplitExplicitFreeSurface(grid; substeps=20))

time_step!(ocean)
time_step!(ocean)

@benchmark time_step!(ocean)
```
on main:
Screenshot 2025-01-07 at 4 01 16 PM
and on this branch with the correction:
Screenshot 2025-01-07 at 4 11 35 PM

This should not impact GPU execution.


glwagner:
Oh. I think this is basically #2996 coming back. This keeps happening so I wonder if there is a way to test it.

glwagner (approved):

simone-silvestri:
We might need to add some performance tests. Possibly very simple ones, for example tests that check that there is no extra allocation that comes from types not being inferred correctly.

PR #4035, opened 2025-01-10
Generalize ntuple type constraints to be tuple{vararg}

wsmoses:
Like in https://github.com/CliMA/Oceananigans.jl/pull/4029

one might now have a tuple like `::Tuple{OffsetArrays.OffsetArray{Float64, 3, ReactantCUDAExt.CuTracedArray{Float64, 3, 1, (35, 34, 16)}}, OffsetArrays.OffsetArray{Float64, 3, ReactantCUDAExt.CuTracedArray{Float64, 3, 1, (34, 35, 16)}}})` which is not an `ntuple{<:Any, offsetarray}` (since they aren't the same underlying type) but is a `tuple{vararg{offsetarray}}` (since this is a tuple of all offsetarrays)

### DIFF
--- a/src/BoundaryConditions/fill_halo_regions.jl
+++ b/src/BoundaryConditions/fill_halo_regions.jl
@@ -44,7 +44,7 @@ end
 @inline extract_bc(bc, ::Val{:bottom_and_top})  = (extract_bottom_bc(bc), extract_top_bc(bc))
 
 # Finally, the true fill_halo!
-const MaybeTupledData = Union{OffsetArray, NTuple{<:Any, OffsetArray}}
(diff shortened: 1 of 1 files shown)


glwagner:
This seems fine to me --- it'd be good to get input from @simone-silvestri since I don't know exactly why the halo filling dispatched on `NTuple` instead of `Tuple`.

simone-silvestri:
I think the `NTuple` are remainders from when we had a different syntax. If I remember correctly we had some issues with inference on GPU that required extracting the tuple size `N` from `NTuple` so that the length of the tuple was known by the compiler. However, the problem was eventually solved using `ntuple() do n`. 

If the tests pass also on GPU then everything should be correct.

simone-silvestri (approved):

Issue #4036, opened 2025-01-10
Scalar indexing with CUDA 5.6.0


simone-silvestri:
Reductions on Oceananigans seem to not work on GPUs using CUDA 5.6.0 which switched to GPUArrays 11.2.0. 
It looks like scalar indexing errors are now popping up.  Seems to be connected with the fact that GPUarrays switched to a KernelAbstraction backend. I ll investigate a bit.

MWE:
```julia
using Oceananigans
grid = RectilinearGrid(GPU(), size=(1, 1, 1), extent=(1, 1, 1))
model = NonhydrostaticModel(; grid)
maximum(abs, model.velocities.u)
```

simone-silvestri:
the issue seems to be 
```
 [16] mapfirst!(f::typeof(identity), R::SubArray{…}, A::Oceananigans.AbstractOperations.ConditionalOperation{…})
```
in particular, just
```julia
maximum(model.velocities.u)
```
works properly

simone-silvestri:
Actually this is easily explainable because `sum` and `prod` use different initialization for reductions
https://github.com/CliMA/Oceananigans.jl/blob/19baa398e793b64ca182ac6c1de5599034f188fa/src/Fields/field.jl#L611-L616

simone-silvestri:
Seems to be connected to the fact that `Base.mapfirst!(f, interior(r), interior(c))` with `c` an `AbstractOperation` now wants to run on a CPU instead than on a GPU, another clue:
```julia
julia> v = model.velocities.u + 5

julia> maximum(v)
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.
```

glwagner:
See what ,ethod gets called with the previous CUDA and also with the updated one. interior(r) is SubArray of CuArray?

simone-silvestri:
For documentation purposes, with @glwagner we found out that the issue is that the new GPUArrays dispatches `map!` onto `AnyGPUArray` also for the destination, defined here
https://github.com/JuliaGPU/GPUArrays.jl/blob/78c9ef079c28cfd4ecb7ce718da4c7490dc12404/src/host/broadcast.jl#L106-L130
while before the destination was an `AbstractArray`
[...]

PR #4037, opened 2025-01-10
(0.95.6) Pin GPUArrays version + fix GPU testing pipeline

simone-silvestri:
a lot of Oceananigans functionalities are broken at the moment because of a change in `GPUArrays` (see issue #4036)
This PR pins to a previous `GPUArrays` version to keep the repository (and the CI) functioning. Once the issue has been resolved we can remove this constraint.


navidcy (approved):

glwagner (approved):

glwagner:
`sverdrup` is down right now so the GPU tests aren't running

navidcy:
Does this closes #4036?

simone-silvestri:
> Does this closes #4036?

Not really, because we want to be able to use the new `GPUArrays` version. This is just patching the problem by pinning the GPUArrays version. I would keep the issue open until we can update CUDA and GPUArrays

PR #4039, opened 2025-01-12
Remove global `allowscalar` from testing suite attempt 2

### LINKED ISSUE #3039: Do we allow scalar indexing in all our tests?
https://github.com/CliMA/Oceananigans.jl/blob/bcc34f07b3f949ea6fb34c7814f4b856d24924c2/test/runtests.jl#L3

We should remove this

simone-silvestri:
Because it does not allow us to catch problems where kernels erroneously run on the CPU (for example #4036), see:
 >             > and none of the tests of Oceananigans caught this?
>
> OK, none of Oceananigans tests caught this because they are all run within a `CUDA.allowscalar()`!
>
> https://github.com/CliMA/Oceananigans.jl/blob/e02564df8ab90fb8f994fddf465b10d123fdfc3a/test/runtests.jl#L19
>
> We should remove this `CUDA.allowscalar() do ... all-the-tests ... end`
>
> cc @glwagner

_Originally posted by @navidcy in https://github.com/CliMA/ClimaOcean.jl/issues/318#issuecomment-2585887372_
            
Probably will require a bit of fiddling to add `allowscalar` granularly in tests

closes #3039

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -16,8 +16,6 @@ end
 ##### Run tests!
 #####
 
-CUDA.allowscalar() do
-
 @testset "Oceananigans" begin
 
     if test_file != :none
@@ -230,5 +228,3 @@ CUDA.allowscalar() do
         include("test_convergence.jl")
     end
 end
-
-end #CUDA.allowscalar()


navidcy:
I'm so much in favor!

I did try to do it one other time long ago and gave up... :( 

I'll help out!

navidcy:
So the output writer [was failing with allow scalar operations](https://buildkite.com/clima/oceananigans/builds/20172#0194673f-fd1e-4867-9f99-12c0820baab6/17-1250) and I added a `CUDA.@allowscalar` at the `Base.getindex(f::Field, inds...)` method; see [1fe5fb6](https://github.com/CliMA/Oceananigans.jl/pull/4039/commits/1fe5fb62388698c291175230758b2765b9ce2216).

Is this bad idea?

simone-silvestri:
I think so, this would defeat the purpose of this PR because it would not allow to catch scalar indexing issues with fields

navidcy:
> I think so, this would defeat the purpose of this PR because it would not allow to catch scalar indexing issues with fields

OK. But we need to do something because [fetch_output was hitting a wall](https://buildkite.com/clima/oceananigans/builds/20172#0194673f-fd1e-4867-9f99-12c0820baab6/17-1250)

navidcy:
What do we do with this: https://buildkite.com/clima/oceananigans/builds/20198#019476ab-1f43-478a-a915-a760bb94157e/31-1301 ?

simone-silvestri:
looks like a bug in `convert_output`. We should be passing `parent(field)` to it, not a field, but from the error we are trying to 
```julia
copyto!(dst::Array{}, scr::Field{})
```

simone-silvestri:
> What do we do with this: https://buildkite.com/clima/oceananigans/builds/20198#019476ab-1f43-478a-a915-a760bb94157e/31-1301 ?

The problem seems to arise when trying to write down a `FunctionField`. this is because we have 
```julia
function fetch_output(field::AbstractField, model)
    return parent(field)
end
```
that returns the same `FunctionField` which then fails in 
```julia
function convert_output(output::AbstractArray, writer)
    if architecture(output) isa GPU
        output_array = writer.array_type(undef, size(output)...)
        copyto!(output_array, output)
    else
        output_array = convert(writer.array_type, output)
    end

    return output_array
end
```
I have added a method that converts the function field to CPU before writing it down.

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,5 @@
 env:
-  JULIA_VERSION: "1.10.6"
+  JULIA_VERSION: "1.10.8"
```
glwagner, line 2: probably better to put this in another PR to get it in sooner

glwagner:
@simone-silvestri are you still pursuing this?

simone-silvestri:
We can probably close this because it is stale. We can maybe pursue this in a fresh PR

Issue #3935, opened 2024-11-18
`FieldTimeSeries` support for `NetCDF`


tomchor:
@ali-ramadhan and I are interesting in expanding `FieldTimeSeries` to support `NetCDF`.

None of us is super familiar with FTS, but it seems like, for the most part, all we need would be to figure out a way to reconstruct `grid`s based on `NetCDF` output and figure out how to deal with memory access if the data is `PartlyInMemory`. Is that correct or are there other items that we pay attention to?

For the former we can output the relevant grid parameters to file and reconstruct it in a way that's already done for JLD2 whenever the `grid` can't be restored directly: https://github.com/CliMA/Oceananigans.jl/blob/82ad840f2b7e03d9b65cf004a216da76a9ec173d/src/OutputReaders/field_time_series.jl#L489-L529

We were also wondering if this is worth doing over one PR and a few small ones.

cc @glwagner

glwagner:
For a very first PR, it is only necessary to figure out how to reconstruct the `grid`. We need the `grid` in order to do any meaningful post-processing, such as computing derivatives, integrals, etc, with the saved data.

I would focus completely on one backend --- probably `InMemory`. We may want to refactor the backends a little bit. For example, partly in memory should be support automatically if we do this right.

ali-ramadhan:
Maybe a good starting point would be to add enough functionality so that a `RectilinearGrid` can be reconstructed? And tests for a the reconstruction of a simple `FieldTimeSeries`. This could be checked pretty thoroughly with some unit tests.

> Small PRs are always better because they lead to better design and easier review.

Definitely agree! I might instead say that a PR should have one clear purpose. Perhaps the narrower the better. I suppose even narrowly-scoped PRs can grow quite large though.

glwagner:
> Maybe a good starting point would be to add enough functionality so that a RectilinearGrid can be reconstructed?

That works! But don't reinvent the wheel or implement something too specific. I would start by looking at the existing functionality for reconstructing grids eg

https://github.com/CliMA/Oceananigans.jl/blob/82ad840f2b7e03d9b65cf004a216da76a9ec173d/src/Grids/rectilinear_grid.jl#L374-L394

I think you have to save the topology and location as strings. Does NetCDF support tuples?

glwagner:
You could have `constructor_arguments(filename::String)` and then something like

```julia
function reconstruct_grid(filename)
    args, kwargs = constructor_arguments(filename)
[...]

PR #4046, opened 2025-01-16
(0.96.0) Improve the `NetCDFOutputWriter` experience

### LINKED ISSUE #1334: Integrating Oceananigans with Python-xgcm
I was thinking that it would be a good idea to make an integration with Python's [`xgcm`](https://github.com/xgcm/xgcm) package easier. The project is relatively new but I expect it to grow given that it has [Pangeo](https://pangeo.io/) support.

In my mind, this is related to #1313, since the primary (only?) communication with Python is done through NetCDF files. So including some [grid metrics](https://xgcm.readthedocs.io/en/latest/grid_metrics.html) in the NetCDF output would be helpful. For now I'm not exactly sure what needs to be there since their docs [aren't very explanatory in that sense](https://github.com/xgcm/xgcm/issues/291), but it includes distances between `Face`s and `Center`s of the grids and other measures that should be straightforward.

Any thoughts?

### LINKED ISSUE #2248: Support curvilinear grids with `NetCDFOutputWriter`?
Right now it looks like we're assuming rectilinear grids, eg:

https://github.com/CliMA/Oceananigans.jl/blob/adffd0cb67373da364a54d7f300d1f33817539bd/src/OutputWriters/netcdf_output_writer.jl#L31-L43

and

https://github.com/CliMA/Oceananigans.jl/blob/adffd0cb67373da364a54d7f300d1f33817539bd/src/OutputWriters/netcdf_output_writer.jl#L28-L29

For rectilinear grids there are just six node ranges (2 locations for each dimension); however on curvilinear grids there are 24 (8 locations for each dimension).

One possibility is to attempt to infer the dimensions of the output and store verbose names like `"xᶠᶠᶜ"`. We can also store the six defaults that we have currently but with new verbose names `"xᶠᶜᶜ"`. Note that in a latitude-longitude coordinate system, `x` becomes `λ`, etc.

I think it's also nice if the grid can be easily reconstructed from the file content. We would _need_ this if we ever support `FieldTimeSeries` with `NetCDFOutputWriter`.

I don't know if it's relevant, but whatever solutions we invent we may want to think a bit about how we are going to make maintaining output writers easier in the future. Because we may have to support even more output formats (like [Zarr](https://pangeo.io/data.html)).

Curious to hear from the NetCDF users out there like @tomchor, @francispoulin ...

### LINKED ISSUE #2770: Output constructor does not work with sliced fields
Hi all,

I've come across a problem a few times with the output writer where it can't set up running with `indices`specified. This presents as a bounds error coming from the `offset_data` function:
https://github.com/CliMA/Oceananigans.jl/blob/c73ef035702b33b8d11c505365a4dd48444444a7/src/Grids/new_data.jl#L33-L48

An example of this occurs when you try to run the "near_global_lat_lon" example when it tries to output surface values:
https://github.com/CliMA/Oceananigans.jl/blob/c73ef035702b33b8d11c505365a4dd48444444a7/validation/near_global_lat_lon/near_global_quarter_degree.jl#L296-L301

Essentially what I think is happening is that when we output data and specify some slice that isn't the dimensions of the grid, in this example just the surface level `grid.Nz`, the data is being correctly (? not sure we actually want the halo here) selected as e.g. `-Hx:Nx+Hx` in dimensions specified as `:` but just whatever indices specified by the user in the other dimensions. In the `offset_data` function I think `ii` gets the halo added in all dimensions so there's a miss match. 

I'm not sure how to fix this cleanly?

### LINKED ISSUE #3997: Helper for outputting a NetCDF "grid file"
After learning a bit about how people work with NetCDF (😺 ) and discussing with @noraloose it seems there may be scope for a helper function that writes out a grid file. Something like:

```julia
write_grid_to_netcdf!(filename, grid)
```

Also internally, this functionality could be added to `NetCDFOutputWriter`, so that users could write something like

```julia
ow = NetCDFOutputWriter(model, grid; filename, with_grid_metrics=true)
```

and include all the same grid metrics that are saved when using `write_grid_to_netcdf!`.

Under the hood, I think this is accomplished by creating a special `NetCDFOutputWriter` and treating the grid metrics like output variables. One possible difference is that unlike other kinds of output the grid metrics do not have a time dimension. So we also need to generalize the scope of outputs for `NetCDFOutputWriter` to outputs that don't have a time dimension? My NetCDF knowledge is a little hazy so please let me know if this isn't right.

@ali-ramadhan @tomchor

ali-ramadhan:
This PR updates the `NetCDFOutputWriter` to:
* Properly work on `RectilinearGrid` and `LatitudeLongitudeGrid` (with correct and useful output attributes).
* Save grid metrics with useful attributes.
* Save immersed boundary information.
* Properly support flat grids.
* Work cleanly with `LagrangianParticles` output.
* Allow for flexible dimension naming to satisfy desired naming conventions.
* Write grid reconstruction metadata into a NetCDF group to support `FieldTimeSeries` construction from NetCDF (full support to be added in a subsequent PR).

Tests have been added for all these features. Thank you to @tomchor and @almacarolina for helpful discussions during this refactor!

There's still a little bit left to do, but if anyone has any thoughts or feedback and would like to leave a review I'd appreciate it! There are a lot of line additions but thankfully the changes are almost fully isolated to just two files.

TODO:
- [x] Output grid areas and volumes too? But maybe it would be nice to first have `xareas`, `volumes`, etc. via `KernelFunctionOperation`?
- [x] Further improve the docstring and documentation for `NetCDFOutputWriter` to highlight all features.
- [x] Use `NetCDFOutputWriter` + some fancier features in an example.
- [x] Figure out which tests are missing.
- [x] Watch out for failing tests/examples and fix them.

Some comments:
* I'm hoping to merge this PR and immediately work on adding support for NetCDF-backed `FieldTimeSeries`. I was planning on working on it here (as the branch name suggests) but it will involve refactoring `field_time_series.jl` which seems best suited for a separate PR.
* Supporting output of the free surface height is a bit hacky. Since it's a regular `Field` but is kind of a `ReducedField` we cannot dispatch on its type and work with it correctly. Will open an issue to discuss.
* I'm tagging v0.96.0 since this PR significantly changes the dimension names used in NetCDF files produced by `NetCDFOutputWriter`.

Resolves #1334 (via the dimension name generator)
Resolves #2248
Resolves #2770 (Maybe? You can save u, v, w, T, S, and η in the same NetCDF file but it's a bit hacky, see above.)
Resolves #3997

This PR makes progress on issue #3935
This PR supercedes PR #2652


ali-ramadhan:
Here's what a NetCDF file looks like now in NCDatasets.jl and xarray to give an idea of the changes introduced by this PR.

NCDatasets.jl:
```
julia> ds = NCDataset("test/test_immersed_grid_latlon_no_halos_GPU.nc")
Dataset: test/test_immersed_grid_latlon_no_halos_GPU.nc
Group: /

Dimensions
   time = 6
   z_c = 16
   z_f = 17
   latitude_c = 16
   latitude_f = 17
   longitude_f = 17
   longitude_c = 16

Variables
  time   (6)
    Datatype:    Float64 (Float64)
    Dimensions:  time
    Attributes:
     units                = seconds
     long_name            = Time

  z_c   (16)
    Datatype:    Float32 (Float32)
    Dimensions:  z_c
    Attributes:
     units                = m
     long_name            = Locations of the cell centers in the z-direction.

  z_f   (17)
    Datatype:    Float32 (Float32)
    Dimensions:  z_f
    Attributes:
     units                = m
     long_name            = Locations of the cell faces in the z-direction.

  latitude_c   (16)
    Datatype:    Float32 (Float32)
    Dimensions:  latitude_c
    Attributes:
     units                = degrees north
     long_name            = Locations of the cell centers in the meridional direction.

  latitude_f   (17)
    Datatype:    Float32 (Float32)
    Dimensions:  latitude_f
    Attributes:
     units                = degrees north
     long_name            = Locations of the cell faces in the meridional direction.

  longitude_f   (17)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_f
    Attributes:
     units                = degrees east
     long_name            = Locations of the cell faces in the zonal direction.

  longitude_c   (16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c
    Attributes:
     units                = degrees east
     long_name            = Locations of the cell centers in the zonal direction.

  bottom_height   (16 × 16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_c

  dlatitude_c   (16)
    Datatype:    Float32 (Float32)
    Dimensions:  latitude_c
    Attributes:
     units                = degrees
     long_name            = Angular spacings between the cell centers in the meridional direction.

  dlatitude_f   (17)
    Datatype:    Float32 (Float32)
    Dimensions:  latitude_f
    Attributes:
     units                = degrees
     long_name            = Angular spacings between the cell faces in the meridional direction.

  dlongitude_c   (16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c
    Attributes:
     units                = degrees
     long_name            = Angular spacings between the cell centers in the zonal direction.

  dlongitude_f   (17)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_f
    Attributes:
     units                = degrees
     long_name            = Angular spacings between the cell faces in the zonal direction.

  dx_cc   (16 × 16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_c
    Attributes:
     units                = m
     long_name            = Geodesic spacings in the zonal direction between the cell located at (Center, Center).

  dx_cf   (16 × 17)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_f
    Attributes:
     units                = m
     long_name            = Geodesic spacings in the zonal direction between the cell located at (Center, Face).

  dx_fc   (17 × 16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_f × latitude_c
    Attributes:
     units                = m
     long_name            = Geodesic spacings in the zonal direction between the cell located at (Face, Center).

  dx_ff   (17 × 17)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_f × latitude_f
    Attributes:
     units                = m
     long_name            = Geodesic spacings in the zonal direction between the cell located at (Face, Face).

  dy_cc   (16 × 16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_c
    Attributes:
     units                = m
     long_name            = Geodesic spacings in the meridional direction between the cell located at (Center, Center).

  dy_cf   (16 × 17)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_f
    Attributes:
     units                = m
     long_name            = Geodesic spacings in the meridional direction between the cell located at (Center, Face).

  dy_fc   (17 × 16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_f × latitude_c
    Attributes:
     units                = m
     long_name            = Geodesic spacings in the meridional direction between the cell located at (Face, Center).

  dy_ff   (17 × 17)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_f × latitude_f
    Attributes:
     units                = m
     long_name            = Geodesic spacings in the meridional direction between the cell located at (Face, Face).

  dz_c   (16)
    Datatype:    Float32 (Float32)
    Dimensions:  z_c
    Attributes:
     units                = m
     long_name            = Spacings between the cell faces (located at the cell centers) in the z-direction.

  dz_f   (17)
    Datatype:    Float32 (Float32)
    Dimensions:  z_f
    Attributes:
     units                = m
     long_name            = Spacings between the cell centers (located at the cell faces) in the z-direction.

  immersed_boundary_mask_ccc   (16 × 16 × 16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_c × z_c

  immersed_boundary_mask_ccf   (16 × 16 × 17)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_c × z_f

  immersed_boundary_mask_cfc   (16 × 17 × 16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_f × z_c

  immersed_boundary_mask_fcc   (17 × 16 × 16)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_f × latitude_c × z_c

  S   (16 × 16 × 16 × 6)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_c × z_c × time
    Attributes:
     units                = practical salinity unit (psu)
     long_name            = Salinity

  T   (16 × 16 × 16 × 6)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_c × z_c × time
    Attributes:
     units                = °C
     long_name            = Temperature

  u   (17 × 16 × 16 × 6)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_f × latitude_c × z_c × time
    Attributes:
     units                = m/s
     long_name            = Velocity in the zonal direction (+ = east).

  v   (16 × 17 × 16 × 6)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_f × z_c × time
    Attributes:
     units                = m/s
     long_name            = Velocity in the meridional direction (+ = north).

  w   (16 × 16 × 17 × 6)
    Datatype:    Float32 (Float32)
    Dimensions:  longitude_c × latitude_c × z_f × time
    Attributes:
     units                = m/s
     long_name            = Velocity in the vertical direction (+ = up).

Global attributes
  Julia                = This file was generated using Julia Version 1.10.7
Commit 4976d05258e (2024-11-26 15:57 UTC)
Build Info:
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 48 × AMD Ryzen Threadripper 7960X 24-Cores
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, znver3)
Threads: 1 default, 0 interactive, 1 GC (on 48 virtual cores)
Environment:
  JULIA_TEST_FAILFAST = true
  LD_PRELOAD = /usr/NX/lib/libnxegl.so
  JULIA_LOAD_PATH = @:/tmp/jl_JcvMoH
  GPU: NVIDIA GeForce RTX 4090

  Oceananigans         = This file was generated using Oceananigans v0.96.0
  date                 = This file was generated on 2025-01-15T17:53:38.863 local time (2025-01-16T00:53:38.863 UTC).
  interval             = 1
  output iteration interval = Output was saved every 1 iteration(s).
  schedule             = IterationInterval
Groups
  Dataset: test/test_immersed_grid_latlon_no_halos_GPU.nc
  Group: grid_reconstruction

  Dimensions
     φ_f = 17
     λ_f = 17
     z_f = 17

  Variables
    φ_f     (17)
      Datatype:    Float32 (Float32)
      Dimensions:  φ_f

    λ_f     (17)
      Datatype:    Float32 (Float32)
      Dimensions:  λ_f

    z_f     (17)
      Datatype:    Float32 (Float32)
      Dimensions:  z_f

  Global attributes
    Hx                   = 2
    Hy                   = 3
    Hz                   = 4
    Nx                   = 16
    Ny                   = 16
    Nz                   = 16
    TX                   = Bounded
    TY                   = Bounded
    TZ                   = Bounded
    eltype               = Float64
    immersed_boundary_type = GridFittedBottom
    type                 = LatitudeLongitudeGrid
    z_spacing            = regular
    λ_spacing            = regular
    φ_spacing            = regular
```

xarray:
```
In [2]: ds = xr.open_dataset("test/test_immersed_grid_latlon_no_halos_GPU.nc")

In [3]: ds
Out[3]: 
<xarray.Dataset> Size: 589kB
Dimensions:                     (time: 6, z_c: 16, z_f: 17, latitude_c: 16,
                                 latitude_f: 17, longitude_f: 17,
                                 longitude_c: 16)
Coordinates:
  * time                        (time) timedelta64[ns] 48B 00:00:00 ... 00:00...
  * z_c                         (z_c) float32 64B -968.8 -906.2 ... -31.25
  * z_f                         (z_f) float32 68B -1e+03 -937.5 ... -62.5 0.0
  * latitude_c                  (latitude_c) float32 64B -9.375 -8.125 ... 9.375
  * latitude_f                  (latitude_f) float32 68B -10.0 -8.75 ... 10.0
  * longitude_f                 (longitude_f) float32 68B -20.0 -17.5 ... 20.0
  * longitude_c                 (longitude_c) float32 64B -18.75 ... 18.75
Data variables: (12/24)
    bottom_height               (latitude_c, longitude_c) float32 1kB ...
    dlatitude_c                 (latitude_c) float32 64B ...
    dlatitude_f                 (latitude_f) float32 68B ...
    dlongitude_c                (longitude_c) float32 64B ...
    dlongitude_f                (longitude_f) float32 68B ...
    dx_cc                       (latitude_c, longitude_c) float32 1kB ...
    ...                          ...
    immersed_boundary_mask_fcc  (z_c, latitude_c, longitude_f) float32 17kB ...
    S                           (time, z_c, latitude_c, longitude_c) float32 98kB ...
    T                           (time, z_c, latitude_c, longitude_c) float32 98kB ...
    u                           (time, z_c, latitude_c, longitude_f) float32 104kB ...
    v                           (time, z_c, latitude_f, longitude_c) float32 104kB ...
    w                           (time, z_f, latitude_c, longitude_c) float32 104kB ...
Attributes:
    Julia:                      This file was generated using Julia Version 1...
    Oceananigans:               This file was generated using Oceananigans v0...
    date:                       This file was generated on 2025-01-15T17:53:3...
    interval:                   1
    output iteration interval:  Output was saved every 1 iteration(s).
    schedule:                   IterationInterval

In [4]: ds.longitude_f
Out[4]: 
<xarray.DataArray 'longitude_f' (longitude_f: 17)> Size: 68B
array([-20. , -17.5, -15. , -12.5, -10. ,  -7.5,  -5. ,  -2.5,   0. ,   2.5,
         5. ,   7.5,  10. ,  12.5,  15. ,  17.5,  20. ], dtype=float32)
Coordinates:
  * longitude_f  (longitude_f) float32 68B -20.0 -17.5 -15.0 ... 15.0 17.5 20.0
Attributes:
    units:      degrees east
    long_name:  Locations of the cell faces in the zonal direction.

In [5]: ds.u
Out[5]: 
<xarray.DataArray 'u' (time: 6, z_c: 16, latitude_c: 16, longitude_f: 17)> Size: 104kB
[26112 values with dtype=float32]
Coordinates:
  * time         (time) timedelta64[ns] 48B 00:00:00 ... 00:00:00.500000
  * z_c          (z_c) float32 64B -968.8 -906.2 -843.8 ... -156.2 -93.75 -31.25
  * latitude_c   (latitude_c) float32 64B -9.375 -8.125 -6.875 ... 8.125 9.375
  * longitude_f  (longitude_f) float32 68B -20.0 -17.5 -15.0 ... 15.0 17.5 20.0
Attributes:
    units:      m/s
    long_name:  Velocity in the zonal direction (+ = east).
```

Hmmm, looks like I forgot to save some attributes for immersed boundary related variables.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -1,135 +1,716 @@
...
+#####
+
+function gather_grid_metrics(grid::RectilinearGrid, indices, dim_name_generator)
+    TX, TY, TZ = topology(grid)
+
+    metrics = Dict()
+
+    if TX != Flat
+        Δxᶠᵃᵃ_name = dim_name_generator("dx", grid, f, nothing, nothing, Val(:x))
+        Δxᶜᵃᵃ_name = dim_name_generator("dx", grid, c, nothing, nothing, Val(:x))
```
tomchor, line 205: It seems that `dx` is hardcoded here. Should we allow for more flexibility here too since this is one of the motivations behind the PR?

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -1,135 +1,716 @@
...
+    op_mask_ccf = KernelFunctionOperation{Center, Center, Face}(peripheral_node, grid, Center(), Center(), Face())
+
+    return Dict(
+        "bottom_height" => Field(grid.immersed_boundary.bottom_height; indices),
+        "immersed_boundary_mask_ccc" => Field(op_mask_ccc; indices),
+        "immersed_boundary_mask_fcc" => Field(op_mask_fcc; indices),
+        "immersed_boundary_mask_cfc" => Field(op_mask_cfc; indices),
+        "immersed_boundary_mask_ccf" => Field(op_mask_ccf; indices)
+    )
+end
```
tomchor, line 334: This function is assuming a `GridFittedBottom`, but it would be nice to extend it to `GridFittedBoundary` (the only difference is that `GridFittedBoundary` doesn't necessary have a bottom height).

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -1,135 +1,716 @@
...
+function default_dimension_attributes(grid::RectilinearGrid, dim_name_generator)
+    xᶠᵃᵃ_name = dim_name_generator("x", grid, f, nothing, nothing, Val(:x))
+    xᶜᵃᵃ_name = dim_name_generator("x", grid, c, nothing, nothing, Val(:x))
+    yᵃᶠᵃ_name = dim_name_generator("y", grid, nothing, f, nothing, Val(:y))
+    yᵃᶜᵃ_name = dim_name_generator("y", grid, nothing, c, nothing, Val(:y))
+
+    Δxᶠᵃᵃ_name = dim_name_generator("dx", grid, f, nothing, nothing, Val(:x))
+    Δxᶜᵃᵃ_name = dim_name_generator("dx", grid, c, nothing, nothing, Val(:x))
+    Δyᵃᶠᵃ_name = dim_name_generator("dy", grid, nothing, f, nothing, Val(:y))
+    Δyᵃᶜᵃ_name = dim_name_generator("dy", grid, nothing, c, nothing, Val(:y))
```
tomchor, line 413: Same comment as before. Should we avoid hardcoding these?

src/OutputWriters/netcdf_output_writer.jl
ali-ramadhan: Ah so I suppose `dx` is the "canonical" name, but the `dim_name_generator` can be used to override this name!

For example, a custom `dim_name_generator` can be passed to the `NetCDFOutputWriter` constructor to change `dx` to `Δx` or `delta_x` before passing it back to `Oceananigans.OutputWriters.default_dim_name`.

This is totally not obvious from the code (and the docstring) so I should add a test for this and probably another docstring example. It's how I was thinking we can satisfy desired naming conventions.

But happy to switch to another better design!

src/OutputWriters/netcdf_output_writer.jl
ali-ramadhan: Yes! This function only dispatches on `GridFittedBottom` (via `GFBIBG`) but I need to also define `gather_immersed_boundary` for `PartialCellBottom` and `GridFittedBoundary`.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -1,135 +1,716 @@
...
+
+    xᶠᵃᵃ_attrs = Dict("long_name" => "Locations of the cell faces in the x-direction.",   "units" => "m")
+    xᶜᵃᵃ_attrs = Dict("long_name" => "Locations of the cell centers in the x-direction.", "units" => "m")
+    yᵃᶠᵃ_attrs = Dict("long_name" => "Locations of the cell faces in the y-direction.",   "units" => "m")
+    yᵃᶜᵃ_attrs = Dict("long_name" => "Locations of the cell centers in the y-direction.", "units" => "m")
```
tomchor, line 418:
```suggestion
    xᶠᵃᵃ_attrs = Dict("long_name" => "Cell face locations in the x-direction.",   "units" => "m")
    xᶜᵃᵃ_attrs = Dict("long_name" => "Cell center locations in the x-direction.", "units" => "m")
    yᵃᶠᵃ_attrs = Dict("long_name" => "Cell face locations in the y-direction.",   "units" => "m")
    yᵃᶜᵃ_attrs = Dict("long_name" => "Cell center locations in the y-direction.", "units" => "m")
```

I always thought these names were pretty long and not super easy to read as axis labels in figures. I made them a bit shorter but I wonder if we can do better. Perhaps `x node locations (faces)` on so on? That way `x` is the first thing in the description, making it easier for the user/reader/viewer.

tomchor: Also we have always assumed SI units for the outputs. I wonder if it would be hard to include a flag to not do that.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -1,135 +1,716 @@
...
+        vertical_dimension_attributes
+    )
+end
+
+function default_dimension_attributes(grid::LatitudeLongitudeGrid, dim_name_generator)
+    λᶠᵃᵃ_name = dim_name_generator("longitude", grid, f, nothing, nothing, Val(:x))
+    λᶜᵃᵃ_name = dim_name_generator("longitude", grid, c, nothing, nothing, Val(:x))
+
+    λᶠᵃᵃ_attrs = Dict("long_name" => "Locations of the cell faces in the zonal direction.",   "units" => "degrees east")
+    λᶜᵃᵃ_attrs = Dict("long_name" => "Locations of the cell centers in the zonal direction.", "units" => "degrees east")
```
tomchor, line 450: Same comment here for the names and units.

tomchor: Ah, I now understand that these are just the default ones! Somewhere down there must be an option to provide user-defined ones. Got it! Disregard the other comments about this,

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -1,135 +1,716 @@
...
+    return attrs, dims
+end
+
+function grid_reconstruction_attributes(ibg::ImmersedBoundaryGrid)
+    attrs, dims = grid_reconstruction_attributes(ibg.underlying_grid)
+
+    immersed_attrs = Dict(
+        "immersed_boundary_type" => string(nameof(typeof(ibg.immersed_boundary)))
+    )
+
```
tomchor, line 690: :heart:

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -173,91 +757,135 @@ function add_schedule_metadata!(global_attributes, schedule::AveragedTimeInterva
...
-                       overwrite_existing = false,
+                       with_halos = false,
+                       include_grid_metrics = true,
+                       overwrite_existing = nothing,
+                       verbose = false,
                        deflatelevel = 0,
                        part = 1,
                        file_splitting = NoFileSplitting(),
-                       verbose = false)
+                       dimension_name_generator = default_dim_name)
```
tomchor, line 803: This is a great way to do things. Maybe in the future we can include functions that already follow a given popular/useful convention. I think we should probably also provide one that uses the exact names Oceananigans uses (I assume that'll be needed for the integration with `FieldTimeSeries`?)

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -397,32 +1032,33 @@ function NetCDFOutputWriter(model, outputs;
...
-    # TODO: This call to dictify is only necessary because "dictify" is hacked to help
-    # with LagrangianParticles output (see the end of the file).
-    # We shouldn't support this in the future; we should require users to 'name' LagrangianParticles output.
-    outputs = dictify(outputs)
-    outputs = Dict(string(name) => construct_output(outputs[name], grid, indices, with_halos) for name in keys(outputs))
+
+    outputs = Dict(
+        string(name) => construct_output(outputs[name], grid, indices, with_halos)
+        for name in keys(outputs)
+    )
```
tomchor, line 1039: Probably we should follow the conventions we already use in Oceananigans?

```suggestion
outputs = Dict(string(name) => construct_output(outputs[name], grid, indices, with_halos)
               for name in keys(outputs))
```

Or in this case it might be better to even do a one-liner.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -431,25 +1067,173 @@ function NetCDFOutputWriter(model, outputs;
...
+                    materialized,
+                    name,
+                    array_type,
+                    deflatelevel,
+                    attributes,
+                    dimensions,
+                    filepath, # for better error messages
+                    dimension_name_generator,
+                    false # time_dependent = false
+                )
```
tomchor, line 1190:
```suggestion
                define_output_variable!(dataset,
                                        materialized,
                                        name,
                                        array_type,
                                        deflatelevel,
                                        attributes,
                                        dimensions,
                                        filepath, # for better error messages
                                        dimension_name_generator,
                                        false # time_dependent = false)
```

src/OutputWriters/netcdf_output_writer.jl
tomchor: Yeah I figured that somewhere down in between this and the last comment haha. There's a comment somewhere about it. I think the way you're doing is great!

tomchor:
I haven't checked the tests yet, but so far it's looking great! I'll try to check the tests tomorrow.

glwagner:
> Supporting output of the free surface height is a bit hacky. Since it's a regular Field but is kind of a ReducedField we cannot dispatch on its type and work with it correctly. Will open an issue to discuss.

It's not a ReducedField, its a WindowedField. But properly supporting WindowedField is intrinsic to NetCDF output writing without halos so there should be a general solution to this.

glwagner:
> I'm tagging v0.96.0 since this PR significantly changes the dimension names used in NetCDF files produced by NetCDFOutputWriter.

I'm not sure a FieldTimeSeries refactor is necessary (note that we also support NetCDF FieldTimeSeries on ClimaOcean, so I don't entirely understand why refactoring is necessary). However, in the case that such a refactor bumps the version again, would it be prudent to be more conservative about version bumping here?

src/OutputWriters/netcdf_output_writer.jl
glwagner: Agree with this comment generally, we can start a discussion to format the code differently throughout, but either way I think we should strive to format consistently.

test/test_netcdf_output_writer.jl
```diff
@@ -46,133 +55,1427 @@ function test_DateTime_netcdf_output(arch)
...
+    @test dimsize(ds_s[:T]) == (y_c=ny, z_c=nz, time=Nt + 1)
+    @test dimsize(ds_s[:S]) == (y_c=ny, z_c=nz, time=Nt + 1)
+
+    close(ds_s)
+    rm(filepath_sliced)
+
+    return nothing
+end
+
+function test_netcdf_rectilinear_column(arch)
```
tomchor, line 1341: I see that the tests are pretty thorough! Well done!

I guess my only comment is that it seems we're re-writing a bunch of code. For example, `test_netcdf_rectilinear_flat_yz()` is pretty much the same thing as their `xy` and `xz` counterparts. I wonder if it would be better to try and re-use much of that code, so that it's easier to maintain.

On the other hand, re-using code like this can make way less readable, so I'm torn. I just wanted to comment this so that we think at least think about it.

@ali-ramadhan what are your thoughts?

tomchor: Also it seems that specifically `test_netcdf_rectilinear_column()` is the only topology where we don't test all three directions...

glwagner:
> > Supporting output of the free surface height is a bit hacky. Since it's a regular Field but is kind of a ReducedField we cannot dispatch on its type and work with it correctly. Will open an issue to discuss.
> 
> It's not a ReducedField, its a WindowedField. But properly supporting WindowedField is intrinsic to NetCDF output writing without halos so there should be a general solution to this.

In case its helpful: a WindowedField is a Field with non-default `indices`. We omit halo regions by producing a windowed field with indices between `(1:Nx, 1:Ny, 1:Nz)` (for example), rather than `(:, :, :)` which span the entire domain including halos. The free surface has indices `(:, :, Nz+1)` --- indicating that it is not a reduced field with physical validity throughout the domain at `z`, but rather a Field that occupies a specific slice of the domain (`k=Nz+1`).

Thus with a general way of "windowing" dimensions, we should support free surface and `with_halos=false`. I'm not sure how the latter is supported without supporting the free surface, though.

src/OutputWriters/netcdf_output_writer.jl
ali-ramadhan: This is done now and is actually tested in `test_lagrangian_particle_tracking.jl`!

src/OutputWriters/netcdf_output_writer.jl
ali-ramadhan: Yeah I didn't want to dig into standards/conventions for this PR, but I'm hoping a custom `dimension_name_generator` would be enough to produce output that conforms to certain conventions. If not, we can extend it!

I didn't make the default use the Oceananigans names as it included unicode and superscripts, which I like and Python 3 supports, but then users may struggle to work with the data efficiently outside of Julia.

Not sure if using the Oceananigans names would make it much easier to integrate with `FieldTimeSeries`, but if it does then I'll definitely add such a `dimension_name_generator`.

Do you think it's necessary for this PR or should we wait for the subsequent PR to support `FieldTimeSeries` before deciding?

src/OutputWriters/netcdf_output_writer.jl
ali-ramadhan: Yes I think I got carried away with the refactoring haha, but 100% agree the code should conform to the existing style. I will format the code consistently.

ali-ramadhan:
> > I'm tagging v0.96.0 since this PR significantly changes the dimension names used in NetCDF files produced by NetCDFOutputWriter.
> 
> I'm not sure a FieldTimeSeries refactor is necessary (note that we also support NetCDF FieldTimeSeries on ClimaOcean, so I don't entirely understand why refactoring is necessary). However, in the case that such a refactor bumps the version again, would it be prudent to be more conservative about version bumping here?

Ah I may have just assumed a large refactor would be necessary without digging into `FieldTimeSeries` first!

I believe that according to semantic versioning which Oceananigans.jl follows (still I think?), this PR would necessitate a minor version bump as it changes the `NetCDFOutputWriter` interface and any code that interfaces with the NetCDF output files will break.

It's probably not nice to increment the minor version very often, but is there a particular reason to be conservative about version bumping?

test/test_netcdf_output_writer.jl
ali-ramadhan: Yeah I also struggled with repeated code vs. less understandable generic code here :( Although I did feel like making it more generic made it quite a bit less readable and debuggable.

Since it was tests where one case can break (e.g. `xy` but not `yz`) I decided to err on the side of repeating code but making it easier to debug and modify specific tests.

glwagner:
> > > I'm tagging v0.96.0 since this PR significantly changes the dimension names used in NetCDF files produced by NetCDFOutputWriter.
> > 
> > 
> > I'm not sure a FieldTimeSeries refactor is necessary (note that we also support NetCDF FieldTimeSeries on ClimaOcean, so I don't entirely understand why refactoring is necessary). However, in the case that such a refactor bumps the version again, would it be prudent to be more conservative about version bumping here?
> 
> Ah I may have just assumed a large refactor would be necessary without digging into `FieldTimeSeries` first!
> 
> I believe that according to semantic versioning which Oceananigans.jl follows (still I think?), this PR would necessitate a minor version bump as it changes the `NetCDFOutputWriter` interface and any code that interfaces with the NetCDF output files will break.
> 
> It's probably not nice to increment the minor version very often, but is there a particular reason to be conservative about version bumping?

For sure if introduces a breaking change to `NetCDFOutputWriter` then we need a new minor version.

src/OutputWriters/netcdf_output_writer.jl
tomchor: nice!

tomchor:
@ali-ramadhan can you double-check my last commit https://github.com/CliMA/Oceananigans.jl/pull/4046/commits/8dece8d6f1b558b709b873730e8843fedc1a02af?

It was the easiest way I found to make NetCDF output `construct_output` for `LagrangianParticles`, but that was the same solution that was implemented before, so I figured you may have had good reason to not use it. Can you advise?

docs/src/model_setup/output_writers.md
```diff
@@ -106,13 +106,12 @@ h(model) = @. model.clock.time * sin(XC) * cos(YF) # x-y slice output (2D array)
 
 outputs = Dict("scalar" => f, "profile" => g, "slice" => h)
 
-dims = Dict("scalar" => (), "profile" => ("zC",), "slice" => ("xC", "yC"))
+dims = Dict("scalar" => (), "profile" => ("z_c",), "slice" => ("x_c", "y_c"))
```
glwagner, line 109: are these going to become our official shorthand for `zᶜᶜᶜ`, etc?

glwagner: if "c" means tracer, this generalizes to other grids, but if it refers to a single location then its not sufficient

ali-ramadhan:
> @ali-ramadhan can you double-check my last commit [8dece8d](https://github.com/CliMA/Oceananigans.jl/commit/8dece8d6f1b558b709b873730e8843fedc1a02af)?
> 
> It was the easiest way I found to make NetCDF output `construct_output` for `LagrangianParticles`, but that was the same solution that was implemented before, so I figured you may have had good reason to not use it. Can you advise?

Looks good and thanks for bringing it back! I think I removed it so the interface would always expect a `Dict` or `NamedTuple` but `(; particles = model.particles)` isn't as nice as just specifying `model.particles.

docs/src/model_setup/output_writers.md
ali-ramadhan: `z_c` is the default dimension name for `zᶜᶜᶜ` on a `RectilinearGrid`. It can be changed by passing a custom `dimension_name_generator`.

I believe @tomchor suggested adding another dimension name generator that uses the Oceananigans names, e.g. `z_ccc`. Is this something we want to implement or use as the default?

I guess it's technically `zᵃᵃᶜ` in the code, but `z_ccc` would be correct. Although one reason I didn't go with the full name is because just having a `z_ccc` variable may imply to the user that `z_fcc`, `z_ffc`, etc. are different but not included. And including `a` would require more explanation and may confuse users. We can use metadata to add explanations though.

I feel like there's a tradeoff between correctness and user friendliness here.

Note that on a `LatitudeLongitudeGrid` you end up with e.g. `longitude_f` and `longitude_c` but `dx_ff`, `dx_fc`, `dx_cf`, and `dx_cc`. So we're basically omitting the `a` and describing the dimensions via metadata in a way that is hopefully clear to the user.

I went with ASCII characters rather than Unicode superscripts as the default since I don't know if every other tool that reads those NetCDF files will be okay with superscripts.

tomchor: > I feel like there's a tradeoff between correctness and user friendliness here.

I think that's true. I personally don't mind either `z_c` or `z_aac`, but I think if we're going with a triple location identifier, then should use what's in the code as the default. So for `RectilinearGrid` we'd use `z_aac`, since that's what's in the `grid` object, not `z_ccc`. (I guess technically `RectilinearGrid` uses `grid.z.cᵃᵃᶜ` these days but the point stands.)

Similarly we'd follow Ocenanigans conventions for other grids.

> I went with ASCII characters rather than Unicode superscripts as the default since I don't know if every other tool that reads those NetCDF files will be okay with superscripts.

I'd also be okay with either choice here. There is some advantage in keeping names as similar as we can to what we use in the code, but I'm guessing most non-julia users will have trouble (or at least be annoyed) to type superscripts. Perhaps we could get the opinion from other NetCDF users?

tomchor: A couple of notes on unicode:

1. I think every widely used software for reading NetCDF _can_ read unicode. It just may need a few extra steps. As an example, python understands `zᵃᵃᶜ` as `zaac`:
```python
>>> zᵃᵃᶜ = 2
>>> zaac
2
```
Which means that accessing `zᵃᵃᶜ` in an xarray dataset can't be done like `ds.zᵃᵃᶜ` (which is usually how it's done) because python would actually look for `ds.zaac`. But it can be done as `ds["zᵃᵃᶜ"]`. An extra step.

2. I also think we'd be well-justified in focusing on Python compliance for non-Julia netcdf readers. The number of people using python grew a lot in the past years (along with the ecosystem) and at this point I'm sure most people (in our community) use python. It's harder to get precise numbers for matlab, but based on data I could find it seems to have plateaued over the past few years. (Plus matlab is paid, and, speaking from experience, non-first-world universities generally don't buy licenses.) Not sure if there are other major software that people have been using these days. (ncview?)

3. Note that python does understand unicode characters, as long as they represent letters in some language. So for example, something `ds.Δφᵃᶠᵃ` wouldn't work in python (and you'd have to resort to `ds["Δφᵃᶠᵃ"]`), but `ds.Δφ_afa` would be perfectly fine.

Lastly, like @ali-ramadhan points out, we are offering the possibility of `dimension_name_generator`, which I think means we're okay picking a default that won't make everyone happy if we feel like there are other advantages.

src/OutputWriters/netcdf_output_writer.jl
```diff
@@ -397,32 +1007,31 @@ function NetCDFOutputWriter(model, outputs;
             @warn "Overwriting existing $filepath."
         end
     end
-    # TODO: This call to dictify is only necessary because "dictify" is hacked to help
-    # with LagrangianParticles output (see the end of the file).
-    # We shouldn't support this in the future; we should require users to 'name' LagrangianParticles output.
+
     outputs = dictify(outputs)
```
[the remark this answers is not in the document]
tomchor: Sorry, by "this" I mean the `outputs = dictify(outputs)` on line 1010. I can't comment on non-changed lines apparently.

tomchor:
It seems like the server (sverdrup?) needs cleaning again:

image

Otherwise I think tests are passing.

glwagner:
should be clear now

ali-ramadhan:
I'm in favor of `NetCDFWriter` and `JLD2Writer`! Agree it would be good to have one breaking release, especially if most people are also in favor of the change.

glwagner:
> I'm in favor of `NetCDFWriter` and `JLD2Writer`! Agree it would be good to have one breaking release, especially if most people are also in favor of the change.

Happy for this. Note, we can have multiple PRs into one release though. Also this should be eliminated as a concern by #3793, but that hasn't progressed in a while.

tomchor:
Created a PR to do this in https://github.com/CliMA/Oceananigans.jl/pull/4210

tomchor:
In addition to the `minimal_dim_name()` generator (which generates names like `x_c` for the `RectilinearGrid`), I also just added the `trilocation_dim_name()` generator (which always generates names like `x_caa`, mimicking the nomenclature we use for `grid`).

Having more time to think about this, I actually think that we should try to follow Oceananigans' native nomenclature as closely as possible in the NetCDF output without using problematic characters such as superscripts. If we agree on this, then:

1. I think we should use the latter generator (`trilocation_dim_name()`, which generates `x_caa` for `RectGrid`) as the default name generator and keep `minimal_dim_name()` as an available option.
2. Based on my points in https://github.com/CliMA/Oceananigans.jl/pull/4046#discussion_r1987577045, I also think we should be using greek letters to keep dimension names as close to Oceananigans standards as possible. So for example spacings in `RectGrid`s would go from the current `dx_c` to `Δx_caa`. Likewise `dlongitude` should become `Δλ_faa`, which is what `LatitudeLongitudeGrid` uses, etc.

And again, these are just the defaults. @ali-ramadhan did a really good job allowing for flexibility with nomenclature, so if the user wants, they can overwrite these choices pretty simply. But I think we should keep the default as close to what the source code uses as possible to foster better connection between NetCDF output and Julia code.

I believe this is the last point to discuss in this PR, so as soon as this is decided we can move forward and merge it.

docs/src/model_setup/output_writers.md
ali-ramadhan: Thanks for pointing this out about how Python deals with Unicode!

ali-ramadhan:
@tomchor Awesome to see the tests passing, thank you!

I'm happy with Greek letters as the default, especially since it can be easily changed and because Python will work with it seamlessly (as opposed to superscripts from what you've described).

I also agree that ensuring seamless Python support is the most worthwhile effort.

docs/src/model_setup/output_writers.md
glwagner: I think we should include the `a` and more generally always 3 letters. There's many advantages to using the same naming system as in the source code. If we think its confusing, we should change the source code too.

tomchor (approved):
Looks good! I think we're ready.

glwagner:
I mentioned this elsewhere, but I am wondering if it would make sense to move `NetCDFWriter` to an extension as part of this breaking change (we can do this in a subsequent PR. This will mean that users have to write `using NCDatasets` at the top of their scripts (a downside), but has the advantage that we can run painlessly on systems that don't support NetCDF out of the box or require special configuration. We keep running into such cases so I think an extension is a good design.

glwagner:
PS to relieve pressure on CI, can I ask for merging only when work is going to be done? There are lots of tiny PRs that are likely irrelevant for this work, so we don't need to merge all of them

ali-ramadhan:
I think making `NetCDFWriter` part of an extension makes sense! It does depend on the NetCDF C library which doesn't run seamlessly everywhere (e.g. NCDatasets.jl doesn't support PowerPC systems like the Satori cluster).

> PS to relieve pressure on CI, can I ask for merging only when work is going to be done? There are lots of tiny PRs that are likely irrelevant for this work, so we don't need to merge all of them independently

Definitely makes sense. Just to clarify, are you suggesting that we line up the three PRs (this one, https://github.com/CliMA/Oceananigans.jl/pull/4210, and the extension one [which could be part of #4210?]) so that there's very little time lag between merging them and between merging and tagging v0.96.0?

glwagner:
Also this will nicely demonstrate the extension pattern for output writers, so that a ZarrWriter will be easier to build too (I've heard noise here and there about desire for that)

PR #4050, opened 2025-01-18
Fix bug in validation of `MutableVerticalDiscretization()`

tomchor:
I tried playing around with it and ended up finding a small typo (leading to a bug) that this PR fixes.


simone-silvestri (approved):

simone-silvestri:
Also, because of last-minute UI changes, you should pass a `vertical_coordinate = Oceananigans.Models.ZStar()` to the model constructor to use zstar.

PR #4053, opened 2025-01-20
Remove all hardcoded Float64 from the `Advection` module

simone-silvestri:
and also correct some `Float64` that slipped in the `SplitExplicitFreeSurface`. This is some preparatory work to allow more flexible user type and eventually fully support Metal.

Note that I have tested `Metal` on an additional branch that implements these changes + some (quite easy actually) support for Metal architectures https://github.com/CliMA/Oceananigans.jl/tree/ss/fix-metal (as well as some other tangential stuff I was trying which is not relevant) and this [example](https://github.com/CliMA/Oceananigans.jl/blob/ss/fix-metal/validation/mesoscale_turbulence/baroclinic_adjustment.jl) seem to run quite nicely with WENO, so after this PR is merged, the route for Metal should be
- enable easy Float32 support
- implement a `MetalGPU` (alias for a possible `GPU{<:Metal.MetalBackend}`) architecture


src/Advection/Advection.jl
```diff
@@ -54,6 +54,9 @@ abstract type AbstractUpwindBiasedAdvectionScheme{B, FT} <: AbstractAdvectionSch
 # Note that it is not possible to compile schemes for `advection_buffer = 41` or higher.
 const advection_buffers = [1, 2, 3, 4, 5, 6]
 
+# To add support for new floating types add them here
+const supported_float_types = [Float32, Float64]
```
glwagner, line 58: This is a bit buried in the Advection module. Should we put this in the top-level file?

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
```diff
@@ -256,7 +256,7 @@ end
     averaging_weights = averaging_weights[1:idx]
     averaging_weights ./= sum(averaging_weights)
 
-    return Δτ, tuple(averaging_weights...)
+    return Δτ, convert.(FT, tuple(averaging_weights...))
```
glwagner, line 259: Always better to use “map” instead of broadcasting

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_field_tuples.jl
```diff
@@ -27,8 +27,8 @@ end
 function hydrostatic_tendency_fields(velocities, free_surface::SplitExplicitFreeSurface, grid, tracer_names)
     u = XFaceField(grid)
     v = YFaceField(grid)
-    U = deepcopy(free_surface.barotropic_velocities.U)
-    V = deepcopy(free_surface.barotropic_velocities.V)
+    U = similar(free_surface.barotropic_velocities.U)
+    V = similar(free_surface.barotropic_velocities.V)
```
glwagner, line 31: This is unrelated right? The difference is that “similar” doesn’t copy boundary conditions.

glwagner (approved):

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_field_tuples.jl
simone-silvestri: Right, this was a leftover from the Metal branch (that I stripped down for this PR). Apparently, there is a bug with `deepcopy` and Metal OffsetArrays:
```julia
julia> a = MtlArray(zeros(Float32, 10));

julia> a = OffsetArray(a, -1);

julia> b = deepcopy(a);

julia> fill!(b, 1);

julia> a
10-element OffsetArray(::MtlVector{Float32, Metal.PrivateStorage}, 0:9) with eltype Float32 with indices 0:9:
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
 1.0
```
I wanted to raise an issue in OffsetArrays.jl but I haven't done it yet (there is some type-piracy of Base going on in that package).

simone-silvestri: with similar, on the other hand, there is no issue:
```julia
julia> a = MtlArray(zeros(Float32, 4));

julia> a = OffsetArray(a, -1);

julia> b = similar(a);

julia> a
4-element OffsetArray(::MtlVector{Float32, Metal.PrivateStorage}, 0:3) with eltype Float32 with indices 0:3:
 0.0
 0.0
 0.0
 0.0

```

glwagner: weird

src/Oceananigans.jl
```diff
@@ -132,6 +132,10 @@ function __init__()
     end
 end
 
+# Supported types for the Advection module.
```
glwagner, line 135: This comment should say:

# List of fully-supported floating point types where applicable.
# Currently used in the Advection module for WENO advection schemes.

This comment communicates that the type is designed to be more general than just for advection, and that the comment should be changed if future development occurs.

Currently, the comment is misleading because one can interpret that a different type should be used for other modules.

src/Oceananigans.jl
```diff
@@ -132,6 +132,10 @@ function __init__()
     end
 end
 
+# Supported types for the Advection module. 
+# To support new floating point types, add them here
+const supported_float_types = [Float32, Float64]
```
glwagner, line 137: I think this should be a tuple rather than array. Otherwise users can change it eg

```julia
push!(Oceananigans.supported_float_types, Float16)
```

but of course this does nothing.

simone-silvestri:
Sounds good, I ll do the changes

PR #4061, opened 2025-01-24
Default free surface for distributed grids

simone-silvestri:
This PR proposes a default free surface for distributed grids.

A distributed grid requires knowing apriori the number of subcycling steps of the barotropic solver.
Here I have hardcoded 60 because I tend to think it is a rather safe default for domains that have a depth similar to the one of the ocean and do not have terribly stretched horizontal coordinates. The free surface speed tends to be about 100 times faster than the maximum advection speed accounting for about 6km depth and 10m2/s gravitational acceleration. 
Taking into account that the CFL for the free surface is typically much larger than the one for advection we reach a factor of about 60 difference. 

However, hardcoding a substep number might be a bit wasteful if the domain is shallow (probably not a big problem as the free surface solver is very cheap), and a little dangerous if we use better time-stepping schemes for the baroclinic step which allow a higher CFL for advection or the domain is very deep.

What do people think?


src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -49,18 +49,28 @@ mutable struct HydrostaticFreeSurfaceModel{TS, E, A<:AbstractArchitecture, S,
          vertical_coordinate :: Z        # Rulesets that define the time-evolution of the grid
 end
 
-default_free_surface(grid::XYRegularRG; gravitational_acceleration=g_Earth) =
+default_free_surface(grid::XYRegularRG; gravitational_acceleration=g_Earth, args...) =
```
glwagner, line 52: What do you think about using the same default for all grids? I know it is maybe less efficient but there is an advantage to having the setup remain the same when you change grids / configurations. It will also probably be less confusing to debug user configurations.

When we eventually finally have tutorials on model-building, we can explain the default and how it might be optimized depending on the setting.

Basically my idea is to have a default that improves our ability to understand the model and simulations (ie maximizes convenience) --- rather than the alternative goal of choosing a default that achieves optimal performance.

glwagner: So the first question to ask is: what is the purpose of a default?

simone-silvestri: I think the right goal to strive for in a default is stability. A user wouldn't mind if the simulation is 2 or 3 times slower than it could theoretically be, it would be more annoying if he gets a NaN and doesn't know why. 

So for the goal of convenience, I am ok uniformizing the free surface for all grids. We can keep the `SplitExplicitFreeSurface` for all grids since it is the only supported one with all grids. I don't know if we should distinguish between a variable and a fixed substep size where supported or, in the spirit of keeping all defaults the same using only a fixed substep size which is the only supported configuration for every grid. The latter option  involves guessing which baroclinic time step the user will use, which might create some stability problems if the user chooses a very large time step and infringes the principle of having a stable simulation with the default configuration.

I would probably keep a variable substep size for serial grids (which should be always stable), while we could make the assumption that users that use distributed grids might have larger problems at hand and tolerate some more tampering with the settings, so as to keep the fixed substep default.

glwagner: Okay, that makes sense. Basically ideally we could use the same configuration for all scenarios including Distributed, but the complexity of our performance optimization for distributed cases introduces an important trade-off. So we will compromise and use split-explicit always, but different approaches for serial vs distributed grids: variable substeps on serial grids to guarantee stability, fixed substeps on distributed grids (the only thing we support currently).

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -103,11 +103,27 @@ mutable struct HydrostaticFreeSurfaceModel{TS, E, A<:AbstractArchitecture, S,
     vertical_coordinate :: Z    # Rulesets that define the time-evolution of the grid
 end
 
-default_free_surface(grid::XYRegularRG; gravitational_acceleration=defaults.gravitational_acceleration) =
-    ImplicitFreeSurface(; gravitational_acceleration)
+default_free_surface(grid; gravitational_acceleration=g_Earth, args...) =
+    SplitExplicitFreeSurface(grid; cfl=0.7, gravitational_acceleration)
```
```suggestion
default_free_surface(grid; gravitational_acceleration=defaults.gravitational_acceleration, args...) =
    SplitExplicitFreeSurface(grid; cfl=0.7, gravitational_acceleration)
```

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -103,11 +103,27 @@ mutable struct HydrostaticFreeSurfaceModel{TS, E, A<:AbstractArchitecture, S,
...
+
+# A heuristic computation of a possible maximum Δt for a HydrostaticFreeSurfaceModel, 
+# given the grid spacings, a hypothetical `maximum_speed` (assumed to be around 3 m/s)
+# and a conservative CFL of 0.3. Note that this computation only considers the ``advective''
+# CFL, as we assume that at a high enough resolution advective processes are the bottleneck.
+function estimate_maximum_Δt(grid; maximum_speed = 3)
+    Δx  = minimum_xspacing(grid)
+    Δy  = minimum_yspacing(grid)
+    Δs  = sqrt(2 / (Δx^2 + Δy^2))^(-1)
```
```suggestion
    Δs  = sqrt((Δx^2 + Δy^2) / 2)
```

glwagner (approved):

PR #4070, opened 2025-01-28
Fix Corsican bug in validation test

glwagner:
@jm-c

### DIFF
--- a/validation/mesoscale_turbulence/immersed_coarse_baroclinic_adjustment.jl
+++ b/validation/mesoscale_turbulence/immersed_coarse_baroclinic_adjustment.jl
@@ -5,7 +5,7 @@ using Oceananigans
 using Oceananigans.Units
 using GLMakie
 using Oceananigans.TurbulenceClosures: IsopycnalSkewSymmetricDiffusivity
-using Oceananigans.TurbulenceClosures: FluxTapering, DiffusiveFormulation, AdvectiveSkewClosure
+using Oceananigans.TurbulenceClosures: FluxTapering, DiffusiveFormulation, AdvectiveFormulation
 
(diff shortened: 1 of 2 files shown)


validation/mesoscale_turbulence/immersed_coarse_baroclinic_adjustment.jl
```diff
@@ -85,12 +85,8 @@ function run_simulation(closure, grid)
...
 
     simulation = Simulation(model; Δt, stop_time)
-
-    wall_clock = Ref(time_ns())
-
-    add_callback!(simulation, progress, IterationInterval(10))
-
-    suffix = closure isa AdvectiveSkewClosure ? "advective" : "diffusive"
+    add_callback!(simulation, progress, IterationInterval(144))
+    suffix = closure isa AdvectiveFormulation ? "advective" : "diffusive"
```
jm-c, line 89: It seems that this line of code does not work as expected: it returns "diffusive" in both cases. How to fix this ?

simone-silvestri:
```suggestion
    suffix = closure isa SkewAdvectionISSD ? "advective" : "diffusive"
```
but I think you need to import it:
```
using Oceananigans.TurbulenceClosures: SkewAdvectionISSD
```

jm-c: I will try.

jm-c: It works ! Thanks @simone-silvestri  I will push this little change.

PR #4073, opened 2025-01-30
Implement a default float type

### LINKED ISSUE #3800: How can we help users change number type more easily?
Running a simulation with `FT=Float32` is fairly onerous because the number type has to be passed to a huge number of model components individually. Eg we need to write

```julia
using Oceananigans
using SeawaterPolynomials: TEOS10EquationOfState

FT = Float32
grid = RectilinearGrid(FT, size=(1, 1, 1), extent=(1, 1, 1))
coriolis = FPlane(FT, f=1)
equation_of_state = TEOS10EquationOfState(FT)
buoyancy = SeawaterBuoyancy(FT; equation_of_state)
```

Here are two ideas to make this easier:

## Use an environment variable

For this we introduce an environment variable, say `OCEANANIGANS_NUMBER_TYPE`. Then we can write a small utility:

```julia
function global_number_type(default=Float64)
    user_number_type = get("OCEANANIGANS_NUMBER_TYPE", ENV, nothing)

    if isnothing(user_number_type)
        FT = default
    else
        try
            FT = eval(Symbol(user_number_type))
        catch
            error("The environment variable OCEANANIGANS_NUMBER_TYPE=$user_number_type is not supported!")
        end
    end

    return FT
end
```

This also permits a script-based approach because users can write

```julia
ENV["OCEANANIGANS_NUMBER_TYPE"] = "Float32"

grid = RectilinearGrid(size=(1, 1, 1), extent=(1, 1, 1))
coriolis = FPlane(f=1)
equation_of_state = TEOS10EquationOfState()
buoyancy = SeawaterBuoyancy(; equation_of_state)
```

## Introducing an internal "settings" dictionary

Then users might write

```julia
Oceananigans.settings[:default_number_type] = Float32
```

This avoids the string-based restriction of environment variables, so may generalize to more types of settings should we need that. It's a slippery slope of course, we want to keep this to a minimum because global settings inhibit programmability (ie one cannot have multiple simulations in a single instance with different settings). Here, the setting only controls a default, so it remains possible to have two simulations in one instance with different number types --- they just cannot be set up simultaneously (or one has to do more work to set the number type "manually").

External packages can opt into using this only if they depend on Oceananigans. Which may be ok, in fact. External packages that extend Oceananigans of course must depend on Oceananigans. There is just the question of what to do with SeawaterPolynomials, where the dependency goes the other direction...

I thought of the environment-variable approach so that SeawaterPolynomials could implement it without depending on Oceananigans. But as I think about it more, it is hacky for SeawaterPolynomials to contain an environment variable called OCEANANIGANS_NUMBER_TYPE. So now I think the second approach may be better in fact, and to support this kind of thing with SeawaterPolynomials, perhaps what we need is a way to extend the key SeawaterPolynomials constructors inside Oceananigans somehow.

I also think it would be nice to have `validate_number_type` utility for comparing number types to the grid eltype to catch inconsistent number types.

@ali-ramadhan may be interested since it I noticed some Float32's creeping into scripts...

glwagner:
This PR implements a new `Defaults` struct with value `Oceananigans.defaults`.

The design is that `defaults` are used in constructors, so changing the content of `Oceananigans.defaults` at the top of a setup script or in a REPL session can change constructed objects afterwards that use the default. For example:

```julia
using Oceananigans
Oceananigans.defaults.float_type = Float32
```

then

```julia
julia> grid = RectilinearGrid(size=(3, 3, 3), extent=(1, 1, 1))
3×3×3 RectilinearGrid{Float32, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.333333
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=0.333333
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=0.333333
```

But of course the default can still be overridden as usual

```julia
julia> grid = RectilinearGrid(Float64, size=(3, 3, 3), extent=(1, 1, 1))
3×3×3 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.333333
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=0.333333
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=0.333333
```

Closes #3800 

Note, I did not add a constructor of the form `set_default_number_type!`. Still, users cannot write anything since the design is based on a mutable `Defaults` struct, so

```julia
julia> Oceananigans.defaults = "Float64"
ERROR: invalid redefinition of constant Oceananigans.defaults
Stacktrace:
 [1] setproperty!
   @ ./Base.jl:51 [inlined]
 [2] setproperty!(x::Module, f::Symbol, v::String)
   @ Base ./Base.jl:49
 [3] top-level scope
   @ REPL[12]:1
```

fails. We could possibly catch these settings and throw errors like "it has to be a DataType". I don't want to retstrict the datatype that migth be used because there are experimental datatypes, for example BFloat16, and who knows what in the future.

TODO:

- [ ] tests


simone-silvestri (approved):
wouldn't it be better to call it `FloatType` since it is a datatype?

navidcy (commented):
So I'm a bit confused: why e.g. `BetaPlane` keeps the `FT = Float64` default while other constructors, e.g. `CostantCartesianCoriolis` gets `FT = Oceananigans.defaults.FloatType`?

Isn't the idea that we change it everywhere so that users have to prescribe it once and then it propagates everywhere?

glwagner:
> So I'm a bit confused: why e.g. `BetaPlane` keeps the `FT = Float64` default while other constructors, e.g. `CostantCartesianCoriolis` gets `FT = Oceananigans.defaults.FloatType`?
> 
> Isn't the idea that we change it everywhere so that users have to prescribe it once and then it propagates everywhere?
> 
> Or is it a matter of our choice (in this PR for example or other) which constructors use those defaults and which do not?

I think this was an omission that I fixed. Just because `BetaPlane` used `T=Float64` so a text search didn't catch it. But it's tested now.

src/Coriolis/beta_plane.jl
```diff
@@ -26,7 +19,7 @@ and `radius` according to the relations `f₀ = 2 * rotation_rate * sind(latitud
 
 By default, the `rotation_rate` and planet `radius` are assumed to be Earth's.
 """
-function BetaPlane(T=Float64; f₀=nothing, β=nothing,
+function BetaPlane(FT=Oceananigans.defaults.FloatType; f₀=nothing, β=nothing,
```
glwagner, line 22: @navidcy see here

navidcy (approved):

glwagner:
@wsmoses @jlk9 any clue what's going on with the Enzyme test? I think this is a new failure, probably related to the Enzyme version being used, because it's unrelated to the developments in this PR

https://buildkite.com/clima/oceananigans/builds/20484#0194dd48-98ec-413e-84f0-82392bda1b7e

glwagner:
This PR will allow us to automatically switch to Float32 which will be useful for work with Reactant.

wsmoses:
@glwagner no the test failure is legitimate, this changes the relative type stability of operators internally. In any case from a quick look at the log, can you see if https://github.com/EnzymeAD/Enzyme.jl/pull/2300 resolves the issue?

glwagner:
> @glwagner no the test failure is legitimate, this changes the relative type stability of operators internally. In any case from a quick look at the log, can you see if [EnzymeAD/Enzyme.jl#2300](https://github.com/EnzymeAD/Enzyme.jl/pull/2300) resolves the issue?

It looks like tests pass when I use this branch

wsmoses:
cool just released a patch with

glwagner:
nice, that fixed the test

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
Yixiao-Zhang, line 1: [@ShuangWang96](https://github.com/ShuangWang96) and I have recently found that a Float32 clock may cause a infinite loop. It is possibly because the `TimeInterval` schedule always uses Float64. `TimeInterval` may decide to take a small timestep, but this timestep is possibly smaller than the machine precision of Float32. A minimal working example \(tested in v0.95.9\) is here:
```Julia
using Oceananigans

# Numerical Technic
const arch = CPU()
const time_stepper = :RungeKutta3
const advection = WENO()

# Output
const output_interval = 1. / 3

# Time Step
const Δt = 1
const stop_time = 10.

# Grid
grid = RectilinearGrid(
    arch,
    size = (4, 4, 4),
    x = (0., 1.),
    y = (0., 1.),
    z = (0., 1.),
)

# Model
model = NonhydrostaticModel(;
    grid = grid,
    clock = Clock{Float32}(time = 0.),
    timestepper = time_stepper,
    advection = advection,
)

# Simulation
simulation = Simulation(model; Δt = Δt, stop_time = stop_time)

# Progress Function
function print_simulation_progress(simulation)
    clock = simulation.model.clock
    @info "Simulation is running" clock.iteration clock.time
    return nothing
end

simulation.callbacks[:progress] = Callback(
    print_simulation_progress,
    IterationInterval(1),
)

# Output
simulation.output_writers[:full] = NetCDFOutputWriter(
     model,
     (),
     filename = "dummy_output.nc",
     schedule = TimeInterval(output_interval),
     overwrite_existing = true,
)

# Run the Simulation
run!(simulation)
```

I am glad that `Clock` now uses Float64 by default. So, this issue has been fixed. But I have also noticed that this change is only applied to `NonhydrostaticModel`. Thus, users of the other models may still encounter similar issues.

glwagner: Do you want to make a PR to fix the other models? It is your prerogative but I am happy to fix it if you like.

glwagner: PS I also noticed this which is why I changed the NonhydrostaticModel clock default but forgot to change the default for other models.

Also I chalk this up to further issues with TimeInterval which should be solved.

glwagner: And finally just to add I think we may also need to convert the clock.time to eltype(grid) when using it for calculations to prevent Float64 promotion. One question I have is how to test this.

Yixiao-Zhang: > Do you want to make a PR to fix the other models? It is your prerogative but I am happy to fix it if you like.

Please fix it for me. Thanks!

PR #4074, opened 2025-01-30
Correct interpolation on longitude lines

simone-silvestri:
This PR enables interpolating between `LatitudeLongitudeGrid`s with different longitude references provided the covered longitude is the same

example:
```julia
julia> grid1 = LatitudeLongitudeGrid(size=(10, 1, 1), longitude=(   0, 360), latitude=(-90, 90), z=(0, 1));

julia> grid2 = LatitudeLongitudeGrid(size=(10, 1, 1), longitude=(-180, 180), latitude=(-90, 90), z=(0, 1));

julia> f1 = CenterField(grid1);

julia> f2 = CenterField(grid2);

julia> set!(f1, (λ, y, z) -> λ);

julia> fill_halo_regions!(f1);

julia> interpolate!(f2, f1);

julia> interior(f2)
10×1×1 view(::Array{Float64, 3}, 4:13, 2:2, 2:2) with eltype Float64:
[:, :, 1] =
 198.0
 234.0
 270.0
 306.0
 342.0
  18.0
  54.0
  90.0
 126.0
 162.0

julia> fill_halo_regions!(f2);

julia> interpolate!(f1, f2);

julia> interior(f1)
10×1×1 view(::Array{Float64, 3}, 4:13, 2:2, 2:2) with eltype Float64:
[:, :, 1] =
  18.0
  54.0
  90.0
 126.0
 162.0
 198.0
 234.0
 270.0
 306.0
 342.0
```

Needed, for example, to perform Atlantic and Mediterranean sea simulations in ClimaOcean.

### DIFF
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -81,12 +81,9 @@ end
     return fractional_index(x, xn, Nx) 
 end
 
-@inline convert_to_0_360(x) = ((x % 360) + 360) % 360
+@inline convert_to_0_360(x) = rem(x, 360, RoundToZero)
 
-@inline function convert_to_minus_180_180(x) 
-    x = convert_to_0_360(x)
-    return ifelse(x > 180, x - 360, x)
-end
+@inline convert_to_minus_180_180(x) = rem(x, 360, RoundNearest)
 
 # When interpolating longitude values, we convert all longitudes to the 0-360 range
 # if the parent grid starts with a positive number (λ₀ > 0), and to a -180-180 range
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -81,7 +81,7 @@ end
     return fractional_index(x, xn, Nx) 
 end
 
-@inline convert_to_0_360(x) = rem(x, 360, RoundToZero)
+@inline convert_to_0_360(x) = ((x % 360) + 360) % 360
 
 @inline convert_to_minus_180_180(x) = rem(x, 360, RoundNearest)
 
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -73,13 +73,6 @@ end
     return convert(FT, (x - x₀) / dx) + 1 # 1 - based indexing 
 end
 
-@inline function fractional_x_index(λ, locs, grid::XRegularLLG)
-    λ₀ = λnode(1, 1, 1, grid, locs...)
-    λ₁ = λnode(2, 1, 1, grid, locs...)
-    FT = eltype(grid)
-    return convert(FT, (λ - λ₀) / (λ₁ - λ₀)) + 1 # 1 - based indexing 
-end
-
 @inline function fractional_x_index(x, locs, grid::RectilinearGrid)
     loc = @inbounds locs[1]
      Tx = topology(grid, 1)()
@@ -88,12 +81,32 @@ end
     return fractional_index(x, xn, Nx) 
 end
 
-@inline function fractional_x_index(x, locs, grid::LatitudeLongitudeGrid)
+@inline convert_to_0_360(x) = ((x % 360) + 360) % 360
+@inline convert_to_minus_180_180(x) = ifelse(x > 180, x - 360, x)
+
+# When interpolating longitude values, we convert all longitudes to the 0-360 range
+# if the parent grid starts with a positive number (λ₀ > 0), and to a -180-180 range
+# if the parent grid starts with a negative number (λ₀ < 0).
+@inline function fractional_x_index(λ, locs, grid::XRegularLLG)
+    λ₀ = λnode(1, 1, 1, grid, locs...)
+    λ₁ = λnode(2, 1, 1, grid, locs...)
+    λc = convert_to_0_360(λ)
+    λc = ifelse(λ₀ < 0, convert_to_minus_180_180(λc), λc)
+    FT = eltype(grid)
+    return convert(FT, (λc - λ₀) / (λ₁ - λ₀)) + 1 # 1 - based indexing 
+end
+
+# When interpolating longitude values, we convert all longitudes to the 0-360 range
+# if the parent grid starts with a positive number (λ₀ > 0), and to a -180-180 range
+# if the parent grid starts with a negative number (λ₀ < 0).
+@inline function fractional_x_index(λ, locs, grid::LatitudeLongitudeGrid)
     loc = @inbounds locs[1]
-     Tx = topology(grid, 1)()
-     Nx = length(loc, Tx, grid.Nx)
-     xn = λnodes(grid, locs...)
-    return fractional_index(x, xn, Nx) 
+     Tλ = topology(grid, 1)()
+     Nλ = length(loc, Tλ, grid.Nx)
+     λc = convert_to_0_360(λ)
+     λn = λnodes(grid, locs...)
+     λc = ifelse(λn[1] < 0, convert_to_minus_180_180(λc), λc)
+    return fractional_index(λc, λn, Nλ) 
 end
 
 @inline fractional_y_index(y, locs, grid::YFlatGrid) = zero(grid)
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -83,7 +83,17 @@ end
 
 @inline convert_to_0_360(x) = ((x % 360) + 360) % 360
 
-@inline convert_to_minus_180_180(x) = rem(x, 360, RoundNearest)
+# Find n for which 360 * n ≤ λ ≤ 360 * (n + 1)
+@inline find_λ_range(λ) = ifelse(λ < 0, λ ÷ 360 - 1, λ ÷ 360)
+
+@inline function convert_to_λ₀_λ₀_plus360(x, λ₀)
+    x  = convert_to_0_360(x)
+    # Convert the interpolating range to a positive-always range
+    n  = find_λ_range(λ₀)
+    λ⁻ = convert_to_0_360(λ₀)
+    n  = ifelse(x > λ⁻, n, n+1)
+    return x + 360 * n
+end
 
 # When interpolating longitude values, we convert all longitudes to the 0-360 range
 # if the parent grid starts with a positive number (λ₀ > 0), and to a -180-180 range
--- a/src/Fields/interpolate.jl
+++ b/src/Fields/interpolate.jl
@@ -86,7 +86,7 @@ end
 # Find n for which 360 * n ≤ λ ≤ 360 * (n + 1)
 @inline find_λ_range(λ) = ifelse((λ < 0) & (mod(λ, 360) != 0), λ ÷ 360 - 1, λ ÷ 360)
 
-# Convert x to lie in the λ₀ : λ + 360 range by accounting for the cyclic
+# Convert x to lie in the λ₀ : λ₀ + 360 range by accounting for the cyclic
 # nature of the longitude coordinate. 
 @inline function convert_to_λ₀_λ₀_plus360(x, λ₀)
     x  = convert_to_0_360(x)
@@ -96,9 +96,9 @@ end
     return x + 360 * n
 end
 
-# When interpolating longitude values, we convert all longitudes to the 0-360 range
-# if the parent grid starts with a positive number (λ₀ > 0), and to a -180-180 range
-# if the parent grid starts with a negative number (λ₀ < 0).
+# When interpolating longitude values, we convert the longitude to 
+# interpolate to lie in the λ₀ : λ₀ + 360 range, where λ₀ is the westernmost node
+# of the interpolating grid.
 @inline function fractional_x_index(λ, locs, grid::XRegularLLG)
     λ₀ = λnode(1, 1, 1, grid, locs...)
     λ₁ = λnode(2, 1, 1, grid, locs...)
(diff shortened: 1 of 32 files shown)


src/Fields/interpolate.jl
```diff
@@ -88,12 +81,32 @@ end
     return fractional_index(x, xn, Nx) 
 end
 
-@inline function fractional_x_index(x, locs, grid::LatitudeLongitudeGrid)
+@inline convert_to_0_360(x) = ((x % 360) + 360) % 360
+@inline convert_to_minus_180_180(x) = ifelse(x > 180, x - 360, x)
```
navidcy, line 85: `convert_to_0_360` works for any `x` _but_ `convert_to_minus_180_180` assumes that the input is within `[0, 360]`

```Julia
julia> convert_to_0_360(100000)
280

julia> convert_to_0_360(-100000)
80

julia> convert_to_minus_180_180(-100000)
-100000

julia> convert_to_minus_180_180(100000)
99640
```

src/Fields/interpolate.jl
```diff
@@ -88,12 +81,32 @@ end
...
+
+# When interpolating longitude values, we convert all longitudes to the 0-360 range
+# if the parent grid starts with a positive number (λ₀ > 0), and to a -180-180 range
```
navidcy, line 88:
```suggestion
# When interpolating longitude values, we convert all longitudes to lie within range [0, 360]
# if the parent grid starts with a positive number (λ₀ > 0), and to range [-180, 180]
```

src/Fields/interpolate.jl
navidcy: How about

```Julia
@inline function convert_to_minus_180_180(x)
    x = convert_to_0_360(x)
    return ifelse(x > 180, x - 360, x)
end
```

simone-silvestri: right, I was using it after `convert_to_0_360`. But we can still embed x in `convert_to_0_360` directly inside `convert_to_minus_180_180 `

navidcy (approved):

src/Fields/interpolate.jl
```diff
@@ -88,12 +81,35 @@ end
     return fractional_index(x, xn, Nx) 
 end
 
-@inline function fractional_x_index(x, locs, grid::LatitudeLongitudeGrid)
+@inline convert_to_0_360(x) = ((x % 360) + 360) % 360
+@inline function convert_to_minus_180_180(x) = 
+    x = convert_to_0_360(x)
+    return ifelse(x > 180, x - 360, x)
+end
```
navidcy: Actually, we can get the same with: 

```suggestion
@inline convert_to_0_360(x) = rem(x, 360, RoundToZero)
@inline convert_to_minus_180_180(x) = rem(x, 360, RoundNearest)
```

Got this from:

```Julia
help?> %
search: %

  rem(x::Integer, T::Type{<:Integer}) -> T
  mod(x::Integer, T::Type{<:Integer}) -> T
  %(x::Integer, T::Type{<:Integer}) -> T

  Find y::T such that x ≡ y (mod n), where n is the number of integers
  representable in T, and y is an integer in [typemin(T),typemax(T)]. If T can
  represent any integer (e.g. T == BigInt), then this operation corresponds to
  a conversion to T.

  Examples
  ≡≡≡≡≡≡≡≡

  julia> x = 129 % Int8
  -127

  julia> typeof(x)
  Int8

  julia> x = 129 % BigInt
  129

  julia> typeof(x)
  BigInt

  ────────────────────────────────────────────────────────────────────────────

  rem(x, y)
  %(x, y)

  Remainder from Euclidean division, returning a value of the same sign as x,
  and smaller in magnitude than y. This value is always exact.

  See also: div, mod, mod1, divrem.

  Examples
  ≡≡≡≡≡≡≡≡

  julia> x = 15; y = 4;

  julia> x % y
  3

  julia> x == div(x, y) * y + rem(x, y)
  true

  julia> rem.(-5:5, 3)'
  1×11 adjoint(::Vector{Int64}) with eltype Int64:
   -2  -1  0  -2  -1  0  1  2  0  1  2

  ────────────────────────────────────────────────────────────────────────────

  rem(x, y, r::RoundingMode=RoundToZero)

  Compute the remainder of x after integer division by y, with the quotient
  rounded according to the rounding mode r. In other words, the quantity

  x - y * round(x / y, r)

  without any intermediate rounding.

    •  if r == RoundNearest, then the result is exact, and in the
       interval [-|y| / 2, |y| / 2]. See also RoundNearest.

    •  if r == RoundToZero (default), then the result is exact, and in
       the interval [0, |y|) if x is positive, or (-|y|, 0] otherwise.
       See also RoundToZero.

    •  if r == RoundDown, then the result is in the interval [0, y) if y
       is positive, or (y, 0] otherwise. The result may not be exact if x
       and y have different signs, and abs(x) < abs(y). See also
       RoundDown.

    •  if r == RoundUp, then the result is in the interval (-y, 0] if y
       is positive, or [0, -y) otherwise. The result may not be exact if
       x and y have the same sign, and abs(x) < abs(y). See also RoundUp.

    •  if r == RoundFromZero, then the result is in the interval (-y, 0]
       if y is positive, or [0, -y) otherwise. The result may not be
       exact if x and y have the same sign, and abs(x) < abs(y). See also
       RoundFromZero.

  │ Julia 1.9
  │
  │  RoundFromZero requires at least Julia 1.9.

  Examples:
  ≡≡≡≡≡≡≡≡≡

  julia> x = 9; y = 4;

  julia> x % y  # same as rem(x, y)
  1

  julia> x ÷ y  # same as div(x, y)
  2

  julia> x == div(x, y) * y + rem(x, y)
  true
```

Seems like we'll need to update the Julia compat to 1.9 since `RoundFromZero` requires that.

src/Fields/interpolate.jl
```diff
@@ -88,12 +81,36 @@ end
     return fractional_index(x, xn, Nx) 
 end
 
-@inline function fractional_x_index(x, locs, grid::LatitudeLongitudeGrid)
+@inline convert_to_0_360(x) = ((x % 360) + 360) % 360
+
+@inline function convert_to_minus_180_180(x) 
+    x = convert_to_0_360(x)
+    return ifelse(x > 180, x - 360, x)
+end
```
navidcy, line 89:
```suggestion
@inline convert_to_0_360(x) = rem(x, 360, RoundToZero)

@inline convert_to_minus_180_180(x) = rem(x, 360, RoundNearest)
```

Got this from 

```Julia
help?> %
search: %

  rem(x::Integer, T::Type{<:Integer}) -> T
  mod(x::Integer, T::Type{<:Integer}) -> T
  %(x::Integer, T::Type{<:Integer}) -> T

  Find y::T such that x ≡ y (mod n), where n is the number of integers
  representable in T, and y is an integer in [typemin(T),typemax(T)]. If T can
  represent any integer (e.g. T == BigInt), then this operation corresponds to
  a conversion to T.

  Examples
  ≡≡≡≡≡≡≡≡

  julia> x = 129 % Int8
  -127

  julia> typeof(x)
  Int8

  julia> x = 129 % BigInt
  129

  julia> typeof(x)
  BigInt

  ────────────────────────────────────────────────────────────────────────────

  rem(x, y)
  %(x, y)

  Remainder from Euclidean division, returning a value of the same sign as x,
  and smaller in magnitude than y. This value is always exact.

  See also: div, mod, mod1, divrem.

  Examples
  ≡≡≡≡≡≡≡≡

  julia> x = 15; y = 4;

  julia> x % y
  3

  julia> x == div(x, y) * y + rem(x, y)
  true

  julia> rem.(-5:5, 3)'
  1×11 adjoint(::Vector{Int64}) with eltype Int64:
   -2  -1  0  -2  -1  0  1  2  0  1  2

  ────────────────────────────────────────────────────────────────────────────

  rem(x, y, r::RoundingMode=RoundToZero)

  Compute the remainder of x after integer division by y, with the quotient
  rounded according to the rounding mode r. In other words, the quantity

  x - y * round(x / y, r)

  without any intermediate rounding.

    •  if r == RoundNearest, then the result is exact, and in the
       interval [-|y| / 2, |y| / 2]. See also RoundNearest.

    •  if r == RoundToZero (default), then the result is exact, and in
       the interval [0, |y|) if x is positive, or (-|y|, 0] otherwise.
       See also RoundToZero.

    •  if r == RoundDown, then the result is in the interval [0, y) if y
       is positive, or (y, 0] otherwise. The result may not be exact if x
       and y have different signs, and abs(x) < abs(y). See also
       RoundDown.

    •  if r == RoundUp, then the result is in the interval (-y, 0] if y
       is positive, or [0, -y) otherwise. The result may not be exact if
       x and y have the same sign, and abs(x) < abs(y). See also RoundUp.

    •  if r == RoundFromZero, then the result is in the interval (-y, 0]
       if y is positive, or [0, -y) otherwise. The result may not be
       exact if x and y have the same sign, and abs(x) < abs(y). See also
       RoundFromZero.

  │ Julia 1.9
  │
  │  RoundFromZero requires at least Julia 1.9.

  Examples:
  ≡≡≡≡≡≡≡≡≡

  julia> x = 9; y = 4;

  julia> x % y  # same as rem(x, y)
  1

  julia> x ÷ y  # same as div(x, y)
  2

  julia> x == div(x, y) * y + rem(x, y)
  true
```

simone-silvestri: nice!

simone-silvestri:
Hmmm, the `convert_to_0_360` does not work for negative angles. For that I guess we can use the previous implementation

glwagner (approved):

src/Fields/interpolate.jl
```diff
@@ -88,12 +81,32 @@ end
...
     loc = @inbounds locs[1]
-     Tx = topology(grid, 1)()
-     Nx = length(loc, Tx, grid.Nx)
-     xn = λnodes(grid, locs...)
-    return fractional_index(x, xn, Nx) 
+     Tλ = topology(grid, 1)()
+     Nλ = length(loc, Tλ, grid.Nx)
+     λn = λnodes(grid, locs...)
+     λ₀ = @inbounds λn[1]     
+     λc = ifelse(λ₀ < 0, convert_to_minus_180_180(λ), convert_to_0_360(λ))
```
glwagner, line 108: indenting messed up

src/Fields/interpolate.jl
```diff
@@ -88,12 +81,32 @@ end
...
 end
 
-@inline function fractional_x_index(x, locs, grid::LatitudeLongitudeGrid)
+@inline convert_to_0_360(x) = ((x % 360) + 360) % 360
+
+@inline convert_to_minus_180_180(x) = rem(x, 360, RoundNearest)
+
+# When interpolating longitude values, we convert all longitudes to the 0-360 range
+# if the parent grid starts with a positive number (λ₀ > 0), and to a -180-180 range
+# if the parent grid starts with a negative number (λ₀ < 0).
```
glwagner, line 90: I don't get it. Is this a generic solution?

glwagner:
What's the point of specifically targeting -180, 180? Why not use periodicity and always move coordinates to 0, 360? Or whatever we decide?

src/Fields/interpolate.jl
glwagner: I'm not sure it makes sense to define valid lat-lon limits here where we do interpolation. It may make more sense to enforce this in the definition of the grids, if we are trying to do this. Or somehow encode information about the coordinates being used in the grid rather than implicitly infer it here.

simone-silvestri: This is quite generic but not 100%, it does not allow grids that have (for example) a range of (-x : 360-x) if x < 180. I have to think about how we can convert to the range `λ₀ : λ₀+360` that would be completely general and allow interpolating from any lat - lon grid to any `λ` value.

simone-silvestri: enforcing it in the grid might be very complicated because the longitude coordinate should be monotone if we want to support interpolation.

simone-silvestri:
Immagine you want to have a north-atlantic grid, that crosses the zero, then you need to provide the grid with negative longitudes. By converting the angles to a common range we can support the interpolation we need, for example, in ClimaOcean. (Note that to support interpolation we need to have a monotone coordinate so we cannot just convert everything to 0 - 360. That would not work, for example, if the grid we interpolate from is defined in the range -180 : 180 because the coordinate ould not be monotonic)

src/Fields/interpolate.jl
simone-silvestri: > This is quite generic but not 100%, it does not allow all type of grids defined on the range (-x : 360-x). I have to think about how we can convert to the range `λ₀ : λ₀+360` that would be completely general and allow interpolating from any lat - lon grid to any `λ` value.

I think this is pretty easy to enforce, so I will be changing this PR to do it

glwagner:
No I understand that, but it seems like you are implicitly assuming that grids are either one way or another way. An alternative approach would be to great an abstraction that represents the coordinate. OR, you can define longitude as cyclical, so that -90 = 270 for example. The latter two approaches seem both better and simpler, don't they? What happens if your longitude is 350, 370 or something.

glwagner:
Basically you are solving a problem that has appeared but you are solving it by baking in an assumption. Whereas it looks to me like the problem is a missing abstraction.

simone-silvestri:
nono, I am trying to make it such that this interpolation is fully general with no assumption in this way

> This is quite generic but not 100%, it does not allow all type of grids defined on the range (-x : 360-x). I have to think about how we can convert to the range `λ₀ : λ₀+360` that would be completely general and allow interpolating from any lat - lon grid to any `λ` value.

Like this there is no assumption and interpolation can happen with any longitude, even if the angle is, for example 35000

glwagner:
> nono, I am trying to make it such that this interpolation is fully general with no assumption in this way
> 
> > This is quite generic but not 100%, it does not allow all type of grids defined on the range (-x : 360-x). I have to think about how we can convert to the range `λ₀ : λ₀+360` that would be completely general and allow interpolating from any lat - lon grid to any `λ` value.
> 
> Like this there is no assumption and interpolation can happen with any longitude, even if the angle is, for example 35000. This is taking advantage of the fact that the longitude is cyclical.

Can you somehow impart that in the code. It's hard to figure out by reading it.

glwagner:
Also fix the desceription of the PR. The description says 

> This PR enables interpolating between LatitudeLongitudeGrids with a -180 : 180 longitude range and LatitudeLongitudeGrids with a 0 : 360 longitude range.

But you are now telling me that it is actually far more general and has nothing specifically with grids on -180, 180. That is just one particular case that is covered by this PR?

Issue #4069, opened 2025-01-28
`DynamicSmagorinsky` is _very_ slow using default arguments and sometimes produces `NaN`s


tomchor:
I've started trying to use the Lagrangian-averaged version of the Dynamic Smagorinsky in simulations with complex bathymetry and it is way slower than I had timed it in https://github.com/CliMA/Oceananigans.jl/pull/3642. Back then iirc I timed the dynamic version as being about 4x slower than the constant version. In my current simulation it is about 12x times slower! Furthermore, it sometimes produces `NaN`s and I haven't been able to figure out why.

I haven't had a chance to investigate this much more for now, but I thought I'd post it here in case someone has experienced this and wants to chime in. I'm planning on writing a MWE sometime soon.

The good news is that results using the much simpler AMD are very close to those using dynamic Smag. At least for simulations where we AMD doesn't produce any [grid-scale noise](https://github.com/CliMA/Oceananigans.jl/discussions/3561#). I think this is some serious incentive to fix that issue on AMD.

glwagner:
Can you post an MWE that people might use for benchmarking?

tomchor:
Okay, turns out that I forgot to set the `schedule` flag in the `DynamicCoefficient`. My bad. The default is to update it at every time step, which causes the severe slowdown of 10+ times. But if I set it to update every 5 time steps, the slowdown is around 4x, which is what I had measured. Virtually everyone updates the coefficient every 5-ish time steps and it's been shown to not degrade the results. Should we set the default value of `schedule` to `iterationInterval(5)`?

That said, the simulation still blows up after some time, which is something that needs to be investigated.

glwagner:
I think that opinionated defaults are a little weird. I do think its good to target convenience but we also usually use that reasoning to decide between two defaults with otherwise similar meanings, like AB2 vs RK3. I'm not sure how we could generically justify `IterationalInterval(2)`, `IterationalInterval(4)`, `IterationalInterval(8)` --- is one more optimal than the other?

A consideration that we often think takes precedence over convenience is self-documentation of scripts. For example, it may be important that one can find things like `IterationInterval(4)` in published scripts.

Side note that 5 is a weird number. Do we choose that because its 10/2, and 10 is a nice round number? Why not 4 or 8 (following a strategy of power-law-based trial-and-error: 1, 2, 4, 8, 16, etc).

tomchor:
[...]

PR #4075, opened 2025-02-02
Improves `DynamicCoefficient` docstring with discussion about `schedule`

tomchor:
Related to https://github.com/CliMA/Oceananigans.jl/issues/4069#issuecomment-2628752818

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
@@ -67,9 +67,10 @@ cost of calculating `DynamicCoefficient`s. Because of this slowdown, it is stand
 literature to recalculate the coefficient only every few time steps, with the assumption that the
 its values don't change much from one time-step to the other. While any frequency choice is possible,
 all studies seem to recalculate every 5 steps (e.g., Bou-Zeid et al. 2005; Chen et al. 2016; Salesky et al. 2017;
-Chor et al 2021). This choice seems to stem from the results by Bou-Zeid et al. (2005) who
+Chor et al 2021). This "tradition" seems to stem from the results by Bou-Zeid et al. (2005) who
 found that considerably speed up simulations while still producing very similar results to an update
-frequency of every time step. We can achieve this by using the `schedule` keyword argument such as:
+frequency of every time step. Users can change the update frequency using the `schedule` keyword argument.
+For example, a `DynamicCoefficient` that gets updated every 4 timesteps is obtained via:
 
 ```jldoctest
 julia> using Oceananigans
@@ -77,12 +78,12 @@ julia> using Oceananigans
 julia> dynamic_coeff = DynamicCoefficient(averaging=(1, 2), schedule=IterationInterval(5))
 DynamicCoefficient with
 ├── averaging = (1, 2)
-├── schedule = IterationInterval(5, 0)
+├── schedule = IterationInterval(4, 0)
 └── minimum_numerator = 1.0e-32
 
 julia> dynamic_smagorinsky = Smagorinsky(coefficient=dynamic_coeff)
 Smagorinsky closure with
-├── coefficient = DynamicCoefficient(averaging = (1, 2), schedule = IterationInterval(5, 0))
+├── coefficient = DynamicCoefficient(averaging = (1, 2), schedule = IterationInterval(4, 0))
 └── Pr = 1.0
 ```
 
--- a/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
@@ -61,14 +61,15 @@ Smagorinsky closure with
 └── Pr = 1.0
 ```
 
-The dynamic Smagorinsky above has its dynamic coefficient recalculated at every time
-step. While this provides the highest level of accuracy, it will almost certainly be very slow
-given the high cost of calculating `DynamicCoefficient`s. Because of this slowdown, it is common in the
+The dynamic Smagorinsky above has its dynamic coefficient recalculated at every time step. While
+this provides the highest level of accuracy, it will almost certainly be very slow given the high
+cost of calculating `DynamicCoefficient`s. Because of this slowdown, it is standard practice in the
 literature to recalculate the coefficient only every few time steps, with the assumption that the
-their values don't change much from one time-step to the other. Usually the update frequency chosen
-is every 5 steps, which considerably speeds up simulations and has been empirically shown to produce
-very similar results to having it be updated at every time step. We can achieve this by using the
-`schedule` keyword argument such as:
+its values don't change much from one time-step to the other. While other frequencies are possible,
+the only update frequency we know to be used in the literature is every 5 steps (e.g. Bou-Zeid et
+al.  2005; Chen et al. 2016; Salesky et al. 2017; Chor et al 2021), which Bou-Zeid et al. (2005)
+found to considerably speed up simulations while still producing very similar results to an update
+frequency of every time step. We can achieve this by using the `schedule` keyword argument such as:
 
 ```jldoctest
 julia> using Oceananigans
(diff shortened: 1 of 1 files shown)


src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -42,6 +42,54 @@ in the `x` and `y` directions).
...
+Smagorinsky closure with
+├── coefficient = DynamicCoefficient(averaging = (1, 2), schedule = IterationInterval(5, 0))
+└── Pr = 1.0
+```
+
+References
+==========
+
+Bou-Zeid, Elie, Charles Meneveau, and Marc Parlange. "A scale-dependent Lagrangian dynamic model for
+large eddy simulation of complex turbulent flows." Physics of fluids 17.2 (2005).
```
navidcy, line 92:
```suggestion
Bou-Zeid, Elie, Meneveau, Charles, and Parlange, Marc. (2005) A scale-dependent Lagrangian dynamic model for
large eddy simulation of complex turbulent flows, **Physics of Fluids_, **17**, 025105.
```

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -42,6 +42,54 @@ in the `x` and `y` directions).
...
+
+julia> dynamic_smagorinsky = Smagorinsky(coefficient=dynamic_coeff)
+Smagorinsky closure with
+├── coefficient = DynamicCoefficient(averaging = (1, 2), schedule = IterationInterval(1, 0))
+└── Pr = 1.0
+```
+
+The dynamic Smagorinsky that was just created above has its dynamic coefficient recalculated at every time
+step. While this will provide the highest level of accuracy, it will almost certainly be very slow
+given the high cost of calculating `DynamicCoefficient`s. Because of that, it is common in the
```
navidcy, line 66:
```suggestion
The dynamic Smagorinsky above has its dynamic coefficient recalculated at every time
step. While this provides the highest level of accuracy, it also brings about slowdown
given the high cost of calculating `DynamicCoefficient`s. Because of this slowdown, it is common in the
```

navidcy (approved):

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
tomchor:
```suggestion
The dynamic Smagorinsky above has its dynamic coefficient recalculated at every time
step. While this provides the highest level of accuracy, it will almost certainly be very slow
given the high cost of calculating `DynamicCoefficient`s. Because of this slowdown, it is common in the
```

navidcy: Sounds good!

tomchor:
I keep getting out of GPU memory errors in some doctests:

image

I've been these errors a lot lately actually. @navidcy can you help?

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -42,6 +42,54 @@ in the `x` and `y` directions).
...
+Smagorinsky closure with
+├── coefficient = DynamicCoefficient(averaging = (1, 2), schedule = IterationInterval(1, 0))
+└── Pr = 1.0
+```
+
+The dynamic Smagorinsky above has its dynamic coefficient recalculated at every time
+step. While this provides the highest level of accuracy, it will almost certainly be very slow
+given the high cost of calculating `DynamicCoefficient`s. Because of this slowdown, it is common in the
+literature to recalculate the coefficient only every few time steps, with the assumption that the
+their values don't change much from one time-step to the other. Usually the update frequency chosen
```
glwagner, line 68: If we say "usually" then we need citation. Eg wikipedia might say [citation needed]. I think a better practice is to give a range so we don't multiply the "collective intellectual debt" that always choosing the same number is having on this community.

glwagner: I'd say try strides of 2-8 or something.

navidcy:
> I keep getting out of GPU memory errors in some doctests:
> 
> image
> 
> I've been these errors a lot lately actually. @navidcy can you help?

I’m on the same ship. I have no idea.. I was wondering about these newly appeared errors in a few PRs…

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
tomchor: I can definitely add some citations. The thing is that I have never seen any other update frequency used in any paper. If you have info on papers that do it differently, let me know and we can give a range. Otherwise, the only number that has been used (afaik) is 5.

And again, I understand the rationale behind going by factors of 2, but it misses the one number that has been widely tested haha.

So please send some papers with different update frequencies my way, but otherwise I'd address your comment by keeping the text the same and adding references so that users can follow up and take their own conclusions.

tomchor: Added citations and made an accurate assessment if the current status: namely that the only values used so far in the literature is every 5 time steps but other values are possible. This way the user can decide what to do by themselves.

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -42,6 +42,66 @@ in the `x` and `y` directions).
...
+```
+
+The dynamic Smagorinsky above has its dynamic coefficient recalculated at every time step. While
+this provides the highest level of accuracy, it will almost certainly be very slow given the high
+cost of calculating `DynamicCoefficient`s. Because of this slowdown, it is standard practice in the
+literature to recalculate the coefficient only every few time steps, with the assumption that the
+its values don't change much from one time-step to the other. While other frequencies are possible,
+the only update frequency we know to be used in the literature is every 5 steps (e.g. Bou-Zeid et
+al.  2005; Chen et al. 2016; Salesky et al. 2017; Chor et al 2021), which Bou-Zeid et al. (2005)
+found to considerably speed up simulations while still producing very similar results to an update
```
navidcy, line 71:
```suggestion
its values don't change much from one time-step to the other. While any frequency choice is possible,
all studies seem to recalculate every 5 steps (e.g., Bou-Zeid et al. 2005; Chen et al. 2016; Salesky et al. 2017;
Chor et al 2021). This choice seems to stemming from the results by Bou-Zeid et al. (2005) who
found that considerably speed up simulations while still producing very similar results to an update
```

navidcy: slight rephrase, but feel free to ignore!

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -42,6 +42,66 @@ in the `x` and `y` directions).
...
+└── Pr = 1.0
+```
+
+The dynamic Smagorinsky above has its dynamic coefficient recalculated at every time step. While
+this provides the highest level of accuracy, it will almost certainly be very slow given the high
+cost of calculating `DynamicCoefficient`s. Because of this slowdown, it is standard practice in the
+literature to recalculate the coefficient only every few time steps, with the assumption that the
+its values don't change much from one time-step to the other. While any frequency choice is possible,
+all studies seem to recalculate every 5 steps (e.g., Bou-Zeid et al. 2005; Chen et al. 2016; Salesky et al. 2017;
+Chor et al 2021). This choice seems to stemming from the results by Bou-Zeid et al. (2005) who
```
tomchor, line 70:
```suggestion
Chor et al 2021). This choice seems to stem from the results by Bou-Zeid et al. (2005) who
```

navidcy:
I see you removed "tradition". :)

tomchor:
> I see you removed "tradition". :)

Wait, did I? I just accepted your suggestion and then fixed the "stemmed" typo! In any case I'm happy with the result :)

navidcy:
it's ok!

navidcy:
I tried to address the "collective intellectual debt" that @glwagner was referring to by using an example with 4 instead of 5.

Unfortunately the CI restarted.....

tomchor:
> I tried to address the "collective intellectual debt" that @glwagner was referring to by using an example with 4 instead of 5.
> 
> Unfortunately the CI restarted.....

Ah, that's a good idea!

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
glwagner: Just a remark that I think its misleading to refer the difference between a stride of 1 as yielding the "highest accuracy". I think a stride of 1 is simply an implementation of the scheme as written out mathematically, so apart from the discretization, there is no approximation or notion of "accuracy". When we use a stride different from 1 (eg 2), only then is the notion of "accuracy" introduced.

I think this can be clarified through a simplification of the text, something like: "to alleviate the high computational cost of the dynamic coefficient calculation, users may introduce an approximation wherein the dynamic coefficient is recomputed only every so often. This approximation is valid when the coefficient changes relatively slowly compared to a single time-step."

tomchor: Reworded everything based on @glwagner comments.

PR #4081, opened 2025-02-06
Bump the compat on GPUArrays to 11.2.2

### LINKED ISSUE #4036: Scalar indexing with CUDA 5.6.0

glwagner:
Closes #4036


Project.toml
```diff
@@ -56,7 +56,7 @@ Distances = "0.10"
 DocStringExtensions = "0.8, 0.9"
 Enzyme = "0.13.14"
 FFTW = "1"
-GPUArrays = "10.3"
+GPUArrays = "11.2.2"
```
giordano, line 59: If you don't need to make changes here to adapt to other breaking changes in GPUArrays maybe you can allow them both?  Like
```suggestion
GPUArrays = "10.3, 11.2.2"
```

navidcy:
Or should we drop the dependency completely as it was before? We added the dependency so we'd be able to pin it to a particular version if I recall correctly.

Project.toml
glwagner: good point

glwagner:
> Or should we drop the dependency completely as it was before? We added the dependency so we'd be able to pin it to a particular version if I recall correctly.

hehh yes if there is a CUDA version we can compat, then we don't need GPUArrays / GPUArrays compat

giordano:
Indeed, it seems to be used only in the tests: https://github.com/search?q=repo%3ACliMA%2FOceananigans.jl+GPUArrays+language%3AJulia&type=code&l=Julia, not in the package itself.

glwagner:
Probably next patch for CUDA.jl we can target. Right now compat is at 11.2.1

https://github.com/JuliaGPU/CUDA.jl/blob/4d85f27cb249efc311e79b6eeb55b3768481a5f3/Project.toml#L62

navidcy (approved):

simone-silvestri (approved):

giordano:
Anything holding this up? :eyes:

glwagner:
no!

Issue #4077, opened 2025-02-03
Bugs when indexing a partly in memory `FieldTimeSeries`


loisbaker:
I've come across a few bugs when reading in a FieldTimeSeries from disk with a partly in memory backend. I'll show them here with a MWE and comment below some suggestions for fixing them, I'm happy to make a PR if the changes seem sensible but haven't contributed before so will start here!

```julia
using Oceananigans
using Oceananigans.Utils: Time 

# First we generate some test data and write to disk
grid = RectilinearGrid(size=(2, 2, 2), extent=(1, 1, 1))
times = 0:0.1:3
filename = "MWE_data_file.jld2"
f_tmp = Field{Center,Center,Center}(grid) 
f = FieldTimeSeries{Center, Center, Center}(grid, times; backend = OnDisk(), path = filename, name = "f")

for (it, time) in enumerate(f.times)
    set!(f_tmp, (x, y, z) -> sin(2* π * time /3) )
    set!(f,f_tmp,it)
end

# Now we load the FTS partly in memory
N_in_mem = 5
f_fts = FieldTimeSeries(filename, "f"; backend = InMemory(N_in_mem))
Nt = length(f_fts.times)
```

**Bug 1:**
We can't access elements from `Nt - N_in_mem + 2` to `Nt` unless we first access a previous element that loads it in memory. 

Code:
```julia
function try_accessing_element(FTS,n)
    println("Trying to access element $n")
    try
        println(FTS[n])
    catch e
        println(e)
    end
end

try_accessing_element(f_fts,Nt-N_in_mem +2 ) 
try_accessing_element(f_fts,Nt-N_in_mem +1 ) 
try_accessing_element(f_fts,Nt-N_in_mem +2 ) 
```
Output:
```julia
Trying to access element 28
BoundsError(0.0:0.1:3.0, (32,))
Trying to access element 27
2×2×2 Field{Center, Center, Center} on RectilinearGrid on CPU
├── grid: 2×2×2 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 2×2×2 halo
├── boundary conditions: Nothing 
└── data: 6×6×6 OffsetArray(view(::Array{Float64, 4}, :, :, :, 1), -1:4, -1:4, -1:4) with eltype Float64 with indices -1:4×-1:4×-1:4
    └── max=-0.743145, min=-0.743145, mean=-0.743145
Trying to access element 28
[...]

PR #4086, opened 2025-02-09
Fix some `FieldTimeSeries` bugs

simone-silvestri:
this PR is a tentative solution to #4077 
and changes the default time indexing setting to `Clamp`

This MWE from #4077 
```julia
# First we generate some test data and write to disk
grid = RectilinearGrid(size=(2, 2, 2), extent=(1, 1, 1))
times = 0:0.1:3
filename = "MWE_data_file.jld2"
f_tmp = Field{Center,Center,Center}(grid) 
f = FieldTimeSeries{Center, Center, Center}(grid, times; backend = OnDisk(), path = filename, name = "f")

for (it, time) in enumerate(f.times)
    set!(f_tmp, (x, y, z) -> sin(2* π * time /3) )
    set!(f,f_tmp,it)
end

# Now we load the FTS partly in memory
N_in_mem = 5
f_fts  = FieldTimeSeries(filename, "f"; backend = InMemory(N_in_mem))
f_fts2 = FieldTimeSeries(filename, "f"; backend = InMemory(N_in_mem), time_indexing = Cyclical())
f_fts3 = FieldTimeSeries(filename, "f"; backend = InMemory(N_in_mem), time_indexing = Clamp())

s1 = []
s2 = []
s3 = []

for i in 0:0.01:10
    push!(s1,  f_fts[Time(i)][1, 1, 1])
    push!(s2, f_fts2[Time(i)][1, 1, 1])
    push!(s3, f_fts3[Time(i)][1, 1, 1])
end
```
yields
Screenshot 2025-02-09 at 12 53 41 PM

Comments are welcome

TODO:
- [x] add interpolation and indexing tests


loisbaker:
Thanks for looking into this! I'm not sure it totally solves the issue as I'm still not getting the right value at `f_fts[Time(0.41)]` for example. (It's now equal to `f_fts[Time(0.4)]`. I think that the backend is still changing from `InMemory(6,5)` to `InMemory(5,5)` in-between the lines (line 147 in `field_time_series_indexing.jl`)
```julia
# Otherwise, make a Field representing a linear interpolation in time
# Make sure both n₁ and n₂ are in memory by first retrieving n₂ and then n₁
ψ₂ = fts[n₂]
ψ₁ = fts[n₁]
```
since the first is still updating the backend so that `n₂` is the first index. Would it make more sense for `n₂ -1` to be the first index?

simone-silvestri:
> Thanks for looking into this! I'm not sure it totally solves the issue as I'm still not getting the right value at `f_fts[Time(0.41)]` for example. (It's now equal to `f_fts[Time(0.4)]`. I think that the backend is still changing from `InMemory(6,5)` to `InMemory(5,5)` in-between the lines (line 147 in `field_time_series_indexing.jl`)

Good catch, I had forgot an `update_field_time_series!` line!

simone-silvestri:
@loisbaker does this branch fix your issue?

loisbaker:
> @loisbaker does this branch fix your issue?

Thanks! I believe it fixes the two issues I mentioned with linear interpolation. I also mentioned a third small issue that occurred for me during some `FieldTimeSeries` interpolations associated with the operator `≈`, but would it be best if I moved that to a new issue?

glwagner:
> > @loisbaker does this branch fix your issue?
> 
> Thanks! I believe it fixes the two issues I mentioned with linear interpolation. I also mentioned a third small issue that occurred for me during some `FieldTimeSeries` interpolations associated with the operator `≈`, but would it be best if I moved that to a new issue?

Yes that'd be great, thank you!

simone-silvestri:
Added a test, should be ready to merge when tests pass

simone-silvestri:
I am not sure why I am getting these reactant errors in this PR since it does not modify the extension and everything seems to work in main:
https://github.com/CliMA/Oceananigans.jl/actions/runs/13364446847/job/37319424452?pr=4086

@glwagner have you encountered this?

simone-silvestri:
The tests on this PR fail with 
```
[2025/02/21 04:03:03.692] ERROR Error during loading of extension OceananigansReactantExt of Oceananigans, use `Base.retry_load_extensions()` to retry. -@-> loading.jl:1371
 ```

Anyone knows what the problem is?

glwagner (approved):

PR #4096, opened 2025-02-12
Set a default Clock depending on the grid

glwagner:
This PR adds a constructor for `Clock` which depends on `grid`, and therefore architecture. This allows the clock to be traced when using `arch = ReactantState()`.

In addition to that, this PR discontinues support for `run!` with `arch = ReactantState()`. Our rationale is that we cannot yet support the `Simulation` paradigm, which executes time-steps in a `while` loop until `!(simulation.running)`, where `simulation.running` is typically controlled by `if` statements that depend, for example, on `clock.iteration`. Basically we're deciding not to try to use `@trace if`, yet.

As a result, time step loops with `Reactant` should be managed manually. To facilitate this, this PR also adds a function `first_time_step!` which first `initialize!(model)` and `update_state!(model)` before taking a time-step (which is limited to Euler step for AB2).

cc @giordano


ext/OceananigansReactantExt/TimeSteppers.jl
```diff
@@ -0,0 +1,24 @@
...
+const ReactantGrid{FT, TX, TY, TZ} = AbstractGrid{FT, TX, TY, TZ, <:ReactantState} where {FT, TX, TY, TZ}
+
+function Clock(grid::ReactantGrid)
+    FT = Float64 # may change in the future
+    t = ConcreteRNumber(zero(FT))
+    iter = ConcreteRNumber(0)
+    stage = ConcreteRNumber(0)
+    last_Δt = ConcreteRNumber(zero(FT))
+    last_stage_Δt = ConcreteRNumber(zero(FT))
+    return Clock(; time=t, iteration=iter, stage, last_Δt, last_stage_Δt)
```
glwagner, line 20: @giordano see the implementation here

giordano: Uhm, that's already what I had (well, what you told me to do) in https://github.com/PRONTOLab/GB-25/pull/12/files#r1953572093.  So this PR simply allows us to skip creating the clock manually?

glwagner: yes! but if there is another feature you think we should put in here I am happy to tackle that

glwagner: or if there is something we should change

glwagner:
I believe this is our error now:

```julia
Reactant Super Simple Simulation Tests: Error During Test at /Users/gregorywagner/Projects/dev/OC1.jl/test/test_reactant.jl:154
  MethodError: no method matching rem(::Reactant.TracedRNumber{Int64}, ::Int64)

  Closest candidates are:
    rem(::Any, ::Any, ::RoundingMode{:FromZero})
     @ Base div.jl:103
    rem(::Any, ::Any, ::RoundingMode{:Nearest})
     @ Base div.jl:100
    rem(::Any, ::Any, ::RoundingMode{:Up})
     @ Base div.jl:99
    ...

  Stacktrace:
    [1] macro expansion
      @ ~/.julia/packages/Reactant/qEA3Y/src/utils.jl:0 [inlined]
    [2] call_with_reactant(::typeof(rem), ::Reactant.TracedRNumber{Int64}, ::Int64)
      @ Reactant ~/.julia/packages/Reactant/qEA3Y/src/utils.jl:767
    [3] IterationInterval
      @ ~/Projects/dev/OC1.jl/src/Utils/schedules.jl:109 [inlined]
```

@wsmoses can we add support for `rem`? If not I think there may be other workarounds.

glwagner:
@giordano @wsmoses I'm hitting a road block and trying to figure out what is going wrong. I put together a very simple prototype of a function that iterates until a stopping criteria:

```julia
using Reactant

mutable struct TestClock{I}
    iteration :: I
end

mutable struct Stopper{I}
    iteration :: I
end

function iterate_until!(clock, stopper)
    iterate = true
    while iterate
        @trace if clock.iteration >= stopper.iteration
            iterate = false
        else
            clock.iteration += 1
        end
    end
    return nothing
end

clock = TestClock(ConcreteRNumber(0))
#stopper = Stopper(ConcreteRNumber(2))
stopper = Stopper(2)

r_iterate! = @compile sync=true iterate_until!(clock, stopper)
r_iterate!(clock, stopper)
@show clock.iteration
```

This fails with 

```julia
ERROR: LoadError: TypeError: non-boolean (Reactant.TracedRNumber{Bool}) used in boolean context
Stacktrace:
  [1] macro expansion
    @ ~/.julia/packages/ReactantCore/t40zc/src/ReactantCore.jl:378 [inlined]
  [2] iterate_until!
    @ ~/Projects/Oceananigans.jl/test/test_reactant.jl:31 [inlined]
  [3] iterate_until!(none::TestClock{Reactant.TracedRNumber{Int64}}, none::Stopper{Int64})
    @ Reactant ./<missing>:0
  [4] iterate_until!
    @ ~/Projects/Oceananigans.jl/test/test_reactant.jl:30 [inlined]
```

where line 31 is `@trace if clock.iteration >= stopper.iteration`.

I'm also wondering whether the `stopper.iteration` needs to be traced, or not? 

I think getting this prototype to work will show me how to get this PR merged

glwagner:
Couple issues open to work on stuff that's come up:

* https://github.com/EnzymeAD/Reactant.jl/issues/793
* https://github.com/EnzymeAD/Reactant.jl/issues/792

glwagner:
@giordano I think we've made progress on this and now `run!` actually "starts to" compile. However, it takes a long time. In fact it is still going

image

glwagner:
Note running the tests requires adding `TestEnv` to your global environment and then

```julia
using TestEnv; TestEnv.activate()
using Pkg; Pkg.add("Reactant", rev="glw-wm/traced-set-path")
include("test/test_reactant.jl")
```

glwagner:
This is ready and tests will hopefully pass soon @wsmoses @simone-silvestri @giordano

simone-silvestri (approved):

ext/OceananigansReactantExt/Simulations/run.jl
```diff
@@ -0,0 +1,19 @@
+import ..TimeSteppers: first_time_step!
+
+""" Step `sim`ulation forward by one time step. """
+time_step!(sim::ReactantSimulation) = time_step!(sim.model, Δt; euler)
```
giordano, line 4: There's no `euler` defined here.

PR #4098, opened 2025-02-13
Precision error when converting to 0 - 360

simone-silvestri:
With the recent change in interpolation in #4074, we are hitting an error in ClimaOcean.jl (see https://github.com/CliMA/ClimaOcean.jl/actions/runs/13301013435/job/37142339229).

This is because the convert function
https://github.com/CliMA/Oceananigans.jl/blob/399695072641a53cfa02f8d9bc34c5f2ad54dffa/src/Fields/interpolate.jl#L84 is susceptible to precision errors when the input is small enough:
```julia
julia> convert_to_0_360(-1e-12)
359.999999999999

julia> convert_to_0_360(-1e-13)
359.9999999999999

julia> convert_to_0_360(-1e-14)
0.0
```
This is not acceptable because a negative number should always be on the "left" side of the zero corresponding to values between 359 and 360, i.e. moving to zero from the left side should result in 360 not 0.
In ClimaOcean we were hitting this error because the bathymetry we interpolate from starts at `-1.4e-14` instead of 0, exposing this issue. 
This PR fixes this problem by making sure we are on the correct side of the zero when converting to the 0 - 360 range

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.95.10"
+version = "0.95.11"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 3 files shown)


glwagner (approved):

glwagner:
I'm constantly wondering whether we are doing the right thing about precision! More so for timestepping though.

PR #4108, opened 2025-02-17
Implement RotatedLatitudeLongitudeGrid

glwagner:
This PR implements a `RotatedLatitudeLongitudeGrid`, which has the structure of a `LatitudeLongitudeGrid` but with a displaced pole.

There are two ways to implement this. One is to make a LatitudeLongitudeGrid, change the latitudes, and then write the data down to `OrthogonalSphericalShellGrid`. The pole displacement is then encoded in the conformal mapping of the grid. Another is to actually embed the coordinate shift into `φnode`. I think the former is probably a simpler approach. The latter saves a little bit of memory though.

This PR moves some files around and implements the basic abstraction. Next we have to write the metrics into `OrthogonalSphericalShellGrid` and add tests.

cc @simone-silvestri

### DIFF
--- a/src/OrthogonalSphericalShellGrids/rotated_latitude_longitude_grid.jl
+++ b/src/OrthogonalSphericalShellGrids/rotated_latitude_longitude_grid.jl
@@ -19,6 +19,7 @@ const RotatedLatitudeLongitudeGrid = OrthogonalSphericalShellGrid{<:Any,
 @inline lat_lon_metric(m, i, j) = @inbounds m[i, j]
 @inline lat_lon_metric(m::AbstractVector, i, j) = @inbounds m[j]
 @inline lat_lon_metric(m::Number, j) = m
+@inline lat_lon_metric(m::Number, i, j) = m
 
 function latitude_longitude_shift((λ₀, φ₀))
     Δλ = - λ₀


glwagner:
Also the name is `RotatedLatitudeLongitudeGrid`. The inputs are the latitude and longitude in "rotated coordinates". This is the simplest API, but it does require users to think a bit to set up their grid. The other input is `north_pole` which specifies the lat-lon of the "grid pole". Then we compute a rotation between the pole of the coordinate system, defined at (0, 90), and the grid north pole to change the metrics.

As an example of how this works, we make a grid spanning the equator and prime meridian (60 deg in both directions), but then rotate the north pole to (180, 0):

```julia
rotated_lat_lon_grid = RotatedLatitudeLongitudeGrid(size = (Nx, Ny, Nz),
                                                    latitude = (-60, 60),
                                                    longitude = (-60, 60),
                                                    north_pole = (180, 0),
                                                    z = (-1000, 0),
```

This moves the grid over the north pole (longitude on left and latitude on right):

image

`north_pole = (0, 0)` rotated the grid over the south pole:

image

I'm still scratching my head whether the longitude is correct.

navidcy:
Does it rotate only the North Pole? Or both Poles?

glwagner:
> Does it rotate only the North Pole? Or both Poles?

Not sure I totally understand but this is a rigid rotation, not a rotation of just one pole. The grid metrics are identical / unchanged, only the coordinates are rotated.

glwagner:
> > > Does it rotate only the North Pole? Or both Poles?
> > 
> > 
> > Not sure I totally understand but this is a rigid rotation, not a rotation of just one pole. The grid metrics are identical / unchanged, only the coordinates are rotated.
> 
> Ok that’s what I thought so. Just from the kwarg it wasn’t clear. The kwarg only mentions north pole. That’s all. All clear now.

True but we have to pick one, because we can't rotate both poles independently.

glwagner:
test works

Uploading splash.mp4…

PR #4112, opened 2025-02-19
(0.95.13) Use active-weighted rather than naive reconstruction for f- and beta-plane Coriolis terms

### LINKED ISSUE #4111: Issue with Noise In Vorticity on Cartesian Grid with Immersed Bottom Boundary
The following is a very simple setup that exhibits grid-scale oscillations in velocity, especially close to the Gaussian ridge specified at the bottom.

```julia
using Oceananigans
using Oceananigans.Units
using CairoMakie

Lx = 2000.0kilometers
Ly = 1000.0kilometers
Lz = 4.0kilometers

Δx = 25.0kilometers
n = 30

filename = "testbed_just_water_velocity_WENO"

grid = RectilinearGrid(size=(80, 40, n), 
                            x=(0, Lx), y=(0, Ly), z=(-Lz, 0),
                            halo=(4,4,4),
                            topology=(Periodic, Bounded, Bounded))

ridge = @inline (x, y) -> -Lz + 1000.0meters * exp(-((x - Lx / 2)^2) / (100000kilometers^2))

ridge_grid = ImmersedBoundaryGrid(grid, GridFittedBottom(ridge))

coriolis = FPlane(f=-1e-4)

model = HydrostaticFreeSurfaceModel(;grid=ridge_grid, 
                                    coriolis,
                                    momentum_advection=WENO(order=5))

function uᵢ(x, y, z)
    return 0.1 * exp(z/500.0)
end

set!(model, u=uᵢ)

simulation = Simulation(model, Δt=20minutes, stop_time=3 * 30days)

#region Callbacks and Checkpointing
using Printf

wall_clock = Ref(time_ns())

function progress(sim)
    elapsed = 1e-9 * (time_ns() - wall_clock[])

    msg = @sprintf("iteration: %d, time: %s, wall time: %s, max|u|: %6.3e, m s⁻¹, max|v|: %6.3e, m s⁻¹, max|w|: %6.3e, m s⁻¹\n",
                   iteration(sim), prettytime(sim), prettytime(elapsed),
                   maximum(abs, sim.model.velocities.u),
                   maximum(abs, sim.model.velocities.v),
                   maximum(abs, sim.model.velocities.w))

    wall_clock[] = time_ns()

    @info msg

    return nothing
end

add_callback!(simulation, progress, name=:progress, IterationInterval(200))

# Write checkpoint files every year
checkpoint_time = 10days
simulation.output_writers[:checkpointer] = Checkpointer(model;
                                                        schedule = TimeInterval(checkpoint_time),
                                                        prefix = filename * "_checkpoint",
                                                        overwrite_existing = true)
#endregion

##

run!(simulation)
```

glwagner:
Resolves #4111. Thanks for the bugfind @kenflat2!

Basically this implements boundary-aware reconstruction for Coriolis terms. Reconstruction is required for Coriolis terms on a staggered grid. Previously, reconstruction was not boundary-aware.

Note, I am not sure about `ConstantCartesianCoriolis` or `NonTraditionalBetaPlane`. Probably there is some work to do there, but reconstruction is different for those so I left it out of this PR.

Here's my updated script:

```julia
using Oceananigans
using Oceananigans.Units
using GLMakie
using Printf

Lx = 2000kilometers
Ly = 1000kilometers
Lz = 4kilometers

grid = RectilinearGrid(size = (128, 64, 64),
                       x = (0, Lx),
                       y = (0, Ly),
                       z = (-Lz, 0),
                       halo = (4, 4, 4),
                       topology = (Periodic, Bounded, Bounded))

ridge(x, y) = -Lz + 1000meters * exp(-((x - Lx / 2)^2) / (100000kilometers^2))
ridge_grid = ImmersedBoundaryGrid(grid, GridFittedBottom(ridge))

coriolis = FPlane(f=-1e-4)
model = HydrostaticFreeSurfaceModel(; grid = ridge_grid, coriolis, momentum_advection = WENO(order=5))

uᵢ(x, y, z) = 0.1 * exp(z / 500)
set!(model, u=uᵢ)
simulation = Simulation(model, Δt=20minutes, stop_iteration=100)

wall_clock = Ref(time_ns())

function progress(sim)
    elapsed = 1e-9 * (time_ns() - wall_clock[])

    msg = @sprintf("iteration: %d, time: %s, wall time: %s, max|u|: (%6.3e, %6.3e, %6.3e), m s⁻¹\n",
                   iteration(sim), prettytime(sim), prettytime(elapsed),
                   maximum(abs, sim.model.velocities.u),
                   maximum(abs, sim.model.velocities.v),
                   maximum(abs, sim.model.velocities.w))

    wall_clock[] = time_ns()
    @info msg

    return nothing
end

add_callback!(simulation, progress, IterationInterval(10))

run!(simulation)

u, v, w = model.velocities
heatmap(view(v, :, 32, :))
current_figure()
```

# on `main`

image

# this PR

image

I'm not sure how to test this, ideas welcome.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -12,7 +12,7 @@ if test_file != :none
 end
 
 #####
-##### Run tests!
+##### Run tests
 #####
 
 CUDA.allowscalar() do
@@ -27,21 +27,7 @@ CUDA.allowscalar() do
 
     # Initialization steps
     if group == :init || group == :all
-        using Reactant
-        using Enzyme
-
-        Pkg.instantiate(; verbose=true)
-        Pkg.precompile(; strict=true)
-        Pkg.status()
-
-        try
-            MPI.versioninfo()
-        catch; end
-
-        try
-            CUDA.precompile_runtime()
-            CUDA.versioninfo()
-        catch; end
+        include("test_init.jl")
     end
     
     # Core Oceananigans
--- a/test/test_init.jl
+++ b/test/test_init.jl
@@ -1,0 +1,15 @@
+using Reactant
+using Enzyme
+Pkg.instantiate(; verbose=true)
+Pkg.precompile(; strict=true)
+Pkg.status()
+
+try
+    MPI.versioninfo()
+catch; end
+
+try
+    CUDA.precompile_runtime()
+    CUDA.versioninfo()
+catch; end
+


src/Coriolis/Coriolis.jl
```diff
@@ -24,6 +25,27 @@ Abstract supertype for parameters related to background rotation rates.
 """
 abstract type AbstractRotation end
 
+@inline active_weighted_ℑxyᶜᶠᵃ(i, j, k, grid, q, args...) = ℑxyᶜᶠᵃ(i, j, k, grid, q, args...)
+@inline active_weighted_ℑxyᶠᶜᵃ(i, j, k, grid, q, args...) = ℑxyᶜᶠᵃ(i, j, k, grid, q, args...)
```
simone-silvestri, line 29: I think we need the same code for immersed and non-immersed boundaries.

simone-silvestri:
Now that the scope for `active_cell` interpolation has expanded out of use for a specific type (`ActiveCellEnstrophyConserving`), we should probably move these functions in `Operators` (and add some tests) so that we can use them for other methods that require interpolation across immersed boundaries

src/Coriolis/Coriolis.jl
simone-silvestri: seemed to make the difference for a coarser test cast (80x40x30)

simone-silvestri:
Looks like masking inactive nodes is always the numerically correct implementation and it is not a user choice, so I think we can remove `ActiveCellEnsthrophyScheme` and leave `EnstrophyScheme` where we mask inactive cells.

glwagner:
@simone-silvestri feel free to make these changes

glwagner:
> @simone-silvestri feel free to make these changes

Nevermind I did it.

glwagner:
> Now that the scope for `active_cell` interpolation has expanded out of use for a specific type (`ActiveCellEnstrophyConserving`), we should probably move these functions in `Operators` (and add some tests) so that we can use them for other methods that require interpolation across immersed boundaries

Not sure what you have in mind for test or something more general but feel free to add it.

Note that we've received feedback that metaprogramming makes it hard to understand Oceananigans code/errors and slows down development. So I think we should avoid it if we don't need it. Not sure how many of the "active-weighted" operators are needed.

simone-silvestri:
I was thinking to use it for sea ice, so I think we can add only the horizontal interpolations for now and add other ones as we see fit

navidcy:
Hm... I can't understand neither from this PR nor from the issue why Coriolis needs to be reconstructed.

glwagner:
> Hm... I can't understand neither from this PR nor from the issue why Coriolis needs to be reconstructed.

The reconstruction is required because of staggering. For example computing `f * v` at `f, c, c` requires a 4-point stencil.

This PR does not change whether we reconstruct or not. Instead, it changes _how_ we reconstruct by omitting inactive cells from the reconstruction.

As for _why_, the evidence is given by the empirical test. I'm not offering a theoretical explanation which would require more work...

navidcy:
Oh ok. I get it now. I wasn’t asking for deeper why.

glwagner:
> Oh ok. I get it now. I wasn’t asking for deeper why.

updated the top-level description

Project.toml
```diff
@@ -74,7 +74,7 @@ OffsetArrays = "1.4"
 OrderedCollections = "1.1"
 Printf = "1.9"
 Random = "1.9"
-Reactant = "0.2.25"
+Reactant = "0.2.31"
```
glwagner, line 77:
```suggestion
Reactant = "0.2.25"
```

simone-silvestri (approved):

PR #4093, opened 2025-02-11
Make the `OceananigansReactantExt` depend explicitly on KA

glwagner:
I suspect that with this change, our distributed tests will fail (possibly, because of a bug with julia 1.10.5 that is used on the caltech cluster, cc @vchuravy). 

This reinstates https://github.com/CliMA/Oceananigans.jl/pull/4085, which we reverted on #4082.

cc @giordano

PR #4115, opened 2025-02-20
Bump to 0.95.13

glwagner:
Captures https://github.com/CliMA/Oceananigans.jl/pull/4093


giordano:
Uhm, I think last tag is 0.95.11, version was already bumped in https://github.com/CliMA/Oceananigans.jl/pull/4093

glwagner:
doh

PR #4116, opened 2025-02-21
Add `finalize!` for callbacks

glwagner:
Adds `finalize!(callback, sim)`. By default, it does nothing. This gets called after the time-stepping loop is done. It only gets called if using `run!`.

Also actually calls `initialize!` for callbacks which previously was not called.

### DIFF
--- a/src/Simulations/callback.jl
+++ b/src/Simulations/callback.jl
@@ -21,6 +21,7 @@ Initialize `callback`. By default, this does nothing, but
 can be optionally specialized on the type parameters of `Callback`.
 """
 initialize!(callback::Callback, sim) = nothing
+finalize!(callback::Callback, sim) = nothing
 
 """
     Callback(func, schedule=IterationInterval(1);
--- a/src/Simulations/callback.jl
+++ b/src/Simulations/callback.jl
@@ -7,8 +7,8 @@ import Oceananigans: initialize!
 struct Callback{P, F, S, CS}
     func :: F
     schedule :: S
-    parameters :: P
     callsite :: CS
+    parameters :: P
 end
 
 @inline (callback::Callback)(sim) = callback.func(sim, callback.parameters)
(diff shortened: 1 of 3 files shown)


tomchor:
I've never used `initialize!()` before. With `MovieMaker` for example, the idea would be to define something like

```julia
finalize!(plotter::MovieMaker, sim) = save(plotter.filename, plotter.io)
```
?

glwagner:
> I've never used `initialize!()` before. With `MovieMaker` for example, the idea would be to define something like
> 
> ```julia
> finalize!(plotter::MovieMaker, sim) = save(plotter.filename, plotter.io)
> ```
> 
> ?

yes, could also write one last frame but not sure you want that

glwagner:
I guess nobody has used `initialize!` because it wasn't called 😂

glwagner:
> > I've never used `initialize!()` before. With `MovieMaker` for example, the idea would be to define something like
> > ```julia
> > finalize!(plotter::MovieMaker, sim) = save(plotter.filename, plotter.io)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> yes, could also write one last frame but not sure you want that

But actually, that wasn't the syntax of the original design here. But I like this idea. I'll change it to that.

tomchor:
As a side note, I think it would be pretty useful if this could be incorporated into `Checkpointer`, perhaps with an extra kwarg. In most, if not all, my scripts I manually save a last checkpoint after the simulation is done. I don't know how many others do this, but I can't be the only one! 😂

glwagner:
For sure, that's easy to extend the idea to `Checkpointer`. I think I'd rather do that in a subsequent PR though just to get this through

tomchor:
> For sure, that's easy to extend the idea to `Checkpointer`. I think I'd rather do that in a subsequent PR though just to get this through

For sure! We should probably add a couple of tests before merging though.

glwagner:
> > For sure, that's easy to extend the idea to `Checkpointer`. I think I'd rather do that in a subsequent PR though just to get this through
> 
> For sure! We should probably add a couple of tests before merging though.

Where is the faith!

Issue #4120, opened 2025-02-21
"KernelAbstractions" not found error and extensions possibly not working


tomchor:
I recently started seeing this error when `using Oceananigans`. Issuing `using Oceananigans` a second time seems to do the trick though:

```julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.10.2 (2024-03-01)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using Oceananigans
[ Info: Oceananigans will use 4 threads
ERROR: KeyError: key "KernelAbstractions" not found
Stacktrace:
  [1] getindex
    @ ./dict.jl:498 [inlined]
  [2] _insert_extension_triggers(parent::Base.PkgId, extensions::Dict{String, Any}, weakdeps::Dict{String, Any})
    @ Base ./loading.jl:1276
  [3] insert_extension_triggers(env::String, pkg::Base.PkgId)
    @ Base ./loading.jl:1219
  [4] insert_extension_triggers
    @ ./loading.jl:1204 [inlined]
  [5] __require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:1817
  [6] #invoke_in_world#3
    @ ./essentials.jl:926 [inlined]
  [7] invoke_in_world
    @ ./essentials.jl:923 [inlined]
  [8] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:1803
  [9] macro expansion
    @ ./loading.jl:1790 [inlined]
 [10] macro expansion
    @ ./lock.jl:267 [inlined]
 [11] __require(into::Module, mod::Symbol)
    @ Base ./loading.jl:1753
 [12] #invoke_in_world#3
    @ ./essentials.jl:926 [inlined]
 [13] invoke_in_world
    @ ./essentials.jl:923 [inlined]
 [14] require(into::Module, mod::Symbol)
    @ Base ./loading.jl:1746

julia> Oceananigans
ERROR: UndefVarError: `Oceananigans` not defined

julia> using Oceananigans

julia> Oceananigans
Oceananigans
```

I wonder if this is due to https://github.com/CliMA/Oceananigans.jl/pull/4093

[...]

PR #4121, opened 2025-02-22
Create `MovieMaker`

### LINKED ISSUE #4110: Should we a `Plotter` to Oceananigans?
This issue was originally posted in https://github.com/tomchor/Oceanostics.jl/issues/186.

The basic idea is that, in the early stages of assembling a simulation, we tend to spend a lot of time (relatively) coding the output writing and plotting/animating parts of the simulation. Most of that code, however, is pretty similar across simulations: setting up a writer with a couple of variables, then reading it, creating a figure, plotting heatmaps, and writing the animation loop. Lots of copy-paste. And then if we want to add other variables to the plot, several lines have to be added/modified to accommodate for that.

I think this process can be automated. As @glwagner suggested, perhaps the best way would be to avoid writing altogether and create a plotter which updates a figure periodically as the simulation is running.

Here's a quick snippet of a proof of concept I tried:

```julia
struct Plotter
    fig
    io
    func
end

function Plotter(func; fig=nothing, io=nothing, kwargs...)
    if fig isa Nothing
        fig = Figure()
        ax1 = Axis(fig[1, 1])
        ax2 = Axis(fig[1, 2])
    end
    if io isa Nothing
        io = VideoStream(fig, kwargs...)
    end
    return Plotter(fig, io, func)
end

(plotter::Plotter)(simulation) = plotter.func(simulation, plotter.fig, plotter.io)

function update_plot(sim, fig, io)
    heatmap!(fig[1, 1], ω; colormap = :balance, colorrange = (-2, 2))
    heatmap!(fig[1, 2], s; colormap = :balance, colorrange = (-2, 2))
    recordframe!(io)
end

plotter = Plotter(update_plot)
add_callback!(simulation, plotter, TimeInterval(0.6))
```

Applied to the 2D turbulence example in the docs this produces:

https://github.com/user-attachments/assets/6a6e2622-2421-42e9-90f6-561d5473c0f8

Is there interest in implementing something like this here? If so I'll close the issue on Oceanostics and open a PR here.

CC @glwagner @jbisits

tomchor:
Closes https://github.com/CliMA/Oceananigans.jl/issues/4110

cc @glwagner 

Note that at the moment some of the functionality is not yet there due to something weird that's going on with the extensions. See https://github.com/CliMA/Oceananigans.jl/issues/4120.


validation/plotters/2d_turbulence_movie.jl
```diff
@@ -0,0 +1,56 @@
...
+
+u, v, w = model.velocities
+ω = ∂x(v) - ∂y(u)
+s = sqrt(u^2 + v^2)
+
+fig = Figure()
+ax1 = Axis(fig[1, 1])
+ax2 = Axis(fig[1, 2])
+
+io = VideoStream(fig, format="mp4", framerate=12, compression=20)
```
glwagner, line 40: can we push this under the hood?

tomchor: That's the idea! But I'm still working around https://github.com/CliMA/Oceananigans.jl/issues/4120. I'm gonna try a newer Julia soon.

glwagner: ah yeah. there's no way to solve that except for updating julia.

ext/OceananigansMakieExt.jl
```diff
@@ -194,4 +196,25 @@ function convert_arguments(pl::Type{<:AbstractPlot}, ξ1::AbstractArray, ξ2::Ab
...
+Create a `MovieMaker`, which makes a movie.
+"""
+function MovieMaker(fig, func; io::VideoStream, filename="movie.mp4")
+    return MovieMaker(fig, func, io, filename)
+end
+
+function MovieMaker(fig, func; filename="movie.mp4", kwargs...)
+    io = VideoStream(fig; kwargs...)
+    return MovieMaker(fig, func, io, filename)
+end
```
tomchor, line 215: @glwagner I was wondering if you can help me out here, because I've been having trouble adding methods via the `MakieExt`. In particular, I always get the following warnings/errors regarding these lines:

```julia
WARNING: Method definition (::Type{Oceananigans.Diagnostics.MovieMaker{F, FN, IO, NM} where NM where IO where FN where F})(Any, Any) in module OceananigansMakieExt at /home/tomas/repos/Oceananigans.jl/ext/OceananigansMakieExt.jl:208 overwritten at /home/tomas/repos/Oceananigans.jl/ext/OceananigansMakieExt.jl:212.
ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
```

And then of course, only the latter of these two methods is actually included:

```julia
julia> methods(MovieMaker)
# 3 methods for type constructor:
 [1] MovieMaker(fig, func; filename, kwargs...)
     @ OceananigansMakieExt ~/repos/Oceananigans.jl/ext/OceananigansMakieExt.jl:212
 [2] MovieMaker()
     @ ~/repos/Oceananigans.jl/src/Diagnostics/plotter.jl:10
 [3] MovieMaker(fig::F, func::FN, io::IO, filename::NM) where {F, FN, IO, NM}
     @ ~/repos/Oceananigans.jl/src/Diagnostics/plotter.jl:4
```

I haven't been able to figure how to export both methods. Do you know what's going on?

PR #4124, opened 2025-02-24
Generalize GPU backend and add Metal extension

glwagner:
This is a WIP. With @simone-silvestri


glwagner:
This seems to work. Funny thing though, its not faster than CPU on my machine (which is Mac Studio). One interesting reason its nice to have this is that we can really test whether Float32 is truly enforced. There's probably a bit more work to do on Float32.

src/Advection/Advection.jl
```diff
@@ -56,6 +56,9 @@ abstract type AbstractUpwindBiasedAdvectionScheme{B, FT} <: AbstractAdvectionSch
 # Note that it is not possible to compile schemes for `advection_buffer = 41` or higher.
 const advection_buffers = [1, 2, 3, 4, 5, 6]
 
+# To add support for new floating types add them here
+const supported_float_types = [Float32, Float64]
+
```
simone-silvestri, line 61:
```suggestion
```
was superceded by #4053

simone-silvestri:
> This seems to work. Funny thing though, its not faster than CPU on my machine (which is Mac Studio). One interesting reason its nice to have this is that we can really test whether Float32 is truly enforced. There's probably a bit more work to do on Float32.

you must have a powerful CPU! On my mac M1, the metal version is about 3 times faster than the CPU version. It is true, that maybe threaded CPU will be faster...

simone-silvestri:
Also, it looks like the cost of launching small kernels is much higher in metal than in a normal GPU. If we do not use the `SplitExplicitFreeSurface` that launches about 150 kernels per time step in the `baroclinic_adjustment.jl` example (i.e., `free_surface=nothing`), on my Mac, the metal version is around 20 times faster than the corresponding CPU version.

Note that the cost on the CPU version between no free surface and split explicit free surface is roughly the same (i.e. the free surface is a negligible cost), while the Metal version has a factor of 8 difference between no free surface and split explicit free surface. 

Probably to have performance on Metal we need some Metal-specific optimizations

glwagner:
> Also, it looks like the cost of launching small kernels is much higher in metal than in a normal GPU. If we do not use the `SplitExplicitFreeSurface` that launches about 150 kernels per time step in the `baroclinic_adjustment.jl` example (i.e., `free_surface=nothing`), on my Mac, the metal version is around 20 times faster than the corresponding CPU version.
> 
> Note that the cost on the CPU version (as well as on a CUDA backend) is roughly the same for no free surface and split explicit free surface (i.e. the free surface is a negligible cost), while the Metal version has a factor of 8 difference in cost between no free surface and split explicit free surface.
> 
> Probably to have performance on Metal we need some Metal-specific optimizations

Could be this overhead is the cause of my demise

src/Grids/latitude_longitude_grid.jl
```diff
@@ -534,6 +530,8 @@ function allocate_metrics(grid::LatitudeLongitudeGrid)
         metric_size = (length(grid.Δλᶜᵃᵃ), length(grid.φᵃᶜᵃ))
     end
 
+    @show FT, metric_size, arch, offsets
+
```
simone-silvestri, line 534:
```suggestion
```

simone-silvestri:
There seem to be some issues with CATKE in Float32 (see https://github.com/CliMA/ClimaOcean.jl/issues/356), should we fix them here?

simone-silvestri:
> There seem to be some issues with CATKE in Float32 (see [CliMA/ClimaOcean.jl#356](https://github.com/CliMA/ClimaOcean.jl/issues/356)), should we fix them here?

Actually, the problem was a missing `Oceananigans.defaults.FloatType = FLoat32` at the beginning of the script

simone-silvestri:
Actually, they were just problems with the default in `OceanSimulations.jl` that was not using the correct floating type

.github/workflows/ci.yml
```diff
@@ -34,7 +34,68 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
         env:
           TEST_GROUP: "reactant"
+  metal:
+    name: Metal - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
```
giordano, line 70: I'm not sure this uses Metal at all, as far as I understand Github-hosted macos runners don't expose the GPU.

glwagner: Doh 🤦 

what are options for testing on Metal then? Either set up our own or convince JuliaGPU?

glwagner: I'll look into this

giordano: Asking Tim to get access to the buildkite runners they use for Metal.jl probably.

glwagner: another link https://github.com/JuliaGPU/Metal.jl/issues/22

PR #4136, opened 2025-02-25
Implement InterpolationIndices for time-interpolating FieldTimeSeries

glwagner:
This PR implements `InterpolationIndices` for time-interpolation of `FieldTimeSeries`. This is useful when interpolating many fields which all have the same interpolation indices (such as a prescribed atmosphere). In that case we need only compute the interpolation indices once, and pass them into a kernel that does the interpolation.

### DIFF
--- a/src/OutputReaders/field_time_series_indexing.jl
+++ b/src/OutputReaders/field_time_series_indexing.jl
@@ -4,11 +4,11 @@ using Oceananigans.Architectures: architecture
 
 import Oceananigans.Fields: interpolate
 
-struct InterpolationIndices{T}
+struct InterpolationIndices{T, N1, N2, N3}
     fractional_index :: T
-    first_index  :: Int
-    second_index :: Int
-    length       :: Int
+    first_index  :: N1
+    second_index :: N2
+    length       :: N3
 end
 
 #####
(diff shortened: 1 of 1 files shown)


glwagner:
cc @wsmoses @simone-silvestri

src/OutputReaders/field_time_series_indexing.jl
```diff
@@ -170,6 +177,18 @@ end
                              from_loc, from_grid, times, backend, time_indexing)
 
     to_time = to_time_index.time
+    ñ, n₁, n₂ = interpolating_time_indices(time_indexing, times, to_time)
+    interp = InterpolationIndices(ñ, n₁, n₂, length(times))
```
simone-silvestri, line 181: why do we need this inside the kernel?

glwagner: This does not change the kernel code. What we have done is reorganize the code so that we always use `InterpolationIndices`. However, there is a second way to interpolate which does not require using `interpolating_time_indices`, and instead constructing `InterpolationIndices` directly -- for example, outside the kernel.

glwagner: In other words with this new feature, this way of writing the code is _required_ to eliminate code duplication. Otherwise, we have to write `interpolate` twice.

simone-silvestri (approved):

glwagner:
Note, we may also be doing binary search for the _spatial_ interpolation which will require additional handling in a future PR.

Issue #4047, opened 2025-01-17
"Number of threads per block exceeds kernel limit" when computing a reduction on a `KernelFunctionOperation` on some GPUs


ali-ramadhan:
Following https://github.com/CliMA/Oceananigans.jl/discussions/4034 I was able to reliably reproduce the error. When using a 16³ `LatitudeLongitudeGrid` with `Float32` on an NVIDIA RTX 4090 with `--check-bounds=yes` (needed!) the MWE below produces the error below.

There is no error with a smaller 8³ grid. There is no error with `Float64`. And there is no error on a `RectilinearGrid`.

I was not able to reproduce on a V100. But I've seen this error show up when running simulations on an immersed `LatitudeLongitudeGrid` with `Float64` on a V100 and H100 without `--check-bounds=yes`.

I will try to reproduce using just CUDA.jl. It's interesting that the error suggests that the RTX 4090 has a "Maximum number of threads per block" of 512 when CUDA deviceQuery says it's 1024.

---

MWE:

```julia
using Oceananigans
using Oceananigans.Advection: cell_advection_timescaleᶜᶜᶜ

grid = LatitudeLongitudeGrid(GPU(), Float32;
    topology = (Bounded, Bounded, Bounded),
    size = (16, 16, 16),
    longitude = (-10, 10),
    latitude = (-10, 10),
    z = (-100, 0)
)

model = HydrostaticFreeSurfaceModel(; grid)

u, v, w = model.velocities
τ = KernelFunctionOperation{Center, Center, Center}(cell_advection_timescaleᶜᶜᶜ, grid, u, v, w)

τ_min = minimum(τ)
```

Error:

```
ERROR: Number of threads per block exceeds kernel limit (640 > 512).
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] diagnose_launch_failure(f::CUDA.CuFunction, err::CUDA.CuError; blockdim::CUDA.CuDim3, threaddim::CUDA.CuDim3, shmem::Int64)
    @ CUDA ~/.julia/packages/CUDA/2kjXI/lib/cudadrv/execution.jl:120
[...]

PR #4139, opened 2025-02-26
Tests on tartarus

simone-silvestri:
Sverdrup has been awfully slow lately, so the idea is to move the GPU testing on tartarus. We already occupy a tartarus GPU for building the docs, so I think there should be no problems with moving the whole testing infrastructure on one GPU on tartarus. 

I changed the testing device to device 3 so that we free device 0 (the default) for benchmarking and prototyping.

Let's see if this work and is faster


glwagner (approved):

.buildkite/pipeline.yml
```diff
@@ -54,19 +35,19 @@ steps:
...
-      JULIA_DEPOT_PATH: "$SVERDRUP_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
+      JULIA_DEPOT_PATH: "$TARTARUS_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
       TEST_GROUP: "unit"
       GPU_TEST: "true"
     commands:
-      - "$SVERDRUP_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
+      - "$TARTARUS_HOME/julia-$JULIA_VERSION/bin/julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
     agents:
       queue: Oceananigans
       architecture: GPU
```
glwagner, line 45: @simone-silvestri try changing this to CPU. Not sure it will help but it might. I think somehow the CPU "queue" is a synonym for tartarus, not sure.

simone-silvestri:
Huh, looks like tartarus has a less powerful GPU? we are encountering the resource limitations in `mapreducedim`

ali-ramadhan:
@simone-silvestri If you mean the errors like

```
Number of threads per block exceeds kernel limit (768 > 640).
```

then I actually don't think it depends on how powerful the GPU is. I've had cases where the code worked on an A100 and an RTX 4090, but then I get this error on a V100 or an H100 :(

X-Ref: https://github.com/CliMA/Oceananigans.jl/issues/4047

simone-silvestri:
Thanks @ali-ramadhan. I wonder where those threads are computed.

ali-ramadhan:
I believe, but can't find the info right now, that most modern NVIDIA GPUs have 1024 as the maximum number of threads per block. And I think this would include both the Quadro P6000 on Sverdrup (if I recall correctly) and the Titan V on Tartarus. So I'm not sure why the error is suggesting that it's 640.

I was hoping that upgrading CUDA.jl versions would fix #4047 but if not, then maybe it's worth doing some debugging this week!

glwagner:
Is this possibly a bug in CUDA?

simone-silvestri:
Yeah (for the sake of speculation) it might be that the kernel threads are inferred by the input to the `mapreducegrid` function, and CUDA does not really know what to do with `ConditionalOperations`

glwagner:
> Yeah (for the sake of speculation) it might be that the kernel threads are inferred by the input to the `mapreducegrid` function, and CUDA does not really know what to do with `ConditionalOperations`

Hmm yes this is a good clue. Perhaps we are missing some part of the AbstractArray interface.

ali-ramadhan:
I wonder if using the `master` branch of CUDA.jl will fix this. We're using v5.6.1 which is the latest but it is almost two months old and there have been 60+ commits since then.

Hmmm, can we even specify `[compat]` with a particular branch or commit? Just for testing, probably not a good idea to ship that. Not sure how to do it without committing the Manifest.toml.

simone-silvestri:
I can try committing the manifest and seeing what happens

simone-silvestri:
Seems like on CUDA master not only the tests do not error, but also previously broken tests in [gpu-abstract-operations-tests](https://buildkite.com/clima/oceananigans/builds/21161#019580e8-847d-47d2-a0c3-c7a0ce205ca0) also pass

I guess we can wait for the new release of CUDA and merge this PR so that tests speed up!

ali-ramadhan:
That's awesome news! Hopefully there will be a new release soon. Otherwise we might be able to ask CUDA.jl to tag a new version since multiple things have been fixed.

glwagner (approved):

simone-silvestri:
I removed the manifest. I am tempted to merge this where at the beginning of the tests we add commit 6fdca8670526456cc9eb94a4a370ba7e096f01e5 of CUDA.jl. We can remove that line when the new version is released

simone-silvestri:
CUDA has a new tag! When tests pass I merge

PR #4142, opened 2025-02-27
Add more UnaryOperations with @unary

### LINKED ISSUE #4130: Extend list of unary operators?
Hello,

I recently noticed that the list of basic functions which are included in Oceananigans as unary operators is relatively short. I needed to use log10, but other useful functions which are missing include log, tan, sinh, and cosh.

While it was simple enough to add log10 as a unary operator in my own code, I was wondering if we could look to extend the current list of unary operators - unless there is a reason the above were not included? I think this change would just require adding the functions to lines 70 and 73 here:

https://github.com/CliMA/Oceananigans.jl/blob/897c71d953926ec11122169a04d9b743235407b5/src/AbstractOperations/AbstractOperations.jl#L69-L75

Thanks!

glwagner:
Closes #4130

### DIFF
--- a/test/test_computed_field.jl
+++ b/test/test_computed_field.jl
@@ -364,7 +364,9 @@ for arch in archs
 
             @testset "Unary computations [$A, $G]" begin
                 @info "      Testing correctness of compute! unary operations..."
-                for unary in (sqrt, sin, cos, exp, tanh)
+                dont_test = tuple(:interpolate_identity)
+                operators_to_test = filter(op -> !(op ∈ dont_test), Oceananigans.AbstractOperations.unary_operators)
+                for unary in operators_to_test
                     @test compute_unary(unary, model)
                 end
             end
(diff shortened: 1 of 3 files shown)


test/test_computed_field.jl
```diff
@@ -364,7 +364,9 @@ for arch in archs
 
             @testset "Unary computations [$A, $G]" begin
                 @info "      Testing correctness of compute! unary operations..."
-                for unary in (sqrt, sin, cos, exp, tanh)
+                dont_test = tuple(:interpolate_identity)
+                operators_to_test = filter(op -> !(op ∈ dont_test), Oceananigans.AbstractOperations.unary_operators)
+                for unary in operators_to_test
                     @test compute_unary(unary, model)
                 end
```
glwagner, line 371: @emmabeniston ended up adding the tests here

PR #4148, opened 2025-02-28
Implement `time_step!(sim::Simulation, Δt)`

glwagner:
This is needed to implement @simone-silvestri's suggested interface for coupling. Leverages the new ability to ensure a constant time-step by setting `sim.align_time_step = false`.

### DIFF
--- a/src/Simulations/run.jl
+++ b/src/Simulations/run.jl
@@ -117,6 +117,13 @@ end
 
 const ModelCallsite = Union{TendencyCallsite, UpdateStateCallsite}
 
+""" Step `sim`ulation forward by Δt. """
+function time_step!(sim::Simulation, Δt)
+    sim.Δt = Δt
+    sim.align_time_step = false # ensure time step
+    return time_step!(sim)
+end
+
 """ Step `sim`ulation forward by one time step. """
 function time_step!(sim::Simulation)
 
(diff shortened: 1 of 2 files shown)


src/Simulations/run.jl
```diff
@@ -117,6 +117,13 @@ end
 
 const ModelCallsite = Union{TendencyCallsite, UpdateStateCallsite}
 
+""" Step `sim`ulation forward by Δt. """
+function time_step!(sim::Simulation, Δt)
+    sim.Δt = Δt
+    sim.align_time_step = false # ensure Δt
```
navidcy, line 123: I'm wondering whether `sim.align_time_step` should be returned to its original state after `time_step!`?
Or whether an info is printed out to the user?

```suggestion
    if sim.align_time_step == true
        sim.align_time_step = false # ensure Δt
        @info "sim.align_time_step set to false"
    end
```

navidcy (approved):

src/Simulations/run.jl
navidcy: hm... on second thought this might be a bit overhead

glwagner: I think that would make it harder to use not easier

glwagner: As for returning it to the original value, I think that might be harder to understand. I like that this is simple

navidcy: fair

tbh I'm not sure what's the intention; at the moment I only use `time_step!(model)` for debugging or something

PR #4150, opened 2025-03-01
New reactant feature: sharding + avoid direct usage of `RArray`

avik-pal:
cc @wsmoses @glwagner


navidcy:
(just tried to make the title more descriptive @avik-pal; feel free to modify to your liking!)

glwagner (approved):
Does `AnyConcreteRArray` still exist or no?

The design to put `sharding` in `ReactantState` makes sense to me.

src/Architectures.jl
```diff
@@ -61,7 +61,11 @@ end
 
 Run Oceananigans on Reactant.
 """
-struct ReactantState <: AbstractSerialArchitecture end
+struct ReactantState{S} <: AbstractSerialArchitecture
+    sharding::S
+end
+
+ReactantState() = ReactantState(missing)
```
glwagner, line 68: should we allow this?

trivial question: `missing` vs `nothing`? The difference in mind is: is `sharding` _required_? If sharding is required then `missing` conveys the right message. If we can also run with "no sharding" then I think `nothing` is clearer, and also consistent with what we typically use in Oceananigans for "not" doing something.

avik-pal:
> Does AnyConcreteRArray still exist or no?

Right now it does as `const AnyConcreteRArray = AnyConcretePJRTArray`. With later releases (potentially a v0.3) it will be removed and ConcreteRArray might become a function that returns a PJRT Array or IFRT Array based on certain conditions

src/Architectures.jl
avik-pal: By default it should be `NoSharding` (or `nothing`) so I think it makes sense to default to nothing. I will update it

PR #4153, opened 2025-03-03
Update spatial_operators.md

briochemc:
Add explanation of what the superscripts (e.g., "faa") mean in the docs.


navidcy (approved):

glwagner:
I love this and will merge without hesitation.

I think we are missing "Developer docs". Eg the superscripts are not really an aspect of the numerical implementation. It's design of the low-level code. 

I think we should merge this but keep in mind we need developer docs.

PR #4156, opened 2025-03-03
Add support for ZStar in RK3

simone-silvestri:
needed for running global simulations with an RK3 timestepper and a ZStar vertical coordinate.

For a hydrostatic model, we save the scaled tracer instead of the tracer and unscale it later once it has been evolved.

I am not 100% sure this is exactly correct, because it seems like there is a (very small) positive drift in integrated tracer.
However, I d like to merge so that we can do some more in-depth testing in ClimaOcean

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
@@ -137,7 +137,7 @@ end
     σᶜᶜⁿ = σⁿ(i, j, k, grid, Center(), Center(), Center())
 
     @inbounds begin
-        ∂t_σθ = α * σᶜᶜⁿ * Gⁿ[i, j, k]
+        ∂t_σθ = σᶜᶜⁿ * Gⁿ[i, j, k]
         
         # We store temporarily σθ in θ. 
         # The unscaled θ will be retrieved with `unscale_tracers!`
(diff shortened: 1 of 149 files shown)


glwagner:
can you show the lock release results

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
```diff
@@ -1,15 +1,15 @@
-using Oceananigans.Fields: location
+using Oceananigans.Fields: location, instantiated_location
 using Oceananigans.TurbulenceClosures: implicit_step!
 using Oceananigans.ImmersedBoundaries: retrieve_interior_active_cells_map, retrieve_surface_active_cells_map
 
-import Oceananigans.TimeSteppers: split_rk3_substep!, _split_rk3_substep_field!
+import Oceananigans.TimeSteppers: split_rk3_substep!, _split_rk3_substep_field!, store_fields!
```
glwagner, line 5: what does "store fields" mean?

simone-silvestri: it's the same concept as `store_tendencies!` of the other RK3, however in this formulation of RK3 instead of storing the tendencies we store prognostic fields

glwagner: maybe `cache_previous_tendencies` and `cache_previous_fields` would be more clear. `store_tendencies` made sense because tendencies are only used for one thing (updating fields). But fields are used for many things, so "storing" fields could be used for a wide variety of purposes, like output, etc.

glwagner (approved):

glwagner:
This seems ready for use basically. Should we export RK3? Make it default?

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
simone-silvestri: sounds good! I ll open a new PR that does this change

simone-silvestri:
I probably need to improve the testing infrastructure a bit. Then make sure it works with all free surfaces (at the moment it does not work for an implicit free surface), then I guess we can make it a default

PR #4157, opened 2025-03-03
Fix reconstruct_global_grid

glwagner:
Because Distributed tests are currently broken, I found this issue running the tests locally. It might be helpful @simone-silvestri if you can reproduce that the tests pass.

Also there is no specific unit test for `reconstruct_global_grid`. This error is found indirectly by the hydrostatic models test. @simone-silvestri I suggest adding a specific unit test in another PR.


simone-silvestri (approved):
nice that you could find it this quickly. I can add a test for `reconstruct_global_grid` in another PR

glwagner:
Could this be the cause of https://github.com/CliMA/ClimaOcean.jl/issues/366 ?

glwagner:
Another design would catch this quicker too btw. Right now we define a fallback for serial grids and specialize on distributed. This allows things to fail silently when the specialization breaks. If we eliminate the fallback, the function will stop working altogether and the specialization failure will be more obvious. We have two serial architectures, CPU and GPU, and don't expect more in the future.

PR #4160, opened 2025-03-04
Clean up of active map utils

glwagner:
This PR will consider changing the active_map from Array{Tuple} to Tuple{Array, Array}.


src/ImmersedBoundaries/active_cells_map.jl
```diff
@@ -176,12 +160,12 @@ end
 # In case of a serial grid, the interior computations are performed over the whole three-dimensional
 # domain. Therefore, the `interior_active_cells` field contains the indices of all the active cells in 
 # the range 1:Nx, 1:Ny and 1:Nz (i.e., we construct the map with parameters :xyz)
-compute_active_cells(ibg) = serially_compute_active_cells(ibg; parameters = :xyz)
+compute_active_cells(grid, ib) = serially_compute_active_cells(grid, ib; parameters = :xyz)
```
simone-silvestri, line 163: I wonder if we can avoid this extra function and just call everything
```julia
compute_active_cells(grid, ib)
```

glwagner: not sure, but I got burned trying to eliminate a function earlier that happened to be extended in a constructor or by distributed, so decided to leave this.

PR #4169, opened 2025-03-05
Expose `tridiagonal_direction` as kwarg for `FourierTridiagonalPoissonSolver`

glwagner:
This also cleans up a number of assumptions that were made within the constructor.


tomchor (approved):

tomchor:
Looks good. Is this needed for something, or just a general improvement?

glwagner:
> Looks good. Is this needed for something, or just a general improvement?

It's needed to test out a different boundary condition for the Poisson solver that requires a vertical tridiagonal solve. Also its a better more general interface and eliminates the assumption that the tridiagonal direction is always linked to the grid stretching

PR #4171, opened 2025-03-06
Bump to 0.95.20

glwagner:


tomchor:
This skips 0.95.19, no?

glwagner:
yeah, we don't need it. We don't need to revert though, just need to merge a few more packages and stay at 0.95.20 and then register

PR #4172, opened 2025-03-06
Expose Simulation dictionaries to constructor

glwagner:
This allows the three dictionaries `diagnostics`, `callbacks`, and `output_writers` to be specified by the user. This allows them to be changed from dictionaries to, say, `NamedTuple`, which are concretely typed.


glwagner:
Doesn't help

PR #4182, opened 2025-03-07
Fill halo regions when lagrangian averaging Smagorinsky coefficient

tomchor:
At the moment halo regions aren't being filled, so periodicity isn't being properly taken into account when calculating the lagrangian averaging. This PR fixes that.

@glwagner  can you double check that I included it in the right place?

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
@@ -1,5 +1,6 @@
 using Oceananigans.Architectures: architecture
 using Oceananigans.Fields: interpolate
+using Oceananigans.BoundaryConditions: fill_halo_regions!
 using Statistics
 
 struct DynamicCoefficient{A, FT, S}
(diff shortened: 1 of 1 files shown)


glwagner:
hmm

I see this

https://github.com/CliMA/Oceananigans.jl/blob/c4185c6621ec26774854ba970f206a6f3058a105/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl#L290-L291

`parent(a)` includes the halo regions of `a`.

Combined with

https://github.com/CliMA/Oceananigans.jl/blob/c4185c6621ec26774854ba970f206a6f3058a105/src/Models/NonhydrostaticModels/update_nonhydrostatic_model_state.jl#L45

it would seem the halos are filled?

Note that

https://github.com/CliMA/Oceananigans.jl/blob/c4185c6621ec26774854ba970f206a6f3058a105/src/Models/NonhydrostaticModels/update_nonhydrostatic_model_state.jl#L45

does more than it needs to, considering the `parent` fill earlier...

this can be optimized using the approach that CATKE uses

https://github.com/CliMA/Oceananigans.jl/blob/c4185c6621ec26774854ba970f206a6f3058a105/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl#L170-L178

glwagner:
maybe it would be more secure to use a simple test rather than reading the code, not sure.

tomchor:
For the record, I don't see any evidence of non-periodicity on the eddy viscosity or diffusivity when using a Lagrangian averaged coeff in a periodic domain. For example the figure below shows a simple seamount simulation. Both figures show the exact same data, but the panel on the right offsets the data in the x-direction:

image

I don't see any signs of discontinuity in the calculation. Although granted that's not exactly a rigorous test.

> maybe it would be more secure to use a simple test rather than reading the code, not sure.

Do you have any test in mind?

glwagner:
I'm not sure if this would catch everything but you can set `𝒥ᴹᴹ⁻` and `𝒥ᴹᴹ` and `model.velocities` to something, call `compute_coefficient_fields!`, and verify the answer you get out is what you expected.

glwagner:
> I don't see any signs of discontinuity in the calculation. Although granted that's not exactly a rigorous test.

I personally think this is pretty strong evidence that all's well, especially given there's no specific reason to suspect an issue.

tomchor:
Yeah I think the plots, combined with your comment in https://github.com/CliMA/Oceananigans.jl/pull/4182#issuecomment-2707745895, are pretty good evidence that we're doing the right thing. Thanks for digging into that.

PR #4185, opened 2025-03-10
Add non-trivial default boundary conditions for `FieldTimeSeries`

glwagner:
These were not add originally because they were not needed, but may be needed now @simone-silvestri


navidcy:
x-ref: https://github.com/CliMA/ClimaOcean.jl/pull/387

simone-silvestri (approved):

Issue #4183, opened 2025-03-09
Missing information in docstring for `BoundaryAdjacentMean`


glwagner:
The function signature in the docstring is missing, and the docstring ends with the phrase "Updated by calling"...

Maybe something was accidentally deleted?

https://github.com/CliMA/Oceananigans.jl/blob/d2e9e272ba2a625fed3582fd19dcfa69d66ee6bb/src/Models/boundary_mean.jl#L12-L15

I was also confused about the purpose of this object. The user interface seems to allow the generation of an operator that can act on any field. But what should this operator used for?

I'm also perplexed by the property names. One property is `flux_field`. But what does `BoundaryAdjacentMean` have to do with fluxes?

https://github.com/CliMA/Oceananigans.jl/blob/d2e9e272ba2a625fed3582fd19dcfa69d66ee6bb/src/Models/boundary_mean.jl#L51

It's also problematic to me that `value` needs to be a `Ref` or stored at all. This creates ambiguity about how the operator is supposed to be used, one can either call the operator and store the value manually, or call it without explicitly storing the value. Probably for the sake of clarity it is better not to store the result in a `Ref`.

Can you clean this up @jagoosw ?

glwagner:
> We need to store the value otherwise, it would have to be computed at every boundary point, and I don't know how else you would do that (other than a 1-element array)?

But it doesn't have to be stored in `BoundaryAdjacentMean`. It is more properly stored in `PertubationAdvectionOpenBoundaryCondition`. In other words, operators don't need to store their values. It's only the _user_ of the operator that needs it.

glwagner:
Why can't `condition` be a `Ref`? This matches the pattern we use elsewhere when it has values that need to be updated (for example, when `condition` is an array). 

Where do we compute the value of `condition`?

jagoosw:
The condition gets computed in `update_boundary_condition!`, I don't understand how we would write a method that computes it without having either `condition` or `classifcation` be something like `BoundaryAdjacentMean` (or we could redesign PAOBC).

glwagner:
My point is that the example:

```julia
julia> bam = BoundaryAdjacentMean(grid, :east)
BoundaryAdjacentMean: (0.0)

julia> bam(:east, cf)
-1.5612511283791264e-18
```

[...]

PR #4189, opened 2025-03-10
Fix `BoundaryAdjacentMean`

jagoosw:
Addresses concerns in #4183 and bug


src/Models/boundary_mean.jl
glwagner, line 1: should we call this `boundary_adjacent_mean` to match the type name?

src/Models/boundary_mean.jl
```diff
@@ -11,85 +11,116 @@ import Oceananigans.BoundaryConditions: update_boundary_condition!
...
+├── top: DefaultBoundaryCondition (FluxBoundaryCondition: Nothing)
+└── immersed: DefaultBoundaryCondition (FluxBoundaryCondition: Nothing)
+
 ```
 """
-struct BoundaryAdjacentMean{FF, BV}
-    flux_field :: FF
-         value :: BV
+struct BoundaryAdjacentMean{NI, BV}
+    boundary_normal_integral :: NI
```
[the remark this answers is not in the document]
jagoosw: That's a good point, I guess `boundary_parallel_integral` is more accurate? boundary normal integal is actually wrong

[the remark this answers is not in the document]
glwagner: I think merely `integral` is enough. The name of the type is "boundary adjacent mean". If we compute the mean by first taking the integral and dividing by area, then `integral` combined with the name of the type give you what yo uneed to know.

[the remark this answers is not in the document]
glwagner: The first rule is "do no harm"

src/Models/boundary_adjacent_mean.jl
```diff
@@ -0,0 +1,156 @@
...
+using Oceananigans.Fields: Center, Face
+using Oceananigans.AbstractOperations: GridMetricOperation, Ax, Ay, Az, @at
+using Oceananigans.BoundaryConditions: BoundaryCondition, Open, PerturbationAdvection
+
+import Adapt: adapt_structure
+import Base: summary, show
+import Oceananigans.BoundaryConditions: update_boundary_condition!
+
+"""
+    BoundaryAdjacentMean
```
glwagner, line 12: the function signature is missing arguments

src/Models/boundary_adjacent_mean.jl
```diff
@@ -0,0 +1,156 @@
...
+# computes the boundary mean and stores/returns it
+function (bam::BoundaryAdjacentMean)(val_side::Val, u)
+    grid = u.grid
+
+    loc = instantiated_location(u)
+    iB, jB, kB = boundary_adjacent_indices(val_side, grid, loc)
+    An = boundary_normal_area(val_side, grid)
+
+    # get the total flux
+    sum!(bam.boundary_parallel_integral, (@at (Face, Center, Center) u) * An)
```
glwagner, line 136: A few comments:

1. It might be nicer to preform this operation --- that way users can look at it, and potentially verify its correctness.
2. Why don't we use `Average` here?
3. `sum!` is incorrect if the grid is stretched, right?

src/Models/boundary_adjacent_mean.jl
```diff
@@ -0,0 +1,156 @@
...
+  boundary_parallel_integral :: NI
+                       value :: BV
+
+   BoundaryAdjacentMean(grid, side; 
+                        boundary_parallel_integral::NI = boundary_reduced_field(Val(side), grid),
+                        value::BV = Ref(zero(grid))) where {NI, BV} = 
+        new{NI, BV}(boundary_parallel_integral, value)
+end
+
+@inline (bam::BoundaryAdjacentMean)(args...) = bam.value[]
```
glwagner, line 100: I just noticed this. I would offer that this code is surprising and hard to understand.

src/Models/boundary_adjacent_mean.jl
jagoosw: Do you mean to return the operation and let the user do it?

I can't exactly remember how I ended up with this design but I think I did it this was so that we don't have to allocate a new average field (since you can't do in place averaging right?)

And I think sum is correct since it is multiplied by the area right?

jagoosw:
> I'm no longer sure that we need the type `BoundaryAdjacentMean`. Can't we construct a `Field` directly with all of the information it needs, and then call `compute!` on that field within `update_boundary_condition!`? What we need is a function `boundary_adjacent_mean(side, field)` which constructs the appropriate Reduction. `getbc` then works correctly (I think); for example the needed value is stored in `field.data`.
> 
> A disadvantage is that we need to use `regularize_boundary_condition` to set it up for users, because users won't have access to `field` before they build the model.
> 
> However, there are some significant advantages:
> 
>     1. More code re-use; no extra type needed.
> 
>     2. The way the code works is easier to understand, provided that you already understand reductions.
> 
>     3. The type is specific to the boundary and its value can be computed without dependencies via `compute!(field)`.
> 
>     4. Because it is a `Field`, users can use an output writer to write its value as a diagnostic. It can also participate in additional operations; for example to compute a flux across the boundary.
> 
>     5. We can simplify `update_boundary_condition!`.

I think the reason I ended up with this was to avoid using `regularize_boundary_condition` but that seems like a reasonable solution. I think we would also need to add a method for `update_boundary_condition!` when the BC has a compute field in the condition.

glwagner:
Ok, so we have to decide which is best: adopt complication of `regularize_boundary_condition` but simplify what we get out. Probably the main thing is that `Field` can participate in abstract operations, so if we think that matters we should use it. Also it might be easier to generalize to open boundary velocities that vary in space (eg not a mean over the whole boundary) ?

glwagner:
This is still needed, isn't it?

PR #4195, opened 2025-03-10
Fix bug in active-weighted interpolation

glwagner:
Thanks to @luraess for identifying this


luraess:
Upon testing it I am getting the same results as for the reference working simulation using OA 0.95.12. So this seems to fix the velocity drift.

glwagner:
Simple test but shows the problem at the boundary:

not bugged (this PR):

https://github.com/user-attachments/assets/4b4ede57-fe42-45c1-b966-0176f685662f

bugged:

https://github.com/user-attachments/assets/750f99ec-edeb-49a0-88bb-71405ed89f84

PR #4196, opened 2025-03-10
Change DiffusivityFields to build_diffusivity_fields and add clock arg

glwagner:
Plus correctly determine type of CATKE's `previous_compute_time`.


glwagner:
when search and replace fails you

glwagner:
thanks @giordano

PR #4198, opened 2025-03-11
[CI] Some improvements to GitHub Actions setup

giordano:
* Trigger on pull requests, pushes to `main` and tags
* Add concurrency group to cancel subsequent jobs in PRs. This is particularly useful to cancelling quickly macOS jobs, since there are only 5 concurrent runners available for the entire organisation
* Use `julia-actions/cache` instead of `actions/cache`
* Add timeouts, to avoid hanging jobs hold on to the runners for 6 hours.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,35 @@
 name: CI
+
 on:
-  - push
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


giordano:
Uhm, not sure why jobs aren't starting at all 🤔

PR #4200, opened 2025-03-11
Implement Distributed{<:ReactantState}

glwagner:
This PR generalizes `Distributed` to work with Reactant / XLA's sharding, so that the distributed computational graph can be optimized by XLA.


glwagner:
Moving forward with this @wsmoses

wsmoses:
wait to merge this please, still testing

simone-silvestri:
The tests here pass (except for the offset array problem which should be solved in the new version of reactant).
Interestingly, the sharded tests pass for a `LatitudeLongitudeGrid` but not for a `TripolarGrid`, I have disabled the tests for a tripolar grid for the moment so we can proceed. Note:
https://github.com/CliMA/Oceananigans.jl/blob/982d65a813e98418457e0305ba38633905730689/test/runtests.jl#L251-L258

I think we should merge this once the tests fully pass and move on.

glwagner:
cc @avik-pal are we ok to merge this? There may be conflicts to resolve with #4262

ext/OceananigansReactantExt/TripolarGrid.jl
```diff
@@ -0,0 +1,92 @@
...
+        throw(ArgumentError("The x partition $(px) is not supported. The partition in x must be an even number. "))
+    end
+
+    # a slab decomposition in x is not supported
+    if px != 1 && py == 1
+        throw(ArgumentError("A x-only partitioning is not supported with the TripolarGrid.\n
+                             Please, use a y partitioning configuration or a x-y pencil \
+                             partitioning."))
+    end
+
```
avik-pal, line 24:
```suggestion

```

simone-silvestri (approved):

test/test_distributed_latitude_longitude.jl
glwagner, line 1: is this a reactant test only?

simone-silvestri: yep, it's a distributed test so it can be used for reactant or MPI interchangeably, depending on the child architecture we pass

glwagner: I think you need a different design to really achieve that fully, so I changed it to reactant test. We can work on generalizing it to run MPi tests in the future if we want.

PR #4204, opened 2025-03-11
Adding a distributed tripolar grid

simone-silvestri:
it's basically a copy - paste of https://github.com/CliMA/OrthogonalSphericalShellGrids.jl/pull/32. I have added the tests to the github action workflow so that we can use different cores without spinning up new jobs (which is fairly slow on the caltec cluster)

The limitation of a distributed tripolar grid are:
- an even number of partitions in x
- only homogenous partitioning in x (no different sizes between ranks)
- partitioning  in x only of there are at least 2 ranks in y
- no limitation in y (variable size or uneven partitions allowed)

I would like to merge #4206 before to use it in the tests

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,38 +1,20 @@
 env:
   JULIA_VERSION: "1.10.8"
   JULIA_MINOR_VERSION: "1.10"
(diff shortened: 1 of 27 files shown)


simone-silvestri:
Distributed tests are running in github action 🚀

glwagner (approved):

PR #4206, opened 2025-03-11
A utility function to mask tripolar grid singularities

simone-silvestri:
which I forgot in #3913

### DIFF
--- a/src/OrthogonalSphericalShellGrids/tripolar_field_extensions.jl
+++ b/src/OrthogonalSphericalShellGrids/tripolar_field_extensions.jl
@@ -94,7 +94,7 @@ function tupled_fill_halo_regions!(full_fields, grid::TripolarGridOfSomeKind, ar
     end
 end
 
-# Mask the singularity of the grid in a region of `radius` degrees around the singularities
+# Mask the singularity of the grid in a region `radius` degrees large around the singularities
 function mask_tripolar_singularities(underlying_grid::TripolarGrid; radius = 5) # degrees
     λp = underlying_grid.conformal_mapping.first_pole_longitude
     φp = underlying_grid.conformal_mapping.north_poles_latitude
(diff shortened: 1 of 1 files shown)


src/OrthogonalSphericalShellGrids/tripolar_field_extensions.jl
```diff
@@ -93,3 +93,20 @@ function tupled_fill_halo_regions!(full_fields, grid::TripolarGridOfSomeKind, ar
         fill_halo_regions!(field, args...; kwargs...)
     end
 end
+
+# Mask the singularity of the grid in a region `radius` degrees large around the singularities
+function analytical_immersed_tripolar_grid(underlying_grid::TripolarGrid; radius = 5) # degrees
```
glwagner, line 98: This is a weird name

src/OrthogonalSphericalShellGrids/tripolar_field_extensions.jl
```diff
@@ -93,3 +93,20 @@ function tupled_fill_halo_regions!(full_fields, grid::TripolarGridOfSomeKind, ar
...
+    λp = underlying_grid.conformal_mapping.first_pole_longitude
+    φp = underlying_grid.conformal_mapping.north_poles_latitude
+    φm = underlying_grid.conformal_mapping.southernmost_latitude
+
+    Lz = underlying_grid.Lz
+
+    # We need a bottom height field that ``masks'' the singularities
+    bottom_height(λ, φ) = ((abs(λ - λp) < radius)       & (abs(φp - φ) < radius)) |
+                          ((abs(λ - λp - 180) < radius) & (abs(φp - φ) < radius)) | (φ < φm) ? 0 : - Lz
```
glwagner, line 107: circular islands around the singularities

maybe write this code more clearly, it's pretty hard to parse

glwagner (approved):
Confused about why this needs to be in the source code.

Also, why do we need to mask a `radius` around the singularity. Is the singularity not just one point? What if we just label the singular points as `inactive` even when not using `ImmersedBoundaryGrid`?

simone-silvestri:
I can just close this PR and use the function in the tests, probably better

PR #4210, opened 2025-03-11
Rename output writers to `JLD2Writer` and `NetCDFWriter`

tomchor:
Let's make these names less verbose. I'm starting with the `JLD2Writer` first and will work on the `NetCDFWriter` after 
https://github.com/CliMA/Oceananigans.jl/pull/4046 is merged.

### DIFF
--- a/docs/src/model_setup/lagrangian_particles.md
+++ b/docs/src/model_setup/lagrangian_particles.md
@@ -120,7 +120,7 @@ model = NonhydrostaticModel(; grid, particles=lagrangian_particles)
 ```
 
 ```@example particles
-JLD2OutputWriter(model, (; particles=model.particles), filename="particles", schedule=TimeInterval(15))
+JLD2Writer(model, (; particles=model.particles), filename="particles", schedule=TimeInterval(15))
 ```
 
 When writing to NetCDF you should write particles to a separate file as the NetCDF dimensions differ for
--- a/docs/src/model_setup/output_writers.md
+++ b/docs/src/model_setup/output_writers.md
@@ -4,17 +4,17 @@
 `Oceananigans` provides three ways to write output:
 
 1. [`NetCDFOutputWriter`](@ref) for output of arrays and scalars that uses [NCDatasets.jl](https://github.com/Alexander-Barth/NCDatasets.jl)
-2. [`JLD2OutputWriter`](@ref) for arbitrary julia data structures that uses [JLD2.jl](https://github.com/JuliaIO/JLD2.jl)
+2. [`JLD2Writer`](@ref) for arbitrary julia data structures that uses [JLD2.jl](https://github.com/JuliaIO/JLD2.jl)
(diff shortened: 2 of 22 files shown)


glwagner:
> Let's make these names less verbose. I'm starting with the `JLD2Writer` first and will work on the `NetCDFWriter` after #4046 is merged.

How about merging #4046 and then changing both in one PR?

Also the comment "working on" makes me a little nervous. This is just a search and replace that takes very little time, isn't it?

glwagner:
Does it make sense to get #4046 in first and then change the names?

tomchor:
> Does it make sense to get #4046 in first and then change the names?

Yeah, sorry that's what I meant. I'm started changing the names for JLD2 now, before #4046 is merged, and after that's done, I'm gonna merge main into this PR, change the `NetCDFWriter` names, and merge. To clarify: I'm proposing doing everything in this PR. Sorry that wasn't clear!

glwagner:
> > Does it make sense to get #4046 in first and then change the names?
> 
> Yeah, sorry that's what I meant. I'm started changing the names for JLD2 now, before #4046 is merged, and after that's done, I'm gonna merge main into this PR, change the `NetCDFWriter` names, and merge. To clarify: I'm proposing doing everything in this PR. Sorry that wasn't clear!

Okay, got it!

glwagner:
I am also wondering if we should move the `NetCDFWriter` to an extension as part of this breaking change. This will mean that users have to write

```julia
using NCDatasets
```

to use `NetCDFWriter`. The advantage is that we won't have to modify the source code to run on systems that don't support NetCDF, which has happened a few times in the past and might happen more and more as we try to run on more unusual systems.

tomchor:
> I am also wondering if we should move the `NetCDFWriter` to an extension as part of this breaking change. This will mean that users have to write
> 
> ```julia
> using NCDatasets
> ```
> 
> to use `NetCDFWriter`. The advantage is that we won't have to modify the source code to run on systems that don't support NetCDF, which has happened a few times in the past and might happen more and more as we try to run on more unusual systems.

I think this makes sense.

tomchor:
There is a file called `multi_region_output_writers.jl` in the `MultiRegion` module that I'm unsure whether or not to rename. For now I kept it the same since it connects to the `OutputWriters` module, where `NetCDFWriter` and `JLD2Writer` live.

glwagner (approved):

glwagner:
> There is a file called `multi_region_output_writers.jl` in the `MultiRegion` module that I'm unsure whether or not to rename. For now I kept it the same since it connects to the `OutputWriters` module, where `NetCDFWriter` and `JLD2Writer` live.

We can probably delete this / discontinue support for multi region output writing.

glwagner:
If this PR gets hung up, I think it will be ok to tag a couple minor releases, eg up to 0.97 once this is ready.

PR #4213, opened 2025-03-13
Make dt constant in CATKE

glwagner:
This makes the time-step constant in CATKE to test if this helps us compile with Reactant @giordano


giordano:
Yes, this seems to have identified the problem in https://github.com/PRONTOLab/GB-25/pull/40#issuecomment-2714150363.

giordano:
Probably superseded by #4220

PR #4216, opened 2025-03-13
Expose pickup functionality in `run!`

navidcy:
This PR exposes the pickup functionality from `run!` (so we can add new methods for picking up a coupled ocean-sea ice simulation in ClimaOcean).

related to https://github.com/CliMA/ClimaOcean.jl/pull/401


src/Simulations/run.jl
```diff
@@ -94,7 +94,7 @@ function run!(sim; pickup=false)
     start_run = time_ns()
 
     if we_want_to_pickup(pickup)
-        set!(sim, pickup)
+        pickup!(sim, pickup)
```
glwagner, line 97: isn't this just changing the name of `set!` to `pickup!`?

navidcy: yes, well for an ocean model in Oceananigans pickup! is just that set!, but in ClimaOcean pickup! does more than that

taimoorsohail: By exposing the pickup! function, we are able to define a OceanSeaIceModel specific pickup!:

```
function pickup!(sim::OSIMSIM, pickup)
    @info "i try to properly pick up"
    set!(sim.model.ocean, pickup)
    return nothing
end
```

glwagner: can you extend `set!` for `OSIMSIM` to achieve the same thing?

glwagner:
```julia
function set!(sim::OSIMSIM, pickup)
    @info "i try to properly pick up"
    set!(sim.model.ocean, pickup)
    return nothing
end
```

taimoorsohail: This is a lot less work than modifying `set!` itself, as that would require changing `write_output!` and `checkpointer_address`, among others

glwagner: I know I suggested this strategy 😭 I didn't realize that it was _already_ exposed neatly

navidcy: OK, I manage to make it work without `pickup!`!

navidcy:
Tricky was to remove the `::AbstractModel` type from `set!(model::AbstractModel, filepath::AbstractString)`. Otherwise I wasn't able to extend `set!` in ClimaOcean.

navidcy:
Actually, with https://github.com/CliMA/ClimaOcean.jl/pull/401/commits/9ab0df109e3cb05aef99f2f44d98b8246857faf9 we don't seem to be needing this PR at all!

taimoorsohail:
Shall we close this PR?

navidcy:
(we can reopen if needed)

PR #4217, opened 2025-03-13
Reduce data precision

simone-silvestri:
This is a PR just to check that tests run with reduced offset precision

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,1006 @@
+# This file is machine-generated - editing it directly is not advised
+
+julia_version = "1.10.9"
+manifest_format = "2.0"
+project_hash = "447cc9b8ae047dde8c69787ed7b427fc45a5622d"
+
+[[deps.AbstractFFTs]]
+deps = ["LinearAlgebra"]
+git-tree-sha1 = "d92ad398961a3ed262d8bf04a1a2b8340f915fef"
+uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
+version = "1.5.0"
+
+    [deps.AbstractFFTs.extensions]
+    AbstractFFTsChainRulesCoreExt = "ChainRulesCore"
+    AbstractFFTsTestExt = "Test"
+
+    [deps.AbstractFFTs.weakdeps]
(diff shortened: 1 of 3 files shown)


Manifest.toml
glwagner, line 1: if you use the `buildkite.yml` approach, it can actually be merged.

simone-silvestri: there are problems with reactant though.

src/Grids/new_data.jl
```diff
@@ -32,19 +32,53 @@ offset_indices(::Nothing, topo, N, H, ::UnitRange) = 1:1
...
+
+# OneTo ranges have offset 0!
+converted_offset(IntType, ::Base.OneTo) = IntType(0)
+converted_offset(IntType, t::Tuple) = Tuple(converted_offset(IntType, i) for i in t)
+
+function find_minimum_precision(ii::Integer)
+    maxInt8  = typemax(Int8)
+    maxInt16 = typemax(Int16)
+    maxInt32 = typemax(Int32)
+    IntType = ii > maxInt8 ? (ii > maxInt16 ? (ii > maxInt32 ? Int64 : Int32) : Int16) : Int8
```
glwagner, line 52: this is confusing. Let's just use Int8 ?

src/Grids/new_data.jl
```diff
@@ -32,19 +32,53 @@ offset_indices(::Nothing, topo, N, H, ::UnitRange) = 1:1
...
+    return IntType
+end
+
+find_minimum_precision(ii::UnitRange) = find_minimum_precision(first(ii))
+    
+function find_minimum_precision(ii::Union{AbstractArray, Tuple, AbstractRange})
+    IntTypes = [find_minimum_precision(i) for i in ii]
```
glwagner, line 59: this is problematic

Manifest.toml
glwagner: also you need to write the code a bit more carefully

glwagner: your code is type unstable

glwagner:
stale?

simone-silvestri:
yeah I think we can close this.

PR #4218, opened 2025-03-13
Mask CATKE's TKE

glwagner:
I noticed that TKE is not masked because `compute_diffusivities!` follows `mask_immersed_field!` within `update_state!`.


glwagner:
Also need to do this for TKEBasedVerticalDiffusivity

simone-silvestri (approved):

PR #4220, opened 2025-03-13
Modify CATKE to work with Reactant when using fixed time step size

glwagner:
@giordano


giordano:
I think this is good to go?

PR #4221, opened 2025-03-13
Define parent type correctly?

glwagner:
@avik-pal am I doing this right?

### DIFF
--- a/src/Fields/field.jl
+++ b/src/Fields/field.jl
@@ -411,8 +411,8 @@ Base.checkbounds(f::Field, I...) = Base.checkbounds(f.data, I...)
 
 @inline Base.fill!(f::Field, val) = fill!(parent(f), val)
 @inline Base.parent(f::Field) = parent(f.data)
-Adapt.parent_type(f::Field) = typeof(parent(f))
 Adapt.adapt_structure(to, f::Field) = Adapt.adapt(to, f.data)
+Adapt.parent_type(T::Type{<:Field}) = T
 
 total_size(f::Field) = total_size(f.grid, location(f), f.indices)
 @inline Base.size(f::Field)  = size(f.grid, location(f), f.indices)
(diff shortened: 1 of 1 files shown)


src/Fields/field.jl
```diff
@@ -411,8 +411,8 @@ Base.checkbounds(f::Field, I...) = Base.checkbounds(f.data, I...)
 
 @inline Base.fill!(f::Field, val) = fill!(parent(f), val)
 @inline Base.parent(f::Field) = parent(f.data)
-Adapt.parent_type(f::Field) = typeof(parent(f))
 Adapt.adapt_structure(to, f::Field) = Adapt.adapt(to, f.data)
+Adapt.parent_type(T::Type{<:Field}) = T
```
avik-pal, line 415: It should be the type of `parent(::Field)`

giordano: Based on https://github.com/CliMA/Oceananigans.jl/blob/73092f2297f0a3a77f730afcf666c6b2a22b5c80/src/Fields/field.jl#L413 and https://github.com/CliMA/Oceananigans.jl/blob/73092f2297f0a3a77f730afcf666c6b2a22b5c80/src/Fields/field.jl#L19-L21 this should be the 7th parameter of `Field{...}`

giordano: More specifically
```suggestion
Adapt.parent_type(::Type{<:Field{LX, LY, LZ, O, G, I, D}}) where {LX, LY, LZ, O, G, I, D} = D
```

glwagner: note `parent` goes down two levels --- `field.data isa OffsetArray`, but `parent(field) = parent(field.data)` is something like Array / CuArray / etc. So we need somethin like `where D <: OffsetArray{T, N, A} = A`?

PR #4222, opened 2025-03-13
Use correct eltype on abstractarray override

wsmoses:
cc @giordano @avik-pal


glwagner (approved):
👀

glwagner:
@wsmoses ready to merge?

PR #4223, opened 2025-03-14
Introduce concept of initialization update state

glwagner:


src/Models/HydrostaticFreeSurfaceModels/set_hydrostatic_free_surface_model.jl
```diff
@@ -60,8 +60,8 @@ model.velocities.u
         @apply_regionally set!(ϕ, value)
     end
 
-    initialize!(model)
-    update_state!(model; compute_tendencies=false)
+    # initialize!(model)
```
glwagner, line 63: @simone-silvestri do you know why this is here? This has to do with free surface. But I am not sure why we need it. Note, it can’t be req for update_state bc the first call is in model constructor. So this may be superfluous, or it needs to happen sooner.

glwagner: Basically we are not guaranteed to call set! So nothing can depend on it

simone-silvestri: This was there to make sure that the barotropic velocities are consistent with the velocities provided by `set!`ting the model. In practice, it just reconciles the barotropic velocities with the total ones. 
If you want to remove it from here, we could put it into `set!(model, ...)` but then it would not cover the case when people would do 
```julia
set!(model.velocities.u, ...)
```

glwagner: Are the barotropic velocities not computed in `update_state!`?

simone-silvestri: Differently from other free surfaces, for the split explicit free surface the "source of truth" is the barotropic velocity, which is used to correct the total velocity in the `pressure_correction` step. Therefore, the barotropic velocity needs to be initialized in some way if we set the velocities. In this step, the barotropic velocity is initialized with the vertical integral of the total velocity.

glwagner: What is the actual function that computes the barotropic velocities. Is there a way to call this independently? It's challenge that its grouped with `initialize!(model)`. Also it's not clear from reading the code that this is intended to compute the barotropic velocities.

PR #4224, opened 2025-03-14
Create a narrower definition for initialization

glwagner:
This PR changes the names of functions and moves a few things around so that:

1) "setup" is defined as the activities that occur prior to running a Simulation / doing a time-stepping loop
2) "initialization" is defined as the action one takes _just prior_ to running a simulation, after setup

Initialization is combined with the concept of the "first time step". Note, this is compiled by Reactant.

This organization will help us avoid launching kernels when we are not able to; eg when we are using Reactant, while still ensuring that a simulation is correct.

Note there is a TODO to allow users to set the barotropic velocity directly for the split explicit free surface stepper. The split explicit barotropic velocity is, strictly speaking, and independent prognostic variable.

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/compute_slow_tendencies.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/compute_slow_tendencies.jl
@@ -41,7 +41,7 @@ end
     return ifelse(immersed, zero(grid), Gⁿ⁺¹)
 end
 
-@inline function compute_split_explicit_forcing!(GUⁿ, GVⁿ, grid, Guⁿ, Gvⁿ, 
+@inline function compute_split_explicit_slow_tendency!(GUⁿ, GVⁿ, grid, Guⁿ, Gvⁿ, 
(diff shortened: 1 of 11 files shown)


src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/compute_slow_tendencies.jl
```diff
@@ -141,12 +141,12 @@ function compute_free_surface_tendency!(grid, model, free_surface::SplitExplicit
     stage = model.clock.stage
 
     @apply_regionally begin
-        compute_split_explicit_forcing!(GUⁿ, GVⁿ, grid, Guⁿ, Gvⁿ, baroclinic_timestepper, Val(stage))
+        compute_split_explicit_slow_tendency!(GUⁿ, GVⁿ, grid, Guⁿ, Gvⁿ, baroclinic_timestepper, Val(stage))
         initialize_free_surface_state!(free_surface, baroclinic_timestepper, barotropic_timestepper, Val(stage))
```
simone-silvestri, line 145:
```suggestion
```

glwagner:
This went stale.

PR #4225, opened 2025-03-14
Make more things mutable

glwagner:
This will hopefully help with compile time.


giordano:
This also seems to break things: https://github.com/PRONTOLab/GB-25/pull/69#issuecomment-2727544771 (and it doesn't seem to help too much with compile times)

giordano:
I think I last tested on 1e16a3b337e7e04660250cf863bb50c6b0e415f6, which is the tip of this branch

glwagner:
ok so even this modest mutability breaks things

PR #4227, opened 2025-03-17
Add adapt for TimeInterpolator

wsmoses:


glwagner:
https://github.com/CliMA/Oceananigans.jl/pull/4228#issuecomment-2729699497

@giordano hmm I think you're right, adding the fact that we need to use `Reactant` though to make this something that can't be adapted automatically

glwagner:
This has been confirmed elsewhere, so we will add a test in a subsequent PR

glwagner:
GPU Reactant test are now taking 2h so we maybe should split that it up into two groups.

PR #4231, opened 2025-03-17
Add a second Reactant test group

glwagner:
The GPU Reactant tests were reaching 2h which is a bit long.


glwagner:
This def helps

glwagner:
looks like `gpu-reactant-extension-tests-2` is 1h12m. That's similar to the distributed test time (1h14m) so I think this PR achieves its goal.

PR #4233, opened 2025-03-17
Try running the docs build with -O0

glwagner:
This may not help exec time, because we do run a bunch of examples. But experience shows that compile dominates, so let's see.


glwagner:
image

It doesn't really matter right now as docs only take 35 min.

glwagner:
I guess we are going to get doctest errors due to the different compilation. It's probably not worth pursuing right now.

PR #4239, opened 2025-03-18
(0.96.0) Move `NetCDFWriter` to an extension

tomchor:
Moves `NetCDFWriter` to an extension. This is a breaking change since scripts that use `NetCDFWriter` now need to be `using NCDatasets` prior to the `NetCDFWriter` call.

This should be merged only after https://github.com/CliMA/Oceananigans.jl/pull/4210 is merged, and tagged in the same release. (Note that at the moment this PR is set to merge to `tc/netcdfwriter`, because I'm using that branch as a base, but I'll change it after https://github.com/CliMA/Oceananigans.jl/pull/4210 is merged.)

cc @glwagner @ali-ramadhan


glwagner:
Please merge this ASAP and I will follow with an extension.

glwagner:
Sorry, now that we are at 0.96.0 we need this to be in by tomorrow, because we will have to continue tagging new versions unfortunately

glwagner:
@ali-ramadhan might use your help to resolve some merge conflicts here

glwagner:
Ah I see you're doing the extension here --- ok

glwagner:
Actually I am confused because this target is not `main`. What is the development strategy being taken here?

tomchor:
No worries, the merge conflicts should be easy and all the hard work is done here. I'll get started as soon as someone approves https://github.com/CliMA/Oceananigans.jl/pull/4210 (so that I know it won't change before merging).

glwagner:
I suggest closing this, finishing the `NetCDFWriter` / `JLD2Writer` PR ASAP. After that, I can take care of the extension which should be quick work.

Project.toml
```diff
@@ -42,12 +41,14 @@ Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
 MakieCore = "20f20a25-4f0e-4fdf-b5d1-57303727442b"
 Metal = "dde4c033-4e86-420c-a63e-0dd931031962"
+NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
```
glwagner, line 44: You need to add `NCDatasets` to the test targets. You also need to change the test files and the test init. You need to add `using NCDatasets` to the `NetCDFWriter` tests specifically.

glwagner:
> No worries, the merge conflicts should be easy and all the hard work is done here. I'll get started as soon as someone approves #4210 (so that I know it won't change before merging).

Note that the target for this PR is the branch on #4210.

glwagner:
You're going to have to split up `test_output_writers` into two files:

```julia
(base)  ~/Projects/Oceananigans.jl/test git:(glw/copyto) grep -r NetCDFOutputWriter ./*
./regression_tests/thermal_bubble_regression_test.jl:    nc_writer = NetCDFOutputWriter(model, outputs, filename=regression_data_filepath, schedule=IterationInterval(10))
./test_jld2_output_writer.jl:        ##### Time-averaging (same test as in NetCDFOutputWriter)
./test_lagrangian_particle_tracking.jl:    sim.output_writers[:particles_nc] = NetCDFOutputWriter(model,
./test_lagrangian_particle_tracking.jl:        sim.output_writers[:particles_nc] = NetCDFOutputWriter(model,
./test_netcdf_output_writer.jl:    simulation.output_writers[:netcdf] = NetCDFOutputWriter(model, fields(model);
./test_netcdf_output_writer.jl:    simulation.output_writers[:netcdf] = NetCDFOutputWriter(model, fields(model);
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:with_metrics_no_halos] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:no_metrics] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:sliced] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:with_metrics_and_halos] = NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:    simulation.output_writers[:no_metrics] = NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:    simulation.output_writers[:with_metrics_no_halos] = NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:    simulation.output_writers[:sliced] = NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:    simulation.output_writers[:with_halos] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:no_halos] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:sliced] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:with_halos] = NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:    simulation.output_writers[:no_halos] = NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:    simulation.output_writers[:sliced] = NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:    simulation.output_writers[:with_halos] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:sliced] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:with_halos] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    simulation.output_writers[:sliced] = NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, fields(model),
./test_netcdf_output_writer.jl:    nc_writer = NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:    nc_sliced_writer = NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:    nc_writer = NetCDFOutputWriter(model,
./test_netcdf_output_writer.jl:    ow = NetCDFOutputWriter(model, (; u=model.velocities.u);
./test_netcdf_output_writer.jl:    ow = NetCDFOutputWriter(model, (; u=model.velocities.u);
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, outputs;
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, outputs;
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model,
./test_netcdf_output_writer.jl:                NetCDFOutputWriter(
./test_netcdf_output_writer.jl:                NetCDFOutputWriter(
./test_netcdf_output_writer.jl:                NetCDFOutputWriter(
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model,
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model,
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model,
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, outputs,
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model,
./test_netcdf_output_writer.jl:    nc_writer = NetCDFOutputWriter(model, outputs;
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, outputs;
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, outputs;
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, outputs;
./test_netcdf_output_writer.jl:        NetCDFOutputWriter(model, outputs;
./test_output_writers.jl:    netcdf_output_writer = NetCDFOutputWriter(model, output,
./test_output_writers.jl:    if output_writer === NetCDFOutputWriter
./test_output_writers.jl:    nc_output_writer = NetCDFOutputWriter(model, nc_outputs,
./test_output_writers.jl:    simulation.output_writers[:nc] = NetCDFOutputWriter(model, nc_outputs,
./test_output_writers.jl:        for output_writer in (NetCDFOutputWriter, JLD2OutputWriter)
```

glwagner:
it also probably makes sense to switch lagrangain particle tracking tets to use JLD2

tomchor:
> > No worries, the merge conflicts should be easy and all the hard work is done here. I'll get started as soon as someone approves #4210 (so that I know it won't change before merging).
> 
> Note that the target for this PR is the branch on #4210.

I see that. I did that to avoid merge conflicts in anticipation of https://github.com/CliMA/Oceananigans.jl/pull/4210 being merged first.

I also just took care of the merge conflicts.

glwagner:
You also need to set up a dummy constructor for `NetCDFWriter` which will emit an error if users try to invoke it without writing `using NCDatasets`. You then need to extend this constructor in the extension in order to implement it properly. I suggest something like this in the file `OutputWriters.jl`:

```julia
function NetCDFWriter(args..., kw...)
    msg = "You must write using NCDatasets in order to construct `NetCDFWriter`"
    error(msg)
    return nothing
end
```

Later, the constructor needs to dispatch so that it doesn't overwrite:

```julia
function NetCDFWriter(model::AbstractModel, outputs; kw...)
```

If you have a better idea def implement that. This is just what I came up with.

tomchor:
> > > > No worries, the merge conflicts should be easy and all the hard work is done here. I'll get started as soon as someone approves #4210 (so that I know it won't change before merging).
> > > 
> > > 
> > > Note that the target for this PR is the branch on #4210.
> > 
> > 
> > I see that. I did that to avoid merge conflicts in anticipation of #4210 being merged first.
> > I also just took care of the merge conflicts.
> 
> Ok great. Are you going to switch this over to target merge to `main` once #4210 is merged then?

Yup.

glwagner:
> > You also need to set up a dummy constructor for `NetCDFWriter` which will emit an error if users try to invoke it without writing `using NCDatasets`. You then need to extend this constructor in the extension in order to implement it properly. I suggest something like this in the file `OutputWriters.jl`:
> > ```julia
> > function NetCDFWriter(args..., kw...)
> >     msg = "You must write using NCDatasets in order to construct `NetCDFWriter`"
> >     error(msg)
> >     return nothing
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Later, the constructor needs to dispatch so that it doesn't overwrite:
> > ```julia
> > function NetCDFWriter(model::AbstractModel, outputs; kw...)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If you have a better idea def implement that. This is just what I came up with.
> 
> The dummy construction is already done here:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/f3effb81d08acc2d8701b97ef75d85d85c7055bd/src/OutputWriters/netcdf_writer.jl#L5-L26
> 
> Which is then imported in
> 
> https://github.com/CliMA/Oceananigans.jl/blob/f3effb81d08acc2d8701b97ef75d85d85c7055bd/ext/OceananigansNCDatasetsExt.jl#L29
> 
> and subsequently expanded upon with a new method. I also couldn't figure out a better way to do it and, according to the threads I read at least, this is best practice.
> 
> It would be nice to implement some sort of useful error message for users who use `NetCDFWriter` but forget to include `NCDatasets`. But it's not clear what the best way to do that is, and we can leave it for a future PR since it won't be a breaking change and we're short on time.

I think this is a good design.

glwagner:
> It would be nice to implement some sort of useful error message for users who use NetCDFWriter but forget to include NCDatasets. But it's not clear what the best way to do that is

You can use the pattern I suggested.

glwagner:
I'm marking this as ready to review. It needs to go in very soon.

glwagner (approved):

tomchor:
> > It would be nice to implement some sort of useful error message for users who use NetCDFWriter but forget to include NCDatasets. But it's not clear what the best way to do that is
> 
> You can use the pattern I suggested.

I don't know if that pattern is the way to go since it's a catch-all. Using that means that, if a user starts `using NCDatasets` and misses the `NetCDFWriter` signature for whatever reason, not only the useful `Method not found error` will not be displayed, but an inaccurate message telling the user to add `NCDatasets` will take its place.

src/OutputWriters/netcdf_writer.jl
```diff
@@ -760,669 +25,10 @@ mutable struct NetCDFWriter{G, D, O, T, A, FS, DN} <: AbstractOutputWriter
...
-
-    dims = dimensions[name]
-    FT = eltype(array_type)
-    defVar(dataset, name, FT, (dims..., "time"); deflatelevel, attrib)
-
+function NetCDFWriter(model, outputs; kw...)
+    msg = string("Building Oceananigans.OutputWriters.NetCDFWriter requires `using NCDatasets`.", '\n',
+                 "If you think you are already `using NCDatasets` , check for warnings or errors.")
+                      
+    error(msg)
```
tomchor, line 32: Again, I think this way to displaying the error message isn't ideal. I'd rather we rely on julia's `Method not found` error and just write `function NetCDFWriter end`.

glwagner: Are you talking about required kwargs? We can add those here

glwagner: Or do you mean a case where they pass an argument other than `model`? That's also easy to deal with. We can make this a `MethodError`

glwagner:
```julia
julia> function h(args...; kw...)
           @warn "Did you load a package?"
           throw(MethodError(h, args))
           end

julia> h(1)
┌ Warning: Did you load a package?
└ @ Main REPL[12]:2
ERROR: MethodError: no method matching h(::Int64)

Closest candidates are:
  h(::Any...; kw...)
   @ Main REPL[12]:1

Stacktrace:
 [1] h(args::Int64; kw::@Kwargs{})
   @ Main ./REPL[12]:3
 [2] h(args::Int64)
   @ Main ./REPL[12]:1
 [3] top-level scope
   @ REPL[14]:1
```

tomchor: I'll mark this as solved. I was under the impression that your way of doing things masked the useful Julia messages if the user makes a mistake after `using NCDatasets`. But I confirmed that's not the case and the message is still useful. See my previous comment.

glwagner: I guess my only comment is that "relying on MethodError" and "writing `function f end` are not necessarily simultaneously required things. You can do one, or the other, or both.

glwagner:
@tomchor don't merge any more if you can --- if the tests pass even when its not updated, its usually fine to merge. We can fix the very rare odd issue if we need to after.

navidcy:
I must say this is great job

navidcy:
should we move JLD2writer to an extension?

tomchor:
Yes, I suggested this at some point in this whole NetCDF refactor and @glwagner seemed to agree.

glwagner:
It's not a horrible idea, because I think it would help us dial in the design of `FieldTimeSeries`. I think that JLD2 is much lighter weight than NetCDF which is something to consider. One of the motivations for moving the NetCDF stuff to extension is that there are some machines where NetCDF binaries are not available, and before having the extension Oceananigans was not usable on them without changing the source code. We haven't experienced that with JLD2.

Another point is that the changes required to move JLD2 to an extension could smoothen the implementation of the NetCDF FieldTimeSeries.

src/OutputWriters/netcdf_writer.jl
```diff
@@ -760,669 +25,7 @@ mutable struct NetCDFWriter{G, D, O, T, A, FS, DN} <: AbstractOutputWriter
...
-
-interpolate_u(model) = interpolate!(coarse_u, model.velocities.u)
-outputs = (; u = interpolate_u)
-
-output_writer = NetCDFWriter(model, outputs;
-                             grid = coarse_grid,
-                             filename = "coarse_u.nc",
-                             schedule = IterationInterval(1))
-```
-"""
```
navidcy, line 972: Was this docstring lost?
I can't see it currently in the Docs/Appendix/Library-API

tomchor: I think you're right! I also can't find it in the code.

PR #4244, opened 2025-03-18
Better name than RankConnectivity

glwagner:
@avik-pal asked me "what's RankConnectivity". I was puzzled at first. Then I looked and I said, "oh, it's the neighboring ranks". So we are changing the name of this object to represent the thing that it is.


glwagner:
@avik-pal if you object because this might disrupt your PR I can wait on it :-D

avik-pal:
go for it, i will rebase

glwagner:
philosophically this helps because now we can understand "connectivity" as the general term for "how things are arranged on the network"; with MPI the key info is local connecctivity, but for sharding the info is global. That's my two cents ✌️

Issue #3485, opened 2024-02-27
Are `WindowedTimeAverage`s working properly when picking up simulations?


tomchor:
I've been using `WindowedTimeAverage`s for my simulations (by setting `schedule = AveragedTimeInterval(...)` in a `NetCDFOutputWriter`). I noticed that whenever I run out of walltime and have to checkpoint my simulations, when I pick them up again I get the following warning for each of the time-averaged outputs:

```
┌ Warning: Returning a WindowedTimeAverage before the collection period is complete.
└ @ Oceananigans.OutputWriters /glade/work/tomasc/.julia/packages/Oceananigans/3LHMs/src/OutputWriters/windowed_time_average.jl:201
```
(which comes from [this call](https://github.com/CliMA/Oceananigans.jl/blob/f2511962ca15f3aaf87d2571e3551e59dc05c694/src/OutputWriters/windowed_time_average.jl#L196-L204).)

Does this mean that the time averages aren't being correctly calculated after picking up? I tried following the trail to figure it out but couldn't determine the answer...

glwagner:
I don't think this is done correctly right now. Somehow the `Checkpointer` needs to know about the simulation for this to work. But right now it only saves model properties.

tomchor:
Ah, I see. Sounds like it wouldn't be trivial to add that support.

I guess a workaround to avoid partially-averaged results when picking up would be to set the `Checkpointer` to only write checkpoints when the `TimeAveraged` results are also written. I'm not sure what that would do to other (more frequent) outputs though, since it'd potentially try to write some time steps twice (and not in monotonic ordering)...

glwagner:
> Ah, I see. Sounds like it wouldn't be trivial to add that support.
> 
> I guess a workaround to avoid partially-averaged results when picking up would be to set the `Checkpointer` to only write checkpoints when the `TimeAveraged` results are also written. I'm not sure what that would do to other (more frequent) outputs though, since it'd potentially try to write some time steps twice (and not in monotonic ordering)...

There are two things. One is to fix the flow of information... that's probably pretty easy because we can either 1) make `Checkpointer` a callback or 2) change `write_output!` to have the syntax `write_output(writer, simulation)` here:

https://github.com/CliMA/Oceananigans.jl/blob/643b484e81e0aeb038b3038266912ad051bce9b8/src/Simulations/run.jl#L147

then with a fallback `write_output!(writer, sim) = write_output!(writer, sim.model)`, very little has to change...

[...]

PR #4250, opened 2025-03-20
Checkpointer updates

navidcy:


navidcy:
@glwagner the CI machine seems out of space?

```julia
ERROR: Unable to automatically download/install artifact 'CUDA_Runtime' from sources listed in '/storage5/buildkite-agent/.julia-21600/packages/CUDA_Runtime_jll/iU7vK/Artifacts.toml'.
Sources attempted:
- https://pkg.julialang.org/artifact/59c3479d656c293c66e1a7a26fcb5d860dbe6235
    Error: SystemError: flush: No space left on device
- https://github.com/JuliaBinaryWrappers/CUDA_Runtime_jll.jl/releases/download/CUDA_Runtime-v0.16.1+0/CUDA_Runtime.v0.16.1.x86_64-linux-gnu-cuda+12.8.tar.gz
    Error: SystemError: flush: No space left on device
```

tomchor:
Do the changes here have any impact on https://github.com/CliMA/Oceananigans.jl/issues/3485?

i.e. will it make it easier or harder to make `Checkpointer` play with time-averaged fields?

glwagner:
We should probably at the least address the problem of model vs simulation checkpointing here

navidcy:
@tomchor it shouldn't affect the time-averaging issue in either way I think.

navidcy:
@glwagner, @tomchor so I see in https://github.com/CliMA/Oceananigans.jl/issues/3485 the discussion/proposal is to have `write_output!(sim::Simulation, ...)` instead of `write_output!(model::AbstractModel, ...)`.

We can do that here. But then it means that every time somebody defines a new model (and I'm thinking of ClimaOcean.jl and beyond) then we need to define a new method for `write_output!(sim::Simulation{MyNewModel}, ...)` instead of `write_output!(model::MyNewModel, ...)`, right?

tomchor:
> @glwagner, @tomchor so I see in #3485 the discussion/proposal is to have `write_output!(sim::Simulation, ...)` instead of `write_output!(model::AbstractModel, ...)`.
> 
> We can do that here. But then it means that every time somebody defines a new model (and I'm thinking of ClimaOcean.jl and beyond) then we need to define a new method for `write_output!(sim::Simulation{MyNewModel}, ...)` instead of `write_output!(model::MyNewModel, ...)`, right?

I think so, but wouldn't a fallback like `write_output!(simulation) = write_output!(simulation.model)` take care of this?

glwagner:
> > @glwagner, @tomchor so I see in #3485 the discussion/proposal is to have `write_output!(sim::Simulation, ...)` instead of `write_output!(model::AbstractModel, ...)`.
> > We can do that here. But then it means that every time somebody defines a new model (and I'm thinking of ClimaOcean.jl and beyond) then we need to define a new method for `write_output!(sim::Simulation{MyNewModel}, ...)` instead of `write_output!(model::MyNewModel, ...)`, right?
> 
> I think so, but wouldn't a fallback like `write_output!(simulation) = write_output!(simulation.model)` take care of this?

Yeah no I think we can have a pretty small PR that simply changes the meaning of `write_output!`. A few lines are all that's needed:

```julia
write_output!(writer, sim::Simulation) = write_output!(writer, sim.model)
```

and then I think 2-3 places in `Simulations` where we call `write_output!`.

I'd prefer if that was self-contained, we should avoid the temptation to make PRs bigger than needed.

The trickier part (but not the hardest thing ever in the world) is to design an interface for checkpointing callbacks, diagnostics, and output writers.

src/OutputWriters/checkpointer.jl
```diff
@@ -5,32 +5,60 @@ using Oceananigans: fields, prognostic_fields
...
 end
 
-function default_checkpointed_properties(model)
-    properties = [:grid, :particles, :clock, :timestepper]
-    #if has_ab2_timestepper(model)
-    #    push!(properties, :timestepper)
-    #end
-    return properties
-end
+default_checkpointed_properties(model) = [:grid, :particles, :clock, :timestepper]
```
simone-silvestri, line 19: I would add the timestepper only if we use AB2, in the end RK3 does not require it.

src/OutputWriters/checkpointer.jl
```diff
@@ -7,30 +7,58 @@ using Oceananigans.TimeSteppers: QuasiAdamsBashforth2TimeStepper
...
 
 has_ab2_timestepper(model) = try
     model.timestepper isa QuasiAdamsBashforth2TimeStepper
 catch
     false
 end
 
+# Certain properties are required for `set!` to pickup from a checkpoint.
+function required_checkpointed_properties(model)
+    properties = [:grid, :particles, :clock]
```
simone-silvestri, line 29:
```suggestion
    properties = [:grid, :clock]
```
I wouldn't say `:particles` are necessary. However, also here we can make the default nothing and extend the `required_checkpointed_properties` in models (after #4530 is merged)

navidcy:
This PR includes few useful updates/refactors for Clock and I'm quite tempted to remove the Checkpointer things, merge and continue with Checkpointer in a new PR. What do you think?

tomchor:
> This PR includes few useful updates/refactors for Clock and I'm quite tempted to remove the Checkpointer things, merge and continue with Checkpointer in a new PR. What do you think?

I'm all for simple PRs!

navidcy:
https://github.com/CliMA/Oceananigans.jl/pull/4637 attempts to separate the Clock enhancements

glwagner:
I think this is superceded by #4892, so we can close it. @navidcy please re-open if this is an error!

PR #4260, opened 2025-03-23
Try removing `LeftBias` and `RightBias` in favor of `true` and `false`

simone-silvestri:
might help with the issue found here
https://github.com/PRONTOLab/GB-25/blob/min-wm/ocean-climate-simulation/minimal_calculate_Gv.jl


simone-silvestri:
confirming that, indeed, with this change, https://github.com/PRONTOLab/GB-25/blob/min-wm/ocean-climate-simulation/minimal_calculate_Gv.jl yields same results for reactant and CPU architectures

simone-silvestri:
Probably not needed anymore

wsmoses:
yeah it's no longer necessary for correctness

glwagner:
I was wondering if this is a positive improvement though, because it removes one level of ifelse. Maybe not important in the near term, but who knows what comes down the line.

glwagner:
i don't think we want this, @simone-silvestri open if i'm wrong

PR #4262, opened 2025-03-23
feat: Distributed Reactant [Take II]

avik-pal:


ext/OceananigansReactantExt/Models.jl
```diff
@@ -1,9 +1,54 @@
...
+
+const ReactantHFSM{TS, E} = Union{
+    HydrostaticFreeSurfaceModel{TS, E, <:ReactantState},
+    HydrostaticFreeSurfaceModel{TS, E, <:Distributed{<:ReactantState}},
+}
 
 initialization_update_state!(::ReactantModel; kw...) = nothing
 
+# We may need this but not sure:
+maybe_extend_halos(TX, TY, grid::ReactantGrid, ::FixedSubstepNumber) = grid
```
simone-silvestri, line 24: This should be removed
```suggestion
```

PR #4266, opened 2025-03-24
Add validation for `condition` kwarg in `ConditionalOperation`

### LINKED ISSUE #4265: Conditional operations with `condition=cond::Array` should warn if `cond` does not match the size of the field
As explained here:

https://github.com/CliMA/Oceananigans.jl/blob/d81ad296475669826cb898a5320a25b99499426c/src/AbstractOperations/conditional_operations.jl#L22-L80

when we construct a conditional operation we can provide the `condition` kwarg with either a function with args `(I, j, k, grid, field)` that returns true or false _or_ an array.

However, if we provide with an array that is missing one dimension compared to the field then unexpected behaviour ensues without any warning. What I believe happens is that the array is multiplied over the first index of the missing dimension. Intuitively I'd expect that the conditional array is broadcasted across the missing dimension.

In either case, we should either update the docstring or spit out an informational warning to the user.

I demonstrate this behaviour below. I create a 3D grid and a field of ones and integrate it to compute the volume of a grid. Then I integrate it over half the grid using a function and a 2D array and 3D array. With a 2D array the answer I get is not `1/2 * volume` but rather `1/2 * volume * grid.z.Δᵃᵃᶠ[1] / grid.Lz `.

```Julia
using Oceananigans

grid = RectilinearGrid(size=(4, 3, 4), x=(0, 2), y=(-1, 2), z=[1, 1.24, 1.5, 2.5, 3])
@show volume = grid.Lx * grid.Ly * grid.Lz

c = CenterField(grid)

set!(c, 1)

@info "without condition"

@show compute!(Field(Integral(c)))[1, 1, 1] ≈ volume

@info "provide function to condition"

using Oceananigans.Grids: xnode
cond(i, j, k, grid, c) = xnode(i, j, k, grid, Center(), Center(), Center()) ≥ 1
compute!(Field(Integral(c, condition=cond)))[1, 1, 1] ≈ grid.Lx/2 * grid.Ly * grid.Lz

@info "provide 2D array to condition"

cond_array2D = trues(grid.Nx, grid.Ny)
cond_array2D[3:grid.Nx, :] .= false
cond_array2D

@show compute!(Field(Integral(c, condition=cond_array2D)))[1, 1, 1] ≈ 1/2 * volume
@show compute!(Field(Integral(c, condition=cond_array2D)))[1, 1, 1] ≈ 1/2 * (grid.z.Δᵃᵃᶠ[1] / grid.Lz) * volume

cond_array3D = trues(grid.Nx, grid.Ny, grid.Nz)
cond_array3D[3:grid.Nx, :, :] .= false

@info "provide 3D array to condition"
@show compute!(Field(Integral(c, condition=cond_array3D)))[1, 1, 1] ≈ 1/2 * grid.Lx * grid.Ly * grid.Lz
```

returns

```Julia
volume = grid.Lx * grid.Ly * grid.Lz = 12.0
[ Info: without condition
(compute!(Field(Integral(c))))[1, 1, 1] ≈ volume = true
[ Info: provide function to condition
[ Info: provide 2D array to condition
(compute!(Field(Integral(c, condition = cond_array2D))))[1, 1, 1] ≈ (1 / 2) * volume = false
(compute!(Field(Integral(c, condition = cond_array2D))))[1, 1, 1] ≈ (1 / 2) * (grid.z.Δᵃᵃᶠ[1] / grid.Lz) * volume = true
[ Info: provide 3D array to condition
(compute!(Field(Integral(c, condition = cond_array3D))))[1, 1, 1] ≈ (1 / 2) * grid.Lx * grid.Ly * grid.Lz = true
true
```

cc found with @taimoorsohail

navidcy:
This PR validates the `condition` kwarg in `ConditionalOperation` and spits out an informative error message in case user provides an array that is not 3D.

Closes #4265


src/AbstractOperations/conditional_operations.jl
```diff
@@ -86,11 +89,22 @@ function ConditionalOperation(operand::AbstractField;
     return ConditionalOperation{LX, LY, LZ}(operand, func, operand.grid, condition, mask)
 end
 
+validate_condition(cond::Function, ::ConditionalOperation) = cond
+
+function validate_condition(cond::AbstractArray, c::ConditionalOperation)
+    if ndims(cond) != 3
+        throw(ArgumentError("when kwarg condition::Array requires a 3D array of size $(size(c))"))
```
glwagner, line 96: i dont totally understand this, what do you mean "when" kwarg. Why just say "The keyword argument condition::AbstractArray requires..."

Note this does not require an array specifically. It can be a `Field`, or an `AbstractOperation` (or a `CuArray`)

glwagner: the dispatch is specifically on `AbstractArray`, so the error is correct to point out exactly that

src/AbstractOperations/conditional_operations.jl
```diff
@@ -86,11 +89,22 @@ function ConditionalOperation(operand::AbstractField;
     return ConditionalOperation{LX, LY, LZ}(operand, func, operand.grid, condition, mask)
 end
 
+validate_condition(cond::Function, ::ConditionalOperation) = cond
+
+function validate_condition(cond::AbstractArray, c::ConditionalOperation)
```
glwagner, line 94:
```suggestion
function validate_condition(cond::AbstractArray, operand)
```

src/AbstractOperations/conditional_operations.jl
glwagner, line 96:
```suggestion
        throw(ArgumentError("when kwarg condition::Array requires a 3D array of size $(size(operand))"))
```

not sure if this was intended but since this validation is in the constructor we don't have the `ConditionalOperation` yet

glwagner: oh, huh, i got confused

src/AbstractOperations/conditional_operations.jl
```diff
@@ -86,11 +89,22 @@ function ConditionalOperation(operand::AbstractField;
...
+    end
+    return cond
+end
+
 function ConditionalOperation(c::ConditionalOperation;
                               func = c.func,
                               condition = c.condition,
                               mask = c.mask)
 
+    condition = validate_condition(condition, c)
```
glwagner, line 106: I'm confused how the input is `ConditionalOperation`. Possibly this is cleaned up in #3794

navidcy: hm.... yes, I got it from there....

src/AbstractOperations/conditional_operations.jl
glwagner: i guess i was confused ignore this

navidcy:
@glwagner have another look?

src/AbstractOperations/conditional_operations.jl
navidcy: true! good catch!

navidcy: hm... should I just leave the `cond` untyped?

navidcy: Oh I got what you mean! you are referring to the error being Array-specific!

src/AbstractOperations/conditional_operations.jl
navidcy: I changed it ;)

src/AbstractOperations/conditional_operations.jl
```diff
@@ -95,10 +95,10 @@ validate_condition(cond::Function, ::AbstractField) = cond
 
 function validate_condition(cond::AbstractArray, operand::AbstractField)
     if ndims(cond) != 3
-        throw(ArgumentError("The keyword argument condition::AbstractArray requires a 3D array of size $(size(operand))"))
+        throw(ArgumentError("The keyword argument condition::AbstractArray requires size $(size(operand))"))
     end
     if eltype(cond) !== Bool
-        throw(ArgumentError("The keyword argument condition::AbstractArray requires an array of booleans"))
+        throw(ArgumentError("The keyword argument condition::AbstractArray requires eltype(condition) == Bool"))
```
glwagner, line 101: This isn't really true because the actual requirement is that `ifelse` can be evaluated. But if we want to make this restriction we can. I'm a little worried it might mess things up, eg if you use `Reactant` you end up with `TracedRNumber{Bool}` (which also has an `ifelse` defined for it)

navidcy: When I try providing an array of ones and zeros I get:

```Julia
ERROR: LoadError: MethodError: no method matching ifelse(::Float64, ::Float64, ::Int64)
The function `ifelse` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  ifelse(::Bool, ::Any, ::Any)
   @ Base essentials.jl:796

Stacktrace:
  [1] getindex
    @ ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Oceananigans.jl-v2/src/AbstractOperations/conditional_operations.jl:118 [inlined]
  [2] _getindex
    @ ./abstractarray.jl:1358 [inlined]
  [3] getindex
    @ ./abstractarray.jl:1312 [inlined]
  [4] macro expansion
    @ ./reducedim.jl:280 [inlined]
  [5] macro expansion
    @ ./simdloop.jl:77 [inlined]
...
```

glwagner: right but if you included another package like ```using GreatIfElseTools` then you can extend `ifelse`

So if you restrict inputs you want to be sure you really don't want anyone to use it any other way

glwagner: that's a general Julia concept that we don't want to over-restrict inputs. I don't know if there is an abstract boolean, that could be used...

navidcy: Agreed; I removed the bool validation.

navidcy:
Thought I check whether things are as expected for an immersed boundary grid.

Ran onto this; is this a bug?

```Julia
using Oceananigans

underlying_grid = RectilinearGrid(size=(6, 4), x=(0, 2.2), z=[1, 1.24, 1.8, 2.5, 3], topology=(Periodic, Flat, Bounded))

bottom_height = underlying_grid.z.cᵃᵃᶠ[1] * ones(underlying_grid.Nx,)
bottom_height[1:3] .= underlying_grid.z.cᵃᵃᶠ[4]
grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom_height))

c = CenterField(grid)
set!(c, 1)

cond(i, j, k, grid, c) = Oceananigans.Grids.xnode(i, j, k, grid, Center(), Center(), Center()) ≤ grid.underlying_grid.Lx/2
compute!(Field(Integral(c, condition=cond)))[1, 1, 1] ≈ (underlying_grid.z.cᵃᵃᶠ[underlying_grid.Nz+1] - underlying_grid.z.cᵃᵃᶠ[4]) * underlying_grid.Lx/2
```

```Julia
ERROR: MethodError: no method matching cond(::Int64, ::Int64, ::Int64, ::ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded, RectilinearGrid{…}, GridFittedBottom{…}, Nothing, Nothing, CPU})

Closest candidates are:
  cond(::Any, ::Any, ::Any, ::Any, ::Any)
   @ Main REPL[12]:1

Stacktrace:
  [1] evaluate_condition
    @ ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Oceananigans.jl-v2/src/AbstractOperations/conditional_operations.jl:120 [inlined]
  [2] evaluate_condition
    @ ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Oceananigans.jl-v2/src/ImmersedBoundaries/immersed_reductions.jl:37 [inlined]
  [3] getindex
    @ ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Oceananigans.jl-v2/src/AbstractOperations/conditional_operations.jl:115 [inlined]
  [4] _getindex
    @ ./abstractarray.jl:1340 [inlined]
  [5] getindex
    @ ./abstractarray.jl:1290 [inlined]
  [6] macro expansion
    @ ./reducedim.jl:310 [inlined]
  [7] macro expansion
    @ ./simdloop.jl:77 [inlined]
  [8] _mapreducedim!(f::typeof(identity), op::typeof(Base.add_sum), R::SubArray{…}, A::Oceananigans.AbstractOperations.ConditionalOperation{…})
    @ Base ./reducedim.jl:309
  [9] mapreducedim!(f::Function, op::Function, R::SubArray{…}, A::Oceananigans.AbstractOperations.ConditionalOperation{…})
    @ Base ./reducedim.jl:326
 [10] sum!(f::Function, r::SubArray{…}, A::Oceananigans.AbstractOperations.ConditionalOperation{…})
    @ Base ./reducedim.jl:1036
 [11] sum!(r::Field{…}, a::Oceananigans.AbstractOperations.ConditionalOperation{…}; condition::Nothing, mask::Int64, kwargs::@Kwargs{})
    @ Oceananigans.Fields ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Oceananigans.jl-v2/src/Fields/field.jl:697
 [12] sum!
    @ ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Oceananigans.jl-v2/src/Fields/field.jl:691 [inlined]
 [13] compute!
    @ ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Oceananigans.jl-v2/src/Fields/scans.jl:69 [inlined]
 [14] compute!(field::Field{…})
    @ Oceananigans.Fields ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Oceananigans.jl-v2/src/Fields/scans.jl:65
 [15] top-level scope
    @ REPL[13]:1
```

By modifying

https://github.com/CliMA/Oceananigans.jl/blob/bf0140e544658158dd17dfea62a814ea902cef2f/src/ImmersedBoundaries/immersed_reductions.jl#L37

to include the `co::ConditionalOperation` arg

```julia
    return !immersed & evaluate_condition(condition.func, i, j, k, ibg, co, args...)
```

makes the above MWE work!

navidcy:
Thinking to add these MWE as test, the one with immersed boundary at least.

navidcy:
Oh I see that #3794 aims to extend the conditional operations functionality to immersed boundaries... So perhaps the immersed boundary example fits there!

src/AbstractOperations/conditional_operations.jl
```diff
@@ -82,10 +85,21 @@ function ConditionalOperation(operand::AbstractField;
...
+    condition = validate_condition(condition, operand)
+
     LX, LY, LZ = location(operand)
     return ConditionalOperation{LX, LY, LZ}(operand, func, operand.grid, condition, mask)
 end
 
+validate_condition(cond, ::AbstractField) = cond # fallback
+
+function validate_condition(cond::AbstractArray, operand::AbstractField)
+    if ndims(cond) != 3
```
taimoorsohail, line 97: Should we instead check if 

```size(cond) == size(operand)```?

`operand` will always have three dimensions so the above should work for both the 3D check but also the size?

taimoorsohail: Especially as your Argument error below says the condition must be the size of the operand

navidcy: hm... I think that's a good point
@glwagner?

glwagner: makes sense to me!

navidcy:
could somebody approve (if you are indeed approving this?)

taimoorsohail (approved):

glwagner (approved):

navidcy:
what's this error? 
seems unrelated to this PR, right?

https://buildkite.com/clima/oceananigans-distributed/builds/6677#0195d135-3deb-488c-94f1-787291eb7060/207-1335

navidcy:
I think https://github.com/CliMA/Oceananigans.jl/pull/4272 fixed it

simone-silvestri (approved):

PR #4178, opened 2025-03-07
Improve definition of `@apply_regionally`

giordano:
By interpolating the function names we can call this macro also from other modules.

PR #4270, opened 2025-03-25
[Reactant] Materialize traced array

wsmoses:
This reduces the size of constants that get baked in, potentially improving compile/runtime.

The only users of the abstractarray interface here are reactant


src/Grids/inactive_node.jl
```diff
@@ -2,22 +2,24 @@
...
         if array
-            return :(($side .< 1) .| ($side .> grid.$dim))
+            return :((ReactantCore.materialize_traced_array($side) .< 1) .| (ReactantCore.materialize_traced_array($side) .> grid.$dim))
         else
             return :(($side < 1) | ($side > grid.$dim))
         end
     elseif Topo == :LeftConnected
         if array
-            return :(($side .> grid.$dim))
+            return :((ReactantCore.materialize_traced_array($side) .> grid.$dim))
```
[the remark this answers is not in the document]
glwagner: yes its necessary

src/Grids/inactive_node.jl
```diff
@@ -2,22 +2,24 @@
...
         else
             return :(($side > grid.$dim))
         end
     else # RightConnected
         if array
-            return :(($side .< 1))
+            return :((ReactantCore.materialize_traced_array($side) .< 1))
```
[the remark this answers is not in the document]
glwagner: this only happens during compilation so your points aren't really relevant

glwagner (approved):

src/Grids/inactive_node.jl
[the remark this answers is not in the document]
glwagner: @gemini-code-assist don't you have a counterargument

glwagner:
@wsmoses ready to merge this and lump into 0.96.3, or do you want to wait?

glwagner:
Distributed Tripolar Grid failure is expected

src/Grids/inactive_node.jl
```diff
@@ -2,22 +2,24 @@
 const c = Center()
 const f = Face()
 
+using ReactantCore
+
 function build_condition(Topo, side, dim, array::Bool) 
     if Topo == :Bounded 
         if array
-            return :(($side .< 1) .| ($side .> grid.$dim))
+            return :((ReactantCore.materialize_traced_array($side) .< 1) .| (ReactantCore.materialize_traced_array($side) .> grid.$dim))
```
giordano, line 10:
```suggestion
            return :(($(ReactantCore).materialize_traced_array($side) .< 1) .| ($(ReactantCore).materialize_traced_array($side) .> grid.$dim))
```

src/Grids/inactive_node.jl
giordano, line 16:
```suggestion
            return :(($(ReactantCore).materialize_traced_array($side) .> grid.$dim))
```

src/Grids/inactive_node.jl
giordano, line 22:
```suggestion
            return :(($(ReactantCore).materialize_traced_array($side) .< 1))
```

giordano (commented):
Similar to #4178, interpolating the module lets you expand the macro also without having `ReactantCore` in scope, which is useful for debugging.

Project.toml
```diff
@@ -27,6 +27,7 @@ OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+ReactantCore = "a3311ec8-5e00-46d5-b541-4f83e724a433"
```
navidcy, line 30: do we need this as main dependency or just weak dependency?

giordano: Yes, because it's used in the main package, not an extension.  Note that `ReactantCore` is a much smaller dependency than `Reactant` itself, [it's fairly small](https://github.com/EnzymeAD/Reactant.jl/blob/main/lib/ReactantCore/src/ReactantCore.jl) and [it has only two dependencies](https://github.com/EnzymeAD/Reactant.jl/blob/0e7287bd1cf543313c0d48f702cce009aaa2a952/lib/ReactantCore/Project.toml#L6), chances are Oceananigans already indirectly depended on at least one of them through other packages.

navidcy: No worries -- great!
Just wanted to ensure that it wasn't there by accident ;)

PR #4272, opened 2025-03-25
Sharding for LatitudeLongitudeGrid and extend Distributed methods for `Distributed{<:ReactantState}`

simone-silvestri:
@avik-pal let me know if this would work or if I misunderstood sharding. 
For the `LatitudeLongitudeGrid` the idea is to have a grid that is similar to the serial grid but shards the underlying metrics. Note that the metrics can have different dimensionalities based on the stretching of the grid.

We also need an extension of the distributed-specific methods for sharding. I have added them in this PR (for example, `partition` and `construct_global_array`). 
If I understand correctly, all the partitioning and reconstruction should be handled by sharding, so all these methods for a `Distributed{<:ReactantState}` architecture fall back to the serial methods (which basically do nothing)


ext/OceananigansReactantExt/ShardedGrids.jl
```diff
@@ -8,6 +14,109 @@ partition_coordinate(c::AbstractVector, n, ::Oceananigans.Distributed{<:Reactant
...
+
+# We assume everything is already synchronized for a sharded architecture
+barrier!(::Oceananigans.Distributed{<:ReactantState}) = nothing
+
+# Reductions are handled by the Sharding framework
+all_reduce(op, val, ::Oceananigans.Distributed{<:ReactantState}) = val
+
+# No need for partitioning and assembling of arrays supposedly
+partition(A::AbstractArray, ::Oceananigans.Distributed{<:ReactantState}, local_size) = A
+construct_global_array(A::AbstractArray, ::Oceananigans.Distributed{<:ReactantState}, local_size) = A
```
simone-silvestri, line 31: @avik-pal these are the sharding extension to the distributed-specific methods

Project.toml
```diff
@@ -13,6 +13,7 @@ Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 GPUArrays = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
+GPUCompiler = "61eb1bfa-7361-4325-ad38-22787b887f55"
```
giordano, line 16: This is fixed in Enzyme now, no need to involve GPUCompiler anymore.

glwagner: looks like tests are still failing, but may be we need to wait a few more min?

giordano: Where are you looking at?  I only see https://github.com/CliMA/Oceananigans.jl/actions/runs/14063874231/job/39381278166?pr=4272#step:6:284, which is a problem in `NCDatasets`, unrelated to Enzyme/Reactant/GPUCompiler

giordano: Ah, https://github.com/CliMA/Oceananigans.jl/actions/runs/14063874231/job/39381277641?pr=4272#step:6:288 just failed, that's a bug in Reactant

ext/OceananigansReactantExt/ShardedGrids.jl
```diff
@@ -1,16 +1,128 @@
...
+                                                         Reactant.to_rarray(z), # Intentionally not sharded
+                                                         (nothing for i=1:10)..., FT(radius))
+
+    if !precompute_metrics
+        return preliminary_grid
+    end
+       
+    # Note! This step requires a kernel that launches on a `ReactantState` architecture.
+    # Would there be issues?
+    grid = with_precomputed_metrics(preliminary_grid)
```
glwagner, line 88:
```suggestion
    grid = @jit with_precomputed_metrics(preliminary_grid) 
```

not sure if this would help

test/distributed_tests_utils.jl
```diff
@@ -128,7 +152,7 @@ function run_tripolar_simulation(grid)
     set!(model, c = ηᵢ, η = ηᵢ)
 
     if architecture(grid) isa ReactantState || child_architecture(grid) isa ReactantState  
-        r_loop! = @compile sync=true raise=true loop!(model)
+        r_loop! = @compile sync=true raise=true reactant_loop!(model)
```
glwagner, line 155: Another test that would be sufficient is to compile a single time step and then loop over that, like

```julia
rstep! = @compile sync=true raise=true time_step!(model, dt)
for n = 1:100
    rstep!(model, dt)
end
```

simone-silvestri: yeah, probably better

ext/OceananigansReactantExt/Fields.jl
```diff
@@ -39,4 +40,8 @@ end
     @inbounds u[i, j, k] = v[i, j, k]
 end
 
+const ShardedDistributedField{LX, LY, LZ, O} = Field{LX, LY, LZ, O, <:ShardedGrid}
+
+_set_to_field!(u::ShardedDistributedField, v::ShardedDistributedField) = copyto!(parent(u), parent(v))
```
glwagner, line 45: @giordano attempted fix

PR #4275, opened 2025-03-25
Remove `multi_region_output_writers.jl`

tomchor:
Based on @glwagner's comment: https://github.com/CliMA/Oceananigans.jl/pull/4210#issuecomment-2731499084

I don't see this being used/tested anywhere so tests should pass as they are.


tomchor:
@glwagner can you confirm that this is indeed desirable?

simone-silvestri:
I guess if the tests pass, we can remove the file.
maybe an easy test would show if we need this file or not. I am not sure the output writing infrastructure is tested.
Maybe something like
```julia
grid = RectilinearGrid(size = (10, 10, 10), extent = (1, 1, 1))
grid = MultiRegionGrid(grid, partition=XPartition(2))
model = HydrostaticFreeSurfaceModel(; grid)
sim = Simulation(model; kw...)

sim.output_writers[:test] = JLD2Writer(args..., kwargs...)
run!(simulation)
```

PR #4276, opened 2025-03-25
Solving Poisson equation for `p * dt` instead of `p`

### LINKED ISSUE #4067: `ConjugateGradientPoissonSolver` fails to converge/is numerically unstable for extremely small timesteps
This MWE is slightly messy due to many moving parts, and I am still left slightly confused about the way to understand the problem, so any help is greatly appreciated.

When the `ConjugateGradientPoissonSolver` is taking extremely small timesteps which are inevitable if some `OutputWriter` with `TimeInterval` is used, it can throw `NaN`s due to numerical instability. My hypothesis is that the numerical instability is caused by extremely strong pressure values when timestep becomes extremely small. Below is the setup of the MWE:

```julia
using Oceananigans
using Oceananigans.Solvers: ConjugateGradientPoissonSolver, FFTBasedPoissonSolver
using Statistics
using Printf

Nx, Nz = 10, 10

arch = CPU()
kw = (size=(Nx, Nz), x=(0, 1), z=(-1, 0), topology=(Periodic, Flat, Bounded))
grid = RectilinearGrid(arch; kw...)
grid = ImmersedBoundaryGrid(grid, GridFittedBottom(x -> -rand()*0.5 - 0.5))
reduced_precision_grid = RectilinearGrid(arch, Float32; kw...)
reltol = 1e-7
abstol = 1e-7
maxiter = 1000
lines(grid.underlying_grid.xᶜᵃᵃ[1:Nx], interior(grid.immersed_boundary.bottom_height, :, 1, 1))

function setup_model_time_step(grid, pressure_solver, Δt)
    model = NonhydrostaticModel(; grid, pressure_solver)
    set!(model.velocities.u, (x, z) -> 1)
    
    for i in 1:3
        time_step!(model, Δt)
        msg = @sprintf("%d, cg iter %d, residual extrema %s, pressure extrema %s", 
                        i,
                        model.pressure_solver.conjugate_gradient_solver.iteration, 
                        extrema(interior(model.pressure_solver.conjugate_gradient_solver.residual)), 
                        extrema(interior(model.pressures.pNHS)))
        @info msg
    end
end
```

First start with no preconditioner. 
If I first timestep with a small but not incredibly small value,
```julia
preconditioner = nothing
pressure_solver = ConjugateGradientPoissonSolver(grid; maxiter, reltol, abstol, preconditioner)

Δt = 1e-8
setup_model_time_step(grid, pressure_solver, Δt)
```
you get
```julia
[ Info: 1, cg iter 44, residual extrema (-2.105994706376415e-5, 1.5442273441399452e-5), pressure extrema (-1.068515840303027, 0.4842691543213293)
[ Info: 2, cg iter 22, residual extrema (-1.0839540972252063e-8, 2.1524501754380456e-8), pressure extrema (-0.7363474597932788, 0.26325038759798114)
[ Info: 3, cg iter 18, residual extrema (-2.350921009681301e-8, 3.075084225992711e-8), pressure extrema (-0.7363474556061929, 0.2632503849309418)
```
So it is working fine. When we decrease timestep further and run:
```julia
Δt = 1e-9
setup_model_time_step(grid, pressure_solver, Δt)
```
it gives
```julia
[ Info: 1, cg iter 42, residual extrema (-0.0001916767690120231, 0.00025824477424041703), pressure extrema (-4.058038143137272, 4.771265835182983)
[ Info: 2, cg iter 42, residual extrema (-10954.661226669075, 12237.981721560594), pressure extrema (0.0, 7.48104921027421e23)
[ Info: 3, cg iter 41, residual extrema (-19817.370733437958, 19244.317591307008), pressure extrema (0.0, 7.481049210274212e23)
```
The model blows up after one timestep, even though the cg algorithm has "converged",

Now when we run with a preconditioner `preconditioner = FFTBasedPoissonSolver(grid.underlying_grid)`,
```julia
preconditioner = FFTBasedPoissonSolver(grid.underlying_grid)
pressure_solver = ConjugateGradientPoissonSolver(grid; maxiter, reltol, abstol, preconditioner)

Δt = 1e-9
setup_model_time_step(grid, pressure_solver, Δt)
```
you get
```julia
[ Info: 1, cg iter 13, residual extrema (-0.00011875682298700312, 0.00014129695299583335), pressure extrema (-2.5228737996504087, 1.3589616116497294)
[ Info: 2, cg iter 1000, residual extrema (-4.27008898390138e-8, 2.488065079073281e-9), pressure extrema (-0.7363474536738828, 0.26325041368886765)
[ Info: 3, cg iter 8, residual extrema (-1.2114022081638616e-8, 4.499327871537952e-8), pressure extrema (-0.7363474411571914, 0.26325040389162996)
```
The CG iteration has converged for iterations 1 and 3, and it doesn't appear to be numerically unstable.

If we now run with an extremely small timestep,
```julia
Δt = 1e-15
setup_model_time_step(grid, pressure_solver, Δt)
```
one gets
```julia
[ Info: 1, cg iter 11, residual extrema (-152.61578261604075, 257.050038622766), pressure extrema (-2.6167556434280328e6, 1.37601961731707e6)
[ Info: 2, cg iter 1000, residual extrema (-0.09157646173359503, 0.0), pressure extrema (-0.7122087004951764, 0.2624839677173069)
[ Info: 3, cg iter 1000, residual extrema (-0.05933479749746563, 0.0102464140126443), pressure extrema (-0.6910966153850525, 0.2524036207636871)
```
where pressure values seem to be extreme for iteration 1 and then becomes "normal", but the cg algorithm has not converged.

A similar thing happens if we used a reduced precision grid for the preconditioner
```julia
preconditioner = FFTBasedPoissonSolver(reduced_precision_grid)
pressure_solver = ConjugateGradientPoissonSolver(grid; maxiter, reltol, abstol, preconditioner)

Δt = 1e-15
setup_model_time_step(grid, pressure_solver, Δt)
```
you get
```julia
[ Info: 1, cg iter 13, residual extrema (-172.46743253724992, 194.21305609177966), pressure extrema (-2.5866277574562617e6, 1.411566084497432e6)
[ Info: 2, cg iter 1000, residual extrema (-0.031877997446973616, 0.0), pressure extrema (-0.731370535185227, 0.2549918725517209)
[ Info: 3, cg iter 1000, residual extrema (-0.019568456612751698, 0.0), pressure extrema (-0.7243117337636092, 0.2536665417367077)
```

Now if we do not use the CG solver and only use FFT:
```julia
function setup_model_time_step_fftsolve(grid, Δt)
    pressure_solver = FFTBasedPoissonSolver(grid.underlying_grid)
    model = NonhydrostaticModel(; grid, pressure_solver)
    set!(model.velocities.u, (x, z) -> 1)
    
    for i in 1:3
        time_step!(model, Δt)
        msg = @sprintf("%d, pressure extrema %s", 
                        i,
                        extrema(interior(model.pressures.pNHS)))
        @info msg
    end
end
```
For normal timesteps,
```julia
Δt = 1e-2
setup_model_time_step_fftsolve(grid, Δt)
```
you get
```julia
[ Info: 1, pressure extrema (-4.518124751680564, 3.1385350743875713)
[ Info: 2, pressure extrema (-1.0771171214148325, 2.1895528232118378)
[ Info: 3, pressure extrema (-0.8057842355807996, 1.5203316471764308)
```
But for a small but not extreme timestep,
```julia
Δt = 1e-9
setup_model_time_step_fftsolve(grid, Δt)
```
you get
```julia
[ Info: 1, pressure extrema (-4.6109498634512275e7, 3.2008724053240076e7)
[ Info: 2, pressure extrema (-1.1934877105394058e7, 2.2234521971178945e7)
[ Info: 3, pressure extrema (-7.623888263773426e6, 1.5749507966054006e7)
```
where pressures are quite high but simulation seems to be numerically stable.

Sorry for the long and not so minimal MWE, I am slightly confused if this is a bug/feature, and whether the solution is to add in some hack for a timestep limiter, or maybe there's a more fundamental issue at play.

@glwagner @simone-silvestri @amontoison any advice/thoughts will be greatly appreciated! Thank you so much

xkykai:
Closes #4067 

This change solves $p \text{d}t$ in the Poisson equation rather than $p$ as it is currently.
The advantage of this is that this avoids numerical instabilities that occur when we compute the RHS of 
```math
\nabla ^2 p = -\frac{\nabla \cdot \boldsymbol{u}_*}{\Delta t}
```

when $\Delta t$ is extremely small, which occurs when we take small timesteps, which are almost inevitable when using some sort of `TimeInterval`.
In this change, we solve for 
```math
\psi  \equiv \int_n^{n+1} p \text{d} t \approx p^{n+1} \Delta t
```
and update the pressure correction with
```math
\boldsymbol{u}_{n+1} = \boldsymbol{u}_* - \nabla \psi.
```
The full derivation can be found here https://github.com/CliMA/Oceananigans.jl/issues/4067#issuecomment-2619120179

This is a huge change as `model.pressures.pNHS` now gives $p \text{d}t$. However if we decide to go through with this change, we can discuss the detail of what to expose to the user/change some of the variable name notation to avoid confusion.
Currently this issue has a workaround using the `minimum_relative_step` flag in `Simulation`, however I think it could be useful to implement this PR as it is a fundamental fix.

Here I provide a staircase convection validation script to illustrate consistent behaviour before and after this PR fix:
```julia
using Oceananigans
using Printf
using Oceananigans.Solvers: ConjugateGradientPoissonSolver, FFTBasedPoissonSolver

#####
##### Model setup
#####
Nx = 64
Nz = 64

grid = RectilinearGrid(CPU(), Float64,
                       size = (Nx, Nz),
                       halo = (4, 4),
                       x = (0, 1),
                       z = (0, 1),
                       topology = (Bounded, Flat, Bounded))

slope(x) = (5 + tanh(40*(x - 1/6)) + tanh(40*(x - 2/6)) + tanh(40*(x - 3/6)) + tanh(40*(x - 4/6)) + tanh(40*(x - 5/6))) / 10

grid = ImmersedBoundaryGrid(grid, GridFittedBottom(slope))
                       
@info "Created $grid"
reltol = 1e-7
abstol = 1e-7
maxiter = 1000
# preconditioner = nothing
preconditioner = FFTBasedPoissonSolver(grid.underlying_grid)
pressure_solver = ConjugateGradientPoissonSolver(grid; maxiter, reltol, abstol, preconditioner)
                            pressure_solver,
                            advection = WENO(order=5),
                            coriolis = FPlane(f=0.1),
                            tracers = (:b, :c),
                            buoyancy = BuoyancyTracer())

@info "Created $model"
@info "with pressure solver $(model.pressure_solver)"

# Cold blob
h = 0.03
x₀ = 0.75
z₀ = 0.9
b_background(x, z) = z
b_blob(x, z) = -5 * exp(-((x - x₀)^2 + (z - z₀)^2) / 2h^2)
bᵢ(x, z) = b_background(x, z) + b_blob(x, z)
set!(model, b=bᵢ, c=1)

#####
##### Simulation
#####
Δt = 1e-2
simulation = Simulation(model, Δt=Δt, stop_time = 10)

# wizard = TimeStepWizard(max_change=1.05, max_Δt=1e-1, cfl=0.6)
# simulation.callbacks[:wizard] = Callback(wizard, IterationInterval(1))

wall_time = Ref(time_ns())

b, c = model.tracers
u, v, w = model.velocities
B = Field(Integral(b))
C = Field(Integral(c))
compute!(B)
compute!(C)
B₀ = B[1, 1, 1]
C₀ = C[1, 1, 1]
δ = ∂x(u) + ∂y(v) + ∂z(w)

function progress(sim)
    elapsed = time_ns() - wall_time[]

    compute!(B)
    compute!(C)
    Bₙ = B[1, 1, 1]
    Cₙ = C[1, 1, 1]

    msg = @sprintf("Iter: %d, next Δt: %s, time: %s, wall time: %s, ΔB: %.3f %%, ΔC: %.3f %%, max|δ|: %.2e, residual extrema %s, pressure extrema %s",
                   iteration(sim), prettytime(sim.Δt), prettytime(sim), prettytime(1e-9 * elapsed),
                   100 * (Bₙ - B₀) / B₀,
                   100 * (Cₙ - C₀) / C₀,
                   maximum(abs, δ),
                   extrema(interior(model.pressure_solver.conjugate_gradient_solver.residual)), 
                   extrema(interior(model.pressures.pNHS)))

    pressure_solver = sim.model.pressure_solver
    if sim.model.pressure_solver isa ConjugateGradientPoissonSolver
        solver_iterations = pressure_solver.conjugate_gradient_solver.iteration 
        msg *= string(", solver iterations: ", solver_iterations)
    end

    @info msg

    wall_time[] = time_ns()

    return nothing
end
                   
simulation.callbacks[:p] = Callback(progress, IterationInterval(50))

prefix = "staircase_convection"

outputs = merge(model.velocities, model.tracers, (; p=model.pressures.pNHS, δ))

simulation.output_writers[:jld2] = JLD2Writer(model, outputs;
                                                    filename = prefix * "_fields",
                                                    # schedule = TimeInterval(0.1),
                                                    overwrite_existing = true)

simulation.output_writers[:timeseries] = JLD2Writer(model, (; B, C);
                                                          filename = prefix * "_time_series",
                                                        #   schedule = TimeInterval(0.1),
                                                          overwrite_existing = true)

run!(simulation)
```
which gives an output of

https://github.com/user-attachments/assets/b8a64348-3569-47f1-84d3-64bbc0f613eb

Here I made a video of the field anomalies before and after the fix:

https://github.com/user-attachments/assets/7046f4a5-0f25-4035-8eb5-b929ca324c2f

This shows that the behavior of the flow did not change after the fix. Not sure what is going on with pressure, perhaps I made a multiplicative error but I essentially just did `model.pressures.pNHS ./ Δt` to the new output...

For reference here is the simulation before applying the fix:

https://github.com/user-attachments/assets/a800d2bb-6426-438e-a5b7-e81fb44f39f0

@glwagner @simone-silvestri interested in your thoughts!

### DIFF
--- a/src/Models/NonhydrostaticModels/pressure_correction.jl
+++ b/src/Models/NonhydrostaticModels/pressure_correction.jl
@@ -24,16 +24,21 @@ end
 #####
 
 """
-Update the predictor velocities u, v, and w with the non-hydrostatic pressure via
+Update the predictor velocities u, v, and w with the timestep-multiplied non-hydrostatic pressure via
 
-    `u^{n+1} = u^n - δₓp_{NH} / Δx * Δt`
+    `u^{n+1} = u^n - δₓp_{NH} / Δx`
+
+    Note that `p_{NH}` is the nonhydrostatic pressure muiltiplied by the timestep which enforces the incompressibility condition.
 """
 @kernel function _pressure_correct_velocities!(U, grid, Δt, pNHS)
     i, j, k = @index(Global, NTuple)
 
-    @inbounds U.u[i, j, k] -= ∂xᶠᶜᶜ(i, j, k, grid, pNHS) * Δt
-    @inbounds U.v[i, j, k] -= ∂yᶜᶠᶜ(i, j, k, grid, pNHS) * Δt
-    @inbounds U.w[i, j, k] -= ∂zᶜᶜᶠ(i, j, k, grid, pNHS) * Δt
+    # @inbounds U.u[i, j, k] -= ∂xᶠᶜᶜ(i, j, k, grid, pNHS) * Δt
+    # @inbounds U.v[i, j, k] -= ∂yᶜᶠᶜ(i, j, k, grid, pNHS) * Δt
+    # @inbounds U.w[i, j, k] -= ∂zᶜᶜᶠ(i, j, k, grid, pNHS) * Δt
+    @inbounds U.u[i, j, k] -= ∂xᶠᶜᶜ(i, j, k, grid, pNHS)
+    @inbounds U.v[i, j, k] -= ∂yᶜᶠᶜ(i, j, k, grid, pNHS)
+    @inbounds U.w[i, j, k] -= ∂zᶜᶜᶠ(i, j, k, grid, pNHS)
 end
 
 "Update the solution variables (velocities and tracers)."
--- a/src/Models/NonhydrostaticModels/solve_for_pressure.jl
+++ b/src/Models/NonhydrostaticModels/solve_for_pressure.jl
@@ -13,28 +13,32 @@ using Oceananigans.Solvers: solve!
     i, j, k = @index(Global, NTuple)
     active = !inactive_cell(i, j, k, grid)
     δ = divᶜᶜᶜ(i, j, k, grid, Ũ.u, Ũ.v, Ũ.w)
-    @inbounds rhs[i, j, k] = active * δ / Δt
+    # @inbounds rhs[i, j, k] = active * δ / Δt
+    @inbounds rhs[i, j, k] = active * δ
 end
 
 @kernel function _fourier_tridiagonal_source_term!(rhs, ::XDirection, grid, Δt, Ũ)
     i, j, k = @index(Global, NTuple)
     active = !inactive_cell(i, j, k, grid)
     δ = divᶜᶜᶜ(i, j, k, grid, Ũ.u, Ũ.v, Ũ.w)
-    @inbounds rhs[i, j, k] = active * Δxᶜᶜᶜ(i, j, k, grid) * δ / Δt
+    # @inbounds rhs[i, j, k] = active * Δxᶜᶜᶜ(i, j, k, grid) * δ / Δt
+    @inbounds rhs[i, j, k] = active * Δxᶜᶜᶜ(i, j, k, grid) * δ
 end
 
 @kernel function _fourier_tridiagonal_source_term!(rhs, ::YDirection, grid, Δt, Ũ)
     i, j, k = @index(Global, NTuple)
     active = !inactive_cell(i, j, k, grid)
     δ = divᶜᶜᶜ(i, j, k, grid, Ũ.u, Ũ.v, Ũ.w)
-    @inbounds rhs[i, j, k] = active * Δyᶜᶜᶜ(i, j, k, grid) * δ / Δt
+    # @inbounds rhs[i, j, k] = active * Δyᶜᶜᶜ(i, j, k, grid) * δ / Δt
+    @inbounds rhs[i, j, k] = active * Δyᶜᶜᶜ(i, j, k, grid) * δ
 end
 
 @kernel function _fourier_tridiagonal_source_term!(rhs, ::ZDirection, grid, Δt, Ũ)
     i, j, k = @index(Global, NTuple)
     active = !inactive_cell(i, j, k, grid)
     δ = divᶜᶜᶜ(i, j, k, grid, Ũ.u, Ũ.v, Ũ.w)
-    @inbounds rhs[i, j, k] = active * Δzᶜᶜᶜ(i, j, k, grid) * δ / Δt
+    # @inbounds rhs[i, j, k] = active * Δzᶜᶜᶜ(i, j, k, grid) * δ / Δt
+    @inbounds rhs[i, j, k] = active * Δzᶜᶜᶜ(i, j, k, grid) * δ
 end
 
 function compute_source_term!(pressure, solver::DistributedFFTBasedPoissonSolver, Δt, Ũ)


glwagner:
> This is a huge change as model.pressures.pNHS now gives p * dt. However if we decide to go through with this change, we can discuss the detail of what to expose to the user/change some of the variable name notation to avoid confusion.

One easy thing we can do is to allocate memory for `pNHS_dt`, but then provide an `AbstractOperation` `pNHS / dt`. This does requiring updating the `dt` as the time-step changes though, but still seems possible. Maybe one could argue that's not even a breaking change since it's still `AbstractField` :-D

glwagner:
An even simpler change is to do `pNHS ./= dt` after the pressure correction.

glwagner:
I think this makes a lot of sense. I'm confused why we didn't do this earlier!

glwagner:
interested if this solves @tomchor's issues with tiny time-steps

I also wish we could find a way to prevent rounding errors in `TimeInterval` but we haven't yet...

tomchor:
> interested if this solves @tomchor's issues with tiny time-steps

Seems to solve it! At least based on the following MWE:

```julia
using Oceananigans

grid_base = RectilinearGrid(topology = (Bounded, Periodic, Bounded), size = (16, 20, 4), extent = (800, 1000, 100),)

model = NonhydrostaticModel(grid = grid, timestepper = :RungeKutta3, buoyancy = BuoyancyTracer(), tracers = :b, hydrostatic_pressure_anomaly=nothing)

N² = 6e-6
b∞(x, y, z) = N² * z
set!(model, b=b∞)

using Statistics: std
using Printf
add_callback!(simulation, progress_message, TimeInterval(100))

run!(simulation)
```

Great, simple solution.

Should this be merged, I think we should test it in more complex cases before considering removing the `minimal_relative_timestep` though.

tomchor:
@xkykai one question though. Am I interpreting this wrong, or does the correction increase the divergence at the boundary?

xkykai:
> @xkykai one question though. Am I interpreting this wrong, or does the correction increase the divergence at the boundary?

it does seem to be the case that in the new fix's case the small divergence at the boundary seems to be more consistent throughout the simulation time, but in the initial case the divergence is in the order of 1e-7 in a short burst but then later on becomes smaller. 

I think this is because for a given `reltol` and `abstol` for the CG solver, in the old scenario it was the residual of $p$ that was the stopping criteria. After the correction we are computing the residual of $p \Delta t$, so now the divergence will be larger by a factor of $\frac{1}{\Delta t}$ given the same tolerance.

glwagner:
> so now the divergence will be larger by a factor of 1 / Δt given the same tolerance.

this is called a "testable hypothesis"

xkykai:
> this is called a "testable hypothesis"

You are right! Not sure what's the best way to test it but I tried comparing the number of CG iterations to convergence before and after the correction using the staircase convection above.

With FFT as preconditioner:
Before correction - mostly 10, sometimes 11
After correction - 11

With no preconditioner:
Before correction - between 224 to 234
After correction - between 236 to 242

So it seems that after the correction cg iterations actually increased slightly. So I think my hypothesis is wrong. I am also not sure why the behavior is as such...

glwagner:
> > this is called a "testable hypothesis"
> 
> You are right! Not sure what's the best way to test it but I tried comparing the number of CG iterations to convergence before and after the correction using the staircase convection above.
> 
> With FFT as preconditioner: Before correction - mostly 10, sometimes 11 After correction - 11
> 
> With no preconditioner: Before correction - between 224 to 234 After correction - between 236 to 242
> 
> So it seems that after the correction cg iterations actually increased slightly. So I think my hypothesis is wrong. I am also not sure why the behavior is as such...

What about changing the tolerances?

xkykai:
> > > this is called a "testable hypothesis"
> > 
> > 
> > You are right! Not sure what's the best way to test it but I tried comparing the number of CG iterations to convergence before and after the correction using the staircase convection above.
> > With FFT as preconditioner: Before correction - mostly 10, sometimes 11 After correction - 11
> > With no preconditioner: Before correction - between 224 to 234 After correction - between 236 to 242
> > So it seems that after the correction cg iterations actually increased slightly. So I think my hypothesis is wrong. I am also not sure why the behavior is as such...
> 
> What about changing the tolerances?

if I reduce the tolerance from `1e-7` to `2e-8`, the cg iteration (with FFT preconditioner) increased by 1, and the maximum divergence is about `2.6e-8`. Further reduction of tolerance from `2e-8` to `1e-8` increases cg iteration by 1 more, and maximum divergence is now around `2e-8`.

Below I attach the simulation video for tolerance = `1e-8`. Note the colorrange for divergence plot is `5e-8`. At this tolerance level with the new fix, the divergence is of similar order to the case before the correction where I set tolerance = `1e-7`.

https://github.com/user-attachments/assets/5a8217dc-09d0-45b7-a5be-dfa3c23713c3

Also, my previous hypothesis is wrong. For the same tolerance, decreasing `Δt` actually decreases maximum divergence. For tolerance = `1e-7`, using `Δt = 1e-2` gives maximum divergence of around `8.7e-8`, while changing to `Δt = 1e-3` gives maximum divergence of around `4e-8`.

glwagner:
I wonder if a good test would be to fix the iteration count and look at the difference in the divergence between the two cases.

I'm happy with the benefit of this simpler algorithm though, even if there are a few minor tradeoffs

xkykai:
> An even simpler change is to do `pNHS ./= dt` after the pressure correction.

I tried doing this after
```julia
calculate_pressure_correction!(model, Δt)
pressure_correct_velocities!(model, Δt)
```
and this works for a pure FFT solver but not for a CG solver. The reason I think is because in `initialize_solution!` here at https://github.com/CliMA/Oceananigans.jl/blob/f922c73eac51956979c134068abd1ddca57c42bb/src/Solvers/conjugate_gradient_solver.jl#L168 we essentially used the previous pressure, which has now been divided by `Δt`. We could multiply `Δt` back at this step here but it seems like we are going back to the old issue if we do that.

glwagner:
> > An even simpler change is to do `pNHS ./= dt` after the pressure correction.
> 
> I tried doing this after
> 
> ```julia
> calculate_pressure_correction!(model, Δt)
> pressure_correct_velocities!(model, Δt)
> ```
> 
> and this works for a pure FFT solver but not for a CG solver. The reason I think is because in `initialize_solution!` here at
> 
> https://github.com/CliMA/Oceananigans.jl/blob/f922c73eac51956979c134068abd1ddca57c42bb/src/Solvers/conjugate_gradient_solver.jl#L168
> 
> we essentially used the previous pressure, which has now been divided by `Δt`. We could multiply `Δt` back at this step here but it seems like we are going back to the old issue if we do that.

You have to make two changes I think:

1. Inside `pressure_correct_velocities!`, add the division after performing the pressure correction
2. Inside `solve_for_pressure!`, multiply back by last_dt before calling `solve_for_pressure!`

I would put these _inside_ the functions rather than exposing them in the time-step, since we have multiple time-step methods

xkykai:
> 2\. Inside `solve_for_pressure!`, multiply back by last_dt before calling `solve_for_pressure!`

This works fine when `Δt` is not too small, but when `Δt = 0` (at the last simulation timestep for example we get division by zero problem.
I propose we do
```julia
function solve_for_pressure!(pressure, solver, Δt, Ũ)
    pressure .*= ifelse(Δt == Inf, 0, max(Δt, eps(eltype(solver.grid))))

    compute_source_term!(pressure, solver, Δt, Ũ)
    solve!(pressure, solver)
    return pressure
end

function pressure_correct_velocities!(model::NonhydrostaticModel, Δt)

    launch!(model.architecture, model.grid, :xyz,
            _pressure_correct_velocities!,
            model.velocities,
            model.grid,
            Δt,
            model.pressures.pNHS)
    
    model.pressures.pNHS ./= max(Δt, eps(eltype(model.grid)))

    return nothing
end
```

In my staircase convection test this performs predictably when taking extremely small timesteps when using `TimeInterval`.

glwagner:
forgive me but I can't read this line

```julia
pressure .*= ifelse(Δt == Inf, 0, max(Δt, eps(eltype(solver.grid))))
```

there is an ifelse, and max, and eps going on?

glwagner:
Here's one idea but feel free to adjust the notation

```julia
ϵ = eps(typeof(Δt))
Δt⁺ = max(ϵ, Δt)
Δt★ = Δt⁺ * isfinite(Δt)
pressure .*= Δt★
```

xkykai (commented):
If there are no further issues, I suppose this is ready to be reviewed and merged?

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -30,25 +30,26 @@ Update the predictor velocities u, v, and w with the timestep-multiplied non-hyd
 
     Note that `p_{NH}` is the nonhydrostatic pressure muiltiplied by the timestep which enforces the incompressibility condition.
 """
-@kernel function _pressure_correct_velocities!(U, grid, Δt, pNHS)
+@kernel function _pressure_correct_velocities!(U, grid, pNHSΔt)
```
glwagner, line 33: Note that [Wagner et al](https://glwagner.github.io/assets/pdf/Oceananigans.pdf) introduces a different notation for `p` (the notation `pNHS` comes from MITgcm). I'll leave it up to you but you should feel free to use something a little cleaner.

glwagner: See equation A6 there

xkykai: are you suggesting we use `p̃` instead? I wonder if this is less explicit than using `pNHS`, as tilde could also be associated with 'not anomaly' terms

glwagner: i think we should change the paper notation then, haha

glwagner: Doesn't that argument also apply to the paper... ?

glwagner: If you think we should change the paper to p_NHS so be it. I think multi-letter subscripts are bad. you could to pn if you want to use a letter.

xkykai: I am not against using `p̃`. Do you mean to change it only internally, or do you suggest revamping the API of `model.pressures.pNHS`?

glwagner: if we want to do this then yes I would suggest changing everywhere (but we can leave it for another PR)

xkykai: Let's do that in the next PR after this is merged then!

tomchor: This is just my personal opinion, but I think `p̃` and `p′` aren't very intuitive for nonhydrostatic and hydrostatic pressures. (In fact at first glance I'd have guessed that they mean the opposite of what is proposed.)

I think having variable names that need explanation is okay on a paper, where you expect the reader to read everything carefully and when the definition of the variable is close by (like in the appendix that was mentioned), but I think code benefits from more clear variable names, even if they need to be more verbose.

If we want to change the names to something shorter, I suggest `pₙ` and `pₕ`, which anyone familiar with the pressure separation can promptly identify as the nonhydrostatic and hydrostatic components.

tomchor: PS: This is beyond the point here, but I'm just realizing that there's a typo in the line right after Eq A6, where the pressure variables are referred to as $p_n$ and $p_h$, which happens to be exactly what I proposed above lol

glwagner: Our philosophy is to use the same names in the code as in our papers, it's important.

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -43,8 +43,11 @@ function pressure_correct_velocities!(model::NonhydrostaticModel, Δt)
             _pressure_correct_velocities!,
             model.velocities,
             model.grid,
-            Δt,
             model.pressures.pNHS)
+    
+    ϵ = eps(typeof(Δt))
+    Δt⁺ = max(ϵ, Δt)
+    model.pressures.pNHS ./= Δt⁺
```
tomchor, line 50: As the code is currently written here, this doesn't prevent extremely high values fore the pressure associated with `TimeInterval`, since `pNHS` can is still be very large if Δt is very small, correct? 

To be clear, gauging from my test in https://github.com/CliMA/Oceananigans.jl/pull/4276#issuecomment-2752761689, this change does seem to help with that issue, but that is  a pretty simple example and I think the division by `Δt` here indicates that the issue can still happen. I don't think this is a problem, especially if this PR helps with the convergence of the `ConjugateGradientPoisson` solver. I just wanted to clarify since, if that's the case, then we still need to keep `maximum_relative_timestep`.

tomchor:
Sorry if that's in the comments and I missed it, but I'm still not clear: do we understand why this correction increases the divergence at the boundary? is it really just a matter of rescaling the tolerance by `Δt`?

src/Models/NonhydrostaticModels/pressure_correction.jl
glwagner: The reason that p_n is tough is because we use "n" to refer to the time step (in paper as well as code).

We have referred to `p′` as the hydrostatic "anomaly" (ie, the deviation of the hydrostatic pressure from the reference pressure associated with p0), so the prime means "anomaly" which is standard.

That said I'm ok to swap the notation or use something different. Whatever we choose, it will be in the paper and code consistently.

tomchor: I wasn't aware of that philosophy. May I suggest reverting the paper notation back to `pₙ` and `pₕ` when the reviews come? I think using the tilde and primed versions are really not intuitive.

tomchor: Sorry, I posted the last comment before your last comment appeared for me. You make a good point about `p_n`. I know you dislike multi-letter subscripts, but in this case I think we're well justified in using `pₙₕ` and `pₕ` then.

src/Models/NonhydrostaticModels/pressure_correction.jl
glwagner: I guess in the case of a time-step that is below the limit resolvable by floating point computation, the pressure is not defined.

xkykai: > As the code is currently written here, this doesn't prevent extremely high values fore the pressure associated with `TimeInterval`, since `pNHS` can is still be very large if Δt is very small, correct?

Yes, doing this means that the divide by Δt is not exposed to the `ConjugateGradientSolver`, therefore there is no convergence issues for the solver when timesteps are extremely small. However, in order to maintain that we store pressure in memory rather than pressure * Δt we do this adjustment after the `ConjugateGradientSolver` computes the appropriate correction.

> To be clear, gauging from my test in [#4276 (comment)](https://github.com/CliMA/Oceananigans.jl/pull/4276#issuecomment-2752761689), this change does seem to help with that issue, but that is a pretty simple example and I think the division by `Δt` here indicates that the issue can still happen. I don't think this is a problem, especially if this PR helps with the convergence of the `ConjugateGradientPoisson` solver. I just wanted to clarify since, if that's the case, then we still need to keep `maximum_relative_timestep`.

I've tried in my simple cases of `TimeInterval` in the staircase convection case and it produces pressure values which are not spuriously large when Δt is small. If we do not use ` Δt⁺ = max(ϵ, Δt)` but rather ` Δt⁺ = max(0, Δt)`, we can still see the same issues as in [#4276 (comment)](https://github.com/CliMA/Oceananigans.jl/pull/4276#issuecomment-2752761689). However I agree that there might be some complicated edge cases that could lead to large pressures when trying to maintain that as we are still dividing by Δt.

xkykai:
> Sorry if that's in the comments and I missed it, but I'm still not clear: do we understand why this correction increases the divergence at the boundary? is it really just a matter of rescaling the tolerance by `Δt`?

No I have found that it is not the case. In fact with this new correction it seems that for the same number of cg iterations, the divergence is larger than before the correction. But if you decrease the tolerance with this new correction cg takes a few more iterations and you can reach the same level of divergence as before.

tomchor (approved):
@xkykai I'm satisfied so I'm approving this! Although it seems tests still need some work before they pass...

glwagner:
Idea: how about single prime for hydrostatic anomaly, double prime for the nonhydrostatic residual?

tomchor:
> Idea: how about single prime for hydrostatic anomaly, double prime for the nonhydrostatic residual?

I personally still find that not very clear/intuitive.

I think the issue is that there's no standard literature notation for these quantities. Therefore, any names that are intuitive/clear at a glance without needing to read an explanation (which I personally think should be the priority here, but feel free to disagree) needs to include "h" and "nh" somehow. The current nomenclature does that with `pNHS` and `pHY′`, so I'm okay leaving that there. Although I agree that it's not the most elegant names, and you probably don't wanna use that in the Oceananigans paper either...

I thought more about it and I couldn't come up with a better compromise than `pₕ` and `pₙₕ` for the hydrostatic and nonhydrostatic component. They're short and it's immediately clear what they mean. If you really wanna avoid the double subscript, perhaps another compromise can be `pₕ` and `p′`? Although mixing subscripts and superscripts is also not the best.

It might also be productive to hear from others.

glwagner:
I don't understand why it matters whether there is standard notation or not. We are not doing standard things and there is always a need to invent notation. Prime and double prime convey the notion of "anomaly" and "anomaly from the anomaly". What is not clear about this?

tomchor:
> I don't understand why it matters whether there is standard notation or not. We are not doing standard things and there is always a need to invent notation. 

It doesn't really matter that there isn't one. I only mentioned it since it might make choosing a name easier if there was a standard that was concise and recognizable.

xkykai:
Shall we merge this for now and decide upon what's the best variable name in a future PR?

glwagner:
> Shall we merge this for now and decide on what's the best variable name in a future PR?

that's fine with me

tomchor:
We probably need to change something in the docs to reflect this choice though, no?

glwagner:
> We probably need to change something in the docs to reflect the change brought by this PR though, no?

Do we have docs on this?

tomchor:
> > We probably need to change something in the docs to reflect the change brought by this PR though, no?
> 
> Do we have docs on this?

I thought we did in the [pressure separation page](https://clima.github.io/OceananigansDocumentation/stable/numerical_implementation/pressure_decomposition/), but upon second look, we don't! So all is good.

ali-ramadhan:
> > > We probably need to change something in the docs to reflect the change brought by this PR though, no?
> > 
> > 
> > Do we have docs on this?
> 
> I thought we did in the [pressure separation page](https://clima.github.io/OceananigansDocumentation/stable/numerical_implementation/pressure_decomposition/), but upon second look, we don't! So all is good.

Does the first section of this page need updating? https://clima.github.io/OceananigansDocumentation/stable/numerical_implementation/elliptic_solvers/#The-elliptic-problem-for-the-pressure

xkykai:
This way of doing things will mean that during output writing, when we encounter extremely small timesteps (which is highly frequent since this is where the issue is discovered), we will have to contend with the pressure fields being incorrect since it has been divided by an extremely small number. Is that okay for people?

Alternatively we can consider ways to maintain the old pressure field before this extremely small timestep is taken to be written in the output, something like skipping a solve when `Δt★` is extremely small for example.

tomchor:
> Alternatively we can consider ways to maintain the old pressure field before this extremely small timestep is taken to be written in the output, something like skipping a solve when `Δt★` is extremely small for example.

Isn't that exactly what the `minimum_relative_timestep` keyword argument does?

xkykai:
Yes 😅 so we should keep `minimum_relative_timestep`

xkykai:
[buildkite/oceananigans/pr](https://buildkite.com/clima/oceananigans/builds/23505)
buildkite/oceananigans/pr — Build #23505 failed (1 minute, 4 seconds)
This seems to be failing due to bus error? I'm trying to merge this but can't seem to fix this @glwagner @simone-silvestri

simone-silvestri:
I restarted the tests; sometimes it happens. I ll check we do not have space issue

PR #4283, opened 2025-03-26
Reactant in weak deps

simone-silvestri:
it looks like now we are depending fully on Reactant, probably this should be in weak deps


wsmoses (approved):
yeah, that must've been an accident (we do need to depend on reactant core though)

PR #4290, opened 2025-03-27
Convert EOS and fix tests

simone-silvestri:
Adds a `convert` of the equation of state before building `SeawaterBuoyancy`, which is also required to make tests pass.
Needs https://github.com/CliMA/SeawaterPolynomials.jl/pull/41


glwagner (approved):

Project.toml
```diff
@@ -83,7 +83,7 @@ Printf = "1.9"
 Random = "1.9"
 ReactantCore = "0.1"
 Rotations = "1.0"
-SeawaterPolynomials = "0.3.5"
+SeawaterPolynomials = "0.3.8"
```
glwagner, line 86:
```suggestion
SeawaterPolynomials = "0.3.9"
```

PR #4293, opened 2025-03-27
Rework sharding tests

simone-silvestri:
A bunch of correctness issues have been solved. 
The sharding tests pass locally. However, over 100 timesteps the difference is not machine precision but rather larger (~1e-11). This is probably expected, given that the split explicit free surface is running O(10) substeps per time step.
For this reason, I have relaxed the tolerance of the tests to `sqrt(eps(Float64)`

### DIFF
--- a/test/distributed_tests_utils.jl
+++ b/test/distributed_tests_utils.jl
@@ -98,16 +98,19 @@ function run_distributed_latitude_longitude_grid(arch, filename)
         longitude = (0, 360),
         latitude = (-90, 90),
         topology = (Periodic, Bounded, Flat))
+    
+    @test isnothing(flat_distributed_grid.z)
 
-    # @test isnothing(flat_distributed_grid.z)
+    distributed_grid = LatitudeLongitudeGrid(arch;
+                                             size = (40, 40, 10),
+                                             longitude = (0, 360),
+                                             latitude = (-10, 10),
+                                             z = (-1000, 0),
+                                             halo = (5, 5, 5))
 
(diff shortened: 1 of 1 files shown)


simone-silvestri:
The sharding tests pass locally!! 🥳

simone-silvestri:
The tripolar tests still lead to a segfault...

test/distributed_tests_utils.jl
```diff
@@ -104,17 +104,21 @@ function run_distributed_latitude_longitude_grid(arch, filename)
...
+    # Check also that the bottom height is reconstructed correctly!
+    b = reconstruct_global_field(model.grid.immersed_boundary.bottom_height)
+
     if arch.local_rank == 0
         jldsave(filename; u = Array(interior(u, :, :, 10)),
                           v = Array(interior(v, :, :, 10)), 
                           c = Array(interior(c, :, :, 10)),
-                          η = Array(interior(η, :, :, 1))) 
+                          η = Array(interior(η, :, :, 1)),
+                          b = Array(parent(b))[:, :, 1])
```
glwagner: is there a reason not to save `parent` for all`?

simone-silvestri: hmm, I have removed that `b` field.

simone-silvestri: we could save parent.

simone-silvestri:
update: the tripolar tests do not pass, the latitude longitude tests do

simone-silvestri:
close because these was moved over in #5389

PR #4299, opened 2025-03-28
Fix some broken tests

simone-silvestri:


simone-silvestri:
I think the only problem remaining is a misalignment in the show method of the `BoussinesqEquationOfState` in the doctests.
I am not sure why there is that problem, but it requires fiddling with SeawaterPolynomials, so I ll merge this when tests pass, then that is the last test to fix

PR #4302, opened 2025-03-28
Switch default behavior of `Distributed` architecture

simone-silvestri:
in main, the default behavior is defined for an asynchronous distributed architecture and methods are further specialized for a synchronized architecture

in light of #4301, it is a better pattern to have the default behavior being the one of a synchronized architecture and specialize the methods for an asynchronous architecture...


glwagner (approved):

simone-silvestri:
Distributed tests pass so I ll merge

PR #4312, opened 2025-03-29
Improve show for distributed grids, models, improve formatting

glwagner:
Posting this up now, will illustrate differences in what things look like soon.


glwagner:
this went stale so superceding with #5066

PR #4319, opened 2025-04-01
Add a `RightFolded` topology

simone-silvestri:
this topology, for the moment, is particular to the tripolar grid. It works in the same exact way as a `RightConnected` topology except for two differences:

- the topology-aware operators (used for the split explicit free surface) are different because they encode the fold in the operator
- the right boundary conditions associated with a `RightFolded` are `ZipperBoundaryConditions`


simone-silvestri:
This implementation has been developed in #5094

PR #4320, opened 2025-04-01
Remove incorrect operators

simone-silvestri:
I think these operators are incorrect for a `RightConnected` topology


simone-silvestri:
I was wrong, these operators are needed

PR #4325, opened 2025-04-02
Fix: accidentally sharding in `z`?

glwagner:
Confused about this. @avik-pal can you take a look? Is this a bug?


avik-pal:
:z is always 1 right?

avik-pal:
from staring at the IR I dont think this is a problem, but yeah we shouldn't shard along a axis with size 1

glwagner:
> :z is always 1 right?

In Oceananigans, the "z" direction is direction 3 / the third index.

glwagner:
Ah but maybe you mean when we create the partition, we always create a partition with `1` in the third dimension?

avik-pal:
>  Ah but maybe you mean when we create the partition, we always create a partition with 1 in the third dimension?


yes

glwagner:
ok ill close this...

PR #4336, opened 2025-04-04
Fix a memory allocation issues

simone-silvestri:
which are deriving mainly from a dynamic dispatch caused by offset kernel objects not being able to infer the type of `KernelOffsets` which was bloating up the memory on kernel launch. This effect is much more evident when using a `SplitExplicitFreeSurface` where we launch a lot of kernels all at the same time.

I have also simplified the recursion for `extract_field_time_series!` to exclude `AbstractGrid`s and `Field`s to simplify the recursion and it seems to help.

This is the difference in a (fairly complicated) lop for the HydrostaticFreeSurface model on main
Untitled 2
where all the spaces indicate GC activity, and on this PR where all those white spaces (indicating an idle GPU) are absent
Untitled

### DIFF
--- a/src/OutputReaders/extract_field_time_series.jl
+++ b/src/OutputReaders/extract_field_time_series.jl
@@ -1,38 +1,40 @@
 using Oceananigans.AbstractOperations: AbstractOperation
-using Oceananigans.Fields: flattened_unique_values
+using Oceananigans.Fields: flattened_unique_values, Field
+using Oceananigans.Grids: AbstractGrid
 
 #####
 ##### Utility for "extracting" FieldTimeSeries from large nested objects (eg models)
 #####
 
-extract_field_time_series(t1, tn...) = extract_field_time_series(tuple(t1, tn...))
+extract_field_time_series(args...) = ()
 
-# Utility used to extract field time series from a type through recursion
-function extract_field_time_series(t) 
-    prop = propertynames(t)
-    if isempty(prop)
-        return nothing
-    end
+# extract_field_time_series(t1, tn...) = extract_field_time_series(tuple(t1, tn...))
 
-    extracted = Tuple(extract_field_time_series(getproperty(t, p)) for p in prop)
-    flattened = flattened_unique_values(extracted)
+# # Utility used to extract field time series from a type through recursion
+# function extract_field_time_series(t) 
+#     prop = propertynames(t)
+#     if isempty(prop)
+#         return nothing
+#     end
 
-    return flattened
-end
+#     extracted = Tuple(extract_field_time_series(getproperty(t, p)) for p in prop)
+#     flattened = flattened_unique_values(extracted)
 
-# Termination (move all here when we switch the code up)
-extract_field_time_series(f::FieldTimeSeries) = f
+#     return flattened
+# end
 
-# For types that do not contain `FieldTimeSeries`, halt the recursion
-CannotPossiblyContainFTS = (:Number, :AbstractArray)
+# # Termination (move all here when we switch the code up)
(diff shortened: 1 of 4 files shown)


simone-silvestri:
Actually the main culprit for the loss of performance is the 
```
extract_field_time_series(t::Field) = extract_field_time_series(t.boundary_conditions) # Only for BCs?
```
which commented or uncommented leads to the discrepancy in profiles that I am showing in the description of the PR. I think it has to do with `flattened_unique_values`. We need this line for FTS boundary conditions, however, we can enforce this extract only in that (not often encountered) case and simplify the recursion in other cases.

glwagner:
> Actually the main culprit for the loss of performance is the
> 
> ```
> extract_field_time_series(t::Field) = extract_field_time_series(t.boundary_conditions) # Only for BCs?
> ```
> 
> which commented or uncommented leads to the discrepancy in profiles that I am showing in the description of the PR. I think it has to do with `flattened_unique_values`. We need this line for FTS boundary conditions, however, we can enforce this extract only in that (not often encountered) case and simplify the recursion in other cases.

Ah yes, `flattened_unique_values` has caused problems in type inference before.

Note it is a performance optimization to use `unique`. Why don't we stop doing this for now and worry about that optimization later?

src/OutputReaders/extract_field_time_series.jl
```diff
@@ -24,15 +25,32 @@ end
...
+const WFTSBCS = FieldBoundaryConditions{<:FTSBC}
+const EFTSBCS = FieldBoundaryConditions{<:Any, <:FTSBC}
+const SFTSBCS = FieldBoundaryConditions{<:Any, <:Any, <:FTSBC}
+const NFTSBCS = FieldBoundaryConditions{<:Any, <:Any, <:Any, <:FTSBC}
+const BFTSBCS = FieldBoundaryConditions{<:Any, <:Any, <:Any, <:Any, <:FTSBC}
+const TFTSBCS = FieldBoundaryConditions{<:Any, <:Any, <:Any, <:Any, <:Any, <:FTSBC}
+const IFTSBCS = FieldBoundaryConditions{<:Any, <:Any, <:Any, <:Any, <:Any, <:Any, <:FTSBC}
+
+const FieldBCsFTS = Union{WFTSBCS, EFTSBCS, SFTSBCS, NFTSBCS, BFTSBCS, TFTSBCS, IFTSBCS}
+const FieldFTS = Field{LX, LY, LZ, O, G, I, D, T, <:FieldBCsFTS} where {LX, LY, LZ, O, G, I, D, T}
```
glwagner, line 48: I don't think this is necessary if you fix the issue with `flattened_unique_values`, is it? You merely need `extract_field_time_series` for `BoundaryCondition` and that's it.

simone-silvestri: I am suspecting that is the issue but haven't confirmed it yet. For sure, this removes the allocations in toto and I think we want something like this anyways because it is a little more explicit.

glwagner: It makes things hard to change though

glwagner: So it has a downside for maintainence. We should avoid complex hardcoded patterns like this if possible so that the codebase remains as nimble as possible, but of course if we can't eliminate allocations while also supporting FTS boundary conditions then we have to accept its downsides.

glwagner: Or rather, we have to see this as a necessary complexity of FTS boundary conditions and then reason about whether the complexity is worth supporting FTS boundary conditions

Issue #4340, opened 2025-04-04
Tracer conservation is broken in `ZStar` with Julia 1.11.4


NoraLoose:
The `ZStar` vertical coordinate is intended to conserve tracers, but currently it appears not to. Additionally, the test designed to verify tracer conservation isn't functioning as expected, so this issue has gone unnoticed in the tests.

### Problem 1: Tests for tracer conservation are broken

We have a test that checks for tracer conservation:
https://github.com/CliMA/Oceananigans.jl/blob/d92a49bcfd2e239c4b6f6f991e27f665b0fff9f8/test/test_zstar_coordinate.jl#L29-L30

However, if I'm not mistaken, these `@test` statements currently end up evaluating:
```julia
0 ≈ 0
0 ≈ 0
```
and therefore always pass. This happens because the integrals are never computed before comparison. The fix is simply to call `compute!` before the test evaluation:
```julia
compute!(∫bᵢ)
compute!(∫b)
compute!(∫cᵢ)
compute!(∫c)
```

### Problem 2: Tracers are not conserved with `ZStar`

When I run the [z-star tests](https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_zstar_coordinate.jl) locally,  several tests fail due to this line:
https://github.com/CliMA/Oceananigans.jl/blob/eade289dc857396d6d73abde0a4037c1a089d980/test/test_zstar_coordinate.jl#L31
This indicates that the vertical velocity is not zero at the top surface, which suggests that something is indeed not working as intended.

To illustrate this point, I replicated one of the [z-star test cases](https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_zstar_coordinate.jl) with additional diagnostics:

```julia

using Oceananigans
using Oceananigans.Units
using Oceananigans.Grids: MutableVerticalDiscretization
using Oceananigans.Models: ZStar, ZCoordinate
using Random
using Printf

grid = RectilinearGrid(
    CPU(); 
    size = (10, 10, 20), 
    x = (0, 100kilometers), 
    y = (-10kilometers, 10kilometers), 
    topology = (Periodic, Periodic, Bounded), 
    z = MutableVerticalDiscretization((-20, 0))
)

free_surface = SplitExplicitFreeSurface(grid; cfl = 0.75)
# free_surface = ImplicitFreeSurface()
# free_surface = ExplicitFreeSurface()

model = HydrostaticFreeSurfaceModel(; 
    grid, 
[...]

PR #4342, opened 2025-04-05
Zstar test that actually test conservation

simone-silvestri:
bug found by @Noraloose (see #4340)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.96.14"
+version = "0.96.15"
 
 [deps]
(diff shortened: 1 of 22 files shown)


NoraLoose:
@simone-silvestri What do you think about adding the `TripolarGrid` to the z-star test suite? I think it would be valuable, especially since tripolar grids combined with z-star will probably be used frequently — for example, in OMIP-style configurations.

simone-silvestri:
Sounds good! We can probably add the same test we are doing for the immersed lat long but for a tripolar grid

glwagner:
remember to do one PR at a time

NoraLoose:
> remember to do one PR at a time

Sorry, my bad. Got a little over-excited with this PR...

glwagner:
> > remember to do one PR at a time
> 
> Sorry, my bad. Got a little over-excited with this PR...

Its ok. It's my never ending burden to say this thing

simone-silvestri:
I ve changed this PR to correct only the tests and remove the tests that do not pass. We can look at those in another PR

PR #4344, opened 2025-04-05
fix: sharding dropped on LatLon Grid

avik-pal:


ext/OceananigansReactantExt/Architectures.jl
```diff
@@ -86,4 +87,27 @@ end
...
+    child_arch = child_architecture(new_arch)
+    old_properties = (old_grid.Δλᶠᵃᵃ, old_grid.Δλᶜᵃᵃ, old_grid.λᶠᵃᵃ,  old_grid.λᶜᵃᵃ,
+                      old_grid.Δφᵃᶠᵃ, old_grid.Δφᵃᶜᵃ, old_grid.φᵃᶠᵃ,  old_grid.φᵃᶜᵃ,
+                      old_grid.z,
+                      old_grid.Δxᶠᶜᵃ, old_grid.Δxᶜᶠᵃ, old_grid.Δxᶠᶠᵃ, old_grid.Δxᶜᶜᵃ,
+                      old_grid.Δyᶠᶜᵃ, old_grid.Δyᶜᶠᵃ,
+                      old_grid.Azᶠᶜᵃ, old_grid.Azᶜᶠᵃ, old_grid.Azᶠᶠᵃ, old_grid.Azᶜᶜᵃ)
+
+    sharding = Sharding.DimsSharding(new_arch.connectivity, (1, 2), (:x, :y))
+    new_properties = Tuple(Reactant.to_rarray(p; sharding) for p in old_properties)
```
glwagner, line 101:
```suggestion
    new_properties = Tuple(on_architecture(new_arch, p) for p in old_properties)
```

what do you think about this instead?

avik-pal: do we have the on_architecture for sharded reactant defined? But yeah, I like this suggestion

glwagner: good question, we can save this for the future esp since tests pass now (assuming this is meaningful)

I was surprised that this worked for the property `old_grid.z` since it is not an array, but perhaps the recurisve behavior works?

glwagner: We may only need to shard the 2D fields, too...

avik-pal: so replicated for 1D and sharded for 2D?

glwagner: I think actually if we use `on_architecture` we can decide within that fairly easily. I can put this in though. I will cook up `on_architecture` for `ShardedDistributed`.

Maybe in the interim we just merge this so we can work on other sharded things concurrently. What do you think?

avik-pal: yeah lets merge and iterate

PR #4348, opened 2025-04-06
Ensure the `substeps` we pass to `SplitExplicitFreeSurface` are the actual used substeps

simone-silvestri:
in `main` this is not the case because the number of substeps is adapted to the averaging function by assuming that we cut the interval `2Δt` in a `substep` number of substeps.

This PR calculates the correct ending fractional time based on the averaging function. Typically this corresponds to the time at which the averaging function reaches zero typically between 1 and 2 (times the baroclinic timesteps). 
Other kernels might have a more complicated ending time (for example a constant function never reaches zero) for which users have to specify it themselves. For this purpose, this PR adds an `AveragingKernel` that stores also the ending fractional time and can be passed to the `SplitExplicitFreeSurface` constructor. 

Finally the weights calculation cuts `last_fractional_time * Δt` in a `substep` number of substeps.


src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
```diff
@@ -202,16 +204,50 @@ function materialize_free_surface(free_surface::SplitExplicitFreeSurface, veloci
...
+
 @inline   cosine_averaging_kernel(τ::FT) where FT = τ ≥ 0.5 && τ ≤ 1.5 ? convert(FT, 1 + cos(2π * (τ - 1))) : zero(FT)
 @inline constant_averaging_kernel(τ::FT) where FT = convert(FT, 1)
 
+@inline last_fractional_time(k::AveragingKernel) = k.last_τ
+
+# Averaging functions for which we know the last fractional time
+@inline last_fractional_time(::typeof(minimal_dispersion_averaging_shape_function)) = 1.4410682589927724
+@inline last_fractional_time(::typeof(cosine_averaging_kernel))                     = 1.5
+@inline last_fractional_time(::typeof(constant_averaging_kernel))                   = 2.0
```
glwagner, line 240: these are hard-coding Float64?

glwagner:
I tried to read to the code to figure out what's going on but I didn't get it quickly.

Isn't another option to change the meaning of what the user passes to `SplitExplicitFreeSurface`? It's unfortunate that these changes make the source code more complicated. Ideally it would simplify the source code.

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
```diff
@@ -202,16 +204,50 @@ function materialize_free_surface(free_surface::SplitExplicitFreeSurface, veloci
                                     timestepper)
 end
 
-# (p = 2, q = 4, r = 0.18927) minimize dispersion error from Shchepetkin and McWilliams (2005): https://doi.org/10.1016/j.ocemod.2004.08.002
+""" An internal function that stores the averaging kernel 
+and the last fractional time stepped in the subcycling """
+struct AveragingKernel{FT}
+    kernel :: Function
```
glwagner, line 210: This isn't concretely typed because `Function` is an abstract type. You need to use a type parameter.

(Also as a passing comment, it's usually not a good idea to subtype or dispatch on `::Function`, because it is not very specific. Callable objects are not functions, and `Function` is only an abstract type).

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
simone-silvestri: yeah, they are changed at regularization.

glwagner:
Can a fix be to take the user input for `substeps` and multiply it by `2` before just reusing the existing machinery?

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
```diff
@@ -202,16 +204,50 @@ function materialize_free_surface(free_surface::SplitExplicitFreeSurface, veloci
...
+@inline (a::AveragingKernel)(τ) = a.kernel(τ)
+
+regularize_averaging_kernel(FT, a::AveragingKernel) = AveragingKernel(a.kernel, convert(FT, a.last_τ))
+
+function regularize_averaging_kernel(FT, f::Function) 
+    last_τ = last_fractional_time(f)
+    return AveragingKernel(f, convert(FT, last_τ))
+end
+
+# Averaging shape function from from Shchepetkin and McWilliams (2005): https://doi.org/10.1016/j.ocemod.2004.08.002
```
simone-silvestri, line 223:
```suggestion
# Averaging shape function from Shchepetkin and McWilliams (2005): https://doi.org/10.1016/j.ocemod.2004.08.002
```

PR #4350, opened 2025-04-06
Use polar criteria to build PolarBoundaryCondition

glwagner:
This PR changes the criteria for `PolarBoundaryCondition` to check whether the north/south end of the grid is at the pole. Previously, it checked whether the _center of the first halo_ was "beyond the pole". This "non-polar" criteria produces undesirable behavior; for example, whether or not the polar boundary condition would be applied depended on resolution.

It also reorganizes code so that `polar_boundary_condition.jl` only contains code relevant to applying the `PolarBoundaryCondition`, and `FieldBoundaryCondition` stuff resides in `field_boundary_condition.jl`.

### DIFF
--- a/src/BoundaryConditions/field_boundary_conditions.jl
+++ b/src/BoundaryConditions/field_boundary_conditions.jl
@@ -259,12 +259,13 @@ regularize_field_boundary_conditions(boundary_conditions::NamedTuple, grid::Abst
 ##### Special behavior for LatitudeLongitudeGrid
 #####
 
+# TODO: these may be incorrect because we have not defined behavior for prognostic fields (which are
+# treated by `regularize`).
 regularize_north_boundary_condition(bc::DefaultBoundaryCondition, grid::LatitudeLongitudeGrid, loc, args...) = 
     regularize_boundary_condition(latitude_north_auxiliary_bc(grid, loc, bc), grid, loc, args...)
 
 regularize_south_boundary_condition(bc::DefaultBoundaryCondition, grid::LatitudeLongitudeGrid, loc, args...) = 
     regularize_boundary_condition(latitude_south_auxiliary_bc(grid, loc, bc), grid, loc, args...)
-
 
 function FieldBoundaryConditions(grid::LatitudeLongitudeGrid, location, indices=(:, :, :);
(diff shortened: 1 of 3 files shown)


src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -254,3 +254,71 @@ regularize_field_boundary_conditions(::Missing,
...
+#####
+##### Special behavior for LatitudeLongitudeGrid
+#####
+
+regularize_north_boundary_condition(bc::DefaultBoundaryCondition, grid::LatitudeLongitudeGrid, loc, args...) = 
+    regularize_boundary_condition(latitude_north_auxiliary_bc(grid, loc, bc), grid, loc, args...)
+
+regularize_south_boundary_condition(bc::DefaultBoundaryCondition, grid::LatitudeLongitudeGrid, loc, args...) = 
+    regularize_boundary_condition(latitude_south_auxiliary_bc(grid, loc, bc), grid, loc, args...)
+
```
glwagner, line 267:
```suggestion
```

src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -254,3 +256,67 @@ regularize_field_boundary_conditions(::Missing,
...
+    # TODO: define special behavior for prognostic fields.
+    north = latitude_north_auxiliary_bc(grid, location, north)
+    south = latitude_south_auxiliary_bc(grid, location, south)
+
+    return FieldBoundaryConditions(indices, west, east, south, north, bottom, top, immersed)
+end
+
+function latitude_north_auxiliary_bc(grid, loc, default_bc=DefaultBoundaryCondition()) 
+    # Check if the halo lies beyond the north pole
+    φnorth = @allowscalar φnode(grid.Ny+1, grid, Face())
```
navidcy, line 290: is this only to be used on a lat-lon grid?
if not, the more general way is to call φnode with i, j, k and 3 locations?

glwagner: for sure, this boundary condition only applies to lat lon

navidcy: then ignore my comment!

PR #4353, opened 2025-04-07
Add validation for `condition` kwarg in `ConditionalOperation` for immersed boundary grids

navidcy:
This PR adds validation for NotImmersed conditions that are Arrays. Continuation from #4266.

### DIFF
--- a/src/AbstractOperations/conditional_operations.jl
+++ b/src/AbstractOperations/conditional_operations.jl
@@ -19,6 +19,7 @@ struct ConditionalOperation{LX, LY, LZ, F, C, O, G, M, T} <: AbstractOperation{L
         if func === Base.identity
             func = nothing
         end
+        condition = validate_condition(condition, operand)
         T = eltype(operand)
         F = typeof(func)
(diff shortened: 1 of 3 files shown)


src/AbstractOperations/conditional_operations.jl
```diff
@@ -19,6 +19,7 @@ struct ConditionalOperation{LX, LY, LZ, F, C, O, G, M, T} <: AbstractOperation{L
         if func === Base.identity
             func = nothing
         end
+        condition = validate_condition(condition, operand)
```
glwagner, line 22: does this have a cost? sometimes it's better to keep the inner constructors clean. Fore xample can this run on GPU?

glwagner: Basically we want to have some simple way of building an object that does not have errors and prints and whatnot embedded in it

navidcy: Hm... I thought it'd be cleaner that way.
But I see your point.

glwagner (approved):

PR #4359, opened 2025-04-07
Use an accumulator in vertical integrals

simone-silvestri:
it might helo with some accumulation problems we were seeing in reactant.


simone-silvestri:
Probably not required anymore

PR #4360, opened 2025-04-07
Test zstar on a tripolar grid

simone-silvestri:


simone-silvestri:
@NoraLoose tests seem to pass here. I would merge then we can look at the difference with your tests failing

NoraLoose:
@simone-silvestri Nice! Which of the Buildkite groups are the z-star tests part of?

simone-silvestri:
They have their specific group in here
https://github.com/CliMA/Oceananigans.jl/blob/db748ea9e7e9d7473edc1046bd6fe1ecadca7a66/.buildkite/pipeline.yml#L670-L705

NoraLoose:
Ah cool, together with

https://github.com/CliMA/Oceananigans.jl/blob/db748ea9e7e9d7473edc1046bd6fe1ecadca7a66/test/runtests.jl#L218-L222

this now makes sense to me.

Yep, let's merge this, and investigate the issue of #4340 (which seems to be a Julia version issue) separately.

test/test_zstar_coordinate.jl
```diff
@@ -202,5 +202,32 @@ end
...
+                                                  tracers = (:b, :c), 
+                                                  buoyancy = BuoyancyTracer(),
+                                                  vertical_coordinate = ZStar())
+
+            bᵢ(x, y, z) = x < 180 ? 0.06 : 0.01 
+
+            set!(model, c = (x, y, z) -> rand(), b = bᵢ)
+
+            Δt = 2minutes
+            test_zstar_coordinate(model, 100, Δt)
```
NoraLoose, line 230: Question since we're starting from rest: I’m guessing that after 200 minutes the velocities have developed enough to significantly differ from zero — so the tracer actually has a chance to be advected? I realize we don’t want the CI to run for much longer than that.

NoraLoose: I'm asking because in #4340 (with Julia 1.11.4) some of the tests for `ImplicitFreeSurface` and `ExplicitFreeSurface` are passing simply because they only run for a few time steps. If they were run longer, they would fail due to tracer conservation issues with Julia 1.11.4. Just want to make sure we’re adding a physically meaningful test here, so that these kinds of failures would be reliably detected.

glwagner: Do you mind putting in a PR to run those tests for longer so they are more meaningful?

NoraLoose: Sounds good. I’ll open a separate PR once I’ve come up with meaningful run times for all zstar test cases. This PR should be ready to merge, @simone-silvestri.

simone-silvestri:
@NoraLoose you were right there was something that was not 100% right. The immersed boundary was too high, leaving very little fluid cells. I have changed the tests to make sure the simulation is more "realistic". Let's see if the tests pass. If not this might be helpful to figure out if something is going wrong

navidcy:
@NoraLoose seems to me that you are approving the PR. Thanks 🙏🏼
Could you also approve formally on Github (otherwise merging is blocked).

NoraLoose (approved):

NoraLoose:
@navidcy I don't think my approval will unblock the merging, but I approve anyways :)

navidcy:
oh I see!

navidcy:
> @navidcy I don't think my approval will unblock the merging, but I approve anyways :)

I sent you an invite to get write access in this repo

PR #4362, opened 2025-04-07
Fix Zstar-RK3 

simone-silvestri:
The tests are being removed by #4342 because it looks like ZStar-RK3 requires a bit more work.
This PR is here to fix the issue and reinstate the tests


src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
```diff
@@ -53,10 +53,6 @@ function rk3_average_free_surface!(free_surface::SplitExplicitFreeSurface, grid,
     launch!(arch, grid, :xy, _rk3_average_free_surface!, Uⁿ, grid, Uⁿ⁻¹, γⁿ, ζⁿ)
     launch!(arch, grid, :xy, _rk3_average_free_surface!, Vⁿ, grid, Vⁿ⁻¹, γⁿ, ζⁿ)
 
-    # Averaging the free surface is only required for a grid with Mutable vertical coordinates,
-    # which needs to update the grid based on the value of the free surface
-    launch!(arch, grid, :xy, _rk3_average_free_surface!, ηⁿ, grid, ηⁿ⁻¹, γⁿ, ζⁿ)
```
simone-silvestri, line 58: This was untrue; this step is not only not necessary but also a bug that I introduced. The correct free surface (which informs the grid update) is the one that is evolved dynamically in the `step_free_surface!` function. After this average step, the free surface is returned to its value at `n-1` for the last substep (3) that evolves the free surface from `n-1` to `n` with the updated slow tendencies.

glwagner (approved):

Issue #4318, opened 2025-04-01
Ax, Ay and Az are negative in a Tripolar grid


taimoorsohail:
I am trying to compute the velocity multiplied by the grid cell area (Ax, Ay or Az in Oceananigans). However, I seem to have stumbled upon a confusing feature of these Operators in that they are negative in some places in a Tripolar grid. 

As a simple example:

```Julia
using ClimaOcean
using Oceananigans

arch = CPU()

# ### Grid and Bathymetry
@info "Defining grid"

Nx = Integer(360*4)
Ny = Integer(180*4)
Nz = Integer(100)

@info "Defining vertical z faces"

r_faces = exponential_z_faces(; Nz, depth=5000, h=34)
z_faces = Oceananigans.MutableVerticalDiscretization(r_faces)

@info "Defining tripolar grid"

underlying_grid = TripolarGrid(arch;
                               size = (Nx, Ny, Nz),
                               z = z_faces,
                               halo = (5, 5, 4),
                               first_pole_longitude = 70,
                               north_poles_latitude = 55)

@info "Defining bottom bathymetry"

@time bottom_height = regrid_bathymetry(underlying_grid;
                                  minimum_depth = 10,
                                  interpolation_passes = 75, # 75 interpolation passes smooth the bathymetry near Florida so that the Gulf Stream is able to flow
				                  major_basins = 2)

# For this bathymetry at this horizontal resolution we need to manually open the Gibraltar strait.
view(bottom_height, 102:103, 124, 1) .= -400

@info "Defining grid"

@time grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom_height); active_cells_map=true)

c = XFaceField(grid)

using Oceananigans.Operators: Ax, Δz

c .= 1

compute!(Field(c * Ax))
```

gives

```Julia
1440×720×100 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 1440×720×100 ImmersedBoundaryGrid{Float64, Periodic, Oceananigans.Grids.RightConnected, Bounded} on CPU with 5×5×4 halo
├── boundary conditions: FieldBoundaryConditions
│   └── west: Periodic, east: Periodic, south: ZeroFlux, north: Zipper, bottom: ZeroFlux, top: ZeroFlux, immersed: Default
├── operand: BinaryOperation at (Face, Center, Center)
├── status: time=0.0
[...]

PR #4368, opened 2025-04-08
Remove assumption about X/YFaceField on TripolarGrid by generalizing `default_auxiliary_boundary_condition`

glwagner:
This PR does a few things:

* add `grid` and `boundary` (eg `::Val{:north}`) as args to `default_auxiliary_bc(grid, boundary, loc, topo)`
* excise the spurious additional Field constructor for TripolarGrid
* fix the auxiliary / regularize distinction so that auxiliary fields have `sign=1` and only :u, :v get sign=-1
* ~~change `fill_halo_regions!` so that it only touches halos, not the interior~~

Questions:

* ~~What about :U, :V barotropic velocities? Do we need to fix those?~~ edit we checked
* There is still an extra Field constructor for distributed tripolar --- can we get rid of that through further design?
* ~~What is the definition of the grid --- are the top row of cells really redundant? Why can't we make it so just the halos are redundant (this is what we expect)?~~

resolves https://github.com/CliMA/Oceananigans.jl/issues/4318

To follow up on crossed out items: One half of the top row of cells must be part of the grid, because they have to be "owned" by somebody (eg tendencies computed, prognostic fields advanced). The other half of the cells are auxiliary and not part of the grid. This necessitates touching interior cells during `fill_halo_regions!` to ensure that fields are consistent.

The above suggests that we should consider changing the name of `fill_halo_regions!` to `fill_auxiliary_cells!` or something.


glwagner:
```julia
using Oceananigans, GLMakie
grid = TripolarGrid(size=(12, 6, 1), z=(0, 1))
f = XFaceField(grid)
set!(f, 1)
Oceananigans.BoundaryConditions.fill_halo_regions!(f)
heatmap(interior(f, :, :, 1))
```

produces

image

glwagner:
```julia
julia> model.free_surface.barotropic_velocities.U
40×30×1 Field{Face, Center, Nothing} reduced over dims = (3,) on OrthogonalSphericalShellGrid on CPU
├── grid: 40×30×10 OrthogonalSphericalShellGrid{Float64, Periodic, Oceananigans.Grids.RightConnected, Bounded} on CPU with 4×15×4 halo and with precomputed metrics
│   └── west: Periodic, east: Periodic, south: ZeroFlux, north: Zipper(-1), bottom: Nothing, top: Nothing, immersed: Default
└── data: 48×60×1 OffsetArray(::Array{Float64, 3}, -3:44, -14:45, 1:1) with eltype Float64 with indices -3:44×-14:45×1:1

julia> model.free_surface.barotropic_velocities.V
40×30×1 Field{Center, Face, Nothing} reduced over dims = (3,) on OrthogonalSphericalShellGrid on CPU
├── grid: 40×30×10 OrthogonalSphericalShellGrid{Float64, Periodic, Oceananigans.Grids.RightConnected, Bounded} on CPU with 4×15×4 halo and with precomputed metrics
│   └── west: Periodic, east: Periodic, south: Open{Nothing}, north: Zipper(-1), bottom: Nothing, top: Nothing, immersed: Default
└── data: 48×60×1 OffsetArray(::Array{Float64, 3}, -3:44, -14:45, 1:1) with eltype Float64 with indices -3:44×-14:45×1:1
```

(we also changed `bc_str` so the sign shows up in Base.summary)

src/BoundaryConditions/fill_halo_regions_zipper.jl
```diff
@@ -77,9 +77,6 @@ end
         end
     end
 
-    # We substitute the redundant part of the last row to ensure consistency
-    @inbounds c[i, Ny, k] = ifelse(i > Nx ÷ 2, sign * c[i′, Ny, k], c[i, Ny, k])
```
simone-silvestri, line 81: This is required if we want only one source of truth. See https://github.com/CliMA/OrthogonalSphericalShellGrids.jl/pull/58

simone-silvestri: so if we remove it from here we must add it somewhere else.

simone-silvestri:
This route seems like a way forward. We need to change this file for the split explicit free surface https://github.com/CliMA/Oceananigans.jl/blob/main/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_field_tuples.jl to make sure that tendencies use the `default_prognostic_bc` (at this point we can change it for all the free surfaces so that the BC are correct even if we don't really need to fill the halos of tendencies except for split explicit)

src/BoundaryConditions/fill_halo_regions_zipper.jl
glwagner: why does the grid overlap? shouldn't the grid only overlap in halos?

glwagner:
> https://github.com/CliMA/Oceananigans.jl/blob/main/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_field_tuples.jl

Are you sure that's enenough, or do we need to use the exact same bcs that apply to the fields?

glwagner:
The diff is still negative on this PR, hoping to keep it that way :-P

src/BoundaryConditions/fill_halo_regions_zipper.jl
```diff
@@ -77,9 +77,6 @@ end
         end
     end
```
glwagner, line 79:
```suggestion

    # We substitute the redundant part of the last row to ensure consistency
    @inbounds c[i, Ny, k] = ifelse(i > Nx ÷ 2, sign * c[i′, Ny, k], c[i, Ny, k])
```

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_field_tuples.jl
```diff
@@ -1,44 +1,54 @@
...
 using Oceananigans.Fields: XFaceField, YFaceField, ZFaceField, TracerFields
 using Oceananigans.TimeSteppers: QuasiAdamsBashforth2TimeStepper, SplitRungeKutta3TimeStepper
+using Oceananigans.Models.HydrostaticFreeSurfaceModels.SplitExplicitFreeSurfaces: QuasiAdamsBashforth2TimeStepper, SplitRungeKutta3TimeStepper
 
-function hydrostatic_velocity_fields(::Nothing, grid, clock, bcs=NamedTuple())
+function hydrostatic_velocity_fields(::Nothing, grid, clock, bcs)
     u = XFaceField(grid, boundary_conditions=bcs.u)
     v = YFaceField(grid, boundary_conditions=bcs.v)
-    w = ZFaceField(grid)
+    w = ZFaceField(grid, boundary_conditions=bcs.w)
```
glwagner, line 9:
```suggestion
    w = ZFaceField(grid)
```

src/BoundaryConditions/fill_halo_regions_zipper.jl
```diff
@@ -65,8 +64,7 @@ end
 
 @inline function fold_north_face_center!(i, k, grid, sign, c)
```
simone-silvestri, line 65: I guess this will be `u` instead of `c`

glwagner, line 65:
```suggestion
@inline function fold_north_face_center!(i, k, grid, sign, u)
```

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_field_tuples.jl
```diff
@@ -2,43 +2,27 @@ using Oceananigans.Grids: Center, Face
...
-    U = similar(free_surface.barotropic_velocities.U)
-    V = similar(free_surface.barotropic_velocities.V)
-    tracers = TracerFields(tracer_names, grid)
-    return merge((u=u, v=v, U=U, V=V), tracers)
-end
-
 previous_hydrostatic_tendency_fields(::Val{:QuasiAdamsBashforth2}, args...) = hydrostatic_tendency_fields(args...)
 previous_hydrostatic_tendency_fields(::Val{:SplitRungeKutta3}, args...) = nothing
-
-function previous_hydrostatic_tendency_fields(::Val{:SplitRungeKutta3}, velocities, free_surface::SplitExplicitFreeSurface, args...)
```
simone-silvestri, line 39: Is this not needed?

simone-silvestri (approved):

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_field_tuples.jl
glwagner: I moved split explicit stuff to the split explicit module

navidcy:
We need to reinstate the requirement that we update `main` before merging. Otherwise PRs overwrite other things and introduce bugs.

navidcy:
I dropped the requirement. Moving on.

Issue #4369, opened 2025-04-09
Using `1days` should throw an error in `IterationInterval` in `OutputWriters`


taimoorsohail:
Hi, 

It took me far too long to realise that putting a `TimeInterval`-based schedule into `IterationInterval` doesn't throw any error, instead, the code just doesn't output anything beyond `t = 0`. 

That is, I think the below should throw an error:

```Julia
ocean.output_writers[:surface] = JLD2Writer(ocean.model, outputs;
                                                 schedule = IterationInterval(5days),
                                                 filename = "global_surface_fields",
                                                 indices = (:, :, grid.Nz),
                                                 with_halos = false,
                                                 overwrite_existing = true,
                                                 array_type = Array{Float32})
```

I think the issue is that `IterationInterval` is defined as an `AbstractSchedule` when it should really be `Integer` or similar, right? 

https://github.com/CliMA/Oceananigans.jl/blob/f67d864953d0946ddc60f96f584ce8fefd8f87ba/src/Utils/schedules.jl#L95

glwagner:
I think what you mean is that you want an error if `interval` is not an `Int`. This is easy: https://github.com/CliMA/Oceananigans.jl/pull/4370

I'm a little worried about this. We have tried not to do this elsewhere because it can be annoying, eg things like `4/2` are not valid intervals, even though `Int(4/2)` is fine (that conversion happens under the hood right now).

But I feel your pain that `1days` is annoying.

Would another solution be to somehow give information about callbacks / output writers after calling `run!`? Then you might notice that something was wrong.

navidcy:
Like, @glwagner, you mean that eg as soon as you call `run!(simulation)` you see info like

```Julia
julia> run!(simulation)
[ Info: Simulation is running...
[ Info: Output :surface_fields will be saved every 2341235235 iterations.
```

etc?

taimoorsohail:
Yes, something like what @navidcy suggests would also work. At the moment, the only sign I have that something is not working is that the size of the output `*.jld2` files doesn't grow as the simulation progresses - and for that, I will have to wait longer as the resolution increases and the dt decreases.

glwagner:
> Like, [@glwagner](https://github.com/glwagner), you mean that eg as soon as you call `run!(simulation)` you see info like
> 
> julia> run!(simulation)
> [ Info: Simulation is running...
> [ Info: Output :surface_fields will be saved every 2341235235 iterations.
> etc?

[...]

PR #4370, opened 2025-04-09
Require `interval::Int` in IterationInterval?

glwagner:
Resolves https://github.com/CliMA/Oceananigans.jl/issues/4369

I am curious what others think. The reason why 86400.0 is valid is because it can be converted to `Int` (which happens under the hood). So if users do math on their iteration interval they will have to manually convert it to `Int`.

### DIFF
--- a/src/Utils/schedules.jl
+++ b/src/Utils/schedules.jl
@@ -108,7 +108,7 @@ For example,
 * `IterationInterval(100)` actuates at iterations `[100, 200, 300, ...]`.
 * `IterationInterval(100, offset=-1)` actuates at iterations `[99, 199, 299, ...]`.
 """
-IterationInterval(interval; offset=0) = IterationInterval(interval, offset)
+IterationInterval(interval::Int; offset=0) = IterationInterval(interval, offset)
 (schedule::IterationInterval)(model) = (model.clock.iteration - schedule.offset) % schedule.interval == 0
 
 next_actuation_time(schedule::IterationInterval) = Inf


navidcy:
I think the proposed change makes sense. It will probably make avoiding more mistakes than the amount of inconvenience it creates?

Yes, people can't do `IterationInterval(20/2)` anymore, but that at least will error quickly. You won't wait for a day to realise that you actually asked for output every `IterationInterval(5days)` which means every 432000 iterations when you wanted to ask for `TimeInterval(5days)`.

But info regarding the callbacks (as discussed in https://github.com/CliMA/Oceananigans.jl/issues/4369) would also be a nice alternative to this PR.

taimoorsohail:
Yes I think the utility of catching peoples' errors might be worth the extra inconvenience of wrapping `20/2` with `Integer()`.

taimoorsohail (approved):

glwagner:
another solution is to use a different strategy for units

I guess I am uncomfortable with this because it is fundamentally a dirty trick rather than actual input validation. For example if we defined `days=86400` then this would not work...

glwagner:
It works because 

https://github.com/CliMA/Oceananigans.jl/blob/f67d864953d0946ddc60f96f584ce8fefd8f87ba/src/Units.jl#L16

glwagner:
maybe some evidence that this is a likely mistake would help? that might justify it

glwagner:
I guess if this causes problems down the line it is easy to undo.

taimoorsohail:
> maybe some evidence that this is a likely mistake would help? that might justify it

I ran into this issue/mistake, so I guess you have a sample size of one. I am sure we would get a better representative sample once more people pick up and start running the model (and maybe I am the anomaly!)

glwagner:
@taimoorsohail @navidcy how about we merge now to close the issue, then in the longer run we can think about more robust solutions to help users catch mistakes (including non-type-related mistakes)?

glwagner:
> > maybe some evidence that this is a likely mistake would help? that might justify it
> 
> I ran into this issue/mistake, so I guess you have a sample size of one. I am sure we would get a better representative sample once more people pick up and start running the model (and maybe I am the anomaly!)

I think Oceananigans is fairly widely used currently, but it takes courage and a desire to make the world better to actually go to the effort to open an issue!

taimoorsohail:
Ah yes - sorry! I forgot we are on `Oceanangians.jl` and not `ClimaOcean.jl`!!!

PR #4371, opened 2025-04-09
Do not hardcode `required_properties` in checkpointer

simone-silvestri:
The checkpointer now assumes that every model has a grid, clock, and particles. This PR switches this definition for a function because some model (for example the `SeaIceModel`) might not have one of these properties.


navidcy:
duplicate of https://github.com/CliMA/Oceananigans.jl/pull/4250
@simone-silvestri I'll work on this there!

simone-silvestri:
I ll close then

PR #4375, opened 2025-04-09
Add a performance benchmark pipeline

simone-silvestri:
Which can be triggered on PRs by adding the `benchmark performance` label.
These tests at the moment run on Tartarus GPU number 1. We probably need a better GPU / a GPU we know is free to perform these tests.


simone-silvestri:
I convert this PR to ready to review just to see if GHA starts, however, it is not really ready to review

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,86 @@
...
+          - ubuntu-latest
+        arch:
+          - x64
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: actions/cache@v4
```
vchuravy, line 52: Better to use the `cache-julia` step.

simone-silvestri:
Just as a test, I have added here a `pipeline-benchmarks.yml` that runs on Tartarus (GPU 0) only if adding the `benchmark performance` label to the PR. 
The pipeline uses nsys to benchmark a representative script and spits out statistics with `nsys stats`. 
I am not sure if this is the correct approach. Is there a way to call the pipeline through a comment in the PR and have it spit out the results in a comment?
@vchuravy

.buildkite/pipeline-benchmarks.yml
```diff
@@ -0,0 +1,97 @@
+env:
+  JULIA_VERSION: "1.10.9"
+  JULIA_MINOR_VERSION: "1.10"
+  TARTARUS_HOME: "/storage5/buildkite-agent"
+  JULIA_DEPOT_PATH: "$TARTARUS_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
+  JULIA_PKG_SERVER_REGISTRY_PREFERENCE: eager
+  JULIA_NUM_PRECOMPILE_TASKS: 24
+  JULIA_NUM_THREADS: 8
+  NSYS: "/storage6/simone/new_nsight/bin/nsys"
```
glwagner, line 9:
```suggestion
  NSYS: "/storage6/nsight/bin/nsys"
```

PR #3475, opened 2024-02-13
AMD GPU Support via an extension for `AMDGPU`

fluidnumerics-joe:
This PR replaces #3468 - editing is allowed by maintainers.

PR #4376, opened 2025-04-09
Add an extension for AMDGPU.jl

amontoison:
The modifications are based on #4124.
It should superseded #3475.


glwagner:
Beautiful! Let us know your ideas for adding tests to CI!

glwagner:
That sounds great if JuliaGPU is willing to help us out!

I don't think we need Metal, because we can (apparently) test that through GHA. At least right now.

amontoison:
I think it is fine to use `JuliaGPU` for small tests of `Oceananigans.jl` for AMD and Intel GPUs.

amontoison:
The modifications of this PR are in #4381.

Issue #4379, opened 2025-04-10
Turbulence closure tests taking a very long time


simone-silvestri:
they are the bottleneck of our testing, but these are rather simple tests so the time spent in them should not be this large. 

There is probably some inference issue that makes them take such a long time.

navidcy:
On CPU or GPU (or both)?
I think I did notice that GPU turbulence closure test suite is the last to finish...

tomchor:
@simone-silvestri should we close this since we got a 30 min speedup on the GPU test due to https://github.com/CliMA/Oceananigans.jl/pull/4380?

Or do you feel like we can do better?

glwagner:
I'd like to close this, please reopen if you disagree

PR #4380, opened 2025-04-10
Reduce turbulence closure tests cost

simone-silvestri:
this PR is here to try reducing the cost of the turbulence closures tests (> 1hr for GPU tests on main!)

see #4379


simone-silvestri:
There is still some investigation to do here, it looks like there is quite some compilation time, but also, for `LagrangianAveragedDynamicSmagorinsky` the cost of building a model is huge (also locally). 
The bottleneck seems to be https://github.com/CliMA/Oceananigans.jl/blob/2fec5015612f1b67183ca9368de7d16f7db52a63/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl#L136-L143
This seems like an innocuous kernel, but I think it has many repeated interpolations that might stress a bit the computation

glwagner:
I tried splitting those into two kernels lets see if it helps

glwagner:
Another idea is to make this struct `mutable`:

https://github.com/CliMA/Oceananigans.jl/blob/73df6ac5f6b25a67bd1cd210a71f0d94d480f7b7/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl#L5

ill test that once the current CI reveals the impact of this change

glwagner:
I think that the changes I made reduced test time down to 46 min.

I believe further improvements would be had by writing out the filter function manually rather than using function composition (not that because it is a 3D filter, the function composition introduces 6 levels of indirection which is perhaps too high).

However we could merge now since I think the objective of the PR has been achieved.

glwagner:
@tomchor may be interested in the improvements here

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -2,13 +2,20 @@ using Oceananigans.Architectures: architecture
 using Oceananigans.Fields: interpolate
 using Statistics
 
-struct DynamicCoefficient{A, FT, S}
+mutable struct DynamicCoefficient{A, FT, S}
```
tomchor, line 5: `CPUDynamicCoefficient` for consistency? Then we can define `DynamicCoefficient = Union{CPUDynamicCoefficient, GPUDynamicCoefficient}` and we (I think) retain the same behavior for previously-defined methods methods.

test/test_turbulence_closures.jl
```diff
@@ -476,16 +481,29 @@ end
...
-            closure = @eval $closurename()
-            compute_closure_specific_diffusive_cfl(closure)
-        end
+        test_closures = (
+            ScalarDiffusivity(),
+            ScalarBiharmonicDiffusivity(),
+            TwoDimensionalLeith(),
+            ConstantSmagorinsky(),
+            SmagorinskyLilly(),
+            # LagrangianAveragedDynamicSmagorinsky(),
```
tomchor, line 490: Are the tests faster because this is commented out?

test/test_turbulence_closures.jl
```diff
@@ -476,16 +481,29 @@ end
     end
 
     @testset "Diagnostics" begin
-        @info "  Testing turbulence closure diagnostics..."
-        for closurename in closures
-            closure = @eval $closurename()
-            compute_closure_specific_diffusive_cfl(closure)
-        end
+        test_closures = (
```
tomchor, line 484: I also don't get why this is being defined here. Isn't it the same as the definition on `closures` in `dependencies_for_runtests.jl`?

tomchor:
Thanks for taking a look into this @simone-silvestri. Have you tested what impact this has on run time? I'd be reluctant to gain ~30 min on tests, if it means a slowdown in runtime for this already-slow closure.

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
glwagner: No because `DynamicCoefficient` is also used for GPU. The distinction between these two objects is more like `CuArray` versus `CuDeviceArray`. Both are for GPU, but only `CuDeviceArray` is used _inside_ kernels. Does that make sense? `GPUDynamicCoefficient` is not user facing, it only crops up internally inside kernels.

glwagner:
@tomchor can you do some benchmarking? I would be surprised if there is a runtime slowdown.

We can still accept a runtime slowdown in this case in my opinion, because Oceananigans is about more than just dynamic smag so we can't prioritize it above everything else.

test/test_turbulence_closures.jl
glwagner, line 490:
```suggestion
            LagrangianAveragedDynamicSmagorinsky(),
```

test/test_turbulence_closures.jl
glwagner: maybe he put this in for testing

glwagner: because then you could change the closures being tested locally in this file and figure out which ones are taking so much time?

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
glwagner: made that change to clarify (there is also a `GPUFieldTimeSeries`, we can clean that up in another PR)

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/scale_invariant_operators.jl
```diff
@@ -41,11 +41,19 @@ using Oceananigans.Operators: volume
...
+# @inline ℱ₂ᶜᶜᶠ(i, j, k, grid, f, args...) = ℑxyzᶜᶜᶠ(i, j, k, grid, ℑxyzᶠᶠᶜ, f, args...)
+# @inline ℱ₂ᶜᶜᶜ(i, j, k, grid, f, args...) = ℑxyzᶜᶜᶜ(i, j, k, grid, ℑxyzᶠᶠᶠ, f, args...)
+
+@inline ℱ₂ᶠᶜᶜ(i, j, k, grid, u::AbstractArray) = @inbounds (6 * u[i, j, k] +
+                                                            u[i+1, j, k] + u[i-1, j, k] +
+                                                            u[i, j+1, k] + u[i, j-1, k] +
+                                                            u[i, j, k+1] + u[i, j, k-1]) / 12
+const ℱ₂ᶜᶠᶜ = ℱ₂ᶠᶜᶜ
+const ℱ₂ᶜᶜᶠ = ℱ₂ᶠᶜᶜ
+const ℱ₂ᶜᶜᶜ = ℱ₂ᶠᶜᶜ
```
glwagner, line 55: @tomchor can you take a look at this change?

I was looking at these operators and I realized that the 4 operators are identical (at least in the case that the reconstruction operators do not depend on grid topology which they don't now). Therefore we only need one filtering function. Also, we only need it for arrays (not functions). I'd like to see how this plays out with compile time but I suspect it will help, since we are compiling just 1 function instead of... 4 * 6 = 24.

tomchor:
> We can still accept some level runtime slowdown in this case in my opinion --- depending on what that is --- because Oceananigans is about more than just dynamic smag so we can't prioritize it above everything else.

That makes sense, and I agree with your argument. However, I'd argue that LES is also an important part of Oceananigans and dynamic Smag is the most well-validated explicit model we have, so it needs to be as performant as we can make it so that we can run state-of-the-art science.

If testing it is becoming a burden, maybe we can consider delegating LES closures to a separate package. Just like everything that goes into `model.biogeochemistry` comes from separate packages, maybe everything `model.closure` should also go to a separate package where it can be maintained separately. Then the "core" Oceananigans would only do implicit  LES, and anything else would require an extra package.

glwagner:
I don't think it makes sense to put the LES closures in a different package.

glwagner:
@tomchor did you want to do any benchmarking? I think this PR is ready to go. Compile times are way down. I am curious whether dynamic smag performance is also improved, but we can look at that later.

tomchor:
I haven't had the change to benchmark it on a GPU yet, but on CPU it seems that compilation is about half the time and runtime also went down to about 75% of the time! :tada: Nice work!

I can benchmark on GPUs tomorrow.

tomchor:
For completeness, there is a similar speedup on GPUs.

glwagner:
runtime is down 75%?!?

tomchor:
> runtime is down 75%?!?

Sorry, no. It's down to 75% of what it was originally. Still a pretty good reduction.

glwagner:
I think the compilation cost may be related to fixing the filtering function, but the runtime improvement seems more likely from isolating the kernel that computes the filtered strain tensor. Not sure. Or just improved inference.

test/test_turbulence_closures.jl
tomchor: Hmm, just realizing we forgot to delete this.

PR #4382, opened 2025-04-10
Remove more splatting

simone-silvestri:
splatting seems to be a very large penalty when done with many different arguments, this PR fixes more performance issue with implicit solvers that seems to depend on the splitting of the arguments inside the flux calculation


simone-silvestri:
I didn't realize that I was removing the implicit step. The search continues...

PR #4383, opened 2025-04-10
Try fixing performance

simone-silvestri:
We still have a performance issue related to the implicit solver. It is very difficult to diagnose. I tried to remove splatting where possible and the `Val(tracer_index)` to see if that creates weird problems. This PR is open just to see if the tests speed up with this change.

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
@@ -172,7 +172,7 @@ end
     end
 end
 
-@inline function implicit_linear_coefficient(i, j, k, grid, closure::FlavorOfCATKE{<:VITD}, K, ::Val{id}, args...) where id
+@inline function implicit_linear_coefficient(i, j, k, grid, ::FlavorOfCATKE{<:VITD}, K, id, args...)
     L = K._tupled_implicit_linear_coefficients[id]
     return @inbounds L[i, j, k]
 end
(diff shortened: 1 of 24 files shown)


glwagner:
might be worthing lookikng at this commit 

https://github.com/CliMA/Oceananigans.jl/commit/8dfb7636cae078ae799fb54b9211925cea281258

simone-silvestri:
Ok, it seems that we have no performance hit, I was mistaken. However, this PR removes a lot of unnecessary splatting making the code clearer, so I would still consider merging it.

simone-silvestri:
(The `@muladd` creeped into here so I will remove it before merging)

simone-silvestri:
This should be ready to merge

simone-silvestri:
@glwagner @navidcy this PR should be ready to merge, I would need a review

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
```diff
@@ -204,6 +206,8 @@ end
     #                  = Lⁱ
     #
     # where ω = ϵ / e ∼ √e / ℓ.
+    
+    active = !inactive_cell(i, j, k, grid)
```
navidcy: this was missing? how was it working previously?

simone-silvestri:
The difference with main is that, not passing `fields` to the implicit solver we could not use field-dependent diffusivities. 
You cannot see it in the methods because the args are splatted in the implicit solver.
In the end performance did not change, but this change makes everything more explicit and understandable from a source code perspective

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
simone-silvestri: oh right, this is extra. I can remove it

navidcy: if it's needed leave it
I just saw that and it seemed beyond the "removing unnecessary splatting" as the PR title suggested so I was wondering...

PR #4386, opened 2025-04-11
Remove the `HeptadiagonalSolver` 

simone-silvestri:
and make the PCG the default. We need some benchmarking before merging this PR, at the time this solver was implemented the PCG was extremely slow (about 3 times difference) which motivated implementing a new strategy. 

A lot of time has passed, so this solver which is more memory demanding might not be necessary anymore, but before merging we should do some benchmarking.


simone-silvestri:
After improving the fill halos, I think this PR might be ready to be merged (if tests pass)

simone-silvestri:
If someone still wants this feature inside oceananigans, speak now or forever hold your peace :)

Project.toml
```diff
@@ -75,8 +74,7 @@ GPUArrays = "10.3, 11.2.2"
 GPUArraysCore = "0.2"
 Glob = "1.3"
 InteractiveUtils = "1.9"
-IterativeSolvers = "0.9"
-JLD2 = "0.4, 0.5, 0.6"
+JLD2 = "0.4, 0.5"
```
navidcy: why drop the JLD2 0.6 compat? on purpose or accidental?

Project.toml
```diff
@@ -119,4 +117,4 @@ TimesDates = "bdfc003b-8df8-5c39-adcd-3a9087f5df4a"
 oneAPI = "8f75cd03-7ff8-4ecb-9b8f-daf728133b1b"
 
 [targets]
-test = ["AMDGPU", "CUDA", "oneAPI", "DataDeps", "SafeTestsets", "Test", "Enzyme", "Reactant", "Metal", "CUDA_Runtime_jll", "MPIPreferences", "TimesDates", "NCDatasets"]
+test = ["AMDGPU", "CUDA", "oneAPI", "DataDeps", "SafeTestsets", "Test", "Enzyme", "Reactant", "Metal", "CUDA_Runtime_jll", "MPIPreferences", "TimesDates", "NCDatasets"]
```
navidcy:
```suggestion
test = ["AMDGPU", "CUDA", "oneAPI", "DataDeps", "SafeTestsets", "Test", "Enzyme", "Reactant", "Metal", "CUDA_Runtime_jll", "MPIPreferences", "TimesDates", "NCDatasets"]

```

Project.toml
simone-silvestri: ah just accidental, I ll correct it

test/test_implicit_free_surface_solver.jl
```diff
@@ -54,7 +53,7 @@ function run_implicit_free_surface_solver_tests(arch, grid, free_surface)
     set_simple_divergent_velocity!(model)
     step_free_surface!(model.free_surface, model, model.timestepper, Δt)
 
-    acronym = free_surface.solver_method == :HeptadiagonalIterativeSolver ? "Matrix" : "PCG"
+    acronym = "PCG"
 
     η = model.free_surface.η
     @info "    " * acronym * " implicit free surface solver test, norm(η_" * lowercase(acronym) * "): $(norm(η)), maximum(abs, η_" * lowercase(acronym) * "): $(maximum(abs, η))"
```
navidcy: do we even need to define `acronym` then? just write it explicitly below?

navidcy:
bump patch?

navidcy:
happy for this to be merged

glwagner:
> bump patch?

true, its breaking.

PR #4389, opened 2025-04-13
add bulidkite pipeline-juliagpu for AMD and oneAPI testing

glwagner:


vchuravy:
You will have to restrict the test to Float32, we only have consumer level hardware available for Intel.

glwagner:
looks like Float64 isn't supported on the oneAPI device on julia GPU. @vchuravy what device is that?

vchuravy:
It's "ntel(R) Iris(R) Xe Graphics"

glwagner:
the tests pass! yay

glwagner:
how are they so fast. thats crazy

PR #4390, opened 2025-04-13
fix: better sharding constructs

avik-pal:


avik-pal (commented):
- changes the nothing trick to use Replicated
- removes `z` sharding since that is causing some optimization issues on enzymejax end

glwagner (approved):

avik-pal:
https://github.com/EnzymeAD/Reactant.jl/pull/1172 is needed for an ambiguity fix

glwagner:
> [EnzymeAD/Reactant.jl#1172](https://github.com/EnzymeAD/Reactant.jl/pull/1172) is needed for an ambiguity fix

copy

avik-pal:
Should be good to go. Reactant version has been tagged

PR #4392, opened 2025-04-13
Change isopycnal_skew_symmetric_diffusivity.jl to use DiffusiveFormulation

glwagner:
i noticed that the only example we have of using this in ClimaOcean used the DiffusiveFormulation. Therefore, it should be the default.


glwagner:
cc @navidcy

glwagner:
I'm thinking this makes sense because `AdvectiveFormulation` is not compatible with all advection schemes (notably it is not compatible with the default advection scheme). So DiffusiveFormulation is a better default.

simone-silvestri (approved):

PR #4394, opened 2025-04-14
Small improvements to `PerturbationAdvectionOpenBoundaryCondition` interface

tomchor:


src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -37,15 +37,25 @@ struct PerturbationAdvection{VT, FT}
...
+Creates a `PerturbationAdvectionOpenBoundaryCondition` with a given `outflow_timescale` and
+`inflow_timescale`. `backward_step` determines whether we assume a backward and forward time
+discretization. For details about this method, refer to the docstring for `PerturbationAdvection`.
+"""
+function PerturbationAdvectionOpenBoundaryCondition(val, FT = Float64;
                                                     backward_step = true,
-                                                    outflow_timescale = Inf, 
-                                                    inflow_timescale = 300.0, kwargs...)
+                                                    outflow_timescale = Inf,
+                                                    inflow_timescale = 0.0, kwargs...)
```
glwagner, line 58:
```suggestion
                                                    inflow_timescale = 0, kwargs...)
```

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -114,7 +114,7 @@ end
 
     τ̃ = Δt / τ
 
-    u₁ⁿ⁺¹ = (uᵢⁿ - U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ - U)
+    u₁ⁿ⁺¹ = isinf(τ̃) ? ūⁿ⁺¹ : (uᵢⁿ - U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ - U)
```
glwagner, line 117: can you organize this code into "paragraphs" (we don't want to use double space code I feel)

tomchor: Is this what you're suggesting?

```suggestion
   if isinf(τ̃)
        u₁ⁿ⁺¹ =  ūⁿ⁺¹
    else
        u₁ⁿ⁺¹ =  (uᵢⁿ - U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ - U)
    end
```

glwagner: no no I just mean removing some vertical white space

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -114,7 +114,7 @@ end
 
     τ̃ = Δt / τ
```
glwagner, line 116:
```suggestion
    τ̃ = Δt / τ
```

for example

eg why do we have so much white space here

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
tomchor: Ah I see. Yeah I agree with you. I also find it harder to read. I'll improve that.

tomchor: Done. Can you check?

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -62,23 +72,18 @@ const FPAOBC = BoundaryCondition{<:Open{<:PerturbationAdvection{Val{false}}}}
...
     U = max(0, min(1, Δt / ΔX * ūⁿ⁺¹))
 
     pa = bc.classification.matching_scheme
-
     τ = ifelse(ūⁿ⁺¹ >= 0, pa.outflow_timescale, pa.inflow_timescale)
-
     τ̃ = Δt / τ
 
-    uᵢⁿ⁺¹ = (uᵢⁿ + U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ + U)
+    uᵢⁿ⁺¹ = isinf(τ̃) ? ūⁿ⁺¹ : (uᵢⁿ + U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ + U)
```
jagoosw, line 86: Shouldn't this be written `uᵢⁿ⁺¹ = ifelse(isinf(τ̃), ūⁿ⁺¹, (uᵢⁿ + U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ + U))`?

tomchor: I thought both had similar performance. Is what I wrote short-circuiting?

tomchor:
@jagoosw can you please write a couple of sentences about `inflow_timescale` and `outflow_timescale`? Maybe expand a bit about the implications about the choice of backward vs forward stepping method?

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
glwagner: Yes, `?` is identical to explicit `if`. Only `ifelse` evaluates both branches.

tomchor:
> Sorry it took me so long to get to this. I have added some sentences about the timescales but am not sure what to put for the stepping options. I think we should have removed the forward stepping before merging because I can't convince myself that it is valid.

I'm okay with that. @glwagner are you okay with this?

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -42,10 +54,20 @@ Adapt.adapt_structure(to, pe::PerturbationAdvection) =
...
+    PerturbationAdvectionOpenBoundaryCondition(val, FT = Float64;
+                                               backward_step = true,
+                                               outflow_timescale = Inf,
+                                               inflow_timescale = 0, kwargs...)
+
+Creates a `PerturbationAdvectionOpenBoundaryCondition` with a given `outflow_timescale` and
+`inflow_timescale`. `backward_step` determines whether we assume a backward and forward time
+discretization. For details about this method, refer to the docstring for `PerturbationAdvection`.
+"""
 function PerturbationAdvectionOpenBoundaryCondition(val, FT = Float64;
```
glwagner, line 67:
```suggestion
function PerturbationAdvectionOpenBoundaryCondition(val, FT = Oceananigans.defaults.FloatType;
```

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -42,10 +54,20 @@ Adapt.adapt_structure(to, pe::PerturbationAdvection) =
...
                                                     backward_step = true,
                                                     outflow_timescale = Inf,
-                                                    inflow_timescale = 300.0, kwargs...)
+                                                    inflow_timescale = 0, kwargs...)
```
glwagner, line 70: 0 is not a valid choice, right? this will make the model blow up. If there's no good default, better just to make it required

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -62,23 +84,20 @@ const FPAOBC = BoundaryCondition{<:Open{<:PerturbationAdvection{Val{false}}}}
...
     pa = bc.classification.matching_scheme
-
     τ = ifelse(ūⁿ⁺¹ >= 0, pa.outflow_timescale, pa.inflow_timescale)
-
     τ̃ = Δt / τ
 
-    uᵢⁿ⁺¹ = (uᵢⁿ + U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ + U)
+    uᵢⁿ⁺¹ = ifelse(isinf(τ̃),
+                   ūⁿ⁺¹,
+                   (uᵢⁿ + U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ + U))
```
glwagner, line 100:
```suggestion
    u★ = (uᵢⁿ + U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ + U)
    uᵢⁿ⁺¹ = ifelse(isinf(τ̃), ūⁿ⁺¹, u★)
```

glwagner:
I'm ok with any improvement. I don't know what "forward stepping" is but with more detail I can give feedback on that.

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
tomchor: @jagoosw can correct me if I'm wrong, but `0` is a valid choice here and it means that the inflow is immediately set to the exterior condition (`val` in this case). Numerically it needs to be treated with a bit of care (hence the `ifelse` statements I included) and mathematically the limit is well-defined too.

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -42,10 +54,20 @@ Adapt.adapt_structure(to, pe::PerturbationAdvection) =
...
 
+"""
+    PerturbationAdvectionOpenBoundaryCondition(val, FT = Float64;
+                                               backward_step = true,
+                                               outflow_timescale = Inf,
+                                               inflow_timescale = 0, kwargs...)
+
+Creates a `PerturbationAdvectionOpenBoundaryCondition` with a given `outflow_timescale` and
+`inflow_timescale`. `backward_step` determines whether we assume a backward and forward time
+discretization. For details about this method, refer to the docstring for `PerturbationAdvection`.
```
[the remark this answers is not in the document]
tomchor: `val` is the exterior value of the normal valocity. I agree it isn't a very intuitive name. Maybe just `exterior_value`?

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
glwagner: okay, after I wrote this I noticed that there is a check for `inflow_timescale==0` basically (although not written that way yet)

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -56,100 +74,47 @@ end
...
     pa = bc.classification.matching_scheme
-
     τ = ifelse(ūⁿ⁺¹ >= 0, pa.outflow_timescale, pa.inflow_timescale)
-
     τ̃ = Δt / τ
 
-    uᵢⁿ⁺¹ = (uᵢⁿ + U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ + U)
+    uᵢⁿ⁺¹ = ifelse(τ == 0,
+                   ūⁿ⁺¹,
+                   (uᵢⁿ + U * uᵢ₋₁ⁿ⁺¹ + ūⁿ⁺¹ * τ̃) / (1 + τ̃ + U))
```
glwagner, line 93: please put the calculation on its own line and assign a variable name (avoid computations inside ifelse statements)

glwagner (approved):

jagoosw:
> I'm ok with any improvement. I don't know what "forward stepping" is but with more detail I can give feedback on that.

At the moment you can make the boundary do a forward or backwards Euler step and backwards is the default. I did some testing with forward stepping (I can't remember why) but I was never convinced that it was sound because we have the interior values at almost $t(n+1)$ not $t(n)$, but then I didn't remove it before we merged the PR (again, don't remember why).

tomchor:
> > I'm ok with any improvement. I don't know what "forward stepping" is but with more detail I can give feedback on that.
> 
> At the moment you can make the boundary do a forward or backwards Euler step and backwards is the default. I did some testing with forward stepping (I can't remember why) but I was never convinced that it was sound because we have the interior values at almost t ( n + 1 ) not t ( n ) , but then I didn't remove it before we merged the PR (again, don't remember why).

I also did a bit of testing and (also don't remember why, but) ended up not liking the forward step too much. So I support removing it :)

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -56,100 +73,45 @@ end
...
-@inline function step_right_boundary!(bc::BPAOBC, l, m, boundary_indices, boundary_adjacent_indices,
+@inline function step_right_boundary!(bc::PAOBC, l, m, boundary_indices, boundary_adjacent_indices,
                                       grid, u, clock, model_fields, ΔX)
     Δt = clock.last_stage_Δt
-
     Δt = ifelse(isinf(Δt), 0, Δt)
 
     ūⁿ⁺¹ = getbc(bc, l, m, grid, clock, model_fields)
-
     uᵢⁿ     = @inbounds getindex(u, boundary_indices...)
```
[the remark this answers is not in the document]
jagoosw: This is where we get the previous value of the boundary point since it hasn't been stepped yet so its still at the previous time step, and then put `uⁿ⁺¹` in the same place to bring the whole field to the current time step.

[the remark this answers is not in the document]
tomchor: Ah I see, thanks for the clarification. Just to clarify even further, the way time-stepping works is that the interior of the flow is time-stepped first, then we do the boundaries?

PR #4396, opened 2025-04-14
Add extension to ConservativeRegridding.jl

glwagner:
This PR adds an extension to ConservativeRegridding which ultimately should allow us to "regrid" fields conservatively between grids with different resolutions or alignments. "Conservative" means that regridding conserves the mean value of the field. This should be useful for nesting as well as coupling.

### DIFF
--- a/ext/OceananigansConservativeRegriddingExt.jl
+++ b/ext/OceananigansConservativeRegriddingExt.jl
@@ -1,0 +1,103 @@
+module OceananigansConservativeRegriddingExt
+
+import GeoInterface as GI, GeometryOps as GO
+
+using ConservativeRegridding
+
+using Oceananigans
+using Oceananigans.Grids: ξnode, ηnode
+using Oceananigans.Fields: AbstractField
+using KernelAbstractions: @index, @kernel
+
+instantiate(L) = L()
+
+function compute_cell_matrix(field::AbstractField)
+    Fx, Fy, _ = size(field)
+    LX, LY, _ = Oceananigans.Fields.location(field)
+    ℓx, ℓy = LX(), LY()
+
+    if isnothing(ℓx) || isnothing(ℓy)
+        throw(ArgumentError("cell_matrix can only be computed for fields with non-nothing horizontal location."))
+    end
+
+    grid = field.grid
+    arch = grid.architecture
+    FT = eltype(grid)
+
+    vertices_per_cell = 5 # convention: [sw, nw, ne, se, sw]
+    ArrayType = Oceananigans.Architectures.array_type(arch)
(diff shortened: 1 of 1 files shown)


glwagner:
This depends on https://github.com/JuliaGeo/ConservativeRegridding.jl/pull/7

@milankl @simone-silvestri

ext/OceananigansConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,103 @@
...
+        cell_matrix[5, linear_idx] = (xsw, ysw)
+    end
+end
+
+function ConservativeRegridding.Regridder(dst_field, src_field)
+    src_cells = compute_cell_matrix(src_field)
+    dst_cells = compute_cell_matrix(dst_field)
+
+    src_polygons = GI.Polygon.(GI.LinearRing.(eachcol(src_cells))) .|> GO.fix
+    dst_polygons = GI.Polygon.(GI.LinearRing.(eachcol(dst_cells))) .|> GO.fix
```
milankl, line 90: I should only be necessary to define a new method for `ConservativeRegridding.get_vertices` I think!

milankl: But `cell_matrix` is a matrix, so you don't have to do anything!

ext/OceananigansConservativeRegriddingExt.jl
```diff
@@ -0,0 +1,103 @@
...
+
+    return ConservativeRegridding.Regridder(src_polygons, dst_polygons)
+end
+
+ConservativeRegridding.regrid!(dst_field::Field, regridder::ConservativeRegridding.regridder, src_field::AbstractField) =
+    regrid!(vec(interior(dst_field)), regridder, vec(interior(src_field)))
```
milankl, line 96: this wouldn't work for the 3D cases right? In that case you only want to unravel the horizontal dimensions into a vector

glwagner: That's true, but I think it will be easy to support such a thing in this PR by making a stack of `vec(interior(field, :, :, k))`

ext/OceananigansConservativeRegriddingExt.jl
glwagner: interesting! I think we may want to exert some control here but not sure. For example, we probably want to interpret user input and throw some specific errors. We also want to support regridding in vertical slices and whatnot. So I believe we will want to extend `regrid!` for Field? Let's see though.

glwagner:
This is growing stale. It will be easy to revisit when we are ready for it.

Issue #4395, opened 2025-04-14
Improve and Verify Z-Star Tracer Conservation Tests


NoraLoose:
This issue is the result of a deep dive into the [z-star tracer conservation tests](https://github.com/CliMA/Oceananigans.jl/blob/main/test/test_zstar_coordinate.jl), with the goal of assessing whether these tests reliably detect tracer non-conservation. Specifically, I explored:
* Are the tests run long enough?
* Is the physical setup appropriate for catching conservation errors?

To evaluate this, I re-ran the z-star tracer conservation tests using the z-coordinate instead, keeping the random seed fixed to match the one used in the original tests.. This provides a strong sanity check: since the z-coordinate does not conserve tracers, the tests should fail — and they do, which is great.

Currently, each test is run for 100 time steps:
* Explicit free surface cases: 10s × 100 = 1000s = 16.67 minutes
* Implicit free surface cases: 2min × 100 = 200 minutes = 3.33 hours

### Key Observations

The relative tracer conservation error for the z-coordinate is plotted below. This mirrors the behavior [tested in the code](https://github.com/CliMA/Oceananigans.jl/blob/bd708ef1a8a5af683cf3677fe3e3f3849750c4da/test/test_zstar_coordinate.jl#L35), which uses the `≈` operator to compare the domain-integrated tracer at the desired time step to its initial value. By default, `≈` checks for approximate equality using a relative tolerance determined by the used precision ([see the docs](https://docs.julialang.org/en/v1/base/math/#Base.isapprox)).

![Image](https://github.com/user-attachments/assets/8491d2be-a0d4-4411-ae0e-b1157be704d0)
![Image](https://github.com/user-attachments/assets/cad2f034-a538-4e5c-925d-8291db6bf573)

**EDIT**: The labels in the legend are explained as follows:
* explicit/implicit/split: describes free surface type
* rect/ll/tripolar: describes grid type, i.e., rectilinear, lat-lon, tripolar
* prefix ib-: immersed boundary
* PPB/PBB/...: describes topology, P = periodic, B = bounded
* uniform/stretched: describes whether uniform or stretched z-levels were used

**Take-aways**:
1. **All tests would fail for the z-coordinate**, which is expected and desirable. (Contrary to a previous comment I made, which was based on now-unreproducible behavior in Julia 1.11.4 — good news!)
[...]

PR #4403, opened 2025-04-15
Improve z-star tracer conservation tests

NoraLoose:
This PR implements the improvements discussed in [#4395](https://github.com/CliMA/Oceananigans.jl/issues/4395):

- Tracer conservation is now tested at every time step, rather than only at the end of the simulation.
- Early stopping is triggered if any conservation test fails, reducing unnecessary computation.
- In the tripolar test, the simulation now starts from randomly initialized velocities instead of rest, helping detect conservation issues earlier.
- The tripolar grid has been reduced in size to improve computational efficiency.
- Tests using a uniform vertical coordinate have been removed to reduce overall test runtime.

Closes [#4395](https://github.com/CliMA/Oceananigans.jl/issues/4395).


NoraLoose:
The new tripolar grid test fails locally for me.

Great — in [#4395](https://github.com/CWorthy-ocean/roms-tools/issues/4395) I had only investigated whether there is a significant deviation for the `z`-coordinate, but I didn't test with `zstar`.

@simone-silvestri, any idea why the modified tripolar test might be failing?

NoraLoose:
> @simone-silvestri, any idea why the modified tripolar test might be failing?

Does it have to do with the fact that the values in the tripolar seem are not appropriately duplicated?

simone-silvestri:
I think there is something wrong with initialization with random velocities. You end up with a state that is not "possible", because you have a divergent barotropic velocity field that is not balanced by a suitable `∂t_σ` (you did not initialize the free surface) and, for this reason, you get a vertical velocity at the top of the domain.

If this suspicion is right
1. the tracer non-conservation happens at the first time step
2. it would happen for every grid
3. initializing with a horizontally divergent-free velocity field would lead to tracer conservation

If you verify this hypothesis (maybe with point 3) we can start initializing with a divergence-free velocity field and in the meantime I ll think how to ensure that the initial state is a valid initialization even when the user provides random divergent velocities.

NoraLoose:
> If this suspicion is right
> 
> 1. the tracer non-conservation happens at the first time step, and not after
> 2. it would happen for every grid provided you initialize with random velocities
> 3. initializing with a horizontally divergent-free velocity field would lead to tracer conservation

Update on point 2:
* Tracer/buoyancy **non-conservation** occurs across all grids using `SplitExplicitFreeSurface`, failing immediately after the first time step when initialized with random velocities.
* In contrast, all grids using `ImplicitFreeSurface` or `ExplicitFreeSurface` maintain conservation of tracers and buoyancy throughout the test period, even when initialized with random velocities.

simone-silvestri:
Ok that is consistent with the hypothesis of incorrect initialization I think

NoraLoose:
Okay, in 853e0fd I changed from random initial velocities to velocities inferred from a random initial streamfunction, to ensure divergence-free initial velocities.

My internal tests show the following:

tracer_conservation_error_random_initial_stream_function_smaller_grid_tripolar_1 10 9
buoyancy_conservation_error_random_initial_stream_function_smaller_grid_tripolar_1 10 9

* The z-star coordinate now conserves tracers.
* The z-coordinate does not conserve tracers, but the relative error is small after 100 time step (see vertical dashed line). I will see if I can change the setup so that we get a more significant violation or tracer conservation with the z-coordinate early on.

NoraLoose:
Let's maybe open an issue about divergent initial velocities creating non-tracer-conservation? I can do that.

NoraLoose:
@simone-silvestri Before merging this PR, I'd like to experiment a bit more with the tripolar setup to see if we can observe a more pronounced tracer conservation violation with the z-coordinate early on. I’ll mark the PR as ready for review once I’ve wrapped up that investigation.

simone-silvestri:
sounds good!

NoraLoose:
Okay, I think we have two options to safely trigger an error for tracer non-conservation with the z-coordinate:

1. Run a (50, 50, 20)-sized tripolar grid for 100 time steps (3.33 hours), as shown by the pink line up to the vertical dashed line.
2. Run a (20, 20, 20)-sized tripolar grid for about 300 time steps (10 hours) to reach the same error as with the larger grid at 100 time steps, as shown by the green line.

The second option seems cheaper, so I suggest that. Sounds good @simone-silvestri?

tracer_conservation_error_random_initial_stream_function_tripolar_1 10 9
buoyancy_conservation_error_random_initial_stream_function_tripolar_1 10 9

NoraLoose:
Are the [buildkite/oceananigans-distributed](https://buildkite.com/clima/oceananigans-distributed/builds/7355) tests failing because I opened this PR from a fork?

glwagner:
> Are the [buildkite/oceananigans-distributed](https://buildkite.com/clima/oceananigans-distributed/builds/7355) tests failing because I opened this PR from a fork?

might be, they are also very tempramental and need lots of encouragement to run

simone-silvestri (approved):

NoraLoose:
@simone-silvestri this PR should be ready to be merged, but I'm not able to due to failing merge requirements (the failing [buildkite/oceananigans-distributed](https://buildkite.com/clima/oceananigans-distributed/builds/7444) check?)

PR #4405, opened 2025-04-16
Mask implicit linear coefficient if inside an immersed boundary

simone-silvestri:
Inside immersed boundaries we should have the diagonal term equal to 1. For this reason, the implicit linear coefficient should be masked

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
@@ -168,7 +168,7 @@ end
     @inbounds begin
         total_Gⁿe = slow_Gⁿe[i, j, k] + fast_Gⁿe
         e[i, j, k] += Δτ * (α * total_Gⁿe - β * G⁻e[i, j, k]) * active
-        G⁻e[i, j, k] = total_Gⁿe * mask
+        G⁻e[i, j, k] = total_Gⁿe * active
     end
 end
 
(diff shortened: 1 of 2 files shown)


src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
```diff
@@ -119,6 +119,7 @@ end
     #       Lᵂ = - Cᵂϵ * √e / Δz.
 
     on_bottom = !inactive_cell(i, j, k, grid) & inactive_cell(i, j, k-1, grid)
+    mask = !inactive_cell(i, j, k, grid)
```
glwagner, line 122: can we call this `not_inactive`?

I'm worried after a discussion with @milankl that we use the term `mask` inconsistently. A mask is 1 when applied, 0 when not. But here, masking it defined by `mask=false`. For the time being I think a good choice is to use more specific names, until we come up with a clean and consistent terminology for this

glwagner, line 122:
```suggestion
    active = !inactive_cell(i, j, k, grid)
```

src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
```diff
@@ -141,7 +142,7 @@ end
     #
     # where ω = ϵ / e ∼ √e / ℓ.
 
-    @inbounds Le[i, j, k] = wb⁻_e - ω + div_Jᵉ_e
+    @inbounds Le[i, j, k] = (wb⁻_e - ω + div_Jᵉ_e) * mask
```
glwagner, line 145:
```suggestion
    @inbounds Le[i, j, k] = (wb⁻_e - ω + div_Jᵉ_e) * active
```

glwagner (approved):

navidcy:
@simone-silvestri, curious if you were running into issues from this?

PR #4406, opened 2025-04-16
Interpret user substeps to SplitExplicitFreeSurface as dividing just one time step

glwagner:
Possibly an alternative to #4348

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
@@ -42,10 +42,8 @@ Keyword Arguments
 
 - `gravitational_acceleration`: the gravitational acceleration (default: `g_Earth`)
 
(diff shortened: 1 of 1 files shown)


src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
```diff
@@ -250,12 +248,13 @@ end
 
 @inline function weights_from_substeps(FT, substeps, averaging_kernel)
 
-    τᶠ = range(FT(0), FT(2), length = substeps+1)
+    # To divide the range (t, t + 2dt):
+    Nτ = 2substeps + 1
```
glwagner, line 252: @simone-silvestri how can we determine the total substeps from the "user substeps" (ie the number of substeps that divide a time-step?

Do we need to introduce an interface where new averaging kernels are defined by:

1. The weights
2. The total number of substeps, as a function of the substeps that divide up a time step

?

We must be computing number 2 at some point (perhaps offline, or it is in a paper somewhere). We just need to elevate this information into this PR.

glwagner:
This won't work as is, more is needed

PR #4408, opened 2025-04-17
Further clean up the buildkite yml

glwagner:


glwagner:
negative diff energy right now

.buildkite/pipeline.yml
```diff
@@ -1,12 +1,14 @@
...
-  JULIA_VERSION: "1.10.8"
+  JULIA_VERSION: "1.10.9"
   JULIA_MINOR_VERSION: "1.10"
   TARTARUS_HOME: "/storage5/buildkite-agent"
+  JULIA_DEPOT_PATH: "$TARTARUS_HOME/.julia-$BUILDKITE_BUILD_NUMBER"
   JULIA_PKG_SERVER_REGISTRY_PREFERENCE: eager
-  JULIA_NUM_PRECOMPILE_TASKS: 24
-  JULIA_NUM_THREADS: 8
+  JULIA_NUM_PRECOMPILE_TASKS: 48
+  JULIA_NUM_THREADS: 24
```
simone-silvestri, line 8: I think this might be a bit too much on Tartarus when running multiple buildkite pipelines in parallel. we only have 64 cores.

glwagner: I’ve always been told that parallelism is composable. Are you concerned based on theory or do have specific info?

glwagner: I’m ok to reduce this but I am just curious. The idea is basically to let task scheduling decide what to do

simone-silvestri: I am not sure how smart the julia scheduler is. In general I don't see an extreme benefit of increasing the number of threads.

glwagner: The reason is that initialization takes 11 minutes right now, a significant portion of the total test time (and we don't learn about any failures until initialization finishes). So I am strategizing that decreasing initialization time is one of the more important ways we can speed up overall development time.

glwagner: But it looks like this change has no impact at all on initialization, so there's no reason to do it.

glwagner: I guess I am not sure whether you are referring to threads or precomplie tasks in this comment

PR #4409, opened 2025-04-17
Integrate `Krylov.jl` for pressure solve

xkykai:
This PR aims to integrate `KrylovSolver` for the pressure solve which has been developed into Oceananigans for the nonhydrostatic model. I begun by developing a new solver called `KrylovPoissonSolver`, which is a Poisson solver which uses Krylov methods available in Krylov.jl.

There are a few considerations on the design that I think would be helpful to hear your thoughts on:

1. Is there a place for both `ConjugateGradientPoissonSolver` and `KrylovPoissonSolver`? Should we keep both methods, or deprecate `ConjugateGradientPoissonSolver`?
3. There are multiple choices of preconditioners we can choose from, both from Oceananigans and Krylov.jl. For now I have kept the preconditioners in the file `src\Solvers\conjugate_gradient_poisson_solver.jl`, while putting the new `KrylovPoissonSolver` in `src\Solvers\krylov_poisson_solver.jl` which is also compatible with the preconditioners we have. Should we reorganize the files so that we put preconditioners in a separate file? Then perhaps their relations would be clearer.
4. Should we develop a unified interface to allow users to use preconditioners from Krylov.jl as well?
5. I am also not good at writing summaries, so if people have ideas on how to write good `Base.summary` I am happy to hear your thoughts on what's important!

@glwagner @simone-silvestri @amontoison @tomchor


amontoison:
I think a dedicated file for the preconditioners is a good idea.

The common API for preconditioners is `ldiv!(y, P, x)` or `mul!(y, P⁻¹, x)` (three arguments).
The current API in Oceananigans.jl is `precondition!` with a variable number of arguments.

It is because the update of the preconditioner and the application of it is fused in Oceananigans.jl.

simone-silvestri:
I would just remove the `ConjugateGradientSolver`

glwagner:

xkykai:
Turns out we have been able to make the Poisson matrix symmetric for all scenarios we have encountered so far, so conjugate gradient works for all cases. Even with the inclusion of free surface (see https://github.com/CliMA/Oceananigans.jl/pull/5706) the matrix is still symmetric, so we probably don't need other Krylov methods to solve for pressure (for now).

So closing it is probably fine.

PR #4410, opened 2025-04-17
Fix rendering + reference `default_checkpointed_properties` in `Checkpointer` docstring

jbisits:
The `:properties` section of [Checkpointer docstring](https://clima.github.io/OceananigansDocumentation/dev/appendix/library/#Oceananigans.OutputWriters.Checkpointer-Tuple{Any}) was not rendering correctly because of a missing `. This PR adds this.

As well (and if you think unnecessary it can be removed) it elaborates on the default setting for `properties` when building a `Checkpointer`.


glwagner (approved):

glwagner:
I like it

PR #4411, opened 2025-04-17
Remove stretched advection

simone-silvestri:
I don't think it is used much or, at the moment, yields much better results.
In addition, I want to heavily rework the advection module by having the topological check _inside_ the reconstruction functions instead of _outside_ in order to significantly reduce the operations we are doing and having a simpler structure helps tremendously towards this goal (i.e. after removing the `coeff_*` from the advection types we will also be able to remove the `buffer_scheme` fields).

I envision around a factor 1.5 - 2 performance improvement by following this path as well as a tremendous code simplification.

We can add stretched advection later on, including also 2d coefficients for 2d grids like the orthogonal spherical shell grid (which was my plan), but a substantial rework anyway. So, in my opinion, it is probably better to excise everything and restart fresh once the advection is completely reworked.

If there are no objections, we can merge it. What do you think @glwagner, @navidcy, @tomchor, @ali-ramadhan?


src/Advection/adapt_advection_order.jl
```diff
@@ -87,24 +87,10 @@ function adapt_advection_order(advection::UpwindBiased{B}, N::Int, grid::Abstrac
...
-new_weno_scheme(::WENO, grid, order, bounds, ::Type{Nothing}, ::Type{Nothing}, ::Type{Nothing},) = WENO(; order, bounds)
-new_weno_scheme(::WENO, grid, order, bounds, XT, YT, ZT)                                         = WENO(grid; order, bounds)
-
-function adapt_advection_order(advection::WENO{B, FT, XT, YT, ZT}, N::Int, grid::AbstractGrid) where {B, FT, XT, YT, ZT}
+function adapt_advection_order(advection::WENO{B}, N::Int, grid::AbstractGrid) where B
     if N >= B
         return advection
     else
-        return new_weno_scheme(advection, grid, 2N-1, advection.bounds, XT, YT, ZT)
+        return WENO(; order=2N-1)
```
glwagner, line 94:
```suggestion
        return WENO(order=2N-1)
```

src/Advection/vector_invariant_advection.jl
```diff
@@ -426,33 +426,27 @@ end
...
+    
+        # Disambiguation
+        @inline symmetric_interpolate_xᶠᵃᵃ(i, j, k, grid::XFlatGrid, ::Centered{$b, $FT}, f::Function, ::AS, args...) = f(i, j, k, grid, args...)
+        @inline symmetric_interpolate_yᵃᶠᵃ(i, j, k, grid::YFlatGrid, ::Centered{$b, $FT}, f::Function, ::AS, args...) = f(i, j, k, grid, args...)
+        @inline symmetric_interpolate_zᵃᵃᶠ(i, j, k, grid::ZFlatGrid, ::Centered{$b, $FT}, f::Function, ::AS, args...) = f(i, j, k, grid, args...)
+            
+        # Disambiguation
+        @inline biased_interpolate_xᶠᵃᵃ(i, j, k, grid::XFlatGrid, ::UpwindBiased{$b, $FT}, bias, f::Function, ::AS, args...) = f(i, j, k, grid, args...)
+        @inline biased_interpolate_yᵃᶠᵃ(i, j, k, grid::YFlatGrid, ::UpwindBiased{$b, $FT}, bias, f::Function, ::AS, args...) = f(i, j, k, grid, args...)
+        @inline biased_interpolate_zᵃᵃᶠ(i, j, k, grid::ZFlatGrid, ::UpwindBiased{$b, $FT}, bias, f::Function, ::AS, args...) = f(i, j, k, grid, args...)
```
glwagner, line 450: We rarely want to dispatch on `Function`. The reason is that not all callable objects are `Function`. The practice we typically use is to use the function-based interface as a fallback and then add more specific methods for array (the usual other case)

glwagner (approved):
good negative diff, just a comment about dispatching on `Function` which is usually not what we want (it makes it so that callable objects can only access those methods by subtyping `Function`)

src/Advection/vector_invariant_advection.jl
glwagner: oh well that isn't a high priority --- was just making a comment not to add it

src/Advection/reconstruction_coefficients.jl
```diff
@@ -2,70 +2,19 @@ using Oceananigans.Grids: ξnodes, ηnodes, rnodes
...
-
-"""
-    @inline biased_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme, bias, ψ, args...)
-
-high order biased reconstruction of variable ψ in the z-direction. ψ can be a `Function`
-with signature `ψ(i, j, k, grid, args...)` or an `AbstractArray`. The `bias` argument is
-either `LeftBias` for a left biased reconstruction, or `RightBias` for a right biased reconstruction
-"""
-@inline biased_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme, bias, ψ, args...)  = inner_biased_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme, bias, ψ, k, Face, args...)
+@inline symmetric_interpolate_zᵃᵃᶜ(i, j, k, grid, scheme, ψ, args...) = symmetric_interpolate_zᵃᵃᶠ(i, j, k+1, grid, scheme, ψ, args...)
```
tomchor, line 10:
```suggestion
@inline symmetric_interpolate_zᵃᵃᶜ(i, j, k, grid, scheme, ψ, args...) = symmetric_interpolate_zᵃᵃᶠ(i, j, k + 1, grid, scheme, ψ, args...)
```

src/Advection/reconstruction_coefficients.jl
```diff
@@ -2,70 +2,19 @@ using Oceananigans.Grids: ξnodes, ηnodes, rnodes
...
-with signature `ψ(i, j, k, grid, args...)` or an `AbstractArray`
-"""
-@inline symmetric_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme, ψ, args...) = inner_symmetric_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme, ψ, k, Face, args...)
-
 """ same as [`symmetric_interpolate_xᶠᵃᵃ`](@ref) but on `Center`s instead of `Face`s """
-@inline symmetric_interpolate_xᶜᵃᵃ(i, j, k, grid, scheme, ψ, args...) = inner_symmetric_interpolate_xᶠᵃᵃ(i+1, j, k, grid, scheme, ψ, i, Center, args...)
+@inline symmetric_interpolate_xᶜᵃᵃ(i, j, k, grid, scheme, ψ, args...) = symmetric_interpolate_xᶠᵃᵃ(i+1, j, k, grid, scheme, ψ, args...)
 """ same as [`symmetric_interpolate_yᵃᶠᵃ`](@ref) but on `Center`s instead of `Face`s """
-@inline symmetric_interpolate_yᵃᶜᵃ(i, j, k, grid, scheme, ψ, args...) = inner_symmetric_interpolate_yᵃᶠᵃ(i, j+1, k, grid, scheme, ψ, j, Center, args...)
+@inline symmetric_interpolate_yᵃᶜᵃ(i, j, k, grid, scheme, ψ, args...) = symmetric_interpolate_yᵃᶠᵃ(i, j+1, k, grid, scheme, ψ, args...)
```
tomchor, line 8:
```suggestion
@inline symmetric_interpolate_yᵃᶜᵃ(i, j, k, grid, scheme, ψ, args...) = symmetric_interpolate_yᵃᶠᵃ(i, j + 1, k, grid, scheme, ψ, args...)
```

src/Advection/reconstruction_coefficients.jl
```diff
@@ -2,70 +2,19 @@ using Oceananigans.Grids: ξnodes, ηnodes, rnodes
...
-"""
-@inline biased_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme, bias, ψ, args...)  = inner_biased_interpolate_zᵃᵃᶠ(i, j, k, grid, scheme, bias, ψ, k, Face, args...)
+@inline symmetric_interpolate_zᵃᵃᶜ(i, j, k, grid, scheme, ψ, args...) = symmetric_interpolate_zᵃᵃᶠ(i, j, k+1, grid, scheme, ψ, args...)
 
 """ same as [`biased_interpolate_xᶠᵃᵃ`](@ref) but on `Center`s instead of `Face`s """
-@inline biased_interpolate_xᶜᵃᵃ(i, j, k, grid, scheme, bias, ψ, args...)  = inner_biased_interpolate_xᶠᵃᵃ(i+1, j, k, grid, scheme, bias, ψ, i, Center, args...)
+@inline biased_interpolate_xᶜᵃᵃ(i, j, k, grid, scheme, bias, ψ, args...)  = biased_interpolate_xᶠᵃᵃ(i+1, j, k, grid, scheme, bias, ψ, args...)
 """ same as [`biased_interpolate_yᵃᶠᵃ`](@ref) but on `Center`s instead of `Face`s """
-@inline biased_interpolate_yᵃᶜᵃ(i, j, k, grid, scheme, bias, ψ, args...)  = inner_biased_interpolate_yᵃᶠᵃ(i, j+1, k, grid, scheme, bias, ψ, j, Center, args...)
+@inline biased_interpolate_yᵃᶜᵃ(i, j, k, grid, scheme, bias, ψ, args...)  = biased_interpolate_yᵃᶠᵃ(i, j+1, k, grid, scheme, bias, ψ, args...)
```
tomchor, line 15:
```suggestion
@inline biased_interpolate_yᵃᶜᵃ(i, j, k, grid, scheme, bias, ψ, args...)  = biased_interpolate_yᵃᶠᵃ(i, j + 1, k, grid, scheme, bias, ψ, args...)
```

src/Advection/reconstruction_coefficients.jl
```diff
@@ -2,70 +2,19 @@ using Oceananigans.Grids: ξnodes, ηnodes, rnodes
...
 """ same as [`biased_interpolate_zᵃᵃᶠ`](@ref) but on `Center`s instead of `Face`s """
-@inline biased_interpolate_zᵃᵃᶜ(i, j, k, grid, scheme, bias, ψ, args...)  = inner_biased_interpolate_zᵃᵃᶠ(i, j, k+1, grid, scheme, bias, ψ, k, Center, args...)
+@inline biased_interpolate_zᵃᵃᶜ(i, j, k, grid, scheme, bias, ψ, args...)  = biased_interpolate_zᵃᵃᶠ(i, j, k+1, grid, scheme, bias, ψ, args...)
```
tomchor, line 17:
```suggestion
@inline biased_interpolate_zᵃᵃᶜ(i, j, k, grid, scheme, bias, ψ, args...)  = biased_interpolate_zᵃᵃᶠ(i, j, k + 1, grid, scheme, bias, ψ, args...)
```

PR #4414, opened 2025-04-18
Don't invoke comm buffers on non-distributed grids, move distributed functionality to DistributedComputations

glwagner:


src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -4,7 +4,7 @@ export
     BCType, Flux, Gradient, Value, Open,
     BoundaryCondition, getbc, setbc!,
     PeriodicBoundaryCondition, OpenBoundaryCondition, NoFluxBoundaryCondition, MultiRegionCommunicationBoundaryCondition, 
-    FluxBoundaryCondition, ValueBoundaryCondition, GradientBoundaryCondition, DistributedCommunicationBoundaryCondition,
+    FluxBoundaryCondition, ValueBoundaryCondition, GradientBoundaryCondition,
```
glwagner, line 7:
```suggestion
    FluxBoundaryCondition, ValueBoundaryCondition, GradientBoundaryCondition, DistributedCommunicationBoundaryCondition
```

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -4,7 +4,7 @@ export
     BCType, Flux, Gradient, Value, Open,
     BoundaryCondition, getbc, setbc!,
     PeriodicBoundaryCondition, OpenBoundaryCondition, NoFluxBoundaryCondition, MultiRegionCommunicationBoundaryCondition, 
-    FluxBoundaryCondition, ValueBoundaryCondition, GradientBoundaryCondition, DistributedCommunicationBoundaryCondition,
+    FluxBoundaryCondition, ValueBoundaryCondition, GradientBoundaryCondition, DistributedCommunicationBoundaryCondition
```
glwagner, line 7:
```suggestion
    FluxBoundaryCondition, ValueBoundaryCondition, GradientBoundaryCondition, DistributedCommunicationBoundaryCondition,
```

src/BoundaryConditions/boundary_condition.jl
```diff
@@ -81,25 +81,26 @@ const VBC  = BoundaryCondition{<:Value}
 const GBC  = BoundaryCondition{<:Gradient}
 const ZFBC = BoundaryCondition{Flux, Nothing} # "zero" flux
 const MCBC = BoundaryCondition{<:MultiRegionCommunication}
-const DCBC = BoundaryCondition{<:DistributedCommunication}
+const DCBC = BoundaryConditions(<:DistributedCommunication}
```
glwagner, line 84:
```suggestion
const DCBC = BoundaryCondition(<:DistributedCommunication}
```

src/BoundaryConditions/boundary_condition.jl
```diff
@@ -81,25 +81,26 @@ const VBC  = BoundaryCondition{<:Value}
 const GBC  = BoundaryCondition{<:Gradient}
 const ZFBC = BoundaryCondition{Flux, Nothing} # "zero" flux
 const MCBC = BoundaryCondition{<:MultiRegionCommunication}
-const DCBC = BoundaryCondition{<:DistributedCommunication}
+const DCBC = BoundaryCondition(<:DistributedCommunication}
 const ZBC  = BoundaryCondition{<:Zipper}
 
+const DistributedCommunicationBoundaryCondition = BoundaryConditions(<:DistributedCommunication}
```
glwagner, line 87:
```suggestion
const DistributedCommunicationBoundaryCondition = BoundaryCondition(<:DistributedCommunication}
```

src/Fields/field.jl
```diff
@@ -99,7 +99,7 @@ function Field(loc::Tuple, grid::AbstractGrid, data, bcs, indices, op=nothing, s
     @apply_regionally indices = validate_indices(indices, loc, grid)
     @apply_regionally validate_field_data(loc, data, grid, indices)
     @apply_regionally validate_boundary_conditions(loc, grid, bcs)
-    buffers = FieldBoundaryBuffers(grid, data, bcs)
+    buffers = nothing
```
simone-silvestri, line 102: maybe this can be a function that we extend in the distributed module if the goal eventually is to remove the special field constructor for distributed

glwagner: we can do that, but why do we need this if we extend Field?

glwagner: oh sorry I read what you wrote too quickly. Yes it would be easy to define `communication_buffer` which returns `nothing` or `CommunicationBuffer`. I think it'd be better to implement that when we do in fact eliminate the special field constructor, rather than anticipating it here.

glwagner:
@simone-silvestri I'm going to need your help to proceed here

simone-silvestri:
seems like we need `CommunicationBuffers` in the field constructor becuase multiregion grids use the general field constructor.

glwagner:
> seems like we need `CommunicationBuffers` in the field constructor becuase multiregion grids use the general field constructor.

for that, define `communication_buffers`, which can return `nothing` or an instances of `CommunicationBuffers`

simone-silvestri:
already done!

PR #4415, opened 2025-04-18
Clean up extra code in `test_turbulence_closures.jl`

tomchor:
Removes some code that seems to be left over from https://github.com/CliMA/Oceananigans.jl/pull/4380


glwagner (approved):

glwagner:
@tomchor I'd like to merge this, but I saw it was marked as a draft? Is there more work. you would like to do?

tomchor:
> @tomchor I'd like to merge this, but I saw it was marked as a draft? Is there more work. you would like to do?

It seemed like the instantiation of LagAveragedSmag wasn't being tested, so I also want to see if we can test that, not that things are faster. I just made this modification.

tomchor:
I'm finding something a bit odd. I see that the turbulence closure tests on GPU passed in 8 minutes!:

image

Which feels suspiciously fast. However, it does seem like everything ran:

image

Am I missing something here? How did this get so fast?

tomchor:
Another minor note is that is seems that CPU tests are running twice on CPU. Once using github actions:

image

and one on buildkite:

image

Is there any reason why we don't run all our CPU tests only on github actions? I don't see why we need to use computing time on our servers when github gives it to us for free.

glwagner:
It woudl be more effort to maintain the two streams, but it could be worth trying. I am curious whether it would be efficient...

tomchor:
> can you make sure that it's running the same tests?

I added a few more print statements for more granularity, but what was printed in this last run already indicates that everything is running the same in both GPU and CPU. The screen output is the same and I don't see anything missing.

tomchor:
Also for the record, the turbulence tests are also similarly fast on main. Here's what I realized 

- Before https://github.com/CliMA/Oceananigans.jl/pull/4380 was merged, turbulence tests on CPU and GPU were passing in ~50 and ~80 minutes respectively
- Immediately after https://github.com/CliMA/Oceananigans.jl/pull/4380 was merged the CPU turbulence tests passed in 21 minutes and the GPU turbulence tests passed in 44 minutes.
- This timing remained the same until https://github.com/CliMA/Oceananigans.jl/pull/4408 was merged. After which CPU turbulence tests passed in ~8 minutes on buildkite, ~15 minutes on github actions, and GPU turbulence tests passed in 10 minutes.

Looking at #4408, it looks like it only reorganized the pipeline, so I don't really see how that made tests faster, but apparently it did I guess. Could it be the switch from Julia 1.10.8 to 1.10.9?

cc @simone-silvestri

PR #4421, opened 2025-04-20
Fix adapt for TKEDissipationDiffusivityFields

wsmoses:
Hit this when doing gradient on your test case


glwagner (approved):

glwagner:
Note to self to test this on GPU

PR #4440, opened 2025-04-23
Compute the difference between the latest main and the PR

simone-silvestri:
This PR adds a method to compute the difference between main and the current PR in terms of profiles (at the moment just a *txt difference).


simone-silvestri:
This seems to work and produces the diff in the artifacts, see https://buildkite.com/clima/oceananigans-benchmarks-1/builds/105#01966836-8f61-486c-bbd9-fd7b269e943f.

We should merge this (so we can start comparing the other PRs) and continue working on the formatting of the outputs.

Issue #4397, opened 2025-04-14
NetCDFWriter scalar indexing error on GPU


xiaozhour:
Hi all,

I got a weird error on a simple output line with NetCDFWriter on GPU. Switching to JLD2Writer removes the error. Any insights would be appreciated!

```
u, v, w = model.velocities
b = model.tracers.b
outputs = (; u, v, w, b)

simulation.output_writers[:fields] = NetCDFWriter(model, outputs;
                                                   filename = "test_drag.nc",
                                                   schedule = TimeInterval(2minutes),
                                                   overwrite_existing = true)
```
The error massage is the following:
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:

glwagner:
@xiaozhour do you still get this error if you don't have the background field?

xiaozhour:
Yep @glwagner

glwagner:
@tomchor @ali-ramadhan

glwagner:
Probably same problem over here: https://github.com/CliMA/Oceananigans.jl/discussions/4433

@xiaozhour it might help if you confirm the Oceanangians version and julia version you're using.

xiaozhour:
Thanks, @glwagner. Oceananigans is v0.96.3 and Julia is 1.10.9.

xiaozhour:
@tomchor Updating to v0.96.21 didn't help

rafacmartins:
Hi all,
Just to highlight this here — I was using the latest version of Oceananigans (v0.96.21) and was getting the same error. Then I downgraded to a previous version v0.92.1 — kind of randomly, and it’s working just fine now.
Note that the function name reverted to `NetCDFOutputWriter` instead of the current `NetCDFWriter`.

glwagner:
It's likely the change from 0.95 to 0.96 that broke this case. If you can confirm it might help. I am somewhat puzzled --- isn't NetCDFWriter tested on GPU?

tomchor:
[...]

PR #4441, opened 2025-04-23
(0.96.22) Make NetCDFWriter work with GPU grids with stretched dimensions

tomchor:
Closes https://github.com/CliMA/Oceananigans.jl/issues/4397

This seems to fix it for me locally. We should also add a test to catch this in the future.


tomchor:
@xiaozhour can you try this branch and see if it solves your problem?

xiaozhour:
> @xiaozhour can you try this branch and see if it solves your problem?

Works well for me!

PR #4446, opened 2025-04-26
Adding reciprocal metric in the z-direction

simone-silvestri:
This PR starts the work of introducing the reciprocal of the metrics starting from the easiest metric (the z direction) which is common to all the grids. Let's see what the effect of this change is.

I am a bit worried thought that we are increasing significantly the parameter space that we are passing to the GPU, so in this quest of improving the performance we might shoot ourselves in the feet


glwagner:
> I am a bit worried thought that we are increasing significantly the parameter space that we are passing to the GPU, so in this quest of improving the performance we might shoot ourselves in the feet

That's a good point. I guess we have to assess the benefit.

> starting from the easiest metric (the z direction)

True that it's common to all grids, but its def not the easiest metric if you want to isolate the effect for just one grid...

glwagner:
I guess one solution is to eliminate the non-reciprocal metrics (eg compute those by taking reciprocal of reciprocal). They are only used in integrals (?) so maybe not as performance critical. 

But yeah its a big change.

src/Grids/vertical_discretization.jl
```diff
@@ -37,14 +37,15 @@ const AbstractStaticGrid  = AbstractUnderlyingGrid{<:Any, <:Any, <:Any, <:Any, <
...
-struct MutableVerticalDiscretization{C, D, E, F, H, CC, FC, CF, FF} <: AbstractVerticalCoordinate
+struct MutableVerticalDiscretization{C, D, E, F, H, U, CC, FC, CF, FF} <: AbstractVerticalCoordinate
       cᵃᵃᶠ :: C
       cᵃᵃᶜ :: D
       Δᵃᵃᶠ :: E
       Δᵃᵃᶜ :: F
     Δ⁻¹ᵃᵃᶠ :: E
     Δ⁻¹ᵃᵃᶜ :: F
         ηⁿ :: H
+        Gⁿ :: U # Storage space used in different ways by different timestepping schemes.
```
glwagner: This doesn't seem belong here. Is this something that belongs in ZStarCoordinate?

We want our grid utilities to be model agnostic so that they can be used in other contexts, eg by land models. This will be an issue very soon.

simone-silvestri: we can move this in the `ZStarCoordinate`. I can open a PR

PR #3146, opened 2023-06-14
Cut Cells

siddharthabishnu:
This PR closes #3123.

PR #4450, opened 2025-04-28
Create Cut-cell2d.py

joedimos:
This is an expanded 2d version of the cut-cell model wrt non-uniform grids. 

This takes coordinates (x_face_coords, z_face_coords), a bathymetry function (bathy_func) and various params as input.
A model grid (x and z coordinates of cell faces), calculating cell dimensions (dx, dz), bathymetry (ocean bottom depth) at grid points using bathy_func.

Crucially, it calculates Face, hFacW, and hFacS 9horizontal and vertical area fractions), which represents a portion of each cell that actually filled with water according to bathymetry.

### DIFF
--- a/examples/Cut-cell2d.py
+++ b/examples/Cut-cell2d.py
@@ -1,0 +1,797 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import matplotlib.patches as patches
+import warnings # To suppress divide-by-zero warnings safely
+
+class CutCellOceanModel2D_NonUniform_Expanded:
+    """
+    An expanded conceptual 2D (x-z slice) ocean model demonstrating cut cells (hFac)
+    on a non-uniform, structured grid. Includes T/S tracers, linear EOS,
+    hydrostatic pressure, vertical velocity from continuity, basic momentum tendency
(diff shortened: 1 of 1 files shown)


glwagner:
The code in this PR isn't mergeable --- it doesn't contribute to Oceananigans. To implement cut cells in Oceananigans, one needs to add a new immersed boundary type. This PR attempts such a thing, but is a WIP: https://github.com/CliMA/Oceananigans.jl/pull/3146 .

navidcy:
@joedimos thanks for this.

Unfortunately it's as @glwagner pointed out above this is not mergeable. It's not even in the language of the package. But even if this was in Julia it doesn't read as if it builds on Oceananigans.

I'm suggesting we close this PR but feel free to continue the discussion in https://github.com/CliMA/Oceananigans.jl/pull/3146.

One thing of concern with partial cells is that sometime we end up with cells  that are  almost filled except a tiny little bit resulting in minuscule cells (which, unfortunately, need minuscule timestep to avoid CFL criterion). Does the algorithm you propose above deals with that in a robust way? If so, explaining how so would be very useful.

glwagner:
Agree, let's continue in #3146

PR #4452, opened 2025-04-29
Time-step TKE using an active cells map

simone-silvestri:
this should improve the performance of the TKE substep in the immersed benchmark testcase


glwagner:
this won't be needed after auto active map though right

PR #4453, opened 2025-04-29
Use a default active cells map for kernel launching

glwagner:
Curious about the performance implications of this.

I think it is better to do the same thing re: active cells for every kernel, rather than picking and choosing.

The main advantage is the simplicity of the algorithm.

Ultimately I think that the active cells should be part of the KernelParameters. It doesn't make sense to think of the kernel parameters and active map as separate concepts.

It should be automatically computed for the majority of kernels, and only changed for edge cases. I am not sure why there are edge cases, but it seems they exist for halo filling?

cc @simone-silvestri can you help improve this design? Perhaps we can discuss here.

Ironing out these issues will become crucial if we want to use Oceananigans finite volume engine for implementing additional models. I also think a distributed algo refactor needs to come at some point, and this is perhaps the first step.

### DIFF
--- a/src/Grids/abstract_grid.jl
+++ b/src/Grids/abstract_grid.jl
@@ -1,3 +1,5 @@
+import Oceananigans.Utils: get_active_cells_map
+
 """
     AbstractGrid{FT, TX, TY, TZ}
 
--- a/src/Utils/kernel_launching.jl
+++ b/src/Utils/kernel_launching.jl
@@ -16,6 +16,8 @@ import Base
 
 struct KernelParameters{S, O} end
 
+get_active_cells_map(grid, any_map_type) = nothing
+
 """
     KernelParameters(size, offsets)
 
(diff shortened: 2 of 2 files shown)


simone-silvestri:
Make sense to have it as a default. I think the default should depend on the `workspec`. If the `workspec` is a default workspec (for example `:xyz` or `:xy`) we can point to the correct active map (for example `Val(:interior)` or `Val(:surface)`, respectively -- maybe we can change the naming --  or `nothing` in case of `:xz` and `:yz`). Also, if the `workspec` is a `KernelParameters`, then I would have the default be `nothing` because it means that we are trying to have ``non-default'' behavior. 

Indeed, following these changes, the distributed algorithm needs to change because at the moment is harcoded with maps for interior, east west and south.

glwagner:
> Indeed, following these changes, the distributed algorithm needs to change because at the moment is harcoded with maps for interior, east, west, north, and south which would not work with this paradigm.

But this is just the default. Doesn't that mean existing code can stay?

glwagner:
@simone-silvestri should we close this?

simone-silvestri:
yeah, we can revisit this in the next round of performance optimizations.

PR #4459, opened 2025-04-30
Fix computation of bottom height for GridFittedBottom

glwagner:
This PR fixes how we compute the bottom height for GridFittedBottom.

I have to confess that I did not completely understand what the code was trying to do prior to this PR. Our objective is to change the bottom height so that it exactly corresponds to what the model "sees". This is useful for computing statistics of the bathymetry, etc.

This might change results --- so we may need to bump the minor version.

@NoraLoose may have an MWE where the previous code was causing an issue (and hopefully this fixes it).


glwagner:
Also `InterfaceImmersedCondition` will not work; there is no appropriate `immersed_cell` method and there are no tests. We can consider removing this property. It might be useful to have though, the original purpose was to produce an abstraction that allows us to exactly test PartialCellBottom. Nobody has taken up the mantle of maturing PartialCellBottom yet, so it's not used much...

glwagner:
Okay also on further inspection I think the code _was_ doing the right thing, so this PR is unecessary.

Issue #4416, opened 2025-04-18
`NetCDFWriter`  doesn't work for `PartialCellBottom` grids


tomchor:
In particular we get this error:

```
ERROR: MethodError: no method matching gather_immersed_boundary(::ImmersedBoundaryGrid{…}, ::Tuple{…}, ::typeof(OceananigansNCDatasetsExt.trilocation_dim_name))

Closest candidates are:
  gather_immersed_boundary(::ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:GridFittedBoundary}, ::Any, ::Any)
   @ OceananigansNCDatasetsExt /glade/work/tomasc/.julia/packages/Oceananigans/L7qIm/ext/OceananigansNCDatasetsExt.jl:380
  gather_immersed_boundary(::ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:GridFittedBottom}, ::Any, ::Any)
   @ OceananigansNCDatasetsExt /glade/work/tomasc/.julia/packages/Oceananigans/L7qIm/ext/OceananigansNCDatasetsExt.jl:364
```

glwagner:
What does this function do?

tomchor:
It looks like it gathers the masks for the immersed grids: https://github.com/CliMA/Oceananigans.jl/blob/e45450166f808304eba572c217a3aea9252c0e9c/ext/OceananigansNCDatasetsExt.jl#L363-L391

So maybe a better name would be `gather_immersed_boundary_masks`? Or `gather_immersed_boundary_metrics`?

I'm not sure what to return for partial cell bottoms though. I don't see anything there that would return something that's between 0 and 1 if the numerical bottom height doesn't exactly align with the cells.

cc @ali-ramadhan

glwagner:
masks = peripheral cells? I recommend using the same nomenclature as the source code. 

The concept of peripheral cells cannot be conflated with partial cells. They are separate concepts. The peripheral cells touch or are within the boundary/immersed boundary (note there is a difference between peripheral and inactive nodes). This concept applies to all immersed boundaries. Additionally, partial cells change the vertical spacing in cells next to the immersed boundary. The change in dz cannot be expressed by booleans.

tomchor:
[...]

PR #4462, opened 2025-05-01
Make `NetCDFWriter` work with `PartialCellBottom` grids

tomchor:
Closes https://github.com/CliMA/Oceananigans.jl/issues/4416

I also changed the metric names from `immersed_boundary_mask` to `peripheral_nodes`, which is what we use in the code.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.96.22"
+version = "0.96.23"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/ext/OceananigansNCDatasetsExt.jl
+++ b/ext/OceananigansNCDatasetsExt.jl
@@ -14,7 +14,7 @@ using Oceananigans.Grids: topology, halo_size, xspacings, yspacings, zspacings, 
 using Oceananigans.Fields: reduced_dimensions, reduced_location, location
 using Oceananigans.AbstractOperations: KernelFunctionOperation
 using Oceananigans.Models: ShallowWaterModel, LagrangianParticles
-using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBottom, GFBIBG, GridFittedBoundary
+using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBottom, GFBIBG, GridFittedBoundary, PartialCellBottom, PCBIBG
 using Oceananigans.TimeSteppers: float_or_date_time
 using Oceananigans.BuoyancyFormulations: BuoyancyForce, BuoyancyTracer, SeawaterBuoyancy, LinearEquationOfState
 using Oceananigans.Utils: TimeInterval, IterationInterval, WallTimeInterval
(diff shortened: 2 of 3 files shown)


glwagner (approved):
Looks good. I wonder if using loops / more programmatic style might be better though. Are we building the four locations because we anticipate the user might need those? Would it be possible to instead build the locations that correspond to outputs? That might result in less code + smaller files (for some cases).

tomchor:
> Looks good. I wonder if using loops / more programmatic style might be better though. Are we building the four locations because we anticipate the user might need those? Would it be possible to instead build the locations that correspond to outputs? That might result in less code + smaller files (for some cases).

My opinion is that we should build the four locations regardless. More than once I've been in situations where I did some post-processing that involved a change of locations for a given variable where a mask in a different location would have been helpful. Plus, compared to the typical size of a NetCDF file (which has many snapshots), a static single field is usually negligible.

@ali-ramadhan thoughts?

tomchor:
Also I wonder if I should make this a breaking release or not. Technically, post processing code using the previous names can break after this PR. But chances are good that this will affect at this point given how new the current version of `NetCDFWriter` is.

glwagner:
> > Looks good. I wonder if using loops / more programmatic style might be better though. Are we building the four locations because we anticipate the user might need those? Would it be possible to instead build the locations that correspond to outputs? That might result in less code + smaller files (for some cases).
> 
> My opinion is that we should build the four locations regardless. More than once I've been in situations where I did some post-processing that involved a change of locations for a given variable where a mask in a different location would have been helpful. Plus, compared to the typical size of a NetCDF file (which has many snapshots), a static single field is usually negligible.
> 
> @ali-ramadhan thoughts?

There are more than four locations. Why did you pick these four?

tomchor:
> There are more than four locations. Why did you pick these four? Notably it is missing the location of vertical vorticity.
> 
> Also, note that all locations can be constructed from the `ccc` value. This is how it works internally.

I didn't pick them, but I assumed these were picked because they're the most common. I'm not necessarily suggesting these particular ones. I'm okay outputting all locations. It's small a small added size to most NetCDFs.

glwagner:
They are 3D fields so not that small, right?

glwagner:
I mean, more generally we actually only need the bottom height to construct the peripheral nodes...

tomchor:
> They are 3D fields so not that small, right?

There are eight static fields, all 3D, one for each location. They are small in the sense that NetCDF files will contain multiple snapshots. So that's equivalent, for example, to two 3D snapshots of `u, v, w, b`. I consider that small, but I guess that's debatable since often people will output slices or directional averages.

> I mean, more generally we actually only need the bottom height to construct the peripheral nodes...

Yes, but the code for `peripheral_node` isn't trivial. While a user using Julia can use it from Oceananigans, users opening NetCDF files using other languages (which is the main point for a NetCDFWriter) will needs some work going from `bottom_height` to `peripheral_nodes`.

glwagner:
Is there scope for a simple python package for analyzing Oceananigans datasets? We can call julia from python, right?

tomchor:
> I'm worried that saving too much data into NetCDF files will eventually block activities that involve very large simulations and generally produce a maintenance burden (NetCDFWriter already need much more love). Up to now we have used JLD2Writer for all of those, but we want to support NetCDF for such cases too.

I agree with the maintenance burden point. Definitely something to think about. For the data size issue, I think we can make all these outputs optional. That's the case with non-immersed grid metrics already.

> Is there scope for a simple python package for analyzing Oceananigans datasets? We can call julia from python, right?

I think so. In fact I have the beginnings of a package called [pynanigans](https://github.com/tomchor/pynanigans) that I've been using for personal projects. It's pretty abandoned at the the moment because I've focused my public development on the Julia ecosystem, but in the past I've discussed with @ali-ramadhan the possibility of interfacing with Oceananigans directly using [JuliaCall](https://juliapy.github.io/PythonCall.jl/stable/juliacall/).

ali-ramadhan (approved):
Looks good to me! Solves the issue and also sticks to the philosophy of naming things in NetCDF similarly to how they're named in Oceananigans to reduce confusion.

I know there's a discussion around which peripheral node masks to save but I would advocate for having this conversation in a new issue so as to keep this PR small and self-contained.

ali-ramadhan:
> Are we building the four locations because we anticipate the user might need those? Would it be possible to instead build the locations that correspond to outputs?

Ah so I built those four (ccc, fcc, cfc, ccf) because I did anticipate them being the most commonly used four (at least for me).

We could add a keyword argument to `NetCDFWriter` called `include_peripheral_nodes_masks`. When `false` they are not included at all to save space. When `true` we can either include all 8 of them or include masks for each location that is used in the outputs. Or we could give the user even more options: `:include_all` for all 8 and `:include_present` to include just what is used. Names could be better lol.

The motivation behind including them is to make it easier to work with Oceananigans output, even if you're working in Julia. When analyzing simulation output, it's nice to just have everything ready for you. There's definitely a storage cost associated with including multiple masks, although I think it should be up to the user to decide the tradeoff they are happy with.

glwagner:
That makes sense @ali-ramadhan. I'd just say that for Julia usage, we really want to encourage / facilitate a situation that creates as much code re-use as possible. 

That's the underlying philosophy of FieldTimeSeries --- the same code that runs simulations is used to analyze them, which overall will increase code quality and features (features, performance improvements, bugfixes then benefit as many people as possible, and the overall magnitude of code written is less, which generally should increase its quality).

I do feel the importance of code reuse cannot be overemphasized in this Earth system modeling situation. Oceananigans is _already_ some 40k lines! And it's still growing. It's a persistent problem with legacy code bases that there is too much code / too much re-implementation and too little re-use.

tomchor:
> I know there's a discussion around which peripheral node masks to save but I would advocate for having this conversation in a new issue so as to keep this PR small and self-contained.

Sounds good! I'm gonna merge this PR for now and tag a new version.

PR #4467, opened 2025-05-02
[WIP] Fix tracer conservation when wind is blowing

### LINKED ISSUE #4466: Passive tracers are not conserved if wind is blowing
I noticed that passive tracers are not conserved in my global simulation (with a tripolar grid and the z-star coordinate) when a prescribed atmosphere is added (i.e., with `JRA55` and `ClimaOcean`). I’ve been able to isolate the issue to an Oceananigans-only MWE, adapted from the tripolar grid tracer conservation test in our suite. The only change is the addition of wind forcing.

Below is the simplified test setup used to reproduce the issue:

```julia
function run_tripolar_test_simulation(arch; with_wind = false)

    z_stretched = MutableVerticalDiscretization(collect(-20:0))
    grid = TripolarGrid(arch; size = (20, 20, 20), z = z_stretched)

    function mtn₁(λ, φ)
        λ₁ = 70
        φ₁ = 55
        dφ = 5
        return exp(-((λ - λ₁)^2 + (φ - φ₁)^2) / 2dφ^2)
    end

    function mtn₂(λ, φ)
        λ₁ = 70
        λ₂ = λ₁ + 180
        φ₂ = 55
        dφ = 5
        return exp(-((λ - λ₂)^2 + (φ - φ₂)^2) / 2dφ^2)
    end

    zb = - 20
    h  = - zb + 10
    gaussian_islands(λ, φ) = zb + h * (mtn₁(λ, φ) + mtn₂(λ, φ))

    grid = ImmersedBoundaryGrid(grid, GridFittedBottom(gaussian_islands))
    free_surface = SplitExplicitFreeSurface(grid; substeps=10)
    
    kwargs = (;
        grid,
        free_surface,
        tracers = (:b, :c),
        buoyancy = BuoyancyTracer(),
        vertical_coordinate = ZStar()
    )

    if with_wind
        u₁₀, cᴰ, ρₒ, ρₐ = 10.0, 2.5e-3, 1026.0, 1.225
        τx = -ρₐ / ρₒ * cᴰ * u₁₀ * abs(u₁₀)
        u_bcs = FieldBoundaryConditions(top = FluxBoundaryCondition(τx))
        model = HydrostaticFreeSurfaceModel(; kwargs..., boundary_conditions = (u = u_bcs,))    
    else
        model = HydrostaticFreeSurfaceModel(; kwargs...)
    end
    
    bᵢ(x, y, z) = y < 0 ? 0.06 : 0.01

    # Instead of initializing with random velocities, infer them from a random initial streamfunction
    # to ensure the velocity field is divergence-free at initialization.
    ψ = Field{Center, Center, Center}(grid)
    set!(ψ, rand(size(ψ)...))
    uᵢ = ∂y(ψ)
    vᵢ = -∂x(ψ)

    set!(model, c = (x, y, z) -> rand(), u = uᵢ, v = vᵢ, b = bᵢ)

    simulation = Simulation(model, Δt=2minutes, stop_time=1day)
    
    c = simulation.model.tracers.c
    tracer_integral = Integral(c, dims=(1, 2, 3))

    if with_wind
        prefix = "test_no_wind"
    else
        prefix = "test_with_wind"
    end
    
    integral_ow = JLD2Writer(
        simulation.model,
        (; tracer_integral),
        filename = "$(prefix)_integrals",
        schedule = IterationInterval(1),
        overwrite_existing = true,
    )
    simulation.output_writers[:integral] = integral_ow

    run!(simulation)
    
    return nothing
end

run_tripolar_test_simulation(GPU(); with_wind = false)
run_tripolar_test_simulation(GPU(); with_wind = true)
```

Visualizing the results shows that the passive tracer `c` is 
* conserved for the simulation without wind, which matches the behavior tested in our CI.
* not conserved for the simulation with wind, despite all other settings being identical.

Image

NoraLoose:
This PR aims to close #4466.

I am adding a test that checks  for tracer conservation on a tripolar grid with wind forcing, using the code snippet provided in #4466. Currently, the test fails due to non-conservation of the passive tracer. Once the underlying issue is resolved, this test should pass.

Feel free to push directly to this PR with any proposed fixes of the underlying issue.

### DIFF
--- a/test/test_zstar_coordinate.jl
+++ b/test/test_zstar_coordinate.jl
@@ -143,8 +143,10 @@ end
     @test all(um.data .≈ us.data)
 end
 
+# vertical z-levels for tracer conservation tests
+z_stretched = MutableVerticalDiscretization(collect(-20:0))
+
 @testset "ZStar tracer conservation testset" begin
-    z_stretched = MutableVerticalDiscretization(collect(-20:0))
     topologies  = ((Periodic, Periodic, Bounded),
                    (Periodic, Bounded, Bounded),
                    (Bounded, Periodic, Bounded),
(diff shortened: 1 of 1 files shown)


test/test_zstar_coordinate.jl
```diff
@@ -234,55 +236,87 @@ end
...
+        model = HydrostaticFreeSurfaceModel(; kwargs..., boundary_conditions = (u = u_bcs,))
+    else
+        model = HydrostaticFreeSurfaceModel(; kwargs...)
+    end
+
+    bᵢ(x, y, z) = y < 0 ? 0.06 : 0.01
+
+    # Instead of initializing with random velocities, infer them from a random initial streamfunction
+    # to ensure the velocity field is divergence-free at initialization.
+    ψ = Field{Center, Center, Center}(grid)
```
glwagner, line 293:
```suggestion
    ψ = Field{Face, Face, Center}(grid)
```

this is the right location for the streamfunction --- then when differentiated we get velocities on the c-grid

glwagner:
Is the problem specific to TripolarGrid, or does it also occur on rectilinear or lat-lon?

test/test_zstar_coordinate.jl
```diff
@@ -234,55 +236,87 @@ end
...
+        free_surface,
+        tracers = (:b, :c),
+        buoyancy = BuoyancyTracer(),
+        vertical_coordinate = ZStar()
+    )
+
+    if with_wind
+        u₁₀, cᴰ, ρₒ, ρₐ = 10.0, 2.5e-3, 1026.0, 1.225
+        τx = -ρₐ / ρₒ * cᴰ * u₁₀ * abs(u₁₀)
+        u_bcs = FieldBoundaryConditions(top = FluxBoundaryCondition(τx))
```
glwagner, line 283:
```suggestion
        u_bcs = FieldBoundaryConditions(top = FluxBoundaryCondition(-1e-4))
```

does this simplify?

NoraLoose:
> Is the problem specific to TripolarGrid, or does it also occur on rectilinear or lat-lon?

Good question - I haven't tested this for rectilinear and lat-lon.

test/test_zstar_coordinate.jl
```diff
@@ -234,55 +236,87 @@ end
...
+    return model
+end
+
+@testset "TripolarGrid ZStar tracer conservation tests" begin
+    for arch in archs
+    	@info "Testing ZStar coordinate with TripolarGrid on $(arch)..."
+
+    	model = make_tripolar_test_model(arch, with_wind=false)
+    	Δt = 2minutes
+    	test_zstar_coordinate(model, 300, Δt)
```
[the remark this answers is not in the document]
NoraLoose: Could be done with a slight refactor. Question is: Do we want to double the number of tests? We are testing quite a number of rectilinear grid cases (different topologies, immersed boundary vs. not, etc.)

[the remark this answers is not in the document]
NoraLoose: It's probably worth doing the refactor and test with wind forcing for any grid. This would also answer the question whether the issue is specific to the TripolarGrid. I will continue working on this in a few hours.

glwagner:
> > Is the problem specific to TripolarGrid, or does it also occur on rectilinear or lat-lon?
> 
> Good question - I haven't tested this for rectilinear and lat-lon.

could be a red herring, but if there is a difference between grids (also when using immersed boundaries or not) that might be an important clue

test/test_zstar_coordinate.jl
```diff
@@ -300,21 +302,16 @@ end
...
-
-@testset "ZStar tracer conservation tests with wind stress" begin
-    for arch in archs
-        @info "Testing ZStar coordinate with TripolarGrid and wind stress on $(arch)..."
-    
-        model = make_tripolar_test_model(arch; with_wind=true)
-        Δt = 2minutes
-        test_zstar_coordinate(model, 300, Δt)
+	for with_wind in (false, true)
+            for with_bathymetry in (false, true)
```
glwagner, line 306: I don't think we can use the TripolarGrid without bathymetry (a downside of Tripolar) because of the coordinate singularities

NoraLoose:
In an attempt to get some more clues, [765be75](https://github.com/CliMA/Oceananigans.jl/pull/4467/commits/765be7559d93e5f0583042c230e5a9d322d54a60) sets up additional tracer conservation tests but without bathymetry (with and without wind).

However, tracer conservation fails at the initial time step for the TripolarGrid without bathymetry (both with and without wind). So this adds more questions than answers...

test/test_zstar_coordinate.jl
NoraLoose: oh okay, that explains 

> However, tracer conservation fails at the initial time step for the TripolarGrid without bathymetry (both with and without wind). So this adds more questions than answers...

NoraLoose: okay, I will roll back

PR #4472, opened 2025-05-04
ReactantExt: Fix conditional operator ordering

wsmoses:
@glwagner this fixes the minimum issue we saw on enzymanigans


src/Operators/spacings_and_areas_and_volumes.jl
```diff
@@ -195,6 +200,14 @@ end
 @inline Δxᶠᶠᵃ(i, j, k, grid::LLGX) = @inbounds grid.Δxᶠᶠᵃ[j]
 @inline Δxᶜᶜᵃ(i, j, k, grid::LLGX) = @inbounds grid.Δxᶜᶜᵃ[j]
 
+for sym in (:Δxᶠᶜᵃ, :Δxᶜᶠᵃ, :Δxᶠᶠᵃ, :Δxᶜᶜᵃ)
+    @eval @inline function $sym(i::AbstractArray, j::AbstractArray, k::AbstractArray, grid::LLGX)
+	  x = $sym(1, j, 1, grid)
+	  Base.Broadcast.materialize(Base.Broadcast.Broadcasted(Base.Broadcast.BroadcastStyle(typeof(x)), Base.identity, (Base.Broadcast.Extruded(x, (false, true,false), (1,1,1)),), (Base.axes(i, 1), Base.axes(j, 2), Base.axes(k, 3))))
```
glwagner, line 206: can this be rewritten to be a little easier to read?

wsmoses: so originally I had this guy with a stack like the others, but it turns out that's horribly slow [as julia will allocate a new var for each i/j/k], whereas this braodcast makes a new array all at once.

open to suggestions though ofc

PR #4479, opened 2025-05-06
Interpret `u, v` within `set!` on OrthogonalSphericalShellGrid as an extrinsic vector by default

glwagner:
This PR changes the default behavior of `set!` so that user input for `(u, v)` is interpreted as the "extrinsic vector". For example, this means that when using `TripolarGrid`, the velocity input is interpreted as giving the zonal and meridional velocity components. Prior to this PR, no interpretation was performed, so that `u, v` in `set!` had to be provided in grid-intrinsic coordinates. Effectively this means that the only way to set a velocity field was to form a streamfunction and differentiate it.

Now, the user has the option to provide the zonal and meridional velocity components for `set!` on `TripolarGrid`. On the other hand, if using a streamfunction, the user can also write

```julia
ui = + ∂y(ψ)
vi = - ∂x(ψ)
set!(model, u=ui, v=vi, intrinsic_velocities=true)
```

A subsequent PR will be needed (which will be a bit more involved) to also interpret the boundary conditions on u, v in a similar manner. Until then, idealized boundary conditions that explicitly specify the zonal and meridional momentum flux components have to also form a "flux streamfunction" and differentiate it to obtain the intrinsic u, v components.

Note that ClimaOcean handles the rotation of the momentum flux vector under the hood when coupling to an atmosphere, so the boundary condition issue is a concern only for manually-supplied boundary conditions on TripolarGrid and friends.

cc @NoraLoose @StevePny


src/Models/HydrostaticFreeSurfaceModels/set_hydrostatic_free_surface_model.jl
```diff
@@ -66,3 +72,61 @@ model.velocities.u
...
+    return nothing
+end
+
+@kernel function _rotate_velocities!(u, v, grid)
+    i, j, k = @index(Global, NTuple)
+    # Rotate u, v from extrinsic to intrinsic coordinate system
+    ur, vr = intrinsic_vector(i, j, k, grid, u, v)
+    @inbounds begin
+        u[i, j, k] = ur[i, j, k]
+        v[i, j, k] = vr[i, j, k]
```
glwagner, line 122:
```suggestion
        u[i, j, k] = ur
        v[i, j, k] = vr
```

src/Models/HydrostaticFreeSurfaceModels/set_hydrostatic_free_surface_model.jl
```diff
@@ -66,3 +72,61 @@ model.velocities.u
...
+    i, j, k = @index(Global, NTuple)
+    # Rotate u, v from extrinsic to intrinsic coordinate system
+    ur, vr = intrinsic_vector(i, j, k, grid, u, v)
+    @inbounds begin
+        u[i, j, k] = ur
+        v[i, j, k] = vr
+    end
+end
+
+@kernel function _interpolate_velocities(u, v, grid, uᶜᶜᶜ, vᶜᶜᶜ)
```
glwagner, line 126:
```suggestion
@kernel function _interpolate_velocities!(u, v, grid, uᶜᶜᶜ, vᶜᶜᶜ)
```

glwagner:
Note to self: for boundary conditions we discussed implementing a new type `UVFluxBoundaryCondition(tx, ty)` which will contain both flux components in meridional and zonal directions.

src/Models/HydrostaticFreeSurfaceModels/set_hydrostatic_free_surface_model.jl
```diff
@@ -66,3 +72,61 @@ model.velocities.u
...
+
+This abstraction is necessary for `OrthogonalSphericalShellGrid` and derivatives thereof,
+where the extrinsic and intrinsic coordinate systems differ.
+"""
+function set_velocities!(model, u, v; intrinsic_velocities=false)
+    if intrinsic_velocities || !(model.grid isa IntrinsicCoordinateGrid)
+        u isa ZeroField || set!(model.velocities.u, u)
+        v isa ZeroField || set!(model.velocities.v, v)
+    else
+        set_from_extrinsic_velocties!(model.velocities, model.grid, u, v)
```
navidcy, line 98:
```suggestion
        set_from_extrinsic_velocities!(model.velocities, model.grid, u, v)
```

PR #4484, opened 2025-05-07
Add extrema(c::AbstractField)

Sbozzolo:
Currently, `extrema(field)` fails with scalar indexing on CUDA. The reason for this is that `extrema(c::AbstractField)` is not defined, so we fall back to `extrema(a::AbstractArray; dims, kw...)`.

This commit adds `extrema(field)` by calling `minimum(field)` and `maximum(field)` separately. Note that there's an opportunity for optimization here, where both the extrama are found in the same reduction.

### DIFF
--- a/src/Fields/field.jl
+++ b/src/Fields/field.jl
@@ -722,6 +722,10 @@ for reduction in (:sum, :maximum, :minimum, :all, :any, :prod)
     end
 end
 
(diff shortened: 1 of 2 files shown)


glwagner:
this is a nice addition, we were talking about the need for this just a few hours ago with @simone-silvestri

navidcy (approved):

PR #4487, opened 2025-05-08
Fix velocity exchanges across tripolar seam

NoraLoose:
As @simone-silvestri noted in [this comment](https://github.com/CliMA/Oceananigans.jl/issues/4466#issuecomment-2862542898), the streamfunction used to initialize the velocity field in our tracer conservation test on the tripolar grid had an incorrect magnitude, resulting in initial velocities on the order of 10⁻⁶ m/s. This PR corrects the streamfunction so that initial velocities are O(1) m/s.

With non-negligible initial velocities, tracer conservation breaks down, see failing test " TripolarGrid ZStar tracer conservation tests".


NoraLoose:
Here is a visualization for how tracer conservation is broken in the modified test.
Screenshot 2025-05-08 at 10 45 00 AM

test/test_zstar_coordinate.jl
```diff
@@ -278,7 +278,7 @@ end
             # Instead of initializing with random velocities, infer them from a random initial streamfunction
             # to ensure the velocity field is divergence-free at initialization.
             ψ = Field{Center, Center, Center}(grid)
-            set!(ψ, rand(size(ψ)...))
+            set!(ψ, 10^6 * rand(size(ψ)...))
```
navidcy, line 281: perhaps add a comment next to it? why 1e6? (is it related to the typical area of grid cells?)

navidcy (approved):

test/test_zstar_coordinate.jl
NoraLoose: I added a comment in [597ccee](https://github.com/CliMA/Oceananigans.jl/pull/4487/commits/597ccee6d00cc964ff20795bd4ccf3b24c1f8341)

NoraLoose:
@navidcy Just a heads-up — we probably don’t want to merge this PR yet. The TripolarGrid ZStar tracer conservation tests are currently failing, because this PR modifies the test in a way so that it exposes a deeper underlying issue. We're tracking that underlying issue in https://github.com/CliMA/Oceananigans.jl/issues/4488. Probably best to resolve the underlying issue first before proceeding with this merge.

glwagner:
We can merge if we get the tests to pass though right??

NoraLoose:
> We can merge if we get the tests to pass though right??

Yes, I just put that label there because @navidcy already approved the PR! Feel free to remove it.

glwagner:
For us "approve" means (good to merge IF tests pass) but we are maybe a little loose sometimes 😂

test/test_zstar_coordinate.jl
```diff
@@ -277,8 +277,12 @@ end
...
             # Instead of initializing with random velocities, infer them from a random initial streamfunction
             # to ensure the velocity field is divergence-free at initialization.
-            ψ = Field{Center, Center, Center}(grid)
-            set!(ψ, rand(size(ψ)...))
+            ψ = Field{Face, Face, Center}(grid)
+	    # Set streamfunction amplitude ~10⁶ to yield non-negligible initial velocities.
+	    # Grid spacing in the test grid ranges from 10⁵ to 10⁶ meters,
+            # so taking spatial derivatives (∂ψ/∂x or ∂ψ/∂y) divides by grid spacing,
+            # resulting in velocity magnitudes on the order of O(1) m/s.
+            set!(ψ, 10^6 * rand(size(ψ)...))
```
navidcy, line 285:
```suggestion

	    mean_xspacing = mean(xspacings(grid, Face(), Face(), Center()))
            mean_yspacing = mean(yspacings(grid, Face(), Face(), Center()))
            Δ = mean((mean_xspacing, mean_yspacing))
	    U = 1

	    # Set streamfunction amplitude to Δ * U to get initial velocities of order U.
            set!(ψ, U * Δ * rand(size(ψ)...))
```

navidcy:
Indeed, my approval was on the basis that tests pass.
I added a small suggestion but feel free to ignore if; up to you.

src/BoundaryConditions/fill_halo_regions_zipper.jl
```diff
@@ -93,6 +93,8 @@ end
         end
     end
 
+    @inbounds v[i, Ny, k] = ifelse(i > Nx ÷ 2, sign * v[i′, Ny + 1, k], v[i, Ny, k])
```
simone-silvestri, line 96:
```suggestion
```

src/BoundaryConditions/fill_halo_regions_zipper.jl
```diff
@@ -93,7 +93,6 @@ end
         end
     end
 
-    @inbounds v[i, Ny, k] = ifelse(i > Nx ÷ 2, sign * v[i′, Ny + 1, k], v[i, Ny, k])
```
NoraLoose: @simone-silvestri I think we need this line. Did you remove it on purpose?

simone-silvestri: Yes, why do we need this line? The v velocity is not duplicated across the seam ad the tracer and u-velocity is

glwagner: Not sure I understand. @NoraLoose's sketch showed that `v` is duplicated. What is the additional information that negates it?

NoraLoose: Oh "evolved", not "involved"

NoraLoose: >  So therefore we have the opposite rule for v than for other fields, and we should be filling in the reverse direction (eg putting interior values from the right of the seam onto the left boundary).

I think we are always taking values from the ride side, and copying them over to the left side (possibly with a sign change).

glwagner: Basically, we compute tendencies only up to Ny, so the Ny+1 must be determined by filling halos

Issue #4490, opened 2025-05-08
`RobinBoundaryCondition` to complement `Gradient` and `Value` boundary conditions


glwagner:
We need to implement `RobinBoundaryCondition`, which is a boundary condition of the form

$$ \partial_n c = a c + b $$

for two inputs $a$ and $b$. On a staggered grid and a "right boundary" (eg $$i=N+1$$) this condition looks like

$$ \frac{c_{N+1} - c_N}{\Delta} = a \frac{c_{N+1} + c_N}{2} + b $$

which implies a rule for the halo point,

$$ c_{N+1} = \frac{c_N \left ( 1 + \frac{a \Delta}{2} \right ) + b \Delta}{1 + \frac{a \Delta}{2}} $$

@joernc, @shriyafruitwala

Right now we have two boundary conditions which invoke rules for filling halos: `Gradient` and `Value`. The way it's currently implemented, these boundary conditions imply rules for determining the across-boundary gradient, eg:

https://github.com/CliMA/Oceananigans.jl/blob/2dc85f08b87b5443bdc0a818d16d945870f2aa60/src/BoundaryConditions/fill_halo_regions_value_gradient.jl#L9

for `GradientBoundaryCondition`, where `getbc` returns `g` for user input `GradientBoundaryCondition(g)`. Another example is `ValueBoundaryCondition(v)`,

https://github.com/CliMA/Oceananigans.jl/blob/2dc85f08b87b5443bdc0a818d16d945870f2aa60/src/BoundaryConditions/fill_halo_regions_value_gradient.jl#L12

This implementation has the effect that reconstruction/interpolation _across_ the boundary returns `v`.

We can rewrite the algebra for RobinBoundaryCondition to fit this pattern, which (I think) yields

$$ \frac{c_{N+1} - c_N}{\Delta}  = \frac{a c_N + b}{1 - \frac{a \Delta}{2}} $$

So we can either refactor how boundary conditions are implemented (first formula) or re-use the infrastructure by just adding a line (above formula).

If we want extra credit, we can allow the `getbc` infrastructure to work for the two Robin values $a$ and $b$. I think for that we actually want to refactor `getbc` first --- it's weird that `getbc` takes in the `BoundaryCondition`, when it looks like a simpler implementation would invoke

```julia
getcondition(bc.condition, i, j, grid, args...)
```

so the first step might be to do this refactor, then implement `RobinBoundaryCondition`.

PR #4491, opened 2025-05-08
Refactor `getbc` interface

glwagner:
This PR refactors the `getbc` interface so that developers can implement methods to extend `bc.condition` directly, rather than having to extend `BoundaryCondition{<:Any, MyNewCondition}`. This simplifies the code and will allow more complex `getbc` cases to be expressed.


glwagner:
Note this PR could go further and also require _callers_ of `getbc` to invoke `bc.condition` directly. I hesitated to make this change because its a much broader change and also hardcodes `BoundaryCondition.condition`. I'm also not sure if other packages invoke `getbc`, cc @jagoosw 

This PR is the first step towards implementing a `RobinBoundaryCondition` as discussed on #4490

PR #4493, opened 2025-05-09
Fix dot implementation

simone-silvestri:
Working on the distributed CG with @xkykai we figured out that the `dot` implementation was wrong, i.e. it was accounting values inside the immersed boundary. 

This might be the issue with the custom PCG being terrible in the implicit free surface implementation?

### DIFF
--- a/src/Fields/field.jl
+++ b/src/Fields/field.jl
@@ -594,7 +594,9 @@ const ReducedAbstractField = Union{XReducedAbstractField,
                                    XYZReducedAbstractField}
 
 # TODO: needs test
-LinearAlgebra.dot(a::AbstractField, b::AbstractField) = mapreduce((x, y) -> x * y, +, interior(a), interior(b))
+LinearAlgebra.dot(a::AbstractField, b::AbstractField; condition=nothing) = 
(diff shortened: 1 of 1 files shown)


navidcy (approved):
Add that example as a test?

glwagner:
ooh, very very good catch

navidcy:
@simone-silvestri a scalar operation appeared in this PR here:

https://buildkite.com/clima/oceananigans/builds/23293#0196b8dd-2b63-48e9-9f69-299bf0e883a3/18-572

fixing this might be all we need to get this through?

simone-silvestri:
Hmmm, I wonder where this scalar operation comes from, given that we don't see it in other tests that use this free surface implementation.

src/Fields/field.jl
```diff
@@ -6,6 +6,7 @@ using Adapt
 using LinearAlgebra
 using KernelAbstractions: @kernel, @index
 using Base: @propagate_inbounds
+using GPUArrays
```
navidcy, line 9: do we still need GPUArrays or was it part of the investigations?

Issue #4367, opened 2025-04-08
Default Poincare coefficient for AMD is inappropriately small


jagoosw:
In the AMD model the default Poincare coefficient is 1/12:
https://github.com/CliMA/Oceananigans.jl/blob/dd778f85778dab21a2b3c8ae7c27d54022824235/src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl#L118-L119

but this is _probably_ not the right default to have.

The Poincare coefficient depends on the advection scheme, and for second order centered should be 1/3 ([Verstappen, 2014](https://github.com/user-attachments/files/19655921/06_verstappen-1.pdf)). Since that's the default advection scheme and (as far as I'm aware) the most common used for explicit LES, I think the defaults for AMD should probably be that too. 

I can't find any derivations for other advection schemes and as far as I'm aware the validity of AMD has only been checked with spectral methods and second order centered. 

I also noticed that reference is made to this in the [LES docs](https://clima.github.io/OceananigansDocumentation/stable/numerical_implementation/large_eddy_simulation/#Anisotropic-minimum-dissipation-models).

Should I open a PR changing the default to 1/3?

glwagner:
The reference in the docstring suggests 1/12: https://www.sciencedirect.com/science/article/pii/S0045793016304005?casa_token=MiYsFFOBzm0AAAAA:O9fn8l8EY_Wy5eJf-YbsdqThYVTGpKQ9aU3GWr88zfg10QLp-I4u7TfioHIF0y_czWUOx2Cm

It's all a bit of numerology to me though.

It might be best to _show_ that there's a problem with the default first to motivate changes, considering how many times it's been used, rather than relying on theoretical arguments only.

PS I wouldn't describe setting your eddy viscosity too high as a "bug".

jagoosw:
> The reference in the docstring suggests 1/12: https://www.sciencedirect.com/science/article/pii/S0045793016304005?casa_token=MiYsFFOBzm0AAAAA:O9fn8l8EY_Wy5eJf-YbsdqThYVTGpKQ9aU3GWr88zfg10QLp-I4u7TfioHIF0y_czWUOx2Cm

I think they're getting there a different way but that in equation 39 they're finding the equivalent of setting it to 1/3?

[...]

PR #4494, opened 2025-05-09
Changes default AMD poincare coefficient

jagoosw:
Following discussion in https://github.com/CliMA/Oceananigans.jl/issues/4367


glwagner (approved):

glwagner:
You'll have to manually change the coefficient in the regression tests (once those pass we can merge right away).

glwagner:
made the necessary change

glwagner:
If you want extra credit, you can expand the discussion in the docstring. You could link the relevant issue or just write a simple independent explanation.

navidcy (approved):

navidcy:
I think the discussion in #4367 worths some mention in the docstring. Even if it's just a short comment + pointing people to the PR discussion?

glwagner:
@jagoosw, can you update the docstring to explain why we chose 1/3 vs 1/12 with appropriate references?

jagoosw:
I'm not sure what to put since the references already there support 1/3 just the value was wrong?

src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl
```diff
@@ -63,7 +63,7 @@ Keyword arguments
         buoyancy modification term. This implementation differs from [Abkar16](@citet)'s proposal
         and the impact of this approximation has not been tested or validated.
 
-By default: `C = Cν = Cκ = 1/12`, which is appropriate for a finite-volume method employing a
+By default: `C = Cν = Cκ = 1/3`, which is appropriate for a finite-volume method employing a
 second-order advection scheme, and `Cb = nothing`, which turns off the buoyancy modification term.
```
glwagner, line 67: @jagoosw please add text here from the issue, with appropriate citations, that explain why 1/3 is the right choice

glwagner: Also put in a reference to https://github.com/CliMA/Oceananigans.jl/issues/4367 which has some relevant plots

glwagner: The discussion will be an elaboration of this remark:

"I think they're getting there a different way but that in equation 39 they're finding the equivalent of setting it to 1/3?"

validation/les/decaying_homogeneous_isotropic_turbulence.jl
```diff
@@ -0,0 +1,85 @@
...
+
+text!(ax, exp(0.8), exp(13); text = L"$E(k)\sim k^{-5/3}$", color = :white) 
+
+Colorbar(fig[1, 2], colormap = :oslo, colorrange = (0, 10), label = "Time (s)")
+
+k_filt = 1/sqrt(closure.Cν * 3 / (1/(2*minimum_xspacing(grid))^2 + 1/(2*minimum_yspacing(grid))^2 + 1/(2*minimum_zspacing(grid))^2))
+
+lines!(ax, ones(2) .* k_filt, [exp(0), exp(10)], color = :red, linestyle = :dash)
+
+text!(ax, k_filt, exp(10); text = "1/δ")
```
glwagner, line 83:
```suggestion
lines!(ax, [exp(1), exp(3)], x->(x/exp(1))^(-5/3)*exp(15), color = :red, linestyle = :dash)
text!(ax, exp(0.8), exp(13); text = L"$E(k)\sim k^{-5/3}$", color = :white) 
Colorbar(fig[1, 2], colormap = :oslo, colorrange = (0, 10), label = "Time (s)")

k_filt = 1/sqrt(closure.Cν * 3 / (1/(2*minimum_xspacing(grid))^2 + 1/(2*minimum_yspacing(grid))^2 + 1/(2*minimum_zspacing(grid))^2))

lines!(ax, ones(2) .* k_filt, [exp(0), exp(10)], color = :red, linestyle = :dash)
text!(ax, k_filt, exp(10); text = "1/δ")
```

src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl
```diff
@@ -49,25 +49,32 @@ Arguments
...
 
 `Cν` or `Cκ` may be numbers, or functions of `x, y, z`.
 
+The default Poincaré constant is appropriate for second order centered advection and is derived by
+descritising the sub grid scale energy production [Verstappen14](@citet). It is 4 times smaller than the 
+staight forward descritisation (which is appropriate for spectral methods) due to the "effective resolution" 
+of second order centered advection schemes.
+
+Discussion demonstraiting the correct energy dissipation behaviour with this Poincaré constant value
+can be found seen in https://github.com/CliMA/Oceananigans.jl/issues/4367.
```
glwagner, line 76:
```suggestion
By default: `C = Cν = Cκ = 1/3`, and `Cb = nothing`, which turns off the buoyancy modification term.
The default Poincaré constant `C = 1/3` is derived by discretizing subgrid scale energy production [Verstappen14](@citet) assuming a second-order advection scheme. `Cν` or `Cκ` may be numbers, or functions of `x, y, z`.

For more information about the default see the discussion at https://github.com/CliMA/Oceananigans.jl/issues/4367.
```

glwagner: Isn't Verstappen 2014 about a different model? Not AMD? Here's a link to the paper: https://authors.library.caltech.edu/records/dds0a-cgz43/files/06_verstappen.pdf?download=1

I couldn't find "C=1/3" written anywhere in the paper. I guess finding C=1/3 needs some interpretation. Another confusing thing is that they use a different definition of "Poincare constant" (which includes the grid scale) than we do.

glwagner: Ok so the logic is "after equation 14 Verstappen (2014) states that C_finite_volume = 3 x C_spectral", and "C_spectral = 1/12", therefore $C_finite_volume = 1/3".

In my opinion the empirical results are much stronger than these arcane numerology arguments! But, we have the empirical results too so that's good.

glwagner: I really do think we have to write this down. Otherwise the reference is more for show than an explicit argumentation. We have to mention how the argument applies to all constants in SGS models including Smagorinsky, AMD, and QR (it's admittedly been a long time since I read Rozema et al. 2015 but this is a new concept to me)

jagoosw: And they also empirically demonstrated it so I said that + that we did it in the issue too

glwagner (approved):

validation/les/decaying_homogeneous_isotropic_turbulence.jl
glwagner: @jagoosw can you address this?

tomchor:
Sorry I'm a bit late here. I understand why these changes were made. I'm still a bit confused though: what should this coefficient be for a WENO of order $n$ advection scheme? Do we know?

xkykai:
> No, we don't know. Although I think that combination exhibits the lowest effective resolution (thus increasing computational expense) compared to 2nd order advection + AMD, or WENO only. @xkykai and @simone-silvestri might know more.

Indeed. AMD + WENO seems to be the most dissipative when compared with 2nd order + AMD and WENO + no SGS closure.

PR #4500, opened 2025-05-12
Fix incorrect type parameter count in ConformalCubedSpherePanel and ConformalCubedSpherePanelGrid definitions

siddharthabishnu:
This PR fixes a bug in the definitions of `ConformalCubedSpherePanel` and `ConformalCubedSpherePanelGrid` by removing an extra sixth `:Any` type parameter that does not belong in the `OrthogonalSphericalShellGrid` type.

Previously:
```
const ConformalCubedSpherePanel = OrthogonalSphericalShellGrid{<:Any, FullyConnected, FullyConnected,
                                                               <:Any, <:Any, <:Any,
                                                               <:CubedSphereConformalMapping}

const ConformalCubedSpherePanelGrid = OrthogonalSphericalShellGrid{<:Any,
                                                                   <:Any,
                                                                   <:Any,
                                                                   <:Any,
                                                                   <:Any,
                                                                   <:Any,
                                                                   <:CubedSphereConformalMapping}
```
The sixth `:<Any` parameter was unnecessary and did not correspond to any field in the underlying `OrthogonalSphericalShellGrid` type. This PR removes that extraneous type to ensure correctness and proper method dispatch.

### DIFF
--- a/src/MultiRegion/cubed_sphere_grid.jl
+++ b/src/MultiRegion/cubed_sphere_grid.jl
@@ -12,16 +12,6 @@ using Distances
 import Oceananigans.Grids: grid_name
 import Oceananigans.BoundaryConditions: fill_halo_regions!
 
-const ConformalCubedSpherePanel = OrthogonalSphericalShellGrid{<:Any, FullyConnected, FullyConnected,
-                                                               <:Any, <:Any, <:CubedSphereConformalMapping}
-
-const ConformalCubedSpherePanelGrid = OrthogonalSphericalShellGrid{<:Any,
-                                                                   <:Any,
-                                                                   <:Any,
-                                                                   <:Any,
-                                                                   <:Any,
-                                                                   <:CubedSphereConformalMapping}
-
 const ConformalCubedSphereGrid{FT, TX, TY, TZ, CZ} = MultiRegionGrid{FT, TX, TY, TZ, CZ, <:CubedSpherePartition}
 
 """
--- a/src/MultiRegion/cubed_sphere_grid.jl
+++ b/src/MultiRegion/cubed_sphere_grid.jl
@@ -1,12 +1,13 @@
 using Oceananigans.Architectures: architecture
-using Oceananigans.Grids: conformal_cubed_sphere_panel,
-                          R_Earth,
+using Oceananigans.Grids: R_Earth,
                           halo_size,
                           size_summary,
                           total_length,
                           topology
 
 using CubedSphere
+using Oceananigans.OrthogonalSphericalShellGrids: conformal_cubed_sphere_panel
+
 using Distances
 
 import Oceananigans.Grids: grid_name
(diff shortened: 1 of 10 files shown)


glwagner (approved):

glwagner:
Also, I don't think the conformal cubed sphere grid should be defined in `Grids`. Probably the cubed sphere should be defined in a dedicated submodule.

siddharthabishnu:
> Also, I don't think the conformal cubed sphere grid should be defined in `Grids`. Probably the cubed sphere should be defined in a dedicated submodule.

Maybe I can move the panel grid definitions to `MultiRegion` for the time being?

glwagner:
We can yes. What's the advantage of using `MultiRegion` vs an independent module?

siddharthabishnu:
> We can yes. What's the advantage of using `MultiRegion` vs an independent module?

@glwagner I can't think of any particular advantage to using `MultiRegion`. I only meant it as a temporary placement. In a subsequent PR, I can move all the cubed sphere related code into an independent module.

glwagner:
ok!

simone-silvestri:
In my opinion the right location for the `ConformalCubedSpherePanelGrid` (the one panel grid) is the `OrthogonalSphericalShellGrids` module, where we also have the `RotatedLatitudeLongitudeGrid` and the `TripolarGrid` implementation. And then the `ConformalCubedSphereGrid` in the `Multiregion` module

siddharthabishnu:
> In my opinion the right location for the `ConformalCubedSpherePanelGrid` (the one panel grid) is the `OrthogonalSphericalShellGrids` module, where we also have the `RotatedLatitudeLongitudeGrid` and the `TripolarGrid` implementation. And then the `ConformalCubedSphereGrid` in the `Multiregion` module

Previously, the `ConformalCubedSpherePanel` and `ConformalCubedSpherePanelGrid` were defined in src/Grids/conformal_cubed_sphere_panel.jl.

src/MultiRegion/cubed_sphere_grid.jl
```diff
@@ -1,12 +1,13 @@
...
-using Oceananigans.Grids: conformal_cubed_sphere_panel,
-                          R_Earth,
+using Oceananigans.Grids: R_Earth,
                           halo_size,
                           size_summary,
                           total_length,
                           topology
 
 using CubedSphere
+using Oceananigans.OrthogonalSphericalShellGrids: conformal_cubed_sphere_panel
```
simone-silvestri, line 9:
```suggestion
```

src/OrthogonalSphericalShellGrids/OrthogonalSphericalShellGrids.jl
```diff
@@ -17,6 +18,7 @@ using Oceananigans.Grids: RightConnected
 using Oceananigans.Grids: R_Earth,
                           halo_size, spherical_area_quadrilateral,
                           lat_lon_to_cartesian, generate_coordinate, topology
+using Oceananigans.Grids: add_halos, fill_metric_halo_regions!
```
simone-silvestri, line 21:
```suggestion
```

PR #4501, opened 2025-05-12
Fix out-of-bounds access in vorticity computation at halo boundaries on conformal cubed sphere panels

siddharthabishnu:
This PR addresses and prevents out-of-bounds memory access during the computation of the vertical vorticity field on the conformal cubed sphere grid, specifically along the southernmost and westernmost halos of each panel.

To reproduce the issue, consider the following minimal working example (MWE) using a Julia session with bounds checking enabled (`--check-bounds=yes`):
```julia
using Oceananigans

arch = CPU()
grid = ConformalCubedSphereGrid(arch; panel_size=(4, 4, 1), z=(-1, 0), horizontal_direction_halo=2, radius=1)
u = XFaceField(grid); set!(u, 1)
v = YFaceField(grid); set!(v, 2)
ζ = Field{Face, Face, Center}(grid)

using Oceananigans.Operators: ζ₃ᶠᶠᶜ

@kernel function _compute_vorticity!(grid, u, v, ζ)
    @inbounds ζ[i, j, k] = ζ₃ᶠᶠᶜ(i, j, k, grid, u, v)
end

using Oceananigans.Grids: halo_size, total_size
using Oceananigans.MultiRegion: fill_halo_regions!
using Oceananigans.Utils: KernelParameters, launch!

function compute_vorticity!(grid, u, v, ζ)
    offset = -1 .* halo_size(grid)
    fill_halo_regions!((u, v))
    @apply_regionally begin
        kernel_parameters = KernelParameters(total_size(ζ[1]), offset)
        launch!(arch, grid, kernel_parameters, _compute_vorticity!, grid, u, v, ζ)
    end
end

compute_vorticity!(grid, u, v, ζ)
```
Running this MWE results in a bounds error:
```julia
ERROR: BoundsError: attempt to access 8×8 OffsetArray(::Matrix{Float64}, -1:6, -1:6) 
with eltype Float64 with indices -1:6×-1:6 at index [-2, -1]
```
Note that the vorticity field `ζ₃ᶠᶠᶜ` at a given grid point is computed by first evaluating the circulation around that point and then dividing it by the corresponding cell area. However, the circulation computation along the outermost line of the western halos `i = 1 - Hx` or the outermost line of the southern halos `j = 1 - Hy` tries to access velocity values beyond their designated halo regions. Specifically, the circulation defined by
```julia
Γ = δxᶠᶠᶜ(i, j, k, grid, Δy_qᶜᶠᶜ, v) - δyᶠᶠᶜ(i, j, k, grid, Δx_qᶠᶜᶜ, u)
```
tries to access `v[i = -Hx]` at `i = 1 - Hx`, and `u[j = -Hy]` at `j = 1 - Hy`, which are out-of-bounds.

Since the vorticity values along the outermost lines of the western and southern halos are not used anywhere, their numerical correctness is not required. So, this PR introduces a simple workaround by clamping the indices used for velocity access when computing vorticity at these halo locations:
```julia
ip = ifelse(i == 1 - Hx, 1, i)
jp = ifelse(j == 1 - Hy, 1, j)
```
These dummy indices `ip` and `jp` are then used in place of `i` and `j` within the vorticity operator, ensuring all array accesses remain within bounds while preserving simulation correctness.


src/Operators/vorticity_operators.jl
```diff
@@ -28,11 +28,13 @@ The vertical circulation associated with horizontal velocities ``u`` and ``v``.
 """
 @inline function Γᶠᶠᶜ(i, j, k, grid::ConformalCubedSpherePanel, u, v)
     Hx, Hy = grid.Hx, grid.Hy
+    ip = ifelse(i == 1 - Hx, 1, i)
+    jp = ifelse(j == 1 - Hy, 1, j)
```
glwagner, line 32:
```suggestion
    ip = max(1, i)
    jp = max(1, j)
```

i guess this means that the circulation cannot be computed in halos, which could make sense. What about on the otherwise of the panel where i = Nx + 1, etc?

navidcy: minor comment: But @glwagner, what you suggest is not exactly equivalent to what was there, right?

for
```Julia
Hx = 3
i = -1
ip = ifelse(i == 1 - Hx, 1, i)
```
would imply `ip = -1` while after your suggestion it'd be `ip = 1`.

siddharthabishnu: @glwagner On the contrary, we do compute vorticity everywhere, including within the halo regions---at least along the innermost lines of the eastern and northern halos at face-face locations. These values are required for the prognostic momentum equations. An alternative would be to exchange vorticity halos like other face-face fields (except at panel corners), but that would reduce performance, so we avoid it.

That said, due to the staggered grid layout, computing vorticity along the outermost lines of the western and southern halos requires velocity values beyond their designated halo regions, leading to out-of-bounds access. This issue does not occur along the outermost lines of the eastern and northern halos.

To illustrate using @navidcy’s example: suppose `Nx = Ny = 6` and `Hx = Hy = 3`. Computing vorticity at `i = 1 - Hx = -2` (outermost line of western halos) requires accessing `v` at `i = -Hx = -3`, which lies outside the halo region for `v`. Similarly, at `j = 1 - Hy = -2` (outermost line of southern halos), vorticity computation requires `u` at `j = -Hy = -3`, again outside the halo region for `u`.

However, computing vorticity along `i = Nx + Hx = 9` (outermost line of eastern halos) and `j = Ny + Hy = 9` (outermost line of northern halos) does not trigger this issue, as all necessary velocity values fall within their designated halo regions.

So, @glwagner, to conclude: the change you suggested would result in incorrect vorticity values across the entire western and southern halo regions, which we want to avoid. Our goal is simply to prevent out-of-bounds errors during vorticity computation along the outermost line of the western (`i = 1 - Hx`) and southern (`j = 1 - Hy`) halos.

Since we don't have access to the velocity values beyond their designated halo regions, the vorticity along these lines would be incorrect regardless. To safely bypass the error without impacting the rest of the domain, we use the workaround:
```
ip = ifelse(i == 1 - Hx, 1, i)
jp = ifelse(j == 1 - Hy, 1, j)
```
This yields placeholder values at those specific locations---still incorrect, but safe in the sense they do not cause memory access errors---while leaving all other vorticity computations unaffected.

glwagner: why not 

```julia
ip = max(2 - Hx, i)
```

for example? It's just easier to understand than introducing a jump to another index

siddharthabishnu: That should work. Incorporated in commit [6ed5bbf](https://github.com/CliMA/Oceananigans.jl/pull/4501/commits/6ed5bbf0fb830ba11154924e079f66dff3cec405).

PR #4503, opened 2025-05-13
Clean up `update_boundary_conditions!`

glwagner:
It keeps tripping me up to read `update_boundary_condition!` which in fact can update many boundary conditions. This PR tweaks the naming so that its easier to read/understand and makes a few other cosmetic changes. I also changed BoundaryAdjacentMean so that it doesn't output anything to test if this PR can resolve that issue too.


jagoosw (approved):

navidcy:
This doctest fails:

https://buildkite.com/clima/oceananigans/builds/23638#0196f898-e5a1-4133-b5dc-7d314e3eae7e

Is it a mistake that was introduced or do we need to update the doctest?

navidcy:
I think I fixed it.

Issue #4100, opened 2025-02-13
Comparing small values using `≈` for `FieldTimeSeries`


loisbaker:
This separates out an issue mentioned in #4077. The operator `≈` sets a relative tolerance that is too strict for values near zero (this caused errors for me in line 251 `field_time_series.jl` and line 10 `set_field_time_series.jl`) 

MWE based on usage in `field_time_series.jl`:
```julia
time_range = range(0, 1, length=11)
time_arr = [0:0.1:1;]
time_arr[1] = 1e-20
println(all(time_range .≈ time_arr)) #false
println(isapprox(time_arr,time_range)) #true
```
Ideally I would want the first print statement to also return `true`, so the `≈` operator (which is based on a relative rather than absolute tolerance) doesn't seem sensible. Some suggestions for where this is an issue:

On line 251 of `field_time_series.jl`
```julia
if all(time_range .≈ times) # good enough for most
```
could be replaced by
```julia
if isapprox(time_range,times)
```

On line 10 of `set_field_time_series.jl`, replace
```julia
findfirst(t -> t ≈ time, file_times)
```
with 
```julia
findfirst(t -> (isapprox(t,time) || isapprox(t,time,atol=1e-14)), file_times)
```
(though that's a bit arbitrary, there's probably a nicer way!)

Environment: Oceananigans.jl main branch.
```
julia> versioninfo()
Julia Version 1.10.5
Commit 6f3fdf7b362 (2024-08-27 14:19 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 32 × Intel(R) Core(TM) i9-14900K
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, goldmont)
Threads: 1 default, 0 interactive, 1 GC (on 32 virtual cores)
```

glwagner:
Thank you for opening this! I'm going to paste links to the code to make it easier to find and discuss solutions:

https://github.com/CliMA/Oceananigans.jl/blob/897c71d953926ec11122169a04d9b743235407b5/src/OutputReaders/field_time_series.jl#L249-L253

and

https://github.com/CliMA/Oceananigans.jl/blob/897c71d953926ec11122169a04d9b743235407b5/src/OutputReaders/set_field_time_series.jl#L10

which is used here

[...]

PR #4505, opened 2025-05-13
Fix some `FieldTimeSeries` indexing bugs

loisbaker:
This PR fixes some small indexing bugs for `FieldTimeSeries` described in https://github.com/CliMA/Oceananigans.jl/issues/4100

- Changes the way that times are compared in https://github.com/CliMA/Oceananigans.jl/blob/d1e9b46586150f2f3508ac8a861d87ef67e22b1c/src/OutputReaders/field_time_series.jl#L252 and https://github.com/CliMA/Oceananigans.jl/blob/d1e9b46586150f2f3508ac8a861d87ef67e22b1c/src/OutputReaders/set_field_time_series.jl#L10 so that `≈` is replaced with a comparison using a vector norm in the first case, and using an absolute tolerance based on a characteristic timescale in the second case. This avoids returning `false` due to the default relative error criterion when comparing values close to zero. 
- Moves the time array that is indexed in https://github.com/CliMA/Oceananigans.jl/blob/d1e9b46586150f2f3508ac8a861d87ef67e22b1c/src/OutputReaders/set_field_time_series.jl#L25 to the CPU so errors aren't thrown when `fts.times` is a `CuArray`.


src/OutputReaders/set_field_time_series.jl
```diff
@@ -7,23 +7,34 @@ using Oceananigans.Architectures: cpu_architecture
 
 iterations_from_file(file) = parse.(Int, keys(file["timeseries/t"]))
 
-find_time_index(time::Number, file_times)       = findfirst(t -> t ≈ time, file_times)
-find_time_index(time::AbstractTime, file_times) = findfirst(t -> t == time, file_times)
+function find_time_index(time::Number, file_times, dt)
+    ϵ = 100 * eps(eltype(file_times))
+    return findfirst(t -> isapprox(t, time, atol=ϵ*dt), file_times)
+end
```
navidcy, line 13: what's the `dt` here? I'm bit confused..
is it the "characteristic time scale" mentioned in the first comment of the PR?

I see below that the dt is the average difference in times...

src/OutputReaders/set_field_time_series.jl
```diff
@@ -7,23 +7,34 @@ using Oceananigans.Architectures: cpu_architecture
...
 
 function set!(fts::InMemoryFTS, path::String=fts.path, name::String=fts.name; warn_missing_data=true)
     file = jldopen(path; fts.reader_kw...)
     file_iterations = iterations_from_file(file)
     file_times = [file["timeseries/t/$i"] for i in file_iterations]
     close(file)
-
+    
+    # Compute a timescale for comparisons
+    dt = mean(file_times[2:end] - file_times[1:end-1])
```
navidcy, line 24:
```suggestion
    dt = mean(diff(file_times))
```

src/OutputReaders/set_field_time_series.jl
navidcy: oh actually I just noticed this followed from issue #4100! let me read there, things might become clear to me!

src/OutputReaders/set_field_time_series.jl
navidcy, line 24: why don't we compute this within `find_time_index`? just a thought...
or something like:

```julia
function find_time_index(time::Number, file_times; time_scale = mean(diff(file_times)))
    ϵ = 100 * eps(eltype(file_times))
    return findfirst(t -> isapprox(t, time, atol=ϵ*time_scale), file_times)
end
```

src/OutputReaders/set_field_time_series.jl
```diff
@@ -7,23 +7,34 @@ using Oceananigans.Architectures: cpu_architecture
 
 iterations_from_file(file) = parse.(Int, keys(file["timeseries/t"]))
 
-find_time_index(time::Number, file_times)       = findfirst(t -> t ≈ time, file_times)
-find_time_index(time::AbstractTime, file_times) = findfirst(t -> t == time, file_times)
```
navidcy, line 11: why delete the ::AbstractTime method?

loisbaker: Sorry, I think it was just moved down not deleted. I thought it would need the dt argument too, but that can be avoided if dt is computed within the function.

navidcy: Oh yes. My bad!

src/OutputReaders/set_field_time_series.jl
glwagner: Hmm, it's really the time increment between time points, so I think `Δt` is more appropriate. Just suggest using the `Δ` not `dt` for readability

glwagner: Precomputing serves two purposes: 1) we don't recompute, which could become important for very long time series, and 2) as a point of design, this allows us to change the estimate of the characteristic time increment later. Right now we use the mean of all time increments, but this is not the only choice one might make.

navidcy: Or `mean_Δt`

glwagner:
@loisbaker should we add tests at all? Ok if that's a todo for the future, but tests will ensure these changes are lasting

src/OutputReaders/field_time_series.jl
```diff
@@ -247,9 +247,9 @@ mutable struct FieldTimeSeries{LX, LY, LZ, TI, K, I, D, G, ET, B, χ, P, N, KW}
         end
 
         if times isa AbstractArray
-            # Try to convert to a range, cuz
+            # Try to convert to a range
```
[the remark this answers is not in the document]
loisbaker: Actually, what is the advantage of this? I wanted to fix it originally because of the error with indexing the `CuArray` in [set_field_time_series.jl](https://github.com/CliMA/Oceananigans.jl/blob/d1e9b46586150f2f3508ac8a861d87ef67e22b1c/src/OutputReaders/set_field_time_series.jl#L25), but it works as an array now.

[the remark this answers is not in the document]
glwagner: Ranges are lighter weight, since you can express an arbitrary long series with just a few numbers and compute the entries of the range on the fly rather than storing/fetching from memory. That said, it might not matter much ultimately...

glwagner:
> > @loisbaker should we add tests at all? Ok if that's a todo for the future, but tests will ensure these changes are lasting
> 
> Sure, I can work on that - a test to check that `find_time_index` will pick the right index when the error between the two compared values is sufficiently small, and doesn't otherwise? Would you also make a test to check that the range conversion in `field_time_series.jl` only happens when the time values are sufficiently close to evenly spaced?

Perhaps just a simple test that resembles the case you presented that illustrated the bug would work (the more fine-grained tests can be nice too, but I don't think necessary in this case and they are more work).

We can also do in a future PR, since I think this is otherwise read to merge -- your call!

glwagner (approved):

loisbaker:
> > > @loisbaker should we add tests at all? Ok if that's a todo for the future, but tests will ensure these changes are lasting
> > 
> > 
> > Sure, I can work on that - a test to check that `find_time_index` will pick the right index when the error between the two compared values is sufficiently small, and doesn't otherwise? Would you also make a test to check that the range conversion in `field_time_series.jl` only happens when the time values are sufficiently close to evenly spaced?
> 
> Perhaps just a simple test that resembles the case you presented that illustrated the bug would work (the more fine-grained tests can be nice too, but I don't think necessary in this case and they are more work).
> 
> We can also do in a future PR, since I think this is otherwise read to merge -- your call!

Great, I'll put this on the to-do list for a future PR!

src/OutputReaders/set_field_time_series.jl
```diff
@@ -7,23 +7,35 @@ using Oceananigans.Architectures: cpu_architecture
...
+function find_time_index(time::Number, file_times, Δt)
+    # Accommodate round-off discrepancies between the FTS times and file times
```
navidcy, line 11:
```suggestion
    # Accommodate round-off discrepancies between the FTS times and file times
    # (see https://github.com/CliMA/Oceananigans.jl/pull/4505)
```

PR #4509, opened 2025-05-14
Try to update_state during initalization for ReactantModel

glwagner:
cc @wsmoses


wsmoses:
Ah wait there is still a pending issue on gb ci with this am looking at

PR #4513, opened 2025-05-15
[TESTING] reduce update hydrostatic model

wsmoses:


glwagner:
I believe this is no longer needed, but re-open if it is

Issue #4483, opened 2025-05-07
Error running a distributed nonhydrostatic model on CPU


tomchor:
I'm trying to get started with the distributed architecture. The goal is to run on multiple GPUs, but for now I'm starting with multiple CPUs. As a starting point I'm trying to run the [distributed_nonhydrostatic_turbulence.jl](https://github.com/CliMA/Oceananigans.jl/blob/main/validation/distributed_simulations/distributed_nonhydrostatic_turbulence.jl) script, but I get the following error when `run!`ning the simulation:

```
[ Info: Initializing simulation...
ERROR: LoadError: MethodError: no method matching interior_work_layout(::RectilinearGrid{Float64, Periodic, Periodic, Flat, Oceananigans.Grids.StaticVerticalDiscretization{Nothing, Nothing, Float64, Float64}, Float64, Float64, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, Distributed{CPU, false, Partition{Nothing, Nothing, Nothing}, Tuple{Int64, Int64, Int64}, Int64, Tuple{Int64, Int64, Int64}, Oceananigans.DistributedComputations.NeighboringRanks{Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing}, MPI.Comm, Vector{MPI.Request}, Base.RefValue{Int64}, Nothing}}, ::Tuple{}, ::Tuple{DataType, DataType, DataType})

Closest candidates are:
  interior_work_layout(::Any, !Matched::Symbol, ::Any)
   @ Oceananigans ~/.julia/packages/Oceananigans/vnQQc/src/Utils/kernel_launching.jl:155

Stacktrace:
  [1] #configure_kernel#7
    @ ~/.julia/packages/Oceananigans/vnQQc/src/Utils/kernel_launching.jl:257 [inlined]
  [2] configure_kernel
    @ ~/.julia/packages/Oceananigans/vnQQc/src/Utils/kernel_launching.jl:246 [inlined]
[...]

PR #4515, opened 2025-05-16
Warn if using a `Distributed` architecture using a single rank

simone-silvestri:
@tomchor this might help catching incorrectly configured MPI, see #4483

### DIFF
--- a/src/DistributedComputations/distributed_architectures.jl
+++ b/src/DistributedComputations/distributed_architectures.jl
@@ -264,6 +264,11 @@ function Distributed(child_architecture = CPU();
     ranks = Rx, Ry, Rz = size(partition)
     partition_ranks = Rx * Ry * Rz
 
+    if partition_ranks == 1
+        @warn "The script has been launched using a single MPI rank. \n
+               This might not be intended but a result of an incorrect MPI configuration."
+    end
+
     # TODO: make this error refer to `partition` (user input) rather than `ranks`
     if partition_ranks != mpi_ranks
         throw(ArgumentError("Partition($Rx, $Ry, $Rz) [$partition_ranks ranks] inconsistent " *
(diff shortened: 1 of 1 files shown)


src/DistributedComputations/distributed_architectures.jl
```diff
@@ -264,6 +264,11 @@ function Distributed(child_architecture = CPU();
     ranks = Rx, Ry, Rz = size(partition)
     partition_ranks = Rx * Ry * Rz
 
+    if partition_ranks == 1
+        @warn "The script has been launched using a single MPI rank. \n" *
+              "This might not be intended but a result of an incorrect MPI configuration."
```
glwagner, line 269:
```suggestion
        @warn "We are building a Distributed architecture on a single MPI rank. \n" *
              "This can occur when MPI is incorrectly configured."
```

glwagner: We don't know this comes from a script. So I changed the comment to be more specific.

Is the configuration issue with MPI.jl or MPI itself?

glwagner (approved):
I think this is a good idea, we just want to make sure the warning is correct and clear.

src/DistributedComputations/distributed_architectures.jl
simone-silvestri: Sure, we can point at MPI.jl config docs. It's a problem only of the julia MPI not finding the MPI implementation and acting as if we launch with one rank. So each rank has an independent execution with `MPI.Comm_size == 0` and `MPI.Comm_rank == 0`

glwagner: ok, so can we refine the warning to indicate that MPI.jl is not configured correctly? I think it is confusing to say "MPI is not configured correctly", but hte problem is actually just MPI.jl

PR #4520, opened 2025-05-19
Use default `prettyurls=true` in docs build

navidcy:


navidcy:
gotcha

docs/make.jl
```diff
@@ -170,7 +170,7 @@ pages = [
 #####
 
 format = Documenter.HTML(collapselevel = 1,
-                         prettyurls = ci_build,
+                         prettyurls = true,
```
glwagner, line 173: is this the default?

navidcy: That’s what the documenter docs suggest

navidcy: Oh I see what you mean. I’ll check!

PR #4521, opened 2025-05-19
Fix a bug in the handshake macro

simone-silvestri:
Renaming `mpi_initialized` was creating problems when used in other packages.


navidcy (approved):

glwagner:
oh yeah, that is a dirty design too, good to fix

PR #4522, opened 2025-05-19
Improve variance dissipation computation

simone-silvestri:
this PR implements some quality of life changes to the `VarianceDissipation` metric as well as some necessary changes to support more involved time-stepping schemes like RK3.
In particular:
- Instead of being "per-model", the metric is switched to be "per-tracer", which simplifies the implementation. (It might lead to an overhead in compute and memory as, for various tracers, the stored transport should be the same, but the additional compute is quite irrelevant, and the memory overhead can be avoided by explicitly passing preconstructed transport to `VarianceDissipation`).
- The metric is now implemented as a struct to use as a `Callback` rather than a diagnostic. This allows us to force the `UpdateStateCallsite`, which is necessary if we want to have an RK3 implementation.
- We allow schedules different than `IterationInterval(1)` by forcing `ConsecutiveIterations(schedule)` if the schedule is different than `IterationInterval(1)`, which allows us to compute the metric at any schedule we want. At the moment, though, the output schedule should be equal to the schedule we pass to `Callback(t::VarianceDissipation; schedule)` if we want to save down meaningful results (the test is modified to capture cases with different schedules). We can improve on this in future PRs

An example of the new UI is in the test and the validation example


simone-silvestri:
`TimeInterval` schedules are not supported at the moment because the time step might change size, invalidating the dissipation computation.

simone-silvestri:
Should be ready to go. The docs failing is because of #4520

src/Simulations/VarianceDissipationComputations/VarianceDissipationComputations.jl
```diff
@@ -0,0 +1,163 @@
...
+    advective_production :: P
+    diffusive_production :: K
+    advective_fluxes :: A
+    diffusive_fluxes :: D
+    previous_state :: S
+    gradient_squared :: G
+    tracer_name :: Symbol
+end
+
+function vector_field(grid)
```
glwagner, line 44: this is wrong if the bc's are not regularized

glwagner: it also doesn't return a `field` --- it returns a NamedTuple

glwagner: `c_staggered_vector`

src/Simulations/VarianceDissipationComputations/VarianceDissipationComputations.jl
```diff
@@ -0,0 +1,163 @@
...
+    previous_state   = (; cⁿ⁻¹, Uⁿ⁻¹, Uⁿ)
+    advective_fluxes = (; Fⁿ, Fⁿ⁻¹)
+    diffusive_fluxes = (; Vⁿ, Vⁿ⁻¹)
+
+    gradients = deepcopy(P)
+
+    return VarianceDissipation(P, K, advective_fluxes, diffusive_fluxes, previous_state, gradients, tracer_name)
+end
+
+function (ϵ::VarianceDissipation)(model)
```
glwagner, line 99: if this is a callback it needs to be a function of `Simulation`

glwagner: or is it using a different callsite?

a docstring is a good idea here

tomchor:
>     the memory overhead can be avoided by explicitly passing preconstructed transport to `VarianceDissipation`).

Can we test that this works on CI?

test/test_implicit_diffusion_diagnostic.jl
```diff
@@ -46,33 +51,50 @@ function test_implicit_diffusion_diagnostic(arch, dim, schedule)
     grid = periodic_grid(arch, Val(dim))
 
     # Change to test pure advection schemes
-    tracer_advection = WENO(order=5)
-    closure = ScalarDiffusivity(κ=1e-3)
+    tracer_advection = (c = WENO(order=7), d = Centered(order=4))
+    closure = ScalarDiffusivity(κ=(c = 1e-3, d=1e-5))
```
simone-silvestri, line 55:
```suggestion
    tracer_advection = (c=WENO(order=7), d = Centered(order=4))
    closure = ScalarDiffusivity(κ=(c=1e-3, d=1e-5))
```

simone-silvestri:
Comments were addressed. If no one has an objection, I would like to merge this to work on #4523 (and generalization to flux-form momentum advection)

src/Simulations/VarianceDissipationComputations/VarianceDissipationComputations.jl
```diff
@@ -0,0 +1,165 @@
...
+    if schedule isa TimeInterval # Time step might change size between iterations invalidating `VarianceDissipation`
+        throw(ArgumentError("a VarianceDissipation computation must be executed on `IterationInterval`s. \n" *
+                            "The provided `TimeInterval` schedule is not supported"))
+    end
+
+    if !(schedule isa ConsecutiveIterations || schedule == IterationInterval(1))
+        @warn "VarianceDissipation callback must be called every Iteration or on `ConsecutiveIterations`. \n" * 
+              "Changing `schedule` to `ConsecutiveIterations(schedule)`."
+        schedule = ConsecutiveIterations(schedule)
+    end
```
glwagner, line 160: I suggest using dispatch to handle this

src/Simulations/VarianceDissipationComputations/compute_dissipation.jl
```diff
@@ -29,7 +21,7 @@ U = 1.5 Uⁿ - 0.5 Uⁿ⁻¹
 
 For an RK3 method (not implemented at the moment), the whole substepping procedure needs to be accounted for.
 """
-function compute_dissipation!(dissipation, model, tracer_name::Symbol)
+function compute_dissipation!(model, dissipation, tracer_name::Symbol)
```
glwagner, line 24: does this modify the model or dissipation?

src/Simulations/VarianceDissipationComputations/update_fluxes.jl
```diff
@@ -2,11 +2,11 @@ using Oceananigans: fields
 using Oceananigans.Grids: topology, Flat
 
 # Store advective and diffusive fluxes for dissipation computation
-function cache_fluxes!(model, dissipation)
-    grid   = model.grid
-    arch   = architecture(grid)
-    sz     = size(model.tracers[1].data)
-    of     = model.tracers[1].data.offsets
+function cache_fluxes!(model, dissipation, tracer_name::Symbol)
```
glwagner, line 5: why dispatch? what happens in the other case?

src/Simulations/Simulations.jl
```diff
@@ -25,4 +25,9 @@ include("simulation.jl")
 include("run.jl")
 include("time_step_wizard.jl")
 
+# Implementation of the diagnostic for computing the dissipation rate
+include("VarianceDissipationComputations/VarianceDissipationComputations.jl")
```
glwagner, line 29: I think this belongs in Models/

src/Simulations/VarianceDissipationComputations/compute_dissipation.jl
simone-silvestri: it modifies the dissipation object, I can move the dissipation in the first position

src/Simulations/Simulations.jl
simone-silvestri: I can move it in models

glwagner: Remember that the Simulations module is supposed to be designed so that it can be lifted out.

One change we can make if we want to clean things up is to put the Simulations before Models, if you want to extend Callback in Models.

src/Simulations/VarianceDissipationComputations/update_fluxes.jl
simone-silvestri: there should be no other case, I think that remained there from earlier dispatch attempts, I can remove it

glwagner:
The main issue is that the VarianceDissipation is a Models feature, not a Simulations feature.

src/Simulations/VarianceDissipationComputations/VarianceDissipationComputations.jl
simone-silvestri: I can add a validation function that I can dispatch on.

src/Simulations/callback.jl
```diff
@@ -142,3 +143,43 @@ function add_callback!(simulation, func, schedule = IterationInterval(1);
...
+function validate_schedule(::VarianceDissipation, schedule::IterationInterval) 
+    if !(schedule == IterationInterval(1))
+        @warn "VarianceDissipation callback must be called every Iteration or on `ConsecutiveIterations`. \n" * 
+              "Changing `schedule` to `ConsecutiveIterations(schedule)`."
+        schedule = ConsecutiveIterations(schedule)
+    end
+    return schedule
+end
+
+function validate_dissipation_schedule(::VarianceDissipation, schedule::ConsecutiveIterations)
```
simone-silvestri, line 179:
```suggestion
function validate_schedule(::VarianceDissipation, schedule::ConsecutiveIterations)
```

PR #4525, opened 2025-05-19
FirstTimeStep: don't call update_state twice in a row

wsmoses:


glwagner:
Then we have to uncomment

https://github.com/CliMA/Oceananigans.jl/blob/7a04d840b24ced83d4bb5926b2e7513511645fa3/ext/OceananigansReactantExt/TimeSteppers.jl#L68

?

glwagner:
If we can support all initialization in the standard way, then there's no need for `first_time_step!`

wsmoses:
No I don't think so, the issue was that the non reactant path was calling update_state twice for the 0th' timestep.

Specificailly non reactant was:
```
function first_time_step!(model::AbstractModel, Δt)
    initialize!(model)
    update_state!(model)
    time_step!(model, Δt)
    return nothing
end
```

but non-reactant, time_step! had if iteration == 0 update_state. 
So as a result non-reactant this would call update_state! twice at the start [not once as previously].

Reactant side doesn't call update_state within time_step, so it does need an update_state explicitly within first_time_Step, which it has

glwagner:
there's no issue calling update_state twice though, is there?

wsmoses:
if it modifies things in place, can't that cause a correctness issue though? when compared with not calling it twice

glwagner:
`update_state!` is supposed to compute the "diagnostic" part of the model state, given the prognostic part of the model state. If it's called twice, it should simply overwrite the diagnostic state with identical values as the previous call.

wsmoses:
so in its current form we've now confirmed this does fix a correctness issue in reactant vs not in catke

glwagner (approved):

PR #4529, opened 2025-05-20
Export `@compute` at the top level

tomchor:
I wonder what people think of this. I find this macro really useful and always find myself using it, getting an error that `@compute` is not defined, and then (properly) `using` it!

### DIFF
--- a/src/Oceananigans.jl
+++ b/src/Oceananigans.jl
@@ -39,7 +39,7 @@ export
     # Fields and field manipulation
     Field, CenterField, XFaceField, YFaceField, ZFaceField,
     Average, Integral, CumulativeIntegral, Reduction, Accumulation, BackgroundField,
-    interior, set!, compute!, regrid!,
+    interior, set!, compute!, @compute, regrid!,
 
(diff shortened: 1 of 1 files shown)


simone-silvestri:
As I understand now, `@compute` is equivalent to `compute!`. 
i.e.
```julia
@compute f = Field(op)
```
is equivalent to 
```julia
f = compute!(Field(op))
```
I think a good change would be to make `@compute` also construct the field, in practice, having
```julia
@compute f = op
```

glwagner:
What about calling `compute!` in the `Field` constructor?

simone-silvestri:
Sounds like a good idea. We could add a keyword argument to
https://github.com/CliMA/Oceananigans.jl/blob/925f4c9cee0fb50743884c4e4b1e55676c216990/src/AbstractOperations/computed_field.jl#L38
which defaults to true.

glwagner:
See #4531

tomchor:
Superseded by https://github.com/CliMA/Oceananigans.jl/pull/4531

PR #4531, opened 2025-05-20
Call `compute!` in constructor for `ComputedField`

glwagner:
This might be a better path to usability than exporting more functions / defining more syntax. Let me know what you think @tomchor .


tomchor:
Just to clarify, this would immediately compute a `Field` right after it's created always?

glwagner:
> Just to clarify, this would immediately compute a `Field` right after it's created always?

yes. @simone-silvestri suggested adding a kwarg like `compute=true` which could be set to `false` if desired for some reason

tomchor:
Yeah I like this solution. I always assumed this wasn't done (i.e. we had to manually call computation sometimes) for performance reasons. But if that's not an issue, then great!

On a side note, if this is the case, do we still want/need to export `compute!()` at the top level?

glwagner:
> Yeah I like this solution. I always assumed this wasn't done (i.e. we had to manually call computation sometimes) for performance reasons. But if that's not an issue, then great!

We did not discuss it at length. But an argument was made that requiring the user to call `compute!` explicitly illustrated the user interface.

> On a side note, if this is the case, do we still want/need to export `compute!()` at the top level?

I don't think this change means that users don't need `compute!`.

tomchor (approved):
I don't 100% understand why the bib modifications were done here, but this is looking good. With this change, I suspect several instances of `compute!()` are now not necessary in the docs. I fixed one of them I saw in the examples. Do we want to hunt down those before merging?

glwagner:
I am thinking that one change we should make is to add `compute=true` default, and then use `compute=false` when we build diagnostics (because that is a redundant computation)

glwagner:
Let's hold off on changing the docs. This change will have larger implications for how we explain how `Field` is a lazy abstraction and I feel we can make that a separate piece of work.

navidcy:
hm... something seems to have broken with the NetCDFWriter?

https://buildkite.com/clima/oceananigans/builds/23644#0196f9a9-f391-482f-9d75-48479e226007/18-7481

src/OutputWriters/output_construction.jl
```diff
@@ -54,7 +54,11 @@ end
 
 function construct_output(user_output::Union{AbstractField, Reduction}, grid, user_indices, with_halos)
     indices = output_indices(user_output, grid, user_indices, with_halos)
-    return Field(user_output; indices)
+
+    # Don't compute AbstractOperations or Reductions
+    kw = user_output isa Field ? (; indices) : (compute=false; indices)
+
+    return Field(user_output; kw...)
```
glwagner, line 61:
```suggestion
    # Don't compute AbstractOperations or Reductions
    additional_kw = user_output isa Field ? NamedTuple() : (; compute=false)

    return Field(user_output; indices, additional_kw...)
```

tomchor:
@glwagner is this ready to merge?

docs/src/operations.md
```diff
@@ -296,7 +296,7 @@ conditional_∫c = Field(Integral(c, dims=(1, 2), condition=cond)) # integrate o
 ├── operand: Integral of ConditionalOperation of BinaryOperation at (Center, Center, Center) with condition cond (generic function with 1 method) over dims (1, 2)
 ├── status: time=0.0
 └── data: 1×1×11 OffsetArray(::Array{Float64, 3}, 1:1, 1:1, -2:8) with eltype Float64 with indices 1:1×1:1×-2:8
-    └── max=0.0, min=0.0, mean=0.0
+    └── max=1.27516e14, min=1.27516e14, mean=1.27516e14
 ```
 
 Above we have attached a condition to the operand.
```
navidcy, line 286: we need to change the text here because it reads as if the field is not yet computed
also need to remove the `compute!(conditional_∫c)` from below?

navidcy:
What about

https://github.com/CliMA/Oceananigans.jl/blob/86765e849cc979c28687522a834cba93b3fdae85/examples/horizontal_convection.jl#L309

https://github.com/CliMA/Oceananigans.jl/blob/86765e849cc979c28687522a834cba93b3fdae85/examples/horizontal_convection.jl#L313

?

navidcy:
OK, I'll merge and we can change more things (eg in the docs) later

PR #4526, opened 2025-05-19
Don't time-step TKE unless warranted (add clock to CATKE and TKEDissipation)

glwagner:
This prevents wrong answers when repeatedly calling `update_state!`

cc @wsmoses @jlk9 

@navidcy this may also affect CATKE checkpointing (but it still is not possible unless we cache `previous_velocities` and the time-averaged buoyancy flux).

PR #4533, opened 2025-05-21
Revert "Don't time-step TKE unless warranted (add clock to CATKE and …

wsmoses:
…TKEDissipation) (#4526)"

This reverts commit 7c561b694a7b5aca4b44746187ddd2cd18c495ba.

### DIFF
--- a/ext/OceananigansReactantExt/OceananigansReactantExt.jl
+++ b/ext/OceananigansReactantExt/OceananigansReactantExt.jl
@@ -27,6 +27,9 @@ using .Fields
 include("TimeSteppers.jl")
 using .TimeSteppers
 
+include("TurbulenceClosures.jl")
+using .TurbulenceClosures
+
 include("Models.jl")
 using .Models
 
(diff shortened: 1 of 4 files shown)


glwagner:
We do need this for correctness though (outside ReactantState) and also for checkpointing with CATKE

wsmoses:
The traced if introduced here is broken for joes code atm, can it be written without the traced if (notably removing a trace on the if will also break if the condition depends on a traced number)

glwagner:
You mean, just remove `@trace`?

wsmoses:
No, I mean restore the old if condition (which apparently used no traces variable)

wsmoses:
Alternatively, could we revert this PR until there is a version of this that doesn’t break joes code, then remand (along with the requisite test)

glwagner:
Right so unfortunately, with this algorithm, we cannot make this correct without using an `if` on a clock variable that is traced in Reactant. I'm ok to temporarily revert this. But the broader question is probably whether we can support `@trace if` (which would simplify a lot of other code too including `time_step!`)

Issue #4451, opened 2025-04-29
`PerturbationAdvectionOpenBoundaryCondition` + `ConjugateGradientSolver` causes simulation to blow-up with versions 0.96.20+


tomchor:
After 0.96.20, I've seen some scripts that were previously running lead to `NaN`s when using `PerturbationAdvectionOpenBoundaryCondition` with `ConjugateGradientSolver` on a `ImmersedBoundaryGrid`, even with extremely simple bathymetries. For example, the simulation below is a domain with a fake flat bottom. It runs fine on 0.96.19, but it blows up with 0.96.20:

```julia
using Oceananigans
using Oceananigans.BoundaryConditions: PerturbationAdvectionOpenBoundaryCondition
using Oceananigans.Solvers: ConjugateGradientPoissonSolver

Lx, Lz = 10, 3
Nx = Nz = 8

grid_base = RectilinearGrid(topology = (Bounded, Flat, Bounded), size = (Nx, Nz), x = (0, Lx), z = (0, Lz))
flat_bottom(x) = 1
grid = ImmersedBoundaryGrid(grid_base, PartialCellBottom(flat_bottom))
U = 1
inflow_timescale = 1e-4
outflow_timescale = Inf
u_boundaries = FieldBoundaryConditions(
    west   = PerturbationAdvectionOpenBoundaryCondition(U; inflow_timescale, outflow_timescale),
    east   = PerturbationAdvectionOpenBoundaryCondition(U; inflow_timescale, outflow_timescale),)

model = NonhydrostaticModel(; grid, boundary_conditions = (; u = u_boundaries),
                              pressure_solver = ConjugateGradientPoissonSolver(grid, maxiter=5),
                              advection = WENO(; grid, order=5))
set!(model, u=U)

Δt = 0.5 * minimum_zspacing(grid) / abs(U)
simulation = Simulation(model; Δt = Δt, stop_time=10)

conjure_time_step_wizard!(simulation, IterationInterval(1), cfl = 0.1)

run!(simulation)
```

In the particular example above there are some tweaks I can make to get it to run, but that's not possible for my production-ready scripts.

I'm assuming this is due to https://github.com/CliMA/Oceananigans.jl/pull/4041, but I can't figure out how to fix this behavior.

cc @amontoison @glwagner

amontoison:
@tomchor 
The issue is the preconditioner used in `ConjugateGradientPoissonSolver`.
[...]

PR #4041, opened 2025-01-13
Interface to Krylov.jl solvers

amontoison:
Replace #3812
@glwagner @michel2323

I added all the routines needed by Krylov, but we probably want them implemented as kernels.

What I don't know is how we can overload the operators such that when we want to perform an operator-vector product with a `KrylovField`, we unwrap the `Field` in the custom vector type.
I think Gregory has an idea how we can do that.

The second question is how can we create a new `KrylovField` from an existing one? 
Can we create a function `KrylovField{T,F}(undef, n)`?  
I think we can't because `Field`s are like 3D arrays, and Krylov.jl should be modified to rely on a function like `similar` when a workspace is created.

PR #4539, opened 2025-05-23
Add tests for `ConjugateGradientPoissonSolver` with `PerturbationAdvectionOpenBoundaryConditions` in `ImmersedBoundaryGrid`s

tomchor:
Originally this PR did what is described below. Then https://github.com/CliMA/Oceananigans.jl/pull/4582 and https://github.com/CliMA/Oceananigans.jl/pull/4554 were merged, which did part of what was originally proposed here. Now this PR will only add tests to ensure that things keep working with the `ConjugateGradientPoissonSolver` when using `PerturbationAdvectionOpenBoundaryConditions` in `ImmersedBoundaryGrid`s. At least in simple cases.

---

The goal of this PR is to fix the initialization of `PerturbationAdvectionOpenBoundaryCondition`. I was playing around with it and did notice that the indices at `i=1` weren't receiving the external boundary condition (the one we get through `getbc()`) which led to, so this PR tests a solution for that. For now this fix is just an `ifelse` statement, but maybe we can find a better one.

I hadn't opened a PR earlier because I was trying to figure out why the implementation of the open boundary condition was different on for left boundary than for the right boundaries (right boundaries seem to work fine). It was unclear to be where the asymmetry was coming from. Then @jagoosw's comment comment earlier mentioned that the left boundary in stepped but apparently the right boundary is not.

As a benchmark, this MWE is currently failing on main, but is running on this branch:

```julia
using Oceananigans

Lx, Lz = 10, 3
Nx = Nz = 8

grid_base = RectilinearGrid(topology = (Bounded, Flat, Bounded), size = (Nx, Nz), x = (0, Lx), z = (0, Lz))
flat_bottom(x) = 1
grid = ImmersedBoundaryGrid(grid_base, PartialCellBottom(flat_bottom))
U = 1
inflow_timescale = 0.0
outflow_timescale = Inf
u_boundaries = FieldBoundaryConditions(
    west   = PerturbationAdvectionOpenBoundaryCondition(U; inflow_timescale, outflow_timescale),
    east   = PerturbationAdvectionOpenBoundaryCondition(U; inflow_timescale, outflow_timescale),)
    
                              boundary_conditions = (; u = u_boundaries),
                              pressure_solver = ConjugateGradientPoissonSolver(grid, maxiter=10),
                              advection = WENO(; grid, order=5)
                             )
u₀(x, z) = U + 1e-2*rand()
set!(model, u=u₀, enforce_incompressibility=true)
```

For now this PR also reverts the FFT preconditioner to what it was before https://github.com/CliMA/Oceananigans.jl/pull/4041 because it makes it much easier for simulations with both `ImmersedBoundaries` and `OpenBoundaryConditions` to run. Although I guess ideally we wouldn't need to do that since the change is needed for Krylov methods (at least that's what I understood).

Ultimately, I think we should also add a test that catches this in the future.

Related to https://github.com/CliMA/Oceananigans.jl/issues/4451#issuecomment-2903837583

Closes https://github.com/CliMA/Oceananigans.jl/issues/4603


tomchor:
cc @glwagner

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
```diff
@@ -101,14 +103,14 @@ end
     Δt = clock.last_stage_Δt
     Δt = ifelse(isinf(Δt), 0, Δt)
 
-    ūⁿ⁺¹ = getbc(bc, l, m, grid, clock, model_fields)
-    uᵢⁿ     = @inbounds getindex(u, boundary_secret_storage_indices...)
+    ūⁿ⁺¹    = getbc(bc, l, m, grid, clock, model_fields)
+    uᵢⁿ     = ifelse(clock.iteration == 0, ūⁿ⁺¹, @inbounds getindex(u, boundary_secret_storage_indices...))
```
tomchor, line 107: @jagoosw this is the fix for the initial condition. It simply overwrites the first iteration to get the external BC, since in the first iteration the secret storage isn't set yet.

tomchor:
> The left boundary for a face field is part of the "interior", I'm not sure it is actually every necessary to step it so the best solution would probably be to work out how to do that.

What do you mean by interior here? The right boundary for a face field also appears in `interior()`.

> I looked through the file changes but I'm not seeing what the fix is here?

I linked you in a code comment pointing to the fix. Let me know if that's not clear.

> Also, just to let you know the mwe did work for me with the initialisation in my comment before with `ConjugateGradientPoissonSolver`.

Can you try with the MWE here? I started a fresh Manifest with the current `main` and it produces NaNs.

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
jagoosw: I see that makes sense!

glwagner: Should you be using regularize boundary condition to obtain this information?

glwagner: I've seen these kinds of designs go badly in other context (mostly with ClimaAtmos) so I think we should endeavor to use a different design if at all possible, which does not require a running simulation for initialization

tomchor: I tested a few different configurations, and for every script I tested, it made no difference if we used a "secret storage" for the left boundary or not. That is, things work equally well whether we treat the left boundary differently from the right boundary, or treat both of them the same.

So with that information, we may also just remove this asymmetry between sides, simplify things, and treat the left boundary the same as the right.

glwagner: i was just referring to the concept of something being "secret"

jagoosw:
I ran the MWE, and it does not work with the removal of the "secret" boundary storage. If I revert to the commit just before you changed that, then the MWE works as is.

jagoosw:
Oh wait maybe it works with the latest commit but I definitely do not think it is correct and it will be wrong in some scenarios

tomchor:
> Oh wait maybe it works with the latest commit but I definitely do not think it is correct and it will be wrong in some scenarios e.g. when the inflow timescale is not zero.

I'm happy to revert that! But for the record, I tried to break it in many different configurations and every single time there was no visible difference between the original scheme, and the one which ignores the secret indices. For example:

https://github.com/user-attachments/assets/2009a808-1db7-48d6-9093-6db3538b4cd0

https://github.com/user-attachments/assets/d6f19944-30ab-4cd7-b1ee-ba08eb3928e7

(tbf, there is one visible difference which is the first time-step.)

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
tomchor: I changed the name to `previous_timestep_storage_indices`, which I believe to be more accurate.

tomchor:
I think this PR is ready for review. This makes simulations with `PerturbationAdvection` OBC run as well as they did before https://github.com/CliMA/Oceananigans.jl/pull/4041. The key to make that happen is to revert `precondition!` to what it was before. In addition to that, this PR also:

1. Makes improvements to `PerturbationAdvection`, including better initialization.
2. Adds several tests to `test_condjugate_gradient_poisson_solver.jl`, including one that tests if the simulation runs with IBG + `PerturbationAdvection`.
3. Cleans up and improves relevant validation scripts

@glwagner @jagoosw I didn't change how the previous timestep is stored for left-side BCs.

src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl
glwagner: What are the indices used for?

tomchor: They point to the location in `model.velocities.u` (and v, w) that store the previous time step. In particular @jagoosw uses `i=0`.

test/test_conjugate_gradient_poisson_solver.jl
```diff
@@ -132,11 +132,7 @@ function test_conjugate_gradient_with_immersed_boundary_grid_and_open_boundaries
         @info "Testing CG solver with ImmersedBoundaryGrid using $preconditioner_name..."
         seed!(198)  # For reproducible results
 
-        # Create immersed boundary grid
-        flat_bottom(x) = 1
-        grid = ImmersedBoundaryGrid(grid_base, PartialCellBottom(flat_bottom))
-
-        # Test that CG solver can be created for immersed boundary grid with specified preconditioner
+        grid = ImmersedBoundaryGrid(grid_base, PartialCellBottom(1.0))
```
glwagner: can we use the naming scheme here that's conventional to Oceananigans, eg `underlying_grid` rather than "grid_base"?

tomchor: Done

src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -127,9 +127,13 @@ const FFTBasedPreconditioner = Union{FFTBasedPoissonSolver, FourierTridiagonalPo
 
 function precondition!(p, preconditioner::FFTBasedPreconditioner, r, args...)
     compute_preconditioner_rhs!(preconditioner, r)
-    shift = - sqrt(eps(eltype(r))) # to make the operator strictly negative definite
-    solve!(p, preconditioner, preconditioner.storage, shift)
-    p .*= -1
+
+    solve!(p, preconditioner)
```
xkykai: After checking the positive definiteness of the `FFTBasedPoissonSolver`, I am confident that the preconditioner is now positive definite and it solves $(\nabla^2 + m) \phi = b$ in this form, rather than a negative Laplacian. Here is a test script that will confirm this 
```julia
using Oceananigans
using Oceananigans.BoundaryConditions: fill_halo_regions!
using Oceananigans.Solvers: FourierTridiagonalPoissonSolver, FFTBasedPoissonSolver, solve!
using Oceananigans.Operators
using Oceananigans.Utils: launch!
using Statistics
using Random
using Oceananigans.Architectures: architecture

arch = CPU()
Nx = Nz = 8
topology = (Periodic, Flat, Bounded)

rng = Xoshiro(123)
x = z = (0, 1)

grid  = RectilinearGrid(arch; x, z, topology, size=(Nx, Nz))

lhs = CenterField(grid)

solver = FFTBasedPoissonSolver(grid)

rhs = CenterField(grid)
set!(rhs, (x, z) -> rand())
rhs .-= mean(rhs)

solver.storage .= interior(rhs) # Copy rhs to storage

solve!(lhs, solver)

@kernel function laplacian!(∇²ϕ, grid, ϕ)
    @inbounds ∇²ϕ[i, j, k] = ∇²ᶜᶜᶜ(i, j, k, grid, ϕ)
end

function compute_laplacian!(∇²ϕ, ϕ)
    grid = ϕ.grid
    arch = architecture(grid)
    fill_halo_regions!(ϕ)
    launch!(arch, grid, :xyz, laplacian!, ∇²ϕ, grid, ϕ)
    return nothing
end

rhs_∇² = CenterField(grid)
compute_laplacian!(rhs_∇², lhs)

@info sum((interior(rhs_∇²) .≈ interior(rhs))[:, 1, :]) == Nx * Nz
```
which gives `true` as an output.

However, as discussed in https://github.com/CliMA/Oceananigans.jl/pull/4554, I do not support reverting  `subtract_and_mask!` back into during preconditioning, as it really should be done on the conjugate gradient level. This PR is awaiting merging approval once the API is finalized. Let me know what would be the best/easiest course of action for you. Should we do this reverting in a different PR? I am happy to do that and you can then merge that branch back into this PR.

xkykai (changes requested):

src/Solvers/conjugate_gradient_poisson_solver.jl
tomchor: Have you tried that the MWE at the top comment in this PR works on #4554?

If it does then, after #4554 is merged, the changes here to the preconditioner become obsolete, no? If so, I can close this PR without merging. I think we key thing here (at least for me) is to make sure that open boundary with immersed boundaries are working at least as well as they were before 
#4041.

xkykai: I have tested the above MWE with https://github.com/CliMA/Oceananigans.jl/pull/4554

Note that https://github.com/CliMA/Oceananigans.jl/pull/4554 the preconditioner remains the way it is like in `main`:

```julia
function precondition!(p, preconditioner::FFTBasedPreconditioner, r, args...)
    compute_preconditioner_rhs!(preconditioner, r)
    shift = - sqrt(eps(eltype(r))) # to make the operator strictly negative definite
    solve!(p, preconditioner, preconditioner.storage, shift)
    p .*= -1

    return p
end
```

Using https://github.com/CliMA/Oceananigans.jl/pull/4554 the above MWE works, with  the `ConjugateGradientPoissonSolver` metrics as

```julia
ConjugateGradientPoissonSolver:
├── grid: 8×1×8 ImmersedBoundaryGrid{Float64, Bounded, Flat, Bounded} on CPU with 3×0×3 halo
└── conjugate_gradient_solver: ConjugateGradientSolver
    ├── maxiter: 10
    ├── reltol: 1.49012e-8
    ├── abstol: 1.49012e-8
    ├── preconditioner: FFTBasedPoissonSolver
    └── iteration: 10
```

Just to be explicit here's the open boundary conditions summary:
```julia
├── west: OpenBoundaryCondition{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}: 1.0
├── east: OpenBoundaryCondition{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}: 1.0
├── south: Nothing
├── north: Nothing
├── top: FluxBoundaryCondition: Nothing
```

Thus with the fix in https://github.com/CliMA/Oceananigans.jl/pull/4554 we don't need further alterations for the MWE to work (even though we should still make the required changes)

tomchor: Sorry, I got confused with so many PRs! But #4554 doesn't fix the issue that the mwe at the top here fixes because after that MWE runs, the velocities start blowing up:

```julia
julia> model.velocities.u
9×1×8 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 8×1×8 ImmersedBoundaryGrid{Float64, Bounded, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Open{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}, east: Open{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 17×1×16 OffsetArray(::Array{Float64, 3}, -3:13, 1:1, -3:12) with eltype Float64 with indices -3:13×1:1×-3:12
    └── max=2.11106e13, min=-2.11106e13, mean=1.00442
```

test/runtests.jl
```diff
@@ -75,7 +75,6 @@ CUDA.allowscalar() do
         @testset "Poisson Solvers 2" begin
             include("test_poisson_solvers_stretched_grids.jl")
             include("test_conjugate_gradient_poisson_solver.jl")
-            include("test_poisson_solvers_immersed_boundaries.jl")
```
tomchor, line 78: These were moved to `test_conjugate_poisson_solver.jl`, so that all tests for the CGSolver are in one place

test/test_conjugate_gradient_poisson_solver.jl
```diff
@@ -1,27 +1,304 @@
...
+
+function test_conjugate_gradient_basic_functionality(grid, preconditioner)
+    preconditioner_name = typeof(preconditioner).name.wrapper
+    @info "  Unit testing ConjugateGradientPoissonSolver with $preconditioner_name..."
+
+    # Test the preconditioner constructor
+    solver = ConjugateGradientPoissonSolver(grid, preconditioner=preconditioner)
+    pressure = CenterField(grid)
+    solve!(pressure, solver.conjugate_gradient_solver, solver.right_hand_side)
+end
```
ali-ramadhan: I think this test doesn't have any `@test` statements? If we're just testing that the basic functionality works, i.e. that we can call `solve!` without erroring, then we can just test that the pressure contains at least one non-zero element? Or if it's supposed to be all zeros, we can just test that `pressure` is still a `Field` afterwards.

tomchor: Ah, good catch! I'll fix that soon

tomchor: Fixed

xkykai (approved):

PR #4542, opened 2025-05-23
Uncomment test block in test_dynamics.jl

glwagner:
I think this was accidentally commented, but I'm not sure.


navidcy:
Looks like accidentally commented.

But also

https://github.com/CliMA/Oceananigans.jl/blob/3ff5c2a14a78d7fc3e7e0bf6f128594a070ce36c/test/test_dynamics.jl#L720

errors:

https://buildkite.com/clima/oceananigans/builds/23685#0196ff19-4dd9-4b29-a330-3bda0486f9f7/18-1136

PR #4546, opened 2025-05-26
Ensure constancy preservation

simone-silvestri:
At the moment, a nonlinear free surface conserves global tracer (except for on a tripolar grid which I ll address after this PR), but does not conserve a uniform tracer.

This PR makes sure that an initially uniform tracer remains uniform by stepping the grid with the same timestepper used for tracers (density). 
I have also added a tracer constancy preservation test in the test suite.

After this problem is solved, we have to deal with the issue of reconciliation of the grid with the free surface evolved in the barotropic step which at the moment are different: i.e. the grid is evolved with the large baroclinic step to ensure consistency with the tracer, while the free surface is evolved either implicitly or split explicitly but with another method that will make the two formulations diverge slightly.

### DIFF
--- a/src/Grids/vertical_discretization.jl
+++ b/src/Grids/vertical_discretization.jl
@@ -35,12 +35,13 @@ const AbstractStaticGrid  = AbstractUnderlyingGrid{<:Any, <:Any, <:Any, <:Any, <
 
 coordinate_summary(topo, z::StaticVerticalDiscretization, name) = coordinate_summary(topo, z.Δᵃᵃᶜ, name)
 
-struct MutableVerticalDiscretization{C, D, E, F, H, CC, FC, CF, FF} <: AbstractVerticalCoordinate
+struct MutableVerticalDiscretization{C, D, E, F, H, U, CC, FC, CF, FF} <: AbstractVerticalCoordinate
     cᵃᵃᶠ :: C
     cᵃᵃᶜ :: D
     Δᵃᵃᶠ :: E
     Δᵃᵃᶜ :: F
       ηⁿ :: H
+     δUⁿ :: U
     σᶜᶜⁿ :: CC
     σᶠᶜⁿ :: FC
     σᶜᶠⁿ :: CF
@@ -68,7 +69,7 @@ or an `AbstractArray`. A `MutableVerticalDiscretization` defines a vertical coor
 following certain rules. Examples of `MutableVerticalDiscretization`s are free-surface following coordinates,
 or sigma coordinates.
 """
-MutableVerticalDiscretization(r) = MutableVerticalDiscretization(r, r, (nothing for i in 1:9)...)
+MutableVerticalDiscretization(r) = MutableVerticalDiscretization(r, r, (nothing for i in 1:10)...)
 
 coordinate_summary(::Bounded, z::RegularMutableVerticalDiscretization, name) =
     @sprintf("regularly spaced with mutable Δr=%s", prettysummary(z.Δᵃᵃᶜ))
@@ -120,6 +121,7 @@ function generate_coordinate(FT, topo, size, halo, coordinate::MutableVerticalDi
     σᶜᶠⁿ = new_data(FT, arch, (Center, Face,   Nothing), args...)
     σᶠᶠⁿ = new_data(FT, arch, (Face,   Face,   Nothing), args...)
     ηⁿ   = new_data(FT, arch, (Center, Center, Nothing), args...)
+    δUⁿ  = new_data(FT, arch, (Center, Center, Nothing), args...) 
     ∂t_σ = new_data(FT, arch, (Center, Center, Nothing), args...)
 
     # Fill all the scalings with one for now (i.e. z == r)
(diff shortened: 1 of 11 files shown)


src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
```diff
@@ -129,30 +211,4 @@ end
...
-end
-
-@kernel function _unscale_tracer!(tracer, grid, ::Val{Hz}, ::Val{Nz}) where {Hz, Nz}
-    i, j = @index(Global, NTuple)
-
-    @unroll for k in -Hz+1:Nz+Hz
-        tracer[i, j, k] /= σⁿ(i, j, k, grid, Center(), Center(), Center())
-    end
-end
+    ℑyᵃᶠᵃ(i, j, k, grid, buoyancy_perturbationᶜᶜᶜ, buoyancy.formulation, model_fields) * ∂y_z(i, j, k, grid)
```
simone-silvestri, line 214:
```suggestion
    ℑyᵃᶠᵃ(i, j, k, grid, buoyancy_perturbationᶜᶜᶜ, buoyancy.formulation, model_fields) * ∂y_z(i, j, k, grid)
    
```

src/Grids/vertical_discretization.jl
```diff
@@ -68,7 +69,7 @@ or an `AbstractArray`. A `MutableVerticalDiscretization` defines a vertical coor
 following certain rules. Examples of `MutableVerticalDiscretization`s are free-surface following coordinates,
 or sigma coordinates.
 """
-MutableVerticalDiscretization(r) = MutableVerticalDiscretization(r, r, (nothing for i in 1:9)...)
+MutableVerticalDiscretization(r) = MutableVerticalDiscretization(r, r, (nothing for i in 1:10)...)
```
navidcy, line 72: this give `r` both for face and for centers, right? how does this work?

simone-silvestri: This is just a constructor for the user interface. The coordinate is populated when constructing the grid (this vertical coordinate also needs horizontal information).
So that users can do
```julia
z = MutableVerticalDiscretization(r_faces)
grid = RectilinearGrid(; z, size= ... , )
```

src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
```diff
@@ -12,55 +13,117 @@ barotropic_velocities(free_surface::SplitExplicitFreeSurface) = free_surface.bar
 barotropic_velocities(free_surface) = nothing, nothing
 
 # Fallback
-update_grid!(model, grid, ztype; parameters) = nothing
+ab2_step_grid!(grid, model, ztype, Δt, χ) = nothing
 
-function update_grid!(model, grid::MutableGridOfSomeKind, ::ZStar; parameters = :xy)
+function ab2_step_grid!(grid::MutableGridOfSomeKind, model, ::ZStar, Δt, χ)
```
navidcy, line 18: model knows the time stepper so why `ab2_step_grid!` and  not just `step_grid!` and dispatch over the time stepper?

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/step_split_explicit_free_surface.jl
```diff
@@ -134,6 +134,15 @@ function step_free_surface!(free_surface::SplitExplicitFreeSurface, model, baroc
     # Wait for setup step to finish
     wait_free_surface_communication!(free_surface, model, architecture(free_surface_grid))
 
+    barotropic_timestepper = free_surface.timestepper
+    baroclinic_timestepper = model.timestepper
+
+    stage = model.clock.stage
+
+    # Reset all the averaged fields to zero and possibly reset also the free surface state
+    # for the last stage of the split RK3 timestepping scheme
```
simone-silvestri, line 143:
```suggestion
    # Reset the filtered fields and the barotropic timestepper to zero. 
    # In case of an RK3 timestepper, reset also the free surface state for the last stage.
```

src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
navidcy: Perhaps doing so it's cleaner? Not sure. I'm happy either way. Wanted to see if there was some other reason I couldn't see myself.

test/test_zstar_coordinate.jl
```diff
@@ -66,42 +69,6 @@ end
 const C = Center
 const F = Face
 
-@testset "ZStar coordinate scaling tests" begin
```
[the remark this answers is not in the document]
simone-silvestri: It assumes that the free surface computed by the barotropic solver is the same one that modifies the grid, while this assumption has been broken in this PR to ensure uniformity of the field. For this reason, this testset is now wrong.

PR #4548, opened 2025-05-27
Correct diffusive dissipation for tuple of closures

simone-silvestri:
the `for` loop does not compile on the GPU. This PR follows the same implementation we have for closure tuples in `TurbulenceClosures`. I have also added a test

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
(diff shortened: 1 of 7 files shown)


navidcy (approved):
Out of curiosity, why we need to have explicit methods for 1, 2, 3-tuples and then use `4:end` for anything longer? Wouldn't explicit methods for 1, 2-tuples + methods for `3:end` work as well?

simone-silvestri:
From a results point of view, yes, from a generated code point of view, I don't think so. Being explicit might help the inferability of a method.

navidcy:
Gotcha!

Issue #4549, opened 2025-05-27
Problems with new output readers


simone-silvestri:
Using the new `find_time_index` in the `validation/implicit_dissipation/variance_dissipation.jl` file results in `FieldTimeSeries` not finding the corresponding snapshot in the jld2 file.

It might be a precision error connected to saving on Float32 or using very small timesteps? 

I ll investigate a bit and open a PR to fix this issue

PR #4550, opened 2025-05-27
Fix indexing issues for `FieldTimeSeries`

simone-silvestri:
I think this is the formulation we want because it is the absolute tolerance relative to the dt used.
In main, if dt is too small, multiplying the `ϵ` times dt might result in an absolute tolerance lower than machine precision.

I am not sure what tests were performed on #4505 to test that the new formulation worked but the previous not did not, so I will try to produce some tests with different precisions and dt size to make sure this is the tolerance we want

### DIFF
--- a/src/OutputReaders/set_field_time_series.jl
+++ b/src/OutputReaders/set_field_time_series.jl
@@ -10,7 +10,7 @@ iterations_from_file(file) = parse.(Int, keys(file["timeseries/t"]))
 function find_time_index(time::Number, file_times, Δt)
     # Accommodate round-off discrepancies between the FTS times and file times
     # (see https://github.com/CliMA/Oceananigans.jl/pull/4505)
-    ϵ = 100 * eps(eltype(file_times))
+    ϵ = 100 * eps(Δt)
     return findfirst(t -> isapprox(t, time, atol=ϵ*Δt), file_times)
 end
 
--- a/src/OutputReaders/set_field_time_series.jl
+++ b/src/OutputReaders/set_field_time_series.jl
@@ -10,8 +10,8 @@ iterations_from_file(file) = parse.(Int, keys(file["timeseries/t"]))
 function find_time_index(time::Number, file_times, Δt)
     # Accommodate round-off discrepancies between the FTS times and file times
     # (see https://github.com/CliMA/Oceananigans.jl/pull/4505)
-    ϵ = 100 * eps(Δt)
-    return findfirst(t -> isapprox(t, time, atol=ϵ*Δt), file_times)
+    ϵ = sqrt(eps(Δt))
+    return findfirst(t -> isapprox(t, time; rtol=ϵ), file_times)
 end
 
 find_time_index(time::AbstractTime, file_times, Δt) = findfirst(t -> t == time, file_times)


src/OutputReaders/set_field_time_series.jl
```diff
@@ -10,8 +10,8 @@ iterations_from_file(file) = parse.(Int, keys(file["timeseries/t"]))
 function find_time_index(time::Number, file_times, Δt)
     # Accommodate round-off discrepancies between the FTS times and file times
     # (see https://github.com/CliMA/Oceananigans.jl/pull/4505)
-    ϵ = 100 * eps(eltype(file_times))
-    return findfirst(t -> isapprox(t, time, atol=ϵ*Δt), file_times)
+    ϵ = sqrt(eps(Δt))
```
navidcy, line 13: why sqrt?

navidcy (approved):
did it fix the issues?

simone-silvestri:
Right, we should add a test for an output writer that was failing before #4505 which is fixed with #4505 to see if this PR does not break anything.

@loisbaker do you have such a case?

src/OutputReaders/set_field_time_series.jl
simone-silvestri: That is the typical value used as a reltol. I am still not sure this is the correct limit, we probably need some tests. This PR changes the absolute tolerance depending on the `Δt` by increasing a bit the absolute tolerance. 
```julia
julia> for n in (-3:3)
                Δt = 10.0^n
                em = 100 * Δt * eps(Float64)
                ep = sqrt(eps(Δt))
                @info @sprintf("Δt: %.2e, on main: %.2e, on this PR: %.2e", Δt, em, ep)
                end
[ Info: Δt: 1.00e-03, on main: 2.22e-17, on this PR: 4.66e-10
[ Info: Δt: 1.00e-02, on main: 2.22e-16, on this PR: 1.32e-09
[ Info: Δt: 1.00e-01, on main: 2.22e-15, on this PR: 3.73e-09
[ Info: Δt: 1.00e+00, on main: 2.22e-14, on this PR: 1.49e-08
[ Info: Δt: 1.00e+01, on main: 2.22e-13, on this PR: 4.21e-08
[ Info: Δt: 1.00e+02, on main: 2.22e-12, on this PR: 1.19e-07
[ Info: Δt: 1.00e+03, on main: 2.22e-11, on this PR: 3.37e-07
```

simone-silvestri: I think the problem is mostly in the small timestep range where the absolute tolerance becomes extremely low on main. For large timesteps a tolerance of order 1e-7 is fine

simone-silvestri: the failing case in issue #4549 has a time step size of 0.002 seconds and a saving time of 0.02 seconds, so in the tens of milliseconds

loisbaker:
> I think the cause of the issue has been correctly identified as such: using `ϵ * Δt` as an absolute tolerance is incorrect because the tolerance dips below machine precision if `Δt` is small, failing to find the correct timestep. I also think that using an absolute tolerance is incorrect because the order of magnitude of the times might change significantly. A relative tolerance is better suited in this case. I am not sure what test case caused the change in #4505. I guess it was trying to fix issues encountered with large timesteps. I am inclined to merge like this since I have added a test for both large and small `Δt`, and with this formulation of the error, they pass for all timestep sizes tested (in the order of 1e-4 to 1e4).

The problem that I encountered with relative tolerance here occurred only at `t=0`, when (for example) `t=1e-16` wasn't recognised as approximately 0 with the default relative tolerance:
```
julia> t = 1e-16
1.0e-16

julia> isapprox(t,0)
false
```

simone-silvestri:
thanks! Do you have an example of a case where the `FieldTimeSeries` was broken so I can include it in the test suite to make sure everything passes smoothly?

simone-silvestri:
However, also in this case I have added from #4077, 
```julia
grid = RectilinearGrid(size=(2, 2, 2), extent=(1, 1, 1))
times = 0:0.1:3
filename = "test_field_time_series_pickup.jld2"
f_tmp = Field{Center,Center,Center}(grid) 
f = FieldTimeSeries{Center, Center, Center}(grid, times; backend=OnDisk(), path=filename, name="f")

for (it, time) in enumerate(f.times)
       set!(f_tmp,   30)
       set!(f,f_tmp, it)
end

N_in_mem = 5
f_fts = FieldTimeSeries(filename, "f"; backend = InMemory(N_in_mem))
Nt = length(f_fts.times)

for t in eachindex(times)
       @test all(interior(f_fts[t]) .== 30)
end
```
the initial time is correctly `0`. What would be the case where the saved time is `1e-16` instead of `0`?

simone-silvestri:
Ok, I see. Is this something we want to support, though? This example is _forcing_ an error. 
I am wondering if we need to cover for cases where `t = 1e-16` instead of `t = 0` because it might happen, or it would never happen unless we force it like in the above?

simone-silvestri:
Ok I see, I can add this example

Issue #4553, opened 2025-05-27
`ConjugateGradientPoissonSolver` with FFT-based preconditioner does not maintain zero mean pressure


xkykai:
With @Yixiao-Zhang we noted that the `ConjugateGradientPoissonSolver` with the FFT-based preconditioner can blow up as the FFT-preconditioned solution does not have zero mean pressure as it is not immersed boundary-aware. Here is a MWE:

```julia
using Oceananigans
using Oceananigans.Models.NonhydrostaticModels: ConjugateGradientPoissonSolver, FFTBasedPoissonSolver
using Oceananigans.Grids: with_number_type
using Printf
using Statistics

const Nx = 400
const Nz = 200
const Lz = 4.0

const Δx = Lz / Nz
const Lx = Nx * Δx

const Rₑ = 30.0

const ν = 1.0
const L₀ = 1.0
const U₀ = Rₑ * ν / L₀

const Δt = 0.4 * Δx^2 / ν

const r₀ = L₀ / 2
const x₀ = 2.0 * L₀
const z₀ = 0.0

const γ = 1.0

initial_u(x, z) = 1e-7 * rand() + U₀
c_tendency(x, z, t, c) = - γ / Δt * (c - sin(40π * z / Lz)) * exp(-γ * x/Δx)
u_tendency(x, z, t, u) = - γ / Δt * (u - U₀) * exp(-γ * x/Δx)

underlying_grid = RectilinearGrid(
    GPU(),
    size = (Nx, Nz),
    x = (0.0, Lx),
    z = (- Lz / 2, Lz / 2),
    topology = (Periodic, Flat, Bounded),
    halo = (4, 4),
)

function is_obstacle(x, z)
    return (x - x₀)^2 + (z - z₀)^2 < r₀^2
end

grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBoundary(is_obstacle))
closure = ScalarDiffusivity(ν = ν)

c_forcing = Forcing(c_tendency, field_dependencies = (:c, ))
u_forcing = Forcing(u_tendency, field_dependencies = (:u, ))

# reduced_precision_grid = with_number_type(Float32, underlying_grid)
# preconditioner = FFTBasedPoissonSolver(reduced_precision_grid)
preconditioner = FFTBasedPoissonSolver(underlying_grid)
# preconditioner = nothing
pressure_solver = ConjugateGradientPoissonSolver(
    grid, maxiter=1000, preconditioner=preconditioner)

[...]

PR #4554, opened 2025-05-27
fix mean pressure in FFT preconditioner for `ConjugateGradientPoissonSolver` for immersed boundaries

xkykai:
Closes https://github.com/CliMA/Oceananigans.jl/issues/4553.

This fixes the immersed-aware mean pressure to be zero at every preconditioning step, effectively avoiding numerical blowup due to floating point error accumulation.

However this also means that without using the preconditioner, the error in mean pressure can and will accumulate and grow over time. However this is perhaps an unlikely use case as it garners no benefits vs. using the FFT preconditioner.

Here I show the results with FFT preconditioner with the test case in https://github.com/CliMA/Oceananigans.jl/issues/4553

```julia
[ Info: Iter: 0, time: 0.000e+00, Δt: 1.600e-04, Poisson iters: 11, max u: 6.646e+01, max v: 0.000e+00, max w: 3.451e+01, max d: 1.653e-05, max pressure: 1.516e+01, mean pressure: -9.107e-18
[ Info:     ... simulation initialization complete (2.146 seconds)
[ Info: Iter: 1, time: 1.600e-04, Δt: 1.600e-04, Poisson iters: 11, max u: 6.743e+01, max v: 0.000e+00, max w: 3.522e+01, max d: 2.001e-07, max pressure: 2.320e+03, mean pressure: -7.460e-16
[ Info:     ... initial time step complete (1.005 seconds).
[ Info: Iter: 2, time: 3.200e-04, Δt: 1.600e-04, Poisson iters: 10, max u: 6.774e+01, max v: 0.000e+00, max w: 3.553e+01, max d: 1.412e-06, max pressure: 2.295e+03, mean pressure: -7.460e-16
[ Info: Iter: 3, time: 4.800e-04, Δt: 1.600e-04, Poisson iters: 10, max u: 6.743e+01, max v: 0.000e+00, max w: 3.541e+01, max d: 9.215e-07, max pressure: 2.254e+03, mean pressure: -1.492e-15
[ Info: Iter: 4, time: 6.400e-04, Δt: 1.600e-04, Poisson iters: 10, max u: 6.688e+01, max v: 0.000e+00, max w: 3.506e+01, max d: 5.800e-07, max pressure: 2.240e+03, mean pressure: 0.000e+00
[ Info: Iter: 5, time: 8.000e-04, Δt: 1.600e-04, Poisson iters: 10, max u: 6.618e+01, max v: 0.000e+00, max w: 3.459e+01, max d: 4.424e-07, max pressure: 2.214e+03, mean pressure: -3.730e-16
[ Info: Iter: 6, time: 9.600e-04, Δt: 1.600e-04, Poisson iters: 11, max u: 6.541e+01, max v: 0.000e+00, max w: 3.406e+01, max d: 8.807e-08, max pressure: 2.160e+03, mean pressure: 0.000e+00
[ Info: Iter: 7, time: 1.120e-03, Δt: 1.600e-04, Poisson iters: 11, max u: 6.473e+01, max v: 0.000e+00, max w: 3.351e+01, max d: 6.340e-08, max pressure: 2.095e+03, mean pressure: 3.730e-16
[ Info: Iter: 8, time: 1.280e-03, Δt: 1.600e-04, Poisson iters: 11, max u: 6.423e+01, max v: 0.000e+00, max w: 3.298e+01, max d: 7.278e-08, max pressure: 2.026e+03, mean pressure: -1.492e-15
[ Info: Iter: 9, time: 1.440e-03, Δt: 1.600e-04, Poisson iters: 11, max u: 6.382e+01, max v: 0.000e+00, max w: 3.252e+01, max d: 7.359e-08, max pressure: 1.955e+03, mean pressure: 0.000e+00
[ Info: Simulation is stopping after running for 3.875 seconds.
[ Info: Model iteration 10 equals or exceeds stop iteration 10.
[ Info: Iter: 10, time: 1.600e-03, Δt: 1.600e-04, Poisson iters: 11, max u: 6.348e+01, max v: 0.000e+00, max w: 3.219e+01, max d: 8.471e-08, max pressure: 1.888e+03, mean pressure: -7.460e-16
```

Here we see that the mean pressure is effectively maintained at zero.

Alternatively, to solve the issue in cases where no preconditioner is used as well, we can use @Yixiao-Zhang 's proposal which is to modify https://github.com/CliMA/Oceananigans.jl/blob/f6463bd4a8280dfa2d7182ccdc2d8e438e49d384/src/Solvers/conjugate_gradient_solver.jl#L242C1-L242C55:
```julia
function update_solution_and_residuals!(x, r, q, p, α)
    xp = parent(x)
    rp = parent(r)
    qp = parent(q)
    pp = parent(p)

    xp .+= α .* pp
    rp .-= α .* qp

    mean_r = mean(r)
    grid = r.grid
    arch = architecture(grid)
    launch!(arch, grid, :xyz, subtract_and_mask!, r, grid, mean_r)

    return nothing
end
```

This touches the internals of the `ConjugateGradientSolver`, and this will fail other arbitrary problems solvable by conjugate gradient where there is no gauge invariance. We can perhaps argue that solving the pressure Poisson equation with Neumann boundary conditions is a special use case of the `ConjugateGradientSolver`. If we decide to address the issue within the conjugate gradient iteration we should perhaps think of modifying the API slightly to not do it for all arbitrary problems. Also if we take this approach we should also do the `subtract_and_mask!` on `x` as well, no reason not to.

Also for reference, here is what the MWE script does:

https://github.com/user-attachments/assets/dace32b5-3b34-4c3b-b7c1-29bc1756f7a7


xkykai:
> I only see a modification to the preconditioner for the Poisson solver. So I don't understand how this PR changes other problems. Can you explain?

I'm referring to doing this alternative approach which I have not implemented:


> Alternatively, to solve the issue in cases where no preconditioner is used as well, we can use @Yixiao-Zhang 's proposal which is to modify [`f6463bd`/src/Solvers/conjugate_gradient_solver.jl#L242C1-L242C55](https://github.com/CliMA/Oceananigans.jl/blob/f6463bd4a8280dfa2d7182ccdc2d8e438e49d384/src/Solvers/conjugate_gradient_solver.jl#L242C1-L242C55):
> 
> ```julia
> function update_solution_and_residuals!(x, r, q, p, α)
>     xp = parent(x)
>     rp = parent(r)
>     qp = parent(q)
>     pp = parent(p)
> 
>     xp .+= α .* pp
>     rp .-= α .* qp
> 
>     mean_r = mean(r)
>     grid = r.grid
>     arch = architecture(grid)
>     launch!(arch, grid, :xyz, subtract_and_mask!, r, grid, mean_r)
> 
>     return nothing
> end
> ```

but yes I agree doing something like `enforce_gauge_condition!` will be a good idea

src/Solvers/conjugate_gradient_solver.jl
```diff
@@ -3,23 +3,25 @@ using Oceananigans.Grids: interior_parent_indices
...
     linear_operator_product :: F
-    search_direction :: F
-    residual :: F
-    preconditioner :: M
-    preconditioner_product :: P
+           search_direction :: F
+                   residual :: F
+             preconditioner :: M
+     preconditioner_product :: P
+    enforce_gauge_condition :: Bool
```
xkykai, line 24: Is this a good interface to apply this gauge condition? @glwagner happy to hear your recommendations

glwagner: No, I would implement a generic interface, because we don't know what the gauge condition will be for an arbitrary `linear_operation!`. Just like we can't have "apply preconditioner" as a boolean for _all_ linear operations, we also need to have some kind of generic interface for enforcing gauge conditions that will work with any linear operation / preconditioner.

glwagner: For example, the implementation of `linear_operation!` is generic --- this can be a function or object which computes the result of applying a "linear operation" to an object.

similarly you can have `enforce_gauge_condition!` which is a function operating on something. Or you can use an object-function interface, eg a property `gauge_enforcer`, which is then involved when calling

```julia
enforce_gauge_condition!(modified_args..., gauge_enforcer::UserDefinedThing, other_args...)
```

xkykai:
I have changed `enforce_gauge_condition!` to take a function: I think this provides the greatest ease for the user to define a gauge condition they wish. The API for it would be 
```julia
enforce_gauge_condition!(x, r)
```

Would there be a case where one might need to access more internals than `x` and `r`? Else this is ready to merge methinks.

src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -56,11 +73,16 @@ Creates a `ConjugateGradientPoissonSolver` on `grid` using a `preconditioner`.
 `ConjugateGradientPoissonSolver` is iterative, and will stop when both the relative error in the
 pressure solution is smaller than `reltol` and the absolute error is smaller than `abstol`. Other
 keyword arguments are passed to `ConjugateGradientSolver`.
+The Poisson solver has a zero mean gauge condition enforced, which pins the pressure field to have a mean of zero.
+This is because the pressure field is defined only up to an arbitrary constant, and the zero mean gauge condition
+is a common choice to remove this degree of freedom.
```
tomchor: The keyword `enforce_gauge_condition` should be explained somewhere here, no?

src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -56,11 +73,16 @@ Creates a `ConjugateGradientPoissonSolver` on `grid` using a `preconditioner`.
...
+
 """
 function ConjugateGradientPoissonSolver(grid;
                                         preconditioner = DefaultPreconditioner(),
                                         reltol = sqrt(eps(grid)),
                                         abstol = sqrt(eps(grid)),
+                                        enforce_gauge_condition! = enforce_zero_mean_gauge!,
```
tomchor, line 86:
```suggestion
                                        gauge_condition = enforce_zero_mean_gauge!,
```
?

xkykai:
Is this ready to go? @glwagner @tomchor @simone-silvestri

tomchor:
Given the importance of the pressure solver, I'm thinking we should add a test for this. @xkykai what do you think?

xkykai:
> Given the importance of the pressure solver, I'm thinking we should add a test for this. @xkykai what do you think?

Sounds great! Should we adopt it from what you've written here https://github.com/CliMA/Oceananigans.jl/blob/tc/perturb-adv-obc-tweaks/test/test_conjugate_gradient_poisson_solver.jl? What else do you wish to see in the test?

tomchor:
> > Given the importance of the pressure solver, I'm thinking we should add a test for this. @xkykai what do you think?
> 
> Sounds great! Should we adopt it from what you've written here https://github.com/CliMA/Oceananigans.jl/blob/tc/perturb-adv-obc-tweaks/test/test_conjugate_gradient_poisson_solver.jl? What else do you wish to see in the test?

I just tested and those tests will fail here. I'd say let's keep it intuitive and test that the immersed-aware mean pressure is zero at every preconditioning step, which is what this PR is designed to do.

I'll add those other tests after the issue of open boundaries is fixed (probably after https://github.com/CliMA/Oceananigans.jl/pull/4582?)

tomchor:
For the record, running

```julia
using Oceananigans

Lx, Lz = 10, 3
Nx = Nz = 8

grid_base = RectilinearGrid(topology = (Bounded, Flat, Bounded), size = (Nx, Nz), x = (0, Lx), z = (0, Lz))
flat_bottom(x) = 1
grid = ImmersedBoundaryGrid(grid_base, PartialCellBottom(flat_bottom))
U = 1
inflow_timescale = 0.0
outflow_timescale = Inf
u_boundaries = FieldBoundaryConditions(
    west   = PerturbationAdvectionOpenBoundaryCondition(U; inflow_timescale, outflow_timescale),
    east   = PerturbationAdvectionOpenBoundaryCondition(U; inflow_timescale, outflow_timescale),)
    
                              boundary_conditions = (; u = u_boundaries),
                              pressure_solver = ConjugateGradientPoissonSolver(grid, maxiter=10),
                              advection = WENO(; grid, order=5)
                             )
u₀(x, z) = U + 1e-2*rand()
set!(model, u=u₀, enforce_incompressibility=true)
```

in this branch produces peaks in velocity approaching infinity, indicating that there's still an issue with open boundaries:

```julia
julia> model.velocities.u
9×1×8 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 8×1×8 ImmersedBoundaryGrid{Float64, Bounded, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Open{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}, east: Open{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 17×1×16 OffsetArray(::Array{Float64, 3}, -3:13, 1:1, -3:12) with eltype Float64 with indices -3:13×1:1×-3:12
    └── max=2.11106e13, min=-2.11106e13, mean=1.00442
```

xkykai:
> For the record, running
> 
> ```julia
> using Oceananigans
> using Oceananigans.BoundaryConditions: PerturbationAdvectionOpenBoundaryCondition
> using Oceananigans.Solvers: ConjugateGradientPoissonSolver
> 
> Lx, Lz = 10, 3
> Nx = Nz = 8
> 
> grid_base = RectilinearGrid(topology = (Bounded, Flat, Bounded), size = (Nx, Nz), x = (0, Lx), z = (0, Lz))
> flat_bottom(x) = 1
> grid = ImmersedBoundaryGrid(grid_base, PartialCellBottom(flat_bottom))
> U = 1
> inflow_timescale = 0.0
> outflow_timescale = Inf
> u_boundaries = FieldBoundaryConditions(
>     west   = PerturbationAdvectionOpenBoundaryCondition(U; inflow_timescale, outflow_timescale),
>     east   = PerturbationAdvectionOpenBoundaryCondition(U; inflow_timescale, outflow_timescale),)
>     
> model = NonhydrostaticModel(; grid,
>                               boundary_conditions = (; u = u_boundaries),
>                               pressure_solver = ConjugateGradientPoissonSolver(grid, maxiter=10),
>                               advection = WENO(; grid, order=5)
>                              )
> u₀(x, z) = U + 1e-2*rand()
> set!(model, u=u₀, enforce_incompressibility=true)
> ```
> 
> in this branch produces peaks in velocity approaching infinity, indicating that there's still an issue with open boundaries:
> 
> ```julia
> julia> model.velocities.u
> 9×1×8 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
> ├── grid: 8×1×8 ImmersedBoundaryGrid{Float64, Bounded, Flat, Bounded} on CPU with 4×0×4 halo
> ├── boundary conditions: FieldBoundaryConditions
> │   └── west: Open{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}, east: Open{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
> └── data: 17×1×16 OffsetArray(::Array{Float64, 3}, -3:13, 1:1, -3:12) with eltype Float64 with indices -3:13×1:1×-3:12
>     └── max=2.11106e13, min=-2.11106e13, mean=1.00442
> ```

Sorry! I just tested this on https://github.com/CliMA/Oceananigans.jl/pull/4563 (which should be the supposed PR to rule them all) the MWE works fine, with no explosion of velocities:

```julia
julia> model.velocities.u
9×1×8 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 8×1×8 ImmersedBoundaryGrid{Float64, Bounded, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Open{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}, east: Open{Oceananigans.BoundaryConditions.PerturbationAdvection{Float64}}, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 17×1×16 OffsetArray(::Array{Float64, 3}, -3:13, 1:1, -3:12) with eltype Float64 with indices -3:13×1:1×-3:12
    └── max=1.00987, min=0.996446, mean=1.00329
```

xkykai:
I've added simple tests and I think it is ready to go!

src/Solvers/conjugate_gradient_solver.jl
```diff
@@ -33,13 +35,17 @@ Base.summary(::ConjugateGradientSolver) = "ConjugateGradientSolver"
 # "Nothing" preconditioner
 @inline precondition!(z, ::Nothing, r, args...) = r
 
+# Default no gauge condition enforcement
+@inline no_gauge_enforcement!(x, r) = nothing
```
tomchor, line 39: Is there a reason why `no_gauge_enforcement!()` is defined here and `enforce_zero_mean_gauge!()` in on another file? If possible, we should probably put both in the same file for simplicity I think.

src/Solvers/conjugate_gradient_solver.jl
```diff
@@ -71,14 +77,24 @@ Arguments
...
 See [`solve!`](@ref) for more information about the preconditioned conjugate-gradient algorithm.
 """
 function ConjugateGradientSolver(linear_operation;
                                  template_field::AbstractField,
                                  maxiter = prod(size(template_field)),
                                  reltol = sqrt(eps(eltype(template_field.grid))),
                                  abstol = 0,
-                                 preconditioner = nothing)
+                                 preconditioner = nothing, 
+                                 enforce_gauge_condition! = no_gauge_enforcement!)
```
tomchor, line 97: I think this should follow the same notation as `ConjugateGradientPoissonSolver`, no?
```suggestion
                                 gauge_condition = no_gauge_enforcement!)
```

test/test_poisson_solvers_immersed_boundaries.jl
tomchor, line 1: Thanks for adding tests! However, I don't see anything that tests that the to gauge conditions are working properly, which I'd say is the main contribution of this PR. Am I missing something?

src/Solvers/conjugate_gradient_solver.jl
xkykai: By default `ConjugateGradientSolver` should be a generic solver that can be used to solve any arbitrary problems amenable to conjugate gradient. In many cases due to the boundary conditions there shouldn't be a gauge invariance. So the default choice for `ConjugateGradientSolver` should be no gauge enforcement. 

On the other hand, when we are solving the `ConjugateGradientPoissonSolver` for the pressure equation with homogeneous Neumann boundary conditions, there is a need to enforce a gauge. Thus `enforce_zero_mean_gauge!` should be at the `ConjugateGradientPoissonSolver` level since this is a problem-specific step.

src/Solvers/conjugate_gradient_solver.jl
xkykai: what if we do it the other way round ie: change https://github.com/CliMA/Oceananigans.jl/blob/fe97955146016c6659ce02ffa420b2f272f24df7/src/Solvers/conjugate_gradient_poisson_solver.jl#L86 to `enforce_gauge_condition!` instead? In my mind I was hoping this argument to be a verb to imply that it is a function-like object, rather than a noun such as `gauge_condition` which to me implies something more like a number. But of course I am happy to do it the other way as well.

test/test_poisson_solvers_immersed_boundaries.jl
xkykai: A proper gauge condition enforcement should result in the resultant pressure field having a zero mean. I opted to test that this way to make sure that the field it produces has zero mean.

https://github.com/CliMA/Oceananigans.jl/blob/fe97955146016c6659ce02ffa420b2f272f24df7/test/test_poisson_solvers_immersed_boundaries.jl#L73

src/Solvers/conjugate_gradient_solver.jl
tomchor: @glwagner is the expert on notation here, so I'm curious about what he has to say here. But my opinion is that if the keyword is `enforce_gauge_condition`, a user would expect it to receive a boolean. Like, "yes enforce gauge condition", or "no don't enforce a gauge condition". When you need pass a function with a specific gauge condition as a kwarg, then I think it's more intuitive for the kwarg to be a noun. Does that make sense?

Of course, that's talking from a user perspective, who most of the time won't read the code. In the code, the kwarg will "become" the function and you'll eventually call `gauge_condition(x, r)` (or `enforce_gauge_condition!(x, r)`). In that context the latter makes more sense.

Given that more people will use the function without reading its source code, I feel like it's better to use `gauge_condition`. But I also don't feel strongly about it.

Maybe `gauge_condition_function` would be a good compromise... ? idk

test/test_poisson_solvers_immersed_boundaries.jl
tomchor: I see, thanks for the clarification. But you're only testing the default gauge enforcement, right? Should we also test the non-default option?

src/Solvers/conjugate_gradient_solver.jl
tomchor: Regardless of what we choose though, both `ConjugateGradientSolver` and `ConjugateGradientPoissonSolver` shold use the same name!

src/Solvers/conjugate_gradient_solver.jl
tomchor: Got it! Thanks

src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -43,24 +43,47 @@ function compute_laplacian!(∇²ϕ, ϕ)
...
+The Poisson solver has a zero mean gauge condition enforced with `gauge_condition = enforce_zero_mean_gauge!`, which pins the pressure field to have a mean of zero.
+This is because the pressure field is defined only up to an arbitrary constant, and the zero mean gauge condition
+is a common choice to remove this degree of freedom.
+
 """
 function ConjugateGradientPoissonSolver(grid;
                                         preconditioner = DefaultPreconditioner(),
                                         reltol = sqrt(eps(grid)),
                                         abstol = sqrt(eps(grid)),
+                                        gauge_condition = enforce_zero_mean_gauge!,
```
glwagner:
```suggestion
                                        enforce_gauge_condition! = enforce_zero_mean_gauge!,
```

src/Solvers/conjugate_gradient_solver.jl
glwagner: I think `enforce_gauge_condition!` is a verb so it can be expected to be a function / callable. (the noun version, which might be a Boolean, might be something like `gauge_condition_enforcing`)

test/test_poisson_solvers_immersed_boundaries.jl
xkykai: I feel like the default is sufficient for now, since non-default ones can be user-defined so in some ways it is quite arbitrary what one wants to do with it

tomchor: Sounds good. Merge away

tomchor (approved):

PR #4558, opened 2025-05-28
Do not check if an `AbstractGrid` contains a function

simone-silvestri:
The `has_reference` function creates problems when trying to checkpoint a grid with an active map on GPU, because it is trying to access the elements of the active map array (which are `Tuple{<:Int, <:Int, <:Int}`), which lead me to find this check on the grid which is not required (grids do not contain functions since when we changed the immersed boundary to always be a `Field`).

I am not sure how this was working before though.


simone-silvestri:
I think the issue I am encountering is another one. I will look into it better and close this PR for the moment

PR #4560, opened 2025-05-29
Fix vector rotation operation

simone-silvestri:
The rotation and inverse rotation matrix seem to be reversed, leading to the issue encountered in https://github.com/CliMA/ClimaOcean.jl/issues/549

I think I will add a couple of more tests here, since this operation seems to be quite elusive and a big source of error.

### DIFF
--- a/src/Operators/vector_rotation_operators.jl
+++ b/src/Operators/vector_rotation_operators.jl
@@ -80,8 +80,8 @@ _intrinsic_ coordinate systems are equivalent. However, for other grids (e.g., f
     cosθ = Rcosθ / R
     sinθ = Rsinθ / R
 
-    uᵢ =   u * cosθ + v * sinθ
-    vᵢ = - u * sinθ + v * cosθ
+    uᵢ = u * cosθ - v * sinθ
+    vᵢ = u * sinθ + v * cosθ
 
     return uᵢ, vᵢ
 end
@@ -121,8 +121,8 @@ end
     cosθ = Rcosθ / R
     sinθ = Rsinθ / R
 
-    uₑ = u * cosθ - v * sinθ
-    vₑ = u * sinθ + v * cosθ
+    uₑ = - u * cosθ + v * sinθ
+    vₑ = + u * sinθ + v * cosθ
 
     return uₑ, vₑ
 end


simone-silvestri:
Should be ready to merge when tests pass

src/Operators/vector_rotation_operators.jl
```diff
@@ -121,8 +127,8 @@ end
     cosθ = Rcosθ / R
     sinθ = Rsinθ / R
 
-    uₑ = u * cosθ - v * sinθ
-    vₑ = u * sinθ + v * cosθ
+    uₑ = + u * cosθ + v * sinθ
+    vₑ = - u * sinθ + v * cosθ
```
siddharthabishnu, line 131:
```
    uₑ = + u * cosθ - v * sinθ
    vₑ =   u * sinθ + v * cosθ
```

navidcy: @siddharthabishnu are you suggesting that there is a mistake in what @simone-silvestri proposed?

I'll double check

simone-silvestri: I think the expression was like that before. The change in this PR is indeed changing 
```
uₑ = + u * cosθ - v * sinθ
vₑ =   u * sinθ + v * cosθ
```
to 
```
uₑ = + u * cosθ + v * sinθ
vₑ = - u * sinθ + v * cosθ
```
This is because in the `intrisic` operation we are rotating the vector to comply with the grid we pass as an argument to the function. Therefore, once the angle is computed we need to use a simple [rotation matrix](https://en.wikipedia.org/wiki/Rotation_matrix). 

In the `extrinsic` operation we are rotating _against_ the grid passed in as an argument, so we are counter-rotating and we need to use the inverse of the rotation matrix, which is:
$$R^{-1} = [cosθ, sinθ; -sinθ, cosθ]$$
complying with the changes made in this PR

navidcy (approved):

src/Operators/vector_rotation_operators.jl
```diff
@@ -80,8 +83,8 @@ _intrinsic_ coordinate systems are equivalent. However, for other grids (e.g., f
     cosθ = Rcosθ / R
     sinθ = Rsinθ / R
 
-    uᵢ =   u * cosθ + v * sinθ
-    vᵢ = - u * sinθ + v * cosθ
+    uᵢ = u * cosθ - v * sinθ
+    vᵢ = u * sinθ + v * cosθ
```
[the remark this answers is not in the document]
siddharthabishnu: @navidcy, apologies for not providing a more detailed explanation earlier. Instead of using a rotation matrix, I sketched the setup by hand and resolved the components of a vector along the new directions of the rotated frame. I realized that my suggestion holds only if the frame containing the intrinsic vector is rotated counterclockwise relative to the frame with the extrinsic vector by an angle θ. If the rotation is in the opposite direction, then @simone-silvestri’s implementation is correct.

src/Operators/vector_rotation_operators.jl
navidcy: > @simone-silvestri, can you clarify if the frame of reference containing the intrinsic vector is rotated **counterclockwise** relative to the frame containing the extrinsic vector by an angle θ? If so, my suggestion is correct else your implementation is.

the angle θ is the angle we need to rotate the intrinsic system to match the extrinsic

that is, the intrinsic system is rotated by an angle -θ compared to the extrinsic one

that's why I think it the PR is correct. But @siddharthabishnu, I admit it wasn't as clear before and I was also confused by that; I added the docstring on `rotation_angle` to clarify this.

siddharthabishnu: @simone-silvestri and @navidcy, thank you both for the detailed explanations. The docstring and the new test helped clarify my confusion. I was unsure about the relative orientation of the intrinsic and extrinsic reference frames, and I always associate a positive angle with counterclockwise rotation by default, which is what led to my misunderstanding. It's all clear now, and I agree that the PR implementation is correct and consistent with the documentation.

PR #4563, opened 2025-05-29
Symmetrizing Laplacian in order to use conjugate gradient in nonuniform grids

xkykai:
Latest Update for this the goals of this PR: This PR will close https://github.com/CliMA/Oceananigans.jl/issues/4634, close https://github.com/CliMA/Oceananigans.jl/issues/4692, an close https://github.com/CliMA/Oceananigans.jl/issues/4693 since they are all closely related.

1. The volume operators in the Laplacian is moved to the RHS i.e. the source term, to ensure that the operator is symmetric in nonuniform grids. This allows conjugate gradient to be used on nonuniform grids as well.
2. With nonuniform grids in mind, the `FourierTridiagonalPoissonSolver` will be fixed so that it can be used both as a preconditioner or on its own with immersed boundaries.
3. Issues with partial cells are also being sorted out, so one can freely use partial cells on uniform and nonuniform grids.
4. With the change in the Laplacian operator, `reltol` and `abstol` of `ConjugateGradientPoissonSolver` will now depend on grid cell volume metrics, so a heuristic default is implemented that scales with the cell sizes of the grid.

@glwagner @simone-silvestri this is ready to merge I think! Let me know of any comments/feedback/suggestions.

---

As noted by @Yixiao-Zhang, the Laplacian operator in its current form is not symmetric when grids are non-uniform. Mathematically, conjugate gradient method should be used for symmetric (semi-) positive definite operators.
This is still very much work in progress, but I will outline the mathematical details of why and how below.

## The Laplacian operator is non-symmetric on nonuniform grids.
We first note that the Laplacian operator $\nabla^2$ on the staggered grid is given by
```julia
@inline function ∇²ᶜᶜᶜ(i, j, k, grid, c)
    return V⁻¹ᶜᶜᶜ(i, j, k, grid) * (δxᶜᵃᵃ(i, j, k, grid, Ax_∂xᶠᶜᶜ, c) +
                                    δyᵃᶜᵃ(i, j, k, grid, Ay_∂yᶜᶠᶜ, c) +
                                    δzᵃᵃᶜ(i, j, k, grid, Az_∂zᶜᶜᶠ, c))
end
```
To see why it is nonsymmetric, consider the second derivative in x $\partial_x^2p$:
```math
(\partial_x^2p)^{ccc} = \frac{1}{V_i^{ccc}} \delta x_i^{caa} \left[\frac{A x_i^{fcc}}{\Delta x_i^{fcc}} \delta x_i^{faa} \left[ p \right] \right]
```
where

```math
\begin{align}
\delta x_i ^{faa} \left[ f \right] &= f[i] - f[i-1], \\
\delta x_i ^{caa} \left[ f \right] &= f[i+1] - f[i].
\end{align}
```

Plugging everything into $(\partial_x^2p)^{caa}$, and grouping terms by their locations we get
```math
(\partial_x^2p)^{ccc} = \frac{1}{V_i^{ccc}} \left( p[i+1] \frac{A x_{i+1}^{fcc}}{\Delta x_{i+1}^{fcc}} - p[i] \left( \frac{A x_{i+1}^{fcc}}{\Delta x_{i+1}^{fcc}} + \frac{A x_{i}^{fcc}}{\Delta x_{i}^{fcc}} \right) + p[i-1] \frac{A x_{i}^{fcc}}{\Delta x_{i}^{fcc}}  \right)
```

Writing this out as a matrix, we get entries of the second derivative operator as
```math
(\partial_x^2)^{ccc}_{ij} = 
\begin{cases}
-\frac{1}{V_i^{ccc}} \left( \frac{A x_{i+1}^{fcc}}{\Delta x_{i+1}^{fcc}} + \frac{A x_{i}^{fcc}}{\Delta x_{i}^{fcc}} \right), & \text{for } i = j,\\
\frac{1}{V_i^{ccc}} \frac{A x_{i+1}^{fcc}}{\Delta x_{i+1}^{fcc}} & \text{for } j = i+1,\\
\frac{1}{V_i^{ccc}} \frac{A x_{i}^{fcc}}{\Delta x_{i}^{fcc}} & \text{for } j = i-1,\\
0 & \text{otherwise}.
\end{cases}
```

From this we can see that $`(\partial_x^2)^{ccc}_{ij}`$ is nonsymmetric when grids are nonuniform. A concrete example can be shown if one tries to write out $`(\partial_x^2)^{ccc}_{12} \neq (\partial_x^2)^{ccc}_{21}`$.

## Symmetrizing the Laplacian operator for nonuniform grids
However, in nonuniform grids, we can perform a symmetrization to create a "symmetrized Laplacian" so that it is still amenable to conjugate gradient methods,
Let our original system be 
$$Lp = b$$
where $L$ is nonsymmetric. To symmetrize the system, we use the cell volume operator $V$ and solve the transformed system
```math
\begin{align}
\tilde{L} \tilde{x} &= \tilde{b}, \\
\tilde{L} &= V^{\frac{1}{2}} L V^{-\frac{1}{2}}, \\
\tilde{x} &= V^{\frac{1}{2}} x, \\
\tilde{b} &= V^{\frac{1}{2}} b, \\
V_{ij} &= V_{i}^{ccc} \delta_{ij}, \\
\end{align}
```
where $\delta_{ij}$ is the Kronecker delta. In words, $V$ is the operator that gives the volume of the cell at specific index $i$.
Now we will show that $\tilde{L}$ is symmetric.

Consider again only the $x$-direction, as it is analagous in $y$ and $z$. In matrix form, 
```math
\tilde{L}_{ij} =  \left( V^{\frac{1}{2}} (\partial_x^2) V^{-\frac{1}{2}} \right)^{ccc}_{ij} = 
\begin{cases}
-\frac{1}{V_i^{ccc}} \left( \frac{A x_{i+1}^{fcc}}{\Delta x_{i+1}^{fcc}} + \frac{A x_{i}^{fcc}}{\Delta x_{i}^{fcc}} \right), & \text{for } i = j,\\
(V_i^{ccc} V_{i+1}^{ccc})^{-\frac{1}{2}} \,  \frac{A x_{i+1}^{fcc}}{\Delta x_{i+1}^{fcc}} & \text{for } j = i+1,\\
(V_i^{ccc} V_{i-1}^{ccc})^{-\frac{1}{2}} \,  \frac{1}{V_i^{ccc}} \frac{A x_{i}^{fcc}}{\Delta x_{i}^{fcc}} & \text{for } j = i-1,\\
0 & \text{otherwise}.
\end{cases}
```
A substitution to show $`\tilde{L}_{12} = \tilde{L}_{21}`$ will illustrate that $\tilde{L}_{ij}$ is indeed symmetric.

### Where to do the volume transform?
There are 2 ways to implement this, and they differ in what goes where.
1. At the beginning of each solve, we can transform $b$ to $\tilde{b}$, then precondition (or not) and solve the system using conjugate gradient to obtain $\tilde{p}$. After `solve!` for CG is completed we then transform back by doing $p = V^{-\frac{1}{2}}\tilde{p}$. In this approach, the linear operator is given $\tilde{p}$, which it then has to first transform back to $p$ to compute the symmetrized laplacian, since in practice the symmetrized linear operator is really doing $V^{\frac{1}{2}} L p$. This is what is currently implemented in my code sketch.
5. Alternatively, the linear operator is given $p$, and it computed $\tilde{L} \tilde{p}$ in return. After the CG iteration, the $\tilde{p} \rightarrow p$ operation is performed before being passed into the linear operation. To me this seemed more clunky, which is why I have opted for option 1 for now.

### What should the API be?
In my mind there are 2 ways to expose the symmetrized Laplacian to the user: 
1. We keep it under `ConjugateGradientPoissonSolver`,
2.  or we create a new solver (something like `ConjugateGradientSymmetrizedPoissonSolver`.

I am inclined to use option 1 since using the asymmetric Laplacian when grids are nonuniform is probably a no go. If we are going with option 1, I have attempted to sketch out how it might look, dispatching the appropriate form of laplacian depending on the grid type (uniform or not)

### Enforcing gauge condition.
This is also another subtlety that came up when I am testing the code, but I haven't really looked closely and will investigate further. Currently the gauge condition is proposed to be enforced after every CG iteration. In the nonsymmetric case, this amounts to enforcing the gauge condition on $\tilde{x}$ rather than $x$. This also means that $x$ will not have zero mean once scaled. Thus we will likely have to think about enforcing gauge condition at different locations depending on which Laplacian is used.

Finally, here's a video of a test case that I was testing using a nonuniform grid in z, with the symmetrized Laplacian in CG with no preconditioner. Unfortunately at this stage using `FourierTridiagonalPoissonSolver` causes numerical instability so it still remains a work in progress.

https://github.com/user-attachments/assets/fc2e879e-88e3-41aa-a4f4-0fc7a32c3b73

I have also tried the symmetrized version on a uniform grid (with and without `FFTBasedPoissonSolver` as preconditioner) which theoretically should perform identically as the old Laplacian we have. The solutions look fine, so perhaps this idea and implementation is still not that crazy.

Sorry this is a long explanation @glwagner @simone-silvestri happy to hear your thoughts, also I'm more than happy to explain over a zoom call since this is honestly too much text.


simone-silvestri:
We can also inspect the matrix visually (for small enough problems)
```julia
julia> x = z = [-i + rand() / 2 for i in 3:-1:0]
4-element Vector{Float64}:
 -2.522319420238921
 -1.7595545526990484
 -0.9419593158192162
  0.2004281087138336

julia> grid = RectilinearGrid(; size = (3, 3), x, z , topology=(Periodic, Flat, Bounded))
3×1×3 RectilinearGrid{Float64, Periodic, Flat, Bounded} on CPU with 3×0×3 halo
├── Periodic x ∈ [-2.52232, 0.200428) variably spaced with min(Δx)=0.762765, max(Δx)=1.14239
├── Flat y
└── Bounded  z ∈ [-2.52232, 0.200428] variably spaced with min(Δz)=0.762765, max(Δz)=1.14239

julia> weights = compute_poisson_weights(grid)
([0.8007389967863805; 0.9653051420453994; 0.778338382840614;;; 0.8582990874606024; 1.0346948579546007; 0.8342882342545385;;; 1.1992610032136195; 1.4457305285525734; 1.1657117657454614], [0.5818102431531194; 0.6236329225598759; 0.871372992553168;;; 0.6236329225598759; 0.6684619713685889; 0.9340105169696403;;; 0.871372992553168; 0.9340105169696403; 1.3050490277312548], [1.0; 1.0718837110534505; 1.497692766340386;;; 0.9653051420453994; 1.0346948579546007; 1.4457305285525734;;; 0.778338382840614; 0.8342882342545385; 1.1657117657454614], [0.0; 0.0; 0.0;;; 0.0; 0.0; 0.0;;; 0.0; 0.0; 0.0], [0.0; 0.0; 0.0;;; 0.0; 0.0; 0.0;;; 0.0; 0.0; 0.0])

julia> solver  = HeptadiagonalIterativeSolver(weights, grid = grid, preconditioner_method = nothing)
Matrix-based iterative solver with:
├── Problem size: (3, 1, 3)
├── Grid: 3×1×3 RectilinearGrid{Float64, Periodic, Flat, Bounded} on CPU with 3×0×3 halo
├── Periodic x ∈ [-2.52232, 0.200428) variably spaced with min(Δx)=0.762765, max(Δx)=1.14239
├── Flat y
└── Bounded  z ∈ [-2.52232, 0.200428] variably spaced with min(Δz)=0.762765, max(Δz)=1.14239
├── Solution method: cg!
└── Preconditioner: nothing

julia> Array(solver.matrix)
9×9 Matrix{Float64}:
 -2.73135    0.965305   0.800739   0.965305   0.0        0.0        0.0        0.0        0.0
  0.965305  -2.77834    0.778338   0.0        1.03469    0.0        0.0        0.0        0.0
  0.800739   0.778338  -3.02481    0.0        0.0        1.44573    0.0        0.0        0.0
  0.965305   0.0        0.0       -3.63664    1.03469    0.858299   0.778338   0.0        0.0
  0.0        1.03469    0.0        1.03469   -3.73797    0.834288   0.0        0.834288   0.0
  0.0        0.0        1.44573    0.858299   0.834288  -4.30403    0.0        0.0        1.16571
  0.0        0.0        0.0        0.778338   0.0        0.0       -3.42333    1.44573    1.19926
  0.0        0.0        0.0        0.0        0.834288   0.0        1.44573   -3.44573    1.16571
  0.0        0.0        0.0        0.0        0.0        1.16571    1.19926    1.16571   -3.53068
```

xkykai:
> I suggest to build the matrix from the operator and checking that, indeed, the matrix is nonsymmetric in your case.

Indeed, here's a MWE to test `compute_laplacian!` which is found here https://github.com/CliMA/Oceananigans.jl/blob/6383a85b6052eaebf7ba11c9eaa20546cb841384/src/Solvers/conjugate_gradient_poisson_solver.jl#L38

```julia
using Oceananigans
using Oceananigans.Solvers: compute_laplacian!
using Oceananigans.BoundaryConditions: fill_halo_regions!
using SparseArrays

xs = collect(range(0, 1, length=6))
xs[2:end-1] .+= rand(length(xs[2:end-1])) * (1 / 6) / 5
@info xs

grid = RectilinearGrid(CPU(), Float64,
                        size = (5), 
                        halo = 4,
                        x = xs,
                        topology = (Bounded, Flat, Flat))

function initialize_matrix(template_field, linear_operator!, args...)
    Nx, Ny, Nz = size(template_field)
    A = spzeros(eltype(template_field.grid), Nx*Ny*Nz, Nx*Ny*Nz)
    make_column(f) = reshape(interior(f), Nx*Ny*Nz)

    eᵢⱼₖ = similar(template_field)
    ∇²eᵢⱼₖ = similar(template_field)

    for k = 1:Nz, j in 1:Ny, i in 1:Nx
        parent(eᵢⱼₖ) .= 0
        parent(∇²eᵢⱼₖ) .= 0
        eᵢⱼₖ[i, j, k] = 1
        fill_halo_regions!(eᵢⱼₖ)
        linear_operator!(∇²eᵢⱼₖ, eᵢⱼₖ, args...)

        A[:, Ny*Nx*(k-1) + Nx*(j-1) + i] .= make_column(∇²eᵢⱼₖ)
    end
    
    return A
end

test_field = CenterField(grid)
A = initialize_matrix(test_field, compute_laplacian!)
display(A)
```

and the output is 
```julia
[ Info: [0.0, 0.22960116301095507, 0.4189282613194224, 0.6149861209945792, 0.804943345574523, 1.0]

5×5 SparseMatrixCSC{Float64, Int64} with 13 stored entries:
 -20.793    20.793      ⋅         ⋅         ⋅
  25.2161  -52.6269   27.4108     ⋅         ⋅
    ⋅       26.4698  -52.8964   26.4266     ⋅
    ⋅         ⋅       27.2753  -54.6216   27.3463
    ⋅         ⋅         ⋅       26.6313  -26.6313
```

This shows that `compute_laplacian!` and hence https://github.com/CliMA/Oceananigans.jl/blob/6383a85b6052eaebf7ba11c9eaa20546cb841384/src/Operators/laplacian_operators.jl#L36 is nonsymmetric indeed. 
Should we fix this? How to fix it? In my mind it seemed to make sense that the Laplacian operator could be nonsymmetric on nonuniform grids, but perhaps I am missing something...

simone-silvestri:
> Should we fix this? How to fix it? In my mind it seemed to make sense that the Laplacian operator could be nonsymmetric on nonuniform grids, but perhaps I am missing something...

The Laplacian should be a symmetric operator. We can use the weights that I posted above to infer the correct discretization of the Laplacian. The discrete Laplacian should satisfy
```julia
    Axᵢ₊₁ pᵢ₊₁ + Axᵢ pᵢ₋₁ + Ayⱼ₊₁ pⱼ₊₁ + Ayⱼ pⱼ₋₁ + Azₖ₊₁ pₖ₊₁ + Azₖ pₖ₋₁
    - 2 ( Axᵢ₊₁ + Axᵢ + Ayⱼ₊₁ + Ayⱼ + Azₖ₊₁ + Azₖ ) pᵢⱼₖ  = b
```
where 
```julia
@kernel function _compute_poisson_weights(Ax, Ay, Az, grid)
    Ax[i, j, k] = Δzᵃᵃᶜ(i, j, k, grid) * Δyᶠᶜᵃ(i, j, k, grid) / Δxᶠᶜᵃ(i, j, k, grid)
    Ay[i, j, k] = Δzᵃᵃᶜ(i, j, k, grid) * Δxᶜᶠᵃ(i, j, k, grid) / Δyᶜᶠᵃ(i, j, k, grid)
    Az[i, j, k] = Δxᶜᶜᵃ(i, j, k, grid) * Δyᶜᶜᵃ(i, j, k, grid) / Δzᵃᵃᶠ(i, j, k, grid)
end
```
If you write it like this, you are guaranteed to end up with a symmetric operator

glwagner:
But precomputing weights isn't required to make the operator symmetric right? What precisely is the issue?

glwagner:
Also I think Ay in @simone-silvestri code has a typo

xkykai:
> But precomputing weights isn't required to make the operator symmetric right? What precisely is the issue?

The issue was that `V` was initially incorporated in the operator on the LHS, thus making the operator nonsymmetric when the grids are not uniform. The fix is to move `V` into the RHS so that the LHS is symmetric even when grids are nonuniform.

xkykai:
Also @glwagner @simone-silvestri I just discovered that there is an issue when `PartialCellBottom` is used in the current main branch and also this branch. I'm trying to find where did this problem appear because I did not seem to have this issue in the past...

simone-silvestri (approved):

xkykai:
I've checked the failed tests and it seems that the incompressibility test fails on the `LatitudeLongitudeGrid` here:

https://buildkite.com/clima/oceananigans/builds/24866/steps/canvas?jid=019876a7-f8b3-4d8d-bcae-e800064f5834#019876a7-f8b3-4d8d-bcae-e800064f5834/18-526

this relates to https://github.com/CliMA/Oceananigans.jl/blob/11e08a652085890727f2364db3904bb731f30910/test/test_nonhydrostatic_models.jl#L194 where in `LatitudeLongitudeGrid`, only the top and bottom layers are set to zero, all interior values are 1s. @glwagner @simone-silvestri do you know why that might be? This seems weird since this test is also not related to `ImmersedBoundaryGrid`...

simone-silvestri:
Hmmm, this might be an indication that there is some issue in the `ConjugatePoissonSolver` with the metrics on the RHS. The rectilinear grid has constant metrics while the lat-lon has variable spacings in the horizontal

glwagner:
> Hmmm, this might be an indication that there is some issue in the `ConjugatePoissonSolver` with the metrics on the RHS. The rectilinear grid has constant metrics while the lat-lon has variable spacings in the horizontal

A bug in computing the divergence maybe?

src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -81,8 +100,8 @@ is a common choice to remove this degree of freedom.
 """
 function ConjugateGradientPoissonSolver(grid;
                                         preconditioner = DefaultPreconditioner(),
-                                        reltol = sqrt(eps(grid)),
-                                        abstol = sqrt(eps(grid)),
+                                        reltol = sqrt(eps(grid)) * minimum_cell_volume(grid),
```
xkykai: The problem the test fails is because of this line, after the volume is moved to the RHS the `reltol` and `abstol` are cell volume dependent.

Once I turned this tolerance way down the test will pass. I will think of a better way to set a good default tolerance.

src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -81,8 +100,8 @@ is a common choice to remove this degree of freedom.
 """
 function ConjugateGradientPoissonSolver(grid;
                                         preconditioner = DefaultPreconditioner(),
-                                        reltol = sqrt(eps(grid)),
-                                        abstol = sqrt(eps(grid)),
+                                        reltol = min(100 * eps(grid), 100 * eps(grid) * minimum_cell_volume(grid)^2, sqrt(eps(grid))),
```
xkykai, line 103: I've found that this could be a good default for setting the tolerances for `reltol` and `abstol`. It tightens the tolerance so that https://github.com/CliMA/Oceananigans.jl/blob/f82bbf5803b38ca1e8c94064e0165a7397a08200/test/test_nonhydrostatic_models.jl#L194 can pass
It also takes into account that since the volume operator has moved to the RHS, for small cell volumes, the tolerance has to decrease correspondingly as well.

@simone-silvestri @glwagner if you think this is okay I think we are ready to merge this!

xkykai:
@glwagner @simone-silvestri is this ok to go? I can merge it!

src/Models/NonhydrostaticModels/solve_for_pressure.jl
```diff
@@ -95,7 +103,6 @@ function solve_for_pressure!(pressure, solver::ConjugateGradientPoissonSolver, 
     rhs = solver.right_hand_side
     grid = solver.grid
     arch = architecture(grid)
-    launch!(arch, grid, :xyz, _compute_source_term!, rhs, grid, args...)
+    launch!(arch, grid, :xyz, _cg_source_term!, rhs, grid, args...)
     return solve!(pressure, solver.conjugate_gradient_solver, rhs)
-end
-
+end
```
simone-silvestri:
```suggestion
end

```

src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -30,9 +31,15 @@ function Base.show(io::IO, ips::ConjugateGradientPoissonSolver)
               "    └── iteration: ", prettysummary(ips.conjugate_gradient_solver.iteration))
 end
 
+@inline function V∇²ᶜᶜᶜ(i, j, k, grid, c)
+    return δxᶜᶜᶜ(i, j, k, grid, Ax_∂xᶠᶜᶜ, c) +
+           δyᶜᶜᶜ(i, j, k, grid, Ay_∂yᶜᶠᶜ, c) +
+           δzᶜᶜᶜ(i, j, k, grid, Az_∂zᶜᶜᶠ, c)
+end
+
 @kernel function laplacian!(∇²ϕ, grid, ϕ)
```
simone-silvestri: This is not just the Laplacian anymore, right? How about

```suggestion
@kernel function _symmetric_laplacian_operator!(∇²ϕ, grid, ϕ)
```

src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -111,30 +130,31 @@ end
...
     i, j, k = @index(Global, NTuple)
-    @inbounds preconditioner_rhs[i, j, k] = Δzᶜᶜᶜ(i, j, k, grid) * rhs[i, j, k]
+    @inbounds preconditioner_rhs[i, j, k] = rhs[i, j, k] * V⁻¹ᶜᶜᶜ(i, j, k, grid)
 end
 
 function compute_preconditioner_rhs!(solver::FFTBasedPoissonSolver, rhs)
     grid = solver.grid
     arch = architecture(grid)
-    launch!(arch, grid, :xyz, fft_preconditioner_rhs!, solver.storage, rhs)
+
```
simone-silvestri:
```suggestion
```

PR #4567, opened 2025-05-30
(0.100.5) Fix outputting index-reduced fields like the free surface height

### LINKED ISSUE #4384: Unable to output η  using NetCDFWriter
I’m trying to output the free surface η using NetCDFWriter, but I encountered an error. Below is an example:
`using Oceananigans
using NCDatasets

grid = RectilinearGrid(size = (20, 40, 100), 
                       extent=(4e5, 8e5, 2e3), 
                       topology = (Periodic, Bounded, Bounded))

model = HydrostaticFreeSurfaceModel(;           grid,
                            buoyancy = BuoyancyTracer(),
                             tracers = :b)

b0=rand(Float64, 100)
using LinearAlgebra
b0=repeat(reshape(b0,(1,1,100)),20,40,1);

set!(model, b=b0)

simulation = Simulation(model, Δt=120, stop_time=24*60*60)

fields = Dict("η" => model.free_surface.η);
simulation.output_writers[:field_writerETA] =
    NetCDFWriter(model, fields, filename="eta.nc", 
    schedule=TimeInterval(60*60), array_type = Array{Float64})

run!(simulation)`

This results in the error:
ArgumentError: view indices (1:20, 1:40, 1:101) do not intersect field indices (Colon(), Colon(), 101:101)

I also tried specifying: indices=(:,:,grid.Nz+1)，but then got: ArgumentError: view indices (1:1, 1:1, 101:100) do not intersect field indices (Colon(), Colon(), 101:101)。

I attempted with_halos=true, but that led to: [ Info: Initializing simulation...
DimensionMismatch: new dimensions (26, 46, 107, 1) must be consistent with array size 1196

How can I correctly output the η variable with NetCDFWriter?

ali-ramadhan:
Resolves #4384

I believe the changes in this PR do fix the main issue of NetCDF output for the free surface height not working. But I'm not quite happy with this solution so I've opened a draft PR for now.

While there should be better support for windowed field output, I ended up having to explicitly check for the free surface =/ Been also wondering if it's almost easier and cleaner to just special case the free surface just like how we do for `LagrangianParticles`.

NetCDF tests passed locally but I will revisit this PR soon with a fresh mind.

Some potential immediate changes/improvements:
* https://github.com/CliMA/Oceananigans.jl/issues/4384#issuecomment-2815787495

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.100.4"
+version = "0.100.5"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"


test/test_netcdf_writer.jl
```diff
@@ -2616,9 +2615,8 @@ function test_netcdf_free_surface_mixed_output(arch)
     Nt = 5
     simulation = Simulation(model, Δt=0.1, stop_iteration=Nt)
 
-    # Kind of a hack because we want η to be a ReducedField.
     free_surface_outputs = (;
-        η = Average(model.free_surface.η, dims=3),
+        η = model.free_surface.η,
```
navidcy, line 2620: nice

navidcy:
One of the tests fail at

https://github.com/CliMA/Oceananigans.jl/blob/c8ff6075b2045a9cabab1a3e899a7f78c4e2a0b7/test/test_jld2_writer.jl#L169-L172

because of how the data is shaped now.

Changing to 

```Julia
 wu = file["timeseries/wu/4"][3] 
 uv = file["timeseries/uv/4"][3] 
 wT = file["timeseries/wT/4"][3] 
```

allows the test to pass. But is this intentional? If so, there are more things that needed to be tweaked later in OutputReaders because FieldTimeSeries assume that all dimensions are present, even if with size = 1.

navidcy:
Btw I've noticed that the changes in this PR affect also the JLD2 output writer; so perhaps the title needs updating?

ali-ramadhan:
Thanks for the comments @navidcy! I've fixed the test in `test_jld2_writer.jl` and also updated the PR title.

I'm revisiting this PR and decided to scrap my previous attempt. I took a different approach which I think actually works while also being simpler. For the purposes of outputs we just needed to consider both "location-based reductions" like having a `Nothing` location (which we currently do) and "index-based reduction at boundaries" like having `(:, :, Nz+1:Nz+1)` indices (this applies to the free surface height).

So I added a new util function `effective_reduced_dimensions` to be used by `NetCDFWriter` to determine output dimensions. Now we can properly output `model.free_surface.η` :tada:

ali-ramadhan:
Also hoping to tag and release v0.100.4 once this PR is merged so we can immediately use the fix.

tomchor:
> Thanks for the comments @navidcy! I've fixed the test in `test_jld2_writer.jl` and also updated the PR title.
> 
> I'm revisiting this PR and decided to scrap my previous attempt. I took a different approach which I think actually works while also being simpler. For the purposes of outputs we just needed to consider both "location-based reductions" like having a `Nothing` location (which we currently do) and "index-based reduction at boundaries" like having `(:, :, Nz+1:Nz+1)` indices (this applies to the free surface height).
> 
> So I added a new util function `effective_reduced_dimensions` to be used by `NetCDFWriter` to determine output dimensions. Now we can properly output `model.free_surface.η` 🎉

@ali-ramadhan can you explain a bit how your revised method for reduced dimensions would work in relation to this query?: https://github.com/CliMA/Oceananigans.jl/pull/4848#issuecomment-3393650076

ali-ramadhan:
@tomchor Thanks for pointing out your PR! The short answer is that this PR, as it stands, would continue to complicate field reconstruction since it drops dimensions (which is the existing behavior).

I would be on board with keeping singleton dimensions and not dropping them. As a breaking change, it might be out of scope for this PR which just aims to fix #4384? Are you thinking that not dropping singleton dimensions should happen in a separate PR, in this PR, or in #4848? I'm happy to widen the scope of this PR if it'll help with NetCDF `FieldTimeSeries` support.

tomchor:
> @tomchor Thanks for pointing out your PR! The short answer is that this PR, as it stands, would continue to complicate field reconstruction since it drops dimensions (which is the existing behavior).
> 
> I would be on board with keeping singleton dimensions and not dropping them. As a breaking change, it might be out of scope for this PR which just aims to fix #4384? Are you thinking that not dropping singleton dimensions should happen in a separate PR, in this PR, or in #4848? I'm happy to widen the scope of this PR if it'll help with NetCDF `FieldTimeSeries` support.

No reason to extend the scope here! I just wanted to understand. I'll deal with that in another PR

PR #4572, opened 2025-06-02
(0.96.32) Fix `zspacing()` for `PartialCellBottom` immersed boundaries and add tests for `ImmersedGrid`s

### LINKED ISSUE #4435: `zspacing` does not adjust for `PartialCellBottom`
I noticed that `zspacing` is blind to whether or not the grid has an immersed boundary. For example:

```julia
julia> grid = RectilinearGrid(size=(3, 3, 3), extent=(1, 1, 1));

julia> bottom(x, y) = - 1/2
bottom (generic function with 1 method)

julia> grid = ImmersedBoundaryGrid(grid, PartialCellBottom(bottom))
3×3×3 ImmersedBoundaryGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo:
├── immersed_boundary: PartialCellBottom(mean(zb)=-0.5, min(zb)=-0.5, max(zb)=-0.5, ϵ=0.2)
├── underlying_grid: 3×3×3 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.333333
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=0.333333
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=0.333333

julia> dz = Field(zspacings(grid, Center(), Center(), Center()));

julia> interior(dz)
3×3×3 view(::Array{Float64, 3}, 4:6, 4:6, 4:6) with eltype Float64:
[:, :, 1] =
 0.333333  0.333333  0.333333
 0.333333  0.333333  0.333333
 0.333333  0.333333  0.333333

[:, :, 2] =
 0.333333  0.333333  0.333333
 0.333333  0.333333  0.333333
 0.333333  0.333333  0.333333

[:, :, 3] =
 0.333333  0.333333  0.333333
 0.333333  0.333333  0.333333
 0.333333  0.333333  0.333333
```

I would expect `dz` to be zero in at `k=1` (inside the immersed boundary), 1/6 at `k=2` and 1/3 at `k=3`. @glwagner suggested that this is a bug.

tomchor:
Fixes issue #4435 where `zspacing()` was not properly adjusting for PartialCellBottom immersed boundaries. Added missing Δz method definitions for all face/center combinations for vertically-static partial cell bottom immersed boundary grids.

This PR will also add more tests for all flavors of `ImmersedGrid`s.

Closes #4435

cc @glwagner @simone-silvestri


glwagner (approved):
might be good for @simone-silvestri to take a look

src/AbstractOperations/grid_metrics.jl
```diff
@@ -2,9 +2,9 @@ using Adapt
 using Oceananigans.Operators
 using Oceananigans.Grids: AbstractGrid
 using Oceananigans.Fields: AbstractField, default_indices, location
-using Oceananigans.Operators: Δx, Δy, Δz, Ax, Δλ, Δφ, Ay, Az, volume
+using Oceananigans.Operators: Δx, Δy, Δz, Δr, Ax, Δλ, Δφ, Ay, Az, volume
```
francispoulin, line 5: What is r used for?  Is this a generalized coordinate for depth?

tomchor: Yes. It should be the same as `z` for a static vertical coordinate.

glwagner: `r` differs from `z` for a z* vertical coordinate. In that case, `r` refers to the static "reference" grid/coordinate, whereas `z` corresponds to the fluctuating physical grid+spacings.

src/ImmersedBoundaries/partial_cell_bottom.jl
```diff
@@ -1,9 +1,10 @@
 using Oceananigans.Utils: prettysummary
 using Oceananigans.Fields: fill_halo_regions!
-using Oceananigans.Grids: bottommost_active_node
+using Oceananigans.Grids: bottommost_active_node, AbstractStaticGrid
 using Printf
 
-import Oceananigans.Operators: Δrᶜᶜᶜ, Δrᶜᶜᶠ, Δrᶜᶠᶜ, Δrᶜᶠᶠ, Δrᶠᶜᶜ, Δrᶠᶜᶠ, Δrᶠᶠᶜ, Δrᶠᶠᶠ
+import Oceananigans.Operators: Δrᶜᶜᶜ, Δrᶜᶜᶠ, Δrᶜᶠᶜ, Δrᶜᶠᶠ, Δrᶠᶜᶜ, Δrᶠᶜᶠ, Δrᶠᶠᶜ, Δrᶠᶠᶠ,
```
francispoulin, line 6: It seems like a lot of work is being done to introduce the variable r, an intermediate variable for depth, and then you assign z to r.  Is this more than cosmetic?

Also, which line exactly fixes the zspacing()? I see zspacings appearing in th tests but not the function intself.

src/ImmersedBoundaries/partial_cell_bottom.jl
```diff
@@ -195,3 +196,14 @@ YFlatPCBIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Flat, <:Any, <:Any, <:Partial
...
+# Vertically-static, partial cell bottom, immergsed boundary grid
+VSPCBIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:AbstractStaticGrid, <:PartialCellBottom}
+@inline Δzᶜᶜᶜ(i, j, k, ibg::VSPCBIBG) = Δrᶜᶜᶜ(i, j, k, ibg)
+@inline Δzᶠᶜᶜ(i, j, k, ibg::VSPCBIBG) = Δrᶠᶜᶜ(i, j, k, ibg)
+@inline Δzᶜᶠᶜ(i, j, k, ibg::VSPCBIBG) = Δrᶜᶠᶜ(i, j, k, ibg)
+@inline Δzᶜᶜᶠ(i, j, k, ibg::VSPCBIBG) = Δrᶜᶜᶠ(i, j, k, ibg)
+@inline Δzᶠᶠᶜ(i, j, k, ibg::VSPCBIBG) = Δrᶠᶠᶜ(i, j, k, ibg)
+@inline Δzᶜᶠᶠ(i, j, k, ibg::VSPCBIBG) = Δrᶜᶠᶠ(i, j, k, ibg)
+@inline Δzᶠᶜᶠ(i, j, k, ibg::VSPCBIBG) = Δrᶠᶜᶠ(i, j, k, ibg)
+@inline Δzᶠᶠᶠ(i, j, k, ibg::VSPCBIBG) = Δrᶠᶠᶠ(i, j, k, ibg)
```
tomchor, line 209: @francispoulin these are the lines that actually fix dz. Basically, for a static vertical coordinate, we just call dr. This PR doesn't implement dz for cases where the vertical coordinate is changing in time.

francispoulin: Thanks for clarifying.

src/ImmersedBoundaries/partial_cell_bottom.jl
```diff
@@ -195,3 +196,14 @@ YFlatPCBIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Flat, <:Any, <:Any, <:Partial
 @inline Δrᶜᶠᶠ(i, j, k, ibg::YFlatPCBIBG) = Δrᶜᶜᶠ(i, j, k, ibg)
 @inline Δrᶠᶠᶜ(i, j, k, ibg::XFlatPCBIBG) = Δrᶜᶠᶜ(i, j, k, ibg)
 @inline Δrᶠᶠᶜ(i, j, k, ibg::YFlatPCBIBG) = Δrᶠᶜᶜ(i, j, k, ibg)
+
+# Vertically-static, partial cell bottom, immergsed boundary grid
+VSPCBIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:AbstractStaticGrid, <:PartialCellBottom}
```
simone-silvestri, line 201: what would happen here for a mutable grid?

tomchor: I haven't specifically tested, but calling `Δz` on a mutable grid should throw a "method not defined" error. Which I think is what we'd want, since we haven't defined it in that case.

simone-silvestri: I don't think you need additional `Δz` methods for PartialCellBottom. It should be covered by 
https://github.com/CliMA/Oceananigans.jl/blob/51a77cc6ab9d7f66eb84ed574709adac3c9210c9/src/ImmersedBoundaries/mutable_immersed_grid.jl#L54-L57
Can we test this also?

tomchor: I remember that without adding these extra method specific to PartialCellBottom things weren't working. But maybe I missed something. I'll test that later today and report back.

tomchor: Just confirmed. Things do not work without these added methods. I get:

```
[2025/06/05 09:53:09.625] INFO    Testing grid spacings and metrics [Float64, CPU]...
Spacings [Float64, CPU]: Test Failed at /home/tomas/repos/Oceananigans.jl2/test/test_immersed_boundary_grid.jl:176
  Expression: interior(Δz_partial) == interior(Δr_partial)
   Evaluated: [0.3333333333333333;;; 0.3333333333333333;;; 0.3333333333333333] == [0.3333333333333333;;; 0.16666666666666669;;; 0.3333333333333333]
```

i.e. `zspacings(::partialcellbg)` falls back to `zspacings(::partialcellbg.underlying_grid)`

Do you have another way of resolving this in mind?

glwagner: @tomchor it may mean that the code @simone-silvestri isn't correctly extending those methods, so you should re-read to make sure it is acting correctly. @simone-silvestri is merely saying that the code you've written is _supposedly_ redundant. So the problem is not that the original code doesn't cover your case, the problem is a bug somewhere, probably

tomchor: Sorry I misunderstood what @simone-silvestri was saying. But zspacings will work (and continue to work) for `MutableVerticalDiscretization` grids. This PR makes it work for Static grids as well. I also added tests for `MutableVerticalDiscretization` grids to make sure things continue to work smoothly. They pass locally but let's see what happens on buildkite.

tomchor: Tests are passing for MutableVerticalDiscretization grids as well! :)

PR #4575, opened 2025-06-03
Delete /paper directory

tomchor:
It seems like this directory was leftover from a while ago and it isn't needed. Can someone please confirm?


glwagner:
confirm

PR #4577, opened 2025-06-04
Update boundary conditions before pressure solve

glwagner:


tomchor:
I don't see any difference in my simulations after this change. As an example, here's a headland simulation (similar to the paper one but with the original concave bottom) without this change, using an fft solver:

https://github.com/user-attachments/assets/2a91f6b7-86b0-438b-9ac3-f1ff3656dca9

and the same simulation with this change:

https://github.com/user-attachments/assets/e06e7d58-aaf4-489f-8847-f57505f87b88

(Note that both simulations above also use the changes done in https://github.com/CliMA/Oceananigans.jl/pull/4539)

glwagner:
is there any change in the predictor velocity?

glwagner:
The difference could be small, so I would look at the pointwise difference betwween the solutions. note that because we are talking about details associated with time-dependence, the difference in results should depend on the time-step, too.

glwagner:
ok, closing this

PR #4587, opened 2025-06-07
Small code refactorings in Utils

thimotedupuch:
So I tried to not change any of the logic but made the code a little more idiomatic. For the ordered dict show function, the code contains more lines now but is more readable in my opinion.

This shouldn't create merge conflits.

### DIFF
--- a/src/Utils/multi_region_transformation.jl
+++ b/src/Utils/multi_region_transformation.jl
@@ -120,8 +120,8 @@ Base.parent(mo::MultiRegionObject) = construct_regionally(parent, mo)
     multi_region_kwargs = isnothing(findfirst(isregional, kwargs)) ? nothing : kwargs[findfirst(isregional, kwargs)]
     isnothing(multi_region_args) && isnothing(multi_region_kwargs) && return regional_func!(args...; kwargs...)
 
-    devs = isnothing(multi_region_args) ? 
-        multi_region_kwargs : multi_region_args |> devices
+    devs = isnothing(multi_region_args) ? multi_region_kwargs : multi_region_args
+    devs = devices(devs)
 
 
     for (r, dev) in enumerate(devs)
--- a/src/Utils/multi_region_transformation.jl
+++ b/src/Utils/multi_region_transformation.jl
@@ -145,8 +145,8 @@ end
     multi_region_kwargs = isnothing(findfirst(isregional, kwargs)) ? nothing : kwargs[findfirst(isregional, kwargs)]
     isnothing(multi_region_args) && isnothing(multi_region_kwargs) && return regional_func(args...; kwargs...)
 
-    devs = isnothing(multi_region_args) ? 
-        multi_region_kwargs : multi_region_args |> devices
+    devs = isnothing(multi_region_args) ? multi_region_kwargs : multi_region_args
+    devs = devices(devs)
 
 
     # Evaluate regional_func on the device of that region and collect
--- a/src/Utils/multi_region_transformation.jl
+++ b/src/Utils/multi_region_transformation.jl
@@ -120,11 +120,9 @@ Base.parent(mo::MultiRegionObject) = construct_regionally(parent, mo)
     multi_region_kwargs = isnothing(findfirst(isregional, kwargs)) ? nothing : kwargs[findfirst(isregional, kwargs)]
     isnothing(multi_region_args) && isnothing(multi_region_kwargs) && return regional_func!(args...; kwargs...)
 
-    if isnothing(multi_region_args)
-        devs = devices(multi_region_kwargs)
-    else
-        devs = devices(multi_region_args)
-    end
+    devs = isnothing(multi_region_args) ? 
+        multi_region_kwargs : multi_region_args |> devices
+
 
     for (r, dev) in enumerate(devs)
         switch_device!(dev)
@@ -147,11 +145,9 @@ end
     multi_region_kwargs = isnothing(findfirst(isregional, kwargs)) ? nothing : kwargs[findfirst(isregional, kwargs)]
     isnothing(multi_region_args) && isnothing(multi_region_kwargs) && return regional_func(args...; kwargs...)
 
-    if isnothing(multi_region_args)
-        devs = devices(multi_region_kwargs)
-    else
-        devs = devices(multi_region_args)
-    end
+    devs = isnothing(multi_region_args) ? 
+        multi_region_kwargs : multi_region_args |> devices
+
 
     # Evaluate regional_func on the device of that region and collect
     # return values
--- a/src/Utils/ordered_dict_show.jl
+++ b/src/Utils/ordered_dict_show.jl
@@ -4,15 +4,27 @@ function ordered_dict_show(dict::OrderedDict, padchar)
     name = "OrderedDict"
     N = length(dict)
 
-    if N === 0
+    if N == 0
         return "$name with no entries"
-    elseif N == 1
-        return string("$name with 1 entry:", "\n",
-                      padchar, "   └── ", dict.keys[1], " => ", summary(dict.vals[1]))
-    else
-        return string(name, " with $N entries:\n",
-                      Tuple(string(padchar, "   ├── $name => ", summary(dict[name]), "\n") for name in dict.keys[1:end-1])...,
-                            string(padchar, "   └── ", dict.keys[end], " => ", summary(dict.vals[end])))
+    else if N == 1
+        k = dict |> keys |> first
+        sum_v = dict |> values |> first |> summary
+        return "$name with 1 entry:\n$padchar   └── $k => $sum_v"
     end
+
+    keys_list = dict |> keys |> collect
+    summary_vals_list = dict |> values |> collect |> v -> summary(v)
+    lines = ["$name with $N entries:"]
+
+    for i in 1:N-1
+        k, sum_v  = keys_list[i], summary_vals_list[i]
+        push!(lines, "$padchar   ├── $k => $sum_v")
+    end
+
+    last_k, last_sum_v = last(keys_list), last(summary_vals_list)
+    push!(lines, "$padchar   └── $last_k => $last_sum_v")
+
+    return join(lines, "\n")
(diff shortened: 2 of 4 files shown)


src/Utils/multi_region_transformation.jl
```diff
@@ -120,11 +120,9 @@ Base.parent(mo::MultiRegionObject) = construct_regionally(parent, mo)
     multi_region_kwargs = isnothing(findfirst(isregional, kwargs)) ? nothing : kwargs[findfirst(isregional, kwargs)]
     isnothing(multi_region_args) && isnothing(multi_region_kwargs) && return regional_func!(args...; kwargs...)
 
-    if isnothing(multi_region_args)
-        devs = devices(multi_region_kwargs)
-    else
-        devs = devices(multi_region_args)
-    end
+    devs = isnothing(multi_region_args) ? 
+        multi_region_kwargs : multi_region_args |> devices
```
glwagner: How about:

```suggestion
    devs = isnothing(multi_region_args) ? multi_region_kwargs : multi_region_args
    devs = devices(devs)
```

I was not sure if the ternary or pipe operator has precedence but you may know

src/Utils/multi_region_transformation.jl
```diff
@@ -147,11 +145,9 @@ end
     multi_region_kwargs = isnothing(findfirst(isregional, kwargs)) ? nothing : kwargs[findfirst(isregional, kwargs)]
     isnothing(multi_region_args) && isnothing(multi_region_kwargs) && return regional_func(args...; kwargs...)
 
-    if isnothing(multi_region_args)
-        devs = devices(multi_region_kwargs)
-    else
-        devs = devices(multi_region_args)
-    end
+    devs = isnothing(multi_region_args) ? 
+        multi_region_kwargs : multi_region_args |> devices
```
glwagner: same comment as above

src/Utils/multi_region_transformation.jl
glwagner: @simone-silvestri why do we need this line?

src/Utils/output_writer_diagnostic_utils.jl
```diff
@@ -8,22 +8,16 @@ using Oceananigans: AbstractOutputWriter, AbstractDiagnostic
...
 defaultname(::AbstractOutputWriter, nelems) = Symbol(:writer, nelems+1)
 
 const DiagOrWriterDict = OrderedDict{S, <:Union{AbstractDiagnostic, AbstractOutputWriter}} where S
 
-function push!(container::DiagOrWriterDict, elem)
-    name = defaultname(elem, length(container))
-    container[name] = elem
-    return nothing
-end
+push!(container::DiagOrWriterDict, elem) = (container[defaultname(elem, length(container))] = elem; nothing)
```
glwagner: I find this a little harder to understand because of the nested calls. (But I love all the other changes on this PR!)

glwagner:
There's a typo in which `elseif` is accidentally `else if`. With that fixed hopefully tests should pass! I also suggest reverting the changes that involve `push!` -- personally, I have trouble reading single lines that combine a few different operations. But @simone-silvestri or @navidcy may feel differently. Otherwise its awesome, thank you @thimotedupuch !

src/Utils/ordered_dict_show.jl
```diff
@@ -4,15 +4,27 @@ function ordered_dict_show(dict::OrderedDict, padchar)
...
+    if N == 0
         return "$name with no entries"
-    elseif N == 1
-        return string("$name with 1 entry:", "\n",
-                      padchar, "   └── ", dict.keys[1], " => ", summary(dict.vals[1]))
-    else
-        return string(name, " with $N entries:\n",
-                      Tuple(string(padchar, "   ├── $name => ", summary(dict[name]), "\n") for name in dict.keys[1:end-1])...,
-                            string(padchar, "   └── ", dict.keys[end], " => ", summary(dict.vals[end])))
+    else if N == 1
```
glwagner:
```suggestion
    elseif N == 1
```

thimotedupuch:
Yeah my bad for the small mistakes here and there. Be free to take the small parts you want and leave the rest unchanged.

Also, are there any other parts of the Oceananigans.jl that you think could be improved ? I happen to have some free time to contribute

glwagner:
> Yeah my bad for the small mistakes here and there. Be free to take the small parts you want and leave the rest unchanged.
> 
> Also, are there any other parts of the Oceananigans.jl that you think could be improved ? I happen to have some free time to contribute

What kind of contributions would you like to make? Numerics, output, abstractions for calculus, output reading, parallelization/MPI, performance optimization... ?

thimotedupuch:
I really don't know. Maybe not too much math-related stuff, because my code would be terrible but I can help for Input / Output in general, utility functions, plotting, ...

src/Utils/output_writer_diagnostic_utils.jl
navidcy: yes, I do prefer it as it was before myself
it's no other change other than line merging, right?

src/Utils/multi_region_transformation.jl
simone-silvestri: I guess because we can launch `apply_regionally!` with no arguments and only keyword arguments (or with global arguments that do not have devices and regional keyword arguments). By default, we infer the devices we are running on from the arguments, however, in that case, we need to infer the devices from the keyword arguments.

glwagner:
One really cool contribution would be to put in proper heatmap plotting on the sphere. The way to do this for various Oceananigans grids has been largely figured out on `Imaginocean.jl`: https://github.com/navidcy/Imaginocean.jl/blob/main/src/Imaginocean.jl

(though I am sure the code can be cleaned up)

Just implementing this for `OrthogonalSphericalShellGrid` would be huge. Some GeoMakie integration might also be interesting to do different projections.

glwagner:
Maybe we can discuss on an issue or discussion?

navidcy:
That would be really cool. @taimoorsohail and @briochemc did something/were working on something along those lines and might be keen to join such a discussion.

src/Utils/multi_region_transformation.jl
glwagner: Are there important cases where it is not possible to pass arguments? It would simply the implementation to impose the requirement that one of the arguments is a multi region object, so that we do not have to check the kwargs. I feel this might also clarify use of this code as well.

src/Utils/output_writer_diagnostic_utils.jl
glwagner: for sure, this is just a refactor

src/Utils/multi_region_transformation.jl
glwagner:
```suggestion
    devs = isnothing(multi_region_args) ? multi_region_kwargs : multi_region_args
    devs = devices(devs)
```

src/Utils/output_writer_diagnostic_utils.jl
glwagner:
```suggestion
function push!(container::DiagOrWriterDict, elem)
    name = defaultname(elem, length(container))
    container[name] = elem
    return nothing
end
```

src/Utils/output_writer_diagnostic_utils.jl
```diff
@@ -19,11 +20,8 @@ function push!(container::DiagOrWriterDict, elem)
...
+
 setindex!(container::DiagOrWriterDict, newvals, inds::Integer...) = setindex!(container.vals, newvals, inds...)
 
-function push!(container::DiagOrWriterDict, elems...)
-    for elem in elems
-        push!(container, elem)
-    end
-    return nothing
-end
+push!(container::DiagOrWriterDict, elems...) = (foreach(e -> push!(container, e), elems); nothing)
```
glwagner:
```suggestion
function push!(container::DiagOrWriterDict, elems...)
    foreach(elem -> push!(container, elem), elems)
    return nothing
end
```

navidcy:
me and @taimoorsohail noticed something strange in the show method of the simulation, in particular the Callback part.

I think the refactor of `ordered_dict_show` in this PR had an effect?
Look what happens in the show method for the callbacks:

After the PR:
```Julia
using Oceananigans

model = NonhydrostaticModel(grid=RectilinearGrid(size=(1, 1, 1), extent=(1, 1, 1)))

simulation = Simulation(model, Δt=1, stop_iteration=10)
```
gives
```julia
Simulation of NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
├── Next time step: 1 second
├── Wall time per iteration: NaN days
├── Stop time: Inf days
├── Stop iteration: 10.0
├── Wall time limit: Inf
├── Minimum relative step: 0.0
├── Callbacks: OrderedDict with 4 entries:
│   ├── stop_time_exceeded => 4
│   ├── stop_iteration_exceeded => -
│   ├── wall_time_limit_exceeded => e
│   └── nan_checker => }
├── Output writers: OrderedDict with no entries
└── Diagnostics: OrderedDict with no entries
```

before the PR (eg using Oceananigans v0.96.33)

```Julia
Simulation of NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
├── Next time step: 1 second
├── Wall time per iteration: NaN days
├── Stop time: Inf days
├── Stop iteration: 10.0
├── Wall time limit: Inf
├── Minimum relative step: 0.0
├── Callbacks: OrderedDict with 4 entries:
│   ├── stop_time_exceeded => Callback of stop_time_exceeded on IterationInterval(1)
│   ├── stop_iteration_exceeded => Callback of stop_iteration_exceeded on IterationInterval(1)
│   ├── wall_time_limit_exceeded => Callback of wall_time_limit_exceeded on IterationInterval(1)
│   └── nan_checker => Callback of NaNChecker for u on IterationInterval(100)
├── Output writers: OrderedDict with no entries
└── Diagnostics: OrderedDict with no entries
```

navidcy:
Shall we revert the `ordered_dict_show` method?

PR #4588, opened 2025-06-09
Docs for zstar and nonlinear free surface

simone-silvestri:
This PR adds docs for the derivation of the generalized vertical coordinate equations and their detailed numerical implementation including the formulation we use for a non-linear free surface

### DIFF
--- a/docs/src/physics/zstar_vertical_coordinate.md
+++ b/docs/src/physics/zstar_vertical_coordinate.md
@@ -1,0 +1,5 @@
+# ZStar vertical coordinate
+
+The user can choose between a `ZCoordinate` and a `ZStar` vertical coordinate. A `ZStar` vertical coordinate conserves tracers
+and volume with the grid following the evolution of the free surface in the domain. To obtain the (discrete) equations evolved 
+in a general framework where the vertical coordinate is moving, we perform a scaling of the continuous primitive equations following ruldes to 


glwagner:
is there a preview?

navidcy:
Always, at: https://clima.github.io/OceananigansDocumentation/previews/PR4588/

At the moment there is not much changes tho

francispoulin:
> Always, at: https://clima.github.io/OceananigansDocumentation/previews/PR4588/grids/
> 
> At the moment there is not much changes tho

I could not find this page in the physics tabe but found it here, in case others want a quick link. 

https://clima.github.io/OceananigansDocumentation/previews/PR4588/physics/zstar_vertical_coordinate/

navidcy:
My bad! I put the wrong link! Fixed it now!

docs/src/physics/zstar_vertical_coordinate.md
```diff
@@ -0,0 +1,128 @@
+# Generalized vertical coordinate
+
+For `HydrostaticFreeSurfaceModel()`, the user can choose between a `ZCoordinate` and a `ZStar` vertical coordinate.
```
navidcy: given that this section is only relevant for `HydrostaticFreeSurfaceModel()` perhaps we move it a bit further down in the docs structure?

navidcy:
@simone-silvestri is this ready to review or are you still working on it? (just asking because it's a Draft PR)

simone-silvestri:
I am thinking of adding also the momentum equations in vector-invariant form today, and a small code tutorial (probably), then it will be ready for review.

navidcy:
Cool -- ping me when you are done.

simone-silvestri:
Ok, I also derived the material derivative of the horizontal velocity in the vector invariant form. All the nomenclature follows very closely the symbols we use in the source code, so it should be very easy to verify the implementation from the docs. I think this is enough for the moment. This PR should be ready to review

simone-silvestri:
I think I will put off the discretization of the momentum equations in this PR because it probably requires its own section.

simone-silvestri:
Also the non-linear free surface requires its own section, this is just the vertical coordinate transformation

glwagner:
I'd like to get rid of the "physics" section of the docs eventually. Can we discuss where this documentation should go?

Note, this is not physics, its numerics.

simone-silvestri:
Right, maybe it makes more sense to add this section in the `numerical implementation` section after large eddy simulations? We can probably add a section like large eddy simulations but more about hydrostatic simulations after which we add the vertical coordinates.

simone-silvestri:
I think this will also be complemented by a section on linear vs non linear free surface which will be similar to the `pressure decomposition` section that is in `numerical implementation`

glwagner:
I would suggest striking the "physics" section, and instead presenting model description and configuration into a single docs section. 

Then for numerics, maybe a docs section on "Discretization", where we introduce the staggered grid, second order difference, derivative, and reconstruction, reductions, upwinding and WENO, time-stepping, diffusion, solvers, and then z*.

simone-silvestri:
Maybe the generalized vertical coordinate section can appear before the "Discretization" and after the hydrostatic free surface model setup so that we refer always to the same equations with $r = z$ or $r = z^\star$

glwagner:
Or we can do away with trying to distinguish between "equations / physics" and "discretization" and come up with another design?

glwagner:
> Maybe the generalized vertical coordinate section can appear before the "Discretization" and after the hydrostatic free surface model setup so that we refer always to the same equations with r = z or r = z ⋆

Is this $r$ the same as the way we use $r$ in the code? In the code we use $r$ for the grid spacings, and it refers to the "reference" grid. But here you seem to be using the MOM notation, where $r$ refers to the "radial coordinate".

We should use identical notation for math and in the code, I think otherwise it will be easy to get _very_ confused. In the code, $z$ is the radial coordinate?

simone-silvestri:
Nono, `r` is exactly the same as in the code, it is the generalized coordinate that we transform into (specifically, in our case it could be only z or zstar). 
On purpose I tried to be as close as possible to the source code so that is would be easy to interpret and understand the implementation.

simone-silvestri:
Note that in the docs $\Delta z = \sigma \Delta r$ which is the same implementation we have here
https://github.com/CliMA/Oceananigans.jl/blob/013f275662aa22fc143ae274c9ad6f8167e028cb/src/Operators/time_variable_grid_operators.jl#L34-L48

simone-silvestri:
> Or we can do away with trying to distinguish between "equations / physics" and "discretization" and come up with another design?

Yeah, I like the idea, we can blend both together and write equations and discretizations together. We should start with grids and the C-grid framework and then move to models where we show equations and discretizations together.

glwagner:
There is already a `Grids` section, so you can add something to that docs section about `MutableVerticalDiscretization`. It'd be nice to add a graphic illustrating the grid too!

navidcy:
I move this new section within the HydrostaticFreeSurfaceModel.

It has a bit of mixture of the physics and the numerics.

navidcy:
A remark I have is that the section includes a generalized coordinate derivation/discussion (which is not HydrostaticFreeSurfaceModel-specific and can fit in a different section. But then it goes into the ZStar.
But there is nowhere a mention of what the generalized coordinate is for zstar, $r = η$? From the paper cited I deduce that it's $r = (z - η) H / (H+η)$, am I right?

simone-silvestri:
Indeed, that is correct, I have added the definition for ``r`` for the two choices at the beginning of the section

navidcy:
The very last expression for the gradient of $η$, if $r=z$ I get that the right hand side is zero, no?

For $r = z$, $σ = 1$ and therefore:

$$
\begin{align}
    \frac{\partial \eta}{\partial x} \bigg\rvert_z & = \frac{\partial \eta}{\partial x}\bigg\rvert_r - \frac{1}{\sigma} \frac{\partial \eta}{\partial r}\frac{\partial z}{\partial x} \nonumber \\
& = \frac{\partial \eta}{\partial x}\bigg\rvert_z - \underbrace{\frac{\partial \eta}{\partial z}\frac{\partial z}{\partial x}}_{\partial_x η \rvert_z} = 0
\end{align}
$$

simone-silvestri:
I am not sure you can do that, note that both $\partial \eta / \partial z$ and $\partial z / \partial x$ are equal to zero. In particular, $\partial z / \partial x = 0$ because it's the partial derivative at constant $y, r, t$, where (in this case) $r = z$

navidcy (approved):

navidcy:
@simone-silvestri I restructured the docs page a bit; gathered all discrete-related discussion at the end.
Have a look if you like. I'm happy to merge it.

navidcy:
> I'd like to get rid of the "physics" section of the docs eventually. Can we discuss where this documentation should go?
> 
> Note, this is not physics, its numerics.

Actually this is fair, not much physics discussed here. I might move the section.

navidcy:
I moved all the generalized vertical coordinates derivation + discretisation in the "Numerics" section. I only left a short discussion for the ZStar coordinate under the Physics/HydrostaticFreeSurfaceModel section.

navidcy:
So I wanted to add a schematic for $z$ and $z^\star$. The schematic would be showing lines of constant $r$ surfaces in $z$. 

To clarify, the $z^\star$ coordinate is defined:

$$
z^\star = (z - η) H / (H + η)
$$

so solving this for $z$ we get:

$$
z = z^\star (H + η) / H + η
$$

Here's a code + fig. I'm not sure if the right panel is the middle one or the right one.

```Julia
using CairoMakie

depth = 20
Lx = 1e3

x = range(-Lx/2, stop=Lx/2, length=200)

σ = Lx/14

# free surface
η = @. -((x/σ)^2 - 1) * exp(-x^2 / 2σ^2)

# bottom
slope_position = -Lx/3
slope = @. 4 * (1 + tanh(-(x - slope_position) / σ))
mountain_position = Lx/3
mountain = @. 2 * exp(-(x - mountain_position)^2 / 2σ^2)
H = @. depth - slope - mountain


fig = Figure(size=(1000, 400))

ax1 = Axis(fig[1, 1], title=L"$z$", titlesize =20)
ax2 = Axis(fig[1, 2], title=L"$z^*$ ?", titlesize = 20)
ax3 = Axis(fig[1, 3], title=L"$z^*$ ?", titlesize = 20)

for ax in (ax1, ax2, ax3)
    band!(ax, x, -H, η, color = (:blue, 0.2))
    band!(ax, x, -depth*1.1, -H, color = (:orange, 0.2))
    lines!(ax, x, η, linewidth=5, label="free surface", color=:darkblue)
    lines!(ax, x, -H, linewidth=5, label="bottom", color=:darkgrey)
end

for rel in 0:1/5:1
    z = -depth * rel * ones(size(x))
    lines!(ax1, x, z, color=:grey, linestyle=:dash)

    r = -depth * rel # r coordinate ∈ [-Lz, 0]
    z = @. r * (H + η) / H + η
    lines!(ax2, x, z, color=:grey, linestyle=:dash)

    r = @. - H * rel # r coordinate ∈ [-H(x), 0]
    z = @. r * (H + η) / H + η
    lines!(ax3, x, z, color=:grey, linestyle=:dash)
end

fig

save("z-zstar.png", fig)
```

z-zstar


From the definition of $z^\star$ I understand that it follows the bathymetry, so that's the right? But that seems like a combination of $z^\star$ and $\sigma$ coords?

navidcy:
I guess given that $z$ depth coordinates are straight lines then I need to compute $z^\star$ coordinates using straight lines of $z$ and thus this implies the correct is the middle!

navidcy:
I included this:

test

PR #4594, opened 2025-06-10
Add optional vertical Coriolis term to hydrostatic pressure integral

glwagner:
This PR adds the vertical Coriolis term, z_f_cross_U, to the hydrostatic pressure integral.

I implemented this so that it is optional. This way, the hydrostatic model can use it, but the nonhydrostatic model will not. We don't want to use this for the nonhydrostatic model because we don't always compute the hydrostatic pressure.

I did a bit of housecleaning too.

cc @francispoulin 

@simone-silvestri I would like to change the kwarg "parameters" to "kernel_parameters" everywhere in a future PR (just mentioning here since I was looking at it). 

Also the kernel parameter functions should not be defined in the same file as the kernels. They should be defined where they are used, eg for each model.


src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
```diff
@@ -3,39 +3,52 @@ using Oceananigans.ImmersedBoundaries: PartialCellBottom, ImmersedBoundaryGrid
...
+@kernel function _update_hydrostatic_pressure!(pHY′, grid, buoyancy, tracers, coriolis, velocities)
     i, j = @index(Global, NTuple)
 
-    @inbounds pHY′[i, j, grid.Nz] = - z_dot_g_bᶜᶜᶠ(i, j, grid.Nz+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, grid.Nz+1, grid)
+    dpdz⁺ = z_dot_g_bᶜᶜᶠ(i, j, grid.Nz+1, grid, buoyancy, tracers) - z_f_cross_U(i, j, grid.Nz+1, grid, coriolis, velocities)
+    @inbounds pHY′[i, j, grid.Nz] = - dpdz⁺ * Δzᶜᶜᶠ(i, j, grid.Nz+1, grid)
 
     for k in grid.Nz-1 : -1 : 1
-        @inbounds pHY′[i, j, k] = pHY′[i, j, k+1] - z_dot_g_bᶜᶜᶠ(i, j, k+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, k+1, grid)
+        dpdz⁺ = z_dot_g_bᶜᶜᶠ(i, j, k+1, grid, buoyancy, tracers) - z_f_cross_U(i, j, k+1, grid, coriolis, velocities)
```
francispoulin, line 13: This looks great!  Does `z_f_cross_U` need to be defined anywhere or is it already computed?

src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
```diff
@@ -3,39 +3,52 @@ using Oceananigans.ImmersedBoundaries: PartialCellBottom, ImmersedBoundaryGrid
...
-update_hydrostatic_pressure!(grid, model; kwargs...) =
-    update_hydrostatic_pressure!(model.pressures.pHY′, model.architecture, model.grid, model.buoyancy, model.tracers; kwargs...)
+"""
+    update_hydrostatic_pressure!(pHY′, grid, buoyancy, tracers, coriolis, velocities; parameters=:xy)
+
+Update the hydrostatic pressure perturbation pHY′. This is done by integrating
+the `buoyancy_perturbationᶜᶜᶜ` downwards:
+
+```math
+pHY′ = ∫ b - ẑ ⋅ (f × u) dz′
```
francispoulin: Can we have a [ before the b and ] before the dz'?

glwagner:
> This looks good to me and I approve the changes, but I think someone else should too.
> 
> To add in the complete Coriolis force we just need to add another body force in the zonal momentum equation. That should be easy enough to do without worrying about whether the methods are energy or enstrophy conserving.

why do we need another body force? Isn't this already included in x_cross_U? We already have non-traditional models for example.

You could set up an example with a non-traditional Coroilis in the hydrostatic model to validate this

glwagner:
for example

https://github.com/CliMA/Oceananigans.jl/blob/6c2944e0b122743147ae4e768188ebb4753c954d/src/Coriolis/constant_cartesian_coriolis.jl#L77

https://github.com/CliMA/Oceananigans.jl/blob/6c2944e0b122743147ae4e768188ebb4753c954d/src/Coriolis/constant_cartesian_coriolis.jl#L68-L69

this is the complete tangent plane approximation for any background rotation around an arbitrary axis. There is also a non-traditional beta plane.

francispoulin:
> > This looks good to me and I approve the changes, but I think someone else should too.
> > To add in the complete Coriolis force we just need to add another body force in the zonal momentum equation. That should be easy enough to do without worrying about whether the methods are energy or enstrophy conserving.
> 
> why do we need another body force? Isn't this already included in x_cross_U? We already have non-traditional models for example.
> 
> You could set up an example with a non-traditional Coroilis in the hydrostatic model to validate this

This is probably my misunderstanding.   I just meant that in the zonal momentum equation we need a term like

```
2Ω * cos(latitude) * w
```

If it's there already, great.   If not, it's easy to add it in.

We need this term to cancel the term added into the vertical momentum equation so that energy is conserved.

Maybe it's there already but I didn't see it.

When I look in the spherical Coriolis script, I see that it only computes the sin of the latitude, not the cosine.  But this should be easy to copy over from the non-traditional term, as you mentioned.

https://github.com/CliMA/Oceananigans.jl/blob/main/src/Coriolis/hydrostatic_spherical_coriolis.jl

src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
navidcy:
```suggestion
pHY′ = ∫ [ b - ẑ ⋅ (f × u) ] dz′
```

glwagner:
> This is probably my misunderstanding. I just meant that in the zonal momentum equation we need a term like
> 
> ```
> 2Ω * cos(latitude) * w
> ```

There are two things going on and I would like to implement them in separate PRs. First, this PR implements a z-component of the Coriolis force which can be used in the hydrostatic model (we already have this for `NonhydrostaticModel`). This allows us to use Coriolis types that include z-components (as well as an additional zonal component). There are two such types right now --- `ConstantCartesianCoriolis` and `NonTraditionalBetaPlane`.

Since we already have these implemented, a simulation that uses these types in `HydrostaticFreeSurfaceModel` can be used as a validation case for this PR.

There is finally an _additional_ Coriolis type we can add, which includes the full spherical variation of the Coriolis force with latitude _in addition_ to the extra vertical and zonal components of the Coriolis force that do not appear in the typical `HydrostaticCoriolisForce`.

This second addition is beyond the scope of this PR and should be tackled in another PR.

glwagner:
Just to clarify further, it is only models on `LatitudeLongitudeGrid` or `OrthogonalSphericalShellGrid` that can use a latitude-dependent Coriolis force. But the quasi-hydrostatic approximation has applications on both `LatitudeLongitudeGrid` _and_ `RectilinearGrid`. Does that make sense?

francispoulin:
Thanks @glwagner for clarifying.  I'm happy with these two parts, and am glad we are starting with the easier of the two.

I just want to repeat that if we include the Coriolis term in the vertical equation, so we have hydrostatic balance, we should always make sure to include the other non-traditional term in the zonal momentum equation.  If you don't include both your energy balance will be all messed up.

glwagner:
> Thanks @glwagner for clarifying. I'm happy with these two parts, and am glad we are starting with the easier of the two.
> 
> I just want to repeat that if we include the Coriolis term in the vertical equation, so we have hydrostatic balance, we should always make sure to include the other non-traditional term in the zonal momentum equation. If you don't include both your energy balance will be all messed up.

But that is guaranteed by every Coriolis implementation. So what is the problem?

glwagner:
Here are two examples:

# `HydrostaticSphericalCoriolis`

https://github.com/CliMA/Oceananigans.jl/blob/013f275662aa22fc143ae274c9ad6f8167e028cb/src/Coriolis/hydrostatic_spherical_coriolis.jl#L62-L64

https://github.com/CliMA/Oceananigans.jl/blob/013f275662aa22fc143ae274c9ad6f8167e028cb/src/Coriolis/hydrostatic_spherical_coriolis.jl#L50

# `ConstantCartesianCoriolis`

https://github.com/CliMA/Oceananigans.jl/blob/013f275662aa22fc143ae274c9ad6f8167e028cb/src/Coriolis/constant_cartesian_coriolis.jl#L77-L79

@francispoulin can you please state specifically where there is a problem in the code?

francispoulin:
> > Thanks @glwagner for clarifying. I'm happy with these two parts, and am glad we are starting with the easier of the two.
> > I just want to repeat that if we include the Coriolis term in the vertical equation, so we have hydrostatic balance, we should always make sure to include the other non-traditional term in the zonal momentum equation. If you don't include both your energy balance will be all messed up.
> 
> But that is guaranteed by every Coriolis implementation. So what is the problem?

Not a problem, I don't think.

francispoulin:
> Here are two examples:
> 
> # `HydrostaticSphericalCoriolis`
> https://github.com/CliMA/Oceananigans.jl/blob/013f275662aa22fc143ae274c9ad6f8167e028cb/src/Coriolis/hydrostatic_spherical_coriolis.jl#L62-L64
> 
> https://github.com/CliMA/Oceananigans.jl/blob/013f275662aa22fc143ae274c9ad6f8167e028cb/src/Coriolis/hydrostatic_spherical_coriolis.jl#L50
> 
> # `ConstantCartesianCoriolis`
> https://github.com/CliMA/Oceananigans.jl/blob/013f275662aa22fc143ae274c9ad6f8167e028cb/src/Coriolis/constant_cartesian_coriolis.jl#L77-L79
> 
> @francispoulin can you please state specifically where there is a problem in the code?

In the first term I don't now where we have `2*Omega*cos(phi)*w - 2*Omega*sin(phi)*u` terms define.  Can you point me to the code that does this?

The second term looks to be correct but I don't know where the terms like `fʸw_minus_fᶻv` are defined.   Can you point me to that as well?

glwagner:
> In the first term I don't now where we have `2*Omega*cos(phi)*w - 2*Omega*sin(phi)*u` terms define. Can you point me to the code that does this?

The `HydrostaticSphericalCoriolis` is hydrostatic (not quasi-hydrostatic), and therefore does _not include the non-traditional term_. This is done consistently, so that z_f_cross_U = 0 as well.

To include the second term, we need a new Coriolis type.

In other words, there is no problem now, because there is no Coriolis implementation that includes a vertical component but then inconsistently does NOT include the corresponding zonal term.

Note, there is _no where else int he code_ where Coriolis forces are defined. The entire Coriolis force must be defined through the `*_f_cross_U` functions.

francispoulin:
@glwagner I added a file in examples called `linear_shear.jl`.  It specifies a shear in the zonal velocity both in `y` and `z`, which are relatec to `fz` and `fy`, respecitvely.

This still needs to be set up to validate that this PR does what it wants, but is this what you were thinking?

It's easy enough to have either the traditional or nontraditional terms, or both.  I suppose it would be worth testing what this done on main before the choices?

src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
```diff
@@ -3,39 +3,52 @@ using Oceananigans.ImmersedBoundaries: PartialCellBottom, ImmersedBoundaryGrid
...
 const PCB = PartialCellBottom
 const PCBIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:PCB}
 
-update_hydrostatic_pressure!(pHY′, arch, ibg::PCBIBG, buoyancy, tracers; parameters = p_kernel_parameters(ibg.underlying_grid)) =
-    update_hydrostatic_pressure!(pHY′, arch, ibg.underlying_grid, buoyancy, tracers; parameters)
-
-update_hydrostatic_pressure!(pHY′, arch, grid, buoyancy, tracers; parameters = p_kernel_parameters(grid)) =
-    launch!(arch, grid, parameters, _update_hydrostatic_pressure!, pHY′, grid, buoyancy, tracers)
+update_hydrostatic_pressure!(pHY′, ibg::PCBIBG, args...; kwargs...)
+    update_hydrostatic_pressure!(pHY′, ibg.underlying_grid, args...; kwargs...)
```
glwagner:
```suggestion
update_hydrostatic_pressure!(pHY′, ibg::PCBIBG, args...; kwargs...) =
    update_hydrostatic_pressure!(pHY′, ibg.underlying_grid, args...; kwargs...)
```

francispoulin:
@glwagner , did you have a chance to look at the file I added into examples?  This should not be an example, but it's something we can play with, I think.  It specifies a mean velocity in two directions that depend on both components of the Coriolis parameters.  I thought this is what you had asked about.

glwagner:
What's the output of the file? Does it validate the new Coriolis term?

francispoulin:
> What's the output of the file? Does it validate the new Coriolis term?

No.  At the moment it just sets up a thermal wind problem with the traditional and non-traditioinal Coriolis terms.  It is using the Cartesian geometry, just as a starting point and will be useful to build from and then for comparison.  

I will extend to to a lat-lon grid and see how things work, if they work.  I just wanted to check to see if this is what you had in mind first.

I can work on this on Monday or Tuesday for the spherical geometry.

glwagner:
Does the script run?

glwagner:
Any progress @francispoulin ?

francispoulin:
Sorry, been traveling but will get to this today.  Thanks!

glwagner:
@francispoulin this PR has gone stale. I think we should close it. However, you should feel free to re-implement this on `main` if you like!

PR #4598, opened 2025-06-11
Remove default particles from checkpointing; also only add time stepper if needed

simone-silvestri:
Not all models have particles and timesteppers. This PR removes the `default_checkpointed_properties` which is redundant, in favor of a `required_checkpointed_properties` function which defaults to `[:grid, :clock]` 
This is extended in the `Models` module for Oceananigans' models


src/OutputWriters/checkpointer.jl
```diff
@@ -72,7 +60,7 @@ Keyword arguments
 
 - `properties`: List of model properties to checkpoint. This list _must_ contain
                 `:grid`, `:particles` and `:clock`, and if using AB2 timestepping then also
-                `:timestepper`. Default: calls [`default_checkpointed_properties`](@ref) on
+                `:timestepper`. Default: calls [`required_checkpointed_properties`](@ref) on
```
glwagner:
```suggestion
                `:timestepper`. Default: calls [`required_checkpoint_properties`](@ref) on
```

glwagner:
Nice!

Minor suggestion: use `required_checkpoint_properties` instead (just better English I think)

Also, could we simply assume that models need grid and clock? I guess this keeps the model interface _very_ general, but I don't know if we need that. Food for thought.

PR #4599, opened 2025-06-11
Add output functionality for conformal cubed sphere fields

siddharthabishnu:
This PR introduces output capabilities for conformal cubed sphere fields, specifically support for output writers. Functionality for checkpointing and field time series will be introduced in a future PR.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.96.33"
+version = "0.96.34"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl
+++ b/src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl
@@ -1,6 +1,6 @@
 module LagrangianParticleTracking
 
-export LagrangianParticles
+export LagrangianParticles, DroguedParticleDynamics
 
 using Printf
 using Adapt
@@ -129,6 +129,7 @@ end
 
 include("update_lagrangian_particle_properties.jl")
 include("lagrangian_particle_advection.jl")
+include("drogued_dynamics.jl")
 
 step_lagrangian_particles!(::Nothing, model, Δt) = nothing
 
--- a/src/Models/LagrangianParticleTracking/drogued_dynamics.jl
+++ b/src/Models/LagrangianParticleTracking/drogued_dynamics.jl
@@ -1,0 +1,72 @@
+"""
+    DroguedParticleDynamics(depths)
+
+`DroguedParticleDynamics` goes in the `dynamics` slot of `LagrangianParticles` 
+and modifies their behaviour to mimic the behaviour of bouys which are 
+drogued at `depths`. The particles remain at the their `z` position
+so the "measurment depth can be set", and then are advected in the `x` and `y`
+directions according to the velocity field at `depths`.
+
+`depths` should be an (abstract) array of length `length(particles)`.
+
+
+Example
+=======
+
+```jldoctest
+julia> using Oceananigans
+
+julia> n = 10
+10
+
+julia> dynamics = DroguedParticleDynamics(-10:10/n:0)
+DroguedParticleDynamics{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}(-10.0:1.0:0.0)
+
+julia> particles = LagrangianParticles(; x = zeros(n), y = zeros(n), z = zeros(n), dynamics)
+10 LagrangianParticles with eltype Particle:
+├── 3 properties: (:x, :y, :z)
+├── particle-wall restitution coefficient: 1.0
+├── 0 tracked fields: ()
+└── dynamics: DroguedParticleDynamics{StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}
+
+```
+"""
+struct DroguedParticleDynamics{DD}
+    depths :: DD
+end
+
+Adapt.adapt_structure(to, dbd::DroguedParticleDynamics) = 
+    DroguedParticleDynamics(adapt(to, dbd.depths))
+
+@inline (::DroguedParticleDynamics)(args...) = nothing
+
+const DroguedParticles = LagrangianParticles{<:Any, <:Any, <:Any, <:DroguedParticleDynamics}
+
+function advect_lagrangian_particles!(particles::DroguedParticles, model, Δt)
+    grid = model.grid
+    arch = architecture(grid)
+    parameters = KernelParameters(1:length(particles))
+
+    launch!(arch, grid, parameters,
+            _advect_drogued_particles!,
+            particles.properties, particles.restitution, model.grid, Δt, total_velocities(model), particles.dynamics.depths)
+
+    return nothing
+end
+
+@kernel function _advect_drogued_particles!(properties, restitution, grid, Δt, velocities, depths)
+    p = @index(Global)
+
+    @inbounds begin
+        x = properties.x[p]
+        y = properties.y[p]
+        z = depths[p]
+    end
+
+    x⁺, y⁺, _ = advect_particle((x, y, z), properties, p, restitution, grid, Δt, velocities)
+
+    @inbounds begin
+        properties.x[p] = x⁺
+        properties.y[p] = y⁺
+    end
+end
--- a/src/Models/Models.jl
+++ b/src/Models/Models.jl
@@ -6,7 +6,7 @@ export
     HydrostaticFreeSurfaceModel, ZStar, ZCoordinate,
     ExplicitFreeSurface, ImplicitFreeSurface, SplitExplicitFreeSurface,
     PrescribedVelocityFields, PressureField,
-    LagrangianParticles,
+    LagrangianParticles, DroguedParticleDynamics,
     BoundaryConditionOperation, ForcingOperation,
     seawater_density
 
@@ -109,7 +109,7 @@ using .HydrostaticFreeSurfaceModels:
 
 using .ShallowWaterModels: ShallowWaterModel, ConservativeFormulation, VectorInvariantFormulation
 
-using .LagrangianParticleTracking: LagrangianParticles
+using .LagrangianParticleTracking: LagrangianParticles, DroguedParticleDynamics
 
 const OceananigansModels = Union{HydrostaticFreeSurfaceModel,
                                  NonhydrostaticModel,
--- a/src/Oceananigans.jl
+++ b/src/Oceananigans.jl
@@ -75,7 +75,7 @@ export
     viscosity, diffusivity,
 
     # Lagrangian particle tracking
-    LagrangianParticles,
+    LagrangianParticles, DroguedParticleDynamics,
 
     # Models
     NonhydrostaticModel, HydrostaticFreeSurfaceModel, ShallowWaterModel,
--- a/test/test_lagrangian_particle_tracking.jl
+++ b/test/test_lagrangian_particle_tracking.jl
@@ -4,6 +4,8 @@ using NCDatasets
 using StructArrays
 using Oceananigans.Architectures: architecture, on_architecture
 
+using Oceananigans.Models.LagrangianParticleTracking: no_dynamics
+
 struct TestParticle{T}
     x::T
     y::T
@@ -41,7 +43,7 @@ function particle_tracking_simulation(; grid, particles, timestepper=:RungeKutta
     return sim, jld2_filepath, nc_filepath
 end
 
-function run_simple_particle_tracking_tests(grid, timestepper=:QuasiAdamsBashforth)
+function run_simple_particle_tracking_tests(grid, dynamics, timestepper=:QuasiAdamsBashforth)
 
     arch = architecture(grid)
 
@@ -55,7 +57,7 @@ function run_simple_particle_tracking_tests(grid, timestepper=:QuasiAdamsBashfor
     ys = on_architecture(arch, 0.58 * ones(P))
     zs = on_architecture(arch, 0.8 * ones(P))
 
-    particles = LagrangianParticles(x=xs, y=ys, z=zs)
+    particles = LagrangianParticles(; x=xs, y=ys, z=zs, dynamics)
     @test particles isa LagrangianParticles
 
     if grid isa RectilinearGrid
@@ -81,7 +83,7 @@ function run_simple_particle_tracking_tests(grid, timestepper=:QuasiAdamsBashfor
 
     x, y, z = on_architecture.(Ref(arch), ([0.0], [0.0], [initial_z]))
 
-    particles = LagrangianParticles(; x, y, z)
+    particles = LagrangianParticles(; x, y, z, dynamics)
     u, v, w = VelocityFields(grid)
 
     Δt = 0.01
@@ -94,8 +96,10 @@ function run_simple_particle_tracking_tests(grid, timestepper=:QuasiAdamsBashfor
 
     time_step!(model, Δt)
 
-    zᶠ = convert(array_type(arch), model.particles.properties.z)
-    @test all(zᶠ .≈ (top_boundary - 0.15))
+    if dynamics == no_dynamics
+        zᶠ = convert(array_type(arch), model.particles.properties.z)
+        @test all(zᶠ .≈ (top_boundary - 0.15))
+    end
 
     #####
     ##### Test custom particle "TestParticle"
@@ -121,7 +125,7 @@ function run_simple_particle_tracking_tests(grid, timestepper=:QuasiAdamsBashfor
     background_v .= 1
 
     # Test second constructor
-    lagrangian_particles = LagrangianParticles(particles; tracked_fields)
+    lagrangian_particles = LagrangianParticles(particles; tracked_fields, dynamics)
     @test lagrangian_particles isa LagrangianParticles
 
     if grid isa RectilinearGrid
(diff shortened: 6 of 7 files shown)


glwagner:
> This PR introduces input/output capabilities for conformal cubed sphere fields, including support for output writers, checkpointing, and field time series.

I suggest reducing the scope a bit and focusing on one thing at a time so that the PR can be merged more quickly (esp since this PR will easily create merge conflicts). Maybe checkpointing for a different PR?

siddharthabishnu:
> > This PR introduces input/output capabilities for conformal cubed sphere fields, including support for output writers, checkpointing, and field time series.
> 
> I suggest reducing the scope a bit and focusing on one thing at a time so that the PR can be merged more quickly (esp since this PR will easily create merge conflicts). Maybe checkpointing for a different PR?

Yes, that's a good idea. I’ll remove the checkpointing functionality from this PR and include it in a future one. I’ve updated the PR description to reflect the revised scope.

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -100,8 +100,10 @@ const OSSG = OrthogonalSphericalShellGrid
 const ZRegOSSG = OrthogonalSphericalShellGrid{<:Any, <:Any, <:Any, <:Any, <:RegularVerticalCoordinate}
 const ZRegOrthogonalSphericalShellGrid = ZRegOSSG
 
-@inline cpu_face_constructor_ξ(grid::OrthogonalSphericalShellGrid) = Array(grid.conformal_mapping.ξᶠᵃᵃ[1:size(grid, 1)+1])
-@inline cpu_face_constructor_η(grid::OrthogonalSphericalShellGrid) = Array(grid.conformal_mapping.ηᵃᶠᵃ[1:size(grid, 2)+1])
+@inline cpu_face_constructor_ξ(grid::OrthogonalSphericalShellGrid) = (grid.conformal_mapping.ξᶠᵃᵃ[1],
```
simone-silvestri:
```suggestion
@inline cpu_face_constructor_ξ(grid::ConformalCubedSpherePanelGrid) = (grid.conformal_mapping.ξᶠᵃᵃ[1],
```
btw, what if the dimension is stretched? Wouldn't this be wrong?

src/Grids/orthogonal_spherical_shell_grid.jl
```diff
@@ -100,8 +100,10 @@ const OSSG = OrthogonalSphericalShellGrid
...
+                                                                      grid.conformal_mapping.ξᶠᵃᵃ[size(grid, 1)+1])
+@inline cpu_face_constructor_η(grid::OrthogonalSphericalShellGrid) = (grid.conformal_mapping.ηᵃᶠᵃ[1],
```
simone-silvestri:
```suggestion
@inline cpu_face_constructor_η(grid:: ConformalCubedSpherePanelGrid) = (grid.conformal_mapping.ηᵃᶠᵃ[1],
```
Like above

src/MultiRegion/multi_region_output_writers.jl
```diff
@@ -13,18 +14,9 @@ function fetch_output(mrf::MultiRegionField, model)
     return parent(field)
 end
 
-function construct_output(mrf::MultiRegionField, grid, user_indices, with_halos)
```
simone-silvestri: Was this not working? Does this PR remove this functionality?

src/OrthogonalSphericalShellGrids/conformal_cubed_sphere_panel.jl
```diff
@@ -91,9 +91,13 @@ function ConformalCubedSpherePanelGrid(filepath::AbstractString, architecture =
                               x = ξ, y = η, z, halo)
 
      ξᶠᵃᵃ = xnodes(ξη_grid, Face())
+     ξᶠᵃᵃ = ntuple(i -> FT(ξᶠᵃᵃ[i]), length(ξᶠᵃᵃ))
```
simone-silvestri: Why are you transforming these into tuples? Should these not be an array like the coordinates of the other grids?

siddharthabishnu: I'm transforming these into tuples primarily for two reasons. First, the `parent` of `xnodes` (or `ynodes` or `znodes`) is not correctly defined. Consider the MRW below:
```
julia> using Oceananigans

julia> grid = RectilinearGrid(size = (4, 4, 4), x = (0, 1), y = (0, 1), z = (0, 1), topology=(Bounded, Bounded, Bounded), halo = (2, 2, 2))
4×4×4 RectilinearGrid{Float64, Bounded, Bounded, Bounded} on CPU with 2×2×2 halo
├── Bounded  x ∈ [0.0, 1.0] regularly spaced with Δx=0.25
├── Bounded  y ∈ [0.0, 1.0] regularly spaced with Δy=0.25
└── Bounded  z ∈ [0.0, 1.0] regularly spaced with Δz=0.25

julia> ξᶠᵃᵃ = xnodes(grid, Face())
5-element view(OffsetArray(::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, -1:7), 1:5) with eltype Float64:
 0.0
 0.25
 0.5
 0.75
 1.0

julia> parent(ξᶠᵃᵃ)
-0.5:0.25:1.5 with indices -1:7

julia> parent(ξᶠᵃᵃ) isa Array
false

julia> typeof(parent(ξᶠᵃᵃ))
OffsetVector{Float64, StepRangeLen{Float64, TwicePrecision{Float64}, TwicePrecision{Float64}, Int64}} (alias for OffsetArrays.OffsetArray{Float64, 1, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}})

julia> Array(parent(ξᶠᵃᵃ))
ERROR: DimensionMismatch: axes must agree, got (Base.OneTo(9),) and (OffsetArrays.IdOffsetRange(values=-1:7, indices=-1:7),)
Stacktrace:
 [1] _checkaxs(axd::Tuple{Base.OneTo{Int64}}, axs::Tuple{OffsetArrays.IdOffsetRange{Int64, Base.OneTo{Int64}}})
   @ Base ./abstractarray.jl:1163
 [2] copyto_axcheck!
   @ ./abstractarray.jl:1166 [inlined]
 [3] Array
   @ ./array.jl:626 [inlined]
 [4] Array(A::OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}})
   @ Core ./boot.jl:605
 [5] top-level scope
   @ REPL[83]:1
```
We can create a new PR to address this issue, but until then, the line
https://github.com/CliMA/Oceananigans.jl/blob/8a1a04ca928eaf553c03e63ed9f3352d8870bf4d/src/OutputWriters/output_writer_utils.jl#L81
results in a dimension mismatch error, as shown above. One workaround is to modify the line:
```
ξᶠᵃᵃ = xnodes(ξη_grid, Face())
```
to:
```
ξᶠᵃᵃ = Array(xnodes(ξη_grid, Face()))
```
but then `Array(parent(ξᶠᵃᵃ))` in line
https://github.com/CliMA/Oceananigans.jl/blob/8a1a04ca928eaf553c03e63ed9f3352d8870bf4d/src/OutputWriters/output_writer_utils.jl#L81
won’t be classified as `bitstype`, which in turn causes a GPU error:
```
julia> isbitstype(typeof(Array(parent(ξᶠᵃᵃ))))
false
```
However, if I convert these arrays into tuples, not only is the result `bitstype`, but also line
https://github.com/CliMA/Oceananigans.jl/blob/8a1a04ca928eaf553c03e63ed9f3352d8870bf4d/src/OutputWriters/output_writer_utils.jl#L83
gets called instead of
https://github.com/CliMA/Oceananigans.jl/blob/8a1a04ca928eaf553c03e63ed9f3352d8870bf4d/src/OutputWriters/output_writer_utils.jl#L81
which avoids the dimension mismatch error altogether.
```
julia> ξᶠᵃᵃ = xnodes(grid, Face())
5-element view(OffsetArray(::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, -1:7), 1:5) with eltype Float64:
 0.0
 0.25
 0.5
 0.75
 1.0

julia> ξᶠᵃᵃ = ntuple(i -> FT(ξᶠᵃᵃ[i]), length(ξᶠᵃᵃ))
(0.0, 0.25, 0.5, 0.75, 1.0)

julia> ξᶠᵃᵃ isa Tuple
true

julia> isbitstype(typeof(ξᶠᵃᵃ))
true
```

src/Grids/orthogonal_spherical_shell_grid.jl
siddharthabishnu: The functions `cpu_face_constructor_ξ` and `cpu_face_constructor_η` are used only to specify the endpoints of the computational domain, which remain the same even in the case of a stretched conformal cubed sphere panel.

src/MultiRegion/multi_region_output_writers.jl
siddharthabishnu: Sorry, I removed it due to a method ambiguity with function `construct_output(user_output::Union{AbstractField, Reduction}, grid::ConformalCubedSphereGridOfSomeKind, user_indices, with_halos)`. What's the best way to avoid that ambiguity if I add it back?

src/Grids/orthogonal_spherical_shell_grid.jl
siddharthabishnu: Suggestion incorporated in commit [d498526](https://github.com/CliMA/Oceananigans.jl/pull/4599/commits/d498526a3d577135b728588c28661db6bbaa69ad).

src/Grids/orthogonal_spherical_shell_grid.jl
siddharthabishnu: Suggestion incorporated in commit [d498526](https://github.com/CliMA/Oceananigans.jl/pull/4599/commits/d498526a3d577135b728588c28661db6bbaa69ad).

simone-silvestri: Are you sure? We typically use `cpu_face_constructor_*` to reconstruct a grid with new halos. For this reason we need to distinguish between stretched coordinates, where the constructor is an array, and uniform spacing coordinates where the `cpu_face_constructor_*` is a tuple.

simone-silvestri: Maybe this is not a problem for the cubed sphere, where `cpu_face_constructor_ξ` and `cpu_face_constructor_η` are not used to construct the grid? But then if you save the grid using these functions, wouldn't you need to specify it?

src/MultiRegion/multi_region_output_writers.jl
simone-silvestri: you would need to add what happens for `construct_output(mrf::MultiRegionField, grid::ConformalCubedSphereGridOfSomeKind, user_indices, with_halos)`.

src/OrthogonalSphericalShellGrids/conformal_cubed_sphere_panel.jl
simone-silvestri: I am not sure about the `isbits` problem. Do you need a bits type to save in jld2? I thought we didn't need that. Long tuples are not very performant and consume a lot of register memory if passed to the GPU. 
Could it be possible to solve the problem with `parent`? In this way you are saving coordinates in a quite unconvenient format:
```julia
 saveproperty!(file, address, p::Tuple) = [saveproperty!(file, address * "/$i", p[i]) for i in 1:length(p)] 
```
this returns something organized like this
```julia
julia> file
JLDFile /Users/simonesilvestri/test.jld2 (read/write)
 └─📂 ξᶠᵃᵃ
    ├─🔢 1
    ├─🔢 2
    ├─🔢 3
    ├─🔢 4
    ├─🔢 5
    ├─🔢 6
    ├─🔢 7
    ├─🔢 8
    └─ ⋯ (92 more entries)
```
where if you want to read it back you have to open individually every address, which is not a problem maybe for a cubed sphere with 10 points in each panel, but it might become if you are going high resolution.

src/MultiRegion/multi_region_output_writers.jl
siddharthabishnu: Done in commit [55de2ca](https://github.com/CliMA/Oceananigans.jl/pull/4599/commits/55de2cadf7f14bed718639358f39252db485e292). 🙂

src/OrthogonalSphericalShellGrids/conformal_cubed_sphere_panel.jl
siddharthabishnu: @simone-silvestri Thank you for the clear explanation!

Maybe we don't need a `bitstype` for saving to JLD2. However, when I tried converting the nodes to arrays and then applying the `parent` operator, I encountered the `isbits` error [here](https://buildkite.com/clima/oceananigans/builds/24111#019782d2-a398-4212-b762-231ff5ffce38) during the numerical bottom height computation of the immersed boundary grid.

siddharthabishnu: What I’m struggling to understand is why only the computational coordinates, and not the coordinate and metric arrays of the cubed-sphere panel are causing this error, even though none of them are `bitstype`.

siddharthabishnu: I was able to resolve the issue using `copy`, which not only reindexes the nodes to start at 1, avoiding dimension mismatch errors from `Array(parent(...))`, but also returns a `bitstype`-compatible object, without the register pressure and memory overhead associated with large `NTuple`s or `SVector`s.

```
julia> using Oceananigans

julia> grid = RectilinearGrid(size = (4, 4, 4), x = (0, 1), y = (0, 1), z = (0, 1), topology=(Bounded, Bounded, Bounded), halo = (2, 2, 2))
4×4×4 RectilinearGrid{Float64, Bounded, Bounded, Bounded} on CPU with 2×2×2 halo
├── Bounded  x ∈ [0.0, 1.0] regularly spaced with Δx=0.25
├── Bounded  y ∈ [0.0, 1.0] regularly spaced with Δy=0.25
└── Bounded  z ∈ [0.0, 1.0] regularly spaced with Δz=0.25

julia> x = xnodes(grid, Face())
5-element view(OffsetArray(::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, -1:7), 1:5) with eltype Float64:
 0.0
 0.25
 0.5
 0.75
 1.0

julia> typeof(x)
SubArray{Float64, 1, OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}}, Tuple{UnitRange{Int64}}, true}

julia> parent(x)
-0.5:0.25:1.5 with indices -1:7

julia> Array(parent(x))
ERROR: DimensionMismatch: axes must agree, got (Base.OneTo(9),) and (OffsetArrays.IdOffsetRange(values=-1:7, indices=-1:7),)
Stacktrace:
 [1] _checkaxs(axd::Tuple{Base.OneTo{Int64}}, axs::Tuple{OffsetArrays.IdOffsetRange{Int64, Base.OneTo{Int64}}})
   @ Base ./abstractarray.jl:1163
 [2] copyto_axcheck!
   @ ./abstractarray.jl:1166 [inlined]
 [3] Array
   @ ./array.jl:626 [inlined]
 [4] Array(A::OffsetArrays.OffsetVector{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}})
   @ Core ./boot.jl:605
 [5] top-level scope
   @ REPL[36]:1

julia> y = copy(x)
0.0:0.25:1.0

julia> typeof(y)
StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}

julia> parent(y)
0.0:0.25:1.0

julia> Array(parent(y))
5-element Vector{Float64}:
 0.0
 0.25
 0.5
 0.75
 1.0

julia> isbitstype(typeof(x))
true

julia> isbitstype(typeof(y))
true
```

src/Grids/orthogonal_spherical_shell_grid.jl
siddharthabishnu: Ah, I see what you mean now. I just checked, and in the current version of the code, `cpu_face_constructor_x` and `cpu_face_constructor_y` return the `x` and `y` endpoints of the domain as tuples for the `RectangularGrid`, and the latitude-longitude extents for the `LatitudeLongitudeGrid`. It’s only for the `OrthogonalSphericalShellGrid` that these functions return arrays of coordinates. The analogous functions `cpu_face_constructor_ξ` and `cpu_face_constructor_η` for the computational coordinates are not used anywhere in the source code, not even for regenerating `OrthogonalSphericalShellGrid`s with new halos.

That said, you're right---for the conformal cubed sphere panel, I’m not using the functions `cpu_face_constructor_ξ` and `cpu_face_constructor_η` to reconstruct the grid, but only to specify the domain endpoints, similar to how they're used in rectangular or latitude-longitude grids. In particular, when the coordinates are stretched, the EKI step used to optimize the grid spacing is computationally expensive. So when reconstructing the grid with new halos, I avoid regenerating the spacing and instead reuse the original conformal mapping, which is passed via the `provided_conformal_mapping` argument.

src/OrthogonalSphericalShellGrids/conformal_cubed_sphere_panel.jl
glwagner: You shouldn't need the copy after #4615 . Note issue #4620

glwagner: My other comment is that I think that `copy` isn't the cleanest solution to this issue -- I don't think the purpose of `copy` is to remove subarray/OffsetArray wrappers, is it?

siddharthabishnu: Right. I was using `copy` primarily for that side effect, but now, I've now come up with a better solution that takes care of both the issues.

PR #4607, opened 2025-06-16
Add `PiecewiseLinearMask`

tomchor:
This PR adds `PiecewiseLinearMask` to the `Forcings` module.

One advantage of `PiecewiseLinearMask` over the existing `GaussianMask` is that it has a well-defined region of action, as opposed to `GaussianMask` which decays smoothly and it's up to the user to decide where the mask is small enough that it doesn't affect the solution. Another advantage is that it's probably faster than `GaussianMask`, although I haven't really tested that.

### DIFF
--- a/src/Forcings/relaxation.jl
+++ b/src/Forcings/relaxation.jl
@@ -129,8 +129,11 @@ Example
 
 Create a Gaussian mask centered on `z=0` with width `1` meter.
 
-```julia
+```jldoctest
+julia> using Oceananigans
+
 julia> mask = GaussianMask{:z}(center=0, width=1)
+GaussianMask{:z, Int64}(0, 1)
 ```
 """
 struct GaussianMask{D, T}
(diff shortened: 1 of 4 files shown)


src/Forcings/relaxation.jl
```diff
@@ -154,6 +157,63 @@ show_exp_arg(D, c) = c == 0 ? "$D^2" :
...
+    function PiecewiseLinearMask{D}(; center, width) where D
+        T = promote_type(typeof(center), typeof(width))
+        return new{D, T}(center, width)
+    end
+end
+
+@inline function (p::PiecewiseLinearMask{:x})(x, y, z)
+    d = abs(x - p.center)
+    return d > p.width ? 0.0 : 1.0 - d/p.width
+end
```
glwagner:
```suggestion
    d = 1 - abs(x - p.center) / p.width
    return max(0, d)
end
```

There are two issues with the code as written:

1. It hardcodes Float64 in `0.0` and `1.0`
2. It invokes short-circuiting logic but doesn't need to

glwagner: the same changes should be made for `y` and `z`

tomchor: Ah, yeah, thanks for catching that. Done.

glwagner (approved):

PR #4608, opened 2025-06-18
Updating barotropic gyre validation script to run with current package interface

matt-graham:
When trying to run the barotropic gyre validation example script with a recent version of Oceananigans package (v0.96.32) I got a series of errors due to changes in Oceananigans interface since the script was written. Specifically

- Import of `years` from `Oceananigans.Units` and `Oceananigans.Utils` fails due to `years` having being removed in #3059.
- Call signature of `xnodes` and `ynodes` changed in #2842 resulting in `MethodError` in current calls.
- `show_axis` not a valid argument to `wireframe!` in current (GL)Makie (v0.11.11).

This PR updates the scripts so that they run for me locally (in Julia v1.11.4 with Oceananigcans v0.96.32, GLMakie v0.11.11, JLD2 v0.5.13) and successfully outputs the animated visualisation of the simulated fields.

### DIFF
--- a/validation/barotropic_gyre/visualize_barotropic_gyre.jl
+++ b/validation/barotropic_gyre/visualize_barotropic_gyre.jl
@@ -26,7 +26,6 @@ function geographic2cartesian(λ, φ, r=1)
 end
 
 function visualize_barotropic_gyre(filepath)
-
     file = jldopen(filepath)
 
     Nx = file["grid/Nx"]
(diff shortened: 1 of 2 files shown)


validation/barotropic_gyre/visualize_barotropic_gyre.jl
```diff
@@ -61,7 +61,7 @@ function visualize_barotropic_gyre(filepath)
     statenames = ["u", "v", "η"]
     for (n, var) in enumerate([u, v, η])
         ax = fig[3:7, 3n-2:3n] = LScene(fig) # make plot area wider
-        wireframe!(ax, Sphere(Point3f(0), 0.99f0), show_axis=false)
```
matt-graham, line 64: I couldn't figure out what the equivalent to `show_axis=false` is in current Makie interface so have just removed the argument here, but this does I assume change the rendering of the output compared to previously.

matt-graham: Now resolved - figured out the `show_axis` argument just needed to be shifted to `LScene` constructor.

navidcy (approved):
Thanks @matt-graham. Looks great! I tweaked a bit more.

PR #4615, opened 2025-06-24
Example with `TEOS10EquationOfState` + "single wrapped" `*nodes`

### LINKED ISSUE #802: Example using nonlinear equation of state
We should add an example that uses `TEOS10EquationOfState` from `SeawaterPolynomials`.

cc @BrodiePearson

glwagner:
Closes #802 

Makes a few other misc improvements on `ocean_wind_mixing_and_convection.jl` and runs on GPU.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -8,6 +8,7 @@ Distributed.addprocs(2)
     using Printf
 
     using CairoMakie # to avoid capturing precompilation output by Literate
+    set_theme!(Theme(fontsize=20))
     CairoMakie.activate!(type = "svg")
 
     using Oceananigans
--- a/docs/src/fields.md
+++ b/docs/src/fields.md
@@ -190,7 +190,7 @@ Let's visualize the situation:
 ```@setup fields
 using Oceananigans
 using CairoMakie
-set_theme!(Theme(fontsize=24))
+set_theme!(Theme(fontsize=20))
 CairoMakie.activate!(type="svg")
 
 grid = RectilinearGrid(topology = (Periodic, Periodic, Bounded),
--- a/docs/src/grids.md
+++ b/docs/src/grids.md
@@ -5,7 +5,7 @@ DocTestSetup = quote
     using Oceananigans
     using CairoMakie
     CairoMakie.activate!(type = "svg")
-    set_theme!(Theme(fontsize=24))
+    set_theme!(Theme(fontsize=20))
 end
 ```
 
@@ -187,7 +187,7 @@ using Oceananigans.Units
 
 using CairoMakie
 CairoMakie.activate!(type = "svg")
-set_theme!(Theme(fontsize=24))
+set_theme!(Theme(fontsize=20))
 
 grid = RectilinearGrid(topology = (Bounded, Bounded, Bounded),
                        size = (20, 20, 20),
@@ -364,7 +364,7 @@ grid = RectilinearGrid(size = (Nx, Ny, Nz),
 using Oceananigans
 using CairoMakie
 CairoMakie.activate!(type = "svg")
-set_theme!(Theme(fontsize=24))
+set_theme!(Theme(fontsize=20))
 
 Nx, Ny, Nz = 64, 64, 32
 Lx, Ly, Lz = 1e4, 1e4, 1e3
--- a/examples/baroclinic_adjustment.jl
+++ b/examples/baroclinic_adjustment.jl
@@ -74,6 +74,7 @@ set!(model, b=bᵢ)
 # Let's visualize the initial buoyancy distribution.
 
 using CairoMakie
+set_theme!(Theme(fontsize = 20))
 
 ## Build coordinates with units of kilometers
 x, y, z = 1e-3 .* nodes(grid, (Center(), Center(), Center()))
@@ -304,8 +305,6 @@ yv = yv ./ 1e3 # convert m -> km
 
 # Next, we set up a plot with 4 panels. The top panels are large and square, while
 # the bottom panels get a reduced aspect ratio through `rowsize!`.
-
-set_theme!(Theme(fontsize=24))
 
 fig = Figure(size=(1800, 1000))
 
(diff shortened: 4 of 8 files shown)


navidcy (approved):

ext/OceananigansMakieExt.jl
navidcy, line 1: @glwagner the plots didn't work; they were hitting scalar operations

navidcy:
preview at https://clima.github.io/OceananigansDocumentation/previews/PR4615/literated/ocean_wind_mixing_and_convection/
seems nice!

examples/ocean_wind_mixing_and_convection.jl
```diff
@@ -14,32 +14,33 @@
...
 # using Pkg
-# pkg"add Oceananigans, CairoMakie"
+# pkg"add Oceananigans, CairoMakie, SeawaterPolynomials"
 # ```
 
 # We start by importing all of the packages and functions that we'll need for this
 # example.
 
+using Oceananigans
+using Oceananigans.Units: minute, minutes, hour
```
navidcy:
```suggestion
using Oceananigans.Units
```

ext/OceananigansMakieExt.jl
```diff
@@ -100,19 +102,19 @@ convert_arguments(pl::Type{<:AbstractPlot}, f::Field) =
 function convert_arguments(pl::Type{<:AbstractPlot}, op::AbstractOperation)
     f = Field(op)
     compute!(f)
-    return convert_arguments(pl, f)
+    return @allowscalar convert_arguments(pl, f)
```
glwagner: we don't want this

glwagner: do you know why scalar operations were induced @navidcy ? What kind of object does `zspacings` return?

navidcy: Returns a KernelFunctorOperation

navidcy: I guess the scalar operations were induced in the evaluation of it?

ext/OceananigansMakieExt.jl
glwagner: I know but that's a bug!

glwagner: The solution is never to add `@allowscalar` (that should only be for scalar operations or debugging)

glwagner: Do you know why they were hitting scalar operations? I can look more into it later but not now

ext/OceananigansMakieExt.jl
```diff
@@ -7,6 +7,8 @@ using Oceananigans.AbstractOperations: AbstractOperation
 using Oceananigans.Architectures: on_architecture, architecture
 using Oceananigans.ImmersedBoundaries: mask_immersed_field!
 
+using CUDA: @allowscalar
+
```
navidcy:
```suggestion
```

ext/OceananigansMakieExt.jl
navidcy: What my investigations led me was that it was coming from that Makie method I decorated the @allowscalar operations. I presume at some point makie loops over to get the values to plot?

Could we convert a KernelFunctorOperator to CPU for plotting purposes? Would that be better solution?

ext/OceananigansMakieExt.jl
glwagner: But then we write `f = Field(op)`. Shouldn't that evaluate the `KernelFunctionOperation` correctly on GPU?

glwagner: something is fishy

navidcy: I see... Let me have another look.
I believe it would be how Makie indexes the input with give it to make the plot... (but this is just a guess at this point)

glwagner: yes but why? We attempt to convert everything to CPU. Is a conversion being missed?

ext/OceananigansMakieExt.jl
glwagner: Ok so I found the issue. It was not with the field itself. It was with the *nodes functions, which were returning view(offset array(array)). This "double wrapper" does not get adapted to GPU properly (CUDA / GPUArrays only support "single wrapped" GPU arrays). I was surprised to find this. I think we should test this, but we can add the test in a future PR if everything works out here.

navidcy:
When I run with the latest commit I made I got this:

https://github.com/user-attachments/assets/27fbce60-ad06-4f73-b337-edfbe15ab875

Does it look like numerical artefacts to you? What's this checker box patterns, especially towards the final stages of the simulations? You can clearly see them in the eddy viscosity plot but also in w.

navidcy:
Confirming that the artifacts didn't come up because I dropped interior and used OceananigansMakieExt to visualise...

ext/OceananigansMakieExt.jl
navidcy: Good find!

navidcy:
Actually the checkerboard patterns were there to begin with; see https://clima.github.io/OceananigansDocumentation/stable/literated/ocean_wind_mixing_and_convection/

ext/OceananigansMakieExt.jl
```diff
@@ -99,19 +98,16 @@ convert_arguments(pl::Type{<:AbstractPlot}, f::Field) =
...
 
 function convert_arguments(pl::Type{<:AbstractPlot}, ξ1::AbstractArray, op::AbstractOperation)
     f = Field(op)
-    compute!(f)
     return convert_arguments(pl, ξ1, f)
 end
 
 function convert_arguments(pl::Type{<:AbstractPlot}, ξ1::AbstractArray, ξ2::AbstractArray, op::AbstractOperation)
     f = Field(op)
-    compute!(f)
```
navidcy, line 114: the `compute!(f)` is dropped here as a result of #4531, right?
(just wondering)

navidcy:
@glwagner do you understand why the CI doesn't start? Seems some reactant-related thing?

glwagner:
We need to pin Reactant I think.

navidcy:
That seems to do the trick.

PR #4623, opened 2025-06-28
Add ExponentialCoordinate + Docs/Grids section on Stretched Coordinates

navidcy:


glwagner:
> I suggest we merge this and leave the StretchedCoordinate to be vertical-coordinate specific in ClimaOcean?

What's the rationale? I think there are disadvantages to overlapping scope between the packages so it'd nice to clarify what-belongs-where. I think ClimaOcean concerns should be focused on OceanSeaIceModel whereas anything to do with grid generation is more fundamental and seems to be an Oceananigans concern?

src/Grids/stretched_coordinates.jl
glwagner, line 1: how about the somewhat more generic `coordinate_utilities.jl`? We want to avoid the impression that this is an implementation of stretched coordinates. These are helper functions, but not the implementation itself

src/Grids/stretched_coordinates.jl
```diff
@@ -0,0 +1,111 @@
+struct ExponentialCoordinate <: Function
```
glwagner: we could relax the `Function` constraint in this PR as well

src/Grids/stretched_coordinates.jl
navidcy: Good point.

navidcy:
I agree with the comments. I'll make the changes and ping you!

navidcy:
> > I suggest we merge this and leave the StretchedCoordinate to be vertical-coordinate specific in ClimaOcean?
> 
> What's the rationale? I think there are disadvantages to overlapping scope between the packages so it'd nice to clarify what-belongs-where. I think ClimaOcean concerns should be focused on OceanSeaIceModel whereas anything to do with grid generation is more fundamental and seems to be an Oceananigans concern?

Yes, feels natural.

navidcy:
@glwagner, I brought everything here and renamed it to `coordinate_utils.jl`. Have another look. When merged, I'll delete from ClimaOcean.

src/Grids/stretched_coordinates.jl
navidcy: I dropped it. But we _do_ need to have a unique type for the coordinate otherwise it's not picked up by `validate_dimension_specification` method. I defined a new abstract type.

navidcy:
@simone-silvestri if you can have a look; preview link at the top.
this PR essentially moves https://clima.github.io/ClimaOceanDocumentation/stable/vertical_grids/ to Oceananigans

docs/src/grids.md
```diff
@@ -538,9 +542,294 @@ hidespines!(axx, :t, :r)
...
+    hidespines!(ax, :t, :r)
+end
+
+rowsize!(fig.layout, 1, Relative(0.1))
+rowsize!(fig.layout, 3, Relative(0.1))
+fig
+```
+
+A downside of [`ExponentialCoordinate`](@ref) coordinate is that we don't have tight control on the minimum spacing at the biased edge.
+To obtain a cordinate with a certain minimum spacing we need to play around with the scale ``h`` and the number of cells.
```
simone-silvestri:
```suggestion
To obtain a coordinate with a certain minimum spacing we need to play around with the scale ``h`` and the number of cells.
```

navidcy:
The docstrings and the docs include doctests.

tomchor (commented):
A couple of questions:

1. What's the rationale for moving this into Oceananigans? I'm all for it, but I think in the past when similar things were suggested (like adding a drag BC for example) the philosophy was that it should be left to the user or more specialized packages. Has this type of thinking changed?
2. We should use these in the examples. I believe two of them use stretched coordinates and it'd the perfect place to showcase this. Maybe one of them could use exponential and the other
3. These utilities can be used for the x and y directions, correct?

navidcy:
> 1. What's the rationale for moving this into Oceananigans? I'm all for it, but I think in the past when similar things were suggested (like adding a drag BC for example) the philosophy was that it should be left to the user or more specialized packages. Has this type of thinking changed?

The idea is that since these functionality revolve around grid generation, then they belong in Oceananigans rather than in a child-package.

https://github.com/CliMA/ClimaOcean.jl/pull/565#pullrequestreview-2968209890

https://github.com/CliMA/Oceananigans.jl/pull/4623#issuecomment-3019700112

> 2. We should use these in the examples. I believe two of them use stretched coordinates and it'd the perfect place to showcase this. Maybe one of them could use exponential and the other could use regular-to-stretched.

Sure. Perhaps in another PR? Perhaps a surface-biased exponential for the ocean mixing example? Perhaps the constant-to-stretched coordinate for the bottom mixing example? What do you think?

> 3. These utilities can be used for the x and y directions, correct?

Yes. The examples in the docs here include coordinates in any direction.

tomchor:
> > 2. We should use these in the examples. I believe two of them use stretched coordinates and it'd the perfect place to showcase this. Maybe one of them could use exponential and the other could use regular-to-stretched.
> 
> Sure. Perhaps in another PR? Perhaps a surface-biased exponential for the ocean mixing example? Perhaps the constant-to-stretched coordinate for the bottom mixing example? What do you think?

Either works for me. The idea is to just showcase both these capabilities. I'd argue including these in the examples and adding tests for this should be a part of this PR. Either way, it'd be good to have some tests before merging. I only see them in the docs.

navidcy:
Regarding tests: there are tests in the docs. I'm not sure what other tests to include.

Regarding examples: OK. I'll let few others chime in. I"m not sure which example is better suited for which. I'm a bit worried that if we put it there in an example then users think that this is the "default" that Oceananigans suggests and go with it. My rational for having a tutorial was to showcase how the coordinate looks like and give you an insight how to tweak it.

tomchor:
I don't feel strongly about changing the examples. I think it's better to do it here, but I'm okay with doing it later (or not at all).

I do think we should add _some_ test to `/test` though. In the future we might change the docs and remove the doctests that are there for these new coordinates in some revamp. Actual tests in the `/test` directory are more permanent and reliable.

navidcy:
Gotcha! I can add few tests there, slight alteration of those in the doctests.

navidcy:
So I wanted to add a schematic for $z$ and $z^*$. The schematic would be showing lines of constant $r$ surfaces in $z$. 

Here's a code + fig. I'm not sure if the right panel is the middle one or the right one.

```Julia
using CairoMakie

depth = 20
Lx = 1e3

x = range(-Lx/2, stop=Lx/2, length=200)

σ = Lx/14

# free surface
η = @. -((x/σ)^2 - 1) * exp(-x^2 / 2σ^2)

# bottom
slope_position = -Lx/3
slope = @. 4 * (1 + tanh(-(x - slope_position) / σ))
mountain_position = Lx/3
mountain = @. 2 * exp(-(x - mountain_position)^2 / 2σ^2)
H = @. -depth + slope + mountain


fig = Figure(size=(1000, 400))

ax1 = Axis(fig[1, 1], title=L"$z$", titlesize =20)
ax2 = Axis(fig[1, 2], title=L"$z^*$ ?", titlesize = 20)
ax3 = Axis(fig[1, 3], title=L"$z^*$ ?", titlesize = 20)

for ax in (ax1, ax2, ax3)
    band!(ax, x, H, η, color = (:blue, 0.2))
    band!(ax, x, -depth*1.1, H, color = (:orange, 0.2))
    lines!(ax, x, η, linewidth=5, label="free surface", color=:darkblue)
    lines!(ax, x, H, linewidth=5, label="bottom", color=:darkgrey)
end

for rel in 0:1/5:1
    z = -depth * rel * ones(size(x))
    lines!(ax1, x, z, color=:grey, linestyle=:dash)

    r = -depth * rel # r coordinate ∈ [-Lz, 0]
    z = @. r * (H - η) / H + η
    lines!(ax2, x, z, color=:grey, linestyle=:dash)

    r = H * rel # r coordinate ∈ [-H(x), 0]
    z = @. r * (H - η) / H + η
    lines!(ax3, x, z, color=:grey, linestyle=:dash)
end

fig

save("z-zstar.png", fig)
```

z-zstar


From the definition of $z^*$ I understand that it follows the bathymetry, so that's the right? But that seems like a combination of $z^*$ and $\sigma$ coords?

navidcy:
@tomchor I added tests; thanks for insisting! I will merge if CI passes tho and leave an example using these utils for a future PR.

tomchor:
> @tomchor I added tests; thanks for insisting! I will merge if CI passes tho and leave an example using these utils for a future PR.

Sounds good!

francispoulin:
A very helpful addition, thanks @navidcy !

tomchor (approved):

navidcy:
The distributed hydrostatic free surface tests get an out of memory error at the end...

https://buildkite.com/clima/oceananigans-distributed/builds/9087#0198349d-1c14-4426-86d4-d187c5808331/189-3984

Seems quite irrelevant with this PR. I'm merging but won't tag a release unless they pass.

PR #4627, opened 2025-06-30
Enforce mass-conservation inside domains with Radiation OpenBoundaryConditions

tomchor:
Recently (I think in https://github.com/CliMA/Oceananigans.jl/pull/4578) we found that mass wasn't being exactly conserved whenever PerturbationAdvection was used as an open boundary condition. This PR (which is a draft for now) explores adding a correction to normal velocities at the boundaries right before `solve_for_pressure!()` is called which enforces that no net mass enters or leaves the domain.

At the moment the algorithm is not optimized (suggestions on how to do that are welcome) but it works by:

1. Calculating the net mass (really, momentum) flux into the domain by integrating the velocities at the boundaries (only for boundaries with an OpenBoundaryCondition with a matching scheme)
2. Dividing the correction equally between all the aforementioned boundaries
3. Adding/subtracting the same correction constant to left/right open boundaries

A [MWE measuring the net mass flux](https://github.com/CliMA/Oceananigans.jl/blob/tc/mass-flux-correction/test/test_mass_conservation.jl) into the domain produces this result without the correction:

```julia
[ Info: time: 0.000, max|u|: 1.011, CFL: 0.20, Net flux: 1.9306e-03
[ Info: time: 0.100, max|u|: 1.010, CFL: 0.20, Net flux: 4.1407e-03
[ Info: time: 0.200, max|u|: 1.009, CFL: 0.20, Net flux: 4.6124e-03
[ Info: time: 0.300, max|u|: 1.009, CFL: 0.20, Net flux: 4.4737e-03
[ Info: time: 0.400, max|u|: 1.008, CFL: 0.20, Net flux: 4.1404e-03
[ Info: time: 0.500, max|u|: 1.008, CFL: 0.20, Net flux: 3.7634e-03
[ Info: time: 0.600, max|u|: 1.008, CFL: 0.20, Net flux: 3.3962e-03
[ Info: time: 0.700, max|u|: 1.007, CFL: 0.20, Net flux: 3.0557e-03
[ Info: time: 0.800, max|u|: 1.006, CFL: 0.20, Net flux: 2.7461e-03
[ Info: time: 0.900, max|u|: 1.007, CFL: 0.20, Net flux: 2.4666e-03
[ Info: time: 1.000, max|u|: 1.006, CFL: 0.20, Net flux: 2.2152e-03
```

and this result with the correction:

```julia
[ Info: time: 0.000, max|u|: 1.011, CFL: 0.20, Net flux: 3.8326e-16
[ Info: time: 0.100, max|u|: 1.008, CFL: 0.20, Net flux: -2.7974e-18
[ Info: time: 0.200, max|u|: 1.007, CFL: 0.20, Net flux: 3.4705e-16
[ Info: time: 0.300, max|u|: 1.007, CFL: 0.20, Net flux: 5.3560e-16
[ Info: time: 0.400, max|u|: 1.006, CFL: 0.20, Net flux: -1.0820e-16
[ Info: time: 0.500, max|u|: 1.006, CFL: 0.20, Net flux: -1.6355e-16
[ Info: time: 0.600, max|u|: 1.006, CFL: 0.20, Net flux: -4.9130e-16
[ Info: time: 0.700, max|u|: 1.006, CFL: 0.20, Net flux: 2.5534e-16
[ Info: time: 0.800, max|u|: 1.005, CFL: 0.20, Net flux: -3.0518e-16
[ Info: time: 0.900, max|u|: 1.005, CFL: 0.20, Net flux: 4.2174e-16
[ Info: time: 1.000, max|u|: 1.005, CFL: 0.20, Net flux: 1.2774e-16
```

Notes:

1. This came about in the context of immersed boundaries, but no immersed boundary is necessary to produce these results. Flows with open (radiation) boundaries seem to be divergent even in regular grids, although the FFTSolver seems to handle this well enough that we haven't had problems.
2. @jagoosw I'm curious about your thoughts here. You mentioned before that this might be due to an asymmetry between left and right boundaries when time-stepping the code, but I don't see how the PerturbationAdvection matching scheme (where boundary velocities are computed locally) can be made to always provide velocities at the boundary that are consistent with zero mass flux without having some sort of nonlocal correction.

cc @glwagner @xkykai @ali-ramadhan

### DIFF
--- a/src/Models/NonhydrostaticModels/pressure_correction.jl
+++ b/src/Models/NonhydrostaticModels/pressure_correction.jl
@@ -1,4 +1,118 @@
 import Oceananigans.TimeSteppers: compute_pressure_correction!, make_pressure_correction!
+using Oceananigans.AbstractOperations: Average
+using Oceananigans.Fields: Field
+using Oceananigans.BoundaryConditions: PerturbationAdvection, PerturbationAdvectionOpenBoundaryCondition
+using Oceananigans.ImmersedBoundaries: immersed_inactive_node
+
+const PAOBC = BoundaryCondition{<:Open{<:PerturbationAdvection}}
+const c = Center()
+const f = Face()
+
+# Left boundary averages for normal velocity components
+west_average(u)   = Field(Average(view(u, 1, :, :), dims=(2, 3)))[]
+south_average(v)  = Field(Average(view(v, :, 1, :), dims=(1, 3)))[]
+bottom_average(w) = Field(Average(view(w, :, :, 1), dims=(1, 2)))[]
+
+# Right boundary averages for normal velocity components
+east_average(u)   = Field(Average(view(u, u.grid.Nx + 1, :, :), dims=(2, 3)))[]
+north_average(v)  = Field(Average(view(v, :, v.grid.Ny + 1, :), dims=(1, 3)))[]
+top_average(w)    = Field(Average(view(w, :, :, w.grid.Nz + 1), dims=(1, 2)))[]
+
+function gather_boundary_fluxes(model::NonhydrostaticModel)
+
+    velocities = model.velocities
+    grid = model.grid
+
+    # Get the boundary conditions for the velocities
+    u_bcs = velocities.u.boundary_conditions
+    v_bcs = velocities.v.boundary_conditions
+    w_bcs = velocities.w.boundary_conditions
+
+    # Collect left and right PAOBC boundary conditions into separate lists
+    left_bcs = Symbol[]
+    right_bcs = Symbol[]
+
+    # Initialize fluxes to zero
+    left_flux = zero(grid)
+    right_flux = zero(grid)
+
+    # Calculate flux through left boundaries
+    if u_bcs.west isa PAOBC
+        left_flux += west_average(velocities.u)
+        push!(left_bcs, :west)
+    end
+    if v_bcs.south isa PAOBC
+        left_flux += south_average(velocities.v)
+        push!(left_bcs, :south)
+    end
+    if w_bcs.bottom isa PAOBC
+        left_flux += bottom_average(velocities.w)
+        push!(left_bcs, :bottom)
+    end
+
+    # Calculate flux through right boundaries
+    if u_bcs.east isa PAOBC
+        right_flux += east_average(velocities.u)
+        push!(right_bcs, :east)
+    end
+    if v_bcs.north isa PAOBC
+        right_flux += north_average(velocities.v)
+        push!(right_bcs, :north)
+    end
+    if w_bcs.top isa PAOBC
+        right_flux += top_average(velocities.w)
+        push!(right_bcs, :top)
+    end
+
+    # Calculate total flux (positive means net inflow)
+    total_flux = left_flux - right_flux
+
+    return total_flux, left_bcs, right_bcs
+end
+
+
+
+
+"""
+correct_boundary_mass_flux!(model::NonhydrostaticModel)
+
+Correct boundary mass fluxes for perturbation advection boundary conditions to ensure
+zero net mass flux through each boundary.
+"""
+function correct_boundary_mass_flux!(model::NonhydrostaticModel)
+    velocities = model.velocities
+    grid = model.grid
+
+    total_flux, left_bcs, right_bcs = gather_boundary_fluxes(model)
+
+    # Calculate flux correction per boundary
+    flux_correction_per_boundary = -total_flux / (length(left_bcs) + length(right_bcs))
+
+    # Subtract correction from left boundaries to reduce inflow
+
+    for bc in left_bcs
+        if bc == :west
+            velocities.u[1, :, :] = velocities.u[1, :, :] .- flux_correction_per_boundary
+        elseif bc == :south  
+            velocities.v[:, 1, :] = velocities.v[:, 1, :] .- flux_correction_per_boundary
+        elseif bc == :bottom
+            velocities.w[:, :, 1] = velocities.w[:, :, 1] .- flux_correction_per_boundary
+        end
+    end
+
+    # Add correction to right boundaries to increase outflow
+    for bc in right_bcs
+        if bc == :east
+            velocities.u[size(grid, 1) + 1, :, :] = velocities.u[size(grid, 1) + 1, :, :] .+ flux_correction_per_boundary
+        elseif bc == :north
+            velocities.v[:, size(grid, 2) + 1, :] = velocities.v[:, size(grid, 2) + 1, :] .+ flux_correction_per_boundary
+        elseif bc == :top
+            velocities.w[:, :, size(grid, 3) + 1] = velocities.w[:, :, size(grid, 3) + 1] .+ flux_correction_per_boundary
+        end
+    end
+end
+
+
 
 """
     compute_pressure_correction!(model::NonhydrostaticModel, Δt)
@@ -10,6 +124,9 @@ function compute_pressure_correction!(model::NonhydrostaticModel, Δt)
     # Mask immersed velocities
     foreach(mask_immersed_field!, model.velocities)
     fill_halo_regions!(model.velocities, model.clock, fields(model))
+
+    correct_boundary_mass_flux!(model)
+
     solve_for_pressure!(model.pressures.pNHS, model.pressure_solver, Δt, model.velocities)
     fill_halo_regions!(model.pressures.pNHS)
 
--- a/src/Models/NonhydrostaticModels/pressure_correction.jl
+++ b/src/Models/NonhydrostaticModels/pressure_correction.jl
@@ -1,10 +1,11 @@
 import Oceananigans.TimeSteppers: compute_pressure_correction!, make_pressure_correction!
 using Oceananigans.AbstractOperations: Average
 using Oceananigans.Fields: Field
-using Oceananigans.BoundaryConditions: PerturbationAdvection, PerturbationAdvectionOpenBoundaryCondition
+using Oceananigans.BoundaryConditions: PerturbationAdvection, FlatExtrapolation
 using Oceananigans.ImmersedBoundaries: immersed_inactive_node
 
-const PAOBC = BoundaryCondition{<:Open{<:PerturbationAdvection}}
+const MatchingScheme = Union{FlatExtrapolation, PerturbationAdvection}
+const ROBC = BoundaryCondition{<:Open{<:MatchingScheme}} # Radiation OpenBoundaryCondition
 
 # Left boundary averages for normal velocity components
 west_average(u)   = Field(Average(view(u, 1, :, :), dims=(2, 3)))[]
@@ -26,7 +27,7 @@ function gather_boundary_fluxes(model::NonhydrostaticModel)
     v_bcs = velocities.v.boundary_conditions
     w_bcs = velocities.w.boundary_conditions
 
-    # Collect left and right PAOBC boundary conditions into separate lists
+    # Collect left and right ROBC boundary conditions into separate lists
     left_bcs = Symbol[]
     right_bcs = Symbol[]
 
@@ -35,29 +36,29 @@ function gather_boundary_fluxes(model::NonhydrostaticModel)
     right_flux = zero(grid)
 
     # Calculate flux through left boundaries
-    if u_bcs.west isa PAOBC
+    if u_bcs.west isa ROBC
         left_flux += west_average(velocities.u)
         push!(left_bcs, :west)
     end
-    if v_bcs.south isa PAOBC
+    if v_bcs.south isa ROBC
         left_flux += south_average(velocities.v)
         push!(left_bcs, :south)
     end
-    if w_bcs.bottom isa PAOBC
+    if w_bcs.bottom isa ROBC
         left_flux += bottom_average(velocities.w)
         push!(left_bcs, :bottom)
     end
 
     # Calculate flux through right boundaries
-    if u_bcs.east isa PAOBC
+    if u_bcs.east isa ROBC
         right_flux += east_average(velocities.u)
         push!(right_bcs, :east)
     end
-    if v_bcs.north isa PAOBC
+    if v_bcs.north isa ROBC
         right_flux += north_average(velocities.v)
         push!(right_bcs, :north)
     end
-    if w_bcs.top isa PAOBC
+    if w_bcs.top isa ROBC
         right_flux += top_average(velocities.w)
         push!(right_bcs, :top)
     end
--- a/src/Models/NonhydrostaticModels/pressure_correction.jl
+++ b/src/Models/NonhydrostaticModels/pressure_correction.jl
@@ -98,11 +98,11 @@ function correct_boundary_mass_flux!(model::NonhydrostaticModel)
     # Add extra flux to right boundaries to increase outflow
     for bc in right_bcs
         if bc == :east
-            velocities.u[size(grid, 1) + 1, :, :] = velocities.u[size(grid, 1) + 1, :, :] .+ extra_flux_per_boundary
+            velocities.u[grid.Nx + 1, :, :] = velocities.u[grid.Nx + 1, :, :] .+ extra_flux_per_boundary
         elseif bc == :north
-            velocities.v[:, size(grid, 2) + 1, :] = velocities.v[:, size(grid, 2) + 1, :] .+ extra_flux_per_boundary
+            velocities.v[:, grid.Ny + 1, :] = velocities.v[:, grid.Ny + 1, :] .+ extra_flux_per_boundary
         elseif bc == :top
-            velocities.w[:, :, size(grid, 3) + 1] = velocities.w[:, :, size(grid, 3) + 1] .+ extra_flux_per_boundary
+            velocities.w[:, :, grid.Nz + 1] = velocities.w[:, :, grid.Nz + 1] .+ extra_flux_per_boundary
         end
     end
 end
--- a/src/Models/NonhydrostaticModels/pressure_correction.jl
+++ b/src/Models/NonhydrostaticModels/pressure_correction.jl
@@ -5,8 +5,6 @@ using Oceananigans.BoundaryConditions: PerturbationAdvection, PerturbationAdvect
 using Oceananigans.ImmersedBoundaries: immersed_inactive_node
 
 const PAOBC = BoundaryCondition{<:Open{<:PerturbationAdvection}}
-const c = Center()
-const f = Face()
 
 # Left boundary averages for normal velocity components
 west_average(u)   = Field(Average(view(u, 1, :, :), dims=(2, 3)))[]
@@ -69,8 +67,6 @@ function gather_boundary_fluxes(model::NonhydrostaticModel)
 
     return total_flux, left_bcs, right_bcs
 end
-
-
 
 
 """
--- a/src/Models/NonhydrostaticModels/pressure_correction.jl
+++ b/src/Models/NonhydrostaticModels/pressure_correction.jl
@@ -86,28 +86,27 @@ function correct_boundary_mass_flux!(model::NonhydrostaticModel)
     total_flux, left_bcs, right_bcs = gather_boundary_fluxes(model)
 
     # Calculate flux correction per boundary
-    flux_correction_per_boundary = -total_flux / (length(left_bcs) + length(right_bcs))
+    extra_flux_per_boundary = total_flux / (length(left_bcs) + length(right_bcs))
 
-    # Subtract correction from left boundaries to reduce inflow
-
+    # Subtract extra flux from left boundaries to reduce inflow
     for bc in left_bcs
         if bc == :west
-            velocities.u[1, :, :] = velocities.u[1, :, :] .- flux_correction_per_boundary
+            velocities.u[1, :, :] = velocities.u[1, :, :] .- extra_flux_per_boundary
         elseif bc == :south  
-            velocities.v[:, 1, :] = velocities.v[:, 1, :] .- flux_correction_per_boundary
+            velocities.v[:, 1, :] = velocities.v[:, 1, :] .- extra_flux_per_boundary
         elseif bc == :bottom
-            velocities.w[:, :, 1] = velocities.w[:, :, 1] .- flux_correction_per_boundary
+            velocities.w[:, :, 1] = velocities.w[:, :, 1] .- extra_flux_per_boundary
         end
     end
 
-    # Add correction to right boundaries to increase outflow
+    # Add extra flux to right boundaries to increase outflow
     for bc in right_bcs
         if bc == :east
-            velocities.u[size(grid, 1) + 1, :, :] = velocities.u[size(grid, 1) + 1, :, :] .+ flux_correction_per_boundary
+            velocities.u[size(grid, 1) + 1, :, :] = velocities.u[size(grid, 1) + 1, :, :] .+ extra_flux_per_boundary
         elseif bc == :north
-            velocities.v[:, size(grid, 2) + 1, :] = velocities.v[:, size(grid, 2) + 1, :] .+ flux_correction_per_boundary
+            velocities.v[:, size(grid, 2) + 1, :] = velocities.v[:, size(grid, 2) + 1, :] .+ extra_flux_per_boundary
         elseif bc == :top
-            velocities.w[:, :, size(grid, 3) + 1] = velocities.w[:, :, size(grid, 3) + 1] .+ flux_correction_per_boundary
+            velocities.w[:, :, size(grid, 3) + 1] = velocities.w[:, :, size(grid, 3) + 1] .+ extra_flux_per_boundary
         end
     end
 end
--- a/test/test_mass_conservation.jl
+++ b/test/test_mass_conservation.jl
@@ -1,0 +1,39 @@
+using Oceananigans
+using Oceananigans.BoundaryConditions: PerturbationAdvectionOpenBoundaryCondition
+using Oceananigans.Diagnostics: AdvectiveCFL
+using Printf
+using Random: seed!
+seed!(156)
+
+Δx = Δz = 0.05
+Nx = Nz = round(Int, 2 / Δx)
+grid = RectilinearGrid(size = (Nx, Nz), halo = (4, 4), extent = (1, 1),
+                       topology = (Bounded, Flat, Bounded))
+
+U₀ = 1.0
+inflow_timescale = 1e-1
+outflow_timescale = Inf
+u_boundary_conditions = FieldBoundaryConditions(west = PerturbationAdvectionOpenBoundaryCondition(U₀; inflow_timescale, outflow_timescale),
+                                                east = PerturbationAdvectionOpenBoundaryCondition(U₀; inflow_timescale, outflow_timescale))
+boundary_conditions = (; u = u_boundary_conditions,)
+
+model = NonhydrostaticModel(; grid, boundary_conditions,
+                            timestepper = :RungeKutta3,)
+
+uᵢ(x, z) = U₀ + 1e-4 * rand()
+fill!(model.velocities.u, U₀)
+set!(model, u=uᵢ)
+simulation = Simulation(model; Δt=0.1Δx/U₀, stop_time=1, verbose=false)
+
+cfl_calculator = AdvectiveCFL(simulation.Δt)
+function progress(sim)
+    u, v, w = model.velocities
+    cfl_value = cfl_calculator(model)
+    west_mass_flux = Field(Average(view(u, 1, :, :)))[]
+    east_mass_flux = Field(Average(view(u, grid.Nx+1, :, :)))[]
+    net_mass_flux = east_mass_flux - west_mass_flux
+    @info @sprintf("time: %.3f, max|u|: %.3f, CFL: %.2f, west mass flux: %.6e, east mass flux: %.6e, Net mass flux: %.4e",
+                   time(sim), maximum(abs, u), cfl_value, west_mass_flux, east_mass_flux, net_mass_flux)
+end
+add_callback!(simulation, progress, IterationInterval(20))
+run!(simulation)
--- a/test/test_mass_conservation.jl
+++ b/test/test_mass_conservation.jl
@@ -15,12 +15,16 @@ inflow_timescale = 1e-1
 outflow_timescale = Inf
 u_boundary_conditions = FieldBoundaryConditions(west = PerturbationAdvectionOpenBoundaryCondition(U₀; inflow_timescale, outflow_timescale),
                                                 east = PerturbationAdvectionOpenBoundaryCondition(U₀; inflow_timescale, outflow_timescale))
-boundary_conditions = (; u = u_boundary_conditions,)
+v_boundary_conditions = FieldBoundaryConditions(south = PerturbationAdvectionOpenBoundaryCondition(0; inflow_timescale, outflow_timescale),
+                                                north = PerturbationAdvectionOpenBoundaryCondition(0; inflow_timescale, outflow_timescale))
+w_boundary_conditions = FieldBoundaryConditions(bottom = PerturbationAdvectionOpenBoundaryCondition(0; inflow_timescale, outflow_timescale),
+                                                top = PerturbationAdvectionOpenBoundaryCondition(0; inflow_timescale, outflow_timescale))
+boundary_conditions = (; u = u_boundary_conditions)
 
 model = NonhydrostaticModel(; grid, boundary_conditions,
-                            timestepper = :RungeKutta3,)
+                              timestepper = :RungeKutta3,)
 
-uᵢ(x, z) = U₀ + 1e-4 * rand()
+uᵢ(x, z) = U₀ + 1e-2 * rand()
 fill!(model.velocities.u, U₀)
 set!(model, u=uᵢ)
 simulation = Simulation(model; Δt=0.1Δx/U₀, stop_time=1, verbose=false)
@@ -29,11 +33,9 @@ cfl_calculator = AdvectiveCFL(simulation.Δt)
 function progress(sim)
     u, v, w = model.velocities
     cfl_value = cfl_calculator(model)
-    west_mass_flux = Field(Average(view(u, 1, :, :)))[]
-    east_mass_flux = Field(Average(view(u, grid.Nx+1, :, :)))[]
-    net_mass_flux = east_mass_flux - west_mass_flux
-    @info @sprintf("time: %.3f, max|u|: %.3f, CFL: %.2f, west mass flux: %.6e, east mass flux: %.6e, Net mass flux: %.4e",
-                   time(sim), maximum(abs, u), cfl_value, west_mass_flux, east_mass_flux, net_mass_flux)
+    net_mass_flux2 = ∫∇u = Field(Integral(∂x(u) + ∂z(w)))[]
+    @info @sprintf("time: %.3f, max|u|: %.3f, CFL: %.2f, Net2: %.4e",
+                   time(sim), maximum(abs, u), cfl_value, net_mass_flux2)
 end
 add_callback!(simulation, progress, IterationInterval(20))
 run!(simulation)


glwagner:
The momentum flux is u\*u\*A but the mass flux is u*A --- we're concerned with mass flux, right?

tomchor:
> The momentum flux is u*u*A but the mass flux is u*A --- we're concerned with mass flux, right?

Yes, we're concerned with mass. I meant that I'm measuring velocity at the boundary. I removed the confusing statement from the comment.

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -1,4 +1,116 @@
...
+
+# Left boundary averages for normal velocity components
+west_average(u)   = Field(Average(view(u, 1, :, :), dims=(2, 3)))[]
+south_average(v)  = Field(Average(view(v, :, 1, :), dims=(1, 3)))[]
+bottom_average(w) = Field(Average(view(w, :, :, 1), dims=(1, 2)))[]
+
+# Right boundary averages for normal velocity components
+east_average(u)   = Field(Average(view(u, u.grid.Nx + 1, :, :), dims=(2, 3)))[]
+north_average(v)  = Field(Average(view(v, :, v.grid.Ny + 1, :), dims=(1, 3)))[]
+top_average(w)    = Field(Average(view(w, :, :, w.grid.Nz + 1), dims=(1, 2)))[]
```
glwagner: These should be prebuilt and stored within the boundary condition

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -1,4 +1,116 @@
...
+    end
+
+    # Add extra flux to right boundaries to increase outflow
+    for bc in right_ROBCs
+        if bc == :east
+            velocities.u[grid.Nx + 1, :, :] = velocities.u[grid.Nx + 1, :, :] .+ extra_flux_per_boundary
+        elseif bc == :north
+            velocities.v[:, grid.Ny + 1, :] = velocities.v[:, grid.Ny + 1, :] .+ extra_flux_per_boundary
+        elseif bc == :top
+            velocities.w[:, :, grid.Nz + 1] = velocities.w[:, :, grid.Nz + 1] .+ extra_flux_per_boundary
```
glwagner: I think we can develop something here based on dispatch and an explicit kernel, probably?

src/Models/NonhydrostaticModels/pressure_correction.jl
tomchor: Good idea! I was wondering what the best way to optimize those was. This seems like it. I guess this would require every matching scheme to have the same property

tomchor: Although this would require us to know the complete grid (and field) at the time of building the boundary condition, no?

glwagner: I think `regularize_boundary_condition` does have this info, which "materializes" the boundary condition within the model constructor

glwagner: for example: https://github.com/CliMA/Oceananigans.jl/blob/8b517313710e2f7b6b730230aca36ce736f24a6d/src/BoundaryConditions/field_boundary_conditions.jl#L226

glwagner: Ah but we still need to do a bit of design work here. Because in `regularize_boundary_condition` we don't have access to the field itself --- we only have access to the field name.

glwagner: Another possibility is to create a new model property to hold this info. That might be better, because otherwise we will have a "circular reference" wherein the boundary condition, which is a property of the field, contains a reference to the field itself.

tomchor:
Also for the record, this does seem to allow for the ConjugateGradientSovler to converge when using `ImmersedBoundaryGrids` + `PerturbationAdvectinoOpenBoundaryCondition` at least on simple setups. As an example, [this code](https://github.com/CliMA/Oceananigans.jl/blob/tc/mass-flux-correction/test/test_ibg_mass_conservation.jlf) (with a flat immersed boundary bottom) blows up immediately on main, but runs fine on this branch.

On main

```julia
[ Info: time: 0.000, max|u|: 1.010, CFL: 0.10, Net flux: 1.1067e-03
[ Info: time: NaN, max|u|: NaN, CFL: NaN, Net flux: NaN
[ Info: time: NaN, max|u|: NaN, CFL: NaN, Net flux: NaN
[ Info: time: NaN, max|u|: NaN, CFL: NaN, Net flux: NaN
[ Info: time: NaN, max|u|: NaN, CFL: NaN, Net flux: NaN
[ Info: time = NaN, iteration = 100: NaN found in field u. Stopping simulation.
[ Info: time: NaN, max|u|: NaN, CFL: NaN, Net flux: NaN
```

This branch:

```julia
...
[ Info: time: 48.497, max|u|: 1.000, CFL: 0.10, Net flux: 5.5511e-17
[ Info: time: 48.997, max|u|: 1.000, CFL: 0.10, Net flux: 4.1633e-17
[ Info: time: 49.497, max|u|: 1.000, CFL: 0.10, Net flux: 6.9389e-17
[ Info: time: 49.997, max|u|: 1.000, CFL: 0.10, Net flux: 9.7145e-17
```

src/Models/NonhydrostaticModels/pressure_correction.jl
tomchor: Couldn't we use `model.auxiliary_fields` for this then?

tomchor: > Another possibility is to create a new model property to hold this info. That might be better, because otherwise we will have a "circular reference" wherein the boundary condition, which is a property of the field, contains a reference to the field itself.

Also something to keep in mind, most radiation  boundary conditions used (Orlanski included) use velocities at past times. So figuring out how to have a field's property (the BC) itself have properties that depend on the field might be needed in general.

glwagner:
Should we make this PR not a draft anymore, or is there a reason to keep it as a draft? It seems like something we want to merge ASAP?

tomchor:
> Should we make this PR not a draft anymore, or is there a reason to keep it as a draft? It seems like something we want to merge ASAP?

I made it a draft because I spent no time optimizing it, so it's very slow (as you can imagine due to the allocations being created every time-step). So that was my attempt as signaling that it's not ready for a full review. I'm okay with making it a full PR though.

glwagner:
> > Should we make this PR not a draft anymore, or is there a reason to keep it as a draft? It seems like something we want to merge ASAP?
> 
> I made it a draft because I spent no time optimizing it, so it's very slow (as you can imagine due to the allocations being created every time-step). So that was my attempt as signaling that it's not ready for a full review. I'm okay with making it a full PR though.

Are we ready to work on turning this into a proper implementation, or do we want more time to tinker? Happy to help whenever!

tomchor:
> > > Should we make this PR not a draft anymore, or is there a reason to keep it as a draft? It seems like something we want to merge ASAP?
> > 
> > 
> > I made it a draft because I spent no time optimizing it, so it's very slow (as you can imagine due to the allocations being created every time-step). So that was my attempt as signaling that it's not ready for a full review. I'm okay with making it a full PR though.
> 
> Are we ready to work on turning this into a proper implementation, or do we want more time to tinker? Happy to help whenever!

Gave it some thought and I think we're ready. The actual algorithm is pretty straightforward. All we need to figure out is the software design.

jagoosw:
I read a bit more about this, and it turns out open boundary conditions often don't conserve mass, and a common way to correct is to scale the boundary values to correct so this seems reasonable.

tomchor:
> I read a bit more about this, and it turns out open boundary conditions often don't conserve mass, and a common way to correct is to scale the boundary values to correct so this seems reasonable.

Okay, that's good in the sense that it matches my understanding. Could you point to some of the literature you read though? I'm having some trouble finding relevant literature for these kinds of "details".

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -52,6 +52,7 @@ mutable struct NonhydrostaticModel{TS, E, A<:AbstractArchitecture, G, T, B, R, S
           timestepper :: TS       # Object containing timestepper fields and parameters
       pressure_solver :: S        # Pressure/Poisson solver
      auxiliary_fields :: AF       # User-specified auxiliary fields for forcing functions and boundary conditions
+ boundary_mass_fluxes :: BM       # Container for the average mass fluxes at boundaries
```
glwagner, line 54:
```suggestion
 boundary_mass_fluxes :: BM       # Container for the integrated mass fluxes at boundaries
```

tomchor: I am actually using the average though, because then I can just do stuff like
```julia
velocities.u[1, :, :] .-= extra_flux_per_boundary
```
instead of having to divide by the area when correcting.

glwagner: hmm ok, that makes sense

tomchor:
These modifications seem to slow down the simulation by a lot. The culprit seems to be the average boundary mass flux calculations. Running [this simulation](https://github.com/CliMA/Oceananigans.jl/blob/tc/mass-flux-correction/test/test_mass_conservation.jl), which has only one radiation open boundary (and therefore needs `1*3` calculations of the mass flux through the boundaries per RK3 time-step) leads to almost a factor ~1.8 slowdown.

Having 4 radiation open boundaries leads to a slowdown of ~6 times...

Not sure what can be done to improve this

(I'm only doing this calculation wherever there is a radiation open boundary (i.e. an OpenBoundaryCondition with a matching scheme))

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -65,4 +68,98 @@ function initialize_boundary_mass_fluxes(velocities::NamedTuple)
...
     return boundary_fluxes
+end
+
+@inline function update_boundary_mass_fluxes!(model)
+    for flux in model.boundary_mass_fluxes
+        compute!(flux)
+    end
+end
+
+@inline function gather_boundary_fluxes(model)
```
glwagner:
```suggestion
@inline function open_boundary_mass_fluxes(model)
```

glwagner: This shouldn't have to be @inline

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -65,4 +68,98 @@ function initialize_boundary_mass_fluxes(velocities::NamedTuple)
     end
 
     return boundary_fluxes
+end
+
+@inline function update_boundary_mass_fluxes!(model)
+    for flux in model.boundary_mass_fluxes
+        compute!(flux)
+    end
+end
```
glwagner: I don't think this needs @inline since it is not appearing inside a kernel

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -112,7 +15,7 @@ function compute_pressure_correction!(model::NonhydrostaticModel, Δt)
     foreach(mask_immersed_field!, model.velocities)
     fill_halo_regions!(model.velocities, model.clock, fields(model))
 
-    #correct_boundary_mass_flux!(model)
+    correct_boundary_mass_flux!(model)
```
glwagner: perhaps we can find a desciptor that is more specific than "correct"? Perhaps `enforce_open_boundary_mass_conservation` or something along those lines (eg indicate that this has to do with _open_ boundaries, and mass fluxes / mass conservation)

tomchor: Changed it to your suggestion since I think it's pretty clear

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -47,52 +47,52 @@ function initialize_boundary_mass_fluxes(velocities::NamedTuple)
 
     # Check west boundary (u velocity)
     if u_bcs.west isa OBC
-        west_flux = get_boundary_flux(u_bcs.west, west_average(velocities.u))
+        west_flux = get_boundary_mass_flux(u_bcs.west, west_average(velocities.u))
         boundary_fluxes = merge(boundary_fluxes, (west = west_flux,))
```
glwagner:
```suggestion
        boundary_fluxes = merge(boundary_fluxes, (; west = west_flux))
```

this is the style we agreed to use for singleton NamedTuple (but we can change convention if you like trailing comma better)

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -17,7 +17,7 @@ const FOBC = BoundaryCondition{<:Open{<:Nothing}} # "Fixed-velocity" OpenBoundar
 @inline north_average(v)  = Average(view(v, :, v.grid.Ny + 1, :), dims=(1, 3)) |> Field
 @inline top_average(w)    = Average(view(w, :, :, w.grid.Nz + 1), dims=(1, 2)) |> Field
 
-@inline function get_boundary_flux(bc, boundary_flux_field)
+function get_boundary_mass_flux(bc, boundary_flux_field)
```
glwagner: dispatch might be simpler than `if` statements here, and will allow easier extension of this function

glwagner: Another way to write this code:

```julia
open_boundary_mass_flux(bc, side, field) = NamedTuple()
open_boundary_mass_flux(bc::OBC{Number}, Val{side}, field) where side = NamedTuple(side => bc.condition)
open_boundary_mass_flux(bc::OBC{Number}, Val{side}, field) where side = NamedTuple(side => 0)
# Unsure what ?? needs to be here:
open_boundary_mass_flux(bc::OBC{??}, ::Val{top}, w) = (; top=Average(view(w, :, :, w.grid.Nz + 1), dims=(1, 2)) |> Field)
# etc

function open_boundary_mass_fluxes(velocities)
    u, v, w = velocities
    u_bcs = u.boundary_conditions
    v_bcs = v.boundary_conditions
    w_bcs = w.boundary_conditions
    boundary_mass_fluxes = NamedTuple()
    west_flux_nt = open_boundary_mass_flux(u_bcs.west, Val(:west), u))
    east_flux_nt = open_boundary_mass_flux(u_bcs.east, Val(:east), u))
    # etc
    return merge(...)
end
```

tomchor:
For now I made a change to `mask_immersed_field!()` that allows us to use `mean()` directly and still work with `ImmersedBoundaryGrid`s. After this change, the code is pretty performant, running in about the same time as a simulation without immersed boundaries:

```julia
julia> simulation = create_mass_conservation_simulation(; use_open_boundary_condition = true);
┌ Warning: `PerturbationAdvection` open boundaries matching scheme is experimental and un-tested/validated
└ @ Oceananigans.BoundaryConditions ~/repos/Oceananigans.jl/src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl:71

julia> @btime time_step!(simulation)
  254.261 μs (1589 allocations: 524.57 KiB)

julia> simulation = create_mass_conservation_simulation(; use_open_boundary_condition = false);

julia> @btime time_step!(simulation)
  255.122 μs (1566 allocations: 521.63 KiB)
```

The limitations of this method are that:

1. It doesn't work for stretched grids (although it can be easily extended)
2. It doesn't work on GPUs because, for whatever reason, `mean(interior(u))` throw a scalar index error on GPUs

tomchor:
I modified the algorithm to use a workaround where the calculate the numerical value of the area of the open boundaries when initializing the model and then recalculate the integral of the velocity at every time step. From these measurements we can correct the mass flux. This appears to work both for regular and immersed grids:

```julia
julia> simulation = create_mass_conservation_simulation(; immersed_bottom = nothing);
┌ Warning: `PerturbationAdvection` open boundaries matching scheme is experimental and un-tested/validated
└ @ Oceananigans.BoundaryConditions ~/repos/Oceananigans.jl/src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl:71

julia> run!(simulation)
[ Info: time: 0.000, max|u|: 1.004, CFL: 0.10, Net flux: -2.6368e-16
[ Info: time: 0.156, max|u|: 1.005, CFL: 0.10, Net flux: -2.0813e-16
[ Info: time: 0.313, max|u|: 1.006, CFL: 0.10, Net flux: 1.4917e-16
[ Info: time: 0.469, max|u|: 1.005, CFL: 0.10, Net flux: 3.5037e-16
[ Info: time: 0.625, max|u|: 1.005, CFL: 0.10, Net flux: -5.2009e-17
[ Info: time: 0.781, max|u|: 1.005, CFL: 0.10, Net flux: -6.2439e-17
[ Info: time: 0.938, max|u|: 1.006, CFL: 0.10, Net flux: -5.5434e-17

julia> simulation = create_mass_conservation_simulation(; immersed_bottom = GridFittedBottom(-0.6));
┌ Warning: `PerturbationAdvection` open boundaries matching scheme is experimental and un-tested/validated
└ @ Oceananigans.BoundaryConditions ~/repos/Oceananigans.jl/src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl:71

julia> run!(simulation)
[ Info: time: 0.000, max|u|: 1.005, CFL: 0.10, Net flux: 4.1628e-17
[ Info: time: 0.156, max|u|: 1.005, CFL: 0.10, Net flux: 1.3824e-17
[ Info: time: 0.313, max|u|: 1.005, CFL: 0.10, Net flux: -3.8107e-17
[ Info: time: 0.469, max|u|: 1.005, CFL: 0.10, Net flux: -1.0760e-16
[ Info: time: 0.625, max|u|: 1.004, CFL: 0.10, Net flux: -8.3230e-17
[ Info: time: 0.781, max|u|: 1.004, CFL: 0.10, Net flux: 9.7115e-17
[ Info: time: 0.938, max|u|: 1.005, CFL: 0.10, Net flux: 1.0337e-17
```

The function above comes from the validation script I'm introducing with this PR. (Note that using the PartialCellBottom sometimes produces NaNs; not sure why)

This method using the integral also appears to be reasonably performant. Using the same validation script:

```julia
julia> simulation = create_mass_conservation_simulation(; use_open_boundary_condition=false); # baseline w/o open boundaries

julia> @btime time_step!(simulation)
  574.736 μs (4428 allocations: 1.50 MiB)

julia> simulation = create_mass_conservation_simulation(; use_open_boundary_condition=true);
┌ Warning: `PerturbationAdvection` open boundaries matching scheme is experimental and un-tested/validated
└ @ Oceananigans.BoundaryConditions ~/repos/Oceananigans.jl/src/BoundaryConditions/perturbation_advection_open_boundary_matching_scheme.jl:71

julia> @btime time_step!(simulation)
  669.849 μs (4689 allocations: 1.52 MiB)
```
so there's a slowdown of ~15% when using 1 radiation open boundary condition, which isn't great but it also isn't too bad as a starting point.

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -1,26 +1,59 @@
...
+# Right boundary integrals for normal velocity components
+@inline east_integral(u)   = Field(Integral(view(u, u.grid.Nx + 1, :, :), dims=(2, 3)))
+@inline north_integral(v)  = Field(Integral(view(v, :, v.grid.Ny + 1, :), dims=(1, 3)))
+@inline top_integral(w)    = Field(Integral(view(w, :, :, w.grid.Nz + 1), dims=(1, 2)))
+
+function get_west_area(grid)
+    f = XFaceField(grid); set!(f, 1) # Create an XFaceField with all ones
+    ∫f = Field(Integral(f, dims=(2, 3))) # Integrate over y and z dimensions
+    @allowscalar return ∫f[1, 1, 1]
+end
```
glwagner, line 15: Isn't this replicating things that are already done within `Average`? I'm thinking of these lines:

https://github.com/CliMA/Oceananigans.jl/blob/b400322ce9602a2faadc18c4c6a085f1bcc5aa38/src/AbstractOperations/metric_field_reductions.jl#L53-L58

tomchor: Yes, but for some reason it's not very performant there. Maybe because the normalization (i.e. `dx / L`) is embedded pointwise. This creates more operations than what I do which is integrate the field once, and only then divide by L... ?

tomchor: Actually, I think it's because L is a reduced Field (I think). Maybe extracting the `Number` there is enough to speed things up. I'll apply this and report back.

tomchor: See https://github.com/CliMA/Oceananigans.jl/pull/4639/files#r2193145570

glwagner: Trying to understand this idea. If `L` is depth and we are doing a vertical average, do we need to allow `L` to be a `Field`? Because the depth varies in `x, y`?

tomchor: I might be missing some particularities of the `Average()` implementation (maybe we should discuss that in #4639), but for the purposes of surface average needed here, it's useful to interpret it as (taking $u$ as an example to make notation simpler):

```math
\bar u = \frac{\int^{yz} u \, dydz}{\int^{yz} dydz} = \frac{\int^{yz} u \,{dydz}}{A_{yz}}
```
So the area shouldn't change unless we're dealing with mutable coordinate systems (which I'm assuming for simplicity that we aren't).

A note here is that if we were calculating the xz-average over the whole domain, then obviously $A_{yz}$ would be a function of $x$, but we only want the average at the boundary.

glwagner: Certainly there are specific cases where the normalization is a constant / number. I'm just wondering if this conclusion is general. For example a vertical average in a domain with variable depth looks like

$$ \langle u \rangle = \frac{\int^z u dz}{\int^z dz} = \frac{1}{H(x, y)} \int_{-H(x, y)}^0 u dz $$

where $H = \int^z u dz$ is the depth.

tomchor:
I think this PR is ready (just gotta merge main and wait for tests to pass). Here is a simulation I added in the validation with PerturbationAdvectionOBC in both west and east boundaries:

https://github.com/user-attachments/assets/f21f917b-73d5-4169-9531-6880aed0a8fd

Despite some spikes in pointwise divergence, the domain-average divergence remains at ~1e-21 throughout the whole simulation. This also makes the ConjugateGradient poisson solver converge much faster. Often it converges in 2 or 3 iterations, and only occasionally it takes more than 5 (at least for the simulation above). I also tried "breaking" the simulation above in different ways (more complex bathymetry, more noise, etc.) and wasn't able to, so I think that bodes well.

There is a bit of an added cost due to the 2D integral required, but it's not prohibitive. A 32^2 simulation without immersed boundaries and without open boundaries on the GPU is timed as

```julia
BenchmarkTools.Trial: 6409 samples with 1 evaluation per sample.
 Range (min … max):  558.528 μs … 27.158 ms  ┊ GC (min … max): 0.00% … 95.58%
 Time  (median):     755.753 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   778.158 μs ±  1.256 ms  ┊ GC (mean ± σ):  7.69% ±  4.63%

    ▂▄▂                                        ▃▇█▇▇▃▁▂▃▁▁      
  ▂▄████▆▅▄▃▃▂▂▂▂▂▂▁▂▂▂▂▁▂▁▂▂▂▂▂▂▂▂▂▁▂▂▁▂▂▂▂▃▃▆███████████▇▅▃▃ ▄
  559 μs          Histogram: frequency by time          806 μs <

 Memory estimate: 549.33 KiB, allocs estimate: 1948.
```

While the same simulation with two open radiation boundary clocks in at:

```julia
BenchmarkTools.Trial: 4694 samples with 1 evaluation per sample.
 Range (min … max):  811.526 μs … 28.853 ms  ┊ GC (min … max): 0.00% … 93.95%
 Time  (median):       1.036 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.063 ms ±  1.382 ms  ┊ GC (mean ± σ):  6.43% ±  4.77%

     ▂▂▂                                     ▃▆▆█▆▄▃▃▂          
  ▂▄▆████▆▅▄▃▃▂▂▂▂▂▂▂▂▁▂▂▂▂▂▁▂▂▁▁▂▂▂▂▂▂▂▂▃▄▅▇██████████▇▆▅▄▄▃▃ ▄
  812 μs          Histogram: frequency by time         1.11 ms <

 Memory estimate: 622.25 KiB, allocs estimate: 3021.
```
so this is about a 35% slowdown. That said, this fixes what I'd say is the last major hurdle in running open boundaries in realistic configurations, so I vote for merging it asap. After that I can open another PR to try and speed up performance here (in addition to https://github.com/CliMA/Oceananigans.jl/pull/4639, in which I'm already exploring performance). This slowdown shouldn't affect simulations with no radiation open boundaries.

test/test_boundary_conditions_integration.jl
```diff
@@ -227,7 +227,8 @@ function test_open_boundary_condition_mass_conservation(arch, FT, boundary_condi
     Δt = 0.1 * minimum_zspacing(grid) / maximum(abs, u)
     simulation = Simulation(model; stop_time=1, Δt, verbose=false)
 
-    ∫∇u = Field(Integral(∂x(u) + ∂y(v) + ∂z(w)))
+    ∇u = Field(∂x(u) + ∂y(v) + ∂z(w))
+    ∫∇u = Field(Integral(∇u))
```
glwagner, line 230: You should be able to define `Integral` of the divergence directly without using an intermediate `Field`.

tomchor: That's what I had before. It compiled on my local machines, but it wasn't compiling for GPU on the CI so I changed it in the last commit. It seems to work now.

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -0,0 +1,238 @@
...
+    return interior(∫dA, 1, grid.Ny+1, 1) |> maximum
+end
+
+function get_bottom_area(grid)
+    dA = GridMetricOperation((Center, Center, Face), Az, grid)
+    ∫dA = sum(dA, dims=(1, 2))
+    return interior(∫dA, 1, 1, 1) |> maximum
+end
+
+function get_top_area(grid)
```
glwagner, line 41: why do we need these? Are we supporting open boundaries at the top and bottom?

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -0,0 +1,238 @@
...
+function get_west_area(grid)
+    dA = GridMetricOperation((Face, Center, Center), Ax, grid)
+    ∫dA = sum(dA, dims=(2, 3))
+    return interior(∫dA, 1, 1, 1) |> maximum
+end
+
+function get_east_area(grid)
+    dA = GridMetricOperation((Face, Center, Center), Ax, grid)
+    ∫dA = sum(dA, dims=(2, 3))
+    return interior(∫dA, grid.Nx+1, 1, 1) |> maximum
```
glwagner: Do we reduce with `maximum` just as a way of extracting a single value, or something more going on here?

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
tomchor: Yes.

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
tomchor: Just to extract a single value. I found `maximum` to be the simplest from all the options I tested. I can add a comment to make it clearer, or extract it another way.

glwagner: I guess just indexing into it might be easier to interpret --- `∫dA[grid.Nx+1, 1, 1]`

tomchor: That's what it used to be but I need `@allowscalar` from CUDA. Since the refactor of CUDA I need an explicit export for that, so I chose to just use `maximum` since it's simpler.

Would it help if I defined a local function as `to_scalar(a) = maximum(a)`, so that at least we can write `interior(∫dA, grid.Nx+1, 1, 1) |> to_scalar`?

Or maybe there's CuArray-compliant way to do `∫dA[grid.Nx+1, 1, 1]` that I'm missing?

glwagner: You need `@allowscalar`. This is also what is invoked inside `maximum` when it returns a scalar.

glwagner: `maximum` does more than just a scalar index, it will allocate memory for the result, launch a kernel that computes a reduction, and then output the result using `@allowscalar` to transfer the data from GPU to CPU. If you simply index into the array with `@allowscalar` in the first place, you avoid the extraneous operations and get the info you need directly.

glwagner: Oh and sorry. `@allowscalar` is not actually defined by CUDA.jl --- it is extended by it. `@allowscalar` is defined in `GPUArraysCore` and this is the correct place to get it from.

tomchor: Ah, I see. I didn't know that. I'll make that change and import it from `GPUArraysCore`.

tomchor: Done! I'm resolving this

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -0,0 +1,239 @@
...
+    return @allowscalar ∫dA[1, 1, grid.Nz+1]
+end
+
+# Left boundary integrals for normal velocity components
+@inline u_west_integral(u)   = Field(Integral(view(u, 1, :, :), dims=(2, 3)))
+@inline v_south_integral(v)  = Field(Integral(view(v, :, 1, :), dims=(1, 3)))
+@inline w_bottom_integral(w) = Field(Integral(view(w, :, :, 1), dims=(1, 2)))
+
+# Right boundary integrals for normal velocity components
+@inline u_east_integral(u)   = Field(Integral(view(u, u.grid.Nx + 1, :, :), dims=(2, 3)))
```
glwagner:
```suggestion
@inline net_east_flux(u)   = Field(Integral(view(u, u.grid.Nx + 1, :, :), dims=(2, 3)))
```

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -0,0 +1,239 @@
...
+enforce_open_boundary_mass_conservation!(model::NonhydrostaticModel)
+
+Correct boundary mass fluxes for perturbation advection boundary conditions to ensure
+zero net mass flux through each boundary.
+"""
+function enforce_open_boundary_mass_conservation!(model)
+    u, v, w = model.velocities
+    grid = model.grid
+
+    total_mass_inflow = open_boundary_mass_inflow(model)
```
glwagner: I think things like this read better if you pair a verbose name with a symbol, so here's an idea:

```julia
∮u_dA = open_boundary_mass_inflow(model)
A = model.boundary_mass_fluxes.total_area_matching_scheme_boundaries
 A⁻¹_∮u dA = ∮u_dA / A
```

or something like that. 

because otherwise we have two names

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -0,0 +1,239 @@
...
+
+    # Subtract flux through right boundaries
+    total_flux -= open_boundary_mass_flux(model, u_bcs.east, Val(:east), u)
+    total_flux -= open_boundary_mass_flux(model, v_bcs.north, Val(:north), v)
+    total_flux -= open_boundary_mass_flux(model, w_bcs.top, Val(:top), w)
+
+    return total_flux
+end
+
+correct_left_boundary_mass_flux!(u, bc::OBC, ::Val{:west},   avg_mass_inflow) = interior(u, 1, :, :) .-= avg_mass_inflow
```
glwagner: here and in some other places, I think defining a mathematical notation will help make this easier to read: eg `δu` instead of `avg_mass_inflow`. But I am ok either way

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -0,0 +1,239 @@
...
+correct_left_boundary_mass_flux!(v, bc::OBC, ::Val{:south},  avg_mass_inflow) = interior(v, :, 1, :) .-= avg_mass_inflow
+correct_left_boundary_mass_flux!(w, bc::OBC, ::Val{:bottom}, avg_mass_inflow) = interior(w, :, :, 1) .-= avg_mass_inflow
+correct_left_boundary_mass_flux!(velocity, bc::IOBC, ::Val{:west}, avg_mass_inflow) = nothing
```
glwagner:
```suggestion
correct_left_boundary_mass_flux!(u, bc::IOBC, ::Val{:west}, avg_mass_inflow) = nothing
```

this is ok in my opinion. or `U` or `q`

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
glwagner: or `east_mass_flux`, which I notice you call it below

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -0,0 +1,239 @@
...
+initialize_boundary_mass_flux(v, bc::FIOBC, ::Val{:south})  = (; south_mass_flux = bc.condition * get_south_area(v.grid), south_area = get_south_area(v.grid))
+initialize_boundary_mass_flux(v, bc::FIOBC, ::Val{:north})  = (; north_mass_flux = bc.condition * get_north_area(v.grid), north_area = get_north_area(v.grid))
+initialize_boundary_mass_flux(w, bc::FIOBC, ::Val{:bottom}) = (; bottom_mass_flux = bc.condition * get_bottom_area(w.grid), bottom_area = get_bottom_area(w.grid))
+initialize_boundary_mass_flux(w, bc::FIOBC, ::Val{:top})    = (; top_mass_flux = bc.condition * get_top_area(w.grid), top_area = get_top_area(w.grid))
+
+initialize_boundary_mass_flux(velocity, ::Nothing, side) = NamedTuple()
+initialize_boundary_mass_flux(velocity, bc, side) = NamedTuple()
+
+needs_mass_flux_correction(::IOBC) = false
+needs_mass_flux_correction(bc::OBC) = bc.classification.matching_scheme !== nothing ? true : false
```
glwagner:
```suggestion
needs_mass_flux_correction(bc::OBC) = bc.classification.matching_scheme !== nothing
```

glwagner: ie this returns the desired Bool directly (a good principle is "never use an if statement to return a Bool")

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -0,0 +1,239 @@
...
+    boundary_fluxes = merge(boundary_fluxes, (; left_matching_scheme_boundaries,
+                                                right_matching_scheme_boundaries,
+                                                total_area_matching_scheme_boundaries))
+
+    return boundary_fluxes
+end
+
+update_open_boundary_mass_fluxes!(model) = map(compute!, model.boundary_mass_fluxes)
+
+open_boundary_mass_flux(model, bc::OBC, ::Val{:west}, u) = model.boundary_mass_fluxes.west_mass_flux |> maximum
```
glwagner: are these `maximum` needed or is this tantamount to an allowscalar

test/test_boundary_conditions_integration.jl
```diff
@@ -448,6 +427,34 @@ test_boundary_conditions(C, FT, ArrayType) = (integer_bc(C, FT, ArrayType),
             @info "  Testing open boundary conditions [$A, $FT]..."
             test_flat_extrapolation_open_boundary_conditions(arch, FT)
             test_perturbation_advection_open_boundary_conditions(arch, FT)
+
+            # Only PerturbationAdvectionOpenBoundaryCondition
+            U₀ = 1.0
```
glwagner:
```suggestion
            U₀ = 1
```

in case an impressionable user reads it

test/test_conjugate_gradient_poisson_solver.jl
```diff
@@ -291,17 +295,30 @@ end
...
+            for (underlying_grid_name, underlying_grid) in underlying_grids
+                @info "  Testing $underlying_grid_name with different bottom types and open boundary conditions [$(typeof(arch))]..."
+                @test test_conjugate_gradient_with_immersed_boundary_grid_and_open_boundaries(underlying_grid, DiagonallyDominantPreconditioner(), bottom)
+                @test test_conjugate_gradient_with_immersed_boundary_grid_and_open_boundaries(underlying_grid, fft_poisson_solver(underlying_grid), bottom)
+            end
         end
 
-        @testset "Conjugate gradient solver with ImmersedBoundaryGrid on a stretched grid and open boundary conditions [$(typeof(arch))]" begin
-            underlying_grid = underlying_grids["stretched grid"]
-            @test test_conjugate_gradient_with_immersed_boundary_grid_and_open_boundaries(underlying_grid, DiagonallyDominantPreconditioner())
```
glwagner, line 303: these function names are too long --- this is not blocking, but if you can improve please do

glwagner (approved):
I think this needs a good proofread because I found a few details to fix. Please read through. Reduce verbosity if possible, very long lines / variable names can impact the legibility of code

test/test_boundary_conditions_integration.jl
tomchor: Changed

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
tomchor: Replaced with allowscalar.

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
tomchor: Done

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
tomchor: Done

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
tomchor: I preferred to use `*_inflow` in some of these names because it makes it clear that a positive value means added mass to the domain. Otherwise this could be a source of confusion I think.

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
tomchor: Changed

test/test_conjugate_gradient_poisson_solver.jl
tomchor: Shortened it to `test_CGSolver_with_immersed_boundary_and_open_boundaries`

tomchor:
Addressed all your comments and gave it a thorough proofread afterwards, as per your suggestion.

tomchor:
https://github.com/CliMA/Oceananigans.jl/pull/4668 seems to have really sped up things. Here are the benchmarks for a simulation without and then with two open boundaries before #4668 was merged:

```julia
julia> b_without_open
BenchmarkTools.Trial: 5785 samples with 1 evaluation per sample.
 Range (min … max):  662.920 μs …   8.982 ms  ┊ GC (min … max): 0.00% … 85.53%
 Time  (median):     871.306 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   861.066 μs ± 559.721 μs  ┊ GC (mean ± σ):  4.45% ±  6.25%

       ▂▄▅▅▃                         ▁▄█▆▄▂▁                     
  ▁▂▃▄▇█████▇▆▅▄▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▂▃▄████████▆▅▄▃▃▃▃▂▃▂▂▂▂▂▂▁▁▁▁ ▃
  663 μs           Histogram: frequency by time         1.02 ms <

 Memory estimate: 731.53 KiB, allocs estimate: 2104.

julia> b_with_open
BenchmarkTools.Trial: 5104 samples with 1 evaluation per sample.
 Range (min … max):  754.581 μs …  12.483 ms  ┊ GC (min … max): 0.00% … 89.36%
 Time  (median):     972.037 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   975.329 μs ± 620.934 μs  ┊ GC (mean ± σ):  4.54% ±  6.48%

        ▄███▆▄▁                 ▂▅▇▇▇▇▇▃                         
  ▂▂▃▄▆████████▇▅▅▄▃▃▃▃▂▂▂▂▂▃▄▅▇█████████▆▅▄▄▅▃▃▃▃▃▃▃▃▂▂▃▂▂▂▁▂▂ ▄
  755 μs           Histogram: frequency by time          1.2 ms <

 Memory estimate: 746.67 KiB, allocs estimate: 2382.
```

and here are the results after it was merged:

```julia
julia> b_without_open
BenchmarkTools.Trial: 6946 samples with 1 evaluation per sample.
 Range (min … max):  532.364 μs …   5.758 ms  ┊ GC (min … max): 0.00% … 78.44%
 Time  (median):     626.273 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   716.930 μs ± 406.932 μs  ┊ GC (mean ± σ):  4.45% ±  6.98%

          ▁▅██▆▂                           ▁▃▂                   
  ▁▁▂▂▂▂▃▅███████▆▄▃▃▂▂▂▂▁▁▁▁▁▁▁▂▂▂▂▂▂▃▃▅▇████▇▅▃▂▂▂▂▁▂▁▂▁▁▁▁▁▁ ▃
  532 μs           Histogram: frequency by time          910 μs <

 Memory estimate: 715.31 KiB, allocs estimate: 1776.

julia> b_with_open
BenchmarkTools.Trial: 6801 samples with 1 evaluation per sample.
 Range (min … max):  539.551 μs …   8.671 ms  ┊ GC (min … max): 0.00% … 88.88%
 Time  (median):     661.378 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   731.322 μs ± 411.323 μs  ┊ GC (mean ± σ):  4.48% ±  7.13%

       ▃██▄▂▁                   ▁                                
  ▂▂▃▄▇██████▇▅▄▅▄▄▄▅▄▄▃▃▃▃▃▃▄▆███▆▄▄▃▃▃▂▂▃▃▂▃▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁ ▃
  540 μs           Histogram: frequency by time         1.04 ms <

 Memory estimate: 723.58 KiB, allocs estimate: 1844.
```

The difference is now much smaller and, perhaps more importantly, both are faster :tada:

Issue #4630, opened 2025-07-01
Should we change `closure_constant()` to `closure_coefficient()`?


tomchor:
I think `closure_constant()` is a bit of a misnomer. It also leads to some misleading lines such as the ones below:

https://github.com/CliMA/Oceananigans.jl/blob/73ddc2b93361378ece386ae9c15fd4f0e083a29a/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/smagorinsky.jl#L108-L109

These lines imply that we have a "coefficient" for the constant Smag (where it is a constant), but a "constant" for the dynamic Smag (where the point is that the "constant" isn't a constant). A better name imo would be `closure_coefficient()`.

Sorry for not bringing this up sooner, but I was not aware of https://github.com/CliMA/Oceananigans.jl/pull/3370 until I was it in yesterday's release merged PRs.

cc @glwagner

glwagner:
There's no need to apologize for anything. Seems like a positive change.

PR #4631, opened 2025-07-01
Change `closure_constant()` to `closure_coefficient()`

tomchor:
Closes https://github.com/CliMA/Oceananigans.jl/issues/4630#issuecomment-3024509112

### DIFF
--- a/src/TurbulenceClosures/TurbulenceClosures.jl
+++ b/src/TurbulenceClosures/TurbulenceClosures.jl
@@ -182,8 +182,8 @@ include("turbulence_closure_implementations/scalar_biharmonic_diffusivity.jl")
 
 # Dispatch on the type of the user-provided AMD model constant.
 # Only numbers, arrays, and functions supported now.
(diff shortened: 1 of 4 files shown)


glwagner (approved):

glwagner:
PS you don't always need to merge main while CI is running for small PRs like this one (we just need to see that turbulence closure tests pass for example)

tomchor:
> PS you don't always need to merge main while CI is running for small PRs like this one (we just need to see that turbulence closure tests pass for example)

Got it. But then you have to be the one that merges it, right? At least I know I can't merge until all the required tests pass.

glwagner:
> > PS you don't always need to merge main while CI is running for small PRs like this one (we just need to see that turbulence closure tests pass for example)
> 
> Got it. But then you have to be the one that merges it, right? At least I know I can't merge until all the required tests pass.

yes and that's fine, happy to

tomchor:
Sounds good. The turbulence closure tests seem to be passing if you wanna merge.

PR #4637, opened 2025-07-06
Clock enhancements

navidcy:
This PR brings the clock-related changes from #4250

### DIFF
--- a/src/TimeSteppers/clock.jl
+++ b/src/TimeSteppers/clock.jl
@@ -33,7 +33,7 @@ end
 """
     Clock(; time, last_Δt=Inf, last_stage_Δt=Inf, iteration=0, stage=1)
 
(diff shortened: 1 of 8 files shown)


src/TimeSteppers/clock.jl
```diff
@@ -49,6 +40,38 @@ function Clock(; time,
...
+    clock.last_stage_Δt = Inf
+    return nothing
+end
+
+"""
+    set_clock!(clock::Clock, new_clock::Clock)
+
+Set `clock` to the `new_clock`.
+"""
+function set_clock!(clock::Clock, new_clock::Clock)
```
navidcy, line 57: Perhaps even better is to just add a method to `set!`?

```suggestion
import  Oceananigans: set!

function set!(clock::Clock, new_clock::Clock)
```

Issue #4633, opened 2025-07-04
Reducing a Field with `mean()` is much slower than reducing its underlying array


tomchor:
I've been noticing this slowdown in https://github.com/CliMA/Oceananigans.jl/pull/4627, where we need to calculate a 2D `Average` for each boundary with a radiation boundary condition. This computation alone proved to be a significant bottleneck, significantly slowing down the simulations.

Here's an example script:

```julia
using Oceananigans
using Statistics: mean
arch = GPU()
grid = RectilinearGrid(arch, size=(8, 8, 8), extent=(1, 1, 1), halo=(1, 1, 1));

u = XFaceField(grid);
ū = Average(view(u, 1, :, :), dims=(2, 3)) |> Field;
```

On GPU, the timings based on this (after running a few times to precompile everything) are:
```
julia> @time compute!(ū);
  0.001258 seconds (1.35 k allocations: 117.500 KiB)

julia> @time mean(interior(u, 1, :, :).parent);
  0.000399 seconds (158 allocations: 5.031 KiB)
```

The issue also persists on the CPU:
```julia
julia> @time compute!(ū);
  0.000697 seconds (914 allocations: 94.312 KiB)

julia> @time mean(interior(u, 1, :, :).parent);
  0.000195 seconds (62 allocations: 2.109 KiB)
```
Basically in both cases there's a ~3.5x slowdown with many more allocations using `Average`.

Another interesting thing is that `Integral` is much faster, and comparable to `sum`:
```julia
julia> ∫u = Integral(view(u, 1, :, :), dims=(2, 3)) |> Field;

julia> @time compute!(∫u); # On the GPU, after precompilation
  0.000335 seconds (185 allocations: 10.781 KiB)

julia> @time sum(interior(u, 1, :, :).parent);
  0.000392 seconds (158 allocations: 5.031 KiB)
```

cc @glwagner @simone-silvestri @ali-ramadhan

glwagner:
What are the timings without using .parent ?

glwagner:
The difference with Integral probably shows what the solution is. The operation for Average is defined with an abstract operation that everywhere divides by the “volume” of the average (a length, area, or volume depending on dimensionality).

It’s straightforward to convert Average into two parts: first compute Integral, and then normalize by the volume in place.

tomchor:
> What are the timings without using .parent ?

[...]

PR #4639, opened 2025-07-07
Add methods to bypass `ConditionalOperation` when `condition==nothing`

tomchor:
Related to https://github.com/CliMA/Oceananigans.jl/issues/4633

Wrote this short script to time the slowdown we have (compared to directly accessing `interior(u)` when using `mean(u)` and `Integral(u)`:

```julia
using Oceananigans, BenchmarkTools

grid = RectilinearGrid(size=(8, 8, 8), extent=(1, 1, 1), halo=(1, 1, 1));
u = XFaceField(grid);

bmean_baseline = @benchmark mean(interior(u))
bmean = @benchmark mean(u)
slowdown_mean = minimum(bmean.times) / minimum(bmean_baseline.times)
@info "Slowdown using mean(u): $slowdown_mean"

ū = Field(Average(u))
bavg = @benchmark compute!(ū)
bsum_baseline = @benchmark sum(interior(u))
slowdown_avg = minimum(bavg.times) / minimum(bsum_baseline.times)
@info "Slowdown using compute!(ū): $slowdown_avg"
```

On main this produces

```julia
[ Info: Slowdown using mean(u): 8.200285408490902
[ Info: Slowdown using compute!(ū): 6.156982321501987
```

On this branch
```julia
[ Info: Slowdown using mean(u): 5.059299971073185
[ Info: Slowdown using compute!(ū): 1.8239949351060463
```

So this definitely speeds things up (when `ConditionalOperation` isn't needed), although there is more work to be done.


src/AbstractOperations/conditional_operations.jl
```diff
@@ -151,6 +151,11 @@ end
     return ifelse(conditioned, value, co.mask)
 end
 
+# Fallbacks for reductions
+for reduction! in (:sum!, :maximum!, :minimum!, :all!, :any!, :prod!)
+    @eval Base.$(reduction!)(id, int_r, co::NoConditionCO, kwargs...) = Base.$(reduction!)(id, int_r, co.operand; kwargs...)
```
glwagner:
```suggestion
    @eval Base.$(reduction!)(f, int_r, co::NoConditionCO, kwargs...) = Base.$(reduction!)(f, int_r, co.operand; kwargs...)
```

glwagner:
If necessary we should do this, but I wonder if a better design would simply avoid creating the pointless ConditionalOperation in the first place?

glwagner:
Something is sketchy because the conditional operation should have no effect due to the constant here:

https://github.com/CliMA/Oceananigans.jl/blob/b400322ce9602a2faadc18c4c6a085f1bcc5aa38/src/AbstractOperations/conditional_operations.jl#L156

glwagner:
and moreover for example: https://github.com/CliMA/Oceananigans.jl/blob/b400322ce9602a2faadc18c4c6a085f1bcc5aa38/src/AbstractOperations/conditional_operations.jl#L146

although I am not sure that method is being hit properly, maybe it isn't.

src/AbstractOperations/metric_field_reductions.jl
```diff
@@ -43,24 +44,15 @@ for information and examples using `condition` and `mask` kwargs.
...
-        L = sum(metric; condition, mask, dims)
-
-        # Construct summand of the Average
-        L⁻¹_field_dx = field * dx / L
-
-        operand = condition_operand(L⁻¹_field_dx, condition, mask)
+    # Compute "size" (length, area, or volume) of averaging region
+    metric = GridMetricOperation(location(field), dx, field.grid)
+    ∫dx = @allowscalar sum(metric; condition, mask, dims)[] # Pluck out L as a Number
+    ∫dx = convert(eltype(field.grid), ∫dx)
```
tomchor: This change (which, importantly, calculated the total length/area/volume as a Number) gives me no slowdown most of the time. For example here's the benchmark for `mean(interior())`:

```julia
julia> @benchmark mean(interior(u))
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  15.223 μs … 82.285 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     15.894 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.362 μs ±  1.748 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▁█▄                                                        
  ▁▃███▇▄▃▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  15.2 μs         Histogram: frequency by time        24.3 μs <

 Memory estimate: 1.92 KiB, allocs estimate: 58.
```
and here's the benchmark using `Average()` (in most cases):
```julia
julia> @benchmark compute!(ū)
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  15.117 μs …  4.930 ms  ┊ GC (min … max): 0.00% … 98.86%
 Time  (median):     15.939 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.929 μs ± 49.189 μs  ┊ GC (mean ± σ):  2.88% ±  0.99%

    ▅█▅▂                                                       
  ▂▅█████▆▄▄▄▄▄▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  15.1 μs         Histogram: frequency by time        25.3 μs <

 Memory estimate: 5.94 KiB, allocs estimate: 58.
```

However, for some reason sometimes I'll run the exact same code to calculate these benchmarks and it'll produce something that's much slower and with much more allocations for Average:

```julia
julia> @benchmark compute!(ū)
BenchmarkTools.Trial: 4213 samples with 1 evaluation per sample.
 Range (min … max):  762.176 μs …   3.352 ms  ┊ GC (min … max): 0.00% … 37.45%
 Time  (median):       1.002 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.182 ms ± 550.772 μs  ┊ GC (mean ± σ):  8.45% ± 11.50%

  ▃▃  ▂██▄▁                                                  ▁  ▁
  ██▇██████▇▇▇▆▅▇▆▅▇▅▇▆▆▆▅▆▆▆▆▆▆▅▃▃▂▃▅▅▅▆▅▅▆▄▇▇▆▆▆▆▆▅▆▅▆▇▆▅▇███ █
  762 μs        Histogram: log(frequency) by time       3.16 ms <

 Memory estimate: 4.37 MiB, allocs estimate: 7373.
```

Not sure and what's happening here but this seems to be a clue.

Also, this inconsistency with results is making it really hard to work on this :/

src/AbstractOperations/metric_field_reductions.jl
```diff
@@ -43,24 +44,15 @@ for information and examples using `condition` and `mask` kwargs.
...
-        metric = GridMetricOperation(location(field), dx, field.grid)
-        L = sum(metric; condition, mask, dims)
-
-        # Construct summand of the Average
-        L⁻¹_field_dx = field * dx / L
-
-        operand = condition_operand(L⁻¹_field_dx, condition, mask)
+    # Compute "size" (length, area, or volume) of averaging region
+    metric = GridMetricOperation(location(field), dx, field.grid)
+    ∫dx = @allowscalar sum(metric; condition, mask, dims)[] # Pluck out L as a Number
```
glwagner: Why doesn't `sum` already return a number in this case?

src/AbstractOperations/metric_field_reductions.jl
glwagner: You mean that the eltype is an issue?

src/AbstractOperations/metric_field_reductions.jl
tomchor: Not sure since I've never used `GridMetricOperation` before. That said, if you `sum()` the variable `metric` (a `GridMetricOperation`), it does return a number:

```julia
infil> metric
Vᶠᶜᶜ at (Face, Center, Center)
├── grid: 8×8×8 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 1×1×1 halo
└── tree: 
    Vᶠᶜᶜ at (Face, Center, Center)

infil> sum(metric)
1.0
```

But when you `sum(metric)` along with a `dims` argument that isn't `Colon()`, then you get a `Field`:

```julia
infil> sum(metric; dims=Colon())
1.0

infil> sum(metric; dims=(1, 2, 3))
1×1×1 Field{Nothing, Nothing, Nothing} reduced over dims = (1, 2, 3) on RectilinearGrid on CPU
├── grid: 8×8×8 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 1×1×1 halo
│   └── west: Nothing, east: Nothing, south: Nothing, north: Nothing, bottom: Nothing, top: Nothing, immersed: ZeroFlux
└── data: 1×1×1 OffsetArray(::Array{Float64, 3}, 1:1, 1:1, 1:1) with eltype Float64 with indices 1:1×1:1×1:1
```

tomchor: That said, I think more generally we do want a `Field` or an `GridMetricOperation` since, whenever we're averaging in less than 3 dimensions, the `metric` will potentially very with non-averaged dimensions. I made this PR into a draft for now to account for the fact that I'm exploring.

src/AbstractOperations/metric_field_reductions.jl
tomchor: I think it's likely, though not sure. And/or something related to the calculation of `GridMetricOperation`s... ?

src/AbstractOperations/metric_field_reductions.jl
glwagner: Can't we use `dims=:` in this case though? Because we are averaging a WindowedField that only has one grid point in the third direction. So we are indeed able to average over all directions.

What if you try to precompute `sum(1 / metric)`?

tomchor: Same results as main: I get a 50-70x slowdown when computing averages with `Average()`.

simone-silvestri:
I think I found one issue. `mean(interior(u))` is slower than it should be. Apparently, all the `map`s we have in `interior` are slowing down considerably all the computations, I ll open a different PR, then we can revisit this one. After fixing the `map`s, the additional cost of adding a conditional operation is concentrated in the `Field` constructor.

simone-silvestri:
See #4668

tomchor:
Closing this since https://github.com/CliMA/Oceananigans.jl/pull/4795 made this PR obsolete

PR #4651, opened 2025-07-17
Fix boundary conditions with general vertical coordinate

simone-silvestri:
in main, the added flux boundary conditions (computed after the tendency computation) are not multiplied by the grid scaling, leading to an incorrect behaviour when using zstar in connection with a flux boundary condition.

This PR corrects this and adds more tests.


src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
```diff
@@ -181,23 +181,46 @@ end
...
+        launch!(architecture(grid), grid, :xyz, _multiply_by_grid_scaling!, G, grid)
+    end
+
+    return nothing
+end
+
+@kernel function _multiply_by_grid_scaling!(G, grid)
+    i, j, k = @index(Global, NTuple)
+    @inbounds G[i, j, k] *= σⁿ(i, j, k, grid, Center(), Center(), Center())
+end
```
glwagner, line 222: The whole tendency has to be multiplied by this?

this complicated all extensions of the code that do substepping — is there no better way?

simone-silvestri: Yeah, we could also just multiply it in the time advancement, but it will come with the same issue. This comes from the fact that a division by the grid scaling pops up in the time derivative once you transform the equations (written up in #4588). Another option is to change the equations by multiplying $\sigma$ throughout but that would change the definition of the grid, and all the equations would have to be modified accordingly.

glwagner: > Another option is to change the equations by multiplying σ throughout but that would change the definition of the grid, and all the equations would have to be modified accordingly.

Right, this is the simplification I am wondering about. If we simply change the definition of the prognostic variables, why would this change the definition of the grid?

simone-silvestri: We could think about it. That requires a quite substantial refactor to include the scaling in the individual tendency term (for example, advective and diffusive fluxes would need to be multiplied by $\sigma$), and using $\Delta r$ instead of $\Delta z$ everywhere else, or just removing $\Delta r$ and $\Delta z$ would just be the $r$-coordinate instead of the actual depth. So I would say to merge this PR and then look at the feasibility of a refactor of the `HydrostaticFreeSurfaceModel`

glwagner: ok well if its not a simplification, then its not worth it

Issue #3943, opened 2024-11-20
Non-hydrostatic curvilinear shell?


YoffeG:
Hi all,

We wish to simulate a part of a spherical shell (e.g., within lat +-70) in a non-hydrostatic regime. Is there a readily available curvilinear grid framework to implement for such a setup?

Thanks in advance,

Gidi

glwagner:
Hi @YoffeG do you mind if I convert this to a discussion? I think our conversation will be better served by that format.

It sounds like you want to use a `LatitudeLongitudeGrid`. However, while we have a conjugate gradient solver for such a grid, we have never tested it. Note that all nonhydrostatic work I am aware of uses `RectilinearGrid` and a direct FFT solver. There is some more experimental work with the CG solver for nonhydrostatic simulations in complex domains.

It could be a fun research project to try to get the CG solver working on LatitudeLongitudeGrid if you are up for it.

YoffeG:
Hey @glwagner, many thanks for touching base and for your prompt reply!

The LatLong grid is definitely our tool of choice, but unfortunately -- it seems that the non-hydrostatic model is not our only problem (I believe that e.g., in MITgcm, it runs with CG/multigrid?), but mainly the lack of the full Coriolis force for a gcm. 
Are you guys planning to implement it at some point?

Thanks!

glwagner:
@YoffeG I am on here every day 😂 

MITgcm uses a CG solver with an analytical preconditioner as described by [Marshall et al 1997](https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/96JC02775). However, a CG solver with a multigrid preconditioner would likely be more performant.

I personally don't have plans to work on physics for nonhydrostatic deep atmosphere/oceans because I am working on developing a thin-shell hydrostatic ocean climate model at the moment. But Oceananigans is designed to facilitate rapid development by... anyone... so the correct question is "does _anybody_ have plans to implement this?" Obviously I don't know the answer! Perhaps, @Yixiao-Zhang or @jm-c have plans, or might know if any friends have plans to look into this. Here are some additional comments:

I'm not sure what a full Coriolis force entails. If it uses the existing grid, it is probably pretty easy. For example, this is the implementation of `FPlane` coriolis force:

https://github.com/CliMA/Oceananigans.jl/blob/2c955dac657ffa3760ffb46c5ff7a36c0215c59b/src/Coriolis/f_plane.jl#L44

[...]

PR #4654, opened 2025-07-17
`NonhydrostaticModel` on a `LatitudeLongitudeGrid`

simone-silvestri:
This PR allows using the `NonhydrostaticModel` on a `LatitudeLongitudeGrid`. We can perform some validation here before merging.

with @francispoulin 

An example of a small hot box positioned at the bottom of a nonhydrostatic latitude-longitude grid leads to

image

after about 10 minutes of evolution. The code is
```julia
grid = LatitudeLongitudeGrid(size = (1, 100, 100), latitude = (-0.01, 0.01), longitude = (-10, 10), z = (-100, 0), topology=(Periodic, Bounded, Bounded))
model = NonhydrostaticModel(; grid, tracers = :b, buoyancy = BuoyancyTracer(), advection=WENO())
bₛ(x, y, z) = 2e-4 * 10 * (10 + 5 * z / 100)
bb(x, y, z) = -90 < z < -70 ? (-0.005 < y < 0.005 ? 2e-4 * 10 * 5 : 0) : 0
bᵢ(x, y, z) = bₛ(x, y, z) + bb(x, y, z)
set!(model, b=bᵢ)
simulation = Simulation(model, Δt=10, stop_time=10minutes)
conjure_time_step_wizard!(simulation, IterationInterval(10); cfl=0.5)
run!(simulation)
```


francispoulin:
This answers the question that I raised in #3943. 

I approve this but hope someone else does also.

francispoulin (approved):

tomchor:
Nice! Ready to approve this after some tests are put in place. I'm assuming this runs significantly slower than an equivalent simulation on a RectilinearGrid, no?

simone-silvestri:
I haven't tried to compare the speed, but if I have to guess, I expect it to definitely run much slower.

tomchor:
> I haven't tried to compare the speed, but if I have to guess, I expect it to definitely run much slower.

Given that, do you think it's worth throwing a warning to the user?

glwagner:
> > I haven't tried to compare the speed, but if I have to guess, I expect it to definitely run much slower.
> 
> Given that, do you think it's worth throwing a warning to the user?

I don't think a warning makes sense. It's intrinsic that the model runs slow; there's little that can be done to solve this problem. If we go down that road we might as well throw a warning all the time "fluid dynamics calculations are slow!"

francispoulin:
Good question @tomchor.

We have a bunch of tests in `test_nonhydrostatic_model.jl`, all of which use the `RectilinearGrid`.  Do you want us to do some of these or all of these for `LatitudeLongitudeGrid`?

francispoulin:
In case anyone is interested, this is an animation of the rectangular bubble in a spherical geometry.
 
https://github.com/user-attachments/assets/3ab6ec78-60b5-46a1-b318-c4c3e1ce5f86

glwagner:
> Good question @tomchor.
> 
> We have a bunch of tests in `test_nonhydrostatic_model.jl`, all of which use the `RectilinearGrid`. Do you want us to do some of these or all of these for `LatitudeLongitudeGrid`?

Can we stick to just a few minimal tests that don't slow down CI too much?

francispoulin:
@glwagner and @tomchor : I add some tests (see the link below).  These should not slow things down.  

What do you think?  Do you want more or less?

https://github.com/CliMA/Oceananigans.jl/blob/ss-fjp/latlon-nonhydrostatic/test/test_nonhydrostatic_models.jl

simone-silvestri:
> @glwagner and @tomchor : I add some tests (see the link below). These should not slow things down.
> 
> What do you think? Do you want more or less?

I think this is ok, we can still regard it as an experimental feature. We could add just another test [here](https://github.com/CliMA/Oceananigans.jl/blob/f02624a5e19cfbb044f5860c2291f62775f3d7e0/test/test_nonhydrostatic_models.jl#L115) to test that the pressure solver works (i.e. incompressibility is set correctly)

francispoulin:
Thanks @simone-silvestri for the new tests.  Much better!

I tried running it and found there were two errors on `all(values_match)`.  Can someone test to see if they also get an error?

```
[2025/07/22 09:21:34.127] INFO    Testing model construction with single tracer and nothing tracer...
[2025/07/22 09:22:06.585] INFO    Testing setting model fields...
Setting model fields: Test Failed at /u/fpoulin/software/Oceananigans.jl/test/test_nonhydrostatic_models.jl:180
  Expression: all(values_match)

Stacktrace:
 [1] macro expansion
   @ /opt/uw/julia/1.10.8/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/software/Oceananigans.jl/test/test_nonhydrostatic_models.jl:180 [inlined]
 [3] macro expansion
   @ /opt/uw/julia/1.10.8/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] macro expansion
   @ ~/software/Oceananigans.jl/test/test_nonhydrostatic_models.jl:115 [inlined]
 [5] macro expansion
   @ /opt/uw/julia/1.10.8/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [6] top-level scope
   @ ~/software/Oceananigans.jl/test/test_nonhydrostatic_models.jl:6
Setting model fields: Test Failed at /u/fpoulin/software/Oceananigans.jl/test/test_nonhydrostatic_models.jl:180
  Expression: all(values_match)

Stacktrace:
 [1] macro expansion
   @ /opt/uw/julia/1.10.8/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/software/Oceananigans.jl/test/test_nonhydrostatic_models.jl:180 [inlined]
 [3] macro expansion
   @ /opt/uw/julia/1.10.8/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] macro expansion
   @ ~/software/Oceananigans.jl/test/test_nonhydrostatic_models.jl:115 [inlined]
 [5] macro expansion
   @ /opt/uw/julia/1.10.8/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [6] top-level scope
   @ ~/software/Oceananigans.jl/test/test_nonhydrostatic_models.jl:6
Test Summary:                                                                                                                                                                                                                                                                                                                                                          | Pass  Fail  Total     Time
Models                                                                                                                                                                                                                                                                                                                                                                 |   76     2     78  5m24.2s
```

francispoulin:
Testing the function setting for u passes on a rectilineaer grid but fails for the lat-lon grid.  Before is the output for the latter.  It seems like the u_cpu are including some ghost cells?

```[-88.25 -87.75 -87.25 -86.75; -87.75 -87.25 -86.75 -86.25; -87.25 -86.75 -86.25 -85.75; -86.75 -86.25 -85.75 -85.25; -86.25 -85.75 -85.25 -84.75;;; -63.25 -62.75 -62.25 -61.75; -62.75 -62.25 -61.75 -61.25; -62.25 -61.75 -61.25 -60.75; -61.75 -61.25 -60.75 -60.25; -61.25 -60.75 -60.25 -59.75;;; -38.25 -37.75 -37.25 -36.75; -37.75 -37.25 -36.75 -36.25; -37.25 -36.75 -36.25 -35.75; -36.75 -36.25 -35.75 -35.25; -36.25 -35.75 -35.25 -34.75;;; -13.25 -12.75 -12.25 -11.75; -12.75 -12.25 -11.75 -11.25; -12.25 -11.75 -11.25 -10.75; -11.75 -11.25 -10.75 -10.25; -11.25 -10.75 -10.25 -9.75]```

```[0.0 0.0 0.0 0.0; -87.75 -87.25 -86.75 -86.25; -87.25 -86.75 -86.25 -85.75; -86.75 -86.25 -85.75 -85.25; 0.0 0.0 0.0 0.0;;; 0.0 0.0 0.0 0.0; -62.75 -62.25 -61.75 -61.25; -62.25 -61.75 -61.25 -60.75; -61.75 -61.25 -60.75 -60.25; 0.0 0.0 0.0 0.0;;; 0.0 0.0 0.0 0.0; -37.75 -37.25 -36.75 -36.25; -37.25 -36.75 -36.25 -35.75; -36.75 -36.25 -35.75 -35.25; 0.0 0.0 0.0 0.0;;; 0.0 0.0 0.0 0.0; -12.75 -12.25 -11.75 -11.25; -12.25 -11.75 -11.25 -10.75; -11.75 -11.25 -10.75 -10.25; 0.0 0.0 0.0 0.0]```

**Comment by francispoulin** [comment]
I looked at this a little more and it seems to me that when we use `set!` to set the initial conditions, it does not set the first and last x-value.  Those arrays are both zero.  This works fine on a cartesian grid but not on the lat-lon grid.

File: test/test_nonhydrostatic_models.jl [hunk 1]
```diff
@@ -5,11 +5,17 @@ using Oceananigans.Grids: required_halo_size_x, required_halo_size_y, required_h
 @testset "Models" begin
     @info "Testing models..."
 
-    @testset "Model constructor errors" begin
-        grid = RectilinearGrid(CPU(), size=(1, 1, 1), extent=(1, 1, 1))
-        @test_throws TypeError NonhydrostaticModel(; grid, boundary_conditions=1)
-        @test_throws TypeError NonhydrostaticModel(; grid, forcing=2)
-        @test_throws TypeError NonhydrostaticModel(; grid, background_fields=3)
+    grids = (RectilinearGrid(CPU(), size=(1, 1, 1), extent=(1, 1, 1)),
+             LatitudeLongitudeGrid(CPU(), size=(1,1,1), longitude=(-180, 180), latitude=(-20, 20), z=(-1, 0)))
```
navidcy:
```suggestion
             LatitudeLongitudeGrid(CPU(), size=(1, 1, 1), longitude=(-180, 180), latitude=(-20, 20), z=(-1, 0)))
```

test/test_nonhydrostatic_models.jl
```diff
@@ -95,13 +101,13 @@ using Oceananigans.Grids: required_halo_size_x, required_halo_size_y, required_h
     @testset "Model construction with single tracer and nothing tracer" begin
         @info "  Testing model construction with single tracer and nothing tracer..."
         for arch in archs
-            grid = RectilinearGrid(arch, size=(1, 1, 1), extent=(1, 2, 3))
-
-            model = NonhydrostaticModel(; grid, tracers=:c, buoyancy=nothing)
-            @test model isa NonhydrostaticModel
+            for grid in grids
```
navidcy, line 104: `grids` seem to be constructed beforehand and live on the CPU(). Thus putting all these within a loop `for arch in archs ... end` seems to have no effect, right?

francispoulin: Thanks for the suggestion.

francispoulin:
Based on #4664, I know that the tests are failing because the rectilinear grids and lat-lon grides have different topologies.  In particular, the first is periodic in x and the second is bounded in x.  

Do we want to change the rectilinear grid to be Bounded in all three directions as well and then change the test functions?  @simone-silvestri

navidcy:
You can also change the LatitudeLongitudeGrid to periodic? (in zonal direction at least)

francispoulin:
> You can also change the LatitudeLongitudeGrid to periodic? (in zonal direction at least)

I can make it periodic in the zonal direction, which will then allow the first test to pass for u.  

But I can't make it periodic in the meridonal direction (with good reason) so the v velocity will fail.  

I guess that would make things somewhat better as we only need to find a new function for v.

(Did you mean to close this?)

navidcy:
I definitely didn't mean to close the PR!

francispoulin:
I made the following changes and the tests pass.

1. Both grids are Periodic, Bounded, Bounded
2. I removed the test for periodicity in the y-direction, since that is no longer true.
3. I removed the test for v at the meridional boundaries, as the function does not satisfy these boundary conditons.

Is this okay with everyone?

We can construct a function for v that satisfies the boundary conditions but we will then need to change the length of the domains.  This is something I could do tomorrow, if people like.

Thanks @navidcy for the help, it was very much appreciated.   You helped me to find the problem and at least find a patch for the test.

francispoulin:
Good changes @navidcy , thanks!

Do people want more tests or anything else from this PR?

navidcy:
Happy to merge this

navidcy (approved):

PR #4660, opened 2025-07-21
Add grid reconstruction utilities to reconstruct Grids from NetCDF outputs

tomchor:
Needed to complete https://github.com/CliMA/Oceananigans.jl/issues/3935

This PR allows us to dump grid data (and metadata) into a NetCDF and reconstruct the same grid from it later on:

```julia
ds = NCDataset(filename, "c")
write_grid_reconstruction_data!(ds, original_grid)
close(ds)

reconstructed_grid = reconstruct_grid_from_netcdf(filename)
original_grid == grid # true
```

It also adds tests for the grid reconstruction utilities that already existed, but which do not use NetCDF. (Namely, `constructor_arguments()`.)


test/test_grid_constructor_arguments.jl
```diff
@@ -0,0 +1,267 @@
...
+
+    # Test coordinate spacings (for regular grids these should be constant numbers)
+    @test reconstructed_grid.Δxᶠᵃᵃ == original_grid.Δxᶠᵃᵃ
+    @test reconstructed_grid.Δyᵃᶠᵃ == original_grid.Δyᵃᶠᵃ
+
+    # Test coordinates match
+    @test all(reconstructed_grid.xᶠᵃᵃ == original_grid.xᶠᵃᵃ)
+    @test all(reconstructed_grid.xᶜᵃᵃ == original_grid.xᶜᵃᵃ)
+    @test all(reconstructed_grid.yᵃᶠᵃ == original_grid.yᵃᶠᵃ)
+    @test all(reconstructed_grid.yᵃᶜᵃ == original_grid.yᵃᶜᵃ)
```
navidcy: why not test z also?

test/test_grid_constructor_arguments.jl
```diff
@@ -0,0 +1,267 @@
...
+    @test reconstructed_grid.Δλᶠᵃᵃ == original_grid.Δλᶠᵃᵃ
+    @test reconstructed_grid.Δφᵃᶠᵃ == original_grid.Δφᵃᶠᵃ
+    @test reconstructed_grid.Δxᶠᶠᵃ == original_grid.Δxᶠᶠᵃ
+    @test reconstructed_grid.Δyᶜᶠᵃ == original_grid.Δyᶜᶠᵃ
+
+    # Test coordinate arrays match
+    @test all(reconstructed_grid.λᶠᵃᵃ == original_grid.λᶠᵃᵃ)
+    @test all(reconstructed_grid.λᶜᵃᵃ == original_grid.λᶜᵃᵃ)
+    @test all(reconstructed_grid.φᵃᶠᵃ == original_grid.φᵃᶠᵃ)
+    @test all(reconstructed_grid.φᵃᶜᵃ == original_grid.φᵃᶜᵃ)
```
navidcy: test z?

test/test_grid_constructor_arguments.jl
```diff
@@ -0,0 +1,267 @@
...
+            if CUDA.functional()
+                push!(archs, GPU())
+            end
+        catch
+            # CUDA not functional, continue with CPU only
+        end
+    end
+
+    float_types = [Float32, Float64]
+
```
navidcy: No need for duplicate redefinitions.

"dependencies_for_runtests.jl" includes definitions for archs and float_types.

```suggestion
```

test/test_grid_constructor_arguments.jl
```diff
@@ -0,0 +1,267 @@
+using Oceananigans
+using Test
```
navidcy:
```suggestion
include("dependencies_for_runtests.jl")
```

test/test_grid_constructor_arguments.jl
```diff
@@ -0,0 +1,267 @@
...
+            test_grid_equality_after_reconstruction(arch, FT)
+        end
+
+        @testset "LatitudeLongitudeGrid reconstruction tests [$FT, $(typeof(arch))]" begin
+            @info "  Testing LatitudeLongitudeGrid reconstruction [$FT, $(typeof(arch))]..."
+            
+            test_latitude_longitude_grid_reconstruction(arch, FT)
+        end
+    end
+end
```
navidcy:
```suggestion
end

```

navidcy:
Oh sorry, I only noticed it was a draft after I made the suggestions! :(

tomchor:
> Oh sorry, I only noticed it was a draft after I made the suggestions! :(

No worries at all! Comments are always appreciated :)

I'll address them soon.

tomchor:
I realized that grid, currently, equality only works properly for non-immersed grids, as it doesn't compare anything related to immersed boundaries. I want this feature to be available for immersed grids as well, as the tests would be much simpler with proper equality testing for immersed grids, so I'm going to open that PR first before testing immersed grids here.

Otherwise this PR is more or less ready.

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -709,11 +711,17 @@ function grid_attributes(ibg::ImmersedBoundaryGrid)
...
 end
 
-netcdfify_dict_values(dict) = OrderedDict(key => netcdfify(value) for (key, value) in dict)
+# Using OrderedDict to preserve order of keys. Important for positional arguments.
+netcdfify_dict_values(dict) = OrderedDict{Symbol, Any}(key => netcdfify(value) for (key, value) in dict)
 netcdfify(x::Number) = x
 netcdfify(x::NTuple{N, Number}) where N = collect(x)
 netcdfify(x) = string(x)
 
+de_netcdfify_dict_values(dict) = OrderedDict(Symbol(key) => de_netcdfify(value) for (key, value) in dict)
```
glwagner:
```suggestion
materialize_from_netcdf(dict) = OrderedDict(Symbol(key) => de_netcdfify(value) for (key, value) in dict)
```

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -709,11 +711,17 @@ function grid_attributes(ibg::ImmersedBoundaryGrid)
     return attrs, dims
 end
 
-netcdfify_dict_values(dict) = OrderedDict(key => netcdfify(value) for (key, value) in dict)
+# Using OrderedDict to preserve order of keys. Important for positional arguments.
+netcdfify_dict_values(dict) = OrderedDict{Symbol, Any}(key => netcdfify(value) for (key, value) in dict)
```
glwagner:
```suggestion
convert_for_netcdf(dict) = OrderedDict{Symbol, Any}(key => netcdfify(value) for (key, value) in dict)
```

ext/OceananigansNCDatasetsExt.jl
tomchor: Much better! I'll make these changes soon

tomchor:
I can't confirm that the tests are passing because CI seems to be down, but it'd be good to get some reviews. For now the reconstruction is supported for Rectilinear and LatLon grids, but not for immersed grids. If possible I'd like to implement the immersed boundary part (which requires a few more tweaks) in another PR.

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -676,52 +659,80 @@ function grid_reconstruction_attributes(grid::LatitudeLongitudeGrid)
...
+materialize_from_netcdf(dict::AbstractDict) = OrderedDict(Symbol(key) => materialize_from_netcdf(value) for (key, value) in dict)
+materialize_from_netcdf(x::Number) = x
+materialize_from_netcdf(x::Array) = Tuple(x)
+materialize_from_netcdf(x::String) = @eval $(Meta.parse(x))
 
-    ds_grid = defGroup(ds, "grid_reconstruction"; attrib = sort(collect(pairs(grid_attrs)), by=first))
+function write_grid_reconstruction_data!(ds, grid; array_type=Array{eltype(grid)}, deflatelevel=0)
+    grid_attrs, grid_dims = grid_attributes(grid)
 
+    sorted_grid_attrs = sort(collect(pairs(grid_attrs)), by=first)
```
glwagner: add a comment about what this is doing?

src/Grids/rectilinear_grid.jl
```diff
@@ -387,7 +389,7 @@ cpu_face_constructor_y(grid::YRegularRG) = y_domain(grid)
 function constructor_arguments(grid::RectilinearGrid)
     arch = architecture(grid)
     FT = eltype(grid)
-    args = Dict(:architecture => arch, :number_type => eltype(grid))
+    args = OrderedDict(:architecture => arch, :number_type => eltype(grid))
```
glwagner, line 394: 👀 

This seems totally fine, but maybe a little explanation why we need it to be ordered is apropos?

glwagner (approved):
Can you put some code into this discussion that illustrates what this does? I presume this allows us to write something like

```julia
grid = construct_grid(filename)
```

when `filename` points to a NetCDF file generated by `NetCDFWriter`.

tomchor:
> Can you put some code into this discussion that illustrates what this does? I presume this allows us to write something like
> 
> ```julia
> grid = construct_grid(filename)
> ```
> 
> when `filename` points to a NetCDF file generated by `NetCDFWriter`.

Yes. I added a quick description at the top. Let me know if that's enough. The user current interface is a bit more verbose, but I'm wondering if I should change it to what you used in your example, which is much simpler and still understandable. Let me know your thoughts.

src/Grids/rectilinear_grid.jl
tomchor: Added this comment:

`# We use OrderedDict to preserve order of keys. Important for positional arguments since we wanna be able to splat them.`

ext/OceananigansNCDatasetsExt.jl
tomchor: Added this comment:

`# Organizes attributes to make it more easily human-readable`

So the ordering isn't really necessary and it was there before my changes, but I do think it makes it nicer to read as a human.

glwagner: That makes sense to me. By sorting them, the attributes appear in more predictable positions.

glwagner:
> > Can you put some code into this discussion that illustrates what this does? I presume this allows us to write something like
> > ```julia
> > grid = construct_grid(filename)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > when `filename` points to a NetCDF file generated by `NetCDFWriter`.
> 
> Yes. I added a quick description at the top. Let me know if that's enough. The user current interface is a bit more verbose, but I'm wondering if I should change it to what you used in your example, which is much simpler and still understandable. Let me know your thoughts.

I think typically fewer and shorter names is better, when we can leverage multiple dispatch / analysis of the arguments to control behavior, the benefit being an interface which is easier to learn and remember. "grid construction" is useful for any file format (not just netcdf) so I'm not sure it's well motivated to append `from_netcdf`. 

Also we might want something like `reconstruct_grid(::Dataset)` in addition to `reconstruct_grid(::String) since the former is what we need for `FieldTimeSeries`.

As for `write_grid_reconstruction_data!` --- is it useful to distinguish between "reconstruction data" and the grid itself? Perhaps just `write_grid!` is enough?

glwagner:
Another thing to consider is that there are a number of additional features for output analysis that we would like to add; notably the ability to inspect the content of a file without having to build `FieldTimeSeries` (ie to see what variables are saved inside). It'd be nice to have a common interface for all file formats for that too.

tomchor:
> > > Can you put some code into this discussion that illustrates what this does? I presume this allows us to write something like
> > > ```julia
> > > grid = construct_grid(filename)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > when `filename` points to a NetCDF file generated by `NetCDFWriter`.
> > 
> > 
> > Yes. I added a quick description at the top. Let me know if that's enough. The user current interface is a bit more verbose, but I'm wondering if I should change it to what you used in your example, which is much simpler and still understandable. Let me know your thoughts.
> 
> I think typically fewer and shorter names is better, when we can leverage multiple dispatch / analysis of the arguments to control behavior, the benefit being an interface which is easier to learn and remember. "grid construction" is useful for any file format (not just netcdf) so I'm not sure it's well motivated to append `from_netcdf`.

Makes sense. Just changed it to `reconstruct_grid`. Although I think `construct_grid` might also be good...

> 
> Also we might want something like `reconstruct_grid(::Dataset)` in addition to `reconstruct_grid(::String) since the former is what we need for `FieldTimeSeries`.

That is already what the main method is. The `reconstruct_grid(::String)` method just opens the NCDataset and calls the main method.

> As for `write_grid_reconstruction_data!` --- is it useful to distinguish between "reconstruction data" and the grid itself? Perhaps just `write_grid!` is enough?

I thought it was useful to be more explicit because otherwise it gives the impression that we're writing the grid itself into a file, which (if I understand correctly) can be done with JLD2 but cannot be done with NetCDF. So I thought it was good to make that distinction. But I'm happy to change it if you think it's best. (Also not sure how correct my understanding on JLD2 is...)

tomchor:
@glwagner @navidcy  the only tests that are failing are the reactant tests, which don't have anything to do with this PR. Can one of you merge this?

PR #4662, opened 2025-07-22
Fix for distributed unit tests

simone-silvestri:
It might be a synchronization issue. Let's try like this


navidcy (approved):

navidcy:
Test seem to pass.

Does this particular sync impede performance? I doubt, right?

simone-silvestri:
No it should not. Well, in any case, the device should be synchronized anyway for MPI to work... I'll merge this so we can procede with the other PRs

PR #4666, opened 2025-07-23
Fix distributed HydrostaticFreeSurfaceModel test that seem to break with memory error

navidcy:


navidcy:
Hm... this tho passed: https://buildkite.com/clima/oceananigans-distributed/builds/9100#0198392b-7d42-48c3-994f-3695c482b433

So perhaps the errors are random/stochastic...

simone-silvestri:
Yeah, I think it is kind of random, it might be connected to the timeout of the yaml. Let me try changing it.

navidcy:
It's happening more and more; eg https://github.com/CliMA/Oceananigans.jl/pull/4674, https://github.com/CliMA/Oceananigans.jl/pull/4672. Doesn't seem so "random"?

navidcy:
I increased the memory request a bit. They passed the first go. Shall we merge? Does it worth more investigating?

simone-silvestri:
Let's merge this then see if the error continues appearing

PR #4671, opened 2025-07-26
Drop CUDA_Runtime_jll

navidcy:
Do we need it?

I see it was introduced by https://github.com/CliMA/Oceananigans.jl/pull/3880? cc @simone-silvestri


navidcy:
Is this the line where we need it?

https://github.com/CliMA/Oceananigans.jl/blob/df160f9a316e10a2859178df7e707d5eeffd3c84/test/utils_for_runtests.jl#L28

PR #4672, opened 2025-07-26
(0.100.0) NamedCoordinate -> NamedDiscretization

navidcy:
Rename `Coordinates` to `Discretization` as per discussions in https://github.com/CliMA/ClimaOcean.jl/pull/573#discussion_r2232526219

### DIFF
--- a/src/Grids/coordinate_utils.jl
+++ b/src/Grids/coordinate_utils.jl
@@ -1,7 +1,7 @@
 abstract type CallableDiscretization end
 
 # Callable coordinates can be indexed just like arrays!
-Base.getindex(dsrc::CallableCoordinate, i) = dsrc(i)
+Base.getindex(dsrc::CallableDiscretization, i) = dsrc(i)
 
 struct ExponentialDiscretization <: CallableDiscretization
     size :: Int
@@ -23,7 +23,8 @@ Base.getindex(dsrc::ExponentialDiscretization, i) = @inbounds dsrc.faces[i]
 """
     ExponentialDiscretization(N::Int, left, right;
                               scale = (right - left) / 5,
-                              bias = :right)
+                              bias = :right,
+                              mutable = false)
 
 Return a one-dimensional discretization with `N` cells that are exponentially spaced
 (or, equivalently, with spacings that grow linearly along the coordinate).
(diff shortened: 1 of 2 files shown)


docs/src/grids.md
```diff
@@ -789,10 +800,10 @@ scatter!(axz2, 0 * zc, zc)
 hidedecorations!(axz2)
 hidespines!(axz2)
 
-z = ConstantToStretchedCoordinate(; extent, bias, bias_edge,
-                                  constant_spacing, constant_spacing_extent,
-                                  stretching = PowerLawStretching(1.03),
-                                  maximum_stretching_extent = 500)
+z = ConstantToStretchedDiscretization(; extent, bias, bias_edge,
```
glwagner: what does "constant to stretched" mean? Is this better than just calling it "stretched"? I'm not sure if its trying to say that you can specify a constant discretization in addition to a stretched one, or if the stretched discretization is associated with a constant reference by a function mapping? In the latter case, the concept of "stretching" actually already implies this on its own (which is perhaps the source of confusion about the meaning)

glwagner: Also rather than "constant" I suggest using "reference" since it helps clarify perhaps the role of the abstraction (a starting point for a computation)

navidcy:
Perhaps we include this breaking change in the v0.98?

navidcy:
Might require to resolve some conflicts after #4727 is merged.

glwagner (approved):

tomchor:
@navidcy could you please write why this is desirable in the top comment?

simone-silvestri:
I can't find the `mutable` kwarg as per https://github.com/CliMA/ClimaOcean.jl/pull/573#discussion_r2232526219, though.
I think it is required since it would be quite awkward to have to write.
```julia
z = ExponentialDiscretization(...)
z = MutableVerticalDiscretization(...)
```

navidcy:
OK. Let's not wait necessarily for this PR for the 0.98

navidcy:
I'll add the kwarg soon

glwagner:
what are the remaining challenges of this PR?

navidcy:
to add the kwarg about the mutable; I'm on it

PR #4673, opened 2025-07-26
Filter warnings from doctest output

navidcy:
This PR figures out to filter warnings from doctest outputs.

Also enhances some docstrings here and there.


src/Grids/latitude_longitude_grid.jl
```diff
@@ -132,7 +132,7 @@ Keyword arguments
   Each is either a:
   1. 2-tuple that specify the end points of the domain,
   2. one-dimensional array specifying the cell interface locations, or
-  3. a single-argument function that takes an index and returns cell interface location.
+  3. single-argument function that take an index and returns cell interface location.
```
tomchor:
```suggestion
  3. single-argument function that takes an index and returns cell interface location.
```

src/OrthogonalSphericalShellGrids/distributed_tripolar_grid.jl
```diff
@@ -42,16 +51,16 @@ function TripolarGrid(arch::Distributed, FT::DataType=Float64;
...
+        throw(ArgumentError("The x partition $(px) is not supported. The partition in x must be an even number."))
     end
 
     # a slab decomposition in x is not supported
     if px != 1 && py == 1
-        throw(ArgumentError("A x-only partitioning is not supported with the TripolarGrid. \n
-                            Please, use a y partitioning configuration or a x-y pencil partitioning."))
+        throw(ArgumentError("An x-only partitioning is not supported for TripolarGrid. \n
+                              Please, use a y partitioning configuration or an x-y pencil partitioning."))
     end
```
tomchor:
```suggestion
        throw(ArgumentError("An x-only partitioning is not supported for TripolarGrid. \n
                             Please, use a y partitioning configuration or an x-y pencil partitioning."))
    end
```

PR #4674, opened 2025-07-26
Only enforce open boundary mass conservation for open boundaries

navidcy:
Following from #4627, I noticed that `enforce_open_boundary_mass_conservation!` was _always_ called and doing stuff even when there was no open boundaries.

This PR deals with that.

with @glwagner

### DIFF
--- a/src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
+++ b/src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
@@ -153,7 +153,11 @@ function initialize_boundary_mass_fluxes(velocities::NamedTuple)
                                                 right_matching_scheme_boundaries,
                                                 total_area_matching_scheme_boundaries))
 
-    return boundary_fluxes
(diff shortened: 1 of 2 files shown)


glwagner:
An additional point we discussed was to use tuples instead of arrays to hold the list of open boundaries.

navidcy:
You can't `push!` into tuples tho. This method:

https://github.com/CliMA/Oceananigans.jl/blob/fb96739d4b1725edc91ec44d0eaec616012f4ba5/src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl#L92

 might require a bit of rethinking/refactoring.

glwagner:
> You can't `push!` into tuples tho. This method:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/fb96739d4b1725edc91ec44d0eaec616012f4ba5/src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl#L92
> 
> might require a bit of rethinking/refactoring.

We just need to convert the arrays to tuples

PR #4680, opened 2025-07-28
[Reactant] Add recursive specialization for BinaryOperation

wsmoses:


wsmoses:
@navidcy given the failures are unrelated, is this good to merge?

wsmoses:
ah they're not, one sec

tomchor:
Could you please a summary in the top comment of what issue this is trying to solve?

src/AbstractOperations/binary_operations.jl
```diff
@@ -51,6 +51,16 @@ choose_location(::Type{Center}, ::Type{Center}, Lc) = Center        #
 choose_location(La::ConcreteLocationType, ::Type{Nothing}, Lc) = La # don't interpolate unspecified locations.
 choose_location(::Type{Nothing}, Lb::ConcreteLocationType, Lc) = Lb #
 
+# Apply the function if the inputs are scalars, otherwise broadcast it over the inputs
+# This can occur in the binary operator code if we index into with an array, e.g. array[1:10]
+@inline @propagate_inbounds function apply_bcast(f, lhs, rhs)
```
glwagner:
```suggestion
@propagate_inbounds function apply_bcast(f, lhs, rhs)
```

or so I have been told is sufficient

src/AbstractOperations/binary_operations.jl
```diff
@@ -51,6 +51,16 @@ choose_location(::Type{Center}, ::Type{Center}, Lc) = Center        #
...
+    f(lhs, rhs)
+end
+
+@inline @propagate_inbounds function apply_bcast(f, lhs::AbstractArray, rhs::AbstractArray)
```
glwagner:
```suggestion
@propagate_inbounds function apply_bcast(f, lhs::AbstractArray, rhs::AbstractArray)
```

Project.toml
```diff
@@ -92,7 +92,7 @@ OffsetArrays = "1.4"
 OrderedCollections = "1.1"
 Printf = "1.9"
 Random = "1.9"
-Reactant = "0.2.141 - 0.2.146"
+Reactant = "0.2.141 - 0.2.149"
```
navidcy: is there a reason to cap it at 0.2.149?

Should we do:
```suggestion
Reactant = "0.2.141"
```
?

glwagner:
I am wondering why we end up trying to call the kernel functions (i, j, k, grid, etc) to get array outputs. The kernel functions are designed to output scalars --- but something about the convention/pattern is failing and creating the need for a fair amount of extra code such as this PR. @wsmoses is there a way around this? Its ok if not, but understanding why this need arises might help us participate more in the design if there is a better way to do it.

wsmoses:
so there are essentially various cases where these are called outside of a kernel context. In particular, this week @jlk9 was hitting some issues that stem from using these inside of a generated mapreduce. In these cases indexes into the arrays with slices (e.g. array[1:10]) to avoid scalar indexing (and also blowing up compile and runtime/etc).

simone-silvestri (approved):
I think this might need to be applied also the other `AbstractOperations`

PR #4683, opened 2025-07-29
Fjp/nonhydrostatic_coriolis

francispoulin:
This is my first attempt to include the complete Coriolis force (both traditional and non-traditional components) on a lat-lon grid.  I admit that I'm guessing in some places and would really appreciate another set of ideas.  

@simone-silvestri 

When I think the equations are correct I will put together a text.

I am not sure how to make this conserve energy or enstrophy, so I'm avoiding that for now, but happy to return to that when I know how.

### DIFF
--- a/src/Coriolis/nonhydrostatic_spherical_coriolis.jl
+++ b/src/Coriolis/nonhydrostatic_spherical_coriolis.jl
@@ -10,16 +10,18 @@ using Oceananigans.ImmersedBoundaries
     struct NonhydrostaticSphericalCoriolis{S, FT} <: AbstractRotation
 
 A parameter object for constant rotation around a vertical axis on the sphere.
+
+This is only designed to work for a LatitudeLongitudeGrid.
+
+There is no attempt to conserve energy or enstrophy at this early stage.
 """
 struct NonhydrostaticSphericalCoriolis{S, FT} <: AbstractRotation
     rotation_rate :: FT
-    scheme :: S
 end
 
 """
     NonhydrostaticSphericalCoriolis([FT=Float64;]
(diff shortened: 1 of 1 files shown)


glwagner:
Would it be better to just have "SphericalCoriolis" with the option to make the hydrostatic approximation (rather than copy/pasting a lot)?

francispoulin:
> Would it be better to just have "SphericalCoriolis" with the option to make the hydrostatic approximation (rather than copy/pasting a lot)?

I think that's a better idea, thanks for the suggestion.

To do that i can take `hydrostatic_sphericdal_coriolis.jl` and make a `spherical_coriolis.jl` function.  Is that what you had in mind @glwagner ?

glwagner:
> > Would it be better to just have "SphericalCoriolis" with the option to make the hydrostatic approximation (rather than copy/pasting a lot)?
> 
> I think that's a better idea, thanks for the suggestion.
> 
> To do that i can take `hydrostatic_sphericdal_coriolis.jl` and make a `spherical_coriolis.jl` function. Is that what you had in mind @glwagner ?

I'm suggesting to first change `HydrostaticSphericalCoriolis` to `SphericalCoriolis`. Then add a property which indicates whether we make the hydrostatic approximation. Then define the type alias for `HydrostaticSphericalCoriolis` (this makes the change not breaking)

francispoulin:
Thanks @glwagner .

I'm closing this PR as it's off to a false start.

I created a new branch `fjp/SphericalCoriolis`, that starts to do what I undertstand you are suggesting. I'll come back to this to get the lines of code that compute the Coriolis terms in general.

PR #4686, opened 2025-07-30
Move storage space to from `MutableVerticalDiscretization` to `ZStarCoordinate`

simone-silvestri:
and while we are at it, also refactor `ZStar` to `ZStarCoordinate`


src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -112,8 +112,8 @@ Keyword arguments
   - `pressure`: Hydrostatic pressure field. Default: `nothing`.
   - `diffusivity_fields`: Diffusivity fields. Default: `nothing`.
   - `auxiliary_fields`: `NamedTuple` of auxiliary fields. Default: `nothing`.
-  - `vertical_coordinate`: Algorithm for grid evolution: `ZStar()` or `ZCoordinate()`.
-                           Default: `default_vertical_coordinate(grid)`, which returns `ZStar()`
+  - `vertical_coordinate`: Algorithm for grid evolution: `ZStarCoordinate()` or `ZCoordinate()`.
+                           Default: `default_vertical_coordinate(grid)`, which returns `ZStarCoordinate()`
```
simone-silvestri, line 116:
```suggestion
                           Default: `default_vertical_coordinate(grid)`, which returns `ZStarCoordinate(grid)`
```

glwagner (approved):
Is `ηⁿ` needed to compute metrics too? Basically I think the only information that should be in the grid is what is required to compute the metrics. Anything to do with specific evolution rules is a model concept?

simone-silvestri:
Yeah, `ηⁿ` is the surface of the domain, so it is needed for `znode` to compute the correct `z` level. It might be considered somewhat a zstar specific quantity, but I guess any vertical coordinate would require the information of the extent of the domain, including the surface height.

glwagner:
> Yeah, `ηⁿ` is the surface of the domain, so it is needed for `znode` to compute the correct `z` level. It might be considered somewhat a zstar specific quantity, but I guess any vertical coordinate would require the information of the extent of the domain, including the surface height.

Okay, so the main improvement would be better notation if we have it

glwagner:
What about the time derivative of the stretching metric? This seems unnecessary for prescribed evolution?

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -112,8 +112,8 @@ Keyword arguments
   - `pressure`: Hydrostatic pressure field. Default: `nothing`.
   - `diffusivity_fields`: Diffusivity fields. Default: `nothing`.
   - `auxiliary_fields`: `NamedTuple` of auxiliary fields. Default: `nothing`.
-  - `vertical_coordinate`: Algorithm for grid evolution: `ZStar()` or `ZCoordinate()`.
-                           Default: `default_vertical_coordinate(grid)`, which returns `ZStar()`
+  - `vertical_coordinate`: Algorithm for grid evolution: `ZStarCoordinate()` or `ZCoordinate()`.
```
navidcy:
```suggestion
  - `vertical_coordinate`: Algorithm for grid evolution: `ZStarCoordinate()` or `ZCoordinate(grid)`.
```
?

validation/z_star_coordinate/lock_release.jl
```diff
@@ -23,7 +23,7 @@ model = HydrostaticFreeSurfaceModel(; grid,
                                     closure = nothing,
                                     tracers = :b,
                                 timestepper = :SplitRungeKutta3,
-                        vertical_coordinate = Oceananigans.Models.ZStar(),
+                        vertical_coordinate = Oceananigans.Models.ZStarCoordinate(),
```
navidcy: or
```suggestion
                        vertical_coordinate = Oceananigans.Models.ZStarCoordinate(grid),
```
?

validation/implicit_free_surface/geostrophic_adjustment_test.jl
```diff
@@ -17,7 +17,7 @@ function geostrophic_adjustment_simulation(free_surface, grid, timestepper=:Quas
                                           coriolis=FPlane(f = 1e-4),
                                           timestepper,
                                           free_surface,
-                                          vertical_coordinate=ZStar())
+                                          vertical_coordinate=ZStarCoordinate())
```
navidcy:
```suggestion
                                          vertical_coordinate=ZStarCoordinate(grid))
```
?

simone-silvestri:
> What about the time derivative of the stretching metric? This seems unnecessary for prescribed evolution?

I think that is required for any type of mutable grid, but I guess that can also be moved...

Might be annoying to rewrite this operator though
https://github.com/CliMA/Oceananigans.jl/blob/fb96739d4b1725edc91ec44d0eaec616012f4ba5/src/Advection/vector_invariant_self_upwinding.jl#L8-L12

navidcy:
Suggest bump patch release with this (or straight after) since it breaks this:

https://github.com/CliMA/ClimaOcean.jl/blob/8479089e933f1c8b29b9aaaac11e51437fbd9c64/src/OceanSimulations/ocean_simulation.jl#L80

PR #4687, opened 2025-07-31
(0.98.0) Modify interface for `OpenBoundaryCondition`s with `schemes` and use Tuples in `boundary_mass_fluxes`

tomchor:
This apparently came up in a conversation between @glwagner and @navidcy  and was mentioned in https://github.com/CliMA/Oceananigans.jl/pull/4674, so I'm opening this. This PR:

- Modifies the interface for creating OpenBoundaryConditions with "matching schemes", which now are just called schemes
- Removes the "experimental" warning for `PerturbationAdvectionOpenBoundaryCondition` now that it has more rigorous testing and seems to be working properly
- Removes FlatExtrapolation based on the discussion in https://github.com/CliMA/Oceananigans.jl/issues/4703#issuecomment-3177580181
- Uses tuples (instead of arrays) to hold the open boundaries in `model.boundary_fluxes`.


glwagner:
> I also took the liberty of exporting PerturbationAdvectionOpenBoundaryCondition at the top level. Let me know if that's not wanted.

Can we change the name to just `OpenBoundaryCondition`? I think we should use `OpenBoundaryCondition` and implement a generic constructor that can be used for the different cases flexibly. `PerturbationAdvectionOpenBoundaryCondition` is too long and also isn't really self-describing, so we don't benefit from the verbosity.
I would make this API change instead of exporting a new name.

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -149,8 +149,8 @@ function initialize_boundary_mass_fluxes(velocities::NamedTuple)
         total_area_matching_scheme_boundaries += boundary_fluxes.top_area
     end
 
-    boundary_fluxes = merge(boundary_fluxes, (; left_matching_scheme_boundaries,
-                                                right_matching_scheme_boundaries,
+    boundary_fluxes = merge(boundary_fluxes, (; left_matching_scheme_boundaries = Tuple(left_matching_scheme_boundaries),
+                                                right_matching_scheme_boundaries = Tuple(right_matching_scheme_boundaries),
                                                 total_area_matching_scheme_boundaries))
```
glwagner: What are the benefits of a NamedTuple vs type here

tomchor:
> > I also took the liberty of exporting PerturbationAdvectionOpenBoundaryCondition at the top level. Let me know if that's not wanted.
> 
> Can we change the name to just `OpenBoundaryCondition`? I think we should use `OpenBoundaryCondition` and implement a generic constructor that can be used for the different cases flexibly. `PerturbationAdvectionOpenBoundaryCondition` is too long and also isn't really self-describing, so we don't benefit from the verbosity. I would make this API change instead of exporting a new name.

Sure, something like `OpenBoundaryCondition(U; matching_scheme=PerturbationAdvection())`?

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
tomchor: Not sure I understand the question. What do you mean by type?

glwagner:
> > > I also took the liberty of exporting PerturbationAdvectionOpenBoundaryCondition at the top level. Let me know if that's not wanted.
> > 
> > 
> > Can we change the name to just `OpenBoundaryCondition`? I think we should use `OpenBoundaryCondition` and implement a generic constructor that can be used for the different cases flexibly. `PerturbationAdvectionOpenBoundaryCondition` is too long and also isn't really self-describing, so we don't benefit from the verbosity. I would make this API change instead of exporting a new name.
> 
> Sure, something like `OpenBoundaryCondition(U; matching_scheme=PerturbationAdvection())`?

right, with `OpenBoundaryCondition()` defaulting to impenetrable perhaps?

Also what is a "perturbation advection scheme"? One thing to consider for the name is what the different alternatives might be so that we can choose a name that meaningfully distinguishes between them. From what I understand, we either have a scheme that strictly prescribes the boundary velocity, or a scheme that can evolve it, is that right?

tomchor:
> > > > I also took the liberty of exporting PerturbationAdvectionOpenBoundaryCondition at the top level. Let me know if that's not wanted.
> > > 
> > > 
> > > Can we change the name to just `OpenBoundaryCondition`? I think we should use `OpenBoundaryCondition` and implement a generic constructor that can be used for the different cases flexibly. `PerturbationAdvectionOpenBoundaryCondition` is too long and also isn't really self-describing, so we don't benefit from the verbosity. I would make this API change instead of exporting a new name.
> > 
> > 
> > Sure, something like `OpenBoundaryCondition(U; matching_scheme=PerturbationAdvection())`?
> 
> right, with `OpenBoundaryCondition()` defaulting to impenetrable perhaps?

:+1: 

> Also what is a "perturbation advection scheme"? One thing to consider for the name is what the different alternatives might be so that we can choose a name that meaningfully distinguishes between them. From what I understand, we either have a scheme that strictly prescribes the boundary velocity, or a scheme that can evolve it, is that right?

Yes, and the schemes that predict what the boundary velocity can be from the flow (rather then prescribe/impose it) typically do so in order to minimize reflection. There is a lot of different nomenclature here, so I don't know what the best name is. When originally implementing this @jagoosw chose to call these schemes "matching schemes", but I haven't found that name elsewhere in the literature (@jagoosw feel free to chime in here; perhaps `radiation_scheme` is more intuitive to understand?). I think they just call these different types of BCs radiation or nonreflective boundary conditions. For the record, here's what Durran 2010 says about it:

image

(So a note is that, us calling boundary conditions that prescribe the normal velocity without trying to minimize reflection as an `OpenBoundaryCondition`s goes against common nomenclature.)

And back to your point, `PerturbationAdvection` is what we ended up calling the "matching scheme" that @jagoosw implemented. I remember we had some back and forth about that name back then, but I couldn't find that scheme in the literature so I'm not sure what it is usually called.

jagoosw:
As I remember, we settled on "matching scheme" since the scheme abstractly matches the interior physics with some prescribed boundary physics. I think that radiation schemes are a subset of this because an open boundary might e.g. be prescribing an inflow, and I gather that radiation schemes gained their name from their derivation from approximating the wave propagation from the Helmholtz equation, which isn't universally how you might want to treat an open boundary. So I think calling them all radiation schemes is misleading. For example, the Stevens boundary condition (https://mitgcm.readthedocs.io/en/latest/phys_pkgs/obcs.html) has nothing to do with that derivation.

Similarly, for the "perturbation advection" we called it this because that describes what it does (i.e. it advects perturbations out of the domain with some mean velocity), this is similar to an Orlanski/radiation condition where the "mean" velocity is guessed by some particular procedure, or could be the boundary mean velocity like [Simone describes here](https://github.com/CliMA/Oceananigans.jl/pull/3482#issuecomment-1965168780). If we can think of something else that describes what this does, then happy for it to be changed, but I think this is quite a good descriptor of what it can be used for. There is also a derivation in the docstring which kind of shows the advecting a pertubation idea.

jagoosw:
> Sure, something like `OpenBoundaryCondition(U; matching_scheme=PerturbationAdvection())`?

Given this API it could maybe just be called `scheme`?

glwagner:
> > Sure, something like `OpenBoundaryCondition(U; matching_scheme=PerturbationAdvection())`?
> 
> Given this API it could maybe just be called `scheme`?

I like that. If there is only one kind of "scheme", it's better to avoid the adjective. Also just to rehash some very old conversations, originally we put off writing a unified API for `OpenBoundaryCondition` because we saw the implementation as experimental. So the unified API should be implemented as part of removing the "experimental" designation.

navidcy:
> If there is only one kind of "scheme", it's better to avoid the adjective. 

I agree; and explain that kwarg `scheme` is about matching in the docstring.

tomchor:
> > > Sure, something like `OpenBoundaryCondition(U; matching_scheme=PerturbationAdvection())`?
> > 
> > 
> > Given this API it could maybe just be called `scheme`?
> 
> I like that. If there is only one kind of "scheme", it's better to avoid the adjective. Also just to rehash some very old conversations, originally we put off writing a unified API for `OpenBoundaryCondition` because we saw the implementation as experimental. So the unified API should be implemented as part of removing the "experimental" designation.

Not reviving anything old at all! It's one of the points I'm making in the top comment :)

If I understand correctly, I'll then remove the warning for this scheme.

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
glwagner: a type other than NamedTuple --- a custom struct, which is an alternative to using a NamedTuple

src/Oceananigans.jl
```diff
@@ -35,6 +35,7 @@ export
     # Boundary conditions
     BoundaryCondition,
     FluxBoundaryCondition, ValueBoundaryCondition, GradientBoundaryCondition, OpenBoundaryCondition,
+    PerturbationAdvectionOpenBoundaryCondition,
```
glwagner: please remove this

tomchor: I added this before our discussion on user interface. That won't exist once I refactor everything.

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
tomchor: I guess just the simplicity of using a common type that everyone's familiar with. I don't see disadvantages of using `NamedTuple`s.

src/BoundaryConditions/perturbation_advection_open_boundary_scheme.jl
```diff
@@ -4,6 +4,10 @@ using Oceananigans: defaults
 """
     PerturbationAdvection
 
+Create a `PerturbationAdvection` scheme to be used with an `OpenBoundaryCondition`.
+This scheme will nudge the boundary velocity to the OpenBoundaryCondition's exterior value `val`,
+using a time-scale `inflow_timescale` for inflow and `outflow_timescale` for outflow.
+
```
navidcy: Should we move the whole docstring on the constructor rather than the type?

test/test_boundary_conditions_integration.jl
```diff
@@ -195,7 +193,7 @@ function test_perturbation_advection_open_boundary_conditions(arch, FT)
         @test all(view(parent(u), :, :, :) .== -1)
         @test all(interior(u) .== -1)
 
-        obc = PerturbationAdvectionOpenBoundaryCondition((t) -> 0.1*t, inflow_timescale = 0.01, outflow_timescale = 0.5)
+        obc = OpenBoundaryCondition((t) -> 0.1*t, scheme = PerturbationAdvection(inflow_timescale = 0.01, outflow_timescale = 0.5))
```
navidcy: I don't think the parenthesis is needed.

```suggestion
        obc = OpenBoundaryCondition(t -> 0.1*t, scheme = PerturbationAdvection(inflow_timescale = 0.01, outflow_timescale = 0.5))
```

src/BoundaryConditions/perturbation_advection_open_boundary_scheme.jl
tomchor: Yes! Done

test/test_boundary_conditions_integration.jl
tomchor: Done

tomchor:
I think this is ready. I bumped the minor version since we're changing the user interface for for open BCs with schemes.

validation/open_boundaries/cylinder.jl
```diff
@@ -214,14 +213,11 @@ end
...
 
-feobcs = (west = OpenBoundaryCondition(u∞), east = FlatExtrapolationOpenBoundaryCondition(),)
-
 paobcs = (west = PerturbationAdvectionOpenBoundaryCondition(u∞; inflow_timescale = 0.1, outflow_timescale = 0.1),
           east = PerturbationAdvectionOpenBoundaryCondition(u∞; inflow_timescale = 1/4, outflow_timescale = Inf),
           )
 
-obcs = (; flat_extrapolation=feobcs,
-          perturbation_advection=paobcs)
+obcs = (; perturbation_advection=paobcs,)
```
glwagner:
```suggestion
obcs = (; perturbation_advection=paobcs)
```

validation/open_boundaries/oscillating_flow.jl
```diff
@@ -143,7 +136,7 @@ for grid in (xygrid, xzgrid)
                                               top    = PerturbationAdvectionOpenBoundaryCondition(v∞; parameters = (; U, T), inflow_timescale, outflow_timescale))
     paobcs = (u = u_boundaries_pa, v = v_boundaries_pa, w = w_boundaries_pa)
 
-    for obcs in (feobcs, paobcs,)
+    for obcs in (paobcs,)
```
glwagner:
```suggestion
    for obcs in tuple(paobcs)
```

tomchor:
@glwagner I implemented your suggestions. This should be good to go. The reactant tests are failing for some reason, but I think that's not related to anything here in this PR.

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -97,63 +97,63 @@ function initialize_boundary_mass_fluxes(velocities::NamedTuple)
...
 
-    boundary_fluxes = merge(boundary_fluxes, (; left_matching_scheme_boundaries,
-                                                right_matching_scheme_boundaries,
-                                                total_area_matching_scheme_boundaries))
+    boundary_fluxes = merge(boundary_fluxes, (; left_scheme_boundaries = Tuple(left_scheme_boundaries),
+                                                right_scheme_boundaries = Tuple(right_scheme_boundaries),
+                                                total_area_scheme_boundaries))
 
-    if length(left_matching_scheme_boundaries) == 0 && length(right_matching_scheme_boundaries) == 0
+    if length(left_scheme_boundaries) == 0 && length(right_scheme_boundaries) == 0
```
ali-ramadhan: Probably doesn't matter, but should this be checking the length of the new tuples `boundary_fluxes.left_scheme_boundaries` and `boundary_fluxes.right_scheme_boundaries` since those are the ones that will be used?

tomchor:
```suggestion
    if length(boundary_fluxes.left_scheme_boundaries) == 0 && length(boundary_fluxes.right_scheme_boundaries) == 0
```

tomchor: Change made!

src/BoundaryConditions/BoundaryConditions.jl
```diff
@@ -41,5 +41,5 @@ include("compute_flux_bcs.jl")
 include("update_boundary_conditions.jl")
 include("polar_boundary_condition.jl")
 
-include("perturbation_advection_open_boundary_scheme.jl")
+include("perturbation_advection_open_scheme.jl")
```
glwagner:
```suggestion
include("perturbation_advection.jl")
```

this is the convention that a file which introduces a type is named after the type

tomchor: Sounds good! Will make the change soon

tomchor: Done

src/Models/NonhydrostaticModels/boundary_mass_fluxes.jl
```diff
@@ -79,7 +79,7 @@ initialize_boundary_mass_flux(velocity, ::Nothing, side) = NamedTuple()
 initialize_boundary_mass_flux(velocity, bc, side) = NamedTuple()
 
 needs_mass_flux_correction(::IOBC) = false
-needs_mass_flux_correction(bc::OBC) = bc.classification.matching_scheme !== nothing
+needs_mass_flux_correction(bc::OBC) = bc.classification.scheme !== nothing
```
simone-silvestri:
```suggestion
needs_mass_flux_correction(::OBC) = true
```

tomchor: I don't think that'd work because that would also flag OpenBoundaryConditions without a matching scheme as needing correction, where we're basically imposing a flow. I think we'd like those boundary conditions to have the exact flow/flux we impose on them, no?

simone-silvestri: Isn't that took care of here?
```julia
needs_mass_flux_correction(::IOBC) = false
```

tomchor: Ah, that's right! I missed that. This PR has been sitting here so long I forgot the details lol

tomchor:
The tests pass, with the exception of reactant tests and an error in the docs. None of these has anything to do with this PR. So @glwagner @navidcy @simone-silvestri can someone please merge this?

glwagner:
I'll merge when the docs build successfully, which I think we should fix before taking 0.98.0. I fixed a doctest error so hopefully it'll build this time.

navidcy:
Thanks @glwagner!

tomchor:
> I'll merge when the docs build successfully, which I think we should fix before taking 0.98.0. I fixed a doctest error so hopefully it'll build this time.

Sounds good, thanks. I thought we wanted to fix this error in another PR since it didn't come from the changes here which is why I didn't fix it myself. I'll keep an eye on the docs and fix any other errors in case that might pop up.

Issue #4689, opened 2025-07-31
`JLD2Writer()` does not save all outputs if `schedule` is passed as `output_interval = TimeInterval()`


taimoorsohail:
Hi all, 

I have found unexpected behaviour with `JLD2Writer` in Oceananigans:

```Julia
output_intervals = TimeInterval(0.1)

@time simulation.output_writers[:surfacetracers] = JLD2Writer(model, tracers;
                                                 schedule = output_intervals,
                                                 filename = "global_surface_fields_test_tracersocngns")

@time simulation.output_writers[:surfacevels] = JLD2Writer(model, velocities;
                                                 schedule = output_intervals,
                                                 filename = "global_surface_fields_test_velocitiesocngns")
```

Only the first output write saves! If instead I do:

```Julia

@time simulation.output_writers[:surfacetracers] = JLD2Writer(model, tracers;
                                                 schedule = TimeInterval(0.1),
                                                 filename = "global_surface_fields_test_tracersocngns")

@time simulation.output_writers[:surfacevels] = JLD2Writer(model, velocities;
                                                 schedule = TimeInterval(0.1),
                                                 filename = "global_surface_fields_test_velocitiesocngns")
```
Then the simulation outputs as expected...

glwagner:
Hm yes you cannot re-use a schedule. To fix that you may need to `deepcopy` a schedule that is passed to an output writer

glwagner:
We could add a line to `JLD2Writer` and `NetCDFWriter` to achieve this. I'm trying to understand if we should define all schedules as independent or whether there are cases where we would not want to deepcopy.

taimoorsohail:
Sorry, I meant to mention that this ONLY breaks with `TimeInterval()`. When I do `IterationInterval()` it works either way...

glwagner:
That's because `TimeInterval` (and possibly other schedules) involve the notion of a "schedule state". Thus schedules cannot be shared between two actions; the schedule can be used only once. IterationInterval is a special case because it does not have a state.

taimoorsohail:
Perhaps we could add a warning that says "schedule is being reused, $(output) will not be saved!" or something? That will at least prompt the user to not do this. I think it is pretty intuitive to reuse schedules, if, for example, you have 5 output writers with the same output schedule, you don't want to be going in and manually adjusting each every time right?

navidcy:
I don't see any notion of state in 

[...]

PR #4690, opened 2025-07-31
Allow users to reuse a predefined `schedule` in many output writers

navidcy:
This PR ensures that a schedule provided to an output writer is deep-copied. This need comes up if a user defines a schedule by which they mean to activate several output writers. Before this PR, TimeInterval schedules after activated the first output writer they wouldn't activate other output writers they were  used in.

Closes https://github.com/CliMA/Oceananigans.jl/issues/4689

### DIFF
--- a/src/Utils/schedules.jl
+++ b/src/Utils/schedules.jl
@@ -94,7 +94,7 @@ end
 """
     IterationInterval(interval; offset=0)
 
-Return a callable `IterationInterval` that "actuates" (schedules output or callback execution)
+Return a callable `IterationInterval` that "actuates" (i.e., schedules output or callback execution)
 whenever the model iteration (modified by `offset`) is a multiple of `interval`.
 
 For example,
@@ -157,7 +157,7 @@ end
 """
     SpecifiedTimes(times)
 
-Return a `schedule::SpecifiedTimes` that "actuates" (schedules output or callback execution)
+Return a `schedule::SpecifiedTimes` that "actuates" (i.e., schedules output or callback execution)
 whenever the model's clock equals the specified values in `times`. For example,
 
 * `SpecifiedTimes([1, 15.3])` actuates when `model.clock.time` is `1` and `15.3`.


src/OutputWriters/jld2_writer.jl
```diff
@@ -164,6 +164,8 @@ function JLD2Writer(model, outputs; filename, schedule,
     outputs = NamedTuple(Symbol(name) => construct_output(outputs[name], model.grid, indices, with_halos)
                          for name in keys(outputs))
 
+    schedule = deepcopy(schedule) # assume user wants to reuse a schedule
+
     # Convert each output to WindowedTimeAverage if schedule::AveragedTimeWindow is specified
     schedule, outputs = time_average_outputs(schedule, outputs, model)
```
taimoorsohail, line 170: Should we do it also for this line, if the user uses `AveragedTimeWindow`? i.e. `time_average_outputs(deepcopy(schedule), outputs, model)`

navidcy: But it's already being deepcopied above; no need to redo, right?

glwagner: what you suggest is just a different way of writing the same code, isn't it?

glwagner:
Actually I realized that we don't want to deepcopy because this invalidates a huge class of possible schedules. We should put in code that only does a deepcopy for `TimeInterval` --- can we do that? Something like

```julia
schedule = materialize_schedule(schedule)
```

which has fallback

```julia
materialize_schedule(s) = s
materialize_schedule(ti::TimeInterval) = deepcopy(ti)
```

can we do that?

navidcy:
> Actually I realized that we don't want to deepcopy because this invalidates a huge class of possible schedules. We should put in code that only does a deepcopy for `TimeInterval` --- can we do that? Something like
> 
> ```julia
> schedule = materialize_schedule(schedule)
> ```
> 
> which has fallback
> 
> ```julia
> materialize_schedule(s) = s
> materialize_schedule(ti::TimeInterval) = deepcopy(ti)
> ```
> 
> can we do that?

Good catch! Yes, will do!

navidcy:
@glwagner looks ok?

src/Utils/schedules.jl
```diff
@@ -28,7 +28,7 @@ end
 #####
 
 """
-    struct TimeInterval <: AbstractSchedule
+    mutable struct TimeInterval{FT} <: AbstractSchedule
```
glwagner: suggest deleting the docstring for the struct (we decided it is just confusing and better to document only constructors)

src/Utils/schedules.jl
```diff
@@ -303,3 +303,5 @@ Base.summary(schedule::ConsecutiveIterations) = string("ConsecutiveIterations(",
                                                        summary(schedule.parent), ", ",
                                                        schedule.consecutive_iterations, ")")
 
+materialize_schedule(s) = s
+materialize_schedule(ti::TimeInterval) = deepcopy(ti) # required to reuse a pre-defined TimeInterval
```
glwagner:
```suggestion
const StatefulSchedules = Union{TimeInterval, SpecifiedTimes, ConsecutiveIterations}
materialize_schedule(ss::StatefulSchedules) = deepcopy(ss) # required to reuse a pre-defined schedule with a state
materialize_schedule(or::OrSchedule) = OrSchedule(Tuple(materialize_schedule(s) for s in or.schedules))
materialize_schedule(and::AndSchedule) = AndSchedule(Tuple(materialize_schedule(s) for s in and.schedules))
```

glwagner: I cooked this up, please check

navidcy: works!

PR #4691, opened 2025-08-01
Update spacings calls to enable array operations.

jlk9:
Similar to the PR here: #4472
We are adding additional methods to some Oceananigans Binary Operations (particularly in `spacings_and_areas_and_volumes.jl`) to extend their use with Reactant.jl.

This PR extends #4680 

@wsmoses


src/Grids/latitude_longitude_grid.jl
```diff
@@ -28,7 +28,7 @@ struct LatitudeLongitudeGrid{FT, TX, TY, TZ, Z, DXF, DXC, XF, XC, DYF, DYC, YF,
     Δxᶠᶜᵃ :: DXFC
     Δxᶜᶠᵃ :: DXCF
     Δxᶠᶠᵃ :: DXFF
-    Δyᶠᶜᵃ :: DYFC
+    Δyᶠᶜᵃ :: DYFC # Problem
```
simone-silvestri: why is this causing a problem?

src/Operators/spacings_and_areas_and_volumes.jl
```diff
@@ -156,7 +158,9 @@ end
 @inline Δλᶠᵃᵃ(i, j, k, grid::LLGX) = @inbounds grid.Δλᶠᵃᵃ
 
 @inline Δφᵃᶜᵃ(i, j, k, grid::LLG)  = @inbounds grid.Δφᵃᶜᵃ[j]
+@inline Δφᵃᶜᵃ(i::AbstractArray, j::AbstractArray, k::AbstractArray, grid::LLG) = Base.stack(collect(collect(transpose(Base.stack(collect(Δφᵃᶜᵃ(1, j, 1, grid) for _ in i)))) for _ in k))
```
simone-silvestri, line 161: It seems like there is a lot of copy and paste going on here to apply these operators to arrays. Would it be possible to maybe collect all these extra methods in one file in the `Operators` module and use metaprogramming to extend them all in case they are used with arrays. Otherwise, we need to continuously add these methods little by little.

src/Grids/latitude_longitude_grid.jl
jlk9: I thought some of the types in LatLongGrid were complicating issues with the conditional operators. That is NOT the case - just added a commit to remove this comment.

src/Operators/spacings_and_areas_and_volumes.jl
jlk9: That's a good point - it would be useful to write helper functions that can define these additional methods and put them all in one place.

There is additional complexity beyond just copy/pasting, since whether the operators iterate over `I`, `j`, or `k` means the arrays need transpose / dimension permutations that vary in each case.

glwagner: It's also very hard to understand the chain of 7 functions... if there is any way to clarify the code by writing it differently that would be nice.

wsmoses: probably we should add a function utility 

broadcast_i_to_ijk(i, j, k, val) = Base.stack(collect(collect(transpose(Base.stack(collect(val for _ in i)))) for _ in k))

broadcast_j_to_ijk(i, j, k, val) = permutedims(Base.stack(collect(collect(transpose(Base.stack(collect(val for _ in i)))) for _ in k)), ...)

broadcast_k_to_ijk(i, j, k, val) = permutedims(Base.stack(collect(collect(transpose(Base.stack(collect(val for _ in i)))) for _ in k)), ...)


that could be called by this and all similar functions

glwagner: This is very helpful!

Are the `j` and `k` broadcast correct? They are identical.

glwagner (approved):
I am approving so we can move forward but I would really like to see a future PR that systematizes the code (eg following @wsmoses comments), and collects all of these Reactant-related extensions into one file which will probably accelerate future development and help future maintenance.

src/Operators/spacings_and_areas_and_volumes.jl
wsmoses: So given the ... I didn't finish them here xD

PR #4695, opened 2025-08-02
fjp/spherical coriolis

francispoulin:
I have created a first attempt at this PR to create the SphericalCoriolis as discussed in #4683 with @glwagner.  

Note that none of the tests will work because I have not done a global search and replace for `HydrostaticSphericalCoriolis` for `SphericalCoriolis`, but I'm happy to do that after we have this set up the way we want.   We hope this will not be a breaking change, if done correctly.

In the new script, [spherical_coriolis.jl](https://github.com/CliMA/Oceananigans.jl/blob/fjp/SphericalCoriolis/src/Coriolis/spherical_coriolis.jl), I currently have an if statement that lets me switch from nonhydrostatic to hydrostatic model.  This is not of course what we want but was the easiest thing for me to set up.  

Below I have what I think are the two most pressing questions.

Question 1:
What can we use (or create) so that this script can determine whether we are using a nonhydrostatic or hydrostatic model?  I am happy to try whatever ideas people think are best. 

Question 2:
Can somone look at the nonhydrostatic terms to see if these are completed correctdly?  Maybe @simone-silvestri ?


src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,127 @@
...
+using Oceananigans.Fields
+using Oceananigans.AbstractOperations: KernelFunctionOperation
+using Oceananigans.ImmersedBoundaries
+
+"""
+    struct SphericalCoriolis{S, FT} <: AbstractRotation
+
+A parameter object for constant rotation around a vertical axis on the sphere.
+"""
+struct SphericalCoriolis{S, FT} <: AbstractRotation
```
simone-silvestri: I think we want to have something like this
```suggestion
struct SphericalCoriolis{H, S, FT} <: AbstractRotation
```
where the parameter `H` informs if the hydrostatic approximation is active or not. With maybe 
```julia
const HydrostaticSphericalCoriolis{S, FT} = SphericalCoriolis{true, S, FT}
const NonhydrostaticSphericalCoriolis{S, FT} = SphericalCoriolis{false, S, FT}
```
or maybe we can define a new type `HydrostaticCoriolis` and `NonhydrostaticCoriolis` and 
```julia
const HydrostaticSphericalCoriolis{S, FT} = SphericalCoriolis{HydrostaticCoriolis, S, FT}
const NonhydrostaticSphericalCoriolis{S, FT} =  SphericalCoriolis{NonhydrostaticCoriolis, S, FT}

```
and then dispatch the pointwise forcing terms on the sperical coriolis type, for example:
```julia
    @inline f̃_ℑx_wᶠᶠᵃ(i, j, k, grid, coriolis::NonhydrostaticSphericalCoriolis, v) = f̃ᶠᶠᵃ(i, j, k, grid, coriolis) * ℑxᶠᵃᵃ(i, j, k, grid, Δx_qᶜᶠᶜ, w)
    @inline f̃_ℑx_wᶠᶠᵃ(i, j, k, grid, ::HydrostaticSphericalCoriolis, v) = zero(grid)
```
Also I think the locations of `f̃_ℑx_wᶠᶠᵃ`and `f̃_ℑz_uᶠᶠᵃ` are not quite correct.

francispoulin: Thanks @simone-silvestri , very helpful.  I will look at this tomorrow and try and push some improvements.

francispoulin:
@simone-silvestri , I have pushed a few changes and wanted to summarize what I have tried to do.

First, I ignored the energy of enstrophy conserving schemes for now. I thought we can first get the dispatch working for hydrostatic and nonhydrostatic, and then distpach between the two conservative schemes.  This seemed easier for me for the moment.

Second, I defined two types in `Coriolis.jl`: `HydrostaticCoriolis` and `NonhydrostaticCoriolis` that we can use to dispatch between.  

Third, in the same file, I also defined `HydrostaticSphericalCoriolis` and `NonhydrostaticSphericalCoriolis` as you suggested. 

Fourth, I changed the averaging operators so f̃*w and f̃*u are zero in the hydrostatic case.  

You pointed out that the term `f̃_ℑx_wᶠᶠᵃ` and `f̃_ℑz_uᶠᶠᵃ` are not correct, and I can believe that as I'm still trying to understand the formulation. 

In the first term, I was wrong as I was using v instead of w.  I fixed that.  Is the averaging done incorrectly too?

I'm also not sure what's wrong in the second term but if you can give me a clue I'm happy to investigate.

src/Coriolis/nonhydrostatic_spherical_coriolis.jl
```diff
@@ -0,0 +1,71 @@
+using Oceananigans.Grids: LatitudeLongitudeGrid, OrthogonalSphericalShellGrid, peripheral_node, φnode
+using Oceananigans.Operators: Δx_qᶜᶠᶜ, Δy_qᶠᶜᶜ, Δxᶠᶜᶜ, Δyᶜᶠᶜ, hack_sind, hack_cosd
+using Oceananigans.Advection: EnergyConserving, EnstrophyConserving
```
simone-silvestri: This file can be deleted right?

src/Coriolis/Coriolis.jl
```diff
@@ -25,14 +25,21 @@ Abstract supertype for parameters related to background rotation rates.
...
 
+abstract type HydrostaticCoriolis end
+
+abstract type NonhydrostaticCoriolis end
+
 const face = Face()
 const center = Center()
 
+const HydrostaticSphericalCoriolis{S, FT} = SphericalCoriolis{HydrostaticCoriolis, S, FT}
+const NonhydrostaticSphericalCoriolis{S, FT} =  SphericalCoriolis{NonhydrostaticCoriolis, S, FT}
```
simone-silvestri: I think this should go in the `spherical_coriolis.jl` file

src/Coriolis/nonhydrostatic_spherical_coriolis.jl
francispoulin: Yes, and it is now deleted.  Thanks.

src/Coriolis/Coriolis.jl
francispoulin: I just moved those as well.

src/Coriolis/spherical_coriolis.jl
```diff
@@ -1,25 +1,27 @@
...
-    struct HydrostaticSphericalCoriolis{S, FT} <: AbstractRotation
+    struct SphericalCoriolis{S, FT} <: AbstractRotation
 
 A parameter object for constant rotation around a vertical axis on the sphere.
 """
-struct HydrostaticSphericalCoriolis{S, FT} <: AbstractRotation
+struct SphericalCoriolis{H, S, FT} <: AbstractRotation
     rotation_rate :: FT
     scheme :: S
+    HydrostaticCoriolis :: H
```
glwagner: Will this work? Whether or not we are using the hydrostatic approximation has to be encoded in a type, which is known at compile time.

src/Coriolis/Coriolis.jl
```diff
@@ -25,6 +25,10 @@ Abstract supertype for parameters related to background rotation rates.
 """
 abstract type AbstractRotation end
 
+abstract type HydrostaticCoriolis end
+
+abstract type NonhydrostaticCoriolis end
```
glwagner: these need to be type aliases, not abstract types

glwagner:
I don't think we need to bother with detecting whether the model is nonhydrostatic or not. I think we should just let the user choose it.

src/Coriolis/spherical_coriolis.jl
francispoulin: I can't say that it will work.  This was a suggestion that I was trying out.  

I think that we need a flag to specify before we define `Coriolis` or `model`, that says whether we want hydrostatic or non-hydrostatic, which is not ideal as that is one more thing the user needs to speicfy.

Do you have a different idea on how to do this?

francispoulin:
> I don't think we need to bother with detecting whether the model is nonhydrostatic or not. I think we should just let the user choose it.

I think that's a simple and attracative idea.  If the user specifies it then we can choose the Coriolis and model types very easily.  

Were you thinking of specifying something like `ModelType` to be Nonhydrostatic or Hydrostatic?  Lots of ways this can go and I admit that I don't have a strong feeling as to what would be ideal for the user.  Any suggestions are welcome.

Maybe you were suggesting that we simply pick the Coriolis type to be eieher `HydrostaticCoriolis` or `NonhydrostaticCoriolis` ?  That makes sense to me and seems that it could work well.

src/Coriolis/Coriolis.jl
francispoulin: I'm happy to change this but maybe best after we decide on how to swtich from one to the other.

src/Coriolis/spherical_coriolis.jl
glwagner: i propose this API:

```julia
hydrostatic_coriolis = HydrostaticSphericalCoriolis()
full_coriolis = SphericalCoriolis()
```

for implementation, we need a type such that

```julia
const HydrostaticSphericalCoriolis{S, FT} = SphericalCoriolis{S, FT, <:HydrostaticFormulation} where {S, FT}
```

note this preserves the type structure on `main` so it is not a breaking PR.

one can introduce the concept of the "formulation" like this

```julia
struct HydrostaticFormulation end
struct NonhydrostaticFormulation end

struct SphericalCoriolis{S, FT, F}
    rotation_rate :: FT
    scheme :: S
    formulation :: F
end
```

francispoulin: Thanks @glwagner .  I'll give this a try  and get back to you when I have something together.

francispoulin:
@glwagner , I have tried to set things up the way you suggested.  Unfortuantely, my programming skills in julia leave a lot to be desired.   Could you take a look at let me know if this is what you had in mind?

Also, any suggests on how to clean this up would be greatly appreciated.

It doesn't run now but I think if we can get the structure set up, the calculations should be easier.

src/Coriolis/Coriolis.jl
glwagner: The user will decide which one to use in their script.

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,136 @@
...
+using Oceananigans.ImmersedBoundaries
+
+
+"""
+Will use the following syntax:
+
+hydrostatic_coriolis = HydrostaticSphericalCoriolis()
+full_coriolis = SphericalCoriolis()
+
+"""
```
simone-silvestri:
```suggestion
```

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,136 @@
...
+
+struct HydrostaticFormulation end
+struct NonhydrostaticFormulation end
+
```
simone-silvestri, line 10:
```suggestion
```

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,136 @@
...
+    rotation_rate :: FT
+    scheme :: S
+    formulation :: F
+end
+
+"""
+    SphericalCoriolis([FT=Float64;]
+                                 rotation_rate = Ω_Earth,
+                                 scheme = EnstrophyConserving())
+                                 formulation = HydrostaticFormulation/NonhydrostaticFormulation
```
simone-silvestri:
```suggestion
                                 formulation = HydrostaticFormulation()
```

francispoulin:
I worked on this a bit more but still not sure if I am getting the dispatch between what `formulation` and what `scheme` to use.  Any advice @simone-silvestri and @glwagner ?

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,130 @@
...
+
+@inline f̃_ℑz_uᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, formulation::HydrostaticFormulation, u) = zero(grid)
+@inline f̃_ℑx_wᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, formulation::HydrostaticFormulation, w) = zero(grid)
+
+@inline x_f_cross_U(i, j, k, grid, coriolis::SphericalCoriolis, U) = @inbounds - ℑyᵃᶜᵃ(i, j, k, grid, f_ℑx_vᶠᶠᵃ, coriolis, U[2]) * Δx⁻¹ᶠᶜᶜ(i, j, k, grid) +
+                                                                                 ℑyᵃᶜᵃ(i, j, k, grid, f̃_ℑx_wᶠᶠᵃ, coriolis, U[3]) * Δx⁻¹ᶠᶜᶜ(i, j, k, grid)
+@inline y_f_cross_U(i, j, k, grid, coriolis::SphericalCoriolis, U) = @inbounds + ℑxᶜᵃᵃ(i, j, k, grid, f_ℑy_uᶠᶠᵃ, coriolis, U[1]) * Δy⁻¹ᶜᶠᶜ(i, j, k, grid)
+@inline z_f_cross_U(i, j, k, grid, coriolis::SphericalCoriolis, U) = @inbounds - ℑxᶜᵃᵃ(i, j, k, grid, f̃_ℑz_uᶠᶠᵃ, coriolis, U[1]) * Δz⁻¹ᶜᶜᶠ(i, j, k, grid)
+
+@inline z_f_cross_U(i, j, k, grid, coriolis::SphericalCoriolis, formulation::HydrostaticFormulation, U) = zero(grid)
```
simone-silvestri: can you do something like
```suggestion
@inline z_f_cross_U(i, j, k, grid, ::HydrostaticSphericalCoriolis, U) = zero(grid)
```

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,130 @@
...
+@inline f̃ᶠᶠᵃ(i, j, k, grid, coriolis) = 2 * coriolis.rotation_rate * hack_cosd(φᶠᶠᵃ(i, j, k, grid))
+
+@inline f_ℑy_uᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, u) = fᶠᶠᵃ(i, j, k, grid, coriolis) * ℑyᵃᶠᵃ(i, j, k, grid, Δy_qᶠᶜᶜ, u)
+@inline f_ℑx_vᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, v) = fᶠᶠᵃ(i, j, k, grid, coriolis) * ℑxᶠᵃᵃ(i, j, k, grid, Δx_qᶜᶠᶜ, v)
+
+@inline f̃_ℑz_uᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, u) = f̃ᶠᶠᵃ(i, j, k, grid, coriolis) * ℑyᵃᶠᵃ(i, j, k, grid, Δz_qᶜᶜᶠ, u)
+@inline f̃_ℑx_wᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, w) = f̃ᶠᶠᵃ(i, j, k, grid, coriolis) * ℑxᶠᵃᵃ(i, j, k, grid, Δx_qᶜᶠᶜ, w)
+
+@inline f̃_ℑz_uᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, formulation::HydrostaticFormulation, u) = zero(grid)
+@inline f̃_ℑx_wᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, formulation::HydrostaticFormulation, w) = zero(grid)
```
simone-silvestri:
```suggestion
@inline f̃_ℑz_uᶠᶠᵃ(i, j, k, grid, ::HydrostaticSphericalCoriolis, u) = zero(grid)
@inline f̃_ℑx_wᶠᶠᵃ(i, j, k, grid, ::HydrostaticSphericalCoriolis, w) = zero(grid)
```

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,130 @@
...
+
+#####
+##### Active Point Enstrophy-conserving scheme
+#####
+
+# It might happen that a cell is active but all the neighbouring staggered nodes are inactive,
+# (an example is a 1-cell large channel)
+# In that case the Coriolis force is equal to zero
+
+const CoriolisEnstrophyConserving = SphericalCoriolis{<:HydrostaticFormulation, <:EnstrophyConserving}
```
simone-silvestri:
```suggestion
const CoriolisEnstrophyConserving = SphericalCoriolis{<:EnstrophyConserving, <:Any, <:HydrostaticFormulation}
```

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,130 @@
...
+
+@inline y_f_cross_U(i, j, k, grid, coriolis::CoriolisEnstrophyConserving, U) =
+    @inbounds + ℑxᶜᵃᵃ(i, j, k, grid, fᶠᶠᵃ, coriolis) *
+                active_weighted_ℑxyᶜᶠᶜ(i, j, k, grid, Δy_qᶠᶜᶜ, U[1]) * Δy⁻¹ᶜᶠᶜ(i, j, k, grid)
+
+#####
+##### Energy-conserving scheme
+#####
+
+const CoriolisEnergyConserving = SphericalCoriolis{<:HydrostaticFormulation, <:EnergyConserving}
```
simone-silvestri:
```suggestion
const CoriolisEnergyConserving = SphericalCoriolis{<:EnergyConserving, <:Any, <:HydrostaticFormulation}
```

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,130 @@
...
+    return SphericalCoriolis(rotation_rate, scheme, formulation)
+end
+
+const HydrostaticSphericalCoriolis{S, FT}    = SphericalCoriolis{S, FT, <:HydrostaticFormulation} where {S, FT}
+const NonhydrostaticSphericalCoriolis{S, FT} = SphericalCoriolis{S, FT, <:NonhydrostaticFormulation} where {S, FT}
+
+# using Oceananigans.Coriolis: HydrostaticFormulation, NonhydrostaticFormulation
+# model_hy = HydrostaticFreeSurfaceModel(; grid, coriolis=SphericalCoriolis(scheme=EnstrophyConserving(), formulation=HydrostaticFormulation()))
+# model_nh = NonhydrostaticModel(; grid=grid, coriolis=SphericalCoriolis(formulation=NonhydrostaticFormulation()))
+
```
simone-silvestri:
```suggestion
```

validation/coriolis/enceladus_convection.jl
```diff
@@ -0,0 +1,225 @@
...
+                             radius = R_enceladus,
+                             topology = (Periodic, Bounded, Bounded))
+
+# ## Model
+#
+# We use a NonhydrostaticModel with buoyancy as a tracer.
+# For Enceladus, we use a simple buoyancy formulation since
+# we're simulating convection in a liquid water ocean.
+using Oceananigans.Coriolis: SphericalCoriolis, NonhydrostaticFormulation
+coriolis = SphericalCoriolis(rotation_rate=Ω_enceladus, formulation=NonhydrostaticFormulation())
```
glwagner, line 156: @francispoulin this is how the user should select the nonhydrostatic formulation. (it is also the default for `SphericalCoriolis`)

glwagner:
@francispoulin I put some work into this:

* there were bugs in the implementation (missing arguments in functions, wrong operators called). @francispoulin can you please look over the math / numerics and verify that it is correct?
* This PR should not be breaking. We will continue to export the constructor `HydrostaticSphericalCoriolis` with the same meaning.
* This PR _adds_ a new function, `SphericalCoriolis`, with a formulation that defaults to `NonhydrostaticFormulation` and hopefully can be used in nonhydrostatic models on curvilinear grids, like `LatitudeLongitudeGrid`.
* I added a validation case that uses the new coriolis term. I did not run it for very long. I was also intrigued by the equation of state on enceladus so there is some random stuff in the script related to that 😂

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,132 @@
...
+@inline φᶠᶠᵃ(i, j, k, grid::LatitudeLongitudeGrid)        = φnode(j, grid, face)
+@inline φᶠᶠᵃ(i, j, k, grid::OrthogonalSphericalShellGrid) = φnode(i, j, grid, face, face)
+@inline φᶠᶠᵃ(i, j, k, grid::ImmersedBoundaryGrid)         = φᶠᶠᵃ(i, j, k, grid.underlying_grid)
+
+@inline fᶠᶠᵃ(i, j, k, grid, coriolis) = 2 * coriolis.rotation_rate * hack_sind(φᶠᶠᵃ(i, j, k, grid))
+@inline f̃ᶠᶠᵃ(i, j, k, grid, coriolis) = 2 * coriolis.rotation_rate * hack_cosd(φᶠᶠᵃ(i, j, k, grid))
+
+@inline f_ℑy_uᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, u) = fᶠᶠᵃ(i, j, k, grid, coriolis) * ℑyᵃᶠᵃ(i, j, k, grid, Δy_qᶠᶜᶜ, u)
+@inline f_ℑx_vᶠᶠᵃ(i, j, k, grid, coriolis::SphericalCoriolis, v) = fᶠᶠᵃ(i, j, k, grid, coriolis) * ℑxᶠᵃᵃ(i, j, k, grid, Δx_qᶜᶠᶜ, v)
```
glwagner:
```suggestion
@inline f_ℑy_uᶠᶠᶜ(i, j, k, grid, coriolis::SphericalCoriolis, u) = fᶠᶠᵃ(i, j, k, grid, coriolis) * ℑyᵃᶠᵃ(i, j, k, grid, Δy_qᶠᶜᶜ, u)
@inline f_ℑx_vᶠᶠᶜ(i, j, k, grid, coriolis::SphericalCoriolis, v) = fᶠᶠᵃ(i, j, k, grid, coriolis) * ℑxᶠᵃᵃ(i, j, k, grid, Δx_qᶜᶠᶜ, v)
```

glwagner:
I am fairly sure that an incorrect discretization was being used. https://github.com/CliMA/Oceananigans.jl/pull/4695/commits/7b7db5468a29065bf9109ddfa9cc8b80fde5d71a tries to address it

francispoulin:
Thank you @glwagner !  That is great that you advanced this so much.  I'm happy to look it over and do some testing.   I will fix whatever I can, including sticking to American spelling.

src/Coriolis/spherical_coriolis.jl
```diff
@@ -0,0 +1,132 @@
...
+using Oceananigans.Fields
+using Oceananigans.ImmersedBoundaries
+
+struct HydrostaticFormulation end
+struct NonhydrostaticFormulation end
+
+"""
+    struct SphericalCoriolis{S, FT, F} <: AbstractRotation 
+
+A parameter object for constant rotation around a vertical axis on the sphere.
```
francispoulin: Can we say `constant rotation rate`? We are saying Omega is constant but don't want people to think that f is constant, which it is not.

glwagner: yes! although, we do not like docstrings on structs. so i removed it anyways. the constructor docstring says rotation rate

francispoulin: Thanks!

validation/coriolis/enceladus_convection.jl
```diff
@@ -0,0 +1,225 @@
...
+const Ω_enceladus = 2π / 32.9hours # 1/s, rotation rate of Enceladus
+
+# ## Plotting thermal expansion to haline contraction ratio
+# Create heatmaps of R = α/β where α is thermal expansion coefficient and β is haline contraction coefficient.
+
+# Parameters
+Nd, NT, NS = 256, 256, 256
+z_min, z_max = -40e3, 0
+T_min, T_max, S_min, S_max = -2, 20, -2, 40
+S1 = 0.0  # psu, fixed salinity for R_TD
```
francispoulin, line 40: Maybe an obvious answer but what to TD and SD stand for?  Maybe we could spell this out in the comments?

glwagner: just "studying temperature-depth" or "salinity-depth". I think we should focus on the source code though, this is just a throw away script to run for fun!

francispoulin:
I amrunning `enceladus_convection.jl` and happy to say that the two figures were generated quickly.  But then it takes half an hour before the grid is defined.  

@glwagner , I presume this didn't happen for you in your testing?

I'll run it to the end to make sure everything works for me and then maybe then run it for longer time.  Seems like a good first step.

validation/coriolis/enceladus_convection.jl
francispoulin: Thanks and I agree.

src/Coriolis/Coriolis.jl
```diff
@@ -2,7 +2,8 @@ module Coriolis
 
 export
     FPlane, ConstantCartesianCoriolis, BetaPlane, NonTraditionalBetaPlane,
-    HydrostaticSphericalCoriolis,
+    SphericalCoriolis, HydrostaticSphericalCoriolis,
+    ActiveCellEnstrophyConserving,
```
giordano: Doesn't exist?  See https://buildkite.com/clima/oceananigans/builds/27269#019ae52d-78a5-41eb-a492-cb29abba771f/17-460
```
Undefined exports detected:
1-element Vector{Symbol}:
 Symbol("Oceananigans.Coriolis.ActiveCellEnstrophyConserving")
Undefined exports: Test Failed at /var/lib/buildkite-agent/.julia-oceananigans/packages/Aqua/MCcFg/src/exports.jl:66
  Expression: isempty(exports)
   Evaluated: isempty([Symbol("Oceananigans.Coriolis.ActiveCellEnstrophyConserving")])
```
```suggestion
```

glwagner:
> I amrunning `enceladus_convection.jl` and happy to say that the two figures were generated quickly. But then it takes half an hour before the grid is defined.
> 
> @glwagner , I presume this didn't happen for you in your testing?
> 
> I'll run it to the end to make sure everything works for me and then maybe then run it for longer time. Seems like a good first step.

No, but maybe that has to do with precompiling CairoMakie or something. It has nothing to do with this PR though -- if you want to define a simpler case for this PR (we also need a test btw) you should!

francispoulin:
The error we are getting now is with `test_quality_assurance.jl`.  I believe this is the line but I have not used Aqua before so not sure what we should expect.  Any thougths?

```
julia> ExplicitImports.check_no_stale_explicit_imports(Oceananigans)
ERROR: StaleImportsException
Module `Oceananigans.Coriolis` has stale (unused) explicit imports for:
* `defaults`
```
============== Addition =================

This problem went away. Please ignore this comment

glwagner:
if tests pass I am inclined to merge this. We can continue validation separately.

francispoulin:
All checks have passed, all we need is someone to approve it and then I am happy to merge, assuming people are happy.

glwagner:
very happy. merge away!

PR #4696, opened 2025-08-06
(0.97.6) Move the computation of flux BC before the time-stepping

simone-silvestri:
This PR is moving the computation of the boundary conditions just before the time-stepping of the prognostic variables, rather than at the end of the time-stepping procedure.

This is because, in Oceananigans, we add the flux bcs to the tendencies, which are computed at the end of a time step. However, if we need to couple a model, the exchange between components is typically computed at the end of the stepping of all the components, meaning that, if we do this, we do not add the updated fluxes to the tendency in an Oceananigans model; rather, the fluxes corresponding to the previous time-step. 

This might become an issue when coupling on large time steps, where the fluxes lag behind and the model can overshoot. 
This change will not affect oceananigans-only models, but will affect coupled models like the `OceanSeaIceModel` in ClimaOcean.jl.

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/explicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/explicit_free_surface.jl
@@ -134,8 +134,7 @@ The tendency is called ``G_η`` and defined via
     k_top = grid.Nz + 1
     model_fields = merge(hydrostatic_fields(velocities, free_surface, tracers), auxiliary_fields)
 
-    return @inbounds (  velocities.w[i, j, k_top]
-                      + forcings.η(i, j, k_top, grid, clock, model_fields))
+    return @inbounds velocities.w[i, j, k_top] + forcings.η(i, j, k_top, grid, clock, model_fields)
 end
 
 compute_free_surface_tendency!(grid, model, ::ExplicitFreeSurface) =


src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
```diff
@@ -134,6 +134,7 @@ include("barotropic_pressure_correction.jl")
 include("hydrostatic_free_surface_tendency_kernel_functions.jl")
 include("compute_hydrostatic_free_surface_tendencies.jl")
 include("compute_hydrostatic_free_surface_buffers.jl")
+include("apply_hydrostatic_flux_bcs.jl")
```
glwagner: we could use this PR to finally fix the rogue "apply" ?

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_ab2_step.jl
```diff
@@ -20,6 +21,7 @@ function ab2_step!(model::HydrostaticFreeSurfaceModel, Δt)
 
     # Step locally velocity and tracers
     @apply_regionally begin
+        multiply_by_grid_scaling!(model.timestepper.Gⁿ, model.tracers, model.grid)
```
glwagner:
```suggestion
        scale_by_stretching_factor!(model.timestepper.Gⁿ, model.tracers, model.grid)
```

glwagner: "grid scaling" might be ambiguous

src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
glwagner: and use `compute_flux_bcs`, for example

navidcy:
> This is because, in Oceananigans, we add the flux bcs to the tendencies, which are computed at the end of a time step. 

So, how was this working for multi-step timestepping schemes? Before this PR the flux bcs were added at the end of the Δt or at the end of each substep?

navidcy (approved):

simone-silvestri:
@navidcy This is the difference between main and this PR in a one-step method like AB2 blue boxes are computed in ClimaOcean, black boxes in Oceanananigans.
RK3 is a little more complex but still has only one blue box per complete Oceananigans timestep.

ab2-timestepping

simone-silvestri:
This is a bit oversemplified, but it is the RK3 timestepper
rk3-timestep

If we want to compute fluxes each substep we probably need a timestepper for the `OceanSeaIceModel` which would probably help a lot with stabiliy as we could substep the sea ice with the ocean

PR #4699, opened 2025-08-06
Fix some inference problems in the `launch!` function

simone-silvestri:
This PR is trying to fix some inference issues of the `launch!` function. It is WIP though.


glwagner:
cc @bgroenks96

navidcy (approved):

bgroenks96:
@maximilian-gelbrecht May also be relevant for our Enzyme tests.

@simone-silvestri and @navidcy will you make a new release soon?

navidcy:
Yes, let me tag a new release.

glwagner:
@bgroenks96 note there are Enzyme and Reactant tests on Oceananigans CI --- so if something is missing / not working, a PR is welcomed to fill out the test suite so that issues get caught.

bgroenks96:
but I guess if it's working in the Oceananigans test suite then that must not be the issue

glwagner:
One possibility is that the issue you're encountering isn't covered by our test suite; in that case we should add more tests. The test suite is fairly barebones, I believe it tests just a few configurations.

PR #4700, opened 2025-08-07
Create methods for equality of ImmersedBoundaryGrids

tomchor:
While testing https://github.com/CliMA/Oceananigans.jl/pull/4660 I realized these methods do not exist. As a consequence, any two IBGs with the same underlying grid return `true` when tested with `==`.


glwagner (approved):
I believe this does not need to be a draft?

tomchor:
It's only a draft because I think there's more to be done for it to be robust. I also want to add more tests.

src/ImmersedBoundaries/immersed_boundary_grid.jl
```diff
@@ -131,3 +131,11 @@ function on_architecture(arch, ibg::IBG)
 end
 
 isrectilinear(ibg::IBG) = isrectilinear(ibg.underlying_grid)
+
+function Base.:(==)(grid1::IBG, grid2::IBG)
+    #check if grids are of the same type
```
navidcy: I don't understand the comment; the grids are of the same type if this method is triggered because both grids need to be `IBG` (from the typed args above), right?

tomchor: yes, sorry, that was a leftover comment that I forgot to cleanup. I'll do a pass for those right now.

Issue #4702, opened 2025-08-07
Unable to use continuous function with ValueBoundaryCondition


bgroenks96:
Hi guys,

Thanks for all of your outstanding work on Oceananigans, really happy to be working with you :)

For unit testing in our land model, I would like to be able to apply a simple periodic upper boundary condition to the temperature field. I figured this would be possible with `ValueBoundaryCondition(f)` where `f` is the function. This works fine for `FluxBoudnaryCondition`, as indicated in the docs. However, with `ValueBoundaryCondition`, I get an error. Here is the MWE:

```julia
using Oceananigans
using Oceananigans.BoundaryConditions: FieldBoundaryConditions, ValueBoundaryCondition, NoFluxBoundaryCondition, fill_halo_regions!

f(x, z, t) = sin(2π*t)
bc = ValueBoundaryCondition(f)
grid = RectilinearGrid(CPU(), size=(1,10), x=(0,1), z=(0,1), topology=(Periodic,Flat,Bounded))
field_bc = FieldBoundaryConditions(grid, (Center, Center, Nothing), top=bc, bottom=NoFluxBoundaryCondition())
field = CenterField(grid, boundary_conditions=field_bc)
fill_halo_regions!(field)
```

Error:
```
ERROR: MethodError: objects of type Oceananigans.BoundaryConditions.ContinuousBoundaryFunction{Nothing, Nothing, Nothing, Nothing, typeof(f), Nothing, Tuple{}, Nothing, Nothing} are not callable
Stacktrace:
  [1] getbc
    @ ~/.julia/packages/Oceananigans/bKiml/src/BoundaryConditions/boundary_condition.jl:113 [inlined]
  [2] getbc
    @ ~/.julia/packages/Oceananigans/bKiml/src/BoundaryConditions/boundary_condition.jl:112 [inlined]
  [3] right_gradient
    @ ~/.julia/packages/Oceananigans/bKiml/src/BoundaryConditions/fill_halo_regions_value_gradient.jl:13 [inlined]
```

@glwagner mentioned that `regularize_field_boundary_conditions` might fix this, but it didn't seem to work when I tried it.

simone-silvestri:
You cannot `fill_halo_regions!` of a field when the boundary conditions has possible dependencies outside the field itself (in this case, model time). For function value boundary conditions, you need to provide the dependencies to the `fill_halo_regions!` function. 
For example, if you have a model, the correct way to fill the halo regions is:
```julia
[...]

PR #4704, opened 2025-08-08
Regularize field boundary conditions with a location rather than a name

simone-silvestri:
Even if `regularize_field_boundary_conditions` is not necessarily thought to be user-facing, it is a bit easier to understand and to use if we pass a location rather than a `field_name`, so that other models can use it without needing to extend the `assumed_field_location` function in oceananigans to add the `field_name` they need.

comes from issue #4702


glwagner:
Something to be careful about here is that this function assumes a c-grid convention and assumes that face+bounded implies impenetrable boundary condition by default.

PR #4705, opened 2025-08-08
RK3 CATKE Implementation

simone-silvestri:
This PR introduces CATKE for RK3.
Substepping is not supported at the moment.

I am trying to reproduce figures 15 and 16 of [Wagner et al](https://agupubs.onlinelibrary.wiley.com/doi/epdf/10.1029/2024MS004522), I think the LES data I have might be wrong, so disregard the LES, but just looking at the time-step dependency, it looks like RK3 does help a bit with decreasing the dependency with respect to vanilla AB2. @glwagner, what do you think? Maybe if it is quick, we can try doing a recalibration with RK3?

## Vanilla AB2 (`tke_time_step = nothing`)
[ab2_step.pdf](https://github.com/user-attachments/files/21685527/ab2_step.pdf)

## Substepped AB2 (`tke_time_step = 2minutes`)
[ab2_step_2min.pdf](https://github.com/user-attachments/files/21685526/ab2_step_2min.pdf)

## Split RKE (`tke_time_step` does not matter)
[rk3_step.pdf](https://github.com/user-attachments/files/21685523/rk3_step.pdf)

Another option is to remove the `time_step_tke!` equation in case of an RK3 and add a term in the tendency computation like I am doing here: https://github.com/CliMA/Oceananigans.jl/blob/ss/omip-branch/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/explicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/explicit_free_surface.jl
@@ -81,7 +81,7 @@ explicit_ab2_step_free_surface!(free_surface, model, Δt, χ) =
 
 @kernel function _explicit_rk3_step_free_surface!(η, Δt, γⁿ, ζⁿ, Gⁿ, η⁻, Nz)
     i, j = @index(Global, NTuple)
-    @inbounds η[i, j, Nz+1] += ζⁿ * η⁻[i, j, Nz+1] + γⁿ * (η[i, j, Nz+1] + Δt * Gⁿ[i, j, Nz+1])
+    @inbounds η[i, j, Nz+1] = ζⁿ * η⁻[i, j, Nz+1] + γⁿ * (η[i, j, Nz+1] + Δt * Gⁿ[i, j, Nz+1])
 end
 
 @kernel function _explicit_ab2_step_free_surface!(η, Δt, χ, Gηⁿ, Gη⁻, Nz)
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_ab2_step.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_ab2_step.jl
@@ -12,7 +12,6 @@ import Oceananigans.TimeSteppers: ab2_step!
 function ab2_step!(model::HydrostaticFreeSurfaceModel, Δt)
 
     grid = model.grid
-
     compute_free_surface_tendency!(grid, model, model.free_surface)
 
     FT = eltype(grid)
@@ -21,7 +20,6 @@ function ab2_step!(model::HydrostaticFreeSurfaceModel, Δt)
 
     # Step locally velocity and tracers
     @apply_regionally begin
-        scale_by_stretching_factor!(model.timestepper.Gⁿ, model.tracers, model.grid)
         ab2_step_grid!(model.grid, model, model.vertical_coordinate, Δt, χ)
         ab2_step_velocities!(model.velocities, model, Δt, χ)
         ab2_step_tracers!(model.tracers, model, Δt, χ)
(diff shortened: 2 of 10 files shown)


src/Models/HydrostaticFreeSurfaceModels/explicit_free_surface.jl
```diff
@@ -81,7 +81,7 @@ explicit_ab2_step_free_surface!(free_surface, model, Δt, χ) =
 
 @kernel function _explicit_rk3_step_free_surface!(η, Δt, γⁿ, ζⁿ, Gⁿ, η⁻, Nz)
     i, j = @index(Global, NTuple)
-    @inbounds η[i, j, Nz+1] += ζⁿ * η⁻[i, j, Nz+1] + γⁿ * (η[i, j, Nz+1] + Δt * Gⁿ[i, j, Nz+1])
+    @inbounds η[i, j, Nz+1] = ζⁿ * η⁻[i, j, Nz+1] + γⁿ * (η[i, j, Nz+1] + Δt * Gⁿ[i, j, Nz+1])
```
glwagner, line 84: :-O

glwagner:
> just looking at the time-step dependency, it looks like RK3 does help a bit with decreasing the dependency with respect to vanilla AB2. @glwagner, what do you think? Maybe if it is quick, we can try doing a recalibration with RK3?

Ok, but the calibrations all used a very short time-step (I think). Is there a difference between AB2 and RK3 even in the limit of 0 time step?

simone-silvestri:
Let me give it a try. I think there are some differences between the two time-stepping schemes even for very little forcing. I ll try to plot them together now (and remove the LES which confuses a bit the story).

glwagner:
It should be sufficient just to compare two solutions with the same time step, but different schemes.

simone-silvestri:
Indeed, there seems to be a systematic difference between RK3 and AB2 (in this case this is the "high-quality" AB2 with TKE substepping and a tke timestep of 2minutes)
[catke_time_step_dependence_weak_wind_strong_cooling.pdf](https://github.com/user-attachments/files/21714932/catke_time_step_dependence_weak_wind_strong_cooling.pdf)

different-tests

glwagner:
Does this mean there is a bug in the time stepping? Shouldn't the solutions converge as dt -> 0?

simone-silvestri:
Yeah it definitely must be a bug with implicit - explicit timestepping with RK3.

Actually rather than I bug, I would call it a wrong implementation, #4710 corrects it.

src/Models/HydrostaticFreeSurfaceModels/explicit_free_surface.jl
simone-silvestri: yeah, probably I should open a different PR and add a test for an explicit free surface 😅

simone-silvestri:
I have merged #4710 to this branch, am producing new results, and updating the figures at the top of the page. Just to report quickly here, the difference between AB2 and RK3 following the correction in #4710.
Results are not too bad!

## RK3 vs AB2 with a 2-minute substepping for TKE:
image

## RK3 vs AB2 without substepping for TKE:
image

simone-silvestri:
Should we merge this?

glwagner:
We need RK3 for TKEDissipationVerticalDiffusivity as well, which is better validated for global simulations.

glwagner:
@taimoorsohail I think in the future, if there are reviewers that have looked at a PR but not yet signed off (in this case, me), please wait for all reviewers to approve before merging.

taimoorsohail:
Sorry @glwagner -- I was told this was ready to merge and I offered to approve and merge, an offer which was accepted. I won't be doing this in the future!

PR #4706, opened 2025-08-08
Reduce boundary conditions slowdown

simone-silvestri:
A lot of slowdown in the `fill_halo_regions!` function comes from the fact that every time we need to fill the halos, we are allocating arrays and permuting them to figure out what is the order of execution of the boundary conditions. This procedure is filled with type instability that leads to a lot of allocation, which eventually needs to be cleaned up. 

This information, however, is known at model construction, so it is relatively easy to store this info in the `FieldBoundaryConditions` and use it when we need it.

This PR does exactly this, avoiding the need for `permute_boundary_conditions`

### DIFF
--- a/src/BoundaryConditions/field_boundary_conditions.jl
+++ b/src/BoundaryConditions/field_boundary_conditions.jl
@@ -47,7 +47,7 @@ default_auxiliary_bc(grid, ::Val{:top}, loc)    = _default_auxiliary_bc(topology
 ##### Field boundary conditions
 #####
 
-mutable struct FieldBoundaryConditions{W, E, S, N, B, T, I}
+mutable struct FieldBoundaryConditions{W, E, S, N, B, T, I, A}
     west :: W
     east :: E
     south :: S
@@ -55,14 +55,13 @@ mutable struct FieldBoundaryConditions{W, E, S, N, B, T, I}
     bottom :: B
     top :: T
     immersed :: I
+    auxiliaries :: A # Auxiliaires used to fill halo regions
 end
 
 function FieldBoundaryConditions(indices::Tuple, west, east, south, north, bottom, top, immersed)
-    # Turn bcs in windowed dimensions into nothing
-    west, east   = window_boundary_conditions(indices[1], west, east)
-    south, north = window_boundary_conditions(indices[2], south, north)
-    bottom, top  = window_boundary_conditions(indices[3], bottom, top)
-    return FieldBoundaryConditions(west, east, south, north, bottom, top, immersed)
+    bcs = FieldBoundaryConditions(west, east, south, north, bottom, top, immersed, nothing)
+    fill_halos!, bcs = permute_boundary_conditions(bcs)
+    return FieldBoundaryConditions(west, east, south, north, bottom, top, immersed, (; fill_halos!, ordered_boundary_conditions = bcs))
 end
 
 FieldBoundaryConditions(indices::Tuple, bcs::FieldBoundaryConditions) =
@@ -83,7 +82,8 @@ on_architecture(arch, fbcs::FieldBoundaryConditions) =
                             on_architecture(arch, fbcs.north),
                             on_architecture(arch, fbcs.bottom),
                             on_architecture(arch, fbcs.top),
-                            on_architecture(arch, fbcs.immersed))
+                            on_architecture(arch, fbcs.immersed), 
+                            on_architecture(arch, fbcs.execution_auxiliaries))
 
 """
     FieldBoundaryConditions(; kwargs...)
@@ -111,14 +111,15 @@ and the topology in the boundary-normal direction is used:
  - `ImpenetrableBoundaryCondition` for `Bounded` directions and `Face`-located fields
  - `nothing` for `Flat` directions and/or `Nothing`-located fields
 """
-FieldBoundaryConditions(default_bounded_bc::BoundaryCondition = NoFluxBoundaryCondition();
+function FieldBoundaryConditions(default_bounded_bc::BoundaryCondition = NoFluxBoundaryCondition();
                         west = DefaultBoundaryCondition(default_bounded_bc),
                         east = DefaultBoundaryCondition(default_bounded_bc),
                         south = DefaultBoundaryCondition(default_bounded_bc),
                         north = DefaultBoundaryCondition(default_bounded_bc),
                         bottom = DefaultBoundaryCondition(default_bounded_bc),
                         top = DefaultBoundaryCondition(default_bounded_bc),
-                        immersed = DefaultBoundaryCondition(default_bounded_bc)) =
+                        immersed = DefaultBoundaryCondition(default_bounded_bc))
+                        
     FieldBoundaryConditions(west, east, south, north, bottom, top, immersed)
 
 """
--- a/src/BoundaryConditions/BoundaryConditions.jl
+++ b/src/BoundaryConditions/BoundaryConditions.jl
@@ -24,6 +24,7 @@ include("boundary_condition_classifications.jl")
 include("boundary_condition.jl")
 include("discrete_boundary_function.jl")
 include("continuous_boundary_function.jl")
+include("boundary_condition_ordering.jl")
 include("field_boundary_conditions.jl")
 include("show_boundary_conditions.jl")
 
--- a/src/BoundaryConditions/boundary_condition_ordering.jl
+++ b/src/BoundaryConditions/boundary_condition_ordering.jl
@@ -1,0 +1,137 @@
+
+retrieve_bc(bc) = bc
+
+# Returns the boundary conditions a specific side for `FieldBoundaryConditions` inputs and
+# a tuple of boundary conditions for `NTuple{N, <:FieldBoundaryConditions}` inputs
+for dir in (:west, :east, :south, :north, :bottom, :top)
+    extract_side_bc = Symbol(:extract_, dir, :_bc)
+    @eval begin
+        @inline $extract_side_bc(bc) = retrieve_bc(bc.$dir)
+        @inline $extract_side_bc(bc::Tuple) = map($extract_side_bc, bc)
+    end
+end
+#
+@inline extract_bc(bc, ::Val{:west})   = tuple(extract_west_bc(bc))
+@inline extract_bc(bc, ::Val{:east})   = tuple(extract_east_bc(bc))
+@inline extract_bc(bc, ::Val{:south})  = tuple(extract_south_bc(bc))
+@inline extract_bc(bc, ::Val{:north})  = tuple(extract_north_bc(bc))
+@inline extract_bc(bc, ::Val{:bottom}) = tuple(extract_bottom_bc(bc))
+@inline extract_bc(bc, ::Val{:top})    = tuple(extract_top_bc(bc))
+
+@inline extract_bc(bc, ::Val{:west_and_east})   = (extract_west_bc(bc), extract_east_bc(bc))
+@inline extract_bc(bc, ::Val{:south_and_north}) = (extract_south_bc(bc), extract_north_bc(bc))
+@inline extract_bc(bc, ::Val{:bottom_and_top})  = (extract_bottom_bc(bc), extract_top_bc(bc))
+
+# In case of a DistributedCommunication paired with a
+# Flux, Value or Gradient boundary condition, we split the direction in two single-sided
+# fill_halo! events (see issue #3342)
+# `permute_boundary_conditions` returns a 2-tuple containing the ordered operations to execute in
+# position [1] and the associated boundary conditions in position [2]
+function permute_boundary_conditions(boundary_conditions)
+
+    split_x_halo_filling = split_halo_filling(extract_west_bc(boundary_conditions),  extract_east_bc(boundary_conditions))
+    split_y_halo_filling = split_halo_filling(extract_south_bc(boundary_conditions), extract_north_bc(boundary_conditions))
+
+    west_bc  = extract_west_bc(boundary_conditions)
+    east_bc  = extract_east_bc(boundary_conditions)
+    south_bc = extract_south_bc(boundary_conditions)
+    north_bc = extract_north_bc(boundary_conditions)
+
+    if split_x_halo_filling
+        if split_y_halo_filling
+            fill_halos! = [fill_west_halo!, fill_east_halo!, fill_south_halo!, fill_north_halo!, fill_bottom_and_top_halo!]
+            sides       = [:west, :east, :south, :north, :bottom_and_top]
+            bcs_array   = [west_bc, east_bc, south_bc, north_bc, extract_bottom_bc(boundary_conditions)]
+        else
+            fill_halos! = [fill_west_halo!, fill_east_halo!, fill_south_and_north_halo!, fill_bottom_and_top_halo!]
+            sides       = [:west, :east, :south_and_north, :bottom_and_top]
+            bcs_array   = [west_bc, east_bc, south_bc, extract_bottom_bc(boundary_conditions)]
+        end
+    else
+        if split_y_halo_filling
+            fill_halos! = [fill_west_and_east_halo!, fill_south_halo!, fill_north_halo!, fill_bottom_and_top_halo!]
+            sides       = [:west_and_east, :south, :north, :bottom_and_top]
+            bcs_array   = [west_bc, south_bc, north_bc, extract_bottom_bc(boundary_conditions)]
+        else
+            fill_halos! = [fill_west_and_east_halo!, fill_south_and_north_halo!, fill_bottom_and_top_halo!]
+            sides       = [:west_and_east, :south_and_north, :bottom_and_top]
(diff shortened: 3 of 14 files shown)


simone-silvestri:
a field in this PR looks something like this
```julia
julia> c = CenterField(grid)
1×1×10 Field{Center, Center, Center} on RectilinearGrid on CPU
├── grid: 1×1×10 RectilinearGrid{Float64, Periodic, Flat, Bounded} on CPU with 1×0×3 halo
│   └── west: Periodic, east: Periodic, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 3×1×16 OffsetArray(::Array{Float64, 3}, 0:2, 1:1, -2:13) with eltype Float64 with indices 0:2×1:1×-2:13

julia> c.boundary_conditions.kernels
3-element Vector{Function}:
 fill_bottom_and_top_halo! (generic function with 5 methods)
 fill_south_and_north_halo! (generic function with 11 methods)
 fill_west_and_east_halo! (generic function with 8 methods)

julia> c.boundary_conditions.ordered_bcs
((FluxBoundaryCondition: Nothing, FluxBoundaryCondition: Nothing), (nothing, nothing), (PeriodicBoundaryCondition, PeriodicBoundaryCondition))
```

src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -47,14 +47,25 @@ default_auxiliary_bc(grid, ::Val{:top}, loc)    = _default_auxiliary_bc(topology
...
+mutable struct FieldBoundaryConditions{W, E, S, N, B, T, I, K, O}
     west :: W
     east :: E
     south :: S
     north :: N
     bottom :: B
     top :: T
     immersed :: I
+    kernels :: K # kernels used to fill halo regions
+    ordered_bcs :: O
```
glwagner, line 59: what does "ordered bcs" mean here? Shouldn't we store the ordering (eg something like a tuple that reads `(:west, :east, :top, :bottom, :south, :north)`? If we store the kernels, can't the ordering be embedded in that object?

glwagner:
If we store the "order" in which the halos are filled in `FieldBoundaryConditions`, then how does it work for tuples? At a glance it looks like this PR is _also_ storing pre-built kernels, in addition to precomputing the order of halo filling. Is that correct?

simone-silvestri:
I was storing just the names of the functions that call the kernels, but it might be a good idea to store really the whole prebuilt kernels. In theory, the tupled halo filling should have the same order of the individual fields, but indeed this approach might become a bit messy in combination with tupled halo filling. 

Probably we can we can weight the advantage of tupled halo filling vs prebuilt kernels and support only one of the two? It would require a lot of refactoring, so I guess I can mark this PR as WIP and leave it as an exploration.
I can work on adding new benchmarks that measure the actual execution time rather than the kernels and then we can add a low resolution and a high resolution so that we cover all the cases and see how we can improve the `fill_halo_regions!` function

glwagner:
> I was storing just the names of the functions that call the kernels, but it might be a good idea to store really the whole prebuilt kernels. In theory, the tupled halo filling should have the same order of the individual fields, but indeed this approach might become a bit messy in combination with tupled halo filling.
> 
> Probably we can we can weight the advantage of tupled halo filling vs prebuilt kernels and support only one of the two? It would require a lot of refactoring, so I guess I can mark this PR as WIP and leave it as an exploration. I can work on adding new benchmarks that measure the actual execution time rather than the kernels and then we can add a low resolution and a high resolution so that we cover all the cases and see how we can improve the `fill_halo_regions!` function

We may not need tupled halo filling if there is a significant speed up to single-field halo filling. So I would pursue that first and then consider using a simpler algorithm in `update_state!`. Maybe the whole concept of tupled halo filling is premature optimization.

simone-silvestri:
@jagoosw as a part of this PR, I am proposing changing the keyword argument `fill_boundary_normal_velocities` to `fill_open_bcs` since `fill_halo_regions!` does not necessarily know if the incoming data is a velocity or a tracer.
I have completely removed the `fill_open_halo_regions!` function and (hopefully) simplified the algorithm. Can you take a look to make sure I am not doing anything wrong?

src/BoundaryConditions/field_boundary_conditions.jl
simone-silvestri: Maybe we can give it another name? Basically, `ordered_bcs` holds the boundary conditions in the correct order to be used by the kernels. In practice, the redundant fields here are the `west`, `east`, `south`, `north`, etc... the "meat" is in `ordered_bcs`.

src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -66,7 +75,7 @@ function FieldBoundaryConditions(indices::Tuple, west, east, south, north, botto
 end
 
 FieldBoundaryConditions(indices::Tuple, bcs::FieldBoundaryConditions) =
-    FieldBoundaryConditions(indices, (getproperty(bcs, side) for side in propertynames(bcs))...)
+    FieldBoundaryConditions(indices, (getproperty(bcs, side) for side in propertynames(bcs)[1:7])...)
```
navidcy: what's 1:7 here? does it deserve a comment?

src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -118,7 +129,7 @@ FieldBoundaryConditions(default_bounded_bc::BoundaryCondition = NoFluxBoundaryCo
                         north = DefaultBoundaryCondition(default_bounded_bc),
                         bottom = DefaultBoundaryCondition(default_bounded_bc),
                         top = DefaultBoundaryCondition(default_bounded_bc),
-                        immersed = DefaultBoundaryCondition(default_bounded_bc)) =
+                        immersed = DefaultBoundaryCondition(default_bounded_bc)) =
```
navidcy:
```suggestion
                        immersed = DefaultBoundaryCondition(default_bounded_bc)) =
```

src/BoundaryConditions/fill_halo_kernels.jl
```diff
@@ -0,0 +1,162 @@
...
+
+"""
+    construct_boundary_conditions_kernels(bcs::FieldBoundaryConditions,
+                                          data::OffsetArray,
+                                          grid::AbstractGrid,
+                                          loc, indices)
+
+Constructs preconfigured boundary condition kernels for a given data arraym, grid, 
+and the provided `FieldBoundaryConditions` object. Returns a new `FieldBoundaryConditions` object
+with the preconfigured kernels and ordered boundary conditions.
```
navidcy:
```suggestion
Construct preconfigured boundary condition kernels for a given `data` array, `grid`, 
and the provided `bcs` (a FieldBoundaryConditions` object).
Return a new `FieldBoundaryConditions` object with the preconfigured kernels and
ordered boundary conditions.
```

src/BoundaryConditions/fill_halo_kernels.jl
```diff
@@ -0,0 +1,162 @@
...
+#####
+
+# A struct to hold the side of the fill_halo kernel
+# These are defined in `src/DistributedComputations/halo_communication.jl`
+struct DistributedFillHalo{S}
+    side :: S
+end
+
+for Side in (:WestAndEast, :SouthAndNorth, :BottomAndTop, :West, :East, :South, :North, :Bottom, :Top)
+    @eval fill_halo_kernel!(::$Side, bc::DCBC, grid, size, offset, data, reduced_dimensions) =  DistributedFillHalo($Side())
```
navidcy:
```suggestion
    @eval fill_halo_kernel!(::$Side, bc::DCBC, grid, size, offset, data, reduced_dimensions) = DistributedFillHalo($Side())
```

src/BoundaryConditions/field_boundary_conditions.jl
glwagner: Looking at `FieldBoundaryConditions`, it's the first 7 boundaries: west, east, south, north, bottom, top, immersed.

I think a better way to write this code would be to define a constant, eg

```julia
const boundarynames = (:west, :east, :south, :north, :bottom, :top, :immersed)
```

simone-silvestri:
an example of the information ordered and stored in the field boundary condition is
```julia
julia> grid = RectilinearGrid(size = (12, 1), x = (0, 1), z = (0,1), topology = (Periodic, Flat, Bounded))
12×1×1 RectilinearGrid{Float64, Periodic, Flat, Bounded} on CPU with 3×0×1 halo
├── Periodic x ∈ [0.0, 1.0) regularly spaced with Δx=0.0833333
├── Flat y
└── Bounded  z ∈ [0.0, 1.0] regularly spaced with Δz=1.0

julia> c = CenterField(grid)
12×1×1 Field{Center, Center, Center} on RectilinearGrid on CPU
├── grid: 12×1×1 RectilinearGrid{Float64, Periodic, Flat, Bounded} on CPU with 3×0×1 halo
│   └── west: Periodic, east: Periodic, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 18×1×3 OffsetArray(::Array{Float64, 3}, -2:15, 1:1, 0:2) with eltype Float64 with indices -2:15×1:1×0:2

julia> c.boundary_conditions
├── west: PeriodicBoundaryCondition
├── east: PeriodicBoundaryCondition
├── south: Nothing
├── north: Nothing
├── top: FluxBoundaryCondition: Nothing

julia> c.boundary_conditions.ordered_bcs
((FluxBoundaryCondition: Nothing, FluxBoundaryCondition: Nothing), (nothing, nothing), (PeriodicBoundaryCondition, PeriodicBoundaryCondition))

julia> c.boundary_conditions.kernels
(KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(12, 1)}, KernelAbstractions.NDIteration.StaticSize{(12, 1)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_bottom_and_top_halo!)}(KernelAbstractions.CPU(false), Oceananigans.BoundaryConditions.cpu__fill_bottom_and_top_halo!), nothing, KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(1, 3)}, Oceananigans.Utils.OffsetStaticSize{(1:1, 1:3)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_periodic_west_and_east_halo!)}(KernelAbstractions.CPU(false), Oceananigans.BoundaryConditions.cpu__fill_periodic_west_and_east_halo!))

julia> c.boundary_conditions.kernels[1]
KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(12, 1)}, KernelAbstractions.NDIteration.StaticSize{(12, 1)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_bottom_and_top_halo!)}(KernelAbstractions.CPU(false), Oceananigans.BoundaryConditions.cpu__fill_bottom_and_top_halo!)

julia> c.boundary_conditions.kernels[2]

julia> c.boundary_conditions.kernels[3]
KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(1, 3)}, Oceananigans.Utils.OffsetStaticSize{(1:1, 1:3)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_periodic_west_and_east_halo!)}(KernelAbstractions.CPU(false), Oceananigans.BoundaryConditions.cpu__fill_periodic_west_and_east_halo!)
```

glwagner:
Should `boundary_conditions.kernels` be a named tuple so we know which kernel fills which boundary?

simone-silvestri:
Ok, the new example of stored info is
```julia
julia> grid = RectilinearGrid(size = (12, 2), x = (0, 1), z = (0, 1), topology = (Periodic, Flat, Bounded));

julia> c = CenterField(grid);

julia> c.boundary_conditions.ordered_bcs
(bottom_and_top = (FluxBoundaryCondition: Nothing, FluxBoundaryCondition: Nothing), south_and_north = (nothing, nothing), west_and_east = (PeriodicBoundaryCondition, PeriodicBoundaryCondition))

julia> c.boundary_conditions.kernels
(bottom_and_top = KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(12, 1)}, KernelAbstractions.NDIteration.StaticSize{(12, 1)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_bottom_and_top_halo!)}(KernelAbstractions.CPU(false), Oceananigans.BoundaryConditions.cpu__fill_bottom_and_top_halo!), south_and_north = nothing, west_and_east = KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(1, 6)}, Oceananigans.Utils.OffsetStaticSize{(1:1, 1:6)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_periodic_west_and_east_halo!)}(KernelAbstractions.CPU(false), Oceananigans.BoundaryConditions.cpu__fill_periodic_west_and_east_halo!))
```

glwagner:
> Ok, the new example of stored info is
> 
> ```julia
> julia> grid = RectilinearGrid(size = (12, 2), x = (0, 1), z = (0, 1), topology = (Periodic, Flat, Bounded));
> 
> julia> c = CenterField(grid);
> 
> julia> c.boundary_conditions.ordered_bcs
> (bottom_and_top = (FluxBoundaryCondition: Nothing, FluxBoundaryCondition: Nothing), south_and_north = (nothing, nothing), west_and_east = (PeriodicBoundaryCondition, PeriodicBoundaryCondition))
> 
> julia> c.boundary_conditions.kernels
> (bottom_and_top = KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(12, 1)}, KernelAbstractions.NDIteration.StaticSize{(12, 1)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_bottom_and_top_halo!)}(KernelAbstractions.CPU(false), Oceananigans.BoundaryConditions.cpu__fill_bottom_and_top_halo!), south_and_north = nothing, west_and_east = KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(1, 6)}, Oceananigans.Utils.OffsetStaticSize{(1:1, 1:6)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_periodic_west_and_east_halo!)}(KernelAbstractions.CPU(false), Oceananigans.BoundaryConditions.cpu__fill_periodic_west_and_east_halo!))
> ```

looks great!

src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -87,6 +87,9 @@ window_boundary_conditions(::UnitRange,  left, right) = nothing, nothing
 window_boundary_conditions(::Base.OneTo, left, right) = nothing, nothing
 window_boundary_conditions(::Colon,      left, right) = left, right
 
+# The only thing we need
+Adapt.adapt_structure(to, fbcs::FieldBoundaryConditions) = (kernels = bcs.kernels, Adapt.adapt(to, fbcs.ordered_bcs))
```
glwagner: why do you need the `kernels` on GPU?

simone-silvestri: I was playing with making `fill_halo_regions!` possible to call with adapted input here 
https://github.com/CliMA/Oceananigans.jl/blob/743610313cf76281597a44ad6fc88024e96e8bf8/src/Fields/field.jl#L809-L827
so that we do not need to adapt the grid a bunch of times, which is, I suspect, the reason for high launch latency.
I can always only adapt the grid.

simone-silvestri:
This should be ready to merge, I think all tests pass except for the reactant ones

navidcy:
@simone-silvestri could you resolve the conflicts that came after #4687?
the doctest should fix when we merge main back here ;)

simone-silvestri:
Tests pass, I would need an approval to merge

tomchor:
Seems like tests pass :tada: 

@simone-silvestri maybe I missed this, but can we expect a significant speedup in simulations after this PR is merged?

simone-silvestri:
Yeah, generally for smaller models on GPU (for example for a one degree ocean model), and more so for models that involve a lot of fill halos. for sea ice dynamics for example we have a speed up of a factor of 5

glwagner:
it might be nice to report a benchmark for some small problem in this PR for record purposes

simone-silvestri:
On it.

simone-silvestri:
This is a hydrostatic model benchmark of a fairly small hydrostatic model (64 x 32 x 8) initialized with a divergent velocity field (the `benchmark_hydrostatic_model.jl` file revamped a bit) ran on my mac M1:

### branch main 
image

### This branch
image

In particular, allocations seem to be down quite a bit. In terms of performance, there is an improvement across the board, more for some free surfaces than the other though.

PR #4710, opened 2025-08-12
Correct the implicit Split-RK3 diffusion implementation

simone-silvestri:
Following discoveries in #4705 it seems like there is some difference in the AB2 and the RK3 timestepper for the hydrostatic free surface model regarding implicit timestepping. This PR is here to try to fix the (possibly) bug in the RK3 timestepper with implicit diffusion

This is a chance to add also dynamics tests for the hydrostatic RK3 timestepper which were lacking.

EDIT: Indeed, the implicit-explicit timestepping was not correctly implemented with RK3. The implicit timestep needs to be applied to the "Forward Euler" evolved variables before averaging, not to the averaged variables that include previous time-step information.

As part of this PR, I also remove the warning on the implicit - explicit stepping with split RK3 since it is tested

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
@@ -23,8 +23,9 @@ function split_rk3_substep!(model::HydrostaticFreeSurfaceModel, Δt, γⁿ, ζ�
     step_free_surface!(free_surface, model, timestepper, Δt)
 
     # Average free surface variables in the second stage
-    model.clock.stage == 2 && 
+    if model.clock.stage == 2 
         @apply_regionally rk3_average_free_surface!(free_surface, grid, timestepper, γⁿ, ζⁿ)
+    end
     
     return nothing
 end
@@ -86,8 +87,10 @@ function rk3_substep_velocities!(velocities, model, Δt, γⁿ, ζⁿ)
                        model.clock,
                        Δt)
 
-        launch!(architecture(grid), grid, :xyz,
+        if model.clock.stage > 1 
+            launch!(architecture(grid), grid, :xyz,
                 _split_rk3_average_field!, velocity_field, γⁿ, ζⁿ, Ψ⁻)
+        end
     end
 
     return nothing
@@ -124,8 +127,10 @@ function rk3_substep_tracers!(tracers, model, Δt, γⁿ, ζⁿ)
                        model.clock,
                        Δt)
 
-        launch!(architecture(grid), grid, :xyz,
-                _split_rk3_average_tracer_field!, c, grid, γⁿ, ζⁿ, Ψ⁻)
+        if model.clock.stage > 1 
+            launch!(architecture(grid), grid, :xyz,
+                    _split_rk3_average_tracer_field!, c, grid, γⁿ, ζⁿ, Ψ⁻)
+        end
     end
 
     return nothing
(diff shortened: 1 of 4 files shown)


src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
```diff
@@ -128,31 +135,20 @@ end
...
     i, j, k = @index(Global, NTuple)
     σᶜᶜⁿ = σⁿ(i, j, k, grid, Center(), Center(), Center())
     σᶜᶜ⁻ = σ⁻(i, j, k, grid, Center(), Center(), Center())
-    @inbounds θ[i, j, k] = (ζⁿ * Ψ⁻[i, j, k] + γⁿ * (σᶜᶜ⁻ * θ[i, j, k] + Δt * Gⁿ[i, j, k])) / σᶜᶜⁿ
-end
-
-@kernel function _split_rk3_substep_tracer_field!(θ, grid, Δt, ::Nothing, ::Nothing, Gⁿ, Ψ⁻)
-    i, j, k = @index(Global, NTuple)
-    @inbounds θ[i, j, k] = (Ψ⁻[i, j, k] + Δt * Gⁿ[i, j, k]) / σⁿ(i, j, k, grid, Center(), Center(), Center())
+    @inbounds θ[i, j, k] = (σᶜᶜ⁻ * θ[i, j, k] + Δt * Gⁿ[i, j, k]) / σᶜᶜⁿ
```
glwagner: `θ` or `c`? 🤔

simone-silvestri: we can call it c, `θ` is probably the potential temperature

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
```diff
@@ -84,6 +86,11 @@ function rk3_substep_velocities!(velocities, model, Δt, γⁿ, ζⁿ)
                        nothing,
                        model.clock,
                        Δt)
+
+        if model.clock.stage > 1 
+            launch!(architecture(grid), grid, :xyz,
+                _split_rk3_average_field!, velocity_field, γⁿ, ζⁿ, Ψ⁻)
```
navidcy:
```suggestion
            launch!(architecture(grid), grid, :xyz,
                    _split_rk3_average_field!, velocity_field, γⁿ, ζⁿ, Ψ⁻)
```

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
```diff
@@ -128,19 +140,20 @@ end
 ##### Tracer update in mutable vertical coordinates
 #####
 
-# σθ is the evolved quantity.
-# We store temporarily σθ in θ. Once σⁿ⁺¹ is known we can retrieve θⁿ⁺¹
-# Ψ⁻ is the previous tracer already scaled by the vertical coordinate scaling factor: ψ⁻ = σ * θ
-@kernel function _split_rk3_substep_tracer_field!(θ, grid, Δt, γⁿ, ζⁿ, Gⁿ, Ψ⁻)
+
+# σθ is the evolved quantity, so tracer fields need to be evolved
+# accounting for the stretching factors from the new and the previous time step.
```
navidcy:
```suggestion
# σc is the evolved quantity, so tracer fields need to be evolved
# accounting for the stretching factors from the new and the previous time step.
```

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_rk3_step.jl
glwagner: agree i think `θ` is more specific

PR #4711, opened 2025-08-13
Adding automated CPU benchmarks

simone-silvestri:
This PR is here in preparation for #4706. It adds some unit benchmarks that will run on the CPU. 
They have to be taken with a grain of salt since Tartarus might be occupied while running the benchmarks so the performance might show high variance. However it is good to have at least a pipeline for tests

### DIFF
--- a/.buildkite/pipeline-benchmarks.yml
+++ b/.buildkite/pipeline-benchmarks.yml
@@ -62,12 +62,17 @@ steps:
         rm \${OUTPUT_PREFIX}.sqlite
       done
 
+      # Unit tests (on CPU) using BenchmarkTools
+      export BENCHMARK_GROUP="unit"
+      $TARTARUS_HOME/julia-$JULIA_VERSION/bin/julia --color=yes --project --check-bounds=no test/benchmark_tests.jl
+
     artifact_paths:
       - "periodic_output.txt"
       - "bounded_output.txt"
       - "periodic_cheap_advection_output.txt"
       - "bounded_cheap_advection_output.txt"
       - "immersed_output.txt"
+      - "unit.txt"
     soft_fail:
       - exit_status: 3
 
@@ -114,3 +119,4 @@ steps:
       - "diff_periodic_cheap_advection_output.txt"
       - "diff_bounded_cheap_advection_output.txt"
       - "diff_immersed_output.txt"
+      - "diff_unit.txt"


benchmark/Manifest.toml
navidcy, line 1: do we need the Manifest? That would pin the Oceananigans version used, right?

glwagner:
Why don't we use the benchmark agent for this pipeline? That way it is guaranteed that only one benchmark runs at a time.

benchmark/Manifest.toml
simone-silvestri: yeah we probably don't need it. I can delete it. Oceananigans is added during the tests here
https://github.com/CliMA/Oceananigans.jl/blob/c7ccfd0419df223556708b7409c09d6af226957a/.buildkite/pipeline-benchmarks.yml#L40

simone-silvestri:
> Why don't we use the benchmark agent for this pipeline? That way it is guaranteed that only one benchmark runs at a time.

Right, this benchmark is run on the benchmark pipeline.

.buildkite/pipeline-benchmarks.yml
```diff
@@ -37,7 +37,7 @@ steps:
 
     command: |
       # Instantiate
-      $TARTARUS_HOME/julia-$JULIA_VERSION/bin/julia --color=yes --project --check-bounds=no -e 'using Pkg; Pkg.add("CUDA"); Pkg.instantiate()'
+      $TARTARUS_HOME/julia-$JULIA_VERSION/bin/julia --color=yes --project=benchmark --check-bounds=no -e 'using Pkg; Pkg.add(path="./"); Pkg.instantiate()'
```
giordano, line 40: With Julia v1.11 you could use [`[sources]`](https://pkgdocs.julialang.org/v1/toml-files/#The-[sources]-section) instead, but I see you're still on v1.10 for the time being

benchmark/benchmark_models_stepping.jl
```diff
@@ -52,21 +50,19 @@ end
...
+    for (model, name) in zip((:nonhydrostatic, :hydrostatic, :shallowwater), ("NonhydrostaticModel", "HydrostaticFreeSurfaceModel", "ShallowWaterModel"))
+        Architectures = has_cuda() ? [CPU, GPU] : [CPU]
+        Float_types = [Float64]
+        Ns = [32, 64, 128, 256]
 
-# Run and summarize benchmarks
+        benchmark_func = Symbol(:benchmark_, model, :_model)
+        @eval begin
+            suite = run_benchmarks($benchmark_func; Architectures=$Architectures, Float_types=$Float_types, Ns=$Ns)
+        end
```
giordano: This won't work in newer versions of julia, with `@eval` you're actually defining `suite` in the global scope.  What you _can_ do is
```suggestion
        suite = @eval run_benchmarks($benchmark_func; Architectures=$Architectures, Float_types=$Float_types, Ns=$Ns)
```

benchmark/benchmark_models_stepping.jl
```diff
@@ -52,21 +50,19 @@ end
...
-    @eval begin
-        suite = run_benchmarks($benchmark_func; Architectures, Float_types, Ns)
+        push!(df, benchmarks_dataframe(suite))
     end
-
-    df = benchmarks_dataframe(suite)
-    benchmarks_pretty_table(df, title=name * " benchmarks")
-end
+    return df
+end
```
giordano: Missing newline
```suggestion
end

```

simone-silvestri:
I guess this PR is being superceded by #4882

PR #4722, opened 2025-08-23
Regularize auxiliary boundary conditions

glwagner:
This PR regularizes auxiliary boundary conditions in the FieldBoundaryCondition constructor. It was made feasible by support for location-based regularization. I believe this provides a simpler solution to the annoying warning that #3961 tries to solve.

cc @simone-silvestri, @bgroenks96


simone-silvestri (approved):
Maybe we can add something like
```julia
hasproperty(:immersed_boundary, grid) ? NoFluxBoundaryCondition() : nothing
```
to line 
https://github.com/CliMA/Oceananigans.jl/blob/d76dd0b0c3145ccac3620427bdfca9b9d18b00e1/src/BoundaryConditions/field_boundary_conditions.jl#L189

glwagner:
> Maybe we can add something like
> 
> ```julia
> hasproperty(:immersed_boundary, grid) ? NoFluxBoundaryCondition() : nothing
> ```
> 
> to line
> 
> https://github.com/CliMA/Oceananigans.jl/blob/d76dd0b0c3145ccac3620427bdfca9b9d18b00e1/src/BoundaryConditions/field_boundary_conditions.jl#L189

To elaborate: (I think) that you are saying it is clearer to omit the immersed boundary condition when there is no immersed boundary. This is consistent with how we treat `Flat` directions I think.

One note, we should not use `hasproperty`, that's an example of bad design which we are trying really hard to avoid due to the complexity of the code base. We have the `AbstractUnderlyingGrid` type which I think covers this case.

glwagner:
@simone-silvestri take a look at what I did and let me know what you think.

simone-silvestri:
Looks good!

navidcy:
I'll merge if all CI passes (I'll ignore Reactant-related fails)

PR #4723, opened 2025-08-23
Implement MixedBoundaryCondition (aka RobinBoundaryCondition)

### LINKED ISSUE #4490: `RobinBoundaryCondition` to complement `Gradient` and `Value` boundary conditions

glwagner:
The implementation requires some refactoring of how boundary conditions are regularized to support the fact that `CombinationCondition` has two properties, rather than one as holds for all other conditions.

See #4490 for details about how the properties of `CombinationCondition` are defined.

Partially overlaps with #4722, but I can deal with conflicts as needed.

Closes #4490.

cc @joernc @shriyafruitwala


glwagner:
CI is down right now so we can't merge, but @joernc @shriyafruitwala if you can give this PR a look through (might be good to check my math too) that will be helpful.

francispoulin:
Great to see this introduction!

Any chance we could call if `MixedBoundaryConditions`?

Just a suggestion.

glwagner:
> Great to see this introduction!
> 
> Any chance we could call if `MixedBoundaryConditions`?
> 
> Just a suggestion.

That's a good term, easy change to make.

src/BoundaryConditions/boundary_condition.jl
```diff
@@ -102,6 +110,41 @@ MultiRegionCommunicationBoundaryCondition(val; kwargs...) = BoundaryCondition(Mu
...
+when using function `coefficient` and/or `combination`.
+"""
+function MixedBoundaryCondition(coefficient, combination=0;
+                                parameters = nothing,
+                                discrete_form = false,
+                                field_dependencies = ())
+
+    coefficient = materialize_condition(coefficient, parameters, discrete_form, field_dependencies)
+    combination = materialize_condition(combination, parameters, discrete_form, field_dependencies)
+    condition = MixedCondition(coefficient, combination)
```
simone-silvestri: I am not really sure I understand the formulation from the word `combination`.

glwagner: The boundary condition may be written

```math
\partial_z c + a c = b
```

we're defining ``a`` as "coefficient" and ``b`` as "combination". What do you think would be a better term for ``b``?

simone-silvestri: is `linear_coefficient` for `a` and `right_hand_side` for `b` too lengthy?

simone-silvestri: or maybe `linear_coefficient` and `constant_coefficient`?

francispoulin: b in the inhomogeneous part of the equation.  Can we call it inhomogeneity?  Not a pretty name but it iis precise.

glwagner: well, it's not a "coefficient" (a coefficient multiplies something). "right hand side" or "inhomogeneity" can work

navidcy:
is this ready to merge?

PR #4727, opened 2025-08-26
Indexing for a `CallableCoordinate`

simone-silvestri:
I think it is pretty useful to be able to index into a coordinate without needing to build a grid first to do some preliminary check (for example, what is the surface layer height? Or the bottom layer?)

This PR introduces indexing for `CallableCoordinate`s so they can be used and inspected as arrays.
This PR was motivated by this change https://github.com/PRONTOLab/GB-25/pull/234/commits/e745054d887d4f5c2d7bb7697a38219db9000a83 which demonstrates that it would be nice to be able to index coordinates.


simone-silvestri:
I am thinking that maybe we can also make `CallableCoordinate` show the actual coordinate rather than just the constructed struct?

I mean I get very little info from 
```julia
julia> z
ExponentialCoordinate
├─ size: 60
├─ left: -6000.0
├─ right: 0.0
├─ scale: 1200.0
└─ bias: :right
```

while if I add the actual array it might be a bit verbose but I am showing the info that I really care about
```julia
julia> z
ExponentialCoordinate
├─ size: 60
├─ left: -6000.0
├─ right: 0.0
├─ scale: 1200.0
├─ faces: [-6000.0, -5517.012141182025, -5072.641859342764, -4663.8014635092995, -4287.650140847878, -3941.574217377802, -3623.168996951477, -3330.222052309417, -3060.6978521087617, -2812.7236171067166  …  -45.46405585821886, -38.57460406483252, -32.236002422469454, -26.40420738485304, -21.03869693323106, -16.102189010581, -11.560382468573408, -7.3817187272724025, -3.5371624914743416, 0.0]
└─ bias: :right
```

src/Grids/coordinate_utils.jl
```diff
@@ -132,11 +136,13 @@ Base.length(coord::ExponentialCoordinate) = coord.size
 Base.summary(::ExponentialCoordinate) = "ExponentialCoordinate"
 
 function Base.show(io::IO, coord::ExponentialCoordinate)
+    array = [coord(i) for i in 1:coord.size+1]
```
glwagner: any reason not to add this as a property?

src/Grids/coordinate_utils.jl
```diff
@@ -1,5 +1,9 @@
 abstract type CallableCoordinate end
 
+# Callable coordinates can be indexed just like arrays!
+Base.getindex(coord::CallableCoordinate, i) = coord(i)
+
+
```
glwagner:
```suggestion
```

glwagner (approved):
seems legit

glwagner:
PS i suggest putting the faces property as second from top, after size.

src/Grids/coordinate_utils.jl
```diff
@@ -135,13 +147,12 @@ Base.length(coord::ExponentialCoordinate) = coord.size
...
 
 function Base.show(io::IO, coord::ExponentialCoordinate)
-    array = [coord(i) for i in 1:coord.size+1]
     return print(io, summary(coord), '\n',
                  "├─ size: ", coord.size, '\n',
                  "├─ left: ", coord.left, '\n',
                  "├─ right: ", coord.right, '\n',
                  "├─ scale: ", coord.scale, '\n',
-                 "├─ faces: ", array, '\n',
+                 "├─ faces: ", coord.faces, '\n',
```
glwagner: suggest moving this up underneath `size` !

src/Grids/coordinate_utils.jl
```diff
@@ -1,13 +1,24 @@
...
     scale :: Float64
     bias :: Symbol
+    function ExponentialCoordinate(size::Int, left::Number, right::Number, scale::Number, bias::Symbol)
+        faces = [construct_exponential_coordinate(i, size, left, right, scale, bias) for i in 1:size]
+        return new(size, faces, left, right, scale, bias)
+    end
 end
 
+# An exponential coordinate actually has faces
+Base.getindex(coord::ExponentialCoordinate, i) = coord.faces[i]
```
navidcy:
```suggestion
Base.getindex(coord::ExponentialCoordinate, i) = @inbounds coord.faces[i]
```
?

navidcy (approved):

src/Grids/coordinate_utils.jl
glwagner: doesn't matter

glwagner: its better practice to use `Base.@propagate_inbounds` though (on the outside) if you need it.

src/Grids/coordinate_utils.jl
```diff
@@ -46,6 +57,7 @@ x = ExponentialCoordinate(N, l, r)
 # output
 ExponentialCoordinate
 ├─ size: 10
+├─ faces: [-1000.0, -564.247649441104, -299.95048878528615, -139.64615757253702, -42.41666580727582, 16.55600197663209, 52.324733072619736, 74.0195651413529, 87.17814594835643, 95.15922864611028]
```
navidcy: Might need to update the doctest

PR #4731, opened 2025-08-28
Get output with ConformalCubedSphereGrid working again

glwagner:
Plus some other misc bug fixes. This is a WIP. cc @siddharthabishnu @simone-silvestri @navidcy

### DIFF
--- a/src/Grids/Grids.jl
+++ b/src/Grids/Grids.jl
@@ -107,9 +107,7 @@ Grid topology for dimensions that are connected to other models or domains only 
 struct RightConnected <: AbstractTopology end
 
 topology_str(T) = string(T)
-topology_str(::Type{RightConnected}) = "RightConnected"
-topology_str(::Type{LeftConnected}) = "LeftConnected"
-topology_str(::Type{FullyConnected}) = "FullyConnected"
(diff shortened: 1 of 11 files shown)


src/Grids/Grids.jl
```diff
@@ -106,6 +106,11 @@ Grid topology for dimensions that are connected to other models or domains only
 """
 struct RightConnected <: AbstractTopology end
 
+topology_str(T) = string(T)
+topology_str(::Type{RightConnected}) = "RightConnected"
+topology_str(::Type{LeftConnected}) = "LeftConnected"
+topology_str(::Type{FullyConnected}) = "FullyConnected"
```
navidcy:
```suggestion
topology_str(::Type{F}) where F = string(F)
```

glwagner: interesting. I don't think this is different, is it?

navidcy: no, it's just 2 lines less code, right?

glwagner: oh i missed that you propose deleting lines

glwagner: this may indeed have the intended effect

glwagner:
closing this, will restart

PR #4732, opened 2025-08-28
Add visualization functions for numerical solutions on the cubed sphere

siddharthabishnu:
This PR introduces functionality to visualize numerical solutions on the cubed sphere. It provides tools for easy visualization across the six panels (faces) of the cubed sphere as well as on the entire spherical surface, enabling clear interpretation of results and comparison with solutions on other grids such as latitude–longitude and tripolar grids.


glwagner:
Can you change the code structure so that there is only one top level module `OceananigansMakieExt` and no submodules? I'd prefer that if possible

glwagner:
Once you're ready it might be further helpful to see some example usage here. In a subsequent PR, we can add a simple example in the docs that showcases a simple cubed sphere simulation.

PR #4735, opened 2025-08-28
Define `keys` for `FieldDataset`

erin-atkinson:
`FieldDataset` ought to have `keys` return the keys of the underlying dictionary.

### DIFF
--- a/test/test_output_readers.jl
+++ b/test/test_output_readers.jl
@@ -495,9 +495,10 @@ end
             ds = FieldDataset(filepath3d, backend=Backend())
 
             @test ds isa FieldDataset
-            @test length(keys(ds.fields)) == 8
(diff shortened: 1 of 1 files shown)


glwagner (approved):
consider adding a test?

erin-atkinson:
I added a check for the correct elements in `keys`. I also changed the length check at the start to the new method. Is returning strings, rather than symbols, appropriate?

glwagner:
> I added a check for the correct elements in `keys`. I also changed the length check at the start to the new method. Is returning strings, rather than symbols, appropriate?

looks great!

erin-atkinson:
This can be merged?

simone-silvestri:
yeah, the test failures  are not related to this PR

PR #4736, opened 2025-08-28
Temporary fix for allocating reductions hitting CUDA bug

glwagner:
This PR puts in a temporary fix that (I think) will fix allocating reductions (but does not address non-allocating reductions) that encounter a CUDA bug in configuring kernels for map reductions.

We don't have to merge but we can consider it, if a fix is not forthcoming.

cc @tomchor @ali-ramadhan


ali-ramadhan:
Nice! It'll be good to be able to use this branch if we're trying to run a simulation that refuses to run because of the kernel limit bug.

glwagner:
Closing this bc it was resolved upstream

tomchor:
I'd keep it open for a little while. It seems the fix causes a big performance regression: https://github.com/JuliaGPU/CUDA.jl/pull/2869#issuecomment-3259241572, plus the fix hasn't been backported yet.

Currently there's a potential fix that doesn't cause a performance degradation in https://github.com/JuliaGPU/CUDA.jl/pull/2880, but who knows what will happen there.

glwagner:
Ok. Maybe we can save CI resources though unless its needed to merge main here.

navidcy:
is this resolved by #4824?

glwagner:
seems like

PR #4737, opened 2025-08-29
A new formulation for the split RK3 timestepper

simone-silvestri:
Recently, I have been trying to adapt zstar to RK3 by making sure that 
- tracer is conserved locally (an originally constant tracer field remains constant in the absence of tracer forcing)
- tracer is conserved globally (the integral tracer content remains constant in the absence of tracer forcing)
- baroclinic and barotropic steps yield the same estimate for the free surface

This has been proving quite complicated to do, especially with our SKR3 that requires different treatment for each substep making the whole development quite cumbersome. In contrast, in recent NEMO development, they implemented a simpler RK3 formulation that is second-order accurate and consists of 3 separate Euler steps with increasing dt, so that the whole timestep can be summarized very simply with

$$y^1 = y^n + \frac{\Delta t}{3}G(y^n)$$
$$y^2 = y^n + \frac{\Delta t}{2}G(y^1)$$
$$y^3 = y^n + \Delta t G(y^2)$$

This is the same RK3 used in https://doi.org/10.1175/MWR-D-13-00068.1 and in https://doi.org/10.1016/j.ocemod.2023.102273 for the evolution of the slow variables.

This PR switches to the new RK3 formulation. To check that the results maintain the same quality, I have 3 different test cases in the validation folder:

## One-dimensional tracer advection (checking tracer evolution)

a one-dimensional tracer profile advected by a prescribed velocity of `u=1` using WENO 7

https://github.com/user-attachments/assets/467cb854-d795-4ad5-8cc5-7d4643de75ac

## CATKE diffusion (checking implicit diffusion within RK3)

single-column CATKE simulation forced by a suite of different forcings. AB2 uses a CATKE formulation that substeps the TKE with a time step of 5 minutes. SRK3 does not substep TKE

image

## Geostrophic adjustment (checking the barotropic evolution)

This is a one-layer model, so the solution is dominated by the barotropic mode.
The video shows a visualization of the free surface evolution; both timesteppers should give the same exact answer, given that the case is barotropic and the barotropic timestepper is forward-backward in both cases

https://github.com/user-attachments/assets/1962e1b4-ee6d-49ff-8e31-69824ade1024

## Lock-exchange test (check tracer and volume conservation)

the output of the `validation/z_star_coordinate/lock_release.jl` script using both SRK3 and AB2. The integral conservation plots track the integral change in volume, buoyancy, and a passive tracer initialized with a value of 1 everywhere in the domain.

image

As a note, it looks like tracer consistency is not perfectly satisfied as the tracer content grows in time in both SRK3 and AB2, I suspect precision errors that propagate. However, the important thing is that the behavior is consistent between AB2 and SRK3 and a better formulation of a split-zstar method will come in a different PR.


simone-silvestri:
Note that the implementation is much simpler, if I hadn't reworked the validation examples, this PR would result is a significant reduction in code lines.

navidcy:
So what is the improvement compared to eg `main`. Meaning, which of the 

> - tracer is conserved locally (an originally constant tracer field remains constant in the absence of tracer forcing)
> - tracer is conserved globally (the integral tracer content remains constant in the absence of tracer forcing)
> - baroclinic and barotropic steps yield the same estimate for the free surface

wasn't satisfied previously and now it does.

src/TimeSteppers/split_hydrostatic_runge_kutta_3.jl
```diff
@@ -22,55 +19,51 @@ end
...
 ```julia
-Uᵐ⁺¹ = ζᵐ * Uⁿ + γᵐ * (Uᵐ + Δt * Gᵐ)
+Δt̃   = Δt / βᵐ
+Uᵐ⁺¹ = Uⁿ + Δt̃ * Gᵐ
 ```
 
 where `Uᵐ` is the state at the ``m``-th substep, `Uⁿ` is the state at the ``n``-th timestep,
-`Gᵐ` is the tendency at the ``m``-th substep, and constants `γ¹ = 1`, `γ² = 1/4`, `γ³ = 1/3`,
-`ζ¹ = 0`, `ζ² = 3/4`, and `ζ³ = 1/3`.
+`Gᵐ` is the tendency at the ``m``-th substep, and constants `β¹ = 1/3`, `β² = 1/2`, `β³ = 1`.
```
navidcy: the docstring talks about `β¹ = 1/3`, `β² = 1/2`, `β³ = 1` but the code below writes `β¹ = 3`, `β² = 2`.

navidcy: I think you simply reported the inverse here in the docstring.

But on another note, reading https://www.sciencedirect.com/science/article/pii/S1463500323001130, if I look at their equations between (1) and (2) I'd say that it should be the other way around, meaning:

 `β¹ = 1`, `β² = 2`, `β³ = 3`

while your implementation has the first substep with 3, then 2, then 1. Am I missing anything?

simone-silvestri:
> So what is the improvement compared to eg `main`. Meaning, which of the
> 
> > * tracer is conserved locally (an originally constant tracer field remains constant in the absence of tracer forcing)
> > * tracer is conserved globally (the integral tracer content remains constant in the absence of tracer forcing)
> > * baroclinic and barotropic steps yield the same estimate for the free surface
> 
> wasn't satisfied previously and now it does.

Condition (1) and (2) was satisfied before and still are in the same way. Condition (3) is a bit more tricky but very difficult to achieve with the previous formulation and (I hope) much easier to do with this one that is basically just a combination of Euler steps.

src/TimeSteppers/split_hydrostatic_runge_kutta_3.jl
simone-silvestri: right, docstrings are wrong. However, the correct order is `3`, `2`, `1`

navidcy: Actually I think I was also wrong.

From this:

Screenshot 2025-09-01 at 9 00 27 am

I understand that 

 `β¹ = 1`, `β² = 3`, `β³ =2`

?

simone-silvestri: Not really, you see from there that

$$y_1 = y_n + \frac{\Delta t}{3}k_1$$
$$y_2 = y_n + \frac{\Delta t}{2}k_2$$
$$y_3 = y_n + \frac{\Delta t}{1}k_3$$

navidcy: OK! My bad! I was reading it the other way around!

simone-silvestri:
Ready to go when tests pass. It will change anyways in the future, so it is good to get this merged quickly as it is blocking further work.

PR #4738, opened 2025-08-29
Remove support for MultiRegionObject split on different devices

simone-silvestri:
At the moment, the `MultiRegion` module is used only for the cubed sphere, which does not support running on multiple devices anyway, so this PR removes the ability to split multi-region variables across different devices.

This can be reimplemented using MPI in the future.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -16,7 +16,7 @@ StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 TimesDates = "bdfc003b-8df8-5c39-adcd-3a9087f5df4a"
 
 [compat]
-CUDA = "< 5.8.3"
+CUDA = "5.4 - 5.8.2"
 CairoMakie = "0.11, 0.12, 0.13, 0.14"
 Documenter = "1"
 DocumenterCitations = "1"


glwagner:
@siddharthabishnu can you take a look at this PR? This substantial changes may impact other PRs you are working on, so we may want to merge this one ASAP.

glwagner:
@michel2323 can you also take a look

siddharthabishnu:
> @siddharthabishnu can you take a look at this PR? This substantial changes may impact other PRs you are working on, so we may want to merge this one ASAP.

I just checked it out, and it looks good to me. This PR removes functionalities that the cubed sphere doesn’t use anyway, so we should be good on that front. Besides, if all tests pass in `test/test_multi_region_cubed_sphere.jl`, that’s further confirmation.

michel2323:
I think for launching the kernels in `multi_region_transformation.jl`, you need to save the architecture or backend in MRO, or use `get_backend()` on an array. Either way, LGTM and much simplified.

glwagner:
> I think for launching the kernels in `multi_region_transformation.jl`, you need to save the architecture or backend in MRO, or use `get_backend()` on an array. Either way, LGTM and much simplified.

I think @simone-silvestri may be assuming that the backend can be inferred from child objects.

simone-silvestri:
Yeah typically a kernel will be called from inside a `regional_function`.

simone-silvestri:
Ok, I guess I can merge this PR then since all tests pass.

siddharthabishnu (approved):

PR #3968, opened 2024-12-02
Implement an implicit free surface solver in the NonhydrostaticModel

glwagner:

PR #4740, opened 2025-08-29
[WIP] implicit free surface formulation for NonhydrostaticModel

glwagner:
This PR adds an implicit free surface solver for `NonhydrostaticModel`. This will be useful for large eddy simulations of internal tide generation, breaking, and turbulent mixing. Work with @shriyafruitwala

### DIFF
--- a/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
+++ b/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
@@ -10,7 +10,7 @@ using Oceananigans.Fields: Field, tracernames, VelocityFields, TracerFields, Cen
 using Oceananigans.Forcings: model_forcing
 using Oceananigans.Grids: inflate_halo_size, with_halo, architecture
 using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid
-using Oceananigans.Models: AbstractModel, extract_boundary_conditions
+using Oceananigans.Models: AbstractModel, extract_boundary_conditions, materialize_free_surface
 using Oceananigans.Solvers: FFTBasedPoissonSolver
 using Oceananigans.TimeSteppers: Clock, TimeStepper, update_state!, AbstractLagrangianParticles
 using Oceananigans.TurbulenceClosures: validate_closure, with_tracers, build_diffusivity_fields, time_discretization, implicit_diffusion_solver
@@ -29,7 +29,7 @@ const BFOrNamedTuple = Union{BackgroundFields, NamedTuple}
 # but for now we use it only for hydrostatic pressure anomalies for now.
 struct DefaultHydrostaticPressureAnomaly end
 
-mutable struct NonhydrostaticModel{TS, E, A<:AbstractArchitecture, G, T, B, R, SD, U, C, Φ, F,
+mutable struct NonhydrostaticModel{TS, E, A<:AbstractArchitecture, G, T, B, R, SD, U, C, Φ, F, FS,
                                    V, S, K, BG, P, BGC, AF, BM} <: AbstractModel{TS, A}
 
          architecture :: A        # Computer `Architecture` on which `Model` is run
@@ -41,6 +41,7 @@ mutable struct NonhydrostaticModel{TS, E, A<:AbstractArchitecture, G, T, B, R, S
          stokes_drift :: SD       # Set of parameters for surfaces waves via the Craik-Leibovich approximation
               forcing :: F        # Container for forcing functions defined by the user
               closure :: E        # Diffusive 'turbulence closure' for all model fields
+         free_surface :: FS       # Diffusive 'turbulence closure' for all model fields
     background_fields :: BG       # Background velocity and tracer fields
             particles :: P        # Particle set for Lagrangian tracking
       biogeochemistry :: BGC      # Biogeochemistry for Oceananigans tracers
@@ -120,6 +121,7 @@ function NonhydrostaticModel(; grid,
                              stokes_drift = nothing,
                              forcing::NamedTuple = NamedTuple(),
                              closure = nothing,
+                             free_surface = nothing,
                              boundary_conditions::NamedTuple = NamedTuple(),
                              tracers = (),
                              timestepper = :RungeKutta3,
@@ -207,6 +209,11 @@ function NonhydrostaticModel(; grid,
     # Ensure `closure` describes all tracers
     closure = with_tracers(tracernames(tracers), closure)
 
+    # TODO: limit free surface to `nothing` (rigid lid) or ImplicitFreeSurface
+    if !isnothing(free_surface)
+        free_surface = materialize_free_surface(free_surface, velocities, grid)
+    end
+
     # Either check grid-correctness, or construct tuples of fields
     velocities         = VelocityFields(velocities, grid, boundary_conditions)
     tracers            = TracerFields(tracers,      grid, boundary_conditions)
@@ -235,7 +242,7 @@ function NonhydrostaticModel(; grid,
     !isnothing(particles) && arch isa Distributed && error("LagrangianParticles are not supported on Distributed architectures.")
 
     model = NonhydrostaticModel(arch, grid, clock, advection, buoyancy, coriolis, stokes_drift,
-                                forcing, closure, background_fields, particles, biogeochemistry, velocities, tracers,
+                                forcing, closure, free_surface, background_fields, particles, biogeochemistry, velocities, tracers,
                                 pressures, diffusivity_fields, timestepper, pressure_solver, auxiliary_fields, boundary_mass_fluxes)
 
     update_state!(model; compute_tendencies = false)
--- a/src/Models/NonhydrostaticModels/pressure_correction.jl
+++ b/src/Models/NonhydrostaticModels/pressure_correction.jl
@@ -13,7 +13,16 @@ function compute_pressure_correction!(model::NonhydrostaticModel, Δt)
 
     enforce_open_boundary_mass_conservation!(model, model.boundary_mass_fluxes)
 
-    solve_for_pressure!(model.pressures.pNHS, model.pressure_solver, Δt, model.velocities)
+    if model.free_surface isa ImplicitFreeSurface
+        g = model.free_surface.gravitational_acceleration
+        η = model.free_surface.η
+    else
+        g = nothing
+        η = nothing
+    end
+    
+    solve_for_pressure!(model.pressures.pNHS, model.pressure_solver, model.velocities, Δt, g, η)
+
     fill_halo_regions!(model.pressures.pNHS)
 
     return nothing
--- a/src/Models/NonhydrostaticModels/solve_for_pressure.jl
+++ b/src/Models/NonhydrostaticModels/solve_for_pressure.jl
@@ -58,16 +58,21 @@ function compute_source_term!(solver::DistributedFourierTridiagonalPoissonSolver
     return nothing
 end
 
-function compute_source_term!(solver::FourierTridiagonalPoissonSolver, Ũ)
+add_inhomogeneous_boundary_terms!(rhs, solver, grid, Ũ, Δt, g, η) = nothing
+
+function compute_source_term!(solver::FourierTridiagonalPoissonSolver, Ũ, Δt, g, η)
     rhs = solver.source_term
     arch = architecture(solver)
     grid = solver.grid
     tdir = solver.batched_tridiagonal_solver.tridiagonal_direction
     launch!(arch, grid, :xyz, _fourier_tridiagonal_source_term!, rhs, tdir, grid, Ũ)
+
+    add_inhomogeneous_boundary_terms!(rhs, solver, grid, Ũ, Δt, g, η)
+
     return nothing
 end
 
-function compute_source_term!(solver::FFTBasedPoissonSolver, Ũ)
+function compute_source_term!(solver::FFTBasedPoissonSolver, Ũ, Δt, g, η)
     rhs = solver.storage
     arch = architecture(solver)
     grid = solver.grid
@@ -80,13 +85,17 @@ end
 #####
 
 # Note that Δt is unused here.
-function solve_for_pressure!(pressure, solver, Δt, args...)
-    compute_source_term!(solver, args...)
+function solve_for_pressure!(pressure, solver, Ũ, Δt, g, η)
+    compute_source_term!(solver, Ũ, Δt, g, η)
+    # update_fourier_tridiagonal_solver!(solver, Ũ, Δt, g, η)
+        #=
+        D[i, j, Nz] = -(-1 / Δzᵃᵃᶠ(i, j, Nz, grid) *((-3 / (2*g*Δt^2) - 1 / Δzᵃᵃᶠ(i, j, Nz, grid))/(1 / Δzᵃᵃᶠ(i, j, Nz, grid) + 1 / (2*g*Δt^2)))) - Δzᵃᵃᶜ(i, j, Nz, grid) * (λx[i] + λy[j])
+        =#
     solve!(pressure, solver)
     return pressure
 end
 
-function solve_for_pressure!(pressure, solver::ConjugateGradientPoissonSolver, Δt, args...)
+function solve_for_pressure!(pressure, solver::ConjugateGradientPoissonSolver, Ũ, Δt, g, η)
     ϵ = eps(eltype(pressure))
     Δt⁺ = max(ϵ, Δt)
     Δt★ = Δt⁺ * isfinite(Δt)
@@ -95,7 +104,21 @@ function solve_for_pressure!(pressure, solver::ConjugateGradientPoissonSolver, �
     rhs = solver.right_hand_side
     grid = solver.grid
     arch = architecture(grid)
-    launch!(arch, grid, :xyz, _compute_source_term!, rhs, grid, args...)
+    launch!(arch, grid, :xyz, _compute_source_term!, rhs, grid, Ũ)
+
     return solve!(pressure, solver.conjugate_gradient_solver, rhs)
 end
 
+# TODO: write a function to add the inhomogeneous boundary contributions to `rhs`
+# for a non-hydrostatic implicit free surface
+#=
+function add_inhomogeneous_boundary_terms!(rhs, solver, grid, Ũ, Δt, g, η)
+    launch!(arch, grid, :xy, _add_implicit_free_surface_source_term!, rhs, grid, Ũ, Δt, g, η)
+end
+
+function _add_implicit_free_surface_source_term(rhs, solver, grid, Ũ, Δt, g, η)
+    # modifies rhs of pressure solve surface boundary condition to allow for free surface
+    if k == grid.Nz && active
+        source_term -= ((η[i,j] + Δt * w[i, j, k+1])/(Δt^2 + Δzᶜᶜᶜ(i, j, k, grid) / (2*g))) * Δt
+    end
+=#
--- a/src/Solvers/fourier_tridiagonal_poisson_solver.jl
+++ b/src/Solvers/fourier_tridiagonal_poisson_solver.jl
@@ -3,8 +3,9 @@ using Oceananigans.Grids: XYRegularRG, XZRegularRG, YZRegularRG, XYZRegularRG, s
 
 import Oceananigans.Architectures: architecture
 
-struct FourierTridiagonalPoissonSolver{G, B, R, S, β, T}
+struct FourierTridiagonalPoissonSolver{G, F, B, R, S, β, T}
     grid :: G
+    tridiagonal_formulation :: F
     batched_tridiagonal_solver :: B
     source_term :: R
     storage :: S
@@ -34,6 +35,10 @@ main_diagonal_launch_configuration(::ZDirection) = :xy
 
 extent(grid) = (grid.Lx, grid.Ly, grid.Lz)
 
+struct InhomogeneousFormulation{D}
+    direction :: D
+end
+
 struct HomogeneousNeumannFormulation{D}
     direction :: D
 end
@@ -43,6 +48,10 @@ tridiagonal_direction(formulation::HomogeneousNeumannFormulation) = formulation.
 const HomogeneousXFormulation = HomogeneousNeumannFormulation{<:XDirection}
 const HomogeneousYFormulation = HomogeneousNeumannFormulation{<:YDirection}
 const HomogeneousZFormulation = HomogeneousNeumannFormulation{<:ZDirection}
+
+const InhomogeneousXFormulation = InhomogeneousFormulation{<:XDirection}
+const InhomogeneousYFormulation = InhomogeneousFormulation{<:YDirection}
+const InhomogeneousZFormulation = InhomogeneousFormulation{<:ZDirection}
 
 """
     FourierTridiagonalPoissonSolver(grid, planner_flag = FFTW.PATIENT; tridiagonal_formulation=nothing)
@@ -130,7 +139,7 @@ function FourierTridiagonalPoissonSolver(grid, planner_flag=FFTW.PATIENT; tridia
     CT = complex(eltype(grid))
     rhs = on_architecture(arch, zeros(CT, size(grid)...))
 
-    return FourierTridiagonalPoissonSolver(grid, btsolver, rhs, sol_storage, buffer, transforms)
+    return FourierTridiagonalPoissonSolver(grid, tridiagonal_formulation, btsolver, rhs, sol_storage, buffer, transforms)
 end
 
 #####
@@ -147,7 +156,11 @@ function compute_main_diagonal!(main_diagonal, tridiagonal_formulation, grid, λ
     return nothing
 end
 
-@kernel function _compute_main_diagonal!(D, grid, λy, λz, ::HomogeneousXFormulation)
+const XFormulation = Union{HomogeneousXFormulation, InhomogeneousXFormulation}
+const YFormulation = Union{HomogeneousYFormulation, InhomogeneousYFormulation}
+const ZFormulation = Union{HomogeneousZFormulation, InhomogeneousZFormulation}
+
+@kernel function _compute_main_diagonal!(D, grid, λy, λz, ::XFormulation)
     j, k = @index(Global, NTuple)
     Nx = size(grid, 1)
 
@@ -162,7 +175,7 @@ end
     end
 end
 
-@kernel function _compute_main_diagonal!(D, grid, λx, λz, ::HomogeneousYFormulation)
+@kernel function _compute_main_diagonal!(D, grid, λx, λz, ::YFormulation)
     i, k = @index(Global, NTuple)
     Ny = size(grid, 2)
 
@@ -177,7 +190,7 @@ end
     end
 end
 
-@kernel function _compute_main_diagonal!(D, grid, λx, λy, ::HomogeneousZFormulation)
+@kernel function _compute_main_diagonal!(D, grid, λx, λy, ::ZFormulation)
     i, j = @index(Global, NTuple)
     Nz = size(grid, 3)
 
@@ -230,7 +243,9 @@ function solve!(x, solver::FourierTridiagonalPoissonSolver, b=nothing)
     # Solutions to Poisson's equation are only unique up to a constant (the global mean
     # of the solution), so we need to pick a constant. We choose the constant to be zero
     # so that the solution has zero-mean.
-    ϕ .= ϕ .- mean(ϕ)
+    if solver.tridiagonal_formulation isa HomogeneousNeumannFormulation
+        ϕ .= ϕ .- mean(ϕ)
+    end
 
     arch = architecture(solver)
     launch!(arch, solver.grid, :xyz, copy_real_component!, x, ϕ, indices(x))
--- a/src/BoundaryConditions/boundary_condition.jl
+++ b/src/BoundaryConditions/boundary_condition.jl
@@ -86,7 +86,7 @@ const PBC  = BoundaryCondition{<:Periodic}
 const OBC  = BoundaryCondition{<:Open}
 const VBC  = BoundaryCondition{<:Value}
 const GBC  = BoundaryCondition{<:Gradient}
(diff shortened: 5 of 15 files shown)


shriyafruitwala:
When I run my deep water wave example, it looks like it's back to being diffusive. In the pressure correction step, does `make_pressure_correction!` ever get called? It also seems like w is not getting updated at Nz+1. 

https://github.com/user-attachments/assets/55652533-af63-4d6e-9643-c84781610542

shriyafruitwala:
@glwagner Got it working again! There were a couple of typos and some larger fixes needed that I have detailed below:

- The “coefficient” in `set_top_pressure_boundary_condition!` Should be $1/(g \Delta t^2)$ instead of $1/(g \Delta t)$, since we are using $q = p\Delta t$ for the pressure in this step. Then, $\frac{\partial q}{\partial z} = \Delta t \frac{\partial p}{\partial z}$, so the original boundary condition gets multiplied by $\Delta t$:
$\frac{\partial q}{\partial z} = -p/ g\Delta t + \eta^*/\Delta t$, but $p$ is $q/\Delta t$, which puts another $\Delta t$ factor in the denominator. 
- Typo in setting the diagonal: should be `@inbounds diagonal[i, j, Nz] = (- 1 / den - 1/Δzᵃᵃᶠ(i, j, Nz, grid)) - Δzᶜ * (λx[i] + λy[j])` (i.e. now have the `1/Δzᵃᵃᶠ(i, j, Nz, grid)` term. 
- Typo in [this](https://github.com/CliMA/Oceananigans.jl/blob/d0272ad9d12e3501084694c8c1f0c7169fe7ee25/src/Models/NonhydrostaticModels/solve_for_pressure.jl#L135) line
- [This](https://github.com/CliMA/Oceananigans.jl/blob/d0272ad9d12e3501084694c8c1f0c7169fe7ee25/src/Models/NonhydrostaticModels/solve_for_pressure.jl#L72) line should be -=, not +=
- The [free surface update](https://github.com/CliMA/Oceananigans.jl/blob/d0272ad9d12e3501084694c8c1f0c7169fe7ee25/src/Models/NonhydrostaticModels/pressure_correction.jl#L77) in the `set_top_pressure_boundary_condition!` kernel should be `η[i, j] = (p_Δt[i, j, Nz]/Δt + (p_Δt[i, j, Nz]/Δt)*(a/b) + η★/(Δt^2*b)) / (2*g)`
- `w` at `Nz+1` needs to be updated in the velocity correction step, i.e. adding something like this to the `make_pressure_correction!` kernel:
```
if k == grid.Nz
   U.w[i, j, k+1] -= (pNHSΔt[i,j,k+1] - pNHSΔt[i,j,k])/Δzᶜᶜᶜ(i, j, k, grid) 
end
```
- In the previous version, I had made a couple of changes to `field_tuples` ([1](https://github.com/shriyafruitwala/Oceananigans.jl/blob/e3a02e1ce5cfdb742335e07b510c9b1a27e1fb19/src/Fields/field_tuples.jl#L193), [2](https://github.com/shriyafruitwala/Oceananigans.jl/blob/e3a02e1ce5cfdb742335e07b510c9b1a27e1fb19/src/Fields/field_tuples.jl#L253)), and that change still seems to be necessary (the original line is commented out and the change is right below). I believe this issue came from w always being set to 0 at the surface.
- [This](https://github.com/CliMA/Oceananigans.jl/blob/d0272ad9d12e3501084694c8c1f0c7169fe7ee25/src/Solvers/fourier_tridiagonal_poisson_solver.jl#L250C4-L252C8) line in the solver that subtracts the mean is not working correctly. The mean is still being subtracted despite the formulation being inhomogeneous (when I comment it out entirely, the test case runs). 

https://github.com/user-attachments/assets/82181f2c-3bc8-4894-9028-21039aa07a95

glwagner:
Nice work @shriyafruitwala ! I think we can merge this as-is then --- we can address the issue of w* computation at Nz+1 in a later PR?

shriyafruitwala:
> In the previous version, I had made a couple of changes to field_tuples ([1](https://github.com/shriyafruitwala/Oceananigans.jl/blob/e3a02e1ce5cfdb742335e07b510c9b1a27e1fb19/src/Fields/field_tuples.jl#L193), [2](https://github.com/shriyafruitwala/Oceananigans.jl/blob/e3a02e1ce5cfdb742335e07b510c9b1a27e1fb19/src/Fields/field_tuples.jl#L253)), and that change still seems to be necessary (the original line is commented out and the change is right below). I believe this issue came from w always being set to 0 at the surface.
>
> [This](https://github.com/CliMA/Oceananigans.jl/blob/d0272ad9d12e3501084694c8c1f0c7169fe7ee25/src/Solvers/fourier_tridiagonal_poisson_solver.jl#L250C4-L252C8) line in the solver that subtracts the mean is not working correctly. The mean is still being subtracted despite the formulation being inhomogeneous (when I comment it out entirely, the test case runs).

@glwagner I think these two points should be addressed before merging, right?

glwagner:
> > In the previous version, I had made a couple of changes to field_tuples ([1](https://github.com/shriyafruitwala/Oceananigans.jl/blob/e3a02e1ce5cfdb742335e07b510c9b1a27e1fb19/src/Fields/field_tuples.jl#L193), [2](https://github.com/shriyafruitwala/Oceananigans.jl/blob/e3a02e1ce5cfdb742335e07b510c9b1a27e1fb19/src/Fields/field_tuples.jl#L253)), and that change still seems to be necessary (the original line is commented out and the change is right below). I believe this issue came from w always being set to 0 at the surface.
> > [This](https://github.com/CliMA/Oceananigans.jl/blob/d0272ad9d12e3501084694c8c1f0c7169fe7ee25/src/Solvers/fourier_tridiagonal_poisson_solver.jl#L250C4-L252C8) line in the solver that subtracts the mean is not working correctly. The mean is still being subtracted despite the formulation being inhomogeneous (when I comment it out entirely, the test case runs).
> 
> @glwagner I think these two points should be addressed before merging, right?

Ah yes ok. We need new boundary conditions on `w` and we need to test that the pressure solver isn't subtracting the mean. I can take a look.

src/Fields/field_tuples.jl
```diff
@@ -135,8 +135,9 @@ function VelocityFields(grid::AbstractGrid, user_bcs = NamedTuple())
 
     u = XFaceField(grid, boundary_conditions=bcs.u)
     v = YFaceField(grid, boundary_conditions=bcs.v)
-    w = ZFaceField(grid, boundary_conditions=bcs.w)
-
+    # w = ZFaceField(grid, boundary_conditions=bcs.w)
+    w = ZFaceField(grid)
```
glwagner: Instead of this, let's add a line to the model constructor that ensures the `w` boundary conditions are correct. I think the `HydrostaticFreeSurfaceModel` uses `nothing` boundary condition on top.

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -41,6 +42,7 @@ mutable struct NonhydrostaticModel{TS, E, A<:AbstractArchitecture, G, T, B, R, S
          stokes_drift :: SD       # Set of parameters for surfaces waves via the Craik-Leibovich approximation
               forcing :: F        # Container for forcing functions defined by the user
               closure :: E        # Diffusive 'turbulence closure' for all model fields
+         free_surface :: FS       # Diffusive 'turbulence closure' for all model fields
```
glwagner:
```suggestion
         free_surface :: FS       # Linearized free surface representation (Nothing for rigid lid)
```

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -34,6 +70,10 @@ Update the predictor velocities u, v, and w with the non-hydrostatic pressure mu
     @inbounds U.u[i, j, k] -= ∂xᶠᶜᶜ(i, j, k, grid, pNHSΔt)
     @inbounds U.v[i, j, k] -= ∂yᶜᶠᶜ(i, j, k, grid, pNHSΔt)
     @inbounds U.w[i, j, k] -= ∂zᶜᶜᶠ(i, j, k, grid, pNHSΔt)
+
+    if k == grid.Nz
+        U.w[i, j, k+1] -= (pNHSΔt[i,j,k+1] - pNHSΔt[i,j,k])/Δzᶜᶜᶜ(i, j, k, grid) 
+    end
```
glwagner: @shriyafruitwala I think we want to put this update into a separate kernel that is only called for the implicit free surface

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -10,15 +10,53 @@ function compute_pressure_correction!(model::NonhydrostaticModel, Δt)
...
+    @inbounds bc_rhs[i, j, 1] = η[i, j, Nz+1] / Δt + w̃[i, j, Nz+1]
+
+    # Compute ηⁿ⁺¹
+    Δzᶠ = Δzᵃᵃᶠ(i, j, Nz+1, grid)
+    a = 1 / Δzᶠ - 1 / (2g * Δt^2)
+    b = 1 / Δzᶠ + 1 / (2g * Δt^2)
+    @inbounds begin
+        η★ = η[i, j, Nz+1] + Δt * w̃[i, j, Nz+1]
+        η[i, j, Nz+1] = (p_Δt[i, j, Nz] / Δt + (p_Δt[i, j, Nz] * a / (b * Δt) + η★ / (b * Δt^2))) / 2g
+    end
```
glwagner, line 57: @shriyafruitwala I cleaned up the math here (spaces around arithmetic operators), and I also added correct indexing for the free surface. It's sort of unfortunate that the 2D indexing "works" for eta --- but note that we should always use 3D indexing for Oceananigans fields.

glwagner: Can you take a look and verify that the changes are correct?

glwagner: Another change, I think, is adding `@inbounds`

src/Models/NonhydrostaticModels/NonhydrostaticModels.jl
```diff
@@ -16,32 +16,40 @@ using Oceananigans.DistributedComputations: DistributedFFTBasedPoissonSolver, Di
...
+nonhydrostatic_pressure_solver(arch, grid::XYZRegularRG, ::Nothing) = FFTBasedPoissonSolver(grid)
+
+function nonhydrostatic_pressure_solver(arch, grid::GridWithFourierTridiagonalSolver, ::Nothing)
+    return FourierTridiagonalPoissonSolver(grid)
+end
+
+function nonhydrostatic_pressure_solver(arch, grid::Union{GridWithFourierTridiagonalSolver, XYZRegularRG}, free_surface)
+    tridiagonal_formulation = InhomogeneousFormulation(ZDirection())
+    return FourierTridiagonalPoissonSolver(grid; tridiagonal_formulation)
+end
```
glwagner, line 45: @shriyafruitwala this is how I enforce the appropriate formulation (which should then avoid the gauge condition correction in the solver later on)

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -196,25 +213,43 @@ function NonhydrostaticModel(; grid,
...
+        w_bcs = FieldBoundaryConditions(top = nothing,
+                                        bottom = w_bcs.bottom,
+                                        east = w_bcs.east,
+                                        west = w_bcs.west,
+                                        south = w_bcs.south,
+                                        north = w_bcs.north,
+                                        immersed = w_bcs.immersed)
+
+        boundary_conditions  =  merge(boundary_conditions, (; w = w_bcs))
+    end
```
glwagner, line 233: @shriyafruitwala note here how we enforce `nothing` boundary condition on `w` at the top

src/Fields/field_tuples.jl
glwagner:
```suggestion
    w = ZFaceField(grid, boundary_conditions=bcs.w)
```

src/Fields/field_tuples.jl
```diff
@@ -191,7 +192,8 @@ function VelocityFields(proposed_velocities::NamedTuple{(:u, :v, :w)}, grid, bcs
 
     u = XFaceField(grid, boundary_conditions=bcs.u, data=proposed_velocities.u.data)
     v = YFaceField(grid, boundary_conditions=bcs.v, data=proposed_velocities.v.data)
-    w = ZFaceField(grid, boundary_conditions=bcs.w, data=proposed_velocities.w.data)
+    # w = ZFaceField(grid, boundary_conditions=bcs.w, data=proposed_velocities.w.data)
+    w = ZFaceField(grid, data=proposed_velocities.w.data)
```
glwagner:
```suggestion
    w = ZFaceField(grid, boundary_conditions=bcs.w, data=proposed_velocities.w.data)
```

shriyafruitwala:
@glwagner Everything looks good to me and works with my surface gravity wave test! These lines:
```
# if solver.tridiagonal_formulation isa HomogeneousNeumannFormulation
#     ϕ .= ϕ .- mean(ϕ)
# end
```
are still commented out in the solver, but other than that and the w update at the surface, I think it is ready!

src/Solvers/fourier_tridiagonal_poisson_solver.jl
```diff
@@ -230,7 +248,9 @@ function solve!(x, solver::FourierTridiagonalPoissonSolver, b=nothing)
     # Solutions to Poisson's equation are only unique up to a constant (the global mean
     # of the solution), so we need to pick a constant. We choose the constant to be zero
     # so that the solution has zero-mean.
-    ϕ .= ϕ .- mean(ϕ)
+    # if solver.tridiagonal_formulation isa HomogeneousNeumannFormulation
+    #     ϕ .= ϕ .- mean(ϕ)
+    # end
```
glwagner:
```suggestion
    if solver.tridiagonal_formulation isa HomogeneousNeumannFormulation
        ϕ .= ϕ .- mean(ϕ)
    end
```

glwagner:
> @glwagner Everything looks good to me and works with my surface gravity wave test! These lines:
> 
> ```
> # if solver.tridiagonal_formulation isa HomogeneousNeumannFormulation
> #     ϕ .= ϕ .- mean(ϕ)
> # end
> ```
> 
> are still commented out in the solver, but other than that and the w update at the surface, I think it is ready!

Ok! If you will do the honor, can you approve this PR officially by reviewing it?

src/Models/NonhydrostaticModels/solve_for_pressure.jl
```diff
@@ -66,16 +66,44 @@ function compute_source_term!(solver::DistributedFourierTridiagonalPoissonSolver
...
-function compute_source_term!(solver::FourierTridiagonalPoissonSolver, Ũ)
+add_inhomogeneous_boundary_terms!(rhs, ::Nothing, grid, Ũ, Δt) = nothing
+
+@kernel function _add_inhomogeneous_boundary_terms!(rhs, grid, w̃, Δt, g, η)
+    i, j = @index(Global, NTuple)
+    Nz = grid.Nz
+    Δzᶠ = Δzᵃᵃᶠ(i, j, Nz+1, grid)
+
+    @inbounds begin
+        num = η[i, j] + Δt * w̃[i, j, Nz+1]
```
shriyafruitwala: $\eta$ is still being indexed as a 2d field here. I'm not sure I totally understand the reasoning behind the 3d indexing. Is it just for consistency with the other fields?

src/Models/NonhydrostaticModels/pressure_correction.jl
shriyafruitwala: What is the plan for this update (i.e. are we keeping it as is in this PR?)?

src/Models/NonhydrostaticModels/solve_for_pressure.jl
glwagner: The underlying arrays are 3D and there's no tests or support for 2D indexing. It works now essentially by coincidence but could break in the future.

src/Models/NonhydrostaticModels/pressure_correction.jl
glwagner: We should add another kernel to update `w` at `k = Nz+1` that is only launched `!isnothing(model.free_surface)`.

glwagner:
@shriyafruitwala I still need you to formally "approve" this PR. This means clicking a few buttons here on this PR.

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -36,15 +74,32 @@ Update the predictor velocities u, v, and w with the non-hydrostatic pressure mu
     @inbounds U.w[i, j, k] -= ∂zᶜᶜᶠ(i, j, k, grid, pNHSΔt)
 end
 
+@kernel function _compute_surface_vertical_velocity!(w, grid, pNHSΔt)
+    i, j = @index(Global, NTuple)
+    k = grid.Nz + 1
+    @inbounds U.w[i, j, k] -= ∂zᶜᶜᶠ(i, j, k, grid, pNHSΔt)
```
shriyafruitwala: I think this line should be `@inbounds w[i, j, k] -= ∂zᶜᶜᶠ(i, j, k, grid, pNHSΔt)`

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -36,15 +74,32 @@ Update the predictor velocities u, v, and w with the non-hydrostatic pressure mu
...
+    launch!(arch, grid, :xyz,
             _make_pressure_correction!,
             model.velocities,
             model.grid,
             model.pressures.pNHS)
 
+    if !isnothing(model.free_surface)
+        launch!(arch, grid, :xy,
+                _compute_surface_vertical_velocity!,
+                model.velocities,
```
shriyafruitwala: This should be `model.velocities.w`

shriyafruitwala (changes requested):
Left a couple comments under some typos in the kernel function

src/Models/NonhydrostaticModels/pressure_correction.jl
glwagner:
```suggestion
    @inbounds w[i, j, k] -= ∂zᶜᶜᶠ(i, j, k, grid, pNHSΔt)
```

src/Models/NonhydrostaticModels/pressure_correction.jl
```diff
@@ -36,15 +74,32 @@ Update the predictor velocities u, v, and w with the non-hydrostatic pressure mu
...
```
glwagner:
```suggestion
                model.velocities.w,
```

src/Models/NonhydrostaticModels/pressure_correction.jl
glwagner: Thank you!

PS you can also make specific suggestions with a "suggest changes" feature which speeds up the review process even more: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-comments-to-a-pull-request

jagoosw:
Hi both, I have been keeping an eye on this PR since it looks really interesting and thought I'd try it out since it seems like its nearing completion! I was wondering, is it within the scope of this PR to allow the free surface to be forced and/or boundary conditions to be applied to it? I guess you could apply boundary conditions by materialising the field before but it doesn't seem to work with the normal `boundary_conditions = (\eta = ...`.

glwagner:
> Hi both, I have been keeping an eye on this PR since it looks really interesting and thought I'd try it out since it seems like its nearing completion! I was wondering, is it within the scope of this PR to allow the free surface to be forced and/or boundary conditions to be applied to it? I guess you could apply boundary conditions by materialising the field before but it doesn't seem to work with the normal `boundary_conditions = (\eta = ...`.

I suggest we try doing that in a new PR! I'm not sure about boundary conditions --- possibly, equivalent boundary conditions can be applied to the momentum equation. Certainly forcing would produce new behavior though and I think that can be added. This would also apply to the hydrostatic model (cc @simone-silvestri)

glwagner:
FYI @joernc posted notes on the implicit free surface formulation for NonhydrostaticModel here: https://github.com/CliMA/Oceananigans.jl/pull/3968#issuecomment-2641296779

glwagner:
Scrutinizing the notes, I think there's scope to add a forcing term to equation 3, which eventually ends up in equation 9 (the boundary condition for the pressure solve). This boundary condition is implemented here:

https://github.com/CliMA/Oceananigans.jl/blob/60d0a23d8d041686491c49cf39b3311eb07cd696/src/Models/NonhydrostaticModels/solve_for_pressure.jl#L79

so that is probably where the forcing would go. Its doable but requires some restructuring, so I think best saved for a future PR.

As for horizontal boundary conditions -- I am less sure. It seems that via eq 3 this would have to be interpreted as a boundary condition on pressure (eg equation 3). In our incompressible situation, this in turn may be more easily interpreted as an open boundary condition of some kind, which moves implementation to modifying the velocity field / momentum equation rather than changing the pressure solver. @jagoosw can you give more specifics on the physics situation you are trying to describe? Similar situations may be described in the MITgcm docs so we could look there.

jagoosw:
> Scrutinizing the notes, I think there's scope to add a forcing term to equation 3, which eventually ends up in equation 9 (the boundary condition for the pressure solve). This boundary condition is implemented here:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/60d0a23d8d041686491c49cf39b3311eb07cd696/src/Models/NonhydrostaticModels/solve_for_pressure.jl#L79
> 
> so that is probably where the forcing would go. Its doable but requires some restructuring, so I think best saved for a future PR.
> 
> As for horizontal boundary conditions -- I am less sure. It seems that via eq 3 this would have to be interpreted as a boundary condition on pressure (eg equation 3). In our incompressible situation, this in turn may be more easily interpreted as an open boundary condition of some kind, which moves implementation to modifying the velocity field / momentum equation rather than changing the pressure solver. @jagoosw can you give more specifics on the physics situation you are trying to describe? Similar situations may be described in the MITgcm docs so we could look there.

I realise now that the random things I initially thought of would probably actually be momentum forcing, for example a model of a wave tank the momentum should be whats being forced. I guess you would need some kind of "forcing" if you wanted to relax to some free surface e.g. at an open boundary?

glwagner:
> I guess you would need some kind of "forcing" if you wanted to relax to some free surface e.g. at an open boundary?

I think you're right that with a prognostic free surface (eg split-explicit) we need to provide conditions on the displacement at an open boundary for things like tides. @simone-silvestri and @jm-c might have more to add.

tomchor:
I am seeing now this PR exists and I notice that there aren't any tests for this other than direct tests for the pressure solver. Do we want to add a more complete test here?

glwagner:
Not a bad idea to add a simple "existence" test that ensures we can take a time-step without an error.

@shriyafruitwala also has a validation case involving a deep water gravity wave. We could come up with an accuracy-based dynamics test with that.

tomchor:
> Not a bad idea to add a simple "existence" test that ensures we can take a time-step without an error.
> 
> @shriyafruitwala also has a validation case involving a deep water gravity wave. We could come up with an accuracy-based dynamics test with that.

I agree, I think we should add at least a time-stepping test as a minimum to ensure that this doesn't break. An accuracy test would be good too I think, but I'll let @shriyafruitwala comment on that.

shriyafruitwala:
> I agree, I think we should add at least a time-stepping test as a minimum to ensure that this doesn't break. An accuracy test would be good too I think, but I'll let @shriyafruitwala comment on that.

Here is some code for a deep water surface gravity wave initialized with a Gaussian bump in the sea surface height that I have been using to test the boundary condition, if that is helpful! It includes code that calculates the analytical solution for comparison. 

```
using Oceananigans
using Oceananigans.Models.HydrostaticFreeSurfaceModels: ImplicitFreeSurface
using GLMakie
using Oceananigans.Units

# grid setup: flat bottom, free surface
Nx, Nz = 50, 50
H = 50meters
L = 50meters
g = 10.0

coriolis = FPlane(latitude=28)
f = coriolis.f
dz = H/Nz
z = -H:dz:0
grid = RectilinearGrid(size = (Nx, Nz),
                            x = (-25meters, 25meters), 
                            z = z,
                            halo = (4,4),

free_surface = ImplicitFreeSurface(gravitational_acceleration=10)
model = NonhydrostaticModel(; grid, free_surface, coriolis, advection=WENO(order=5), buoyancy = BuoyancyTracer(), tracers = :b) 

# initialize with gaussian bump surface
η = model.free_surface.η
bump(x) = 1*exp(-(x^2)/(2 * (5meters)^2))
x_vals = LinRange(-L/2, L/2, Nx)
η₀_vals = bump.(x_vals)
η₀ = reshape(η₀_vals, (Nx,1,1))
set!(η, η₀)

# sets up simulation run
using Statistics
using Printf

simulation = Simulation(model, Δt=0.01, stop_time=5)

function print_stats(sim)
    w_max = maximum(abs(model.velocities.w))
    w_min = minimum(abs(model.velocities.w))
    w_mean = mean(model.velocities.w)

    u_max = maximum(abs(model.velocities.u))
    u_min = minimum(abs(model.velocities.u))
    u_mean = mean(model.velocities.u)    

    @info @sprintf("Time: %.2f, max|w|: %.3e, min|w|: %.3e, mean(w): %.3e",
                   time(sim), w_max, w_min, w_mean)
    @info @sprintf("Time: %.2f, max|u|: %.3e, min|u|: %.3e, mean(u): %.3e",
                   time(sim), u_max, u_min, u_mean)
    return nothing
end

progress(sim) = @info string(iteration(sim), ": ", time(sim))
add_callback!(simulation, progress, IterationInterval(10))
add_callback!(simulation, print_stats, IterationInterval(10)) 

output_file = "nonhydro_robin_new_test.jld2"
ow = JLD2Writer(model, merge(model.velocities, (; b = model.tracers[:b], η=model.free_surface.η)),
                      filename = output_file,  
                      schedule = IterationInterval(1),
                      overwrite_existing = true)

simulation.output_writers[:jld2] = ow

run!(simulation)

# produces animation comparing simulation with analytical solution
using FFTW
using Printf

# load field
ηt = FieldTimeSeries(output_file, "η") 
Nt = length(ηt)

# parameters
x_sim = xnodes(ηt)
N = length(x_sim)
L = 50.0
a = 1.0
δ = 5.0
Δx = x_sim[2] - x_sim[1]

times = ηt.times
k = [2π * i / L for i in 0:(N÷2)]
n = 0:(N÷2)
η0 = a .* exp.(-(x_sim .^ 2) ./ (2 * δ^2))
η0_hat = rfft(η0) ./ 2

function η_fourier(t)
    ω = sqrt.(g .* abs.(k))
    η_hat_t = η0_hat .* exp.(-1im .* ω .* t) .+ η0_hat .* exp.(1im .* ω .* t)
    return irfft(η_hat_t, N)
end

#plot
fig = Figure(; size=(1000, 400))

ax_sim = Axis(fig[1, 1], xlabel="x (m)", ylabel="η", title="Oceananigans Free Surface")
ax_fft = Axis(fig[1, 2], xlabel="x (m)", ylabel="η", title="Fourier-Analytical Free Surface")
n = Observable(1)
η_sim = @lift interior(ηt[$n], :, 1, 1)
η_fft = Observable(η_fourier(0.0))

lines!(ax_sim, x_sim, η_sim)
lines!(ax_fft, x_sim, η_fft)

ylims!(ax_sim, -1.5, 1.5)
ylims!(ax_fft, -1.5, 1.5)

record(fig, output_file * "eta.mp4", 1:Nt; framerate=24) do frame
    @info "Drawing frame $frame of $Nt..."
    n[] = frame
    η_fft[] = η_fourier(times[frame])
end
```

tomchor:
@shriyafruitwala any change you can clean up the example you just shared into a simple, small (as in very few lines of code and also as in "it runs fast because it's a small problem") function that returns true if the result is expected? That way I can open a PR and include that as a test.

PS: If you feel like it's easier, feel free to open PR yourself too. I do think it's important to have a test otherwise we risk this amazing new capability silently breaking in the future.

shriyafruitwala:
> @shriyafruitwala any change you can clean up the example you just shared into a simple, small (as in very few lines of code and also as in "it runs fast because it's a small problem") function that returns true if the result is expected? That way I can open a PR and include that as a test.

@tomchor I'm so sorry -- I am only just now seeing this! Do you mean something like this?

```
using Oceananigans
using FFTW
using Statistics

function test_free_surface(; Nx=25, Nz=25, H=50, L=50, g=10, a=1.0, δ=5.0, 
                            stop_time=10.0, tolerance=0.15)

    grid = RectilinearGrid(size=(Nx, Nz), 
                            x=(-L/2, L/2), z=(-H, 0), halo=(4, 4), 
                            topology=(Periodic, Flat, Bounded))

    free_surface = ImplicitFreeSurface(gravitational_acceleration=g)
    model = NonhydrostaticModel(; grid, free_surface, 
                                advection=WENO(order=5),
                                tracers=())
    
    # Gaussian bump
    x_vals = xnodes(model.free_surface.η)
    η₀ = a .* exp.(-(x_vals .^ 2) ./ (2 * δ^2))
    set!(model.free_surface.η, reshape(η₀, (Nx, 1, 1)))
    
    # run sim
    simulation = Simulation(model, Δt=0.01, stop_time=stop_time)
    run!(simulation)
    
    # check analytical solution
    k = [2π * i / L for i in 0:(Nx÷2)]
    η0_hat = rfft(η₀) / 2
    ω = sqrt.(g .* abs.(k))
    η_hat_final = η0_hat .* exp.(-1im .* ω .* stop_time) .+ η0_hat .* exp.(1im .* ω .* stop_time)
    η_analytical = irfft(η_hat_final, Nx)
    
    # compare solutions
    η_numerical = interior(model.free_surface.η, :, 1, 1)
    error = sqrt(mean((η_numerical .- η_analytical).^2))
    max_amplitude = maximum(abs.(η_analytical))
    relative_error = error / max_amplitude
    
    @info "Relative RMSE: $relative_error"
    
    return relative_error < tolerance
end

@assert test_free_surface() "Free surface wave test failed!"
```

PR #4748, opened 2025-08-31
Fill additional CenterField halos on cubed sphere panels for correct SSH interpolation

siddharthabishnu:
This PR fills additional halo regions of `CenterField` on cubed sphere panels. Previously, when interpolating sea surface height (SSH), η, from cell centers to edge locations (Center-Face and Face-Center), uninitialized corner halo values of η produced incorrect interpolated values at the edges. This, in turn, affected the computed column depths at those edge locations.

With this update, the missing η halos are filled, ensuring that interpolation to edge locations produces consistent results. This eliminates the spurious values that could previously lead to inaccuracies and potential instabilities.

Finally, this PR enlarges the extended halo region to include the extra layer needed for `CenterField` to correctly interpolate to the outermost southern and western edges of cubed sphere panels.


siddharthabishnu:
Closing this PR, as I realized that filling these additional halos for a `CenterField` in corner regions is not only ambiguous but also unnecessary, provided that the relevant operators (including the above-mentioned interpolation operator) at halo edges are correctly extended for the extended grid.

PR #4749, opened 2025-09-01
(0.100.2) Refactor checkpointing restoring + set particles from checkpointing only if they exist

simone-silvestri:
`set!`ing a model from a checkpoint requires trying to set the particles. Some models do not have particles, this PR allows to use checkpointing with models without particles (in particular sea-ice)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.100.1"
+version = "0.100.2"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"


glwagner:
I don't think this is the design we want.

We should use dispatch so that each model can define a set/pickup function which is appropriate for the specific model.

The issue here is that we previously only checkpointed two models (NonhydrostaticModel and HydrostaticFreeSurfaceModel) which are very similar.

simone-silvestri:
Then it should be better to define `set!` for `OceananigansModels` directly in the Models module, but the we have to extend set for every new model.

simone-silvestri:
Makes sense, however, most models just need to fill prognostic fields and clock, so it is good to keep a fallback given that we have a generic `Checkpointer`

simone-silvestri:
> Define
> 
> ```julia
> set!(model::NonhydrostaticModel, filepath::AbstractString)
> set!(model::HydrostaticFreeSurfaceModel, filepath::AbstractString)
> ```
> 
> in the `Models` module. All new models that want to support checkpointing will define this function.

This should be done, this PR should be ready to merge if someone approves

src/Models/set_model.jl
```diff
@@ -0,0 +1,88 @@
...
+        else
+            @warn "Particles do not exist in checkpoint and could not be restored."
+        end
+
+        checkpointed_clock = file["$addr/clock"]
+
+        # Update model clock
+        model.clock.iteration = checkpointed_clock.iteration
+        model.clock.time = checkpointed_clock.time
+        model.clock.last_Δt = checkpointed_clock.last_Δt
```
navidcy: Should we use `set_clock!` method here?

https://github.com/CliMA/Oceananigans.jl/blob/ed950ba05bf2bf0d249a9cd649dbe97e6fd2e56a/src/TimeSteppers/clock.jl#L52-L65

src/Models/set_model.jl
```diff
@@ -0,0 +1,88 @@
...
+
+#####
+##### set! for checkpointer filepaths and `OceananigansModels`
+#####
+
+"""
+    set!(model::OceananigansModels, filepath::AbstractString)
+
+Set data in `model.velocities`, `model.tracers`, `model.timestepper.Gⁿ`, and
+`model.timestepper.G⁻` to checkpointed data stored at `filepath`.
```
navidcy:
```suggestion
Set data in `model.velocities`, `model.tracers` and the timestepper tendencies `model.timestepper.Gⁿ` and
`model.timestepper.G⁻` (for non-self starting timesteppers) to the checkpointed data stored at `filepath`.
```

src/Models/set_model.jl
navidcy: Btw I'm also very tempted to change `set_clock!` to just `set!(clock::Clock, new_clock::Clock)`

navidcy: see 5cad6dfb7b78aa74130a2b959581b801bbb7a88a

I think we should use the method here, but I'm not totally sure if we don't want to set the stage and last_stage from the saved clock...

navidcy: I did it in [c258f7c](https://github.com/CliMA/Oceananigans.jl/pull/4749/commits/c258f7cac8fbbc865b97f43925ba351aff47c05c) but revert if that's not what we want.

navidcy:
@glwagner since you requested changed your approval is required

src/OutputWriters/checkpointer.jl
```diff
@@ -225,49 +218,11 @@ function set!(model, filepath::AbstractString)
...
-
-            # Tendency "n-1"
-            parent_data = on_architecture(arch, file["$addr/timestepper/G⁻/$name/data"])
-
-            tendency⁻_field = timestepper.G⁻[name]
-            @apply_regionally copyto!(parent(tendency⁻_field), parent_data)
-        elseif tendency_in_model && !tendency_in_checkpoint
-            @warn "Tendencies for $name do not exist in checkpoint and could not be restored."
-        end
+        set!(model, checkpointed_clock)
```
simone-silvestri, line 224: Isn't it like this?
```suggestion
        set!(model.clock, checkpointed_clock)
```

navidcy: There is also a shorthand method that takes model as first arg. Either would work!

navidcy: https://github.com/CliMA/Oceananigans.jl/blob/db5937d508f4885abe0cb1efc6b7f4b017c8b0a3/src/Models/Models.jl#L119

navidcy:
I think we merge this when CI passes.

PR #4755, opened 2025-09-03
Move Buildkite CI to Nautilus

glwagner:


ali-ramadhan:
Tests are running :tada: Hopefully they all pass! @tomchor Looks like Reactant tests started successfully.

Initializing the environment took almost 20 minutes but maybe it's just the first time. I'd like to modify the Buildkite steps so that the Julia depot is cached, this should save a lot of time by making initialization only take ~1 minute in most cases.

But I should be able to just modify this PR.

simone-silvestri:
Nice! And the GPU reactant tests which are failing on main pass on nautilus

tomchor:
And it's so much faster too! :rocket:

ali-ramadhan:
Seems like it's just the docs that are failing, but I'm looking through the log file and not seeing any errors or warnings. It's huge but maybe I'm also rusty at this.

Some of the tests are taking longer than expected compared to https://github.com/CliMA/Oceananigans.jl/discussions/4678#discussioncomment-13938024, especially the Enzyme tests. Will look into this, but we should probably get the tests up and running first.

ali-ramadhan:
Never mind it was just a doctest. Side note but this kind of doctest expecting very specific numbers might not be the best:

```
│ Expected output:
│ 
│ 16×16×16 Field{Center, Center, Center} on RectilinearGrid on CPU
│ ├── grid: 16×16×16 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
│ ├── boundary conditions: FieldBoundaryConditions
│ │   └── west: Periodic, east: Periodic, south: Periodic, north: Periodic, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
│ └── data: 22×22×22 OffsetArray(::Array{Float64, 3}, -2:19, -2:19, -2:19) with eltype Float64 with indices -2:19×-2:19×-2:19
│     └── max=0.980785, min=-0.980785, mean=1.10534e-16

```

ali-ramadhan:
Should the string be "IRFT" or "IFRT" in this part of `pipeline.yml`?

```bash
      # Force IRFT runtime for Reactant (this can be moved to Project.toml on 1.11)
      touch LocalPreferences.toml
      echo "[Reactant]" >> LocalPreferences.toml
      echo "xla_runtime = \"IFRT\"" >> LocalPreferences.toml
      cat LocalPreferences.toml
```

ali-ramadhan:
Some builds might die as I'm refactoring the docker container and the pipeline.yaml to use `juliaup` and share the Julia depot between builds.

glwagner:
> Should the string be "IRFT" or "IFRT" in this part of `pipeline.yml`?
> 
> ```shell
>       # Force IRFT runtime for Reactant (this can be moved to Project.toml on 1.11)
>       touch LocalPreferences.toml
>       echo "[Reactant]" >> LocalPreferences.toml
>       echo "xla_runtime = \"IFRT\"" >> LocalPreferences.toml
>       cat LocalPreferences.toml
> ```

The string should be `"IFRT"`: https://enzymead.github.io/Reactant.jl/stable/tutorials/sharding#sharding

glwagner:
> Never mind it was just a doctest. Side note but this kind of doctest expecting very specific numbers might not be the best:
> 
> ```
> │ Expected output:
> │ 
> │ 16×16×16 Field{Center, Center, Center} on RectilinearGrid on CPU
> │ ├── grid: 16×16×16 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 3×3×3 halo
> │ ├── boundary conditions: FieldBoundaryConditions
> │ │   └── west: Periodic, east: Periodic, south: Periodic, north: Periodic, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
> │ └── data: 22×22×22 OffsetArray(::Array{Float64, 3}, -2:19, -2:19, -2:19) with eltype Float64 with indices -2:19×-2:19×-2:19
> │     └── max=0.980785, min=-0.980785, mean=1.10534e-16
> ```

agree, we should filter these numbers out of the doctest. cc @navidcy

glwagner:
> Seems like it's just the docs that are failing, but I'm looking through the log file and not seeing any errors or warnings. It's huge but maybe I'm also rusty at this.
> 
> Some of the tests are taking longer than expected compared to [#4678 (comment)](https://github.com/CliMA/Oceananigans.jl/discussions/4678#discussioncomment-13938024), especially the Enzyme tests. Will look into this, but we should probably get the tests up and running first.

I usually search for "error". But also, it would be nice to decrease the size of the output. I think `JULIA_DEBUG=Documenter` is a reason for this?

ali-ramadhan:
> I usually search for "error". But also, it would be nice to decrease the size of the output. I think `JULIA_DEBUG=Documenter` is a reason for this?

Yeah I needed to search for "failure" or "doctest failure" to find it, but I think I didn't find it originally because I was using the `bat` command to process all the ASCII color codes . Found it when I used plain old `cat` haha.

But yeah if we don't need all the debug statements we can turn them off.

glwagner:
> > I usually search for "error". But also, it would be nice to decrease the size of the output. I think `JULIA_DEBUG=Documenter` is a reason for this?
> 
> Yeah I needed to search for "failure" or "doctest failure" to find it, but I think I didn't find it originally because I was using the `bat` command to process all the ASCII color codes . Found it when I used plain old `cat` haha.
> 
> But yeah if we don't need all the debug statements we can turn them off.

I don't know if the debug statements are really useful. I think the main thing we need to do in the log is to find the errors.

glwagner:
Amazing!

I would be ok to merge this as-is since it is an improvement. We can always iterate on it in the future or revert if its not working...

A future optimization could be to move some of the cheaper tests back to tartarus.

Also, it looks like we need to split up the enzyme tests since they are a bottleneck.

@ali-ramadhan can you remind me what GPU nautilus has?

A final note... rather than moving unit tests and/or docs build back to tartarus, we could consider 1) using two GPUs on tartarus for distributed tests 2) using a second GPU for the ClimaOcean docs build, which is heavy (we are using 1 GPU right now).

ali-ramadhan:
Also agree with merging as-is since it works. We can keep making future improvements.

We might just need to fix a tolerance and a couple of race conditions? These test groups passed when retried but we should probably fix them:

1. https://buildkite.com/clima/oceananigans/builds/25420#019910fc-724a-4909-9295-52416071c31f/45-1106 (tolerance needs to be updated)
2. https://buildkite.com/clima/oceananigans/builds/25420#019910fc-7252-4f3d-93d0-0fbc729ef246/17-326 (file might be removed by another test?)
3. https://buildkite.com/clima/oceananigans/builds/25420#019910fc-724c-442c-b85b-0020b24bbdaf failing with `SystemError: opening file "/var/lib/buildkite-agent/Oceananigans.jl-25420/test/cubed_sphere_output_Float32_CPU_IG.jld2"` (file might also be removed by another test?)

Happy to fix these later today!

> @ali-ramadhan can you remind me what GPU nautilus has?

A V100 32GB PCIe: https://buildkite.com/clima/oceananigans/builds/25420#019910fc-7247-429c-8b3a-98647f9a9c54/45-926

navidcy:
> A final note... rather than moving unit tests and/or docs build back to tartarus, we could consider 1) using two GPUs on tartarus for distributed tests 2) using a second GPU for the ClimaOcean docs build, which is heavy (we are using 1 GPU right now).

I like this suggestion.

ali-ramadhan:
Tests should hopefully pass without retrying now!

Not sure why the `buildkite-agent pipeline upload` step took 30 minutes in https://buildkite.com/clima/oceananigans/builds/25425 but took 5 seconds (as expected) in the latest build. Hopefully it's just a one-time thing.

test/test_multi_region_cubed_sphere.jl
```diff
@@ -773,11 +773,6 @@ end
 end
 
 @testset "Testing simulation on conformal and immersed conformal cubed sphere grids" begin
-    for f in readdir(".")
-        if occursin(r"^cubed_sphere_(output|checkpointer)_.*\.jld2$", f)
-            rm(f; force=true)
-        end
-    end
```
ali-ramadhan, line 780: Hope this removal is okay. The reason for doing this is that this pattern, when executed by the GPU test group, can delete files being used by the CPU test group and vice versa.

And now the cleanup step deletes the entire Oceananigans.jl directory so the files are deleted at the end.

ali-ramadhan:
Also just a heads up that I've changed the CI design a bit.

Now different builds share the same Julia depot, i.e. the depot is cached. This saves a lot of time as instantiating a fresh environment can take 5-10 minutes. Initializing the environment now takes 1 minute.

Also all test groups run in the same Oceananigans.jl directory. This means they share the same Manifest.toml (resolved by the `init` step) and do not need to resolve package versions. This has also fixed an issue for me in the past where different test groups started precompiling the same package. This also means that some test groups started failing due to race conditions but it was only a couple which are now hopefully fixed.

The test groups do say `Could not use exact versions of packages in manifest, re-resolving` and I'm not sure why, but I have some ideas for fixes which I can try in a future PR.

Also re-iterating a good point brought up by @simone-silvestri that we should probably split up the tests into unit tests and integration tests at least.

tomchor (approved):

simone-silvestri:
@ali-ramadhan how did you cache the initialization step? Did you just remove the julia load path? I think we want to use the same approach for the distributed tests, where the initialization also takes > 20mins

ali-ramadhan:
> @ali-ramadhan how did you cache the initialization step? Did you just remove the julia load path? I think we want to use the same approach for the distributed tests, where the initialization also takes > 20mins

I just reused the Julia depot! Before we had

https://github.com/CliMA/Oceananigans.jl/blob/4cf7ceb98918949d2e889e394872a4f478a7a7f5/.buildkite/pipeline.yml#L5

which means each build has to create a Julia depot then download, install, and precompile every package from scratch. 

Now all builds use the same Julia depot:

https://github.com/CliMA/Oceananigans.jl/blob/a02649d768fdb734544da93c0140e487ae2b412b/.buildkite/pipeline.yml#L4

I just called it `.julia-oceananigans` instead of `.julia` so that different packages / Buildkite pipelines can have their own depots.

Not sure how the Caltech cluster works, but as long as you can persist a Julia depot somewhere (hopefully on an SSD) then it should reduce the initialization time down to 1-2 minutes, although this probably also depends on what CPUs the cluster uses.

simone-silvestri:
Looks good! I ll open a PR and see what happens

ali-ramadhan:
Another (probably important) note: All test groups/builds share the same working directory. You'll see that I set this environment variable:

https://github.com/CliMA/Oceananigans.jl/blob/a02649d768fdb734544da93c0140e487ae2b412b/.buildkite/pipeline.yml#L7

and all steps `cd $OCEANANIGANS_DIR` (with the init step using `rsync` to create the directory).

This is to share the exact same Manifest.toml and it helps ensure that the init step can do all the precompiling. Otherwise I've found that sometimes all the different steps will precompile a few packages then they conflict and have to wait which increases build times.

I'm not sure I fully understand why this works, but I do see that in the main Oceananigans pipeline, only the init step precompiles packages which is good and what we want.

simone-silvestri:
Thanks for the clarification. I am actually noticing that the distributed tests do not have a JULIA_DEPOT_PATH specified, so in principle, they should use the same one?

ali-ramadhan:
> Thanks for the clarification. I am actually noticing that the distributed tests do not have a JULIA_DEPOT_PATH specified, so in principle, they should use the same one?

I believe it should default to `$HOME/.julia` so yeah if it's not set they should use the same one. Unless `$HOME/.julia` is not persisted between Slurm jobs on the Caltech cluster?

PR #4757, opened 2025-09-03
Add function for resetting the model clock

glwagner:
xref: https://github.com/TUM-PIK-ESM/Terrarium.jl/issues/26#issuecomment-3248713100

### DIFF
--- a/src/Models/Models.jl
+++ b/src/Models/Models.jl
@@ -116,9 +116,6 @@ const OceananigansModels = Union{HydrostaticFreeSurfaceModel,
                                  ShallowWaterModel}
 
 set_clock!(model::OceananigansModels, new_clock) = set_clock!(model.clock, new_clock)
-
-# Fallback. Models without clocks should extend this function.
-reset_clock!(model::AbstractModel) = reset!(model.clock)
     
 """
     possible_field_time_series(model::OceananigansModels)
--- a/src/Models/Models.jl
+++ b/src/Models/Models.jl
@@ -116,7 +116,7 @@ const OceananigansModels = Union{HydrostaticFreeSurfaceModel,
                                  ShallowWaterModel}
 
 set_clock!(model::OceananigansModels, new_clock) = set_clock!(model.clock, new_clock)
-    
+
 """
     possible_field_time_series(model::OceananigansModels)
 
--- a/src/Models/Models.jl
+++ b/src/Models/Models.jl
@@ -117,6 +117,16 @@ const OceananigansModels = Union{HydrostaticFreeSurfaceModel,
 
 set_clock!(model::OceananigansModels, new_clock) = set_clock!(model.clock, new_clock)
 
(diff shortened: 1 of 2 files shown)


glwagner:
cc @bgroenks96

src/Models/Models.jl
```diff
@@ -117,6 +117,16 @@ const OceananigansModels = Union{HydrostaticFreeSurfaceModel,
...
 
+# Fallback. Models without clocks should extend this function.
+function reset_clock!(model::AbstractModel)
+    model.clock.time = 0
+    model.clock.last_Δt = Inf
+    model.clock.iteration = 0
+    model.clock.stage = 1
+    return nothing
+end
+
```
navidcy: This seems redundant? We can make use of 

https://github.com/CliMA/Oceananigans.jl/blob/cd238510315615d587e62a8918e6ef7b1b35d5c9/src/TimeSteppers/clock.jl#L43-L50

?

src/Simulations/simulation.jl
```diff
@@ -182,10 +183,7 @@ run_wall_time(sim::Simulation) = prettytime(sim.run_wall_time)
 Reset `sim`ulation, `model.clock`, and `model.timestepper` to their initial state.
 """
 function reset!(sim::Simulation)
-    sim.model.clock.time = 0
-    sim.model.clock.last_Δt = Inf
-    sim.model.clock.iteration = 0
-    sim.model.clock.stage = 1
+    reset_clock!(sim.model)
```
navidcy, line 185: I think that's the only line needed here; no  need to (re)define `reset_clock!` method

```suggestion
    reset!(sim.model.clock)
```

src/Models/Models.jl
glwagner: done

src/Simulations/simulation.jl
glwagner: The purpose of this addition is to relax the restricting that all subtypes of `AbstractModel` need to have a `clock` property. But I did incorporate the suggestion to re-use `reset!(model.clock)` within `reset_clock!(model)`.

bgroenks96:
Thanks for making this change so quickly!

Just to clarify though, I just meant `reset!` to be an example. I think we would need to go through all of the `Simulations` module and update all of the `sim.model.clock` references to use the corresponding methods to really make this work, no?

glwagner:
A good way to do this is to write a test, and then make changes as-needed until the test / simple example works. I doubt there is much to do.

I think we want to keep the notion that fallback models have a `clock`. It's important that models keep track of their time, and its nice also to encourage usage of `TimeSteppers.Clock` for a more uniform user experience. I think it also reduces the learning curve if users can come to expect to find `model.clock.time` and `model.clock.iteration` in this ecosystem. That said it's nice to allow exceptions if there is good motivation; hence this PR.

PR #4760, opened 2025-09-03
Validate location input to FieldBoundaryConditions

### LINKED ISSUE #4759: `no method matching default_prognostic_bc` error when building `FieldBoundaryConditions`
The following is failing on `main`:

```julia
julia> FieldBoundaryConditions(grid, (Center, Center, Center); top = GradientBoundaryCondition(1))
ERROR: MethodError: no method matching default_prognostic_bc(::Bounded, ::Type{Center}, ::Oceananigans.BoundaryConditions.DefaultBoundaryCondition{BoundaryCondition{Oceananigans.BoundaryConditions.Flux, Nothing}})

Closest candidates are:
  default_prognostic_bc(::Flat, ::Any, ::Any)
   @ Oceananigans ~/repos/Oceananigans.jl/src/BoundaryConditions/field_boundary_conditions.jl:17
  default_prognostic_bc(::Oceananigans.Grids.FullyConnected, ::Any, ::Any)
   @ Oceananigans ~/repos/Oceananigans.jl/src/BoundaryConditions/field_boundary_conditions.jl:16
  default_prognostic_bc(::Bounded, ::Nothing, ::Any)
   @ Oceananigans ~/repos/Oceananigans.jl/src/BoundaryConditions/field_boundary_conditions.jl:27
  ...

Stacktrace:
 [1] _default_auxiliary_bc(topo::Bounded, loc::Type)
   @ Oceananigans.BoundaryConditions ~/repos/Oceananigans.jl/src/BoundaryConditions/field_boundary_conditions.jl:34
 [2] default_auxiliary_bc(grid::RectilinearGrid{…}, ::Val{…}, loc::Tuple{…})
   @ Oceananigans.BoundaryConditions ~/repos/Oceananigans.jl/src/BoundaryConditions/field_boundary_conditions.jl:40
 [3] top-level scope
   @ REPL[6]:1
```

glwagner:
Resolves #4759

### DIFF
--- a/src/MultiRegion/multi_region_field.jl
+++ b/src/MultiRegion/multi_region_field.jl
@@ -7,7 +7,7 @@ using Oceananigans.OutputWriters: output_indices
 using Base: @propagate_inbounds
 
 import Oceananigans.DistributedComputations: reconstruct_global_field, CommunicationBuffers
-import Oceananigans.BoundaryConditions: regularize_field_boundary_conditions
+import Oceananigans.BoundaryConditions: regularize_field_boundary_conditions, FieldBoundaryConditions
 import Oceananigans.Grids: xnodes, ynodes
 import Oceananigans.Fields: set!, compute!, compute_at!, interior, validate_field_data, validate_boundary_conditions
 import Oceananigans.Fields: validate_indices, communication_buffers


src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -171,7 +172,7 @@ and the topology in the boundary-normal direction is used:
 - `nothing` for `Bounded` directions and `Face`-located fields
 - `nothing` for `Flat` directions and/or `Nothing`-located fields
 """
-function FieldBoundaryConditions(grid::AbstractGrid, loc, indices=(:, :, :);
+function FieldBoundaryConditions(grid::AbstractGrid, loc::Tuple, indices=(:, :, :);
                                  west     = default_auxiliary_bc(grid, Val(:west),   loc),
```
simone-silvestri: We could also just always `instantiate` the location, so that the constructors would work with both instantiated and non-instantiated locations.

glwagner: I think it may be a better/simpler design to require instantiation. I don't think it is inconvenient practically and simplifies using/reading code.

src/MultiRegion/multi_region_field.jl
```diff
@@ -193,7 +193,7 @@ function inject_regional_bcs(grid, connectivity, loc, indices;
     return FieldBoundaryConditions(indices, west, east, south, north, bottom, top, immersed)
 end
 
-FieldBoundaryConditions(mrg::MultiRegionGrids, loc, indices; kwargs...) =
+FieldBoundaryConditions(mrg::MultiRegionGrids, loc::Tuple, indices::MultiRegionObject; kwargs...) =
```
glwagner, line 196: @simone-silvestri is this the issue?

simone-silvestri: oh, I guess we are missing an `import Oceananigans.BoundaryConditions: FieldBoundaryConditions`?

glwagner: that was definitely missing (otherwise this method doesn't extend anything) but also that didn't fix it hmm

glwagner:
this went stale so I am reopening it

PR #4762, opened 2025-09-04
materialize `DefaultBoundaryCondition` from bcs in `Field` constructor

### LINKED ISSUE #4758: `no method matching fill_halo_kernel!` error when building `Fields` with explicit boundary conditions
The following fails on `main`, possibly due to the recently refactor:

```julia
julia> using Oceananigans

julia> grid = RectilinearGrid(topology=(Bounded, Bounded, Bounded), size=(4, 4, 4), extent=(1, 1, 1));

julia> boundary_conditions = FieldBoundaryConditions(top = GradientBoundaryCondition(1));

julia> b = CenterField(grid; boundary_conditions)
ERROR: MethodError: no method matching fill_halo_kernel!(::Oceananigans.BoundaryConditions.BottomAndTop, ::Oceananigans.BoundaryConditions.DefaultBoundaryCondition{…}, ::RectilinearGrid{…}, ::Symbol, ::Tuple{…}, ::OffsetArrays.OffsetArray{…}, ::Tuple{})

Closest candidates are:
  fill_halo_kernel!(::Oceananigans.BoundaryConditions.BottomAndTop, ::BoundaryCondition{<:Oceananigans.BoundaryConditions.MultiRegionCommunication}, ::Any, ::Any, ::Any, ::Any, ::Any)
   @ Oceananigans ~/repos/Oceananigans.jl/src/BoundaryConditions/fill_halo_kernels.jl:162
  fill_halo_kernel!(::Oceananigans.BoundaryConditions.BottomAndTop, ::Oceananigans.BoundaryConditions.DistributedCommunicationBoundaryCondition, ::Any, ::Any, ::Any, ::Any, ::Any)
   @ Oceananigans ~/repos/Oceananigans.jl/src/BoundaryConditions/fill_halo_kernels.jl:148
  fill_halo_kernel!(::Oceananigans.BoundaryConditions.BottomAndTop, ::BoundaryCondition{<:Oceananigans.BoundaryConditions.Periodic}, ::Any, ::Any, ::Any, ::Any, ::Any)
   @ Oceananigans ~/repos/Oceananigans.jl/src/BoundaryConditions/fill_halo_kernels.jl:132
  ...

Stacktrace:
  [1] fill_halo_kernels
    @ ~/repos/Oceananigans.jl/src/BoundaryConditions/fill_halo_kernels.jl:54 [inlined]
  [2] construct_boundary_conditions_kernels(bcs::FieldBoundaryConditions{…}, data::OffsetArrays.OffsetArray{…}, grid::RectilinearGrid{…}, loc::Tuple{…}, indices::Tuple{…})
    @ Oceananigans.BoundaryConditions ~/repos/Oceananigans.jl/src/BoundaryConditions/fill_halo_kernels.jl:19
  [3] #construct_regionally#82
    @ ~/repos/Oceananigans.jl/src/Utils/multi_region_transformation.jl:159 [inlined]
  [4] construct_regionally
    @ ~/repos/Oceananigans.jl/src/Utils/multi_region_transformation.jl:154 [inlined]
  [5] macro expansion
    @ ~/repos/Oceananigans.jl/src/Utils/multi_region_transformation.jl:244 [inlined]
  [6] (Field{…})(grid::RectilinearGrid{…}, data::OffsetArrays.OffsetArray{…}, bcs::FieldBoundaryConditions{…}, indices::Tuple{…}, op::Nothing, status::Nothing, buffers::Nothing)
    @ Oceananigans.Fields ~/repos/Oceananigans.jl/src/Fields/field.jl:35
  [7] Field
    @ ~/repos/Oceananigans.jl/src/Fields/field.jl:106 [inlined]
  [8] Field(loc::Tuple{…}, grid::RectilinearGrid{…}, T::DataType; indices::Tuple{…}, data::OffsetArrays.OffsetArray{…}, boundary_conditions::FieldBoundaryConditions{…}, operand::Nothing, status::Nothing)
    @ Oceananigans.Fields ~/repos/Oceananigans.jl/src/Fields/field.jl:196
  [9] CenterField
    @ ~/repos/Oceananigans.jl/src/Fields/field.jl:208 [inlined]
```

simone-silvestri:
this PR should allow the pattern described in #4758, for example:
```julia
julia> grid = RectilinearGrid(topology=(Bounded, Bounded, Bounded), size=(4, 4, 4), extent=(1, 1, 1));

julia> boundary_conditions = FieldBoundaryConditions(top = GradientBoundaryCondition(1));

julia> CenterField(grid; boundary_conditions).boundary_conditions
├── west: FluxBoundaryCondition: Nothing
├── east: FluxBoundaryCondition: Nothing
├── south: FluxBoundaryCondition: Nothing
├── north: FluxBoundaryCondition: Nothing
├── top: GradientBoundaryCondition: 1.0
└── immersed: Nothing

julia> XFaceField(grid; boundary_conditions).boundary_conditions
├── west: OpenBoundaryCondition{Nothing}: Nothing
├── east: OpenBoundaryCondition{Nothing}: Nothing
├── south: FluxBoundaryCondition: Nothing
├── north: FluxBoundaryCondition: Nothing
├── top: GradientBoundaryCondition: 1.0
└── immersed: Nothing
julia> grid = ImmersedBoundaryGrid(grid, GridFittedBottom(-1));

julia> YFaceField(grid; boundary_conditions).boundary_conditions
├── west: FluxBoundaryCondition: Nothing
├── east: FluxBoundaryCondition: Nothing
├── south: OpenBoundaryCondition{Nothing}: Nothing
├── north: OpenBoundaryCondition{Nothing}: Nothing
├── top: GradientBoundaryCondition: 1.0
```

closes #4758


simone-silvestri:
I guess we cannot really `regularize_boundary_conditions` because fields do not have the information of dependencies, but we can add another method that correctly replaces the `DefaultBoundaryCondition`

simone-silvestri:
As part of this PR, I have removed the `boundary_condition` inside the `DefaultBoundaryCondition`. 
In my opinion, the default is uniquely defined and should depend on location and grid; user-defined BCs should be explicitly specified rather than changing a default globally (which is not even possible because `Face` and `Center` locations do not accept the same bcs so this approach could very well lead to error)

src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -6,18 +6,14 @@ using GPUArraysCore
 ##### Default boundary conditions
 #####
 
-struct DefaultBoundaryCondition{BC}
-    boundary_condition :: BC
-end
-
-DefaultBoundaryCondition() = DefaultBoundaryCondition(NoFluxBoundaryCondition())
+struct DefaultBoundaryCondition end
```
glwagner, line 9: why did you change this design?

glwagner:
> I guess we cannot really `regularize_boundary_conditions` because fields do not have the information of dependencies, but we can add another method that correctly replaces the `DefaultBoundaryCondition`

No, this is ok --- only prognostic fields can have field dependencies, and those are built separately inside the model constructor. This PR applies to auxiliary fields only.

src/BoundaryConditions/field_boundary_conditions.jl
```diff
@@ -127,14 +124,13 @@ and the topology in the boundary-normal direction is used:
...
-                        bottom = DefaultBoundaryCondition(default_bounded_bc),
-                        top = DefaultBoundaryCondition(default_bounded_bc),
-                        immersed = DefaultBoundaryCondition(default_bounded_bc)) =
+FieldBoundaryConditions(; west = DefaultBoundaryCondition(),
+                          east = DefaultBoundaryCondition(),
+                          south = DefaultBoundaryCondition(),
+                          north = DefaultBoundaryCondition(),
+                          bottom = DefaultBoundaryCondition(),
+                          top = DefaultBoundaryCondition(),
+                          immersed = DefaultBoundaryCondition()) =
```
glwagner, line 133: I'd like to go back to the other design. It is meant to provide more generality for model developers, even though it does not apply to Oceananigans now. Some models may want to use `nothing` as the default.

glwagner: I feel this PR removes a nice feature that I designed for changing the default boundary condition. Note that this code is meant to be applicable to a wide class of models --- not just `NonhydrostaticModel` and `HydrostaticFreeSurfaceModel`.

glwagner: Also, this PR makes the previous code more useful...

src/BoundaryConditions/field_boundary_conditions.jl
simone-silvestri: > As part of this PR, I have removed the `boundary_condition` inside the `DefaultBoundaryCondition`. 
> In my opinion, the default is uniquely defined and should depend on location and grid; user-defined BCs should be explicitly specified rather than changing a default globally (which is not even possible because `Face` and `Center` locations do not accept the same bcs so this approach could very well lead to error)

simone-silvestri: I thought to remove it for this reason and in particular because `validate_boundary_conditions` is only grid and location dependent (not model dependent), so it basically forces a limited default choice. But of course, I am open to discuss about it and keep this feature if this is the consensus 😅

glwagner: I guess, what do you think about models / situations in which "no flux" is irrelevant. For example a model that has no derivatives.

simone-silvestri: We could have the default be a `nothing` like on `Face`s, this would then be general.

glwagner: That isn't general because models with gradients require `NoFluxBoundaryCondition()` as a default.

src/BoundaryConditions/field_boundary_conditions.jl
simone-silvestri: Ok, I can keep the default in the `DefaultBoundaryCondition`.  I guess having it or not is not really the scope of this PR

simone-silvestri:
This last commit should have allowed the possibility of choosing a general default without needing to use a field internal to `DefaultBoundaryCondition`.
The UI remains the same, i.e. on main and on this PR
```julia
bcs = FieldBoundaryConditions(FluxBoundaryCondition(10))
```
will have the same effect

PR #4765, opened 2025-09-04
Do not specify `JULIA_NUM_THREADS` or `JULIA_NUM_PRECOMPILE_TASKS` for Buildkite CI

ali-ramadhan:
Just seeing if this has an impact on Buildkite CI performance.


ali-ramadhan:
Seems to shave ~20 minutes off the total build time due to the two Enzyme tests, but otherwise no big difference. Removing `JULIA_NUM_THREADS` does not slow down the tests so I'm inclined to keep it for now. And I don't think we need `JULIA_NUM_PRECOMPILE_TASKS` as Julia should use as many threads as are available.

glwagner:
nice work

PR #4769, opened 2025-09-06
Enforce (temporarily) CUDA.jl < v5.8.3 for Buildkite CI docs

ali-ramadhan:
I see the docs are currently failing in Buildkite. I feel like there are a few issues with CUDA.jl right now (e.g. https://github.com/JuliaGPU/CUDA.jl/issues/2879) so this PR tries to fix things by using CUDA.jl v5.8.2. Hopefully this won't be needed once v5.8.4 is tagged and released.

Not sure why the main tests are okay and just the docs fail...

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -16,7 +16,7 @@ StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 TimesDates = "bdfc003b-8df8-5c39-adcd-3a9087f5df4a"
 
 [compat]
-CUDA = "5.4"
+CUDA = "< 5.8.3"
 CairoMakie = "0.11, 0.12, 0.13, 0.14"
 Documenter = "1"
 DocumenterCitations = "1"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -16,7 +16,7 @@ StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 TimesDates = "bdfc003b-8df8-5c39-adcd-3a9087f5df4a"
 
 [compat]
-CUDA = "< 5.8.3"
+CUDA = "5.4"
 CairoMakie = "0.11, 0.12, 0.13, 0.14"
 Documenter = "1"
 DocumenterCitations = "1"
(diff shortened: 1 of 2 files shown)


ali-ramadhan:
Looks like that did the trick!

navidcy:
hm... I'm bit confused since I could also seeing Buildkite passing with CUDA 5.8.3.

navidcy:
You referring to errors like, e.g., 

https://buildkite.com/clima/oceananigans/builds/25467#0199217d-7ad9-4917-9c71-414f31e223aa/17-416

?

navidcy:
Seems like all rest of the CI are using CUDA 5.8.2?

Here's a log _from a different PR_! Look at:

https://buildkite.com/clima/oceananigans/builds/25467#0199217d-7ad2-4952-ab7e-f7710ff8a62f/17-51

navidcy (approved):

navidcy:
OK, I see. CUDA.jl v5.8.3 can't be used yet in conjunction with AMDGPU.jl v2; x-ref: https://github.com/JuliaGPU/AMDGPU.jl/commit/22edc31c7ae8e3b31166c645341f51af3eb7b6c4#commitcomment-165302092

All CI does uses CUDA v5.8.3 so we might as well force the docs to do so. CUDA v5.9 would hopefully smooth this issue.

navidcy:
@ali-ramadhan, I'm merging this when CI passes.

docs/Project.toml
```diff
@@ -16,7 +16,7 @@ StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 TimesDates = "bdfc003b-8df8-5c39-adcd-3a9087f5df4a"
 
 [compat]
-CUDA = "5.4"
+CUDA = "< 5.8.3"
```
giordano, line 19: This is wrong, it should be
```suggestion
CUDA = "5.4 - 5.8.2"
```

navidcy: fair point

does it really matter for the docs/Project? the package itself imposes its proper compat entries (with both upper and lower bounds)

giordano: Didn't notice it was the docs environment, it's less critical there (but it wouldn't have been accepted in the registry if it was a dependency in the main environment), but good to be precise with compat bounds anyway.

ali-ramadhan: Thanks for pointing this out @giordano! Will definitely more careful about updating bounds in the future.

PR #4772, opened 2025-09-08
(0.98.1) Make `convert_for_netcdf()` accesible to users

tomchor:
The function `convert_for_netcdf()` (which converts objects for use in NetCDFs as attributes) was used only in the `OceananigansNCDatasets.jl` extension to write grid-reconstruction arguments to NetCDF, but I think it would be nice to make it possible for users to use this interface. It can be useful to write all sorts of objects to NetCDF, and this function makes it easy to do it. It requires minimal code changes and no extra maintenance on our part.


glwagner:
very nice to see NetCDF functionality becoming first-class!

PR #4782, opened 2025-09-12
(0.99.2) Add XESMF.jl extension to use xESMF to compute tracer regridding weights

taimoorsohail:
I have created a wrapper for xESMF (python package) so we can generally regrid any field from source to destination. At the moment, the PR doesn't include vertical regrinding (i.e., source and destinations vertical grids need to be the same).

Note: 
1) There are some speed ups that could be achieved in the actual matrix multiplication step.
2) I know an extension might be better but perhaps we can work on that together - I am just opening a PR so the collaboration can begin!

### DIFF
--- a/test/test_xesmf.jl
+++ b/test/test_xesmf.jl
@@ -35,7 +35,7 @@ tg = TripolarGrid(; size=(360, 170, 1), z, southernmost_latitude, radius)
         width = 12         # degrees
         set!(src_field, (λ, φ, z) -> exp(-((λ - λ₀)^2 + (φ - φ₀)^2) / 2width^2))
 
-        R = Oceananigans.Fields.Regridder(dst_field, src_field)
+        R = XESMF.Regridder(dst_field, src_field)
         @test R.weights isa SparseMatrixCSC
 
         regrid!(dst_field, R, src_field)


glwagner:
Great! A few comments off the bat:

1. We should implement some of this functionality in an extension so that we don't add a dependency on `PyCall`
2. I believe we also want to use `PythonCall` rather than `PyCall`: https://github.com/JuliaPy/PythonCall.jl. It appears to be actively developed (unlike `PyCall`) which is why ClimaOcean also uses `PythonCall`.

I am happy to implement these changes if that's ok @taimoorsohail ?

glwagner:
Ah and another thing --- `regrid!` is defined in the `Fields` module rather than the `Diagnostics` module. The idea is that this concept is purely associated with `Field`, similar to `interpolate!`. The `Diagnostics` module is actually model specific; eg it is a place for things that compute quantities that depend on the type of `AbstractModel` being used, such as the CFL constraint

taimoorsohail:
Yep @glwagner very happy for you to implement. Note that I did switch to PythonCall at some point and then I reverted back. I can't remember why exactly, it should be totally fine though...

taimoorsohail:
By the way, xESMF requires calling MPI to work, as far as I could tell. That means that if you are using the regridder you need to call `using MPI` and `MPI.init()` even if you don't actually use MPI in the code. Perhaps there's a way to bypass this.

ext/OceananigansPythonCallExt.jl
```diff
@@ -0,0 +1,171 @@
...
+    lon_b = Array(grid.λᶠᵃᵃ[1:grid.Nx+1])
+
+    lat,   lon   = two_dimensionalize(lat,   lon)
+    lat_b, lon_b = two_dimensionalize(lat_b, lon_b)
+
+    return structured_coordinate_dataset(lat, lon, lat_b, lon_b)
+end
+
+function coordinate_dataset(grid::SomeTripolarGrid)
+    lat = Array(grid.φᶜᶜᵃ[1:grid.Nx, 1:grid.Ny])
```
glwagner: Here, the preferred syntax is

```julia
lon, lat, z = nodes(grid, Center(), Center(), Center())
```

navidcy:
> Great! A few comments off the bat:
> 
> 1. We should implement some of this functionality in an extension so that we don't add a dependency on `PyCall`
> 2. I believe we also want to use `PythonCall` rather than `PyCall`: https://github.com/JuliaPy/PythonCall.jl. It appears to be actively developed (unlike `PyCall`) which is why ClimaOcean also uses `PythonCall`.
> 
> I am happy to implement these changes if that's ok @taimoorsohail ?

I started working on transitioning to an extension

glwagner:
> > Great! A few comments off the bat:
> > 
> > 1. We should implement some of this functionality in an extension so that we don't add a dependency on `PyCall`
> > 2. I believe we also want to use `PythonCall` rather than `PyCall`: https://github.com/JuliaPy/PythonCall.jl. It appears to be actively developed (unlike `PyCall`) which is why ClimaOcean also uses `PythonCall`.
> > 
> > I am happy to implement these changes if that's ok @taimoorsohail ?
> 
> I started working on transitioning to an extension

We are committing on top of each other, so lets coordinate

ext/OceananigansXESMFExt.jl
```diff
@@ -0,0 +1,96 @@
...
+        pyimport(name)
+    catch
+        add_package(name, channel)
+        pyimport(name)
+    end
+
+    return pkg
+end
+
+function regridding_weights(dst_field, src_field; method="conservative")
```
navidcy:
```suggestion

"""
    regridding_weights(dst_field, src_field; method="conservative")

Return a SparseMatrix with the weights to regrid `src_field` to the `dst_field`.
The regrinding weights are obtained via xESMF Python package using the
specified regridding method.

The `method` keyword argument determines one of the available regrinding methods
 that are exposed in xESMF from the ESMF library. Those include:

- `"bilinear"`: ESMF.RegridMethod.BILINEAR
- `"conservative"`: ESMF.RegridMethod.CONSERVE
- `"conservative_normed"`: ESMF.RegridMethod.CONSERVE
- `"patch"`: ESMF.RegridMethod.PATCH
- `"nearest_s2d"`: ESMF.RegridMethod.NEAREST_STOD
- `"nearest_d2s"`: ESMF.RegridMethod.NEAREST_DTOS

where `conservative_normed` is just the conservative method with the normalization set
to `ESMF.NormType.FRACAREA` instead of the default `norm_type=ESMF.NormType.DSTAREA`.

For more information, refer to the xESMF documentation at:
> https://xesmf.readthedocs.io/en/latest/notebooks/Compare_algorithms.html
"""
function regridding_weights(dst_field, src_field; method="conservative")
```

ext/OceananigansPythonCallExt.jl
```diff
@@ -0,0 +1,96 @@
...
+
+    # Extract center coordinates from both fields
+    λd = λnodes(dst_field)
+    φd = φnodes(dst_field)
+    λs = λnodes(src_field)
+    φs = φnodes(src_field)
+
+    # Extract boundary coordinates
+    
+
```
navidcy:
```suggestion
    
```

ext/OceananigansPythonCallExt.jl
```diff
@@ -0,0 +1,96 @@
...
+import Oceananigans.Grids: λnodes, φnodes, Center, Face
+import Oceananigans.Architectures: on_architecture, CPU
+
+"""
+    install_xesmf()
+
+Install the xesmf package using CondaPkg.
+Returns a NamedTuple containing package information if successful.
+"""
+function add_package(name, channel="conda-forge"; verbose=true)
```
navidcy: docstring needs updating

ext/OceananigansXESMFExt.jl
glwagner: the "bilinear" method is just interpolation right?

glwagner: that is interesting because we have tried to distinguish between interpolation and regridding, but this interface provides both. Maybe we will need a new name for the function

ext/OceananigansPythonCallExt.jl
```diff
@@ -0,0 +1,96 @@
...
+        λs_b = repeat(λs_b', size(φs_b, 1))
+        φs_b = repeat(φs_b, 1, size(λs_b, 2))
+    end
+
+    # Create coordinate datasets
+    src_ds = structured_coordinate_dataset(φs, λs, φs_b, λs_b, numpy, xarray)
+    dst_ds = structured_coordinate_dataset(φd, λd, φd_b, λd_b, numpy, xarray)
+
+    # Create regridder and extract weights
+    regridder = xesmf.Regridder(src_ds, dst_ds, method, periodic=pytrue)
```
simone-silvestri: the periodic flag should depend on the x-topology of the source field I think

src/Diagnostics/regridder.jl
```diff
@@ -0,0 +1,188 @@
...
+
+    function coordinate_dataset(grid::SomeTripolarGrid)
+        lat = Array(grid.φᶜᶜᵃ[1:grid.Nx, 1:grid.Ny])
+        lon = Array(grid.λᶜᶜᵃ[1:grid.Nx, 1:grid.Ny])
+
+        lat_b = Array(grid.φᶠᶠᵃ[1:grid.Nx+1, 1:grid.Ny+1])
+        lon_b = Array(grid.λᶠᶠᵃ[1:grid.Nx+1, 1:grid.Ny+1])
+
+        return structured_coordinate_dataset(lat, lon, lat_b, lon_b)
+    end
```
simone-silvestri:
```suggestion
```
There is no need to have two methods for this function, the one that you implemetented for `SomeLatitudeLongitudeGrid` would also work for `SomeTripolarGrid`

src/Diagnostics/regridder.jl
```diff
@@ -0,0 +1,188 @@
...
+
+        return lat, lon
+    end
+
+    function coordinate_dataset(grid::SomeLatitudeLongitudeGrid)
+        lat = Array(φnodes(grid, Center(), Center(), Center()))
+        lon = Array(λnodes(grid, Center(), Center(), Center()))
+
+        lat_b = Array(φnodes(grid, Face(), Face(), Center()))
+        lon_b = Array(grid.λᶠᵃᵃ[1:grid.Nx+1])
```
simone-silvestri:
```suggestion
        lon_b = Array(λnodes(grid, Face(), Face(), Center()))
```

src/Diagnostics/regridder.jl
```diff
@@ -0,0 +1,188 @@
...
+        return W
+    end
+
+    """
+        regrid_tracers!(src::Field, dst::Field, W::SparseMatrixCSC)
+    Regrid the `src` field onto the `dst` field using the provided weights `W`.
+    The function assumes that the vertical grid (z) of both fields is the same.
+    """
+    
+    function regrid_tracers!(dst::Field, W::SparseMatrixCSC, src::Field)
```
simone-silvestri: I would change this by providing a `Regridder` object that holds two sparse matrices (one in both directions) and two work arrays with contiguous memory

ext/OceananigansPythonCallExt.jl
glwagner: dont worry about these details just yet

src/Diagnostics/regridder.jl
glwagner: i was hoping to avoid dispatch altogether

src/Diagnostics/regridder.jl
glwagner: this will be deleted

ext/OceananigansXESMFExt.jl
navidcy: yes, I guess I've been a bit confused about what's regrid and what's interpolation...

ext/OceananigansPythonCallExt.jl
navidcy: yes!

glwagner:
@simone-silvestri @navidcy just to clarify here --- we will delete the code in `Diagnostics`; everything will be implemented in the PythonCall extension. Please focus attention there.

I just sent some commits. I think we can use the pure numpy interface documented here:

https://xesmf.readthedocs.io/en/stable/notebooks/Pure_numpy.html

Here's the code I'm trying to get to work:

```julia
using Oceananigans, PythonCall, CondaPkg

tg = TripolarGrid(size=(360, 170, 1), z=(0, 1))
ll = LatitudeLongitudeGrid(size=(360, 180, 1), longitude=(0, 360), latitude=(-90, 90), z=(0, 1))

ctg = CenterField(tg)
cll = CenterField(ll)

W = Oceananigans.Fields.regridding_weights(cll, ctg)
```

glwagner:
Do any of you understand why we need to provide the coordinates of the center of the cells? For ConservativeRegridding we needed only to provide the cell vertices. I don't understand what additional information the cell centers provide.

glwagner:
Also... to regrid velocities, we actually need to allocate memory because we need to interpolate the velocities to cell centers, rotate, and regrid the rotated fields. Right? Perhaps this PR can implement regridding for tracers only.

glwagner:
```julia
julia> include("mwe.jl")
64800×61200 SparseArrays.SparseMatrixCSC{Float64, Int64} with 373646 stored entries:
⎡⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣿⣷⡄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎤
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣿⣿⣿⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣿⣿⣿⣿⣿⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣿⣿⣿⣿⣿⣿⣿⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣾⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣦⣤⣤⣤⣤⣤⣤⣤⣤⣤⣶⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢹⠋⠁⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣿⣿⣿⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡟⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣿⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣿⣿⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠘⢿⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠙⢿⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⣿⣿⣆⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⣿⣿⣿⣷⣄⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⣿⣿⣿⣿⣿⣷⣄⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣷⣄⠀⎥
⎢⣤⣤⣤⣤⣤⣤⣤⣤⣴⣶⣾⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⣿⡿⠿⠿⠿⠿⠿⠿⠿⠷⎥
⎢⠈⠻⣿⣿⣿⣿⣿⣿⣿⣿⣿⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠏⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠈⠻⣿⣿⣿⣿⣿⣿⣿⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠈⠻⣿⣿⣿⣿⣿⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠈⠻⣿⣿⣿⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠙⢿⣿⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠙⣧⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎣⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣦⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎦
```

glwagner:
cc @juliasloan25

navidcy:
> Also... to regrid velocities, we actually need to allocate memory because we need to interpolate the velocities to cell centers, rotate, and regrid the rotated fields. Right? Perhaps this PR can implement regridding for tracers only.

Let this PR be only for tracers. @taimoorsohail has been thinking about the velocity quite a bit. Personally I'm still not in place to wrap my head around it. It's bit trickier because you wanna conserve the flux there, not the total volume so multiply `u * area` and then there is the rotation also.

glwagner:
@taimoorsohail take a look and let me know what you think! I think we can remove the files in `Diagnostics` now. In a future PR, we will add an interface for regridding fields. I think we also need to implement our own `Regridder` object to persist these weights. We can design this object to work for tracers or velocities.

ext/OceananigansXESMFExt.jl
```diff
@@ -0,0 +1,121 @@
...
+y_node_array(x::AbstractMatrix, Nx, Ny) = x_node_array(x, Nx, Ny)
+y_vertex_array(x::AbstractMatrix, Nx, Ny) = x_vertex_array(x, Nx, Ny)
+
+function regridding_weights(dst_field, src_field; method="conservative")
+
+    ℓx, ℓy, ℓz = Oceananigans.Fields.instantiated_location(src_field)
+
+    # We only support regridding between centered fields.
+    @assert ℓx isa Center
+    @assert ℓy isa Center
```
navidcy: should assert that the `dst_field` is a CenterField, right?

glwagner: Hmm yes. Actually, I think we should require the fields at the same location.

glwagner: done

glwagner: I think error handling should happen at a higher level PS. So not really for this PR, but for the `regrid!` interface.

navidcy: but for this PR the error should happen here..?

vchuravy:
Is there a strong argument of using the Python interface, rather than directly using the C interface. I am very hesitant on Python dependencies.

navidcy:
Shall I try to add a test with your mwe?

navidcy:
> Is there a strong argument of using the Python interface, rather than directly using the C interface. I am very hesitant on Python dependencies.

The main argument (personally) is that I don't know C and don't know anyone that does. No other reason as far as I can tell!

simone-silvestri:
I guess we could try to adapt https://earthsystemmodeling.org/docs/release/ESMF_5_3_0/ESMC_crefdoc/node2.html

ext/OceananigansReactantExt/TimeSteppers.jl
navidcy, line 1: why do we see these changes from #4784 in this PR?

ext/OceananigansPythonCallExt.jl
```diff
@@ -0,0 +1,145 @@
+module OceananigansPythonCallExt
+
+using Oceananigans
+using PythonCall
+using CondaPkg
+using SparseArrays
+
+import Oceananigans.Fields: regridding_weights
+import Oceananigans.Grids: λnodes, φnodes, Center, Face
+import Oceananigans.Architectures: on_architecture, CPU
```
navidcy:
```suggestion
using PythonCall
using CondaPkg
using SparseArrays

using Oceananigans
using Oceananigans.Grids: λnodes, φnodes, Center, Face
using Oceananigans.Architectures: on_architecture, CPU

import Oceananigans.Fields: regridding_weights
```

ext/OceananigansXESMFExt.jl
glwagner: "regrid" in our terminology is conservative.

navidcy: gotcha

test/test_pythoncall_regridder.jl
navidcy, line 1: can somebody help me set up the tests? seems like they are not picking up the right conda env? I'm not sure... complains about numpy > 2 which probably comes from Reactant dependancy?

glwagner: @giordano do you have some advice here?

giordano: Python packaging isn't exactly my forte 🫠 

I honestly don't know how CondaPkg handles environments from other repos. I see Reactant has indeed the requirement on numpy v2 https://github.com/EnzymeAD/Reactant.jl/blob/6b9f46be505b35a03a04f2cd1bd29c77980d8431/CondaPkg.toml#L4, which honestly sounds like a reasonable idea nowadays, but I wouldn't have expected CondaPkg to pick up the requirements from there (but I may be completely wrong, I just don't know how this stuff works). However I think the more interesting question is "what's requiring a version of numpy different from 2?"

glwagner: Possibly the conda stuff gets shared between tests, so we inherit the restriction here? 

But also, why is that a problem? Looks liek the error is:

```
  Python: AttributeError: module 'numpy' has no attribute '__version__'
```

Wonder if this is PythonCall or CondaPkg issue

giordano: > complains about numpy > 2

I had read this message without actually looking at the error message (a direct link to it would have made my life easier), now that I see it https://github.com/CliMA/Oceananigans.jl/actions/runs/17686903205/job/50273566529?pr=4782#step:6:442
```
PythonCall extension: Error During Test at /home/runner/work/Oceananigans.jl/Oceananigans.jl/test/test_pythoncall_regridder.jl:6
  Python: AttributeError: module 'numpy' has no attribute '__version__'
  Python stacktrace:
   [1] <module>
     @ /tmp/jl_Ii8QfW/.CondaPkg/.pixi/envs/default/lib/python3.13/site-packages/pandas/compat/numpy/__init__.py:9
   [2] <module>
     @ /tmp/jl_Ii8QfW/.CondaPkg/.pixi/envs/default/lib/python3.13/site-packages/pandas/compat/__init__.py:26
   [3] <module>
     @ /tmp/jl_Ii8QfW/.CondaPkg/.pixi/envs/default/lib/python3.13/site-packages/pandas/__init__.py:26
   [4] <module>
     @ /tmp/jl_Ii8QfW/.CondaPkg/.pixi/envs/default/lib/python3.13/site-packages/xarray/coding/times.py:12
   [5] <module>
     @ /tmp/jl_Ii8QfW/.CondaPkg/.pixi/envs/default/lib/python3.13/site-packages/xarray/coders.py:6
   [6] <module>
     @ /tmp/jl_Ii8QfW/.CondaPkg/.pixi/envs/default/lib/python3.13/site-packages/xarray/__init__.py:3
   [7] <module>
     @ /tmp/jl_Ii8QfW/.CondaPkg/.pixi/envs/default/lib/python3.13/site-packages/xesmf/util.py:4
   [8] <module>
     @ /tmp/jl_Ii8QfW/.CondaPkg/.pixi/envs/default/lib/python3.13/site-packages/xesmf/__init__.py:3
```
I'm even more confused: the numpy package seems to be installed, but it doesn't have the version attribute? @cjdoris would you have an inkling of an idea of what's going on?

navidcy: Good point about the link. Sorry, my bad

glwagner: I guess the error originates from this line:

https://github.com/pandas-dev/pandas/blob/cc40732889b59d0ebd867b087691f02221e5666c/pandas/compat/numpy/__init__.py#L10

Project.toml
```diff
@@ -57,11 +59,13 @@ OceananigansMetalExt = "Metal"
 OceananigansNCDatasetsExt = "NCDatasets"
 OceananigansOneAPIExt = "oneAPI"
 OceananigansReactantExt = ["Reactant", "KernelAbstractions", "ConstructionBase"]
+OceananigansPythonCallExt = ["PythonCall", "CondaPkg"]
```
vchuravy: So this seems a bit weird to me. 

An extension is something that is triggered implicitly when someone adds both "PythonCall" and "CondaPkg". That doesn't imply that the person wants to use xESMF 

The right way of doing this is to create an xESMF/ESMF Julia package (which might ot might not depend on CondaPkg/PythonCall) and then add an Oceananigans extension that depends on that as a trigger package. 

This makes the users intention clear. They didn't just incidentally have PythonCall in their manifest, but they want to use the ESMF functionality

ext/OceananigansPythonCallExt.jl
```diff
@@ -0,0 +1,146 @@
...
+    return pkg
+end
+
+"""
+    add_import_pkg(package_name, channel="conda-forge")
+
+Import and return `package_name` with `PythonCall.pyimport`,
+installing it with `add_package` if it is not found.
+"""
+function add_import_pkg(name, channel="conda-forge")
```
cjdoris: This isn't how we install Conda packages inside Julia packages - declare them in a CondaPkg.toml file instead.

test/test_pythoncall_regridder.jl
cjdoris: Not clear what's going on - the CondaPkg output is clear that the numpy bound is found.

It might be a result of https://github.com/CliMA/Oceananigans.jl/pull/4782#discussion_r2345965182 - if you modify the python environment after PythonCall is started (which this does) then the state of imported python packages can quickly get messed up.

cjdoris:
This functionality seems to me like it belongs in a new package, not an extension.

Extensions are for providing compatibility between packages, which this doesn't. This adds a new function which has zero methods until you import PythonCall. This is a slight abuse of the extension mechanism, though I can see why it's nice to have the function part of the main Oceananigans API.

As @vchuravy mentioned, it would probably make more sense to make a separate trigger package for this, instead of triggering on PythonCall itself. But then if you're going to do that, you may as well put the functionality into this new package and not have an extension at all.

If you do stick with using an extension, you should put the Conda dependency into a CondaPkg.toml file at the root of the repo. One day it would be nice if CondaPkg could support dependencies in extensions that are only installed if the relevant weakdeps are installed, but it doesn't do that yet.

Project.toml
glwagner: That does seem like a better design. This advice also applies to ClimaOcean's Copernicus functionality:

https://github.com/CliMA/ClimaOcean.jl/blob/main/ext/ClimaOceanPythonCallExt.jl

My main concern, which may not matter, is that this creates a lot more work in the event that we add functionality for a large number of python packages. We will then need to generate wrappers for every package that we want to use in our workflow? Perhaps that is unavoidable but I wanted to bring it up as a downside.

glwagner:
> This functionality seems to me like it belongs in a new package, not an extension.
> 
> Extensions are for providing compatibility between packages, which this doesn't. This adds a new function which has zero methods until you import PythonCall. This is a slight abuse of the extension mechanism, though I can see why it's nice to have the function part of the main Oceananigans API.
> 
> As @vchuravy mentioned, it would probably make more sense to make a separate trigger package for this, instead of triggering on PythonCall itself. But then if you're going to do that, you may as well put the functionality into this new package and not have an extension at all.

Okay, I understand. I think we will still need an extension at some point, because there is an Oceananigans-defined function `regrid!` which we would like to embed with functionality from xESMF. So I think this workflow requires two steps:

1) defining a new Julia package that uses PythonCall to provide `xESMF` functionality in Julia, which I think also will require defining a user interface into that package, tests for it, etc.
2) Building an extension in Oceananigans that depends on the new Julia wrapper package which is triggered eg if one writes `using xESMF` 

This makes sense to me and definitely sounds cleaner, my only concern is that it is more work. If we only need xESMF its ok, my main concern is that there will probably be more (possibly, quite a few more). Is there any way to streamline this / provide a different design so that it does not always require a Julia wrapper package? I do feel like it will raise the bar for providing seamless interface into python packages (ie it is not just a PR but rather a new package with tests, etc that will be required each time) but perhaps this is worth it / the only way to provide it cleanly.

> If you do stick with using an extension, you should put the Conda dependency into a CondaPkg.toml file at the root of the repo. One day it would be nice if CondaPkg could support dependencies in extensions that are only installed if the relevant weakdeps are installed, but it doesn't do that yet.

If I understand correctly --- if we write a wrapper package `xESMF.jl`, we will not need `CondaPkg.toml` here anymore, right?

glwagner:
I am giving this a shot here: https://github.com/glwagner/xESMF.jl

glwagner:
To answer my own question... one possible alternative design is to create a package that's something like `OceananigansPythonInterface.jl` or `PythonForOceananigans.jl`. Then this package could "collect" all of the python packages that might provide useful extension capabilities to Oceananigans. Then we have just one package for all python dependencies, plus one extension in Oceananigans; users write

```julia
using Oceananigans
using OceananigansPythonInterface
```

to access it.

For the time being I have added a wrapper for just `xESMF`. But I think we may want to take a different approach with ClimaOcean stuff, because ClimaOcean has a far broader scope and I think there will be quite a few python packages we may want to use.

cc @cjdoris @vchuravy

Project.toml
navidcy: I do like the design. It wasn't feeling quite right to have xESMF extension triggered by PyCall.

glwagner:
🎉  Ok, we managed to get a new package `XESMF.jl` registered in the Julia registry: https://github.com/JuliaRegistries/General/pull/138592#issuecomment-3314972935

Reflecting on this experience... getting this merged required a bit of back-and-forth about the package name. This seems like a generic problem that will come up for wrappers around python packages. So I'd like to highlight the downsides of always registering python extensions as independent packages --- which I guess is mainly an issue when said python package doesn't have much of an _independent_ use case.

I am wondering, is it possible there will be a way to trigger extensions that is not explicitly tied to the registry? This would allow opting into extended functionality (which may depend on PythonCall, CondaPkg, etc) without having to crowd the registry with packages who's only purpose is auxiliary to some primary package like Oceananigans.

vchuravy:
Unfortunately no, package extension are tightly coupled to packages so that mechanism won't work (and your only option would be an extension triggering on PythonCall which is way to broad)

One can guard functionality behind a preference (but why?) and include it in the main package, but then Oceananigans has a hard dependency on Python. 

The bigger question is a philosophical one, if you only ever intend to use a small fraction of a Python package it does feel like bad form to "block" the name. 

Personally, if I am only interested in a small piece of functionality I would first look at rewriting the code (LLM are perfect for this)

That's also why I was questioning the need to go through Python at all, if there is a C library that one could use. 

(My 2cents is that I don't want to deal with Python packaging in an HPC context)

navidcy:
OK, the XESMF.jl extension is successfully loaded now and the test_xesmf.jl test suit runs.
But we need some actual test cases.
@taimoorsohail could you post here the example(s) you had going on in ocean-ensembles?

navidcy:
I added a non-trivial test that conservatively regrids a field of ones from a lat-lon grid to a tripolar.

https://github.com/CliMA/Oceananigans.jl/blob/f06db719c6ece3e3ad3362dcd52c70f813555023/test/test_xesmf.jl#L8-L39

I chose `southernmost_latitude = -30`. This means that the Integral of the original field should be 3πR² since

$$
\int_0^{2π} dλ \int_{-π/6}^{π/2} dφ R^2 \cos φ = 3πR^2
$$

However, the tripolar grid doesn't have its southernmost boundary at a line of constant latitude at -30 degrees. Thus the two grids don't span exactly the same area and therefore it's hard to check whether the conservative regrinding happens OK by checking whether `∫ dst_field dA == ∫ src_field dA`. See error at https://github.com/CliMA/Oceananigans.jl/actions/runs/17890374414/job/50869929175?pr=4782#step:6:420

Any ideas how to have a simple, non-trivial test?

test/test_xesmf.jl
```diff
@@ -0,0 +1,57 @@
...
+        # ∫ src_field dA = 3πR²
+        @test Field(Integral(src_field, dims=(1, 2)))[1, 1, 1] ≈ 3π * radius^2
+
+        W = Oceananigans.Fields.regridding_weights(dst_field, src_field)
+        @test W isa SparseMatrixCSC
+
+        regrid_conservatively!(dst_field, W, src_field)
+
+        # ∫ dst_field dA = ∫ src_field dA
+        @test Field(Integral(dst_field, dims=(1, 2)))[1, 1, 1] ≈ Field(Integral(src_field, dims=(1, 2)))[1, 1, 1]
```
glwagner:
```suggestion
        @test first(Field(Integral(dst_field, dims=(1, 2)))) ≈ first(Field(Integral(src_field, dims=(1, 2))))
```

taimoorsohail:
Here's the test I was using @navidcy:

```Julia
using Oceananigans
using ConservativeRegridding

coarse_grid = TripolarGrid(CPU();
                    size = (360,180, 1),
                    z =  (0,1),
                    halo = (6, 6, 3),
                    first_pole_longitude = 70,
                    north_poles_latitude = 55)
fine_grid   = LatitudeLongitudeGrid(size=(360, 180, 1), longitude=(0, 360), latitude=(-90, 90), z=(0, 1))

dst = Field{Center, Center, Nothing}(fine_grid)
src = Field{Center, Center, Nothing}(coarse_grid)

regridder = ConservativeRegridding.Regridder(dst, src)

set!(src, (x, y) -> rand())
new_vals = reshape(ConservativeRegridding.regrid!(dst, regridder, src), 360,180)

@show mean(dst)
@show mean(src)
```

navidcy:
Were you getting machine precision agreement?

ext/OceananigansXESMFExt.jl
```diff
@@ -0,0 +1,113 @@
+module OceananigansXESMFExt
+
+using XESMF
+using PythonCall
```
simone-silvestri:
```suggestion
```

ext/OceananigansXESMFExt.jl
```diff
@@ -0,0 +1,113 @@
...
+    ℓx, ℓy, ℓz = Oceananigans.Fields.instantiated_location(src_field)
+
+    # We only support regridding between centered fields
+    @assert ℓx isa Center
+    @assert ℓy isa Center
+    @assert (ℓx, ℓy, ℓz) == Oceananigans.Fields.instantiated_location(dst_field)
+
+    dst_coordinates, src_coordinates = extract_xesmf_coordinates_structure(dst_field, src_field)
+
+    periodic = Oceananigans.Grids.topology(src_field.grid, 1) === Periodic ? PythonCall.pybuiltins.True : pybuiltins.False
```
simone-silvestri: This probably should go in XESMF.jl?

test/test_xesmf.jl
```diff
@@ -0,0 +1,57 @@
...
+
+using XESMF
+using SparseArrays
+using LinearAlgebra
+
+function regrid_conservatively!(dst_field, weights, src_field)
+    @assert src_field.grid.Nz == dst_field.grid.Nz
+
+    for k in 1:src_field.grid.Nz
+        LinearAlgebra.mul!(vec(interior(dst_field, :, :, k)), weights, vec(interior(src_field, :, :, k)))
```
simone-silvestri: for the test this might be ok, but as a reminder: it is > 1000x more expensive to do this rather than having contiguous memory, for example
```julia
wrk .= vec(interior(src_field, :, :, k) # wrk is a previously allocated array
LinearAlgebra.mul!(wrk2, weights, wrk)
vec(interior(src_field, :, :, k)) .= wrk2 # wrk2 is a previously allocated array
```

simone-silvestri:
> Any ideas how to have a simple, non-trivial test?

You can regrid from tripolar to lat-lon with the lat-lon having a southernmost boundary that encompasses the tripolar grid i.e. if for the tripolar grid southernmost_latitude = -80, then you can use a lat-lon that goes from -85 to 90

ext/OceananigansXESMFExt.jl
```diff
@@ -9,13 +9,21 @@ using Oceananigans.Architectures: on_architecture, CPU
 
 import Oceananigans.Fields: regridding_weights
 
-x_node_array(x::AbstractVector, Nx, Ny) = view(x, 1:Nx) |> Array
-y_node_array(x::AbstractVector, Nx, Ny) = view(x, 1:Ny) |> Array
-x_node_array(x::AbstractMatrix, Nx, Ny) = view(x, 1:Nx, 1:Ny) |> Array
+function x_node_array(x::AbstractVector, Nx, Ny)
```
simone-silvestri, line 12:
```suggestion
function x_node_matrix(x::AbstractVector, Nx, Ny)
```
I guess this is really a matrix rather than an array then

test/test_xesmf.jl
navidcy: Gotcha! That's probably part of the slowness I'm observing.....! Thanks!!

ext/OceananigansXESMFExt.jl
navidcy: I'm rewriting those because they were wrong...

test/test_xesmf.jl
navidcy: for the case I was testing it was actually 40000x slower!

ext/OceananigansXESMFExt.jl
navidcy: hm...  perhaps I missed your point
yes, all the coords need to be 2D arrays before we give them to xesmf.Regridder
but what do you mean "matrix" instead of "array"?
I guess matrix is more precise since Array can be 1D, 2D, 3D, ... Is this what you were referring to?

test/test_xesmf.jl
glwagner: We need to build a `Regridder` object. Inside that, we can allocate scratch memory.

navidcy:
Hm... I'm a bit concerned with how conservative is the conservative... The integrals don't really match???

here's a MWE

```Julia
using Oceananigans, XESMF, LinearAlgebra

function regrid_with_xesmf!(dst_field, weights, src_field)
    @assert src_field.grid.Nz == dst_field.grid.Nz

    for k in 1:src_field.grid.Nz
        LinearAlgebra.mul!(vec(interior(dst_field, :, :, k)), weights, vec(interior(src_field, :, :, k)))
    end

    return nothing
end

z = (-1, 0)
southernmost_latitude = -80
radius = Oceananigans.Grids.R_Earth

llg = LatitudeLongitudeGrid(; size=(360, 180, 1),
                            longitude=(0, 360),
                            latitude=(southernmost_latitude, 90),
                            z, radius)

tg = TripolarGrid(; size=(360, 170, 1), z, southernmost_latitude, radius)

src_grid, dst_grid = tg, llg

src_field = CenterField(src_grid)
dst_field = CenterField(dst_grid)

λ₀, φ₀ = 150, 30.  # degrees
width = 12         # degrees
set!(src_field, (λ, φ, z) -> exp(-((λ - λ₀)^2 + (φ - φ₀)^2) / 2width^2))

W_conservative = Oceananigans.Fields.regridding_weights(dst_field, src_field, method="conservative")
W_bilinear = Oceananigans.Fields.regridding_weights(dst_field, src_field, method="bilinear")

∫src = first(Field(Integral(src_field, dims=(1, 2))))

regrid_with_xesmf!(dst_field, W_conservative, src_field)

∫dst_conservative = first(Field(Integral(dst_field, dims=(1, 2))))

regrid_with_xesmf!(dst_field, W_bilinear, src_field)

∫dst_bilinear = first(Field(Integral(dst_field, dims=(1, 2))))

@info "integral of src field: ∫src = $∫src"
@info "integral of dst field with conservative regridding: ∫dst_conservative = $∫dst_conservative"
@info "integral of dst field with bilinear regridding: ∫dst_bilinear = $∫dst_bilinear"
```

```Julia
[ Info: integral of src field: ∫src = 9.478054020020943e12

[ Info: integral of dst field with conservative regridding: ∫dst_conservative = 9.477993881980473e12

[ Info: integral of dst field with bilinear regridding: ∫dst_bilinear = 9.477704845568057e12
```

test/test_xesmf.jl
navidcy: Yes. Exactly my thought yesterday. Will do it!

ext/OceananigansXESMFExt.jl
simone-silvestri: ok, I guess array is fine.

simone-silvestri:
These grids 
```julia
llg = LatitudeLongitudeGrid(; size=(360, 180, 1),
                            longitude=(0, 360),
                            latitude=(southernmost_latitude, 90),
                            z, radius)

tg = TripolarGrid(; size=(360, 170, 1), z, southernmost_latitude, radius)
```
are not completely aligned because the southernmost latitude in a tripolar grid is approximate, in practice the first latitude will oscillate between
```julia
southernmost_latitude - ϵ, southernmost_latitude + ϵ
```
where `ϵ` depends on the grid size. You should try with a lat-lon grid that completely encompasses a tripolar one, something like
```julia
llg = LatitudeLongitudeGrid(; size=(360, 180, 1),
                            longitude=(0, 360),
                            latitude=(southernmost_latitude - 2, 90),
                            z, radius)

tg = TripolarGrid(; size=(360, 170, 1), z, southernmost_latitude, radius)
```

ext/OceananigansXESMFExt.jl
navidcy: done

ext/OceananigansXESMFExt.jl
navidcy: But then XESMF needs to depend on Oceananigans (to know about grid and topologies?)

test/test_xesmf.jl
navidcy: Done

navidcy: https://github.com/CliMA/Oceananigans.jl/blob/f2c77353a099e5478615e03e894dad8e2a0a3933/src/Fields/regridding_fields.jl#L34-L81

navidcy: The for loop in k could be done asynchronously? launch a kernel? @glwagner/@simone-silvestri feel free to suggest

src/Fields/regridding_fields.jl
```diff
@@ -2,17 +2,88 @@ using KernelAbstractions: @kernel, @index
...
 
+struct Regridder{A, S, V, G1, G2, M}
+    architecture :: A
+    method :: M
+    weights :: S
+    src_grid :: G1
+    dst_grid :: G2
+    src_temp :: V
+    dst_temp :: V
+end
```
simone-silvestri: I would have the `Regridder` struct only contain info for regridding, for example:
```suggestion
struct Regridder{S, V, V1, V2, M}
    method :: M
    weights :: S
    src_temp :: V1
    dst_temp :: V2
end
```
This will allow us to use it more generally, even for other grid types that are not oceananigans native.

simone-silvestri: note that the architecture is easily inferrable from the`temp` arrays, also we might need it to work for heterogeneous grids (for example one on the cpu one on the gpu) for which a single architecture is not enough

src/Fields/regridding_fields.jl
```diff
@@ -2,17 +2,88 @@ using KernelAbstractions: @kernel, @index
...
+Base.summary(r::Regridder{A, S, V, G1, G2, M}) where {A, S, V, G1, G2, M} = "$(r.method) Regridder on $A"
+
+function Base.show(io::IO, r::Regridder)
+    print(io, summary(r), '\n')
+    print(io, "├── src_grid: ", summary(r.src_grid), '\n')
+    print(io, "├── dst_grid: ", summary(r.dst_grid), '\n')
+    print(io, "├── weights: ", summary(r.weights), '\n')
+    print(io, "├── src_temp: ", summary(r.src_temp), '\n')
+    print(io, "└── dst_temp: ", summary(r.dst_temp))
+end
```
simone-silvestri:
```suggestion
function Base.show(io::IO, r::Regridder)
    print(io, summary(r), '\n')
    print(io, "├── weights: ", summary(r.weights), '\n')
    print(io, "├── src_temp: ", summary(r.src_temp), '\n')
    print(io, "└── dst_temp: ", summary(r.dst_temp))
end
```

src/Fields/regridding_fields.jl
```diff
@@ -2,17 +2,88 @@ using KernelAbstractions: @kernel, @index
...
+```
+"""
+function regrid!(dst_field, r::Regridder, src_field)
+    Nz = size(src_field.grid)[3]
+    topo_z = topology(src_field)[3]()
+    ℓz = location(src_field)[3]()
+
+    dst_temp, W, src_temp = r.dst_temp, r.weights, r.src_temp
+
+    for k in 1:total_length(ℓz, topo_z, Nz)
```
simone-silvestri:
```suggestion
    for k in 1:size(grid, 3)
```

src/Fields/regridding_fields.jl
```diff
@@ -2,17 +2,88 @@ using KernelAbstractions: @kernel, @index
...
+regrid!(dst_field, regridder, src_field)
+
+first(Field(Integral(dst_field, dims=(1, 2))))
+```
+"""
+function regrid!(dst_field, r::Regridder, src_field)
+    Nz = size(src_field.grid)[3]
+    topo_z = topology(src_field)[3]()
+    ℓz = location(src_field)[3]()
+
```
simone-silvestri:
```suggestion
```

ext/OceananigansXESMFExt.jl
```diff
@@ -0,0 +1,159 @@
...
+
+    xesmf = XESMF.xesmf
+    regridder = xesmf.Regridder(src_coordinates, dst_coordinates, method; periodic)
+    method = uppercasefirst(string(regridder.method))
+    weights = XESMF.sparse_regridder_weights(regridder)
+
+    arch = architecture(src_field)
+    FT = eltype(src_field)
+
+    weights = on_architecture(arch, weights)
```
simone-silvestri, line 139: I think we removed the `on_architecture` method for sparse matrices when removeing the `HeptadiagonalIterativeSolver` so I am not sure this will work?

src/Fields/regridding_fields.jl
simone-silvestri: also, is not better to have this struct in `XESMF.jl` so we can reuse  it easily in ClimaOcean? The constructor for oceananigans fields can remain here.

ext/OceananigansXESMFExt.jl
```diff
@@ -0,0 +1,159 @@
...
+llg = LatitudeLongitudeGrid(; size=(360, 180, 1), z,
+                            longitude=(0, 360), latitude=(-82, 90))
+
+src_field = CenterField(tg)
+dst_field = CenterField(llg)
+
+regridder = Oceananigans.Fields.Regridder(dst_field, src_field, method="conservative")
+```
+"""
+function Regridder(dst_field, src_field; method="conservative")
```
simone-silvestri:
```suggestion
function Regridder(dst_field::AbstractField, src_field::AbstractField; method="conservative")
```

ext/OceananigansXESMFExt.jl
```diff
@@ -0,0 +1,159 @@
...
+end
+function y_vertex_array(x::AbstractVector, Nx, Ny)
+    return Array(repeat(view(x, 1:Ny+1)', Nx+1, 1))'
+end
+x_vertex_array(x::AbstractMatrix, Nx, Ny) = Array(view(x, 1:Nx+1, 1:Ny+1))'
+
+y_node_array(x::AbstractMatrix, Nx, Ny) = x_node_array(x, Nx, Ny)
+y_vertex_array(x::AbstractMatrix, Nx, Ny) = x_vertex_array(x, Nx, Ny)
+
+function extract_xesmf_coordinates_structure(dst_field, src_field)
```
simone-silvestri: I think should be a general function of the `XESMF.jl` package, imported here and extended for the case of 

```suggestion
function extract_xesmf_coordinates_structure(dst_field::AbstractField, src_field::AbstractField)
```

ext/OceananigansXESMFExt.jl
```diff
@@ -0,0 +1,159 @@
...
+
+using XESMF
+using CUDA
+using SparseArrays
+using Oceananigans
+using Oceananigans.Architectures: CPU, architecture
+using Oceananigans.Grids: λnodes, φnodes, Center, Face
+
+import Oceananigans.Architectures: on_architecture
+import Oceananigans.Fields: Regridder
```
simone-silvestri:
```suggestion
import XESMF: Regridder, extract_xesmf_coordinates_structure
```

src/Fields/regridding_fields.jl
navidcy: I like all these suggestions! I'll implement them.

ext/OceananigansXESMFExt.jl
navidcy: what PR was that? I remember it but can't find it!

src/Fields/regridding_fields.jl
navidcy: I think I've implemented everything.

ext/OceananigansXESMFExt.jl
simone-silvestri: It was this one https://github.com/CliMA/Oceananigans.jl/pull/4386 should I open a PR to readd the method?

navidcy: Yes. We'll need it, right?

simone-silvestri: This should do the trick #4815

ext/OceananigansXESMFExt.jl
```diff
@@ -129,26 +130,68 @@ function Regridder(dst_field, src_field; method="conservative")
...
-    Nx, Ny, Nz = size(dst_field)
-    temp_dst = on_architecture(arch, zeros(FT, Nx * Ny))
+    for k in 1:total_length(ℓz, topo_z, Nz)
+        src_temp .= vec(interior(src_field, :, :, k))
+        LinearAlgebra.mul!(dst_temp, W, src_temp)
+        vec(interior(dst_field, :, :, k)) .= dst_temp
+    end
 
-    return Regridder(arch, method, weights, src_field.grid, dst_field.grid, temp_src, temp_dst)
+    return nothing
```
simone-silvestri:
```suggestion
    return dst_field
```

ext/OceananigansXESMFExt.jl
```diff
@@ -129,26 +130,68 @@ function Regridder(dst_field, src_field; method="conservative")
...
+    ℓz = location(src_field)[3]()
+
+    dst_temp, W, src_temp = regrider.dst_temp, regrider.weights, regrider.src_temp
 
-    Nx, Ny, Nz = size(dst_field)
-    temp_dst = on_architecture(arch, zeros(FT, Nx * Ny))
+    for k in 1:total_length(ℓz, topo_z, Nz)
+        src_temp .= vec(interior(src_field, :, :, k))
+        LinearAlgebra.mul!(dst_temp, W, src_temp)
+        vec(interior(dst_field, :, :, k)) .= dst_temp
```
simone-silvestri: If we merge https://github.com/NumericalEarth/XESMF.jl/pull/16 this would become
```suggestion
        src = vec(interior(src_field, :, :, k))
        dst = vec(interior(dst_field, :, :, k))
        regridder(dst, src)
```

navidcy: without the `.=`?

navidcy:
@simone-silvestri , using https://github.com/NumericalEarth/XESMF.jl/pull/16 with the MWE below I get:

```Julia
using Oceananigans
using XESMF

tg = TripolarGrid(; size=(360, 170, 1), z = (-1, 0), southernmost_latitude = -80)
llg = LatitudeLongitudeGrid(; size=(360, 180, 1), z = (-1, 0),
                            longitude=(0, 360), latitude=(-82, 90))

src_field = CenterField(tg)
dst_field = CenterField(llg)

λ₀, φ₀ = 150, 30.  # degrees
width = 12         # degrees
set!(src_field, (λ, φ, z) -> exp(-((λ - λ₀)^2 + (φ - φ₀)^2) / 2width^2))

regridder = XESMF.Regridder(dst_field, src_field, method="conservative")

regrid!(dst_field, regridder, src_field)
```

```Julia
julia> using Oceananigans

julia> using XESMF

julia> tg = TripolarGrid(; size=(360, 170, 1), z = (-1, 0), southernmost_latitude = -80)
360×170×1 OrthogonalSphericalShellGrid{Float64, Periodic, RightConnected, Bounded} on CPU with 4×4×4 halo and with precomputed metrics
├── centered at (λ, φ) = (70.0, 1.8005)
├── longitude: Periodic  extent 360.156 degrees                        variably spaced with min(Δλ)=0.00798438, max(Δλ)=1.05373
├── latitude:  Oceananigans.Grids.RightConnected  extent 171.0 degrees variably spaced with min(Δφ)=0.0127345, max(Δφ)=1.00592
└── z:         Bounded  z ∈ [-1.0, 0.0]                                regularly spaced with Δz=1.0

julia> llg = LatitudeLongitudeGrid(; size=(360, 180, 1), z = (-1, 0),
                                   longitude=(0, 360), latitude=(-82, 90))
360×180×1 LatitudeLongitudeGrid{Float64, Periodic, Bounded, Bounded} on CPU with 3×3×1 halo and with precomputed metrics
├── longitude: Periodic λ ∈ [0.0, 360.0)  regularly spaced with Δλ=1.0
├── latitude:  Bounded  φ ∈ [-82.0, 90.0] regularly spaced with Δφ=0.955556
└── z:         Bounded  z ∈ [-1.0, 0.0]   regularly spaced with Δz=1.0

julia> src_field = CenterField(tg)
360×170×1 Field{Center, Center, Center} on OrthogonalSphericalShellGrid on CPU
├── grid: 360×170×1 OrthogonalSphericalShellGrid{Float64, Periodic, RightConnected, Bounded} on CPU with 4×4×4 halo and with precomputed metrics
│   └── west: Periodic, east: Periodic, south: ZeroFlux, north: Zipper(1.0), bottom: ZeroFlux, top: ZeroFlux, immersed: Nothing
└── data: 368×178×9 OffsetArray(::Array{Float64, 3}, -3:364, -3:174, -3:5) with eltype Float64 with indices -3:364×-3:174×-3:5

julia> dst_field = CenterField(llg)
360×180×1 Field{Center, Center, Center} on LatitudeLongitudeGrid on CPU
├── grid: 360×180×1 LatitudeLongitudeGrid{Float64, Periodic, Bounded, Bounded} on CPU with 3×3×1 halo and with precomputed metrics
│   └── west: Periodic, east: Periodic, south: ZeroFlux, north: Value, bottom: ZeroFlux, top: ZeroFlux, immersed: Nothing
└── data: 366×186×3 OffsetArray(::Array{Float64, 3}, -2:363, -2:183, 0:2) with eltype Float64 with indices -2:363×-2:183×0:2

julia> λ₀, φ₀ = 150, 30.  # degrees
(150, 30.0)

julia> width = 12         # degrees
12

julia> set!(src_field, (λ, φ, z) -> exp(-((λ - λ₀)^2 + (φ - φ₀)^2) / 2width^2))
360×170×1 Field{Center, Center, Center} on OrthogonalSphericalShellGrid on CPU
├── grid: 360×170×1 OrthogonalSphericalShellGrid{Float64, Periodic, RightConnected, Bounded} on CPU with 4×4×4 halo and with precomputed metrics
│   └── west: Periodic, east: Periodic, south: ZeroFlux, north: Zipper(1.0), bottom: ZeroFlux, top: ZeroFlux, immersed: Nothing
└── data: 368×178×9 OffsetArray(::Array{Float64, 3}, -3:364, -3:174, -3:5) with eltype Float64 with indices -3:364×-3:174×-3:5
    └── max=0.998852, min=3.66505e-85, mean=0.0130418

julia> regridder = XESMF.Regridder(dst_field, src_field, method="conservative")
Conservative Regridder
├── weights: 64800×61200 SparseArrays.SparseMatrixCSC{Float64, Int64} with 263516 stored entries
├── src_temp: 61200-element Vector{Float64}
└── dst_temp: 64800-element Vector{Float64}

julia> regrid!(dst_field, regridder, src_field)
ERROR: UndefVarError: `regridder` not defined
Stacktrace:
 [1] regrid!(dst_field::Field{…}, regrider::XESMF.Regridder{…}, src_field::Field{…})
   @ OceananigansXESMFExt ~/Library/CloudStorage/OneDrive-TheUniversityofMelbourne/Documents/Research/Oceananigans.jl-v4/ext/OceananigansXESMFExt.jl:190
 [2] top-level scope
   @ REPL[34]:1
```

navidcy:
oh I have a typo... nevermind

simone-silvestri:
I get this:
```julia
julia> regridder = XESMF.Regridder(dst_field, src_field, method="conservative")
conservative Regridder
├── weights: 64800×61200 SparseArrays.SparseMatrixCSC{Float64, Int64} with 263516 stored entries
├── src_temp: 61200-element Vector{Float64}
└── dst_temp: 64800-element Vector{Float64}

julia> regrid!(dst_field, regridder, src_field)
ERROR: UndefVarError: `topology` not defined
Stacktrace:
 [1] regrid!(dst_field::Field{…}, regrider::XESMF.Regridder{…}, src_field::Field{…})
   @ OceananigansXESMFExt ~/development/TestOceananigans.jl/ext/OceananigansXESMFExt.jl:182
 [2] top-level scope
   @ REPL[13]:1
```
I think we need 
```julia
using Oceananigans.Grids: topology
```

simone-silvestri:
just with this
```julia
regridder(vec(interior(dst_field)), vec(interior(src_field)))
```
it works

xkykai:
Looks great! Should we consider the case for regridding in GPU as well? I was trying it out here
https://github.com/CliMA/ClimaOceanCalibration.jl/blob/4c379b8de655006722a1cf024da80e3b6b9a78aa/src/DataWrangling/bilinear_interpolator.jl
and came up with a (not so smart) way to do it temporarily:
```julia
function regrid!(dst, weights::CuSparseMatrixCSC, src)
    vec(dst) .= weights * CuArray(vec(src))
end
```

By the same approach to do it over the entire 3D `AbstractField` another not so smart way to do it is probably 
```julia
function regrid!(dst::AbstractField, src::AbstractField, interpolator::BilinearInterpolator)
    weights = interpolator.weights
    
    # Get the interior data
    dst_data = interior(dst)
    src_data = interior(src)
    
    Nz = size(src_data, 3)
    
    for k in 1:Nz
        src_slice = view(src_data, :, :, k)
        dst_slice = view(dst_data, :, :, k)
        regrid!(dst_slice, weights, src_slice)
    end
    
    return dst
end
```

I haven't put too much thought into this, and it looks inefficient as I had to allocate another `CuArray` for sparse matrix multiply for `CuSparseMatrixCSC` to work. This doesn't make use of the kernel approach as well. I haven't managed to get sparse matrix multiply to work within a kernel.

Perhaps a better way to go about this is to actually write our own sparse matrix multiply? Doing that can allow us to dispatch kernels to do the regridding over all vertical levels at the same time. It doesn't seem too hard to write but I also haven't thought very deeply about it.

navidcy:
I'm merging this as a good starting point and will continue working on it.

Some outstanding issues are:
- Cleanup of:
  https://github.com/CliMA/Oceananigans.jl/blob/21ba055122b3fadd3e90680ca3b49dd80f8f14de/ext/OceananigansXESMFExt.jl#L12-L29
- Ensure all is smooth with GPU architectures; add tests for GPU (perhaps move from GitHub actions to buildkite)
- Does the regridding reproduces the Python xESMF regridding?
- (Perhaps related to the above) Can the integral before and after regridding be better than `rtol = 1e-4`?
- Documentation/Example for how to do a regridding using XESMF.jl (docstrings with examples should suffice for now; convert some example(s) to doctest(s)).

PR #4783, opened 2025-09-12
Try reusing the julia path for distributed tests

simone-silvestri:
To clear the depot:
If you ever need to clear the depot (e.g. if some packages get corrupted and can't be correctly accessed), you can go to the [Clear Depot pipeline](https://buildkite.com/clima/clear-depot), click "new build", and put your pipeline's name as the value for the environment variable `SLUG`, i.e. `slug=oceananigans-distributed` for this pipeline).

### DIFF
--- a/.buildkite/distributed/pipeline.yml
+++ b/.buildkite/distributed/pipeline.yml
@@ -23,7 +23,7 @@ steps:
       - echo "--- Initialize distributed tests"
       - "julia -O0 --project -e 'using Pkg; Pkg.test()'"
     agents:
-      slurm_mem: 8G
+      slurm_mem: 50G
       slurm_ntasks: 1
       slurm_gpus_per_task: 1
 
@@ -37,7 +37,7 @@ steps:
     commands:
       - "srun julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
     agents:
-      slurm_mem: 8G
+      slurm_mem: 50G
       slurm_ntasks: 4
     retry:
       automatic:
@@ -52,7 +52,7 @@ steps:
     commands:
       - "srun julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
     agents:
-      slurm_mem: 8G
+      slurm_mem: 50G
       slurm_ntasks: 4
       slurm_gpus_per_task: 1
     retry:
@@ -115,7 +115,7 @@ steps:
       - "srun julia -O0 --color=yes --project -e 'using Pkg; Pkg.test()'"
     timeout_in_minutes: 1440
     agents:
-      slurm_mem: 100G # Apparently the GPU tests require more memory
+      slurm_mem: 150G # Apparently the GPU tests require more memory?
       slurm_ntasks: 4
       slurm_gpus_per_task: 1
     retry:


.buildkite/distributed/pipeline.yml
```diff
@@ -5,6 +5,7 @@ agents:
 
 env:
   JULIA_LOAD_PATH: "${JULIA_LOAD_PATH}:${BUILDKITE_BUILD_CHECKOUT_PATH}/.buildkite/distributed"
+  JULIA_DEPOT_PATH: "${JULIA_LOAD_PATH}:${BUILDKITE_BUILD_CHECKOUT_PATH}/.buildkite/distributed/.julia-oceananigans"
```
ali-ramadhan: I think `$BUILDKITE_BUILD_CHECKOUT_PATH` might be different for different builds?

https://buildkite.com/docs/pipelines/configure/environment-variables#BUILDKITE_BUILD_CHECKOUT_PATH

simone-silvestri: hmmm, you might be right. The `initialize` still takes about 25 minutes.

juliasloan25: can you try:
`JULIA_DEPOT_PATH: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/depot/default"`

That's what we use for most repos

.buildkite/distributed/pipeline.yml
juliasloan25: can you change this to:
```suggestion
  queue: central
```
The cluster is changing to a new filesystem soon and in updating for it we also renamed this queue

juliasloan25, line 4:
```suggestion
  modules: climacommon/2025_07_30
```
you should be able to use newer modules unless there are compatibility issues with e.g. newer Nsight systems versions. You can see the package versions for this climacommon [here](https://github.com/CliMA/ClimaModules/blob/main/climacommon/2025_07_30)

juliasloan25, line 25: If this pipeline is run often, it's good to add a concurrency group so multiple builds don't try to modify the cache at the same time
```suggestion
    concurrency: 1
    concurrency_group: 'depot/oceananigans-distributed'
    agents:
```

juliasloan25:
If you ever need to clear the depot (e.g. if some packages get corrupted and can't be correctly accessed), you can go to the [Clear Depot pipeline](https://buildkite.com/clima/clear-depot), click "new build", and put your pipeline's name as the value for the environment variable `SLUG`, i.e. `slug=oceananigans-distributed` for this pipeline).

juliasloan25:
I tried re-running the pipeline [here](https://buildkite.com/clima/oceananigans-distributed/builds/10848) and it still takes 30 mins to init so I'm not sure the cache is being hit. I can look into it more next week

.buildkite/distributed/pipeline.yml
juliasloan25, line 24: In some of the other CliMA repos we only call `Pkg.instantiate` (and sometimes `Pkg.precompile`) in the "initialize" step, e.g. in [ClimaCore](https://github.com/CliMA/ClimaCore.jl/blob/f5963ca8668bef961fcf808dab78b021de412fda/.buildkite/pipeline.yml#L22-L23). Does this also run all the tests in this step, or does the `TEST_GROUP: "init"` flag make it do something else?

juliasloan25: Ah I just saw that the `test_init.jl` does instantiate and precompile

juliasloan25:
I don't know why the cache isn't improving the initialization time. I made the same changes in ClimaOcean [here](https://github.com/CliMA/ClimaOcean.jl/pull/671) and initialization decreased from 49 to 22 minutes.

.buildkite/distributed/pipeline.yml
```diff
@@ -1,10 +1,10 @@
 agents:
-  queue: new-central
+  queue: central
   slurm_mem: 8G # Note that the tests run on shared nodes, so limiting the memory usage might help in avoiding long queues
-  modules: climacommon/2024_10_10
+  modules: climacommon/2025_07_30
 
 env:
-  JULIA_LOAD_PATH: "${JULIA_LOAD_PATH}:${BUILDKITE_BUILD_CHECKOUT_PATH}/.buildkite/distributed"
+  JULIA_DEPOT_PATH: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/depot/default"
```
giordano:
```suggestion
  JULIA_DEPOT_PATH: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/depot/default:"
```

giordano: So, this worked in the sense that now stdlibs aren't being precompiled anymore, compare [previous job](https://buildkite.com/clima/oceananigans-distributed/builds/10848#019a60d0-5b95-452c-b58d-eb9cd9de31cc/213-766) to [latest one](https://buildkite.com/clima/oceananigans-distributed/builds/10881#019a70a5-5d5f-4e7b-a8af-d588ce5599ef/214-659), and search for `✓ Distributed` or `✓ Test` in the two logs (spoiler: you'll find them only in the former)

giordano:
Are these machines homogeneous (i.e. all same ISA)?  Can you add a command like
```sh
julia -e 'using InteractiveUtils; versioninfo(; verbose=true)'
```
at the beginning of the job?  The following step for debugging why cache isn't being hit is to set the env var
```sh
JULIA_DEBUG=loading
```
(maybe only for the init step, it's very verbose)

giordano:
Also, https://buildkite.com/clima/oceananigans-distributed/builds/10881#019a70a5-5d5f-4e7b-a8af-d588ce5599ef/214-887
```
1.2519645e6 ms  ✓ Enzyme
```
is rather pathological: Enzyme isn't the fastest package to precompile, but this shouldn't really take 20 minutes, it takes less than one on my laptop. Are you sure these machines aren't oversubscribed?

juliasloan25:
This is a tricky time to be setting/changing the depot because Caltech's cluster is in the process of switching from one filesystem to another. @nefrathenrici noticed that these two subsequent runs are accessing the depot on the old and new filesystem respectively, which is why the init step didn't speed up at all with the depot ([run 1](https://buildkite.com/clima/oceananigans-distributed/builds/10847#019a60a6-35b5-47c2-846e-8212bae9471c), [run 2](https://buildkite.com/clima/oceananigans-distributed/builds/10848#_)).

We need to update both the queue for the pipeline YAML (done in this PR) and the buildkite steps ([here](https://buildkite.com/clima/oceananigans-distributed/settings/steps)) to use `queue: central` so that they both use the new filesystem. However if the steps are updated and the queue change from this PR isn't merged in yet, builds on other branches will fail because of the mismatch (this happened the other day e.g. in [this build](https://buildkite.com/clima/oceananigans-distributed/builds/10879)). So it would be good to verify the speedup from accessing the depot on the correct filesystem, then we need to quickly merge this PR so CI isn't broken.

juliasloan25:
Updating climacommon to 2025_07_30 switches the pipeline from using Julia 1.10 to 1.11. I switched back to climacommon/2024_10_10 so that we keep using 1.10 and we can see how only the depot change affects the time to init

juliasloan25:
The init time is down to 21 minutes ([build](https://buildkite.com/clima/oceananigans-distributed/builds/10937)) so I think this is good to go. We should merge it quickly to avoid the issue I [mentioned above](https://github.com/CliMA/Oceananigans.jl/pull/4783#issuecomment-3523388663)

nefrathenrici:
It looks like the GPU distributed unit tests are running out of memory https://buildkite.com/clima/oceananigans-distributed/builds/10941#019a7e07-ecea-4f9a-bef8-c33b98ea3d6d

simone-silvestri:
Sometimes it happens that the tests take more than the allowed time (1h7m) and then fail with an oom error. I am not sure why this happens... Restarting them clears the error

simone-silvestri:
I can merge, then we can try to reduce the init time further (21 mins is still a lot, the serial tests take 1 minute)

nefrathenrici:
> Sometimes it happens that the tests take more than the allowed time (1h7m) and then fail with an oom error. I am not sure why this happens... Restarting them clears the error

From the timestamps in the buildkite log, I think it fails with the OOM error but then hangs until it times out or gets cancelled manually.
```
2025-11-13 11:31:36 EST
	slurmstepd: error: Detected 1 oom_kill event in StepId=54805575.2. Some of the step tasks have been OOM Killed.
2025-11-13 11:31:37 EST
	srun: error: hpc-25-24: task 3: Out Of Memory
2025-11-13 11:41:18 EST
	# Received cancellation signal, interrupting
2025-11-13 11:41:18 EST
	srun: forcing job termination
2025-11-13 11:41:18 EST
	srun: Job step aborted: Waiting up to 32 seconds for job step to finish.
2025-11-13 11:41:18 EST
	slurmstepd: error: *** STEP 54805575.2 ON hpc-25-24 CANCELLED AT 2025-11-13T08:41:18 ***
```
 It may be worth adding more memory to the job if it is flaky.

simone-silvestri:
amazing!

PR #4788, opened 2025-09-16
Change the user interface so that closure tracers are automatically added 

glwagner:
This PR changes how model constructors work so that "auxiliary tracers" associated with closures --- which right now are just turbulent kinetic energy (TKE) for `CATKEVerticalDiffusivity`, and TKE + TKE dissipation (denoted \epsilon) for `TKEDissipationVerticalDiffusivity` are automatically added to models.

The rationale is that this is the interface we want in order to support "one line" changes between closures. Otherwise, users have to introduce their own logic (boilerplate) into scripts to switch between closures.

The original rationale for requiring explicit tracers is that this "self-documents" the tracer names. This is helpful for output. Another rationale is if users need to write forcing functions or set initial conditions for closure auxiliary variables. The latter two are rarely needed. The former remains, but can be solved with ordinary documentation. I think the pros of automatically adding closure auxiliary variables outweighs the cons.

### DIFF
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
@@ -17,8 +17,7 @@ using Oceananigans.Advection:
 
 using Oceananigans.BuoyancyFormulations:
     validate_buoyancy,
-    materialize_buoyancy,
-    SeawaterBuoyancy
+    materialize_buoyancy
 
 using Oceananigans.Biogeochemistry:
     validate_biogeochemistry,
@@ -29,11 +28,9 @@ using Oceananigans.Fields:
     Field,
     CenterField,
     tracernames,
-    VelocityFields,
     TracerFields
 
 using Oceananigans.Grids:
-    AbstractCurvilinearGrid,
     AbstractHorizontallyCurvilinearGrid,
     architecture,
     halo_size,
--- a/test/test_time_stepping.jl
+++ b/test/test_time_stepping.jl
@@ -355,7 +355,7 @@ timesteppers = (:QuasiAdamsBashforth2, :RungeKutta3)
                     C = nameof(typeof(closure))
                     @info "  Testing HydrostaticFreeSurfaceModel time stepping with datetime clocks [$A, $FT, $C]"
 
-                    tracers = (:b, :c, :e, :ϵ)
+                    tracers = (:b, :c)
                     clock = Clock(; time=DateTime(2020, 1, 1))
                     grid = RectilinearGrid(arch; size=(2, 2, 2), extent=(1, 1, 1))
                     @test eltype(grid) == FT


glwagner:
I think this is also a better design because it prevents potential conflict between BGC tracers and closure tracers.

test/test_turbulence_closures.jl
```diff
@@ -209,14 +209,17 @@ function run_time_step_with_catke_tests(arch, closure)
     grid = RectilinearGrid(arch, size=(2, 2, 2), extent=(1, 2, 3))
     buoyancy = BuoyancyTracer()
 
-    # These shouldn't work (need :e in tracers)
-    @test_throws ArgumentError HydrostaticFreeSurfaceModel(; grid, closure, buoyancy, tracers=:b)
-    @test_throws ArgumentError HydrostaticFreeSurfaceModel(; grid, closure, buoyancy, tracers=(:b, :E))
+    # These should work now (:e is auto-appended)
```
navidcy: Remove comment? It doesn't serve much...
```suggestion
```

glwagner:
> Let's bump a patch release?
> 
> Question: is this breaking change?

It's breaking if we keep the error, but we can use a warning instead (perhaps for some limited period of time)

simone-silvestri:
Definitely agree with this change! Agreed also with not bumping the version.

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -141,14 +141,35 @@ function HydrostaticFreeSurfaceModel(; grid,
...
+    user_tracer_names = tracernames(tracers)
+    closure_tracer_names = closure_required_tracers(closure)
+
+    # Throw an error in case of a conflict between user-specified tracers and any other tracers.
+    if any(name ∈ user_tracer_names for name in closure_tracer_names)
+        msg = string("The tracer names $(user_tracer_names) overlap with the closure auxiliary", '\n',
+                     "tracer names $(closure_tracer_names) associated with $(summary(closure)).", '\n',
+                     "The names $(closure_tracer_names) cannot be specified explicitly", '\n',
+                     "or be the names of biogeochemical tracers.")
+        throw(ArgumentError(msg))
```
simone-silvestri, line 212: I think this should be a warning, because the user should be able to manually add tracers even if using a closure that needs that tracer. This is in the spirit of making the code understandable and making sure users know what they are doing. We had this conversation some time ago in https://github.com/CliMA/ClimaOcean.jl/issues/125, where we decided that it is propedeutic to require the users to manually add tracers that are needed and not automatically add them. I think that conversation still holds here. I like the idea of automatically adding tracers (like I was stating in that conversation), but I wouldn't preclude the possibility of explicitly adding tracers that are needed for parameterizations.

glwagner: The reason we need to throw an error is because users may want to add _passive_ tracers whose names conflict with the active closure tracers. I think this is a significant enough issue to warrant an error.

I don't think there is a benefit to allowing users to write `tracers = :e` and have the same effect as not adding tracers at all. Rather, I think this behavior is confusing and surprising.

Regarding the conversation in ClimaOcean:

> We had this conversation some time ago in https://github.com/CliMA/ClimaOcean.jl/issues/125, where we decided that it is propedeutic to require the users to manually add tracers that are needed and not automatically add them.

This PR overturns that decision: after this PR, users will not manually add closure tracers.

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -8,10 +8,60 @@ using Oceananigans.Fields: Field, CenterField, tracernames, TracerFields
...
+    TimeStepper,
+    update_state!,
+    AbstractLagrangianParticles,
+    SplitRungeKutta3TimeStepper
+
+using Oceananigans.TurbulenceClosures:
+    validate_closure,
+    with_tracers,
+    build_closure_fields,
+    add_closure_specific_boundary_conditions
```
```suggestion
    add_closure_specific_boundary_conditions,
```

PR #4792, opened 2025-09-17
Use more efficient native broadcasting in `mean!` and `_mean`

glwagner:
Broadcasting over `Field` directly is still slow; this PR changes flavors of `Statistics.mean` to use broadcasting over the underlying data rather than `Field` as an optimization. The objective is to make these two more similar:

```julia
using Oceananigans
using BenchmarkTools

grid = RectilinearGrid(size=(8, 8, 8), extent=(1, 1, 1), halo=(1, 1, 1))
u = XFaceField(grid)
U = Field{Nothing, Nothing, Nothing}(grid)
Ui = zeros(1, 1, 1)

@benchmark mean!(U, u)
@benchmark mean!(Ui, interior(u))
```

I think this takes the Oceananigans version from something like 20x slower to 10x slower.

### DIFF
--- a/src/Fields/field.jl
+++ b/src/Fields/field.jl
@@ -793,7 +793,7 @@ function Statistics._mean(f, c::AbstractField, dims; condition = nothing, mask =
     operand = condition_operand(f, c, condition, mask)
     r = sum(operand; dims)
     n = conditional_length(operand, dims)
-    r ./= n
+    parent(r) ./= n
     return r
 end
 
@@ -804,7 +804,7 @@ function Statistics.mean!(f::Function, r::ReducedAbstractField, a::AbstractField
     sum!(f, r, a; condition, mask, init=true)
     dims = reduced_dimension(location(r))
     n = conditional_length(condition_operand(f, a, condition, mask), dims)
-    r ./= n
+    parent(r) ./= n
     return r
 end
 


glwagner:
Discussed on #4633

glwagner:
looks like we get

```julia
julia> @benchmark mean!(U, u)
BenchmarkTools.Trial: 10000 samples with 9 evaluations per sample.
 Range (min … max):  2.403 μs …  1.054 ms  ┊ GC (min … max):  0.00% … 99.20%
 Time  (median):     2.727 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   3.423 μs ± 20.359 μs  ┊ GC (mean ± σ):  12.69% ±  2.21%

  ▇▇██▇▆▄▃▂▁▁                                                ▂
  ████████████▇██▇▇▇▇▇▆▅▇▆▆▅▆▆▆▃▅▄▄▅▃▃▄▅▄▅▂▄▄▃▂▂▄▂▃▂▄▄▃▄▃▂▃▅ █
  2.4 μs       Histogram: log(frequency) by time     8.33 μs <

 Memory estimate: 4.70 KiB, allocs estimate: 21.

julia> @benchmark mean!(Ui, interior(u))
BenchmarkTools.Trial: 10000 samples with 10 evaluations per sample.
 Range (min … max):  1.562 μs …   9.575 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.596 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.617 μs ± 189.755 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▂▇█▇▇█▇▅▄▁▁    ▁▁▁ ▁ ▁                                      ▃
  ████████████▇▆████▇███▇██▇▆▁▆▆▅▄▅▁▄▄▅▃▄▅▅▆▆▅▅▅▅▅▅▁▅▅▄▄▃▄▅▅▅ █
  1.56 μs      Histogram: log(frequency) by time         2 μs <

 Memory estimate: 176 bytes, allocs estimate: 4.
```

maybe sufficient to close #4633? probably there is still a slow down for immersed boundary grids. But we may not be able to solve that issue without the ability to precompute the condition length, and this strategy cannot be implemented for `mean` itself (only for `Average` and friends can we do this)

glwagner:
@giordano do you know why `TracedStepRangeLen` isn't available here:

https://buildkite.com/clima/oceananigans/builds/25586#019957d5-d418-4818-80d8-d1726808e1f8/17-57

specifically

```

WARNING: could not import TracedRNumberOverrides.TracedStepRangeLen into OutputReaders
--
  | ERROR: LoadError: UndefVarError: `TracedStepRangeLen` not defined
  | Stacktrace:
  | [1] top-level scope
  | @ ~/Oceananigans.jl-25586/ext/OceananigansReactantExt/OutputReaders.jl:7
  | [2] include(mod::Module, _path::String)
  | @ Base ./Base.jl:495
  | [3] include(x::String)
  | @ OceananigansReactantExt ~/Oceananigans.jl-25586/ext/OceananigansReactantExt/OceananigansReactantExt.jl:1
  | [4] top-level scope
  | @ ~/Oceananigans.jl-25586/ext/OceananigansReactantExt/OceananigansReactantExt.jl:39
  | [5] include
  | @ ./Base.jl:495 [inlined]
  | [6] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::String)
  | @ Base ./loading.jl:2292
  | [7] top-level scope
  | @ stdin:4
  | in expression starting at /var/lib/buildkite-agent/Oceananigans.jl-25586/ext/OceananigansReactantExt/OutputReaders.jl:1
  | in expression starting at /var/lib/buildkite-agent/Oceananigans.jl-25586/ext/OceananigansReactantExt/OceananigansReactantExt.jl:1


```

giordano:
Might have been removed from inside `TracedRNumberOverrides` by https://github.com/EnzymeAD/Reactant.jl/commit/9d852a24d62f7b10d22ce595544d1cb3a7292253#diff-00e36184ac94da3cbddb9822c2a52c9e0307e2be03aa381a81577bbe4657d594  I guess the solution is to reference the actual module defining the type, which was kind of the goal of that PR. CC: @avik-pal

avik-pal:
`Reactant.TracedStepRangeLen`

taimoorsohail:
@glwagner thanks for opening this! I was just hitting this problem myself. I made a MWE that shows that immersed boundary grids are 2x as slow as non-immersed when using Average(). Can I ask - mean! is just the numerical mean, which doesn't take into account grid cell area/volume right? I am using Average(), but the process of wrapping the Reduction in Field() really slows down the code - something that has been noted before. 

Here's the MWE - let me know what you think: https://github.com/taimoorsohail/ocean-ensembles/pull/96/files#diff-a3594d1d71225656678ca43e3331329f441dfbbb017b0fc6d05167c1e65d4794

glwagner:
> @glwagner thanks for opening this! I was just hitting this problem myself. I made a MWE that shows that immersed boundary grids are 2x as slow as non-immersed when using Average(). Can I ask - mean! is just the numerical mean, which doesn't take into account grid cell area/volume right? I am using Average(), but the process of wrapping the Reduction in Field() really slows down the code - something that has been noted before.
> 
> Here's the MWE - let me know what you think: https://github.com/taimoorsohail/ocean-ensembles/pull/96/files#diff-a3594d1d71225656678ca43e3331329f441dfbbb017b0fc6d05167c1e65d4794

yes `mean!` is the naive mean over cells. We have to use `Average` to account for grid metrics.

To set expectations: `Average` _must_ be more expensive than `sum` or `mean` because `Average` requires more computation (eg it has to fetch the metrics).

If there is a huge difference though, its reasonable to think that we might be able to improve or optimize its performance.

glwagner:
Also these optimizations don't affect `Average` on a stretched or curvilinear grid unfortunately because they use pretty different codepaths. I'm open to working on the stretched/curvilinear case if we have a nice benchmark.

navidcy:
https://buildkite.com/clima/oceananigans/builds/25599#01995a72-63aa-497f-a107-889fb8a98167/17-330

<head></head>


Expression: minimum_xspacing(grid) ≈ FT(π / 3)
  | MethodError: condition_operand(::typeof(identity), ::KernelFunctionOperation{Center, Center, Center, ImmersedBoundaryGrid{Float32, Periodic, Periodic, Bounded, RectilinearGrid{Float32, Periodic, Periodic, Bounded, Oceananigans.Grids.StaticVerticalDiscretization{OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, Float32, Float32}, Float32, Float32, OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, CPU}, GridFittedBottom{Field{Center, Center, Nothing, Nothing, RectilinearGrid{Float32, Periodic, Periodic, Bounded, Oceananigans.Grids.StaticVerticalDiscretization{OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, Float32, Float32}, Float32, Float32, OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float64, Float64, Int64}}, CPU}, Tuple{Colon, Colon, Colon}, OffsetArray{Float32, 3, Array{Float32, 3}}, Float32, FieldBoundaryConditions{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, Nothing, Nothing, Nothing, @NamedTuple{bottom_and_top::Nothing, south_and_north::KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(9, 1)}, Oceananigans.Utils.OffsetStaticSize{(1:9, 1:1)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_periodic_south_and_north_halo!)}, west_and_east::KernelAbstractions.Kernel{KernelAbstractions.CPU, KernelAbstractions.NDIteration.StaticSize{(9, 1)}, Oceananigans.Utils.OffsetStaticSize{(1:9, 1:1)}, typeof(Oceananigans.BoundaryConditions.cpu__fill_periodic_west_and_east_halo!)}}, @NamedTuple{bottom_and_top::Tuple{Nothing, Nothing}, south_and_north::Tuple{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}}, west_and_east::Tuple{BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}, BoundaryCondition{Oceananigans.BoundaryConditions.Periodic, Nothing}}}}, Nothing, Nothing}, Oceananigans.ImmersedBoundaries.CenterImmersedCondition}, Nothing, Nothing, CPU}, Float32, typeof(Δx), Tuple{Center, Center, Center}}, ::Nothing, ::Float64) is ambiguous.
  |  
  | Candidates:
  | condition_operand(::typeof(identity), operand, ::Nothing, mask)
  | @ Oceananigans.Fields ~/Oceananigans.jl-25599/src/Fields/field.jl:719
  | condition_operand(func, op::AbstractField{<:Any, <:Any, <:Any, <:ImmersedBoundaryGrid}, condition, mask)
  | @ Oceananigans.ImmersedBoundaries ~/Oceananigans.jl-25599/src/ImmersedBoundaries/immersed_reductions.jl:133
  |  
  | Possible fix, define
  | condition_operand(::typeof(identity), ::AbstractField{<:Any, <:Any, <:Any, <:ImmersedBoundaryGrid}, ::Nothing, ::Any)
 

<br class="Apple-interchange-newline" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;">

PR #4793, opened 2025-09-17
Fix `TracedStepRangeLen` reference for latest Reactant

glwagner:
cc @giordano @avik-pal @jlk9

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.99.0"
+version = "0.99.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
@@ -90,7 +90,7 @@ OffsetArrays = "1.4"
 OrderedCollections = "1.1"
 Printf = "1.9"
 Random = "1.9"
-Reactant = "0.2.141"
+Reactant = "0.2.162"
 ReactantCore = "0.1"
 Rotations = "1.0"
 SeawaterPolynomials = "0.3.9"


glwagner:
should we update compat on Reactant too?

ext/OceananigansReactantExt/OutputReaders.jl
```diff
@@ -2,9 +2,7 @@ module OutputReaders
 
 import Oceananigans.OutputReaders: find_time_index
 
-import Reactant.TracedRNumberOverrides: TracedStepRangeLen
-
-@inline function find_time_index(times::TracedStepRangeLen, t)
+@inline function find_time_index(times::Reactant.TracedStepRangeLen, t)
```
giordano: where's `Reactant` coming from here?  You need to import it, right?  Or even nicer:
```
using Reactant: TracedStepRangeLen
```

giordano (commented):
I think we'll need to fix the compat bounds in the registry to avoid troubles to users

giordano:
> should we update compat on Reactant too?

@avik-pal correct me if I'm wrong, but I think `Reactant.TracedStepRangeLen` should always work?

ext/OceananigansReactantExt/OutputReaders.jl
glwagner: oops

avik-pal:
> @avik-pal correct me if I'm wrong, but I think Reactant.TracedStepRangeLen should always work?

No it wasn't defined outside before. As a temp solution, we can import `TracedStepRangeLen` into `TracedRArrayOverrides` to keep old code functional

ext/OceananigansReactantExt/OutputReaders.jl
glwagner: thanks

glwagner:
> > @avik-pal correct me if I'm wrong, but I think Reactant.TracedStepRangeLen should always work?
> 
> No it wasn't defined outside before. As a temp solution, we can import `TracedStepRangeLen` into `TracedRArrayOverrides` to keep old code functional

or move boldly into the future 🚀

giordano:
> No it wasn't defined outside before.

Argh, then that change was slightly breaking 😬

> or move boldly into the future 🚀

Then you need to bump the compat bound, yes.

giordano:
> As a temp solution, we can import TracedStepRangeLen into TracedRArrayOverrides to keep old code functional

Actually, since in GB-25 we're stuck with an old version of Oceananigans we very probably want to do this.  Currently tests in Enzyme-JAX are broken too, and this PR won't help.

glwagner:
@giordano should we avoid the compat then?

avik-pal:
The compat will still be needed, unless we yank `0.2.162`

giordano:
Either

1. you change the code & bump the compat bound (what the PR is doing now)
2. or you don't do anything and this package will be broken only with Reactant v0.2.162 (unless you yank it, but people don't like that because it causes the resolver to fail badly)

I'd go with 1 and we're all happy, we might want only to change compat bounds of Oceananigans older than this PR to Reactant < 0.2.162 in the registry, but after this PR + https://github.com/EnzymeAD/Reactant.jl/pull/1675 that'd be less urgent.

glwagner:
ok I'll merge this and press forward! thanks!

PR #4798, opened 2025-09-19
Move AbstractModel interface to top level and add docstrings

### LINKED ISSUE #4797: Improve documentation of `AbstractModel` interface
I think it would be helpful to more thoroughly document (and use) the method interface for `AbstractModel` defined in [`Models.jl`](https://github.com/CliMA/Oceananigans.jl/blob/main/src/Models/Models.jl#L28).

First, this comment:

```julia
# We assume that model has some properties, eg:
#   - model.clock::Clock
#   - model.architecture.
#   - model.timestepper with timestepper.G⁻ and timestepper.Gⁿ :spiral_eyes:
```
should probably be moved to the docstring for `AbstractModel`.

Second, I would suggest the following docstrings for these methods:

```julia
"""
    $SIGNATURES

Returns the current iteration of the `model.clock`.
"""
iteration(model::AbstractModel) = model.clock.iteration

"""
    $SIGNATURES

Returns the device architecture defined by the underlying model `grid`.
"""
architecture(model::AbstractModel) = model.grid.architecture

"""
    $SIGNATURES

TODO: what does this method do in addition to the constructor?
"""
initialize!(model::AbstractModel) = nothing


"""
    $SIGNATURES

TODO
"""
total_velocities(model::AbstractModel) = nothing

"""
    $SIGNATURES

Returns the `TimeStepper` used by the given `model`.
"""
timestepper(model::AbstractModel) = model.timestepper

"""
    $SIGNATURES

Defines initialization routines for the first call to `update_state!` after initialization. Defaults to invoking `update_state!(model; kw...)`.
"""
initialization_update_state!(model::AbstractModel; kw...) = update_state!(model; kw...) # fallback

"""
    $SIGNATURES

Updates all `FieldTimeSeries` fields defined on the given `model. The default implementation assumes that the `model` that does not contain any `FieldTimeSeries` and returns `nothing`.
"""
update_model_field_time_series!(model::AbstractModel, clock::Clock) = nothing

"""
    Base.time(model::AbstractModel) 

Returns the current `time` from the given `model.clock`.
"""
Base.time(model::AbstractModel) = model.clock.time

"""
    Base.eltype(model::AbstractModel)

Returns the numeric `eltype` of `model.grid` and all associated `Field`s.
"""
Base.eltype(model::AbstractModel) = eltype(model.grid)
```

I would also be in favor of adding `get_clock` (or just `clock`) and `get_grid` methods but this isn't strictly necessary.

I guess we should also add a doc page on the `AbstractModel` interface to the appendix.

bgroenks96:
Resolves #4797

### DIFF
--- a/src/Models/Models.jl
+++ b/src/Models/Models.jl
@@ -25,27 +25,6 @@ import Oceananigans.Solvers: iteration
 import Oceananigans.Simulations: timestepper
 import Oceananigans.TimeSteppers: reset!, set_clock!
 
-# A prototype interface for AbstractModel.
-#
-# TODO: decide if we like this.
-#
-# We assume that model has some properties, eg:
-#   - model.clock::Clock
-#   - model.architecture.
-#   - model.timestepper with timestepper.G⁻ and timestepper.Gⁿ :spiral_eyes:
-
-iteration(model::AbstractModel) = model.clock.iteration
-Base.time(model::AbstractModel) = model.clock.time
-Base.eltype(model::AbstractModel) = eltype(model.grid)
-architecture(model::AbstractModel) = model.grid.architecture
-initialize!(model::AbstractModel) = nothing
-total_velocities(model::AbstractModel) = nothing
-timestepper(model::AbstractModel) = model.timestepper
-initialization_update_state!(model::AbstractModel; kw...) = update_state!(model; kw...) # fallback
-
-# Fallback for any abstract model that does not contain `FieldTimeSeries`es
-update_model_field_time_series!(model::AbstractModel, clock::Clock) = nothing
-
 #####
 ##### Model-building utilities
 #####
--- a/src/Oceananigans.jl
+++ b/src/Oceananigans.jl
@@ -84,6 +84,9 @@ export
     ConservativeFormulation, VectorInvariantFormulation,
     PressureField, fields, ZCoordinate, ZStarCoordinate,
 
+    # Model interface
+    iteration, architecture, initialize!, timesteppper, initialization_update_state!,
+
     # Hydrostatic free surface model stuff
     VectorInvariant, ExplicitFreeSurface, ImplicitFreeSurface, SplitExplicitFreeSurface,
     HydrostaticSphericalCoriolis, PrescribedVelocityFields,
@@ -92,7 +95,7 @@ export
     Clock, TimeStepWizard, conjure_time_step_wizard!, time_step!,
 
     # Simulations
-    Simulation, run!, Callback, add_callback!, iteration,
+    Simulation, run!, Callback, add_callback!,
     iteration_limit_exceeded, stop_time_exceeded, wall_time_limit_exceeded,
 
     # Diagnostics
@@ -162,8 +165,78 @@ const defaults = Defaults()
     AbstractModel
 
 Abstract supertype for models.
+
+Models are required to have the following fields:
+  - `model.clock::Clock`
+  - `model.architecture`
+  - `model.timestepper` with `timestepper.G⁻` and `timestepper.Gⁿ`
 """
 abstract type AbstractModel{TS, A} end
+
+# Method interface for AbstractModel
+
+"""
+    $SIGNATURES
+
+Returns the current iteration of the `model.clock`.
+"""
+iteration(model::AbstractModel) = model.clock.iteration
+
+"""
+    $SIGNATURES
+
+Returns the device architecture defined by the underlying model `grid`.
+"""
+architecture(model::AbstractModel) = model.grid.architecture
+
+"""
+    $SIGNATURES
+
+TODO: what does this method do in addition to the constructor?
+"""
+initialize!(model::AbstractModel) = nothing
+
+"""
+    $SIGNATURES
+
+TODO
+"""
+total_velocities(model::AbstractModel) = nothing
+
+"""
+    $SIGNATURES
+
+Returns the `TimeStepper` used by the given `model`.
+"""
+timestepper(model::AbstractModel) = model.timestepper
+
+"""
+    $SIGNATURES
+
+Defines initialization routines for the first call to `update_state!` after initialization. Defaults to invoking `update_state!(model; kw...)`.
+"""
+initialization_update_state!(model::AbstractModel; kw...) = update_state!(model; kw...) # fallback
+
+"""
+    $SIGNATURES
+
+Updates all `FieldTimeSeries` fields defined on the given `model. The default implementation assumes that the `model` that does not contain any `FieldTimeSeries` and returns `nothing`.
+"""
+update_model_field_time_series!(model::AbstractModel, clock::Clock) = nothing
+
+"""
+    Base.time(model::AbstractModel) 
+
+Returns the current `time` from the given `model.clock`.
+"""
+Base.time(model::AbstractModel) = model.clock.time
+
+"""
+    Base.eltype(model::AbstractModel)
+
+Returns the numeric `eltype` of `model.grid` and all associated `Field`s.
+"""
+Base.eltype(model::AbstractModel) = eltype(model.grid)
 
 """
     AbstractDiagnostic
--- a/ext/OceananigansReactantExt/Models.jl
+++ b/ext/OceananigansReactantExt/Models.jl
@@ -2,7 +2,7 @@ module Models
 
 import Oceananigans
 
-import Oceananigans.Models: initialization_update_state!
+import Oceananigans: initialization_update_state!
 import Oceananigans.Models.HydrostaticFreeSurfaceModels.SplitExplicitFreeSurfaces: maybe_extend_halos, FixedSubstepNumber
 import Oceananigans: initialize!
 
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
@@ -16,8 +16,7 @@ using Oceananigans.TurbulenceClosures: validate_closure, with_tracers, build_dif
 using Oceananigans.TurbulenceClosures: time_discretization, implicit_diffusion_solver
 using Oceananigans.Utils: tupleit
 
-import Oceananigans: initialize!
-import Oceananigans.Models: total_velocities, timestepper
+import Oceananigans: initialize!, total_velocities, timestepper
 
 PressureField(grid) = (; pHY′ = CenterField(grid))
 
--- a/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
+++ b/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
@@ -18,8 +18,8 @@ using Oceananigans.TurbulenceClosures.TKEBasedVerticalDiffusivities: FlavorOfCAT
 using Oceananigans.Utils: tupleit
 using Oceananigans.Grids: topology
 
+import Oceananigans: total_velocities, timestepper
 import Oceananigans.Architectures: architecture
-import Oceananigans.Models: total_velocities, timestepper
 
 const ParticlesOrNothing = Union{Nothing, AbstractLagrangianParticles}
 const AbstractBGCOrNothing = Union{Nothing, AbstractBiogeochemistry}
--- a/src/Simulations/Simulations.jl
+++ b/src/Simulations/Simulations.jl
@@ -18,9 +18,6 @@ using OrderedCollections: OrderedDict
 
 import Base: show
 
-# To be extended in the `Models` module
-timestepper(model) = nothing
-
 include("callback.jl")
 include("simulation.jl")
 include("run.jl")


bgroenks96:
@glwagner Are you OK with moving the model interface methods out of the `Models` submodule up to the top-level module?

src/Oceananigans.jl
```diff
@@ -162,9 +165,65 @@ const defaults = Defaults()
...
+    $SIGNATURES
+
+Returns the current iteration of the `model.clock`.
+"""
+iteration(model::AbstractModel) = model.clock.iteration
+
+"""
+    $SIGNATURES
+
+TODO: what does this method do in addition to the constructor?
```
simone-silvestri: ensures consistency between internal variables in the `AbstractModel` and user-specified initial conditions provided by the user through the `set(::AbstractModel; kwargs...)` function

src/Oceananigans.jl
```diff
@@ -162,9 +165,65 @@ const defaults = Defaults()
...
+"""
+initialize!(model::AbstractModel) = nothing
+
+"""
+    $SIGNATURES
+
+TODO
+"""
+total_velocities(model::AbstractModel) = nothing
```
simone-silvestri, line 198: Not really well defined at the moment. It returns the sum of background + prognostic velocities for the non hydrostatic model and the prognostic velocities in the hydrostatic model

src/Oceananigans.jl
```diff
@@ -162,9 +165,65 @@ const defaults = Defaults()
     AbstractModel
 
 Abstract supertype for models.
+
+Models are required to have the following fields:
+  - `model.clock::Clock`
+  - `model.architecture`
+  - `model.timestepper` with `timestepper.G⁻` and `timestepper.Gⁿ`
```
simone-silvestri: I don't think we want to assume a timestepper: the `OceanSeaIceModel` does not have a timestepper

src/Oceananigans.jl
```diff
@@ -162,9 +165,65 @@ const defaults = Defaults()
     AbstractModel
 
 Abstract supertype for models.
+
+Models are required to have the following fields:
+  - `model.clock::Clock`
+  - `model.architecture`
```
simone-silvestri: I think it would be best to assume a `grid` rather than an architecture, the architecture should be only a grid field

src/Oceananigans.jl
```diff
@@ -162,9 +165,65 @@ const defaults = Defaults()
...
+Returns the `TimeStepper` used by the given `model`.
+"""
+timestepper(model::AbstractModel) = model.timestepper
+
+"""
+    $SIGNATURES
+
+Defines initialization routines for the first call to `update_state!` after initialization. Defaults to invoking `update_state!(model; kw...)`.
+"""
+initialization_update_state!(model::AbstractModel; kw...) = update_state!(model; kw...) # fallback
```
simone-silvestri: missing initialize!
```suggestion
function initialization_update_state!(model::AbstractModel; kw...) 
    initialize!(model)
    update_state!(model; kw...) # fallback
end
```
(it's a bug we had to correct anyways)

src/Oceananigans.jl
simone-silvestri: and `timestepper.G⁻` is an AB2-specific field

src/Oceananigans.jl
bgroenks96: Fixed in [ea3bb23](https://github.com/CliMA/Oceananigans.jl/pull/4798/commits/ea3bb23421fd9e30917556cf2faabf74034183d8)

src/Oceananigans.jl
bgroenks96: Fixed in [be79014](https://github.com/CliMA/Oceananigans.jl/pull/4798/commits/be790147d1058a632b8721df37172ae6ba676c8d)

src/Oceananigans.jl
bgroenks96: Removed in [be79014](https://github.com/CliMA/Oceananigans.jl/pull/4798/commits/be790147d1058a632b8721df37172ae6ba676c8d)

bgroenks96: What about the `timestepper` method?

src/Oceananigans.jl
bgroenks96: Updated in [be79014](https://github.com/CliMA/Oceananigans.jl/pull/4798/commits/be790147d1058a632b8721df37172ae6ba676c8d)

src/Oceananigans.jl
bgroenks96: Updated in [be79014](https://github.com/CliMA/Oceananigans.jl/pull/4798/commits/be790147d1058a632b8721df37172ae6ba676c8d)

src/Oceananigans.jl
simone-silvestri: I don't think we need it. It is used only in the `reset!(::AbstractModel)` method, which should know wether we have a timestepper or not.

simone-silvestri: oh, actually it is currently used in the `initialize!(simulation)` step... But this should call `initialize!(model)` and not `reset!(timestepper(model))`

src/Oceananigans.jl
```diff
@@ -162,9 +165,52 @@ const defaults = Defaults()
...
+Ensures consistency between internal variables in the AbstractModel and user-specified
+initial conditions provided by the user via `set(::AbstractModel; kwargs...)`.
+Default implementation does nothing.
+"""
+initialize!(model::AbstractModel) = nothing
+
+"""
+    $SIGNATURES
+
+Returns the sum of all velocities defined on the model (background + prognostic, where applicable).
```
simone-silvestri:
```suggestion
Returns the sum of the user-prescribed background + prognostic velocities, where applicable.
```
we have many more velocities than only background + prognostic, which are not considered with this definition (biogeochemical velocities, advective forcing, closure velocities...). For this reason, the name of this function is a bit of a misnomer.

src/Oceananigans.jl
jagoosw: Don't we really assume that there is a method for `architecture(model)` rather than explicitly assuming that there is a grid or architecture? For example you might have a that is 0D in space so doesn't need a grid. Maybe there are instances in the code that assume `model.grid` exists but I don't know if thats the best practice that we want to assume.

glwagner: I am also not sure why `architecture(model)` would be required for _every_ model.

src/TimeSteppers/TimeSteppers.jl
```diff
@@ -24,6 +27,23 @@ function update_state! end
 function compute_tendencies! end
 function compute_flux_bc_tendencies! end
 
+"""
+    $SIGNATURES
+
+Defines initialization routines for the first call to `update_state!` after initialization. Defaults to invoking `update_state!(model; kw...)`.
```
glwagner:
```suggestion
Define initialization routines for the first call to `update_state!` after initialization.
Defaults to invoking `update_state!(model; kw...)`.
```

src/Models/Models.jl
```diff
@@ -19,31 +21,22 @@ using Oceananigans.OutputReaders: update_field_time_series!, extract_field_time_
...
-initialize!(model::AbstractModel) = nothing
-total_velocities(model::AbstractModel) = nothing
-timestepper(model::AbstractModel) = model.timestepper
-initialization_update_state!(model::AbstractModel; kw...) = update_state!(model; kw...) # fallback
 
-# Fallback for any abstract model that does not contain `FieldTimeSeries`es
+"""
+    $SIGNATURES
+
+Updates all `FieldTimeSeries` fields defined on the given `model. The default implementation assumes that the `model` that does not contain any `FieldTimeSeries` and returns `nothing`.
```
glwagner:
```suggestion
Update all `FieldTimeSeries` fields defined on the given `model.
The default implementation assumes that the `model` that does not contain any `FieldTimeSeries` and returns `nothing`.
```

src/Oceananigans.jl
glwagner: I don't think "velocities" should be intrinsic to the concept of a model.

glwagner:
I think there needs to be some thought / intentionality about what functions belong to a "generic model" interface, versus what functions are only associated with `NonhydrostaticModel` / `HydrostaticFreeSurfaceModel`. Here are some examples

* models should not need to have "time steppers". The intrinsic thing is a time stepping rule, ie `time_step!(model, dt)`. It doesn't matter if there is a property `model.timestepper`.
* models should not need to have grids or architectures. Its not well defined in many cases. A coupled model does not have a single grid. 

To develop the scope of models I would consider the following cases:

* a fluids model, like `NonhydrostaticModel` (which used to be the only model)
* a coupled model, like an EarthSystemModel with a mixture of prescribed and prognostic componets
* a 0D model with a very simple evolution rule, ie something like `next = prev + dt * exp(next)` (exponential growth). This model has neither grid nor architecture, only `time_step!(model, dt)`.

The big danger is that we go in the opposite direction that we have been generally striving for --- a simple, useful interface which is disentangled from the "finite volume" (grids, fields) part of the code.

glwagner:
> @glwagner Are you OK with moving the model interface methods out of the `Models` submodule up to the top-level module?

Does this move us in the direction of being able to disentangle finite volume stuff (grids, fields, etc) from model implementation? What is your vision? Usually moving things to top-level acts to further entangle...

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
```diff
@@ -1,5 +1,5 @@
+using Oceananigans: total_velocities
```
glwagner, line 1: I am wary of this. It's a hack to begin with and this sort of cements it. I think we want to go the other way...

src/Oceananigans.jl
bgroenks96: I agree, but I didn't want to rock the boat too much.

src/TimeSteppers/TimeSteppers.jl
```diff
@@ -24,6 +27,23 @@ function update_state! end
...
+"""
+function initialization_update_state!(model::AbstractModel; kw...) 
+    initialize!(model)
+    update_state!(model; kw...) # fallback
+end
```
glwagner, line 39: I think we will be able to get rid of `initialization_update_state!` soon FYI. It was introduced to support work with Reactant but I do not think it is needed anymore.

src/TimeSteppers/TimeSteppers.jl
```diff
@@ -24,6 +27,23 @@ function update_state! end
...
+
+"""
+    $SIGNATURES
+
+Returns the `TimeStepper` used by the given `model`.
+"""
+timestepper(model::AbstractModel) = model.timestepper
```
glwagner, line 46: I'm wary of baking in the notion of a time stepper

glwagner:
I said this in a chat so I'll say it here so we can all discuss --- I think that the "definition" of a model belongs to the `Simulations` module. This makes sense because the requirements of a model are determined by `Simulation` and its functions. For example, `Simulation` requires a model to have a function `time_step!(model, dt)`.

I am not sure about all of the other functions. We have things like `initialize!`, `reset!`. Maybe we need `finalize!`. We definitely don't want any physics assumptions baked in, it `total_velocities` is bad. I am not even sure that we want to enshrine the notion of "prognostic fields", but that is more debatable. I'd be ok with accepting fields as a requirement.

At the top-level, I think we want to move towards an import order that looks somethin like

1. Grids, Fields, Solvers, FieldTimeSeries, utils
2. Simulation, OutputWriters
3. Models

Then someday we may pull out 1 & 2 into another package. We could also have 3 packages (but that's probably overkill. Then, the "models" will basically be the only part of Oceananigans that remains (which makes sense --- ocean specific stuff).

src/Oceananigans.jl
```diff
@@ -162,9 +165,52 @@ const defaults = Defaults()
...
+  - `model.grid::AbstractGrid`
 """
 abstract type AbstractModel{TS, A} end
 
+# Method interface for AbstractModel
+
+"""
+    $SIGNATURES
+
+Returns the current iteration of the `model.clock`.
```
navidcy:
```suggestion
Return the current iteration of the `model.clock`.
```

src/Oceananigans.jl
```diff
@@ -162,9 +165,52 @@ const defaults = Defaults()
...
+"""
+iteration(model::AbstractModel) = model.clock.iteration
+
+"""
+    $SIGNATURES
+
+Ensures consistency between internal variables in the AbstractModel and user-specified
+initial conditions provided by the user via `set(::AbstractModel; kwargs...)`.
+Default implementation does nothing.
```
navidcy:
```suggestion
Ensure consistency between internal variables in the `AbstractModel` and user-specified
initial conditions provided by via `set(::AbstractModel; kwargs...)`.
Default implementation does nothing.
```

src/Oceananigans.jl
```diff
@@ -162,9 +165,52 @@ const defaults = Defaults()
...
+    Base.time(model::AbstractModel) 
+
+Returns the current `time` from the given `model.clock`.
+"""
+Base.time(model::AbstractModel) = model.clock.time
+
+"""
+    Base.eltype(model::AbstractModel)
+
+Returns the numeric `eltype` of `model.grid` and all associated `Field`s.
```
navidcy:
```suggestion
Return the current `time` from the `model.clock`.
"""
Base.time(model::AbstractModel) = model.clock.time

"""
    Base.eltype(model::AbstractModel)

Return the numeric `eltype` of the `model.grid` and all associated `Field`s.
```

src/TimeSteppers/TimeSteppers.jl
navidcy: True. Is this used anywhere?

src/Models/LagrangianParticleTracking/lagrangian_particle_advection.jl
bgroenks96: You mean the method in general, right? Or just the fact that it's at the top-level?

glwagner:
@bgroenks96 do you mind if I help on this PR?

I think that clearly defining the interface for a model is very useful. I would also propose to add documentation with a simple example of defining a new model. This might also motivate improvements.

I would like to propose that the model interface be implemented in `Simulations`. I believe that `Simulation` is the appropriate object wherefrom the definition of a "model" should be derived. For example, `Simulation` determines the fundamental property that model needs to define `time_step!(model, dt)`, and so on.

This avoids putting a definition of `AbstractModel` in the top-level of Oceananigans which preserves encapsulation. Although encapsulation of package functionality is poor right now, it needs to be improved to decrease the work required to split up this package if need be. Adding more to the top-level may make this harder rather than easier.

bgroenks96:
Sure, I think that's where we left it anyway. It might be easier at this point, though, to just close this PR and open a new one? It's gotten quite out of date, and it's not really that many changes.

PR #4803, opened 2025-09-20
Add Reactant compat

navidcy:
this is an attempt to see if the Distributed CI passes


navidcy:
@wsmoses, @giordano any idea about this error?

https://buildkite.com/clima/oceananigans-distributed/builds/9996#01996929-d7d2-46ad-91cf-20e9671982aa

It only appears on the distributed CI so perhaps we need to tweak something there? cc @simone-silvestri

wsmoses:
what version of cuda is on the machine?

wsmoses:
can you add the reelvant prints from https://enzymead.github.io/Reactant.jl/dev/introduction/configuration#Debugging-installation-with-Nvidia-GPUs

```
julia> rm(joinpath(Base.DEPOT_PATH[1], "compiled", "v$(VERSION.major).$(VERSION.minor)", "Reactant_jll"); recursive=true, force=true)

julia> ENV["JULIA_DEBUG"] = "Reactant_jll";

julia> import Pkg

julia> Pkg.add("Reactant_jll")
  [...]
  1 dependency had output during precompilation:
┌ Reactant_jll
│  ┌ Debug: Detected CUDA Driver version 12.2.0
│  └ @ Reactant_jll ~/.julia/packages/Reactant_jll/daenT/.pkg/platform_augmentation.jl:60
│  ┌ Debug: Adding include dependency on /lib/x86_64-linux-gnu/libcuda.so.1
│  └ @ Reactant_jll ~/.julia/packages/Reactant_jll/daenT/.pkg/platform_augmentation.jl:108
```

giordano:
Short story is that XLA officially dropped support for CUDA < 12.6, and we incorrectly assumed this may have worked for CUDA 12.4-12.5, but that wasn't the case.  Should be fixed by https://github.com/JuliaPackaging/Yggdrasil/pull/12111 by completely disabling CUDA support if the version doesn't match.  If you do need the GPU version of Reactant that won't work with current version of the CUDA toolkit available, but I don't think that's the case for these CI runs, right?

giordano:
Actually, on a closer inspection of the error message https://buildkite.com/clima/oceananigans-distributed/builds/9996#01996929-d7d2-46ad-91cf-20e9671982aa/189-967
```
  LoadError: InitError: could not load library "/central/scratch/esm/slurm-buildkite/oceananigans-distributed/9996/depot/default/artifacts/8c190c28f25d44809e766c0c139af6fd05dcbadc/lib/libReactantExtra.so"
  /central/scratch/esm/slurm-buildkite/oceananigans-distributed/9996/depot/default/artifacts/8c190c28f25d44809e766c0c139af6fd05dcbadc/lib/libReactantExtra.so: undefined symbol: cusolverDnXsyevBatched_bufferSize, version libcusolver.so.11
  Stacktrace:
    [1] dlopen(s::String, flags::UInt32; throw_error::Bool)
      @ Base.Libc.Libdl ./libdl.jl:117
    [2] dlopen(s::String, flags::UInt32)
      @ Base.Libc.Libdl ./libdl.jl:116
    [3] macro expansion
      @ /central/scratch/esm/slurm-buildkite/oceananigans-distributed/9996/depot/default/packages/JLLWrappers/m2Pjh/src/products/library_generators.jl:63 [inlined]
    [4] __init__()
      @ Reactant_jll /central/scratch/esm/slurm-buildkite/oceananigans-distributed/9996/depot/default/packages/Reactant_jll/Bwh5k/src/wrappers/x86_64-linux-gnu-cxx11-cuda_version+12.8-gpu+cuda-mode+opt.jl:33
```
the path of the erroring file tells us that the detected CUDA toolkit version is 12.7 or 12.8 (12.8 in the path is the maximum of the two), so now I'm even more confused by why the symbol `cusolverDnXsyevBatched_bufferSize` (which was introduced in CUDA 12.6, see https://rocm.docs.amd.com/projects/HIPIFY/en/docs-6.3.2/tables/CUSOLVER_API_supported_by_HIP.html) isn't found. Looks like somehow an older `libcusolver.so.11` is being found.

giordano:
Out of curiosity, why was this closed? Did the issue go away?

navidcy:
Yes. Sorry I should have commented! I started seeing the distributed CI passing!

simone-silvestri:
I have removed `using Reactant`, which was not used in the distributed pipelines from the distributed tests, to allow tests to pass (we cannot stay without tests).

However, the issue still stands

PR #4805, opened 2025-09-21
Eliminate GridMetricOperation and create node metrics

glwagner:
This PR gets rid of `GridMetricOperation`, as it seems redundant with `KernelFunctionOperation`. In addition it creates new capabilities to form arithmetic expressions with "coordinates", ie so we can write

```julia
using Oceananigans
grid = RectilinearGrid(size=(4, 4, 4), extent=(4, 4, 4))
c = CenterField(grid)
using Oceananigans.Operators: z
c_z = Field(c * z)
```

### DIFF
--- a/src/AbstractOperations/binary_operations.jl
+++ b/src/AbstractOperations/binary_operations.jl
@@ -123,8 +123,8 @@ function define_binary_operator(op)
         $op(Lc::Tuple, f::Function, b::AbstractField) = $op(Lc, FunctionField(location(b), f, b.grid), b)
         $op(Lc::Tuple, a::AbstractField, f::Function) = $op(Lc, a, FunctionField(location(a), f, a.grid))
 
-        $op(Lc::Tuple, m::AbstractGridMetric, b::AbstractField) = $op(Lc, GridMetricOperation(location(b), m, b.grid), b)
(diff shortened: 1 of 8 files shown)


glwagner:
@tomchor can you advise about this question:

> Finally, I think we should change xnodes, ynodes, etc to return the grid_metric_operation (in the same way as xspacing, etc). But I am also not sure why we don't do that now.

simone-silvestri (approved):

tomchor:
> @tomchor can you advise about this question:
> 
> > Finally, I think we should change xnodes, ynodes, etc to return the grid_metric_operation (in the same way as xspacing, etc). But I am also not sure why we don't do that now.

Sorry, I'm only seeing this now' :(

That said, I'm not aware of any good reason why we made the change for spacings but not for nodes.

tomchor:
@glwagner can you clarify the intended usage of the changes to XNode and related stuff? I thought you were making `xnode` return a `KernelFunctionOperation`, but that doesn't seem to be the case:

```julia
julia> using Oceananigans

julia> xnodes(grid, Center());

julia> xnodes(grid, Center())
0.5:1.0:3.5
```

glwagner:
In this PR, I proposed making such a change:

> Finally, I think we should change xnodes, ynodes, etc to return the grid_metric_operation (in the same way as xspacing, etc). But I am also not sure why we don't do that now.

However, I didn't get any feedback on the proposal so I did not implement it. I still think it is a good idea.

PR #4807, opened 2025-09-22
Bump Enzyme compat

wsmoses:
In principle this version and later should bring good compile time improvements [lets find out]

cc @vchuravy


simone-silvestri:
Much faster, 13 mins vs 45 mins! I'll merge even if the distributed tests do not pass as the problem is in main and not in this PR.

Issue #4729, opened 2025-08-27
`nodes()` returns the full grid when given a smaller view of a `Field`


tomchor:
I'm not sure this is an issue per se, but I noticed that when a view of a `Field` is given to `nodes()` it always returns the nodes for the full grid, regardless of what the view is. For example:

```julia
julia> grid = RectilinearGrid(size=(8, 2, 2), extent=(1, 1, 1));

julia> c = XFaceField(grid);

julia> nodes(c)
(0.0:0.125:0.875, 0.25:0.5:0.75, -0.75:0.5:-0.25)

julia> nodes(view(c, 2:4, :, :))
(0.0:0.125:0.875, 0.25:0.5:0.75, -0.75:0.5:-0.25)
```

Looking at the code I see maybe this was always the intended the behavior, but I think it might be more useful to return the nodes corresponding to the indices of the view in this case. Curious about what everyone thinks.

simone-silvestri:
I guess we never really implemented nodes for windowed fields, but I agree that it would be more useful to return the nodes of the actual windowed field.

glwagner:
I think that would be a great addition. Windowing is design to be fully general, but is currently mostly used for "slices" (ie taking a single vertical level, plus the entirety of the other dimensions). As a result, this limitation of `node` hasn't come up. But if we want to properly support windowing in more than one direction, we need a corresponding `nodes` that extracts nodal values on the `Field`'s indices.

tomchor:
Okay, I'll probably tackle that soon then. I'll wait for some of PRs get merged since they're starting to pile up.

For context, this came up while coding https://github.com/CliMA/Oceananigans.jl/pull/4730, but I suspect it can also make the code for our Makie extension simpler.

glwagner:
I actually think that fields which are windowed in two dimensions cannot be plotted with the Makie extension. We already use: https://github.com/CliMA/Oceananigans.jl/blob/52847291258d1abd881f0374457a7a12ec027a2d/ext/OceananigansMakieExt.jl#L150

and we only accommodate windowing in 1D.

However, an _additional_ simplification would be to also adapt the tuple of coordinates; eg returning just two coordinates for 2D fields, etc. This is not an issue of windowing but rather an issue of inferring dimensionality via either `Flat` topology or `Nothing` location. I think the windowing will be simple to implement, but improvements to how we conceive of field dimensionality may run deeper and require more effort to fully clean up.

navidcy:
[...]

PR #4814, opened 2025-09-25
Make `nodes()` return the correct indices when passed a `view` of a `Field`

tomchor:
Resolves https://github.com/CliMA/Oceananigans.jl/issues/4729

In addition to improving the behavior of `nodes()`, this PR, together with https://github.com/CliMA/Oceananigans.jl/pull/4730/, makes it possible to write different slices of a `Field` (or several `Field`s) into one NetCDF file.

### DIFF
--- a/docs/src/fields.md
+++ b/docs/src/fields.md
@@ -76,7 +76,7 @@ of the primary mesh. The nodes of the primary mesh are
 znodes(grid, Center(), with_halos=true)
 
 # output
-6-element OffsetArray(::Vector{Float64}, 0:5) with eltype Float64 with indices 0:5:
+6-element view(OffsetArray(::Vector{Float64}, 0:5), :) with eltype Float64 with indices 0:5:
  -0.05
   0.05
   0.2
@@ -94,7 +94,7 @@ Finally, note that the nodes of the staggered mesh coincide with the cell interf
 znodes(grid, Center())
 
 # output
-4-element Vector{Float64}:
+4-element view(::Vector{Float64}, 2:5) with eltype Float64:
  0.05
  0.2
  0.44999999999999996
--- a/test/test_field.jl
+++ b/test/test_field.jl
@@ -115,9 +115,6 @@ function run_field_reduction_tests(grid)
     return nothing
 end
 
-@inline interpolate_xyz(x, y, z, from_field, from_loc, from_grid) =
-    interpolate((x, y, z), from_field, from_loc, from_grid)
-
 # Choose a trilinear function so trilinear interpolation can return values that
 # are exactly correct.
 @inline func(x, y, z) = convert(typeof(x), exp(-1) + 3x - y/7 + z + 2x*y - 3x*z + 4y*z - 5x*y*z)


glwagner:
Should `nodes` return a `KernelFunctionOperation`? If it did, would this PR be easier?

tomchor:
> Should `nodes` return a `KernelFunctionOperation`?

Not sure. I imagine this being more cumbersome for the end user, but possibly reducing code and makings things more flexible.

> If it did, would this PR be easier?

A little, yes. But I don't think enough as to motivate this change by itself.

src/Fields/field.jl
```diff
@@ -859,3 +859,9 @@ function fill_halo_regions!(field::Field, positional_args...; kwargs...)
 
     return nothing
 end
+
+#####
+##### nodes
+#####
+
+nodes(f::Field; kwargs...) = nodes(f.grid, instantiated_location(f)...; indices=indices(f), kwargs...)
```
simone-silvestri:
```suggestion
nodes(f::Field; kwargs...) = nodes(f.grid, instantiated_location(f)...; indices=indices(f), kwargs...)

```

src/Grids/latitude_longitude_grid.jl
```diff
@@ -647,15 +647,23 @@ end
...
+@inline λnodes(grid::LLG, ℓx, ℓy, ℓz; with_halos=false, indices=Colon()) = λnodes(grid, ℓx; with_halos, indices)
+@inline φnodes(grid::LLG, ℓx, ℓy, ℓz; with_halos=false, indices=Colon()) = φnodes(grid, ℓy; with_halos, indices)
 @inline xnodes(grid::LLG, ℓx, ℓy, ℓz; with_halos=false) = xnodes(grid, ℓx, ℓy; with_halos)
 @inline ynodes(grid::LLG, ℓx, ℓy, ℓz; with_halos=false) = ynodes(grid, ℓy; with_halos)
 
-@inline λnodes(grid::LLG, ℓx::F; with_halos=false) = _property(grid.λᶠᵃᵃ, ℓx, topology(grid, 1), grid.Nx, grid.Hx, with_halos)
-@inline λnodes(grid::LLG, ℓx::C; with_halos=false) = _property(grid.λᶜᵃᵃ, ℓx, topology(grid, 1), grid.Nx, grid.Hx, with_halos)
-@inline φnodes(grid::LLG, ℓy::F; with_halos=false) = _property(grid.φᵃᶠᵃ, ℓy, topology(grid, 2), grid.Ny, grid.Hy, with_halos)
-@inline φnodes(grid::LLG, ℓy::C; with_halos=false) = _property(grid.φᵃᶜᵃ, ℓy, topology(grid, 2), grid.Ny, grid.Hy, with_halos)
+@inline λnodes(grid::LLG, ℓx::F; with_halos=false, indices=Colon()) = getindex(_property(grid.λᶠᵃᵃ, ℓx, topology(grid, 1), grid.Nx, grid.Hx, with_halos), indices)
```
simone-silvestri: I guess this will not work on GPU right? Probably it will return a scalar indexing issue

simone-silvestri (approved):

src/Fields/field.jl
tomchor: I think I'm missing something; What's different here?

src/Grids/latitude_longitude_grid.jl
tomchor: Yeah I think that's why GPU tests are failing. I'll try to come up with an alternative soon

src/Fields/field.jl
simone-silvestri: the extra line at the bottom of the file

glwagner: he added whitespace below

tomchor: Ah, okay. Are we trying to always leave an extra line at the bottom of files?

glwagner:
After https://github.com/CliMA/Oceananigans.jl/pull/4805, it will be pretty trivial to return `KernelFunctionOperation` for `nodes`. Do you want me to wait for this PR to do that, or should I do that before this PR so you don't have to worry about CPU vs GPU for nodes?

glwagner:
Basically if `nodes` return `KernelFunctionOperation`, then the way to get correct indices will be the same as building a `Field` and passing `indices` to the `Field` constructor. We won't need any "special" functionality for windowing the nodes, compared to windowing `Field`.

tomchor:
> After #4805, it will be pretty trivial to return `KernelFunctionOperation` for `nodes`. Do you want me to wait for this PR to do that, or should I do that before this PR so you don't have to worry about CPU vs GPU for nodes?

You can go ahead and make `nodes` return `KernelFunctionOperation` first!

PR #4819, opened 2025-09-27
Further enhancements of the XESMF extension

### LINKED ISSUE #4818: I'm merging this as a good starting point and will continue working on it.
I'm merging this as a good starting point and will continue working on it.

Some outstanding issues are:
- Cleanup of:
  https://github.com/CliMA/Oceananigans.jl/blob/21ba055122b3fadd3e90680ca3b49dd80f8f14de/ext/OceananigansXESMFExt.jl#L12-L29
- Ensure all is smooth with GPU architectures; add tests for GPU (perhaps move from GitHub actions to buildkite)
- Does the regridding reproduces the Python xESMF regridding?
- (Perhaps related to the above) Can the integral before and after regridding be better than `rtol = 1e-4`?
- Documentation/Example for how to do a regridding using XESMF.jl (docstrings with examples should suffice for now; convert some example(s) to doctest(s)).

_Originally posted by @navidcy in https://github.com/CliMA/Oceananigans.jl/issues/4782#issuecomment-3341573444_

navidcy:
This PR further enhances the XESMF extension.

- [x] Cleanup of:
  https://github.com/CliMA/Oceananigans.jl/blob/ff6d022f280e2d1d9b680843857bc6fffa637645/ext/OceananigansXESMFExt.jl#L12-L29
- [x] Ensure all is smooth with GPU architectures; add tests for GPU (perhaps move from GitHub actions to buildkite)
- [ ] Does the regridding reproduces the Python xESMF regridding?
- [ ] (Perhaps related to the above) Can the integral before and after regridding be better than `rtol = 1e-4`?
- [ ] Documentation/Example for how to do a regridding using XESMF.jl (docstrings with examples should suffice for now; convert some example(s) to doctest(s)).

Partially deals with #4818


test/test_xesmf.jl
```diff
@@ -4,44 +4,53 @@ using XESMF
...
+                                        z, radius)
 
-tg = TripolarGrid(; size=(360, 170, 1), z, southernmost_latitude, radius)
+        llg_fine = LatitudeLongitudeGrid(arch; size=(360, 170, 1),
+                                        longitude=(0, 360),
+                                        latitude=(southernmost_latitude, 90),
+                                        z, radius)
 
-@testset "XESMF extension" begin
+        tg = TripolarGrid(arch; size=(360, 170, 1), z, southernmost_latitude, radius)
```
xkykai, line 25: I wonder if one should add tests for 3D fields too? i.e. `TripolarGrid(..., size=(360, 170, 2))` then check if the two layers agree approximately. But I am fine not testing the 3D fields too!

navidcy: Let's merge this and we can have a meaningful test later? I want to also optimise the part of `regrid!` method because a serial for doesn't seem appropriate.

xkykai: Sounds good to me!

PR #4811, opened 2025-09-23
(0.104.0) Reformulate hydrostatic model timestepping

simone-silvestri:
This is a work-in-progress PR that aims to reformulate the timestepping scheme of the `HydrostaticFreeSurfaceModel` to 
make sure that three numerical properties are satisfied:

- tracers are globally conserved (the integral of a tracer field remains constant in the absence of forcings)
- tracers are locally conserved (an initially constant tracer remains constant in the absence of forcings)
- The free surface is evolved consistently with the grid thickness.

This is easier to achieve with an explicit discretization of the free surface (`ExplicitFreeSurface` and `SplitExplicitFreeSurface`) rather than an implicit free surface discretization and with RK3 rather than AB2.
However, this PR reformulates much of the timestepping algorithm, so it will be a lengthy WIP that will need to be thoroughly validated for all combinations of barotropic and baroclinic timestepping schemes.

More changes are performed to correctly abstract the timestepping schemes, which, at the moment, are tailored specifically to the `NonhydrostaticModel` (for example, the timesteppers include a `compute_pressure_correction!` or a ` correct_velocities` function, which are meaningful only for a non-hydrostatic Navier Stokes solver)

Edit: another objective I have is to remove any timestepping from the `update_state!` function. We should be able to call `update_state!` multiple times in a row without problems and timestepping inside `update_state!` negates this possibility.

TODO:
- [ ] <s> Remove timestepping from `update_state!` </s> In another PR
- [x] Clean up the time-stepping code
- [x] Introduce transport weights for transport averaging in the SplitExplicitFreeSurface
- [x] Implement a conservative, low-storage Runge Kutta method (of arbitrary order) using the Explicit free surface
- [x] Implement a conservative, low-storage Runge Kutta method (of arbitrary order) coupled with the SplitExplicitFreeSurface
- [x] Implement a conservative, low-storage Runge Kutta method (of arbitrary order) using the (PCG) ImplicitFreeSurface
- [x] Ensure that the conservation is satisfied on `DistributedGrids`
- [ ] <s> Implement a semi - conservative method for AB2  </s> In another PR
- [ ] <s> Address precision errors that tend to destroy conservation for constant fields </s> In another PR

PR #4823, opened 2025-09-30
More work/enhancements of the XESMF extension

navidcy:
- [ ] Further cleanup of:
  https://github.com/CliMA/Oceananigans.jl/blob/ff6d022f280e2d1d9b680843857bc6fffa637645/ext/OceananigansXESMFExt.jl#L12-L29
- [ ] Does the regridding reproduces the Python xESMF regridding?
- [ ] (Perhaps related to the above) Can the integral before and after regridding be better than `rtol = 1e-4`?
- [ ] Documentation/Example for how to do a regridding using XESMF.jl (docstrings with examples should suffice for now; convert some example(s) to doctest(s)).

Partially deals with #4818

### DIFF
--- a/ext/OceananigansXESMFExt.jl
+++ b/ext/OceananigansXESMFExt.jl
@@ -9,22 +9,20 @@ using Oceananigans.Grids: λnodes, φnodes, Center, Face, total_length
 import Oceananigans.Fields: regrid!
 import XESMF: Regridder, extract_xesmf_coordinates_structure
 
-# permutedims below is used because Python's xESMF expects
-# 2D arrays with (x, y) coordinates with y varying in dim=1 and x varying in dim=2
+node_array(ξ::AbstractMatrix, Nx, Ny) = view(ξ, 1:Nx, 1:Ny)
 
-node_array(ξ::AbstractMatrix, Nx, Ny) = permutedims(view(ξ, 1:Nx, 1:Ny), (2, 1))
-vertex_array(ξ::AbstractMatrix, Nx, Ny) = permutedims(view(ξ, 1:Nx+1, 1:Ny+1), (2, 1))
-
-x_node_array(x::AbstractVector, Nx, Ny) = permutedims(repeat(view(x, 1:Nx), 1, Ny), (2, 1))
+x_node_array(x::AbstractVector, Nx, Ny) = repeat(view(x, 1:Nx), 1, Ny)
 x_node_array(x::AbstractMatrix, Nx, Ny) = node_array(x, Nx, Ny)
 
-y_node_array(y::AbstractVector, Nx, Ny) = repeat(view(y, 1:Ny), 1, Nx)
+y_node_array(y::AbstractVector, Nx, Ny) = repeat(transpose(view(y, 1:Ny)), Nx, 1)
 y_node_array(y::AbstractMatrix, Nx, Ny) = node_array(y, Nx, Ny)
 
-x_vertex_array(x::AbstractVector, Nx, Ny) = permutedims(repeat(view(x, 1:Nx+1), 1, Ny+1), (2, 1))
+vertex_array(ξ::AbstractMatrix, Nx, Ny) = view(ξ, 1:Nx+1, 1:Ny+1)
+
+x_vertex_array(x::AbstractVector, Nx, Ny) = repeat(view(x, 1:Nx+1), 1, Ny+1)
 x_vertex_array(x::AbstractMatrix, Nx, Ny) = vertex_array(x, Nx, Ny)
 
-y_vertex_array(y::AbstractVector, Nx, Ny) = repeat(view(y, 1:Ny+1), 1, Nx+1)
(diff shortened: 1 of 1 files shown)


navidcy:
@simone-silvestri, @glwagner:

Could we optimise the for loop for k at:

https://github.com/CliMA/Oceananigans.jl/blob/cbfb6f24fd0c6852e730f49904ed8b27c6d033ee/ext/OceananigansXESMFExt.jl#L179-L191

?

glwagner:
> @simone-silvestri, @glwagner:
> 
> Could we optimise the for loop for k at:
> 
> https://github.com/CliMA/Oceananigans.jl/blob/cbfb6f24fd0c6852e730f49904ed8b27c6d033ee/ext/OceananigansXESMFExt.jl#L179-L191
> 
> ?

I can't think of anything

simone-silvestri:
We could leverage batched multiplies from the CUBLAS and BLAS libraries.
Something like this https://github.com/FluxML/NNlib.jl/blob/13d901d130d29662d81bbc6c44c67a4513182002/ext/NNlibCUDAExt/batchedmul.jl#L3

navidcy:
Related to reproducing python's xESMF results: https://forum.access-hive.org.au/t/conservative-regridding-with-xesmf-puzzled-because-the-integrals-dont-quite-agree/5343

docs/make.jl
```diff
@@ -184,7 +184,7 @@ for m in [Oceananigans, OceananigansNCDatasetsExt, OceananigansXESMFExt]
     end
 end
 
-makedocs(sitename = "Oceananigans.jl",
+makedocs(; sitename = "Oceananigans.jl",
          authors = "Climate Modeling Alliance and contributors",
          format = format,
          pages = pages,
```
glwagner:
```suggestion
makedocs(sitename = "Oceananigans.jl"; format, pages,
         authors = "Climate Modeling Alliance and contributors",
```

glwagner:
@juliasloan25 once this PR is in, you will probably be ready to use XESMF for tripolar -> lat lon interpolation. (@navidcy can say more maybe). You might be able to use it now.

navidcy:
You can use it now! This PR cleans up some pf the code and adds up examples. But doesn’t add functionality. 

I wanted to confirm that we reproduce pythons xesmf results. I’m pretty sure we do. Im gonna wrap that up and include it.

bottom line: no need to wait for this PR @juliasloan25

navidcy:
I confirmed that the regridder here reproduces python's results.

In particular:

```Julia
using Oceananigans
using XESMF

arch = CPU()

z = (-1, 0)

radius = Oceananigans.defaults.planet_radius

llg_coarse = LatitudeLongitudeGrid(arch; z, radius,
                                   size = (180, 90, 1),
                                   longitude = (0, 360),
                                   latitude = (-90, 90))

llg_fine = LatitudeLongitudeGrid(arch; z, radius,
                                 size = (360, 180, 1),
                                 longitude = (0, 360),
                                 latitude = (-90, 90))

src_field = CenterField(llg_coarse)
dst_field = CenterField(llg_fine)

width = 12         # degrees
set!(src_field, (λ, φ, z) -> exp(-((λ - 150)^2 + (φ - 30)^2) / 2width^2) - 2exp(-((λ - 270)^2 + (φ + 20)^2) / 2width^2))

regridder = XESMF.Regridder(dst_field, src_field, method="conservative")

regrid!(dst_field, regridder, src_field)

println("integral src_field = ", first(Field(Integral(src_field, dims=(1, 2)))))
println("integral dst_field = ", first(Field(Integral(dst_field, dims=(1, 2)))))
```

gives

```Julia
integral src_field = -1.1089929769375445e13
integral dst_field = -1.109066523206981e13
```

And in Python, here's a Jupyter notebook getting exactly the same numbers:

[python-xesmf.pdf](https://github.com/user-attachments/files/22938987/python-xesmf.pdf)

navidcy:
This is ready to review/merge. Could somebody have a look?

simone-silvestri:
wow, not so conservative after all! (with zstar at #4811 I am stressing over a relative error in tracer conservation of 1e-13 😅, maybe I don't need to be so pedantic...)

simone-silvestri:
Shall we add the validation somewhere in the XESMF.jl package?

navidcy:
You mean the Python validation?
Because otherwise it's part of the test. Perhaps not exactly...

I can add both in a validation script.

I'm amazed how it's not giving the same integral to machine precision. But I don't know how xESMF computes their weights. Do they compute them in a way to ensure that the integrals are the same?

simone-silvestri:
I would have suspected so. In general, if they compute intersections between surface areas it should be quite simple to ensure that the integral is conserved. Maybe they assume the areas are planes rather than spherical sectors? This could introduce an error. If this is the case, a rectilinear grid should lead to exact conservation.

navidcy:
The grid I tried didn't have overlapping areas. I used a lat-lon grid of 1deg and 2deg.
But I also computed the spherical area of the two grids and I got exactly 4πR^2 both for the source and for the destination grid... So I'm confused.

navidcy:
I added the validation scripts; I'll merge when CI passes.

PR #4833, opened 2025-10-06
Move defaults to central location for gravitational acceleration, rotation rate, and planet radius

glwagner:
This PR expands the existing `Defaults` struct to include `gravitational_acceleration`, `planet_rotation_rate`, and `planet_radius`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.100.0"
+version = "0.100.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
@@ -63,7 +63,7 @@ OceananigansXESMFExt = ["XESMF"]
 [compat]
 AMDGPU = "1.3.6, 2"
 Adapt = "4.1.1"
-CUDA = "5.7"
+CUDA = "5.9"
 ConstructionBase = "1"
 Crayons = "4"
 CubedSphere = "0.2, 0.3"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -15,7 +15,7 @@ StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 TimesDates = "bdfc003b-8df8-5c39-adcd-3a9087f5df4a"
 
 [compat]
-CUDA = "5.4 - 5.8.2"
+CUDA = "5.9"
 CairoMakie = "0.11, 0.12, 0.13, 0.14"
 Documenter = "1"
 DocumenterCitations = "1"


src/Oceananigans.jl
```diff
@@ -149,9 +149,22 @@ const fully_supported_float_types = (Float32, Float64)
...
+end
+
+function Defaults(;
+    FloatType = Float64,
+    # [m s⁻²] conventional standard value for Earth's gravity https://en.wikipedia.org/wiki/Gravitational_acceleration#Gravity_model_for_Earth
+    gravitational_acceleration = 9.80665,
+    # [m] Earth's radius; see https://en.wikipedia.org/wiki/Earth%27s_radius
+    planet_radius = 6.371e6,
+    # [s⁻¹] Earth's angular speed; see https://en.wikipedia.org/wiki/Earth%27s_rotation#Angular_speed
+    planet_rotation_rate = 7.292115e-5)
```
navidcy: why call it "planet" which is ambiguous?

```suggestion
    Earth_rotation_rate = 7.292115e-5)       
```

glwagner: In general it isn't Earth's rotation rate if the number is changed. It is the rotation rate of a hypothetical planet from which the Coriolis force is derived

glwagner: the default value corresponds to Earth

navidcy: true; agreed! ignore my suggestion, I take it back

navidcy:
Is this a breaking change (eg for packages that depend on Oceananigans?). I'm thinking ClimaSeaIce, ClimaOcean..

navidcy (approved):
let's consider tag new patch release (or minor if it's breaking?)

src/Coriolis/hydrostatic_spherical_coriolis.jl
```diff
@@ -18,19 +18,20 @@ end
...
                                  scheme = EnstrophyConserving())
 
 Return a parameter object for Coriolis forces on a sphere rotating at `rotation_rate`.
 
 Keyword arguments
 =================
 
-- `rotation_rate`: Sphere's rotation rate; default: [`Ω_Earth`](@ref).
+- `rotation_rate`: Sphere's rotation rate;
+   default: [`Oceananigans.defaults.planet_rotation_rate`](@ref).
```
navidcy: does `Oceananigans.defaults.planet_rotation_rate` have a docstring? the `(@ref)` implies that there is a docstring?

navidcy (approved):

simone-silvestri:
It shouldn't be breaking, right? It just moves the defaults from the constructor definition to a global variable. Users can still modify individual types with kwargs.

simone-silvestri (approved):

glwagner:
> Is this a breaking change (eg for packages that depend on Oceananigans?). I'm thinking ClimaSeaIce, ClimaOcean..

No, this is not a breaking change.

navidcy:
merge?

ali-ramadhan:
> > Is this a breaking change (eg for packages that depend on Oceananigans?). I'm thinking ClimaSeaIce, ClimaOcean..
> 
> No, this is not a breaking change.

I think this is technically a breaking change, e.g. https://github.com/OceanBioME/OceanBioME.jl/issues/299, but since `g_Earth` was not exported then it was not a breaking change to the public user interface or API and thus did not warrant bumping the minor version in accordance with [semantic versioning](https://semver.org/).

EDIT: Actually reading over the semver spec again, it does not mandate updating the Y in 0.Y.z when breaking changes are introduced, but I believe we have been trying to follow this rule because Julia `[compat]` bounds treat updating the Y in `0.Y.z` as a potentially breaking change.

PR #4836, opened 2025-10-09
Update CI to Julia version to 1.12.0

navidcy:


simone-silvestri:
I am very interested in this. Let's hope it works and we can move on from julia 1.10

simone-silvestri:
I am disabling the reactant tests for the moment to check if the rest works.

simone-silvestri:
If docs still break on the `internal_tide.jl` example `NaN`ing, I would try with a `GridFittedBottom` instead of a `PartialCellBottom`, which is not even tested, and, if tests pass, merge this PR and move forward trying to assess what the problems with partial cells are.

simone-silvestri:
Seems that we are hitting the same NaN issue on the internal tide example

navidcy:
> Seems that we are hitting the same NaN issue on the internal tide example

the ghosts of the past still haunt us....

simone-silvestri:
Apparently also `GridFittedBottom` fails...

simone-silvestri:
If I run the example locally, it works. Why would it error on CI? Do we have a way to reproduce this error locally?

ali-ramadhan:
> If I run the example locally, it works. Why would it error on CI? Do we have a way to reproduce this error locally?

One thing to try might be to run the example locally and on CI using the exact same Manifest.toml if possible. We can commit a Manifest.toml to this branch for debugging. I can't think of which dependency would lead to such a big difference but it's one thing we can control for.

navidcy:
From the [Julia v1.11 chat](#3836) I recall that the error was showing up only for unix, not for mac?

giordano:
With this environment (manifests for v1.11 and v1.12 both included) [`environment.tar.gz`](https://github.com/user-attachments/files/23153216/environment.tar.gz), https://github.com/CliMA/Oceananigans.jl/blob/ea25179c7af868175fc295c2bf1dbfe78ec3cd4f/examples/internal_tide.jl works on macOS but not on Linux (tested only Ubuntu so far), doesn't seem to be a matter of package versions.

giordano:
I can make the simulation error early with
```diff
diff --git a/src/Diagnostics/nan_checker.jl b/src/Diagnostics/nan_checker.jl
index 57945c5dc..893a9e283 100644
--- a/src/Diagnostics/nan_checker.jl
+++ b/src/Diagnostics/nan_checker.jl
@@ -5,7 +5,7 @@ mutable struct NaNChecker{F}
     erroring :: Bool
 end
 
-NaNChecker(fields) = NaNChecker(fields, false) # default
+NaNChecker(fields) = NaNChecker(fields, true) # default
 default_nan_checker(model) = nothing
 
 function Base.summary(nc::NaNChecker)
@@ -28,7 +28,7 @@ a container with key-value pairs like a dictionary or `NamedTuple`.
 
 If `erroring=true`, the `NaNChecker` will throw an error on NaN detection.
 """
-NaNChecker(; fields, erroring=false) = NaNChecker(fields, erroring)
+NaNChecker(; fields, erroring=true) = NaNChecker(fields, erroring)
 
 hasnan(field::AbstractArray) = any(isnan, parent(field))
 hasnan(model) = hasnan(first(fields(model)))
```
I presume there's also a way to set the `NaNChecker` explicitly for a simulation, but I couldn't quickly figure it out.

Can we use a callback to print out to file all the steps, so that we can compare 1:1 the progress on different machines?  Presumably we're initially interested in the field `u`, that's what the error is about at least:
```
julia> run!(simulation)
[ Info:     ... simulation initialization complete (17.810 ms)
[ Info:     ... initial time step complete (9.452 ms).
ERROR: time = 60000.0, iteration = 200: NaN found in field u. Aborting simulation.
```

giordano:
Before https://github.com/CliMA/Oceananigans.jl/blob/ea25179c7af868175fc295c2bf1dbfe78ec3cd4f/examples/internal_tide.jl#L175 I see
```julia-repl
julia> simulation.model.velocities.u
256×1×128 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 256×1×128 ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Periodic, east: Periodic, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 264×1×136 OffsetArray(::Array{Float64, 3}, -3:260, 1:1, -3:132) with eltype Float64 with indices -3:260×1:1×-3:132
    └── max=0.281029, min=0.281029, mean=0.281029
```
on both machines, if I'm looking at the right field and this display says enough about it, then they're the same at the beginning, but then on macOS I have
```julia-repl
julia> time_step!(simulation); simulation.model.velocities.u
[ Info: Iter: 0, time: 0 seconds, wall time: 2.256 minutes, max|w|: 2.089e-03, m s⁻¹
[ Info:     ... simulation initialization complete (887.307 ms)
[ Info:     ... initial time step complete (128.489 ms).
256×1×128 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 256×1×128 ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Periodic, east: Periodic, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 264×1×136 OffsetArray(::Array{Float64, 3}, -3:260, 1:1, -3:132) with eltype Float64 with indices -3:260×1:1×-3:132
    └── max=0.31715, min=0.265116, mean=0.280967

julia> time_step!(simulation); simulation.model.velocities.u
256×1×128 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 256×1×128 ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Periodic, east: Periodic, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 264×1×136 OffsetArray(::Array{Float64, 3}, -3:260, 1:1, -3:132) with eltype Float64 with indices -3:260×1:1×-3:132
    └── max=0.335864, min=0.264486, mean=0.280859
```
and on Ubuntu
```julia-repl
julia> time_step!(simulation); simulation.model.velocities.u
[ Info: Iter: 0, time: 0 seconds, wall time: 2.391 minutes, max|w|: 2.089e-03, m s⁻¹
[ Info:     ... simulation initialization complete (1.130 seconds)
[ Info:     ... initial time step complete (20.645 ms).
256×1×128 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 256×1×128 ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Periodic, east: Periodic, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 264×1×136 OffsetArray(::Array{Float64, 3}, -3:260, 1:1, -3:132) with eltype Float64 with indices -3:260×1:1×-3:132
    └── max=0.31715, min=0.265116, mean=0.280967

julia> time_step!(simulation); simulation.model.velocities.u
256×1×128 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 256×1×128 ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Periodic, east: Periodic, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 264×1×136 OffsetArray(::Array{Float64, 3}, -3:260, 1:1, -3:132) with eltype Float64 with indices -3:260×1:1×-3:132
    └── max=0.333478, min=0.264645, mean=0.280863
```
so there's a significant divergence already after two timesteps.

***Update***:
```julia-repl
julia> time_step!(simulation); simulation.model.velocities.u
[ Info: Iter: 0, time: 0 seconds, wall time: 2.269 minutes, max|w|: 2.089e-03, m s⁻¹
[ Info:     ... simulation initialization complete (11.788 seconds)
[ Info:     ... initial time step complete (12.640 seconds).
256×1×128 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 256×1×128 ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Periodic, east: Periodic, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 264×1×136 OffsetArray(::Array{Float64, 3}, -3:260, 1:1, -3:132) with eltype Float64 with indices -3:260×1:1×-3:132
    └── max=0.31715, min=0.265116, mean=0.280967

julia> time_step!(simulation); simulation.model.velocities.u
256×1×128 Field{Face, Center, Center} on ImmersedBoundaryGrid on CPU
├── grid: 256×1×128 ImmersedBoundaryGrid{Float64, Periodic, Flat, Bounded} on CPU with 4×0×4 halo
│   └── west: Periodic, east: Periodic, south: Nothing, north: Nothing, bottom: ZeroFlux, top: ZeroFlux, immersed: ZeroFlux
└── data: 264×1×136 OffsetArray(::Array{Float64, 3}, -3:260, 1:1, -3:132) with eltype Float64 with indices -3:260×1:1×-3:132
    └── max=0.335864, min=0.264486, mean=0.280859
```
is also what I see on Ubuntu with Julia v1.10, which is consistent with all versions of Julia on macOS.

giordano:
The plot thickens: it works correctly in Julia v1.12 on Ampere eMAG (aarch64) with AlmaLinux 8.10 as operating system, which rules out an operating system difference.  aarch64 is also the architecture on macOS, so I'm starting to suspect there's an architecture dependence.  Can someone point me to the operation performed on the `u` field?  Is there any chance there you're running any BLAS operation?  ***Edit***: using in x86-64 Julia v1.12 the libopenblas shipped in Julia v1.10 doesn't fix the issue, so that'd rule out BLAS version differences.  ***Edit 2***: also works on Intel(R) Xeon(R) CPU X5670 with CentOS 8 Stream, I'm getting more and more confused, sigh.  Pointers to the relevant code would still be very welcome.

glwagner:
> The plot thickens: it works correctly in Julia v1.12 on Ampere eMAG (aarch64) with AlmaLinux 8.10 as operating system, which rules out an operating system difference. aarch64 is also the architecture on macOS, so I'm starting to suspect there's an architecture dependence. Can someone point me to the operation performed on the `u` field? Is there any chance there you're running any BLAS operation? _**Edit**_: using in x86-64 Julia v1.12 the libopenblas shipped in Julia v1.10 doesn't fix the issue, so that'd rule out BLAS version differences. _**Edit 2**_: also works on Intel(R) Xeon(R) CPU X5670 with CentOS 8 Stream, I'm getting more and more confused, sigh. Pointers to the relevant code would still be very welcome.

Nice work so far though!!

The entire time-step is a complex chain of operations. I do think it is a good start to save down all fields every time-step. We may find that differences arise in one field versus another. Note that the NaNChecker checks `u` only as a proxy for the entire state. Here the prognostic state should be `model.velocities` (u, v, w) and `model.tracers.b`.

glwagner:
To save every iteration chnage this line

https://github.com/CliMA/Oceananigans.jl/blob/ea25179c7af868175fc295c2bf1dbfe78ec3cd4f/examples/internal_tide.jl#L170

to `schedule = IterationInterval(1)`. Also I think we should add `v` (I see `u` and `w` but not `v` there).

The difference should arise in the very first time-step? We could compare those. It seems annoying laborious to do this across architectures, but maybe @giordano you have good ideas how to do this efficiently

giordano:
We made some progress during a pair-debugging session with @simone-silvestri (during which we discovered the typo fixed by #4901), we found that already after the first step the pressure is different, which is updated by https://github.com/CliMA/Oceananigans.jl/blob/4265addd082f0462dfd4fc7114b89efa2a420ece/src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl#L12-L20 Replacing `z_dot_g_bᶜᶜᶠ` with
```julia
@inline my_z_dot_g_bᶜᶜᶠ(i, j, k, grid::Oceananigans.Grids.AbstractGrid{FT}, bf, C) where FT = FT(0.5) * (C.b[i, j, k] + C.b[i, j, k-1])
```
seems to be a workaround, but that's puzzling because that's pretty much the same as https://github.com/CliMA/Oceananigans.jl/blob/5b9538b6fa781e50277f72e1c4659145e6c0845d/src/Operators/interpolation_operators.jl#L15 minus the `@inbounds`.  Putting the `@inbounds` in the definition of `my_z_dot_g_bᶜᶜᶠ` makes the pressure diverge and the `NaN`s pop up again.  I still don't have a full explanation of what's happening though, nor why this happens only on some systems.

giordano:
Issue with Julia v1.11+ seems to be fixed by https://github.com/JuliaGPU/KernelAbstractions.jl/pull/653, resolved by @vchuravy independently from Oceananigans troubles, but very timely nonetheless 😁

simone-silvestri:
I will switch to a local accumulator in that kernel like we did in the `compute_w_from_continuity!` function so we can merge this PR independently on the fix.

simone-silvestri:
It worked! The simulation does not produce NaN anymore. Now we have other problems to solve, but they all seem easier.

src/Models/NonhydrostaticModels/update_hydrostatic_pressure.jl
```diff
@@ -12,10 +12,12 @@ the `buoyancy_perturbationᶜᶜᶜ` downwards:
...
     i, j = @index(Global, NTuple)
 
-    @inbounds pHY′[i, j, grid.Nz] = - z_dot_g_bᶜᶜᶠ(i, j, grid.Nz+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, grid.Nz+1, grid)
+    pᵏ = - z_dot_g_bᶜᶜᶠ(i, j, grid.Nz+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, grid.Nz+1, grid)
+    @inbounds pHY′[i, j, grid.Nz] = pᵏ
 
     for k in grid.Nz-1 : -1 : 1
-        @inbounds pHY′[i, j, k] = pHY′[i, j, k+1] - z_dot_g_bᶜᶜᶠ(i, j, k+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, k+1, grid)
+        pᵏ -= z_dot_g_bᶜᶜᶠ(i, j, k+1, grid, buoyancy, C) * Δzᶜᶜᶠ(i, j, k+1, grid)
+        @inbounds pHY′[i, j, k] = pᵏ
```
giordano, line 20: If we require KernelAbstractions.jl v0.9.39 we don't need the workaround anymore.

navidcy: https://github.com/CliMA/Oceananigans.jl/pull/4836/commits/82caad073e70aa4cf9955a843ac953901692e898

giordano:
Doctests failures are due to https://github.com/CliMA/Oceananigans.jl/pull/4840#discussion_r2487853689

navidcy:
I disabled the doctests temporarily. If everything else works, I'm happy to work through fixing the doctests!

simone-silvestri:
doctests should be working now. The last problems are connected to reactant and abstract operations on GPU

giordano:
Reactant doesn't fully support Julia v1.12 yet (same for Enzyme.jl, but not sure that's relevant for Oceananigans tests).

simone-silvestri:
Shall we just move the tests to julia 1.11 then until Reactant and Enzyme are ready?

giordano:
How easy would it be to run Reactant/Enzyme tests only with 1.11 in Buildkite? It shouldn't be hard in github actions, but I'm less familiar with buildkite.

simone-silvestri:
I can try giving it a go. We can actually add another pipeline, and probably need a new queue. @ali-ramadhan would you know how to set this up on nautilus if I separate out the tests in a new pipeline?

simone-silvestri:
I think I have managed to separate them out.

simone-silvestri:
I think we need to create a new Manifest for version 11

.buildkite/pipeline.yml
```diff
@@ -24,13 +25,31 @@ steps:
       mkdir -p "$OCEANANIGANS_DIR"
       rsync -a . "$OCEANANIGANS_DIR/"
       cd "$OCEANANIGANS_DIR"
+      mv Manifest.toml Manifest-v12.0.toml
```
giordano:
```suggestion
      mv Manifest.toml Manifest-v1.12.toml
```

giordano:
Why is there a manifest?  Anything that can be accomplished by using the [`[sources]`](https://pkgdocs.julialang.org/v1/toml-files/#The-[sources]-section) section of the project file instead?  That makes many things a lot easier.

simone-silvestri:
Apparently the manifest is generated automatically. @giordano do you know how to achieve running from the same folder from two different julia versions using `[sources]`?

giordano:
Possibly, but I need to understand what information is in manifest which isn't in the Project.toml.  Is it the path to the top-level Oceananigans?  Where's the Manifest you're fighting with?

simone-silvestri:
I think we just need the manifest that is automatically generated by 
```julia
julia +$JULIA_VERSION_ENZYME -O0 --color=yes --project -e 'using Pkg; Pkg.test()'
```

I just naively added two initializations with different versions that led to this error http://buildkite.com/clima/oceananigans/builds/26543#019a4e6e-a2a6-47ac-8234-65f3c4986045
After which I thought in here
https://github.com/CliMA/Oceananigans.jl/blob/e776c216339cabdb7ce159affbb8d3d67e8100dc/.buildkite/pipeline.yml#L26
We were also copying a manifest. Apparently, this is not the case; the manifest is automatically generated looking from https://buildkite.com/clima/oceananigans/builds/26545#019a4e76-acae-454d-8940-789074cebe22. 

All information should already be in the project. However, I guess we need to distinguish between the 1.11 and 1.12 version to avoid  http://buildkite.com/clima/oceananigans/builds/26543#019a4e6e-a2a6-47ac-8234-65f3c4986045

.buildkite/pipeline.yml
```diff
@@ -99,14 +111,51 @@ steps:
...
+      fi
+
+      # Strip emoji for environment variable
+      group="{{ matrix.group }}"
+      export TEST_GROUP="\${group#* }"
+      echo $TEST_GROUP
+
+      # Run tests
+      cd "$OCEANANIGANS_DIR"
+      julia +$JULIA_VERSIO_ENZYME -O0 --color=yes --project -e 'using Pkg; Pkg.test()'
```
giordano:
```suggestion
      julia +$JULIA_VERSION_ENZYME -O0 --color=yes --project -e 'using Pkg; Pkg.test()'
```

.buildkite/pipeline.yml
```diff
@@ -24,13 +25,31 @@ steps:
       mkdir -p "$OCEANANIGANS_DIR"
       rsync -a . "$OCEANANIGANS_DIR/"
       cd "$OCEANANIGANS_DIR"
+      mv Manifest.toml Manifest-v1.12.toml
```
giordano:
```suggestion
      touch Manifest-v$(echo "${JULIA_VERSION}" | cut -d. -f1-2).toml
```

.buildkite/pipeline.yml
```diff
@@ -24,13 +25,31 @@ steps:
...
+  - label: "🏕️ initialize enzyme environment"
+    key: "init_enzyme"
+    depends_on: "init"
+    agents:
+      queue: "Oceananigans-nautilus"
+    env:
+      TEST_GROUP: "init"
+    command: |
+      juliaup add $JULIA_VERSION_ENZYME
+      cd "$OCEANANIGANS_DIR"
```
giordano, line 47:
```suggestion
      cd "$OCEANANIGANS_DIR"
      touch Manifest-v$(echo "${JULIA_VERSION_ENZYME}" | cut -d. -f1-2).toml
```

.buildkite/pipeline.yml
```diff
@@ -99,14 +111,51 @@ steps:
...
+      fi
+
+      # Strip emoji for environment variable
+      group="{{ matrix.group }}"
+      export TEST_GROUP="\${group#* }"
+      echo $TEST_GROUP
+
+      # Run tests
+      cd "$OCEANANIGANS_DIR"
+      julia +$JULIA_VERSION_ENZYME -O0 --color=yes --project -e 'using Pkg; Pkg.test()'
```
giordano, line 148:
```suggestion
      touch Manifest-v$(echo "${JULIA_VERSION_ENZYME}" | cut -d. -f1-2).toml
      julia +$JULIA_VERSION_ENZYME -O0 --color=yes --project -e 'using Pkg; Pkg.test()'
```

giordano:
I think I'm missing something fundamental

1. are these builds not happening in clean directories?
2. what's generating these manifests?

giordano:
Alright, I believe I found the right incantation in 94bc181dd83d20f396c6bb30704baf1ab5cf5e0a.

The summary is

* the step `🏕️ initialize general environment` was instantiating the first environment with Julia v1.12, creating the `Manifest.toml`
* the step `🏕️ initialize enzyme environment` would try to use the `Manifest.toml` generated by the previous step (the two steps happen one after the other, in the same machine & directory) with Julia v1.11, failing.

My solution was to

* rename the manifest generated by the first step to the version-specific name _after_ instantiation
* also rename the Enzyme manifest to the version-specific name after instantiation.  This is strictly unnecessary if the previous step is already disambiguating the manifests, but having version-specific names everywhere is a more robust approach anyway.

navidcy:
The docs built and the internal tide example looks great!
https://clima.github.io/OceananigansDocumentation/previews/PR4836/literated/internal_tide/

There seem to be some errors in the  GPU abstract operations CI:

https://buildkite.com/clima/oceananigans/builds/26558#019a515f-fc97-4a52-96e9-a8a5894dffdd/7-53

If that smooths out, I'll work on bringing the doctests back in!

giordano:
https://buildkite.com/clima/oceananigans/builds/26676#019a73b9-9cc9-46be-adaf-f291b4ea1eca/37-407
```
WARNING: Imported binding XESMF.xesmf_coordinates was undeclared at import time during import to OceananigansXESMFExt.
ERROR: LoadError: invalid method definition in OceananigansXESMFExt: exported function XESMF.xesmf_coordinates does not exist
Stacktrace:
 [1] top-level scope
   @ ~/Oceananigans.jl-26676/ext/OceananigansXESMFExt.jl:29
 [2] include(mod::Module, _path::String)
   @ Base ./Base.jl:306
 [3] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2996
 [4] top-level scope
   @ stdin:5
 [5] eval(m::Module, e::Any)
   @ Core ./boot.jl:489
 [6] include_string(mapexpr::typeof(identity), mod::Module, code::String, filename::String)
   @ Base ./loading.jl:2842
 [7] include_string
   @ ./loading.jl:2852 [inlined]
 [8] exec_options(opts::Base.JLOptions)
   @ Base ./client.jl:315
 [9] _start()
   @ Base ./client.jl:550
in expression starting at /var/lib/buildkite-agent/Oceananigans.jl-26676/ext/OceananigansXESMFExt.jl:1
```

***Edit 1***: oh, is this fixed by #4926?  ***Edit 2***: uhm, no, this is already on top on current `main` :thinking: ***Edit 3***: oh, is this fixed by https://github.com/NumericalEarth/XESMF.jl/pull/20?

simone-silvestri:
Interesting how julia 1.10 does not error if an imported method does not exist but julia 1.12 does.

simone-silvestri:
I think I might have found the reason for the abstract operation test breaking. Apparently, the GPU compiler (and the CPU as well) do not really like this method which leads to dynamic evaluation

https://github.com/CliMA/Oceananigans.jl/blob/465335b26a5d0895beb89bddb82a56ce8bca515a/src/AbstractOperations/multiary_operations.jl#L16-L17

I have added a workaround that works for multiary operations with up to 6 operands; however, there might be a better (more general) solution?

simone-silvestri:
Anyway, if we agree with this workaround, I am eager to merge this PR and finally Oceananigans can be officially used with julia 1.12!

giordano:
ntuple + splatting are probably hard to reason about, although N is a static parameter 🤔

docs/make.jl
```diff
@@ -195,7 +195,7 @@ makedocs(; sitename = "Oceananigans.jl",
          plugins = [bib],
          warnonly = [:cross_references],
          draft = false,        # set to true to speed things up
-         doctest = true,       # set to false to speed things up
+         doctest = true,      # set to false to speed things up
```
simone-silvestri:
```suggestion
         doctest = true,       # set to false to speed things up
```

src/AbstractOperations/binary_operations.jl
```diff
@@ -53,7 +53,7 @@ choose_location(::Nothing, Lb::ConcreteLocationType, Lc) = Lb #
 
 # Apply the function if the inputs are scalars, otherwise broadcast it over the inputs
 # This can occur in the binary operator code if we index into with an array, e.g. array[1:10]
-@inline @propagate_inbounds apply_op(op, a, b) = op(a, b)
+@inline @propagate_inbounds apply_op(op, a::Number, b::Number) = op(a, b)
```
simone-silvestri:
```suggestion
@inline @propagate_inbounds apply_op(op, a, b) = op(a, b)
```

examples/internal_tide.jl
```diff
@@ -43,7 +43,7 @@ width = 20kilometers
 hill(x) = h₀ * exp(-x^2 / 2width^2)
 bottom(x) = - H + hill(x)
 
-grid = ImmersedBoundaryGrid(underlying_grid, PartialCellBottom(bottom))
+grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom))
```
simone-silvestri:
```suggestion
grid = ImmersedBoundaryGrid(underlying_grid, PartialCellBottom(bottom))
```

src/AbstractOperations/multiary_operations.jl
```diff
@@ -13,9 +13,44 @@ struct MultiaryOperation{LX, LY, LZ, N, O, A, IN, G, T} <: AbstractOperation{LX,
     end
 end
 
+# NOTE!!! The GPU compiler has trouble inferring this operation which might lead to dynamic evaluations.
 @inline Base.getindex(Π::MultiaryOperation{LX, LY, LZ, N}, i, j, k)  where {LX, LY, LZ, N} =
     Π.op(ntuple(γ -> Π.▶[γ](i, j, k, Π.grid, Π.args[γ]), Val(N))...)
 
+# Try to improve inferrability by hardcoding methods for small N
+@inline Base.getindex(Π::MultiaryOperation{LX, LY, LZ, 1}, i, j, k)  where {LX, LY, LZ} =
```
giordano, line 21: I don't think this is _bad_, ntuple is implemented with similar manual splitting. This might cause longer compilation latency, but so does Val.

simone-silvestri:
There seems to be one last issue connected to the docs deployment.

giordano:
```
 ! [remote rejected] HEAD -> gh-pages (cannot lock ref 'refs/heads/gh-pages': is at a00f0e6e5641011ad80648a72f2e2f21ed6ec749 but expected 534c22e3519adcadb1802bc44ce20e67bcac43c6)
```
Is that a race condition during a push?  Can you restart that job? (I can't myself).

Side note, [in `Breeze.jl` I split the docs job in two parts](https://github.com/NumericalEarth/Breeze.jl/pull/107), build and deployment, so that if the deploy fails for _reasons_ and you need to restart it you don't need to rebuild the docs all over again.

simone-silvestri:
Ok, let's merge away.

navidcy:
O.M.G.!
Amazing!

navidcy:
I'm changing my `.zshrc` so that `julia` is now `julia +1.12`!!

.buildkite/pipeline.yml
wsmoses: a follow up comment, but reactant should work on 1.12. It is not bound by the version limitations of enzyme

PR #4837, opened 2025-10-09
Materialize buoyancy gradients for use in parameterizations

simone-silvestri:
This PR introduces the precomputation of buoyancy gradients, which are required for most parameterization. 
By doing this, we avoid heavyl interpolation of buoyancy computations (for example in slope computations or in Richardson number computations) speeding up significantly the parameterizations.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -102,6 +102,8 @@ the features they describe! Also, if you have developed a new feature in Oceanan
 
 If you have work using Oceananigans that you would like to have listed here, please open a pull request to add it or let us know!
 
+1. Zhang, Y., Kang, W., and Marshall, J. (2025) [How does ice shell geometry shape ocean dynamics on icy moons?](https://doi.org/10.48550/arXiv.2510.25988), _arXiv preprint_, arXiv:2510.25988. DOI: [10.48550/arXiv.2510.25988](https://doi.org/10.48550/arXiv.2510.25988)
+
 1. Shikanian, A. and Parfenyev, V. (2025) [The effects of no-slip boundaries and external force torque on two-dimensional turbulence in a square domain](https://doi.org/10.48550/arXiv.2508.13590), _arXiv preprint_, arXiv:2508.13590. DOI: [10.48550/arXiv.2508.13590](https://doi.org/10.48550/arXiv.2508.13590)
 
 1. Souza, A. N., Silvestri, S., Deck, K., Bischoff, T., Ferrari, R., and Flierl, G. R. (2025) [Surface to seafloor: A generative AI framework for decoding the ocean interior state](https://doi.org/10.48550/arXiv.2504.15308), _arXiv preprint_, arXiv:2504.15308. DOI: [10.48550/arXiv.2503.12845](https://doi.org/10.48550/arXiv.2504.15308)


src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -161,7 +161,7 @@ function HydrostaticFreeSurfaceModel(; grid,
     advection = NamedTuple(name => adapt_advection_order(scheme, grid) for (name, scheme) in pairs(advection))
 
     validate_buoyancy(buoyancy, tracernames(tracers))
-    buoyancy = regularize_buoyancy(buoyancy)
+    buoyancy = regularize_buoyancy(buoyancy, grid)
```
```suggestion
    buoyancy = materialize_buoyancy(buoyancy, grid)
```

src/BuoyancyFormulations/buoyancy_force.jl
```diff
@@ -44,11 +46,31 @@ NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
 └── coriolis: Nothing
 ```
 """
-function BuoyancyForce(formulation; gravity_unit_vector=NegativeZDirection())
+function BuoyancyForce(grid; formulation, gravity_unit_vector=NegativeZDirection(), precompute_gradients=true)
```
glwagner:
```suggestion
function BuoyancyForce(grid; formulation, gravity_unit_vector=NegativeZDirection(), materialize_gradients=true)
```

I suggest this term because it is more literal. Whether or not these quantities are computed (and where and when) is a model choice. Here, all we know is that the gradients are materialized vs empty.

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
glwagner: This function will need the `closure` too eventually?

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
```diff
@@ -83,6 +84,9 @@ function compute_auxiliaries!(model::HydrostaticFreeSurfaceModel; w_parameters =
     P    = model.pressure.pHY′
     arch = architecture(grid)
 
+    # Maybe compute buoyancy gradients
+    compute_buoyancy_gradients!(buoyancy, grid, tracers; parameters = κ_parameters)
```
glwagner, line 88: why does it take "κ_parameters"? These are the parameters for the turbulence closure? It's not clear why we would use that.

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -171,7 +171,7 @@ function NonhydrostaticModel(; grid,
     all_auxiliary_fields = merge(auxiliary_fields, biogeochemical_auxiliary_fields(biogeochemistry))
     tracers, auxiliary_fields = validate_biogeochemistry(tracers, all_auxiliary_fields, biogeochemistry, grid, clock)
     validate_buoyancy(buoyancy, tracernames(tracers))
-    buoyancy = regularize_buoyancy(buoyancy)
+    buoyancy = regularize_buoyancy(buoyancy, grid)
```
glwagner:
```suggestion
    buoyancy = materialize_buoyancy(buoyancy, grid)
```

src/Models/HydrostaticFreeSurfaceModels/update_hydrostatic_free_surface_model_state.jl
simone-silvestri: right, I think we need some `interior_parameters` and `surface_parameters`.

simone-silvestri:
The difference between main and this PR is staggering for all the CATKE functions

## On main
CATKE was the most expensive kernel
image

## On this PR
Tendencies take back their role of most heavy computations in the model
image

simone-silvestri:
> Looks good except for general confusion about the parameters and a few name change suggestions

I will merge this PR and open a new PR immediately to fix the parameter naming (which I was already doing in #4811)

PR #4838, opened 2025-10-09
Some quality of life changes to distributed functions

simone-silvestri:
`barrier!` does not mutate the input so it should be `barrier`, and `global_barrier` should not accept a communicator as the `global_` qualifier implies that the barrier operates on the whole MPI processor grid.

### DIFF
--- a/.buildkite/pipeline-benchmarks.yml
+++ b/.buildkite/pipeline-benchmarks.yml
@@ -54,7 +54,7 @@ steps:
         export BENCHMARK_GROUP
         OUTPUT_PREFIX="\${BENCHMARK_GROUP}_output"
 
-        $NSYS profile --output=\${OUTPUT_PREFIX} --trace=cuda $TARTARUS_HOME/julia-$JULIA_VERSION/bin/julia --color=yes --project --check-bounds=no test/benchmark_tests.jl
+        $NSYS profile --output=\${OUTPUT_PREFIX} --force-overwrite true --trace=cuda $TARTARUS_HOME/julia-$JULIA_VERSION/bin/julia --color=yes --project --check-bounds=no test/benchmark_tests.jl
         $NSYS stats \${OUTPUT_PREFIX}.nsys-rep > \${OUTPUT_PREFIX}.txt
 
         # Remove generated output files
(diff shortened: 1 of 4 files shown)


simone-silvestri:
The error with reactant popped up again. However, this should not be connected to this PR

simone-silvestri:
I am not sure why this PR would result in the Reactant tests failing. In practice, all it is is a refactoring of method names.
@wsmoses @giordano would you have an idea on why this happens?

giordano:
As far as I understand the error is https://buildkite.com/clima/oceananigans/builds/26165#0199dce2-6b9d-4efa-91df-630b0c035615/17-457
```
[2025/10/13 09:29:41.127] ERROR Error during loading of extension OceananigansReactantExt of Oceananigans, use `Base.retry_load_extensions()` to retry. -@-> loading.jl:1371
```
which is then causing the test to error out because methods in the extension aren't being loaded.  Is the method you're renaming used in the Reactant extension?

simone-silvestri:
Yes, but it is only renamed. 
Might this be the problem of the extension not loading?
```
WARNING: Method definition barrier(Any) in module DistributedComputations at /var/lib/buildkite-agent/Oceananigans.jl-26165/src/DistributedComputations/distributed_macros.jl:12 overwritten at /var/lib/buildkite-agent/Oceananigans.jl-26165/src/DistributedComputations/partition_assemble.jl:12.
  | 2025-10-13 09:29:12 UTC | ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
  | 2025-10-13 09:29:20 UTC | WARNING: Method definition barrier(Any) in module DistributedComputations at /var/lib/buildkite-agent/Oceananigans.jl-26165/src/DistributedComputations/distributed_macros.jl:12 overwritten at /var/lib/buildkite-agent/Oceananigans.jl-26165/src/DistributedComputations/partition_assemble.jl:12.
```

giordano:
Possibly, fixing an error might be a good idea anyway.

simone-silvestri:
Thanks, that seems to have done the trick. So, in general, for the extension to be able to load, the main package needs to be able to precompile always.

giordano:
I didn't know that!

PR #4840, opened 2025-10-09
Fix `mt` problem with julia 1.12

### LINKED ISSUE #4839: Julia 1.12.0: ERROR: LoadError: FieldError: type Core.TypeName has no field `mt`
Hi,

Just a minor issue with Julia 1.12.0:
```julia
particles = LagrangianParticles(init_particles, restitution=1.0,
    dynamics=bubble_dynamics!,
    tracked_fields=(T=tracers.T, S=tracers.S,
        pHY′=hydrostatic_pressure_anomaly, pNHS=nonhydrostatic_pressure)
)

@show particles
```
@show particles gives
```
ERROR: LoadError: FieldError: type Core.TypeName has no field `mt`
```

Everything else works just like in Julia 1.11.7.
Best,
Alexander

simone-silvestri:
closes #4839

### DIFF
--- a/src/Utils/prettysummary.jl
+++ b/src/Utils/prettysummary.jl
@@ -5,7 +5,7 @@ prettysummary(x, args...) = summary(x)
 
 function prettysummary(f::Function, showmethods=true)
     ft = typeof(f)
-    if if VERSION >= v"1.12.0"
+    if VERSION >= v"1.12.0"
         name = ft.name.name
     else
         mt = ft.name.mt


src/Utils/prettysummary.jl
```diff
@@ -5,8 +5,12 @@ prettysummary(x, args...) = summary(x)
...
 function prettysummary(f::Function, showmethods=true)
     ft = typeof(f)
-    mt = ft.name.mt
-    name = mt.name
+    if VERSION >= v"1.12.0"
+        name = ft.name.name
+    else
+        mt = ft.name.mt
+        name = mt.name
+    end
```
giordano, line 13: Why not simply using
```suggestion
    name = Base.nameof(f)
```
everywhere?  At the moment in Julia v1.12 this is returning
```julia-repl
julia> f = sin
sin (generic function with 14 methods)

julia> if VERSION >= v"1.12.0"
           name = ft.name.name
       else
           mt = ft.name.mt
           name = mt.name
       end
Symbol("#sin")
```
The symbol has an extra leading `#`, which is screwing up doctests in #4836

simone-silvestri: sounds good, we can change it in the julia 1.12 #4836

PR #4842, opened 2025-10-10
Improve benchmark tests

simone-silvestri:
I am afraid that, given the very low values given to velocity, and an almost constant temperature and salinity, the benchmark tests do not really run in a fair way. I had already noticed a case where the performance depended on the initial conditions (i.e. if your velocities are zeros, the kernel automatically does not compute the reconstructions).

This PR changes the initial conditions of the benchmark tests to make sure they are fair, i.e. a stratified fluid in z with much higher velocity fluctuations that cannot die out in the first iteration

### DIFF
--- a/src/Models/VarianceDissipationComputations/VarianceDissipationComputations.jl
+++ b/src/Models/VarianceDissipationComputations/VarianceDissipationComputations.jl
@@ -35,13 +35,12 @@ using KernelAbstractions: @kernel, @index
 
 const RungeKuttaScheme = Union{RungeKutta3TimeStepper, SplitRungeKutta3TimeStepper}
 
-struct VarianceDissipation{P, K, A, D, S, G}
+struct VarianceDissipation{P, K, A, D, S}
     advective_production :: P
     diffusive_production :: K
     advective_fluxes :: A
     diffusive_fluxes :: D
     previous_state :: S
-    gradient_squared :: G
     tracer_name :: Symbol
 end
 
@@ -99,9 +98,7 @@ function VarianceDissipation(tracer_name, grid;
     advective_fluxes = (; Fⁿ, Fⁿ⁻¹)
     diffusive_fluxes = (; Vⁿ, Vⁿ⁻¹)
 
-    gradients = deepcopy(P)
-
-    return VarianceDissipation(P, K, advective_fluxes, diffusive_fluxes, previous_state, gradients, tracer_name)
+    return VarianceDissipation(P, K, advective_fluxes, diffusive_fluxes, previous_state, tracer_name)
 end
 
 function (ϵ::VarianceDissipation)(model)
--- a/src/Models/VarianceDissipationComputations/flatten_dissipation_fields.jl
+++ b/src/Models/VarianceDissipationComputations/flatten_dissipation_fields.jl
@@ -10,7 +10,7 @@ Flatten the dissipation fields of a `VarianceDissipation` object into a named tu
 function flatten_dissipation_fields(t::VarianceDissipation) 
     A = t.advective_production
     D = t.diffusive_production
-    G = t.gradient_squared
+    
     tracer_name = t.tracer_name
 
     dirs = (:x, :y, :z)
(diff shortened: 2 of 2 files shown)


simone-silvestri:
The benchmarks seem to be fair, as this PR has no effect on the performance results.
We probably can close this PR even if I would prefer merging this so we know we have a test case which is solid.
However, do you guys know what all these test failures that do not pertain to this PR are?
@navidcy @glwagner @siddharthabishnu

simone-silvestri:
Also, a random bathymetry with values going from -5000 to -1000, as it is right now, is not really representative of an ocean simulation where the bathymetry is somewhat smoother (the benchmarks here are hitting way more reduced advection than what would happen in a real simulation). Therefore, I think it is better to use a sloped bathymetry.

glwagner:
The benchmarks should measure time per time-step (not SYPD).

glwagner:
Where are the benchmarks uploaded? Can we put a link into the documentation or readme or somewhere?

simone-silvestri:
> The benchmarks should measure time per time-step (not SYPD).

They do better (for GPU only)! They measure the timing of each kernel independently. For the CPU tests, I have this open PR #4711, which probably still needs some work for the formatting of the output.

> are there benchmarks for NonhydrostaticModel?

No, but we can include them for sure. The tests are run on Tartarus, and a .txt file summarizing the performance of the tested simulation is uploaded as a buildkite artifact.

glwagner:
> and a .txt file summarizing the performance of the tested simulation is uploaded as a buildkite artifact.

How can one view the artifact? Is there documentation or instructions somewhere?

glwagner:
> They do better (for GPU only)! They measure the timing of each kernel independently.

Why does it matter what the bathymetry is then?

simone-silvestri:
>  Where are the benchmarks uploaded? Can we put a link into the documentation or readme or somewhere?

The tests are run on Tartarus, and a .txt file summarizing the performance of the tested simulation is uploaded as a buildkite artifact. I think they are not really user-friendly, more tests for developers to make sure we don't hit some performance regression:

This is an example of the output of the benchmark tests
[buildkiteartifacts.com_52b8ca89-71e4-4c6b-b9a2-94f4dcea2227_019661c8-be0b-455f-80e2-9a259b307d77_0199cdc9-54e6-4211-a2f6-4bb0c7921248_0199cdc9-5dc2-4d76-a0dd-849eb1184893_diff_immersed_output.txt_response-content-type=text%2Fplain&X-Amz-Algorithm=AWS4.pdf](https://github.com/user-attachments/files/22850510/buildkiteartifacts.com_52b8ca89-71e4-4c6b-b9a2-94f4dcea2227_019661c8-be0b-455f-80e2-9a259b307d77_0199cdc9-54e6-4211-a2f6-4bb0c7921248_0199cdc9-5dc2-4d76-a0dd-849eb1184893_diff_immersed_output.txt_response-content-type.text.2Fplain.X-Amz-Algorithm.AWS4.pdf)

The artifacts are uploaded on the github pipeline (for example) 
https://buildkite.com/clima/oceananigans-benchmarks-1/builds/3442#0199cdc9-5dc2-4d76-a0dd-849eb1184893
under the `artifacts` tab
image


The bathymetry has a role because of the different methods that are used near the immersed boundary which will modify the 
profile of the simulation.

glwagner:
> The bathymetry has a role because of the different methods that are used near the immersed boundary which will modify the
> profile of the simulation.

But if we are using non-short-circuiting logic, the same methods are called in every cell regardless of the immersed boundary.

simone-silvestri:
I am not sure that always happens on a GPU; there might be some optimizations that occur under the hood.

glwagner:
> I am not sure that always happens on a GPU; there might be some optimizations that occur under the hood.

I don't think it's possible because the bathymetry is determined by an array whose values can change / are only known at runtime (not compile time)

simone-silvestri:
Yeah, but at runtime, how CUDA organizes execution might depend on possible zeros. 
I noticed that if you multiply by zero a very complicated computation (like, for example, a WENO reconstruction), the reconstruction is avoided altogether.

simone-silvestri:
@navidcy do you know why the documenter key is not found here?
https://buildkite.com/clima/oceananigans/builds/26163#0199dc9b-eb29-42ab-8f54-4daaa060ddd4

giordano:
To make Greg happy, we could automatically upload the benchmark results somewhere so that we can produce charts like https://enzymead.github.io/Enzyme-JAX/benchmarks/ and https://enzymead.github.io/Reactant.jl/benchmarks/.  Those are using https://github.com/benchmark-action/github-action-benchmark, but we can explore alternative visualisations if you prefer, just to give an idea.  This could be another thing I can work on next month, if you're interested

simone-silvestri:
I would be very happy if you would like to take this on! Also, this could help with #4711. I can work on adding some relevant non-hydrostatic benchmarks in another PR (if I can find some time).

giordano:
I opened #4851 to track this task.

simone-silvestri:
In the meantime, I merge this PR that just changes the initial condition and physical setup slightly.

PR #4848, opened 2025-10-11
Support `ImmersedBoundaryGrid` reconstruction using NetCDF

tomchor:
Another necessary step towards supporting `FiedTimeSeries` construction from NetCDF output

### DIFF
--- a/ext/OceananigansNCDatasetsExt.jl
+++ b/ext/OceananigansNCDatasetsExt.jl
@@ -18,7 +18,8 @@ using Oceananigans.Grids: Center, Face, Flat, Periodic, Bounded,
                           topology, halo_size, xspacings, yspacings, zspacings, λspacings, φspacings,
                           parent_index_range, nodes, ξnodes, ηnodes, rnodes, validate_index, peripheral_node,
                           constructor_arguments, architecture
-using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBottom, GFBIBG, GridFittedBoundary, PartialCellBottom, PCBIBG
+using Oceananigans.ImmersedBoundaries: ImmersedBoundaryGrid, GridFittedBottom, GFBIBG, GridFittedBoundary, PartialCellBottom, PCBIBG,
+                                       CenterImmersedCondition, InterfaceImmersedCondition
 using Oceananigans.Models: ShallowWaterModel, LagrangianParticles
 using Oceananigans.Utils: TimeInterval, IterationInterval, WallTimeInterval, materialize_schedule,
                           versioninfo_with_gpu, oceananigans_versioninfo, prettykeys
@@ -743,12 +744,13 @@ convert_for_netcdf(x::Bool) = string(x)
 convert_for_netcdf(x::NTuple{N, Number}) where N = collect(x)
 convert_for_netcdf(x) = string(x)
 convert_for_netcdf(::GPU) = "GPU()"
+convert_for_netcdf(::CenterImmersedCondition) = "CenterImmersedCondition()"
+convert_for_netcdf(::InterfaceImmersedCondition) = "InterfaceImmersedCondition()"
 
 materialize_from_netcdf(dict::AbstractDict) = OrderedDict(Symbol(key) => materialize_from_netcdf(value) for (key, value) in dict)
 materialize_from_netcdf(x::Number) = x
 materialize_from_netcdf(x::Array) = Tuple(x)
 materialize_from_netcdf(x::String) = @eval $(Meta.parse(x))
-
 
 function netcdf_grid_constructor_info(grid)
     underlying_grid_args, underlying_grid_kwargs = constructor_arguments(grid)
@@ -773,16 +775,18 @@ function netcdf_grid_constructor_info(grid::ImmersedBoundaryGrid)
 end
 
 function write_immersed_boundary_data!(ds, grid::ImmersedBoundaryGrid, immersed_grid_args)
-    ibg_group = defGroup(ds, "immersed_grid_reconstruction_args")
-
+    group_name = "immersed_grid_reconstruction_args"
     if (grid.immersed_boundary isa GridFittedBottom) || (grid.immersed_boundary isa PartialCellBottom)
         bottom_height = pop!(immersed_grid_args, :bottom_height)
-        defVar(ibg_group, "bottom_height", bottom_height; attrib=convert_for_netcdf(immersed_grid_args))
+        ibg_group = defGroup(ds, group_name; attrib=convert_for_netcdf(immersed_grid_args))
+        defVar(ibg_group, "bottom_height", bottom_height)
 
     elseif grid.immersed_boundary isa GridFittedBoundary
         mask = pop!(immersed_grid_args, :mask)
-        defVar(ibg_group, "mask", mask; attrib=convert_for_netcdf(immersed_grid_args))
+        ibg_group = defGroup(ds, group_name; attrib=convert_for_netcdf(immersed_grid_args))
+        defVar(ibg_group, "mask", mask)
     end
+
     return ds
 end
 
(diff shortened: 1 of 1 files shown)


tomchor:
One thing that's making this harder is the fact that we started flattening singleton dimensions in NetCDF output. For example, while a 2D `Field` in Oceananigans may have `size(field) == (Nx, Ny, 1)`, when it is written to NetCDF is has the size `(Nx, Ny)`. In the past this wasn't a problem (and was even desirable to some extent), now when getting information back from NetCDF to Oceananigans, this requires some extra work.

I wonder if it's worth going back on that decision and keeping singleton dimensions. @ali-ramadhan @glwagner thoughts?

simone-silvestri:
If it makes the source code simpler, I vote for keeping the third dimension. In the end, Oceananigans' fields are always three-dimensional.

glwagner:
I think it makes sense to keep singleton dimensions in NetCDF output if it simplifies the interface between NetCDF and Oceananigans. I think the disadvantages are minor. If we can identify significant disadvantages in the future, then we can motivate investing energy into supporting the removal of singleton dimensions later.

tomchor:
Btw, turns out `test_netcdf_writer.jl` was being called twice. Once in `runtest.jl`, and again inside `test_output_writer.jl`. I fixed that, so `simultion-tests` should run faster now.

If tests pass this PR is ready to merge/review.

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -797,27 +749,54 @@ materialize_from_netcdf(x::Number) = x
...
+    grid_metadata = Dict(:immersed_boundary_type => nothing,
+                         :underlying_grid_type => underlying_grid_type)
+    return underlying_grid_args, underlying_grid_kwargs, immersed_grid_args, grid_metadata
+end
+
+function netcdf_grid_constructor_info(grid::ImmersedBoundaryGrid)
+    underlying_grid_args, underlying_grid_kwargs, immersed_grid_args = constructor_arguments(grid)
+
+    immersed_boundary_type = typeof(grid.immersed_boundary).name.wrapper |> string # Save type of immersed boundary for reconstruction
+    underlying_grid_type   = typeof(grid.underlying_grid).name.wrapper |> string # Save type of underlyinh grid for reconstruction
```
ali-ramadhan:
```suggestion
    underlying_grid_type   = typeof(grid.underlying_grid).name.wrapper |> string # Save type of underlying grid for reconstruction
```

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -829,23 +808,49 @@ function reconstruct_grid(filename::String)
...
+    grid_reconstruction_metadata          = ds.group["grid_reconstruction_metadata"].attrib |> materialize_from_netcdf
+
+    # Pop out infomration about the underlying grid
+    underlying_grid_type = grid_reconstruction_metadata[:underlying_grid_type]
+    underlying_grid = underlying_grid_type(values(underlying_grid_reconstruction_args)...; underlying_grid_reconstruction_kwargs...)
 
-    # Reconstruct the grid which may or may not be an underlying grid to an ImmersedBoundaryGrid
-    maybe_underlying_grid = grid_type(values(grid_reconstruction_args)...; grid_reconstruction_kwargs...)
+    # If this is an ImmersedBoundaryGrid, reconstruct the immersed boundary, otherwise underlying grid is the final grid
+    if grid_reconstruction_metadata[:immersed_boundary_type] isa Nothing
```
ali-ramadhan:
```suggestion
    if isnothing(grid_reconstruction_metadata[:immersed_boundary_type])
```

test/test_grid_reconstruction.jl
```diff
@@ -355,14 +361,14 @@ N = 6
             test_netcdf_grid_reconstruction(stretched_latlon_grid)
 
             # TODO: Make the functionality below work
```
ali-ramadhan: Can probably remove this TODO now?

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -829,23 +808,49 @@ function reconstruct_grid(filename::String)
...
 end
 
+function reconstruct_immersed_boundary(ds)
+    ibg_group = ds.group["immersed_grid_reconstruction_args"]
+
+    grid_reconstruction_metadata = ds.group["grid_reconstruction_metadata"].attrib |> materialize_from_netcdf
+    immersed_boundary_type = grid_reconstruction_metadata[:immersed_boundary_type]
+    if immersed_boundary_type == GridFittedBottom
+        bottom_height = Array(ibg_group["bottom_height"])
+        immersed_boundary = immersed_boundary_type(bottom_height)
```
ali-ramadhan: Should we also be passing in `ibg_group.attrib["immersed_condition"]`?

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -829,23 +808,49 @@ function reconstruct_grid(filename::String)
...
+
+    elseif immersed_boundary_type == PartialCellBottom
+        bottom_height = Array(ibg_group["bottom_height"])
+        immersed_boundary = immersed_boundary_type(bottom_height)
+
+    elseif immersed_boundary_type == GridFittedBoundary
+        mask = Array(ibg_group["mask"])
+        immersed_boundary = immersed_boundary_type(mask)
```
ali-ramadhan, line 832: Should we also be passing in `ibg_group.attrib["minimum_fractional_cell_height"]`?

test/test_grid_reconstruction.jl
tomchor: yes! Thanks

ext/OceananigansNCDatasetsExt.jl
tomchor: `GridFittedBoundary` only takes `mask` as a argument! `minimum_fractional_cell_height` is an argument only for `PartialCellBottom`.

ext/OceananigansNCDatasetsExt.jl
tomchor: True!

ext/OceananigansNCDatasetsExt.jl
tomchor: I should change that for `PartialCellBottom` though! (Which maybe is where you meant to comment this?)

tomchor:
@glwagner I have an approval and the PR is ready, yet I can't merge. Are there new rules in place to merge PRs?

image

ali-ramadhan:
Not sure when this happened but it sounds like we need to be part of https://github.com/orgs/CliMA/teams/clima-code-reviewers for our reviews to count now. How do we become a CliMA code reviewer?

simone-silvestri (approved):
It's an issue that popped up recently because, apparently, other clima repos had some bot spam, so, for security reasons, the clima organization decided to make this change.

simone-silvestri:
@cmbengue and @rafferrari, we would need to add more people, all of whom are major contributors to oceananigans over the years, to the climate code reviewers list. This is required to accelerate development toward pre-clima-code-reviewer restrictions (see comments above).
@ali-ramadhan, @tomchor, @francispoulin @giordano @jagoosw.
Did I forget someone?

Or more generally, every co-author of this preprint: 
```
@article{Oceananigans-overview-paper-2025,
  title = {{High-level, high-resolution ocean modeling at all scales with Oceananigans}},
  author = {G. L. Wagner and S. Silvestri and N. C. Constantinou and A. Ramadhan and J.-M. Campin and C. Hill and T. Chor and J. Strong-Wright and X. K. Lee and F. Poulin and A. Souza and K. J. Burns and S. Bishnu and J. Marshall and R. Ferrari},
  journal = {arXiv preprint},
  year = {2025},
  archivePrefix = {arXiv},
  eprint = {2502.14148},
  doi = {10.48550/arXiv.2502.14148},
  notes = {submitted to the Journal of Advances in Modeling Earth Systems},
}
```

Another option would be to lift this requirement for Oceananigans, since we already have checks in place: only the restricted pool of people that have write access can accept and merge PRs

PR #4856, opened 2025-10-14
Accomodate `CUDA@5.8.5` in Project.toml temporarily

xkykai:
See this comment on this PR and its related issue:
https://github.com/JuliaGPU/CUDA.jl/pull/2907#issuecomment-3361599935

In summary, there are bugs in `sum(...; dims=1)` in `CUDA@5.9.0`, which will be fixed when a new tagged version of `CUDA@5.9.1` is released. That is not done yet. In the meantime, there is a backport `CUDA@5.8.5` which fixes the bug.

We should open up CUDA compat for `5.8.5` as well right now since I wouldn't be comfortable using `CUDA@5.9.0` given this known issue.

Once `CUDA@5.9.1` is released I will bump compat up.

@navidcy @simone-silvestri @ali-ramadhan @tomchor @glwagner

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -15,7 +15,6 @@ StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 TimesDates = "bdfc003b-8df8-5c39-adcd-3a9087f5df4a"
 
 [compat]
-CUDA = "5.9"
 CairoMakie = "0.11, 0.12, 0.13, 0.14"
 Documenter = "1"
 DocumenterCitations = "1"


navidcy:
I tweaked the compat a bit because the way it was before Oceananigans would use CUDA v5.9.0; [this way](https://github.com/CliMA/Oceananigans.jl/pull/4856/commits/cc22ae86074d253d3f95fd3638cbe6a72023d0be) it won't!

navidcy:
We [don't need compat entries in docs/Project](https://github.com/CliMA/Oceananigans.jl/pull/4856/commits/e0df714c77ab7dc8301657f9ff114ba221ecf387) for Oceananigans dependencies since the Oceananigans compat will be enforced.

cc @glwagner, @simone-silvestri, @tomchor

navidcy:
We've hit the bug in GeophysicalFlows as well. (see https://github.com/FourierFlows/GeophysicalFlows.jl/pull/397#issue-3489514669).

@xkykai feel free to merge

PR #4859, opened 2025-10-15
Add fields to `implicit_step!` in `tke_timestep!`

simone-silvestri:
This part of the code was not updated to work with #4383.
This PR fixes it.


simone-silvestri:
Maybe we can merge this before #4567 so we tag the fix in the new version?

glwagner (approved):

PR #4860, opened 2025-10-15
Add missing import when extending `FieldBoundaryConditions` in `MultiRegion`

navidcy:
Perhaps this is related to those weird MultiRegion-related methods appearing with Distributed? See https://github.com/CliMA/ClimaOcean.jl/issues/647#issue-3485830061

Since the method wasn't imported, the attempt below to extent it might resulted in overwriting it?

What was missing was:
```julia
import Oceananigans.BoundaryConditions: FieldBoundaryConditions
```

Julia v1.12 warned me for that; that's how I found it.

```Julia
  1 dependency had output during precompilation:
┌ Oceananigans
│  WARNING: Constructor for type "FieldBoundaryConditions" was extended in `MultiRegion` without explicit qualification or import.
│    NOTE: Assumed "FieldBoundaryConditions" refers to `BoundaryConditions.FieldBoundaryConditions`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function FieldBoundaryConditions end`.
│    Hint: To silence the warning, qualify `FieldBoundaryConditions` as `BoundaryConditions.FieldBoundaryConditions` in the method signature or explicitly `import BoundaryConditions: FieldBoundaryConditions`.
└
```

cc @taimoorsohail

### DIFF
--- a/src/MultiRegion/multi_region_field.jl
+++ b/src/MultiRegion/multi_region_field.jl
@@ -6,7 +6,7 @@ using Oceananigans.OutputWriters: output_indices
 
 using Base: @propagate_inbounds
 
-import Oceananigans.BoundaryConditions: regularize_field_boundary_conditions, FieldBoundaryConditions
+import Oceananigans.BoundaryConditions: regularize_field_boundary_conditions
 import Oceananigans.Diagnostics: hasnan
 import Oceananigans.DistributedComputations: reconstruct_global_field, CommunicationBuffers
 import Oceananigans.Fields: set!, compute!, compute_at!, interior, communication_buffers,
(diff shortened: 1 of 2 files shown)


taimoorsohail:
The error persists with your first tranche of changes @navidcy

navidcy:
hm... I still think this PR resolves a mistake.
I'll wait for @simone-silvestri or @siddharthabishnu to confirm

simone-silvestri:
indeed, it does solve an issue (we had that change in #4760)

simone-silvestri (approved):

PR #4865, opened 2025-10-16
Pass correct `kwargs...` in distributed tripolar `fill_halo_event!`

simone-silvestri:
Keyword arguments were thrown away in the distributed zipper, causing it to fill distributed halos also when `only_local_halo = false`. As a consequence, MPI transfers were also synchronized in "local" fill halos, destroying the computation - communication overlap.
This PR fixes this behavior. Correctness does not change, only performance.

cc @xkykai @taimoorsohail

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.100.5"
+version = "0.100.6"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"


simone-silvestri:
The difference between the main branch and this PR is shown in the following snapshots for the same exact case (a global third of a degree simulation on 2 tartarus GPUs).
The blue boxes are GPU executions and the red boxes are MPI transfers, in the first case the transfers are not overlapped to compute, in this PR they are.
### main
image

### This PR
image

simone-silvestri:
If anyone wants to dive into the reports, they are downloadable from this [link](https://www.dropbox.com/scl/fo/2gw0baboumzfshl0fv1ki/AEYOlSWOT4-PYZcGYXB9w_Q?rlkey=1jsjssy5q72f083rmo40q00ia&st=tjqrkztw&dl=0)

The reports can be inspected using [nsight system](https://developer.nvidia.com/nsight-systems)

navidcy (approved):

Issue #4733, opened 2025-08-28
Post-processing and saving Fields in a manner compatible with FieldTimeSeries


erin-atkinson:
Excuse me if I'm ignorant of the correct way to do this. Right now my current workflow for post-processing is to read a `FieldTimeSeries` into a dependency field which is then repeatedly used to compute a result, which is saved (perhaps with the in-built jld2output! for cleanliness)

```julia
# Post-processing
using Oceananigans
using Oceananigans.OutputWriters: jld2output!
using Oceananigans.Fields: compute_at!
using JLD2

# Read timeseries
us = FieldTimeSeries("raw.jld2", "u"; backend=OnDisk())
vs = FieldTimeSeries("raw.jld2", "v"; backend=OnDisk())

# Get grid
grid = us.grid

# Create fields
u = us[1]
v = vs[1]

# Computation
s = Field(KernelFunctionOperation{Center, Center, Center}(mykernelfunction, grid, u, v))

outputs = (; s)
println()
# Repeat
for (n, t) in zip(axes(us, 4), us.times)
    set!(u, us[n])
    set!(v, vs[n])

    compute_at!(outputs, n)
    jld2output!("post-processed.jld2", n, t, outputs, ())
    print("\r$n")
end
```

In my ideal world, I would be able to treat `post-processed.jld2` as if it were simulation output, reading from it with `FieldTimeSeries` etc. As far as I understand it, this isn't possible without a lot of prep, as all output writers require initialisation with a model. Would it make sense to introduce functionality to save fields in this way?

glwagner:
If I recall correctly, @simone-silvestri worked on capabilities for saving down data series in a `FieldTimeSeries`-compatible format. I think for this you need to create a time series with `OnDisk` backend:

```julia
path = "post_processed.jld2"
name = "s"
st = FieldTimeSeries{Center, Center, Center}(grid, times=ut.times; path, name, backend=OnDisk())
```

you may then be able to use `set!`:

```julia
for n = 1:length(ut.times)
    set!(u, us[n])
    set!(v, vs[n])
    compute!(s)
    set!(st, s, n)
end
```

might be nice to brainstorm ways to streamline this!

the method is here: 

https://github.com/CliMA/Oceananigans.jl/blob/52847291258d1abd881f0374457a7a12ec027a2d/src/OutputReaders/set_field_time_series.jl#L100

erin-atkinson:
[...]

PR #4866, opened 2025-10-16
Allow initialization of empty `FieldDataset`s similar to methods for `FieldTimeSeries` and define `set!` behaviour for `FieldDataset`

erin-atkinson:
This adds behaviour for making an on-disk (and in-memory) field dataset similar to how a field timeseries can be constructed. `set!` passes through to `set!` for each timeseries corresponding to each keyword argument entered. There is also support for initializing a `FieldDataset` with entries similar a given named tuple of fields, intended for easily outputting post-processed data, for instance. 

The test is just a shameless copy of the test for `FieldTimeSeries{OnDisk}`.

In my opinion, `FieldDataset` deserves a rework. I envision it implementing `AbstractDict{String, FieldTimeSeries}`, for some set of `FieldTimeSeries` that share the same `grid`, `backend`, `times`, `path`, `reader_kw` (and possibly `time_indexing`?, though maybe there is a use case where a user wants this to differ between fields in the same file). I think this could be done without breaking changes given its only current use is reading in all timeseries in the same file, though I didn't commit to it here. It feels clunky currently. The type definition allows it to contain anything, which is why, for instance, support for writing metadata is avoided here. I'm open to working on a rework in this PR.


erin-atkinson:
This was discussed briefly in #4733

glwagner:
One way to motivate design is to come up with a few canonical examples / use cases that we want to support. Then we can work backwards from there to implement a design that is clean, intuitive, efficient and achieves the use case / user interface that was laid out.

erin-atkinson:
Like I mentioned in the issue, I'd like to be able to do `set!(fds, i, t; a, b, c)` to write `a`, `b`, `c` to a file, similar to `FieldTimeSeries` which this minimally implements, along with new constructors for creating `FieldDataset` not from a file. When manipulating simulation output, I find myself wanting to output some set of post-processed fields, for instance, and explicitly creating a `FieldTimeSeries` for each seems unnecessary when I could just make a single `FieldDataset` and then do something like

```julia
for i in r
    # (set arguments for operations in output_fields)
    compute!(output_fields)
    set!(fds, i; output_fields...)
end
```

The problem is because `FieldDataset` doesn't contain information related to the reading and writing (the properties that I mention), `set!` may have quite different behaviour on each contained `FieldTimeSeries`, depending on what they are.

glwagner:
> Like I mentioned in the issue, I'd like to be able to do `set!(fds, i, t; a, b, c)` to write `a`, `b`, `c` to a file, similar to `FieldTimeSeries` which this minimally implements, along with new constructors for creating `FieldDataset` not from a file. When manipulating simulation output, I find myself wanting to output some set of post-processed fields, for instance, and explicitly creating a `FieldTimeSeries` for each seems unnecessary when I could just make a single `FieldDataset` and then do something like
> 
> ```julia
> for i in r
>     # (set arguments for operations in output_fields)
>     compute!(output_fields)
>     set!(fds, i; output_fields...)
> end
> ```
> 
> The problem is because `FieldDataset` doesn't contain information related to the reading and writing (the properties that I mention), `set!` may have quite different behaviour on each contained `FieldTimeSeries`, depending on what they are.

Thank you for the example! If I understand correctly, your changes in this PR support the example you've given. So what is the motivation for introducing the restriction that all `FieldTimeSeries` in a dataset have to be on the same grid, or share the same times? I do see that if that is the case, then more operations between fields might be supported. But I am not sure what _requires_ it / what problem is solved by enforcing grid/times sameness.

erin-atkinson:
`backend` should be shared between all entries:

- The `metadata` variable is associated with a dataset, but it's not until `set!` is called on a timeseries that behaviour specific to the backend is available. For a dataset that consists of timeseries with varying `backend`, to save `metadata` we'd need to either check if an `OnDiskFTS` exists in the dictionary, or pass metadata to a new method for `set!(::FieldTimeSeries, )` that saves it, which seems strange. 
- `set!` has a different signature for `OnDisk` and `InMemory` timeseries, with the additional `time` positional argument for the former. 

The others, it wouldn't be required to restrict them right now. The only constructor for `FieldDataset` currently loads all fields in the same file, which must share `grid` and `path`. I suppose `times` may be different if the simulation has multiple output writers that save to the same file on different schedules.

glwagner:
> `backend` should be shared between all entries:
> 
> * The `metadata` variable is associated with a dataset, but it's not until `set!` is called on a timeseries that behaviour specific to the backend is available. For a dataset that consists of timeseries with varying `backend`, to save `metadata` we'd need to either check if an `OnDiskFTS` exists in the dictionary, or pass metadata to a new method for `set!(::FieldTimeSeries, )` that saves it, which seems strange.
> * `set!` has a different signature for `OnDisk` and `InMemory` timeseries, with the additional `time` positional argument for the former.

This makes sense!

> The others, it wouldn't be required to restrict them right now. The only constructor for `FieldDataset` currently loads all fields in the same file, which must share `grid` and `path`. I suppose `times` may be different if the simulation has multiple output writers that save to the same file on different schedules.

Got it. I think it's reasonable to have a convenience constructor that takes in just one `grid` and `times` even if `FieldDataset` does not require it. Probably there isn't much of a use case for heterogeneous grid/times. But we might save some effort if we do not actively enforce / encode a common grid/times assumption.

src/OutputReaders/field_dataset.jl
```diff
@@ -80,3 +83,114 @@ function Base.show(io::IO, fds::FieldDataset)
 
     return print(io, s)
 end
+
+"""
+    FieldDataset(grid, field_names, times;
```
glwagner:
```suggestion
    FieldDataset(grid, times, field_names;
```

this feels more natural to me, what do you think?

src/OutputReaders/field_dataset.jl
```diff
@@ -80,3 +83,114 @@ function Base.show(io::IO, fds::FieldDataset)
...
+    )
+
+    grid = fields[1].grid
+    any([field.grid != grid for field in fields]) && throw(ArgumentError("All fields must be defined on the same grid"))
+
+    loc = map(instantiated_location, fields)
+    inds = map(Fields.indices, fields)
+    bcs = map(Fields.boundary_conditions, fields)
+
+    FieldDataset(grid, keys(fields), times;
```
glwagner:
```suggestion
    return FieldDataset(grid, keys(fields), times; 
```

We try to enforce a style that always uses `return` for multiline functions

src/OutputReaders/field_dataset.jl
```diff
@@ -80,3 +83,114 @@ function Base.show(io::IO, fds::FieldDataset)
...
+- `metadata`: `Dict` containing metadata entries
+"""
+function FieldDataset(grid, fields::NTuple{N, Symbol}, times;
+    backend=OnDisk(),
+    path=nothing,
+    location=NamedTuple(),
+    indices=NamedTuple(),
+    boundary_conditions=NamedTuple(),
+    metadata=Dict(),
+    reader_kw=NamedTuple()) where {N}
```
glwagner: possibly, can use `fts_kw...` as a placeholder for all the kwargs that get passed without modification/inspection into `FieldTimeSeries`.

glwagner:
Maybe I should know the answer to this (and tangential to this PR) but what is the purpose of the `metadata` property? It's a concern because there is also a `Metadata` object that ClimaOcean defines.

erin-atkinson:
`metadata` contains a dictionary of key-value pairs which are extracted from an array of group names in the simulation output file (`metadata_paths=["metadata"]` by default). I typically write any information (author, text description, date, numerical parameters) to the output file using the `init` kwarg in `JLD2Writer` when creating a simulation, then these can be read using `FieldDataset`.

Writing `metadata` using an on-disk `FieldDataset` is useful for descriptions of post-processed fields and the aforementioned info.

src/OutputReaders/field_dataset.jl
erin-atkinson: I don't think there are any, should I add for future proofing?

erin-atkinson:
I think everything is cleaned up.

src/OutputReaders/field_dataset.jl
```diff
@@ -80,3 +84,141 @@ function Base.show(io::IO, fds::FieldDataset)
...
+"""
+    FieldDataset(grid, times, fields;
+                 backend=OnDisk(),
+                 path=nothing,
+                 location=NamedTuple(),
+                 indices=NamedTuple(),
+                 boundary_conditions=NamedTuple(),
+                 metadata=Dict(),
+                 reader_kw=NamedTuple(),
+                 )
```
glwagner:
```suggestion
                 backend = OnDisk(),
                 path = nothing,
                 location = NamedTuple(),
                 indices = NamedTuple(),
                 boundary_conditions = NamedTuple(),
                 metadata = Dict(),
                 reader_kw = NamedTuple())
```

src/OutputReaders/field_dataset.jl
```diff
@@ -80,3 +84,141 @@ function Base.show(io::IO, fds::FieldDataset)
...
+end
+
+
+"""
+    FieldDataset(times, fields;
+                 backend=OnDisk(),
+                 path=nothing,
+                 metadata=Dict(),
+                 reader_kw=NamedTuple()
+                 )
```
glwagner:
```suggestion
                 backend = OnDisk(),
                 path = nothing,
                 metadata = Dict(),
                 reader_kw = NamedTuple())
```

glwagner:
I'll take a look at the tests once the next batch runs

erin-atkinson:
Inconsistent style is likely a product of my own inexperience... If you are happy this can be merged.

glwagner:
> Inconsistent style is likely a product of my own inexperience... If you are happy this can be merged.

oh well, no worries then! I will merge.

PR #4870, opened 2025-10-17
Fix type inference failure in `AbstractOperations`

### LINKED ISSUE #4869: Type instability in reduction operators
It seems that reduction operators like `Average` and `Integral` are not type stable. This is a problem both for autodifferentiability (Enzyme chokes when using `Average` to compute the mean of a field, for example) and for processes that rely on integrated quantities; e.g. some land surface processes rely on balance equations with terms that involve integrating over the ground domain.

I haven't looked very deeply into the code yet, so I can't speculate as to the cause.

MWE:
```julia
using Oceananigans
using Statistics

import Random

const prng = Random.MersenneTwister(1234)

grid = RectilinearGrid(CPU(), size=(1,10), x=(0,1), z=(0,1), topology=(Periodic, Flat, Bounded))
X = CenterField(grid)
X .= randn(prng, size(X))

function mean_square_field(X::Field)
    Y = X^2
    return Field(Average(Y, dims=(1,2,3)))[1,1,1]
end

function mean_square_field!(Ȳ::Field, X::Field)
    Y = X^2
    set!(Ȳ, Field(Average(Y, dims=(1,2,3))))
    return nothing
end

@assert mean_square_field(X) ≈ mean(X^2)

@code_warntype mean_square_field(X)

Ȳ = CenterField(grid)

@code_warntype mean_square_field!(Ȳ, X)

function integrate_field(Ȳ::Field, X::Field)
    set!(Ȳ, Field(Integral(X, dims=(1,2,3))))
    return nothing
end

@code_warntype integrate_field(Ȳ, X)
```

Verified on Oceananigans v0.100.6

```
julia> versioninfo()
Julia Version 1.10.10
Commit 95f30e51f41 (2025-06-27 09:51 UTC)
Build Info:
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 128 × AMD EPYC 9354 32-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, znver3)
Threads: 1 default, 0 interactive, 1 GC (on 128 virtual cores)
Environment:
  JULIA_NUM_THREADS = 
  JULIA_VSCODE_REPL = 1
  JULIA_EDITOR = code
```

Possibly related to #4633

simone-silvestri:
closes #4869


src/AbstractOperations/conditional_operations.jl
```diff
@@ -91,12 +91,11 @@ julia> d[2, 1, 1]
...
 ```
 """
-function ConditionalOperation(operand::AbstractField;
+function ConditionalOperation(operand::AbstractField{LX, LY, LZ};
                               func = nothing,
                               condition = nothing,
-                              mask = zero(eltype(operand)))
+                              mask = zero(eltype(operand))) where {LX, LY, LZ}
     condition = validate_condition(condition, operand)
-    LX, LY, LZ = location(operand)
```
glwagner, line 99: I don't think this has an effect (but tell me if its not true). I think this compiles identically. We use the function since its the recommendation of YASGuide (eg type parameters should be used for dispatch, not to access type info). It's not that important, but wanted to mention.

simone-silvestri: It is a matter of passing arguments directly as parameteric types, which is type unstable, rather than inferring a type. The constructor cannot infer the values of an argument, but can infer the type of the arguments that are passed. Check the difference between these two constructors (which basically build the same object with just a different name)
```julia
struct InstableType{T1, T2, T3} end
struct StableType{T1, T2, T3} end

StableType(t::Tuple{T1, T2, T3}) where {T1, T2, T3} = StableType{T1, T2, T3}()
InstableType(T::Tuple) = InstableType{T[1], T[2], T[3]}()

@code_warntype StableType((1.0, 1.f0, Float16(1.0))) # This is type stable
@code_warntype InstableType((Float64, Float32, Float16)) # This is type - unstable
```

simone-silvestri: Actually, you are right. In the above case, there is no issue because `LX, LY, LZ` should be inferred in the function call. I will revert it. The problem arises only for constructors like this
https://github.com/CliMA/Oceananigans.jl/blob/f07428b824f1b2365beada01eeb7b5435a4afd0e/src/AbstractOperations/binary_operations.jl#L38-L43

which should actually be
```julia
 function _binary_operation(Lc::Tuple{LX, LY, LZ}, op, a, b, La, Lb, grid) where {LX, LY, LZ}
      ▶a = interpolation_operator(La, Lc) 
      ▶b = interpolation_operator(Lb, Lc) 
  
     return BinaryOperation{LX, LY, LZ}(op, a, b, ▶a, ▶b, grid) 
 end 
```

simone-silvestri:
After fixing this type - instability, I have noticed I introduced another type - instability in the Field constructor in #4706, which basically just moved the already existing type instability out of the fill halo functions and inside the constructor. 
This might be a bit more difficult ot solve

src/AbstractOperations/conditional_operations.jl
glwagner: got it!

src/AbstractOperations/at.jl
```diff
@@ -43,11 +57,12 @@ Modify the `abstract_operation` so that it returns values at
 `location`, where `location` is a 3-tuple of `Face`s and `Center`s.
 """
 macro at(location, abstract_operation)
+    location = instantiate_location_expression(location)
     insert_location!(abstract_operation, location)
 
     # We wrap it all in an interpolator to help "stubborn" binary operations
     # arrive in the right place.
-    wrapped_operation = quote
+    wrapped_operation = quote
```
giordano:
```suggestion
    wrapped_operation = quote
```

glwagner:
> After fixing this type - instability, I have noticed I introduced another type - instability in the Field constructor in #4706, which basically just moved the already existing type instability out of the fill halo functions and inside the constructor. This might be a bit more difficult ot solve

I missed this comment. Generally though a type inference issue in a constructor is not as important as something that gets hit every time step. If you're more specific about the issue maybe we can help

simone-silvestri:
The type instability is in https://github.com/CliMA/Oceananigans.jl/blob/bc4299048e26143971808b708feb737f0ca7ad8f/src/BoundaryConditions/fill_halo_kernels.jl#L42, which previously was in the `fill_halo_regions!` function. 
However, I'll work on merging this PR and then we can open a new one dealing with this issue

glwagner:
> The type instability is in
> 
> https://github.com/CliMA/Oceananigans.jl/blob/bc4299048e26143971808b708feb737f0ca7ad8f/src/BoundaryConditions/fill_halo_kernels.jl#L42
> 
> , which previously was in the `fill_halo_regions!` function.
> However, I'll work on merging this PR and then we can open a new one dealing with this issue

it may not be too hard:

https://github.com/CliMA/Oceananigans.jl/blob/bc4299048e26143971808b708feb737f0ca7ad8f/src/BoundaryConditions/boundary_condition_ordering.jl#L17-L46

its been written to minimize memory allocation maybe, but at the cost of infer ability, now we can reverse this

PR #4871, opened 2025-10-17
Add on_architecture method for XESMF.Regridder

xkykai:
Adding a simple convenience function to move regridder between architectures.

### DIFF
--- a/test/test_xesmf.jl
+++ b/test/test_xesmf.jl
@@ -48,11 +48,10 @@ for arch in archs
                 @test regridder.weights isa CUDA.CUSPARSE.CuSparseMatrixCSC
             end
 
-            cpu_regridder = on_architecture(CPU(), regridder)
-            gpu_regridder = on_architecture(GPU(), regridder)
-
-            @test cpu_regridder.weights isa SparseMatrixCSC
-            if arch isa GPU{CUDABackend}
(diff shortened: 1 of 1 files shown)


ext/OceananigansXESMFExt.jl
```diff
@@ -145,6 +146,11 @@ function Regridder(dst_field::AbstractField, src_field::AbstractField; method="c
     return XESMF.Regridder(method, weights, temp_src, temp_dst)
 end
 
+on_architecture(on, r::XESMF.Regridder) = XESMF.Regridder(on_architecture(on, r.method),
```
navidcy, line 149: `r.method` is a string; do we need that? does it even work?

xkykai: yes it does work, it does nothing to the string. Keeping it or removing it is fine either way, in the future if `method` extends to using architecture dependent objects this can still take care of it

glwagner (approved):

glwagner:
Can you add a test?

ext/OceananigansXESMFExt.jl
navidcy: no worries then!

navidcy:
merge?

PR #4876, opened 2025-10-21
Enable `AveragedSpecifiedTimes` to be used in output writers

### LINKED ISSUE #4875: `AverageSpecifiedTimes` cannot be used
Not sure if `AveragedSpecifiedTimes` is meant to be used externally but when I tried to use it it doesn't work

Given an MWE like:
```julia
using Oceananigans
using Oceananigans.OutputWriters: AveragedSpecifiedTimes
using Printf

arch = CPU()
grid = RectilinearGrid(arch, size=(4, 4, 4), extent=(1, 1, 1))

simulation = Simulation(model, Δt=0.01, stop_time=1)

times = [0.1, 0.25, 0.5, 0.6, 1]

wall_time = Ref(time_ns())

function set_synthetic_values!(sim)
    t = time(sim)
    set!(sim.model, u=t, v=0, w=0)
    return nothing
end

simulation.callbacks[:set_synthetic_values] = Callback(set_synthetic_values!, IterationInterval(1))

simulation.output_writers[:jld2] = JLD2Writer(model, model.velocities;
                                              filename = "averagedspecified_times",
                                              schedule = AveragedSpecifiedTimes(times, window=0.1),
                                              # schedule = AveragedTimeInterval(0.2, window=0.1),
                                              # schedule = TimeInterval(0.2),
                                              overwrite_existing = true)

run!(simulation)

u = FieldTimeSeries("averagedspecified_times.jld2", "u")

u.times
u_values = interior(u, 1, 1, 1, :)
```
we end up with 
```julia
julia> u.times
102-element Vector{Float64}:
 0.0
 0.01
 0.02
 0.03
 0.04
 0.05
 0.060000000000000005
 0.07
 0.08
 0.09
 0.09999999999999999
 ⋮
 0.9099999999999971
 0.919999999999997
 0.9299999999999969
 0.9399999999999968
 0.9499999999999967
 0.9599999999999966
 0.9699999999999965
 0.9799999999999964
 0.9899999999999963
 0.9999999999999962
 1.0

julia> u_values
102-element view(::Array{Float64, 4}, 4, 4, 4, :) with eltype Float64:
 0.0
 0.009999999776482582
 0.019999999552965164
 0.029999999329447746
 0.03999999910593033
 0.05000000074505806
 0.05999999865889549
 0.07000000029802322
 0.07999999821186066
 0.09000000357627869
 0.10000000149011612
 ⋮
 0.9100000262260437
 0.9200000166893005
 0.9300000071525574
 0.9399999976158142
 0.949999988079071
 0.9599999785423279
 0.9700000286102295
 0.9800000190734863
 0.9900000095367432
 1.0
 1.0
```

which is not expected behavior. The outputwriter is expected to only write fields at `times` so that it only has 6 time frames in this case. What eventually ended up happening is that fields at all timesteps are written into the file.

xkykai:
Closes #4875.

This PR aims to extend output writers so that we can pass in schedules which are `AveragedSpecifiedTimes(times::Array{<:Number}, window=w::Array{<:Number}, ...)` into output writers. 

One compelling use case for this (at least for me) is to perform time averaging in `Year` units when running realistic coupled simulations in `ClimaOcean.jl`. For example, if I want to compute yearly averages over realistic years, each year would have different number of days, and the time interval between the end of each year is also not uniform. This calls for `schedule = AveragedSpecifiedTimes(...)`, thus making this feature request a compelling one.


xkykai:
Apart from writing some tests I think this is now ready for review.

For a script like:
```julia
using Oceananigans
using Oceananigans.OutputWriters: AveragedSpecifiedTimes
using Printf
using Random

arch = CPU()
grid = RectilinearGrid(arch, size=(4, 4, 4), extent=(1, 1, 1))

simulation = Simulation(model, Δt=0.0001, stop_time=1)

times = [0.1, 0.25, 0.5, 0.6, 1]
window = [0.1, 0.05, 0.2, 0.1, 0.3]
true_u = vcat([0], times .- window ./ 2)
# window = 0.1

perm = randperm(length(times))
times .= times[perm]
window .= window[perm]

schedule = AveragedSpecifiedTimes(times; window)

wall_time = Ref(time_ns())

function progress(sim)
    msg = @sprintf("iter: %d, time: %s, Δt: %.4f",
                    iteration(sim), prettytime(time(sim)), sim.Δt)

    elapsed = 1e-9 * (time_ns() - wall_time[])

    msg *= @sprintf(", max u: %6.3e, max v: %6.3e, max w: %6.3e, wall time: %s",
                    maximum(sim.model.velocities.u),
                    maximum(sim.model.velocities.v),
                    maximum(sim.model.velocities.w),
                    prettytime(elapsed))

    @info msg
    wall_time[] = time_ns()

    return nothing
end

function set_synthetic_values!(sim)
    t = time(sim)
    set!(sim.model, u=t, v=0, w=0)
    return nothing
end

simulation.callbacks[:progress] = Callback(progress, IterationInterval(1000))
simulation.callbacks[:set_synthetic_values] = Callback(set_synthetic_values!, IterationInterval(1))

simulation.output_writers[:jld2] = JLD2Writer(model, model.velocities;
                                              filename = "averagedspecified_times",
                                              schedule,
                                              overwrite_existing = true)

run!(simulation)

u = FieldTimeSeries("averagedspecified_times.jld2", "u")

@info "Recorded times: ", u.times
@info "Recorded u: ", interior(u, 1, 1, 1, :)
@info "Expected u: ", true_u
```
it gives an output of
```julia
[ Info: ("Recorded times: ", [0.0, 0.1, 0.25, 0.5, 0.6, 1.0])
[ Info: ("Recorded u: ", [0.0, 0.049949999898672104, 0.22495000064373016, 0.3999499976634979, 0.5499500036239624, 0.8499500155448914])
[ Info: ("Expected u: ", [0.0, 0.05, 0.225, 0.4, 0.5499999999999999, 0.85])
```

Now if `window` is a number i.e.
```julia
window = 0.1
```
we get results which are
```julia
[ Info: ("Recorded times: ", [0.0, 0.1, 0.25, 0.5, 0.6, 1.0])
[ Info: ("Recorded u: ", [0.0, 0.049949999898672104, 0.19994999468326569, 0.4499500095844269, 0.5499500036239624, 0.9499499797821045])
[ Info: ("Expected u: ", [0.0, 0.05, 0.2, 0.45, 0.5499999999999999, 0.95])
```
which are "correct". These answers are approximate as the `u` field is only set to `t` at every iteration, so the averaged results can't be exact. I was trying to come up with exact tests but I couldn't think of one so if people have suggestions I'd happy to use them!

I will write tests like this if people think this is good enough of when we can come up with tests that return correct answers exactly.

Now in terms of edge cases. When the windows overlap e.g. (note that the 4th window overlaps with the 3rd time point
```julia
times = [0.1, 0.25, 0.5, 0.6, 1]
window = [0.1, 0.05, 0.2, 0.11, 0.3]
```
we get an error of
```julia
ERROR: ArgumentError: Averaging windows overlap. Ensure that for each specified time tᵢ, tᵢ - windowᵢ ≥ tᵢ₋₁.
Stacktrace:
 [1] AveragedSpecifiedTimes(times::Vector{Float64}, window::Vector{Float64}; kw::@Kwargs{})
   @ Oceananigans.OutputWriters c:\Users\xinle\Dropbox\MIT\main_oceananigans\Oceananigans.jl\src\OutputWriters\windowed_time_average.jl:130
 [2] AveragedSpecifiedTimes
   @ c:\Users\xinle\Dropbox\MIT\main_oceananigans\Oceananigans.jl\src\OutputWriters\windowed_time_average.jl:120 [inlined]
 [3] #AveragedSpecifiedTimes#21
   @ c:\Users\xinle\Dropbox\MIT\main_oceananigans\Oceananigans.jl\src\OutputWriters\windowed_time_average.jl:117 [inlined]
 [4] top-level scope
   @ c:\Users\xinle\Dropbox\MIT\main_oceananigans\Oceananigans.jl\test_averagedspecified_times.jl:23
```

Similarly when `window` is a number e.g.
```julia
window = 0.11
```
we get 
```julia
ERROR: ArgumentError: Averaging window 0.11 is too large and causes overlapping windows. Ensure that for each specified time tᵢ, tᵢ - window ≥ tᵢ₋₁.
Stacktrace:
 [1] AveragedSpecifiedTimes(times::Vector{Float64}, window::Float64; kw::@Kwargs{})
   @ Oceananigans.OutputWriters c:\Users\xinle\Dropbox\MIT\main_oceananigans\Oceananigans.jl\src\OutputWriters\windowed_time_average.jl:135
 [2] AveragedSpecifiedTimes
   @ c:\Users\xinle\Dropbox\MIT\main_oceananigans\Oceananigans.jl\src\OutputWriters\windowed_time_average.jl:135 [inlined]
 [3] #AveragedSpecifiedTimes#21
   @ c:\Users\xinle\Dropbox\MIT\main_oceananigans\Oceananigans.jl\src\OutputWriters\windowed_time_average.jl:117 [inlined]
 [4] top-level scope
   @ c:\Users\xinle\Dropbox\MIT\main_oceananigans\Oceananigans.jl\test_averagedspecified_times.jl:23
```

In order to use the type dispatch style on `window` which in the API is a keyword argument I currently use a wrapper
https://github.com/CliMA/Oceananigans.jl/blob/35c41a517021ea1f0a8751d7f9c2611717425dea/src/OutputWriters/windowed_time_average.jl#L117
to reroute the function to the right dispatch with `window` as positional argument.

Alternatively there is also a method which is currently commented out 
https://github.com/CliMA/Oceananigans.jl/blob/6d168250410165f3cc61b319cf87415ec7d6a769/src/OutputWriters/windowed_time_average.jl#L140-L160
which checks the type of `window` in the function loop.

@glwagner @simone-silvestri @navidcy let me know what do you think and give suggestions on better model design!

PS: also why is `window` forced to be `Float64` rin `AveragedSpecifiedTimes` right now? 
https://github.com/CliMA/Oceananigans.jl/blob/f07428b824f1b2365beada01eeb7b5435a4afd0e/src/OutputWriters/windowed_time_average.jl#L107-L112
Why was that choice made?

src/OutputWriters/windowed_time_average.jl
```diff
@@ -102,17 +102,70 @@ Base.copy(sch::AveragedTimeInterval) = AveragedTimeInterval(sch.interval, window
 
 A schedule for averaging over windows that precede SpecifiedTimes.
 """
-mutable struct AveragedSpecifiedTimes <: AbstractSchedule
+mutable struct AveragedSpecifiedTimes{W <: Union{Float64, Vector{Float64}}} <: AbstractSchedule
```
glwagner:
```suggestion
mutable struct AveragedSpecifiedTimes{W} <: AbstractSchedule
```

src/OutputWriters/windowed_time_average.jl
```diff
@@ -102,17 +102,70 @@ Base.copy(sch::AveragedTimeInterval) = AveragedTimeInterval(sch.interval, window
...
     specified_times :: SpecifiedTimes
-    window :: Float64
+    window :: W
     stride :: Int
     collecting :: Bool
 end
 
+const VaryingWindowAveragedSpecifiedTimes = AveragedSpecifiedTimes{Vector{Float64}}
```
glwagner:
```suggestion
const VaryingWindowAveragedSpecifiedTimes = AveragedSpecifiedTimes{<:Vector}
```

glwagner: I don't think you want to limit to Float64. That won't work with dates.

xkykai: Should we relax `AveragedTimeInterval` as well?
https://github.com/CliMA/Oceananigans.jl/blob/f07428b824f1b2365beada01eeb7b5435a4afd0e/src/OutputWriters/windowed_time_average.jl#L15-L22

glwagner: yes! good catch. I can do that in another PR if you like

glwagner: in the above, `interval` and `window` have a "time period" type, while `first_actuation_time` has a "time" type. These are different when using dates.

xkykai: is there currently a `TimePeriod` type? I can try to define one.

Should we also make a `SpecifiedWindows` like `SpecifiedTimes` in the case for `AveragedSpecifiedTimes`?

glwagner: We don't want a specific type, because sometimes a "period" is `Float64` (the assumption here), but with dates, it may need to be `Dates.Second` or `Dates.Day` (for example). However, a "time type" can be `Float64` (the assumption here) or `DateTime`.

xkykai: should we relax it and make it parametric? or constrain it so that `interval :: Union{Float64, Dates.Period}`

glwagner: Right, we need to use type parameters, and there need to be two of them.

We don't use the type union approach because it is not concrete; with that approach types can no longer be inferred.

src/OutputWriters/windowed_time_average.jl
```diff
@@ -102,63 +103,54 @@ Base.copy(sch::AveragedTimeInterval) = AveragedTimeInterval(sch.interval, window
 
 A schedule for averaging over windows that precede SpecifiedTimes.
 """
-mutable struct AveragedSpecifiedTimes{W <: Union{Float64, Vector{Float64}}} <: AbstractSchedule
+mutable struct AveragedSpecifiedTimes{W} <: AbstractSchedule
     specified_times :: SpecifiedTimes
```
glwagner: Since `SpecifiedTimes` is not a concrete type, this should also be a type parameter

src/OutputWriters/windowed_time_average.jl
```diff
@@ -102,63 +103,54 @@ Base.copy(sch::AveragedTimeInterval) = AveragedTimeInterval(sch.interval, window
...
-function AveragedSpecifiedTimes(times, window::Vector{Float64}; kw...)
+function determine_epsilon(eltype)
+    if eltype <: AbstractFloat
+        return eps(eltype)
+    elseif eltype <: Period
+        return Second(0)
+    else
+        return 0
+    end
+end
```
glwagner:
```suggestion
determine_epsilon(eltype) = 0
determine_epsilon(eltype::AbstractType) = eps(eltype)
determine_epsilon(::Period) = Second(0)
```

the name of the function is slightly problematic

src/OutputWriters/windowed_time_average.jl
```diff
@@ -102,63 +103,54 @@ Base.copy(sch::AveragedTimeInterval) = AveragedTimeInterval(sch.interval, window
...
 
-    any(time_diff .- sorted_window .< -eps(eltype(window))) && throw(ArgumentError("Averaging windows overlap. Ensure that for each specified time tᵢ, tᵢ - windowᵢ ≥ tᵢ₋₁."))
+    epsilon = determine_epsilon(eltype(window))
+    any(time_diff .- sorted_window .< -epsilon) && throw(ArgumentError("Averaging windows overlap. Ensure that for each specified time tᵢ, tᵢ - windowᵢ ≥ tᵢ₋₁."))
 
     return AveragedSpecifiedTimes(SpecifiedTimes(sorted_times); window=sorted_window, kw...)
 end
 
-function AveragedSpecifiedTimes(times, window::Float64; kw...)
+function AveragedSpecifiedTimes(times, window::Union{<:Number, <:Period}; kw...)
```
glwagner:
```suggestion
function AveragedSpecifiedTimes(times, window; kw...)
```

can this be the fallback?

src/OutputWriters/windowed_time_average.jl
xkykai: you meant 
https://github.com/CliMA/Oceananigans.jl/blob/0efdbc82ccfe6b32d40a52245ac357ab0354c2b8/src/OutputWriters/windowed_time_average.jl#L120-L122
right?

I agree the name is terrible, but I haven't figured out a better alternative...

src/OutputWriters/windowed_time_average.jl
```diff
@@ -103,8 +103,8 @@ Base.copy(sch::AveragedTimeInterval) = AveragedTimeInterval(sch.interval, window
 
 A schedule for averaging over windows that precede SpecifiedTimes.
 """
-mutable struct AveragedSpecifiedTimes{W} <: AbstractSchedule
-    specified_times :: SpecifiedTimes
+mutable struct AveragedSpecifiedTimes{S<:SpecifiedTimes, W} <: AbstractSchedule
```
glwagner:
```suggestion
mutable struct AveragedSpecifiedTimes{S, W} <: AbstractSchedule
```

not that its bad, but we don't use this style right now in Oceananigans. Basically we want type restrictions to have a specific purpose

xkykai:
I believe this is now ready to be merged!

@glwagner @simone-silvestri @navidcy

src/OutputWriters/windowed_time_average.jl
```diff
@@ -91,30 +92,84 @@ function (sch::AveragedTimeInterval)(model)
...
+const PeriodTypeWindows = Union{Period, Vector{<:Period}}
+
+
+validate_windows(times, window) = nothing  # Fallback method
+
+function validate_windows(times, window::NumberTypeWindows)
+    tol = 10 * determine_epsilon(eltype(times))
+
+    gaps = diff(vcat(0, times))  # Prepend 0 to check first window against t=0
+    any(gaps .- window .< -tol) && throw(ArgumentError("Averaging windows overlap: some gaps between specified times are less than the window size."))
```
glwagner: is this returning something or `return nothing`? our style is to always have a `return`

src/OutputWriters/windowed_time_average.jl
```diff
@@ -91,30 +92,84 @@ function (sch::AveragedTimeInterval)(model)
...
+
 AveragedSpecifiedTimes(specified_times::SpecifiedTimes; window, stride=1) =
     AveragedSpecifiedTimes(specified_times, window, stride, false)
 
-AveragedSpecifiedTimes(times; kw...) = AveragedSpecifiedTimes(SpecifiedTimes(times); kw...)
+AveragedSpecifiedTimes(times; window, kw...) = AveragedSpecifiedTimes(times, window; kw...)
+
+determine_epsilon(eltype) = 0
+determine_epsilon(::Type{T}) where T <: AbstractFloat = eps(T)
+determine_epsilon(::Type{<:Period}) = Second(0)
```
glwagner: does this function exist elsewhere?

src/OutputWriters/windowed_time_average.jl
```diff
@@ -91,30 +92,84 @@ function (sch::AveragedTimeInterval)(model)
...
+
+const NumberTypeWindows = Union{Number, Vector{<:Number}}
+const PeriodTypeWindows = Union{Period, Vector{<:Period}}
+
+
```
glwagner:
```suggestion

```

src/OutputWriters/windowed_time_average.jl
```diff
@@ -91,30 +92,84 @@ function (sch::AveragedTimeInterval)(model)
...
+function validate_windows(times, window::PeriodTypeWindows)
+    if length(times) >= 2
+        window_starts = times .- window
+        prev_window_ends = times[1:end-1]
+        any(window_starts[2:end] .< prev_window_ends) && throw(ArgumentError("Averaging windows overlap: some gaps between specified times are less than the window size."))
+    end
+
+    # Note: We cannot check if the first window extends before the simulation start
+    # because the model clock is not available at construction time
```
glwagner: what should this return?

glwagner: we could in principle validate schedules in `run!`. I am not sure if that would be complicated.

src/OutputWriters/windowed_time_average.jl
glwagner, line 1: do you think its worth having two files, or one is ok?

src/Utils/prettytime.jl
```diff
@@ -64,3 +64,4 @@ function prettytimeunits(t, longform=true)
 end
 
 prettytime(dt::AbstractTime) = "$dt"
+prettytime(t::Array) = prettytime.(t)
```
glwagner:
```suggestion
prettytime(t::Array) = map(prettytime, t)
```

src/OutputWriters/windowed_time_average.jl
glwagner: I guess this is not really an eps. It's somethign else? bc there is no `eps(::Period)`?

src/OutputWriters/windowed_time_average.jl
xkykai: it is probably long and specific enough for a separate file

src/OutputWriters/windowed_time_average.jl
xkykai: as far as I can tell not explicitly, but `find_time_index` uses something similar. I've changed tolerance to be consistent with `find_time_index`

xkykai:
@glwagner @simone-silvestri @navidcy I believe this is ready for the next round of comments!

xkykai:
@glwagner let me know if there are issues. is this ready to go?

src/OutputWriters/averaged_specified_times.jl
```diff
@@ -0,0 +1,307 @@
...
+
+AveragedSpecifiedTimes(specified_times::SpecifiedTimes; window, stride=1) =
+    AveragedSpecifiedTimes(specified_times, window, stride, false)
+
+AveragedSpecifiedTimes(times; window, kw...) = AveragedSpecifiedTimes(times, window; kw...)
+
+determine_epsilon(eltype) = 0 # Fallback method
+determine_epsilon(::Type{T}) where T <: AbstractFloat = eps(T)
+determine_epsilon(::Type{<:Period}) = Second(0)
+determine_epsilon(::Type{<:DateTime}) = Second(0)
```
glwagner: I think this is confusing, can you call this function `overlap_tolerance(T)` and return `100 * eps(T)` for Float, and 0 otherwise?

test/test_averaged_specified_times.jl
glwagner, line 1: this test file is 642 lines long --- are all the tests necessary? make the tests parsimonious before merging.

src/OutputWriters/averaged_specified_times.jl
```diff
@@ -37,7 +36,7 @@ function validate_nonoverlapping_windows(times, window::NumberTypeWindows)
     # Check for overlapping windows between consecutive times
     # Works for both scalar and vector windows through broadcasting
     if length(times) >= 2
-        tol = 100 * determine_epsilon(eltype(times))
+        tol = 100 * overlap_tolerance(eltype(times))
```
glwagner: i think this should be

```julia
        tol = overlap_tolerance(eltype(times))
```

for it to make sense (that's what I suggested originally, but might have garbled it)

PR #4878, opened 2025-10-22
Clean up minor irregularities in tilted_bottom_boundary_layer

glwagner:
Sadly because diffusivity is \kappa, we need \varkappa for the Von Karman constant.


navidcy:
Docs break at 

```Julia
Δt₀ = 0.5 * minimum([minimum_zspacing(grid) / V∞, minimum_zspacing(grid)^2/κ])
```

because `κ` is never defined.

Suggestion: should we use the advective/diffusive CFL diagnostic to estimate `Δt₀`? this way we can showcase how those can be used/called.

glwagner:
> Docs break at
> 
> ```julia
> Δt₀ = 0.5 * minimum([minimum_zspacing(grid) / V∞, minimum_zspacing(grid)^2/κ])
> ```
> 
> because `κ` is never defined.
> 
> Suggestion: should we use the advective/diffusive CFL diagnostic to estimate `Δt₀`? this way we can showcase how those can be used/called.

For sure, that seems like a better approach. I think just `CFL` can do both?

Alternatively `model.closure.κ.b` would work

navidcy:
how about https://github.com/CliMA/Oceananigans.jl/pull/4878/commits/b5faac0ac120669454a46eb1269d23e2d4f404ee?

glwagner:
its great!

PR #4879, opened 2025-10-22
Fix RotatedLatitudeLongitudeGrid and add more tests

glwagner:
It had a few bugs.

### DIFF
--- a/src/OrthogonalSphericalShellGrids/rotated_latitude_longitude_grid.jl
+++ b/src/OrthogonalSphericalShellGrids/rotated_latitude_longitude_grid.jl
@@ -1,5 +1,5 @@
 using CubedSphere.SphericalGeometry
-using Oceananigans.Grids: LatitudeLongitudeGrid, Bounded
+using Oceananigans.Grids: LatitudeLongitudeGrid, Bounded, validate_lat_lon_grid_args
 using Oceananigans.Utils: KernelParameters
 using StaticArrays
 using LinearAlgebra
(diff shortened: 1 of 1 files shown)


navidcy:
There is a conflict; could you resolve?

src/OrthogonalSphericalShellGrids/rotated_latitude_longitude_grid.jl
```diff
@@ -90,14 +90,18 @@ function RotatedLatitudeLongitudeGrid(arch::AbstractArchitecture = CPU(),
                                       radius = Oceananigans.defaults.planet_radius,
                                       topology = (Bounded, Bounded, Bounded))
 
+    _, φ₀ = north_pole
+
+    if φ₀ < 0
+        throw(ArgumentError("North pole latitude must be >= 0."))
+    elseif φ₀ > 90
+        throw(ArgumentError("North pole latitude must be < 90 degrees."))
```
navidcy: what about φ₀ == 90?
the validation doesn't include it but the ArgumentError suggests that we can't have 90

navidcy (approved):

src/OrthogonalSphericalShellGrids/rotated_latitude_longitude_grid.jl
glwagner: ah sorry

glwagner: changing the comment to <=

PR #4889, opened 2025-10-29
Export HorizontalVectorInvariantScalarBiharmonicDiffusivity

siddharthabishnu:


glwagner:
it certainly is a mouthful. Would something like

```julia
closure = HorizontalBiharmonicViscosity(scheme=VectorInvariant())
```

be easier to read?

PR #4890, opened 2025-10-29
Improvements and fixes to `NetCDFWriter`

tomchor:
This PR contains: 

- A few improvements that will make https://github.com/CliMA/Oceananigans.jl/pull/4848 easier.
- A couple of fixes for edge cases.
- Include `dimension_type` to `NetCDFWriter` which controls the precision type for dimensions, with a default value of `Float64`.
- Perhaps most importantly, it brings the treatment of arrays by the `NetCDFWriter` closer to what Oceananigans does. In particular singleton dimensions are kept if the location has a well-defined value (i.e. the value is not `nothing`, which NetCDF can't handle). This means will remove singleton dimensions _only_ when:
  - the location on that dimension is nothing (e.g. a reduced field)
  - the topology in that dimension is `Flat`

 To- do list:

- [x] Should we enforce that spatial dimension is always Float64 to avoid alignment issues? We already do this with time for the same reason (see https://github.com/CliMA/Oceananigans.jl/issues/3056).
- [x] Special treatment for free surfaces
- [x] Do not squeeze singleton dimensions in slices
- [x] Add test for when the grid passed to `NetCDFWriter` is different from `model.grid`
- [x] Add test to make sure that singleton dimensions aren't improperly squeezed/removed

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -142,6 +142,9 @@ serve(dir="docs/build")
 - Add references to papers from the literature by adding bibtex to `oceananigans.bib`, and then
   a corresponding citation
 - Make use of cross-references with equations
+- In example code, use explicit imports as sparingly as possible. NEVER explicitly import a name that
+  is already exported by the user interface. Always rely on `using Oceananigans` for imports and keep
+  imports clean. Explicit imports should only be used for source code.
 
 ### Writing examples
 - Explain at the top of the file what a simulation is doing
@@ -152,6 +155,9 @@ serve(dir="docs/build")
 - Look at previous examples. New examples should add as much value as possible while
   remaining simple. This requires judiciously introducing new features and doing creative
   and surprising things with simulations that will spark readers' imagination.
+- Don't "over import". Use names that are exported by `using Oceananigans`. If there are
+  names that are not exported, but are needed in common/basic examples, consider
+  exporting those names from `Oceananigans.jl`.
 
 ## Important Files to Know
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.102.0"
+version = "0.102.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/TurbulenceClosures/TurbulenceClosures.jl
+++ b/src/TurbulenceClosures/TurbulenceClosures.jl
@@ -184,6 +184,10 @@ include("turbulence_closure_implementations/scalar_biharmonic_diffusivity.jl")
 # Only numbers, arrays, and functions supported now.
 @inline closure_coefficient(i, j, k, grid, C::Number) = C
 @inline closure_coefficient(i, j, k, grid, C::AbstractArray) = @inbounds C[i, j, k]
+@inline function closure_coefficient(i, j, k, grid, C)
+    x, y, z = node(i, j, k, grid, c, c, c)
+    return C(x, y, z)
+end
 
 include("turbulence_closure_implementations/anisotropic_minimum_dissipation.jl")
 include("turbulence_closure_implementations/Smagorinskys/Smagorinskys.jl")
--- a/src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/anisotropic_minimum_dissipation.jl
@@ -1,4 +1,5 @@
 using Oceananigans.Operators
+using Adapt: Adapt
 
 """
     AnisotropicMinimumDissipation{FT} <: AbstractTurbulenceClosure
@@ -16,6 +17,11 @@ struct AnisotropicMinimumDissipation{TD, PK, PN, PB} <: AbstractScalarDiffusivit
         return new{TD, PK, PN, PB}(Cν, Cκ, Cb)
     end
 end
+
+Adapt.adapt_structure(to, closure::AnisotropicMinimumDissipation{TD}) where TD =
+    AnisotropicMinimumDissipation{TD}(adapt(to, closure.Cν),
+                                      adapt(to, closure.Cκ),
+                                      adapt(to, closure.Cb))
 
 const AMD = AnisotropicMinimumDissipation
 
--- a/test/test_turbulence_closures.jl
+++ b/test/test_turbulence_closures.jl
@@ -159,6 +159,28 @@ function time_step_with_variable_discrete_diffusivity(arch)
     return true
 end
 
+function time_step_with_variable_AMD_coefficient(arch; use_field_coefficient=false)
+    grid = RectilinearGrid(arch, size=(4, 5, 6), extent=(1, 2, 3))
+
+    Cν_func(x, y, z) = exp(z) * cos(x) * cos(y)
+    Cκ_func(x, y, z) = exp(z) * cos(x) * cos(y)
+
+    if use_field_coefficient
+        Cν = CenterField(grid)
+        Cκ = CenterField(grid)
+        set!(Cν, Cν_func)
+        set!(Cκ, Cκ_func)
+    else
+        Cν = Cν_func
+        Cκ = Cκ_func
+    end
+
+    closure = AnisotropicMinimumDissipation(; Cν, Cκ)
+    model = NonhydrostaticModel(; grid, closure)
+    time_step!(model, 1)
+    return true
+end
+
 function time_step_with_tupled_closure(FT, arch)
     closure_tuple = (AnisotropicMinimumDissipation(FT), ScalarDiffusivity(FT))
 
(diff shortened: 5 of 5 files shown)


ext/OceananigansNCDatasetsExt.jl
```diff
@@ -57,48 +57,74 @@ const BuoyancyBoussinesqEOSModel = BuoyancyForce{<:BoussinesqSeawaterBuoyancy, g
 ##### Extend defVar to be able to write fields to NetCDF directly
 #####
 
+
+function squeeze_reduced_dimensions(data, reduced_dims)
+    # Fill missing indices from the right with 1s
+    indices = Any[:,:,:]
```
glwagner:
```suggestion
    indices = Any[:, : ,:]
```

what does this mean exactly?

tomchor: The comment is meant for the whole algorithm, not just that line. Basically it gets `Field`s that have `Nothing` in any of the locations and it gets rid of that dimension. (I'm calling these `reduced_dims` but there's probably a better name.)

So for example, if `data` comes from an array that has `loc = Center, Center, Nothing`, it'll reduce its parent data by indexing it as `data[:, :, 1]`; i.e. a 2D array in this case.

I'm testing this out (hence a draft PR) after some discussions with @ali-ramadhan. Basically we want to keep all NetCDF arrays 3D (to mimic Oceananigans' behavior) as much as possible, but in NetCDF there's no concept of location, so we can't easily pass `LZ = Nothing` to NetCDF. In these cases it probably makes sense to squeeze that dimension (or whatever the verb for this is) so that we know that dimension is reduced.

glwagner: Can you add attributes that indicate the location? We can add custom attributes to data even if they are not built in, right?

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -57,48 +57,78 @@ const BuoyancyBoussinesqEOSModel = BuoyancyForce{<:BoussinesqSeawaterBuoyancy, g
...
+
+    # Main.@infiltrate
+    # Get spatial dimensions excluding reduced dimensions (i.e. dimensions where `loc isa Nothing``)
+    reduced_dims = effective_reduced_dimensions(field)
 
-    spatial_dims_dict = Dict(dim_name => dim_data for (dim_name, dim_data) in zip(spatial_dims, nodes(field)))
-    create_spatial_dimensions!(ds, spatial_dims_dict, dimension_attributes; array_type=Array{eltype(field)})
+    # At the moment, this returns the full nodes even when the field is sliced.
+    # https://github.com/CliMA/Oceananigans.jl/pull/4814 will fix this in the future.
+    node_data = nodes(field; with_halos)
```
tomchor: Note to future self: revisit this after https://github.com/CliMA/Oceananigans.jl/pull/4814 is merged. It should make this line work out of the box and tests pass

ext/OceananigansNCDatasetsExt.jl
tomchor: I can (and in fact I probably will do it regardless), but that doesn't affect calculations with any of the software (that I'm aware of) which reads NetCDF. So broadcasting and other useful things wouldn't work. Without the singleton dimensions (i.e. with the change that I'm exploring here), those things will work out of the box for most software.

Also if we include the singleton reduced dimension, we still would need to pick a value for that length-1 coordinate. In Oceananigans, that value is `nothing` I think, but there's no analog to that for NetCDF, so we'd need to pick an arbitrary value like 1 or 0, which I'd like to avoid.

tomchor: Also to be clear, this PR would move the NetCDF behavior _closer_ to Oceananigans' behavior. Currently the NetCDF writer also drops singleton dimensions whenever we slice a field, even though that sliced dimension (or length 1) has a proper value. This PR would change the behavior and keep that sliced length-1 dimension in the NetCDF, just like Oceananigans does.

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -104,6 +104,13 @@ function squeezed_data(fd::AbstractField; array_type=Array{eltype(fd)}, with_hal
     return getindex(field_data_cpu, indices...)
 end
 
+function squeezed_data(fd::AbstractField; with_halos=false, kwargs...)
+    field_data = with_halos ? parent(fd) : interior(fd)
```
glwagner: does `construct_output` work here?

https://github.com/CliMA/Oceananigans.jl/blob/d3f0df86233828fb7c93b67591690c77102c3479/src/OutputWriters/output_construction.jl#L56

tomchor: No, they do different things. `squeeze_data()` should be specific to `NetCDFWriter` since it modifies the shape of the underlying array.

tomchor:
I think this PR is ready to be reviewed (the only failure is some keyring docs thing). I updated the top post to include a description of what was done. The most important thing is that slices with a singleton dimension now keep the singleton dimension, which is what Oceanigans does.

I'm only removing singleton dimensions in the data when there isn't a well-defined coordinate value associated with it. Namely when the dimension is `Flat` and when that direction is `Reduced`. In both cases the coordinate value in the singleton dimension is `nothing`, which NetCDF doesn't accept. To be clear, this was already done before this PR; I'm just keeping this behavior.

For the record, if I try to replace `nothing` with a placeholder value (like a `missing` flag which gets filled or a `NaN`) then broadcasting will not work properly (at least with all of the tools that I am aware of to read NetCDF). So I do think that's a necessary departure from the Oceananigans behavior to ensure a smooth user experience with the NetCDF files.

tomchor:
@glwagner @ali-ramadhan one thing I'd like to discuss is whether we should make the spatial dimensions always Float64. I've faced alignment problems before when working with two separate NetCDF files from the same simulation: one in Float64 for calculations, and another in Float32 for plotting. We already enforce Float64 with the time dimension, so I think it'd be natural to do it with xyz as well.

tomchor:
Btw, after this PR it'll be straightforward to merge https://github.com/CliMA/Oceananigans.jl/pull/4848 and then it should be straightforward to tackle https://github.com/CliMA/Oceananigans.jl/issues/3935 in one PR!

glwagner:
> @glwagner @ali-ramadhan one thing I'd like to discuss is whether we should make the spatial dimensions always Float64. I've faced alignment problems before when working with two separate NetCDF files from the same simulation: one in Float64 for calculations, and another in Float32 for plotting. We already enforce Float64 with the time dimension, so I think it'd be natural to do it with xyz as well.

happy for that. You could also make it user choice, with default Float64? That would be the Oceananigans Way.

ext/OceananigansNCDatasetsExt.jl
glwagner: Hmm ok. The only thing is that interpreting `with_halos` as the difference between `parent` and `interior` is different from what `construct_output` does? In particular the jld2 output writer uses the function `restrict_to_interior`:

https://github.com/CliMA/Oceananigans.jl/blob/d3f0df86233828fb7c93b67591690c77102c3479/src/OutputWriters/output_construction.jl#L11

Which I believe is more general than `interior` but I could be wrong.

glwagner: these would be different if requesting some custom index range. They arent different for the majority of cases, basically only for windowed fields, for hwich the user requests _further_ restriction to no halos.

tomchor:
> > @glwagner @ali-ramadhan one thing I'd like to discuss is whether we should make the spatial dimensions always Float64. I've faced alignment problems before when working with two separate NetCDF files from the same simulation: one in Float64 for calculations, and another in Float32 for plotting. We already enforce Float64 with the time dimension, so I think it'd be natural to do it with xyz as well.
> 
> happy for that. You could also make it user choice, with default Float64? That would be the Oceananigans Way.

Sounds good! Just added that option with a Float64 default, plus a test :)

ext/OceananigansNCDatasetsExt.jl
tomchor: `construct_output()` is already used before squeeze is called, so by the time we reach `squeeze_data()` the user field `fd` should already have the proper indices. So I guess maybe `interior()` isn't needed? (In which case `with_halos` also isn't needed.) I'll try to make this change and see if tests pass.

tomchor: Ah, okay, the difference between `interior` and `parent` is indeed unnecessary when writing an output through `NetCDFWriter` because `construct_output()` was already called on the output. However, it matter when writing a `Field` directly into a `NetCDF` file.

That said, you be right that I might be able to simplify the logic here and re-use code from `output_construction.jl` to decrease the amount of lines. I'll try doing that.

tomchor: I made the change. Indeed that's simpler code, thanks for catching that. Tests pass locally for me, so let's see on the CI.

tomchor: Tests are passing! I'm going to mark this as solved

glwagner: great work!

tomchor: Thanks! Can you approve this PR if there's no other comments? I really wanna start working on https://github.com/CliMA/Oceananigans.jl/pull/4848 again!

glwagner (approved):

PR #4005, opened 2024-12-17
Generalize Checkpointer for Distributed architectures

glwagner:
This PR generalized the Checkpointer for Distributed architectures by adding "_rank$rank" to the Checkpointer prefix. I believe this is the only change needed to support distributed checkpointing and pickup for simulations that use the same number of MPI ranks.

This PR depends on #4003. I added it because @simone-silvestri suggested "revamping" a PR for distributed Checkpointing. However, I think all that is really needed are these 5 lines so best to start simple.

TODO:
- [ ] test

Issue #3206, opened 2023-07-31
Restarting simulation with different halo size


simone-silvestri:
At the moment, restarting from checkpoints does not allow to restart from data with different halo size.
In particular, I encountered this issue when I wanted to restart a simulation with a different advection order from already evolved data.

We might want to support this feature since, when dealing with large complex simulations, a lot of trial and error has to be done in order to find the optimal parameters, which include order of advection (and sub-stepping size for the free surface, which involves halo size for distributed simulations)

If there is interest in this feature, I can start implementing it.

glwagner:
I would implement this as an extension of `set!(model, checkpoint_file)`.

glwagner:
Possibly you want to invent a new function (or keyword argument) for this? Something like

```julia
set!(model, checkpoint_file; with_halos=false)
```

Hopefully this can be done simply. I was going to suggest using `FieldTimeSeries`. But then I realized that can't be done with `Checkpointer`.

glwagner:
Note, the `Checkpointer` currently does not save field location --- it only serializes the field data:

https://github.com/CliMA/Oceananigans.jl/blob/f7acd8d0bd30dbe1ccb72854b6ea0ccab1eae0b5/src/OutputWriters/checkpointer.jl#L170

For a clean implementation of this feature, you'll need to extend the `Checkpointer` to also serialize the field location (and indices). Then you can use something like

```julia
parent_data = file["$name/data"]
loc = file["$name/location"]
indices = file["$name/indices"]
data = offset_data(parent_data, grid, loc, indices)
model_field = model_fields[name]

if with_halos
    copyto!(parent(model_field), parent_data)
else
    set!(model_field, data)
end
```

which extends these lines

https://github.com/CliMA/Oceananigans.jl/blob/f7acd8d0bd30dbe1ccb72854b6ea0ccab1eae0b5/src/OutputWriters/checkpointer.jl#L215-L217

PR #4892, opened 2025-10-30
(v0.104.0) Checkpointing simulations

### LINKED ISSUE #1249: Convenience function: `checkpoint(simulation)`
In some scripts manually checkpointing, e.g. at the end, is useful so having a `checkpoint(simulation)` function would be nice.

### LINKED ISSUE #2866: Checkpointer does not work for ShallowWaterModel
Checkpointer does not work with a ShallowWaterModel
```julia
using Oceananigans
using Oceananigans.Models: ShallowWaterModel
Lx, Ly, Lz = 2π, 20, 10
Nx, Ny = 128, 128

grid = RectilinearGrid(size = (Nx, Ny),
                       x = (0, Lx), y = (-Ly/2, Ly/2),
                       topology = (Periodic, Bounded, Flat))
gravitational_acceleration = 1
coriolis = FPlane(f=1)
model = ShallowWaterModel(; grid, coriolis, gravitational_acceleration,
                                 timestepper = :RungeKutta3,
                                 momentum_advection = WENO())
simulation.output_writers[:checkpointer] = Checkpointer(model,
            schedule=TimeInterval(5), prefix="model_checkpoint",
            properties = [:architecture, :grid, :clock, :coriolis, :closure, :velocities, 
            :tracers, :timestepper])
```
Errors with:

```
┌ Warning: particles is required for checkpointing. It will be added to checkpointed properties
└ @ Oceananigans.OutputWriters ~/.julia/packages/Oceananigans/42rFA/src/OutputWriters/checkpointer.jl:78
ERROR: Cannot checkpoint particles, it is not a model property!
```

This line https://github.com/CliMA/Oceananigans.jl/blob/c568df3d3729cc243de5453d2c60c1b34b8dd2b2/src/OutputWriters/checkpointer.jl#L77 could be modified to:

```julia
if rp ∉ properties && p ∈ propertynames(model)
```

However I don't know if the thing will look for it when trying to pickup.

### LINKED ISSUE #3670: Error in AveragedTimeInterval Output
Hi all,

Thanks, @glwagner for creating the [PR](https://github.com/CliMA/Oceananigans.jl/pull/3660) for solving the issue of `TimeInterval`. After merging it, I noticed a potential problem related to `AveragedTimeInterval`. Specifically, when resuming from a checkpoint, discontinuity seems to occur. For instance, in the figure below, the tidally averaged velocity is saved for each tidal period. The first run covers days 0-20, and the second run covers days 20-40. A noticeable discontinuity appears at the checkpoint transition from the first run.
image

Here is a MWE for this issue:
```julia
# this is a MWE for reproducing errors from using AveragedTimeInterval
using Oceananigans
using Printf
using Plots

""" Set up a simple simulation to test picking up from a checkpoint. """

function test_simulation(stop_time, Δt, δt, overwrite)
    grid = RectilinearGrid(size=(2), z = (-1,1), topology=(Oceananigans.Flat, Oceananigans.Flat, Oceananigans.Bounded))
    uᵢ(z) = 0
    T=2
    u_forcing(z, t) = 10*sin(2*pi/T*t)

                                  forcing = (u = u_forcing,),
                                  timestepper = :RungeKutta3)

   
    set!(model, u=uᵢ)

    u = model.velocities.u

    progress_message(sim) = @info string("Iter: ", iteration(sim), ", time: ", prettytime(sim))
    simulation.callbacks[:progress] = Callback(progress_message, TimeInterval(δt))

    simulation.output_writers[:timeavg] = NetCDFOutputWriter(model, (u=u,),
                                        filename = "timeavg.nc",
                                        schedule = AveragedTimeInterval(δt, window=δt),
                                        # schedule = TimeInterval(δt),
                                 
                                prefix = "test",
                                cleanup = true)

    return simulation
end

run(`sh -c "rm test_iteration*.jld2"`)

Δt = .01    # timestep (s)
T1 = 5      # first simulation stop time (s)
δt = .06   # progress message interval and output saving interval

simulation = test_simulation(T1, Δt, δt, true)
run!(simulation)

N = iteration(simulation)
checkpoint = "test_iteration$N.jld2"
simulation = test_simulation(T2, Δt, δt, false)
run!(simulation, pickup=checkpoint)

using NCDatasets
file = "timeavg.nc"
ds = Dataset(file)
t = ds["time"][:]
z = ds["zC"][:]
u = dropdims(ds["u"][:,:,:,:],dims = (1,2))
ln = Plots.plot(t[:],u[end,:],xlabel="time", ylabel="u",legend=nothing)
Plots.savefig("ln.png")
close(ds)
```
image

### LINKED ISSUE #3845: Restarting from a checkpoint should restore the time step?
It looks like that restarting from a checkpoint is not bit-for-bit? I think the issue is that when restarting from a checkpoint the time step is not restored -- it is still using the initial time step defined in `Simulation()`, not the `last_Δt` from the saved `Clock` object in the checkpoint file. See the example below, which is the output of the attached test case. 

In the pickup run I changed the onscreen output from every 10 iterations to every 1 iterations to see the time step. Rather than using the previous time step (5.973 s) from the checkpoint, the pickup run is using a time step of 10 s which is the value when defining `simulation = Simulation(model, Δt=10, stop_iteration=220)`. I’m using v0.91.5.

Initial run
```
Iteration: 0000, time: 0 seconds, Δt: 11 seconds, max(|u|) = 2.5e-01 ms⁻¹, wall time: 0 seconds
[ Info:     ... simulation initialization complete (13.909 seconds)
[ Info:     ... initial time step complete (4.696 seconds).
Iteration: 0010, time: 1.833 minutes, Δt: 11.212 seconds, max(|u|) = 2.6e-01 ms⁻¹, wall time: 19.116 seconds
Iteration: 0020, time: 3.702 minutes, Δt: 10.681 seconds, max(|u|) = 2.8e-01 ms⁻¹, wall time: 19.345 seconds
Iteration: 0030, time: 5.482 minutes, Δt: 10.215 seconds, max(|u|) = 2.9e-01 ms⁻¹, wall time: 19.629 seconds
Iteration: 0040, time: 7.185 minutes, Δt: 9.802 seconds, max(|u|) = 3.0e-01 ms⁻¹, wall time: 19.854 seconds
Iteration: 0050, time: 8.819 minutes, Δt: 9.433 seconds, max(|u|) = 3.1e-01 ms⁻¹, wall time: 20.082 seconds
Iteration: 0060, time: 10.391 minutes, Δt: 9.100 seconds, max(|u|) = 3.2e-01 ms⁻¹, wall time: 20.306 seconds
Iteration: 0070, time: 11.907 minutes, Δt: 8.798 seconds, max(|u|) = 3.3e-01 ms⁻¹, wall time: 20.559 seconds
Iteration: 0080, time: 13.374 minutes, Δt: 8.523 seconds, max(|u|) = 3.4e-01 ms⁻¹, wall time: 20.773 seconds
Iteration: 0090, time: 14.794 minutes, Δt: 8.270 seconds, max(|u|) = 3.5e-01 ms⁻¹, wall time: 21.001 seconds
Iteration: 0100, time: 16.173 minutes, Δt: 8.036 seconds, max(|u|) = 3.6e-01 ms⁻¹, wall time: 21.230 seconds
Iteration: 0110, time: 17.512 minutes, Δt: 7.820 seconds, max(|u|) = 3.7e-01 ms⁻¹, wall time: 21.456 seconds
Iteration: 0120, time: 18.815 minutes, Δt: 7.617 seconds, max(|u|) = 3.8e-01 ms⁻¹, wall time: 21.688 seconds
Iteration: 0130, time: 20.085 minutes, Δt: 7.424 seconds, max(|u|) = 3.9e-01 ms⁻¹, wall time: 21.914 seconds
Iteration: 0140, time: 21.322 minutes, Δt: 7.238 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 22.140 seconds
Iteration: 0150, time: 22.528 minutes, Δt: 7.046 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 22.366 seconds
Iteration: 0160, time: 23.703 minutes, Δt: 6.792 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 22.597 seconds
Iteration: 0170, time: 24.835 minutes, Δt: 6.500 seconds, max(|u|) = 4.2e-01 ms⁻¹, wall time: 22.822 seconds
Iteration: 0180, time: 25.918 minutes, Δt: 6.230 seconds, max(|u|) = 4.3e-01 ms⁻¹, wall time: 23.047 seconds
Iteration: 0190, time: 26.956 minutes, Δt: 5.973 seconds, max(|u|) = 4.3e-01 ms⁻¹, wall time: 23.271 seconds
Iteration: 0200, time: 27.952 minutes, Δt: 5.930 seconds, max(|u|) = 4.3e-01 ms⁻¹, wall time: 23.504 seconds
Iteration: 0210, time: 28.940 minutes, Δt: 5.979 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 24.139 seconds
[ Info: Simulation is stopping after running for 24.362 seconds.
[ Info: Model iteration 220 equals or exceeds stop iteration 220.
Iteration: 0220, time: 29.937 minutes, Δt: 5.898 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 24.362 seconds
```

Pickup run
```
[ Info:     ... simulation initialization complete (51.655 ms)
[ Info:     ... initial time step complete (5.017 seconds).
Iteration: 0201, time: 28.118 minutes, Δt: 10 seconds, max(|u|) = 4.2e-01 ms⁻¹, wall time: 0 seconds
Iteration: 0202, time: 28.285 minutes, Δt: 10 seconds, max(|u|) = 4.2e-01 ms⁻¹, wall time: 10.477 seconds
Iteration: 0203, time: 28.452 minutes, Δt: 10 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 10.499 seconds
Iteration: 0204, time: 28.618 minutes, Δt: 10 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 10.520 seconds
Iteration: 0205, time: 28.785 minutes, Δt: 10 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 10.544 seconds
Iteration: 0206, time: 28.952 minutes, Δt: 10 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 10.570 seconds
Iteration: 0207, time: 29.118 minutes, Δt: 10 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 10.595 seconds
Iteration: 0208, time: 29.285 minutes, Δt: 10 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 10.621 seconds
Iteration: 0209, time: 29.452 minutes, Δt: 10 seconds, max(|u|) = 4.1e-01 ms⁻¹, wall time: 10.647 seconds
Iteration: 0210, time: 29.618 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 10.673 seconds
Iteration: 0211, time: 29.722 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.392 seconds
Iteration: 0212, time: 29.825 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.413 seconds
Iteration: 0213, time: 29.928 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.435 seconds
Iteration: 0214, time: 30.031 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.459 seconds
Iteration: 0215, time: 30.134 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.485 seconds
Iteration: 0216, time: 30.237 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.510 seconds
Iteration: 0217, time: 30.340 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.536 seconds
Iteration: 0218, time: 30.443 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.559 seconds
Iteration: 0219, time: 30.546 minutes, Δt: 6.183 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.582 seconds
[ Info: Simulation is stopping after running for 14.606 seconds.
[ Info: Model iteration 220 equals or exceeds stop iteration 220.
Iteration: 0220, time: 30.649 minutes, Δt: 6.023 seconds, max(|u|) = 4.0e-01 ms⁻¹, wall time: 14.606 seconds
```
 
[run.jl.zip](https://github.com/user-attachments/files/17320901/run.jl.zip)

@glwagner

### LINKED ISSUE #4516: Warning message regarding free surface when using checkpoint is not consistent
Hi Oceananigans,

I would like to report a warning message that could be confusing.
Continuing the [discussion](https://github.com/CliMA/Oceananigans.jl/discussions/4504#discussioncomment-13170934), I am getting the following error when restarting the simulation from a pickup file:

```
julia> chkpt = 67500
julia> run!(simulation, pickup=chkpt)
------------------------------------
┌ Warning: Tendencies for η do not exist in checkpoint and could not be restored.
└ @ Oceananigans.OutputWriters [~/.julia/packages/Oceananigans/YRhAo/src/OutputWriters/checkpointer.jl:277](http://127.0.0.1:8989/lab/tree/Oceananigans/ocean_wind_mixing_and_convection/~/.julia/packages/Oceananigans/YRhAo/src/OutputWriters/checkpointer.jl#line=276)
```

The set up I have is the following:

```julia
using Oceananigans
using Oceananigans.Units

Lx = 100kilometers # east-west extent [m]
Ly = 100kilometers # north-south extent [m]
Lz = 500    # depth [m]

grid = RectilinearGrid(GPU(), size = (512, 512, 128),
                       x = (0, Lx),
                       y = (-Ly/2, Ly/2),
                       z = (-Lz, 0),
                       topology = (Periodic, Periodic, Bounded)
)

model = HydrostaticFreeSurfaceModel(; grid, 
            			    buoyancy = BuoyancyTracer(),
                                    coriolis = FPlane(f=1e-4),
                                    closure = CATKEVerticalDiffusivity(),
                                    tracers = (:b, :e), 
            			    boundary_conditions = (u=u_bcs, b=b_bcs),
                                    momentum_advection = WENO(order=5),
                                    tracer_advection = WENO(order=5)
)
```

The model seems to run despite the warning message.

### LINKED ISSUE #4857: Checkpointing that works
_Note_: This is a first draft. I might come back to edit this post with more refined thoughts. But thoughts are always welcome! Would be nice to get some consensus before any rewriting/refactoring.

I know that checkpointing has never really been great and I'd love to work on improving (or revamping, more likely) how Oceananigans.jl does checkpointing so it can be used across all models. I don't think I need to motivate the need for working checkpointing: ability to run on clusters with schedulers like Slurm and job time limits, the ability to spin up a model and checkpoint it, seeding an ensemble, etc.

## Current state

As has been pointed out many times, I think the main issue is that the current `Checkpointer` only serializes the `simulation.model` to disk but there is tons of important information in the `simulation` that needs to be serialized and picked up too.

To refresh my memory I'm just summarizing some of the existing documented issues with checkpointing:
* `WindowedTimeAverage`s (and all schedules in general) need to be checkpointed and picked up: https://github.com/CliMA/Oceananigans.jl/issues/3485, https://github.com/CliMA/Oceananigans.jl/issues/3670
* When picking up, `simulation.Δt` should also be picked up: https://github.com/CliMA/Oceananigans.jl/issues/3845
* Ability to pickup with a grid with different halo sizes: https://github.com/CliMA/Oceananigans.jl/issues/3206 (this might necessitate using new NetCDF output writers if saving the halos to NetCDF)
* The free surface height is not checkpointed: https://github.com/CliMA/Oceananigans.jl/issues/4516 (but only needed for `ExplicitFreeSurface`?)
* Should be easy to checkpoint manually: https://github.com/CliMA/Oceananigans.jl/issues/1249
* Right now we cannot checkpoint `ShallowWaterModel`: https://github.com/CliMA/Oceananigans.jl/issues/2866
* I think we can and should mandate (or mandate by default) checkpointing after `run!(simulation)` is done so avoid issues like https://github.com/CliMA/Oceananigans.jl/issues/2535 (although they may still occur if the Julia process crashes)

Issues that I'm not sure whether to tackle now or down the line:
* https://github.com/CliMA/Oceananigans.jl/issues/3399
* https://github.com/CliMA/Oceananigans.jl/issues/3543

It seems like a necessary refactor to fix these two issues is to only initialize output files after the checkpointer has had a chance to pickup, so it may be in scope.

## Proposed solution

I'm tempted to not think of the `Checkpointer` as an output writer[1] which depends on a `model` but rather as a `Callback` that depends on a `simulation`. The callback can then be easily called manually as well via `checkpoint(simulation)`. I would then advocate for moving checkpointing code out of the `OutputWriters` module and into a new `Checkpointers` module to isolate the checkpointing code and all the object-specific `state` and `restore_state!` functions.

Some new features that we'll need:
* A standardized interface for saving any object to a checkpoint file and reading back into a `simulation`. I know we have `saveproperty!` and `serializeproperty!` but I think we want something like e.g. `state(schedule::TimeInterval) -> Dict` and `restore_state!(schedule::TimeInterval, state::Dict)`.
* Unfortunately we cannot serialize functions to disk but maybe it would be good to print a warning when picking up and a function or BC or forcing function that was there when checkpointing is not there when picking up?

[1]: Even though the checkpointer writes data to disk, I don't think the data should be considered output to be analyzed. If anything, it is designed to serve as simulation _input_ in the future.

ali-ramadhan:
This PR refactors how the `Checkpointer` works by now checkpointing simulations, rather than just models. This is needed as the simulations (+ output writers, callbacks, etc.) all contain crucial information needed to properly restore/pickup a simulation and continue time stepping.

Basic design idea:
* We now have two new functions: `prognostic_state(obj)` which returns a named tuple corresponding to the prognostic state of `obj` and `restore_prognostic_state!(obj, state)` which restores `obj` based on information contained in `state` (which is a named tuple and is read from a checkpoint file).
* Objects are checkpointed recursively by serializing prognostic information to the JLD2 checkpoint file.
* The goal is for checkpointing to be flexible enough that we can very easily use it for different types of simulations, e.g. coupled simulations in ClimaOcean.jl by just defining `prognostic_state` and `restore_prognostic_state!`.

Right now I've only implemented proper checkpointing for non-hydrostatic model but it looks like it'll be straightforward to do it for hydrostatic and shallow water models. I'm working on adding comprehensive testing too.

Will continue working on this PR, but any feedback is very welcome!

Resolves #1249
Resolves #2866
Resolves #3670
Resolves #3845
Resolves #4516 
Resolves #4857

---

### Rhetorical aside

In general, the checkpointer is assuming that the simulation setup is the same. So only prognostic state information that changes will be checkpointed (e.g. field data, `TimeInterval.actuations`, etc.). The approach I have been taking (based on #4857) is to only checkpoint the prognostic state.

Should we operate under this assumption? I think so because not doing so can lead to a lot of undefined behavior. The checkpointer should not be responsible for checking that you set up the same simulation as the one that was checkpointed.

For example, take the `SpecifiedTimes` schedule. It has two properties `times` and `previous_actuation`. Since `previous_actuation` changes as the simulation runs, only `previous_actuation` needs to be checkpointed.

This leads to the possibility of the user changing `times` then picking up `previous_actuation` which can lead to undefined behavior. I think this is fine, because the checkpointer only works assuming you set up the same simulation as the one that was checkpointed.

Checkpointing both `times` and `previous_actuation` allows us to check that `times` is the same when restoring. But I don't think this is the checkpointer's responsibility.

### DIFF
--- a/docs/src/simulations/checkpointing.md
+++ b/docs/src/simulations/checkpointing.md
@@ -73,8 +73,8 @@ running the simulation:
 
 ```julia
 set!(simulation; checkpoint="path/to/file.jld2")  # restore from specific file
-set!(simulation; checkpoint=:latest)               # restore from latest checkpoint (requires Checkpointer)
-set!(simulation; iteration=12345)                  # restore from specific iteration (requires Checkpointer)
+set!(simulation; checkpoint=:latest)              # restore from latest checkpoint (requires Checkpointer)
+set!(simulation; iteration=12345)                 # restore from specific iteration (requires Checkpointer)
 ```
 
 ## Checkpointing on wall-clock time
--- a/src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
@@ -393,11 +393,9 @@ end
 #####
 
 function prognostic_state(fs::SplitExplicitFreeSurface)
-    return (
-        displacement = prognostic_state(fs.displacement),
-        barotropic_velocities = prognostic_state(fs.barotropic_velocities),
-        timestepper = prognostic_state(fs.timestepper),
-    )
+    return (displacement = prognostic_state(fs.displacement),
+            barotropic_velocities = prognostic_state(fs.barotropic_velocities),
+            timestepper = prognostic_state(fs.timestepper))
 end
 
 function restore_prognostic_state!(fs::SplitExplicitFreeSurface, state)
--- a/src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_timesteppers.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_timesteppers.jl
@@ -171,15 +171,13 @@ prognostic_state(::ForwardBackwardScheme) = nothing
 restore_prognostic_state!(ts::ForwardBackwardScheme, ::Nothing) = ts
 
 function prognostic_state(ts::AdamsBashforth3Scheme)
-    return (
-        ηᵐ   = prognostic_state(ts.ηᵐ),
-        ηᵐ⁻¹ = prognostic_state(ts.ηᵐ⁻¹),
-        ηᵐ⁻² = prognostic_state(ts.ηᵐ⁻²),
-        Uᵐ⁻¹ = prognostic_state(ts.Uᵐ⁻¹),
-        Uᵐ⁻² = prognostic_state(ts.Uᵐ⁻²),
-        Vᵐ⁻¹ = prognostic_state(ts.Vᵐ⁻¹),
-        Vᵐ⁻² = prognostic_state(ts.Vᵐ⁻²),
-    )
+    return (ηᵐ   = prognostic_state(ts.ηᵐ),
+            ηᵐ⁻¹ = prognostic_state(ts.ηᵐ⁻¹),
+            ηᵐ⁻² = prognostic_state(ts.ηᵐ⁻²),
+            Uᵐ⁻¹ = prognostic_state(ts.Uᵐ⁻¹),
+            Uᵐ⁻² = prognostic_state(ts.Uᵐ⁻²),
+            Vᵐ⁻¹ = prognostic_state(ts.Vᵐ⁻¹),
+            Vᵐ⁻² = prognostic_state(ts.Vᵐ⁻²))
 end
 
 function restore_prognostic_state!(ts::AdamsBashforth3Scheme, state)
--- a/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
@@ -333,19 +333,20 @@ timestepper(model::HydrostaticFreeSurfaceModel) = model.timestepper
 buoyancy_force(model::HydrostaticFreeSurfaceModel) = model.buoyancy
 buoyancy_tracers(model::HydrostaticFreeSurfaceModel) = model.tracers
 
-# For checkpointing
+#####
+##### Checkpointing
+#####
+
 function prognostic_state(model::HydrostaticFreeSurfaceModel)
-    return (
-        clock = prognostic_state(model.clock),
-        particles = prognostic_state(model.particles),
-        velocities = prognostic_state(model.velocities),
-        tracers = prognostic_state(model.tracers),
-        closure_fields = prognostic_state(model.closure_fields),
-        timestepper = prognostic_state(model.timestepper),
-        free_surface = prognostic_state(model.free_surface),
-        auxiliary_fields = prognostic_state(model.auxiliary_fields),
-        vertical_coordinate = prognostic_state(model.vertical_coordinate, model.grid),
-    )
+    return (clock = prognostic_state(model.clock),
+            particles = prognostic_state(model.particles),
+            velocities = prognostic_state(model.velocities),
+            tracers = prognostic_state(model.tracers),
+            closure_fields = prognostic_state(model.closure_fields),
+            timestepper = prognostic_state(model.timestepper),
+            free_surface = prognostic_state(model.free_surface),
+            auxiliary_fields = prognostic_state(model.auxiliary_fields),
+            vertical_coordinate = prognostic_state(model.vertical_coordinate, model.grid))
 end
 
 function restore_prognostic_state!(model::HydrostaticFreeSurfaceModel, state)
--- a/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
@@ -158,9 +158,7 @@ end
 #####
 
 function prognostic_state(fs::ImplicitFreeSurface)
-    return (
-        displacement = prognostic_state(fs.displacement),
-    )
+    return (; displacement = prognostic_state(fs.displacement))
 end
 
 function restore_prognostic_state!(fs::ImplicitFreeSurface, state)
--- a/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
+++ b/src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
@@ -313,18 +313,19 @@ end
 buoyancy_force(model::NonhydrostaticModel) = model.buoyancy
 buoyancy_tracers(model::NonhydrostaticModel) = model.tracers
 
-# For checkpointing
+#####
+##### Checkpointing
+#####
+
 function prognostic_state(model::NonhydrostaticModel)
-    return (
-        clock = prognostic_state(model.clock),
-        particles = prognostic_state(model.particles),
-        velocities = prognostic_state(model.velocities),
-        tracers = prognostic_state(model.tracers),
-        closure_fields = prognostic_state(model.closure_fields),
-        timestepper = prognostic_state(model.timestepper),
-        auxiliary_fields = prognostic_state(model.auxiliary_fields),
-        boundary_mass_fluxes = prognostic_state(model.boundary_mass_fluxes)
-    )
+    return (clock = prognostic_state(model.clock),
+            particles = prognostic_state(model.particles),
+            velocities = prognostic_state(model.velocities),
+            tracers = prognostic_state(model.tracers),
+            closure_fields = prognostic_state(model.closure_fields),
+            timestepper = prognostic_state(model.timestepper),
+            auxiliary_fields = prognostic_state(model.auxiliary_fields),
+            boundary_mass_fluxes = prognostic_state(model.boundary_mass_fluxes))
 end
 
 function restore_prognostic_state!(model::NonhydrostaticModel, state)
--- a/src/Models/ShallowWaterModels/shallow_water_model.jl
+++ b/src/Models/ShallowWaterModels/shallow_water_model.jl
@@ -243,14 +243,12 @@ shallow_water_fields(velocities, solution, tracers, ::VectorInvariantFormulation
 #####
 
 function prognostic_state(model::ShallowWaterModel)
-    return (
-        clock = prognostic_state(model.clock),
-        solution = prognostic_state(model.solution),
-        velocities = prognostic_state(model.velocities),
-        tracers = prognostic_state(model.tracers),
-        closure_fields = prognostic_state(model.closure_fields),
-        timestepper = prognostic_state(model.timestepper),
-    )
+    return (clock = prognostic_state(model.clock),
+            solution = prognostic_state(model.solution),
+            velocities = prognostic_state(model.velocities),
+            tracers = prognostic_state(model.tracers),
+            closure_fields = prognostic_state(model.closure_fields),
+            timestepper = prognostic_state(model.timestepper))
 end
 
 function restore_prognostic_state!(model::ShallowWaterModel, state)
--- a/src/OutputWriters/checkpointer.jl
+++ b/src/OutputWriters/checkpointer.jl
@@ -259,9 +259,7 @@ restore_prognostic_state!(r::Ref, value) = (r[] = value; r)
 #####
 
 function prognostic_state(checkpointer::Checkpointer)
-    return (
-        schedule = prognostic_state(checkpointer.schedule),
-    )
+    return (; schedule = prognostic_state(checkpointer.schedule))
 end
 
 function restore_prognostic_state!(checkpointer::Checkpointer, state)
@@ -284,11 +282,9 @@ function prognostic_state(writer::Union{JLD2Writer, NetCDFWriter})
                              for (name, output) in pairs(writer.outputs)
                              if output isa WindowedTimeAverage)
 
-    return (
-        schedule = prognostic_state(writer.schedule),
-        part = writer.part,
-        windowed_time_averages = isempty(wta_outputs) ? nothing : wta_outputs,
-    )
+    return (schedule = prognostic_state(writer.schedule),
+            part = writer.part,
+            windowed_time_averages = isempty(wta_outputs) ? nothing : wta_outputs)
 end
 
 function restore_prognostic_state!(writer::Union{JLD2Writer, NetCDFWriter}, state)
--- a/src/TimeSteppers/TimeSteppers.jl
+++ b/src/TimeSteppers/TimeSteppers.jl
@@ -80,12 +80,13 @@ function first_time_step!(model::AbstractModel{<:QuasiAdamsBashforth2TimeStepper
     return nothing
 end
 
-# Checkpointing
+#####
+##### Checkpointing
+#####
+
 function prognostic_state(timestepper::AbstractTimeStepper)
-    return (
-        Gⁿ = prognostic_state(timestepper.Gⁿ),
-        G⁻ = prognostic_state(timestepper.G⁻),
-    )
+    return (Gⁿ = prognostic_state(timestepper.Gⁿ),
+            G⁻ = prognostic_state(timestepper.G⁻))
 end
 
 function restore_prognostic_state!(timestepper::AbstractTimeStepper, state)
--- a/src/TimeSteppers/split_hydrostatic_runge_kutta_3.jl
+++ b/src/TimeSteppers/split_hydrostatic_runge_kutta_3.jl
@@ -169,12 +169,13 @@ function cache_previous_fields!(model)
     return nothing
 end
 
-# Checkpointing
+#####
+##### Checkpointing
+#####
+
 function prognostic_state(timestepper::SplitRungeKutta3TimeStepper)
-    return (
-        Gⁿ = prognostic_state(timestepper.Gⁿ),
-        Ψ⁻ = prognostic_state(timestepper.Ψ⁻),
-    )
+    return (Gⁿ = prognostic_state(timestepper.Gⁿ),
+            Ψ⁻ = prognostic_state(timestepper.Ψ⁻))
 end
 
 function restore_prognostic_state!(timestepper::SplitRungeKutta3TimeStepper, state)
--- a/src/Models/HydrostaticFreeSurfaceModels/explicit_free_surface.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/explicit_free_surface.jl
@@ -179,9 +179,7 @@ end
 #####
 
 function prognostic_state(fs::ExplicitFreeSurface)
-    return (
-        η = prognostic_state(fs.η),
-    )
+    return (; η = prognostic_state(fs.η))
 end
 
 function restore_prognostic_state!(fs::ExplicitFreeSurface, state)
--- a/src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
+++ b/src/Models/HydrostaticFreeSurfaceModels/z_star_vertical_spacing.jl
@@ -233,15 +233,13 @@ restore_prognostic_state!(::ZCoordinate, grid, ::Nothing) = nothing
 
 function prognostic_state(zc::ZStarCoordinate, grid)
     z = grid.z
-    return (
-        storage = prognostic_state(zc.storage),
-        ηⁿ   = prognostic_state(z.ηⁿ),
-        σᶜᶜⁿ = prognostic_state(z.σᶜᶜⁿ),
-        σᶠᶜⁿ = prognostic_state(z.σᶠᶜⁿ),
-        σᶜᶠⁿ = prognostic_state(z.σᶜᶠⁿ),
-        σᶠᶠⁿ = prognostic_state(z.σᶠᶠⁿ),
-        σᶜᶜ⁻ = prognostic_state(z.σᶜᶜ⁻),
-    )
+    return (storage = prognostic_state(zc.storage),
+            ηⁿ   = prognostic_state(z.ηⁿ),
+            σᶜᶜⁿ = prognostic_state(z.σᶜᶜⁿ),
+            σᶠᶜⁿ = prognostic_state(z.σᶠᶜⁿ),
+            σᶜᶠⁿ = prognostic_state(z.σᶜᶠⁿ),
+            σᶠᶠⁿ = prognostic_state(z.σᶠᶠⁿ),
+            σᶜᶜ⁻ = prognostic_state(z.σᶜᶜ⁻))
 end
 
 function restore_prognostic_state!(zc::ZStarCoordinate, grid, state)
--- a/src/OutputWriters/windowed_time_average.jl
+++ b/src/OutputWriters/windowed_time_average.jl
@@ -108,13 +108,14 @@ end
 TimeInterval(sch::AveragedTimeInterval) = TimeInterval(sch.interval)
 Base.copy(sch::AveragedTimeInterval) = AveragedTimeInterval(sch.interval, window=sch.window, stride=sch.stride)
 
-# Checkpointing
+#####
+##### Checkpointing
+#####
+
 function prognostic_state(schedule::AveragedTimeInterval)
-    return (
-        first_actuation_time = schedule.first_actuation_time,
-        actuations = schedule.actuations,
-        collecting = schedule.collecting,
-    )
+    return (first_actuation_time = schedule.first_actuation_time,
+            actuations = schedule.actuations,
+            collecting = schedule.collecting)
 end
 
 function restore_prognostic_state!(schedule::AveragedTimeInterval, state)
@@ -214,12 +215,9 @@ function end_of_window(schedule::AveragedSpecifiedTimes, clock)
     return clock.time >= next_time - eps(next_time)
 end
 
-# Checkpointing
 function prognostic_state(schedule::AveragedSpecifiedTimes)
-    return (
-        specified_times = prognostic_state(schedule.specified_times),
-        collecting = schedule.collecting,
-    )
+    return (specified_times = prognostic_state(schedule.specified_times),
+            collecting = schedule.collecting)
 end
 
 function restore_prognostic_state!(schedule::AveragedSpecifiedTimes, state)
@@ -350,15 +348,12 @@ end
 # So it can be used as a Diagnostic
 run_diagnostic!(wta::WindowedTimeAverage, model) = advance_time_average!(wta, model)
 
-# Checkpointing
 function prognostic_state(wta::WindowedTimeAverage)
-    return (
-        result = prognostic_state(wta.result),
-        window_start_time = wta.window_start_time,
-        window_start_iteration = wta.window_start_iteration,
-        previous_collection_time = wta.previous_collection_time,
-        schedule = prognostic_state(wta.schedule),
-    )
+    return (result = prognostic_state(wta.result),
+            window_start_time = wta.window_start_time,
+            window_start_iteration = wta.window_start_iteration,
+            previous_collection_time = wta.previous_collection_time,
+            schedule = prognostic_state(wta.schedule))
 end
 
 function restore_prognostic_state!(wta::WindowedTimeAverage, state)
--- a/src/Simulations/callback.jl
+++ b/src/Simulations/callback.jl
@@ -139,9 +139,7 @@ end
 validate_schedule(func, schedule) = schedule
 
 function prognostic_state(callback::Callback)
-    return (
-        schedule = prognostic_state(callback.schedule),
-    )
+    return (; schedule = prognostic_state(callback.schedule))
 end
 
 function restore_prognostic_state!(callback::Callback, state)
--- a/src/Simulations/simulation.jl
+++ b/src/Simulations/simulation.jl
@@ -284,17 +284,15 @@ write_output!(writer::JLD2Writer,   sim::Simulation) = write_output!(writer, sim
 write_output!(writer::NetCDFWriter, sim::Simulation) = write_output!(writer, sim.model)
 
 function prognostic_state(sim::Simulation)
-    return (
-        model = prognostic_state(sim.model),
-        Δt = sim.Δt,
-        diagnostics = prognostic_state(sim.diagnostics),
-        output_writers = prognostic_state(sim.output_writers),
-        callbacks = prognostic_state(sim.callbacks),
-        run_wall_time = sim.run_wall_time,
-        align_time_step = sim.align_time_step,
-        verbose = sim.verbose,
-        minimum_relative_step = sim.minimum_relative_step
-    )
+    return (model = prognostic_state(sim.model),
+            Δt = sim.Δt,
+            diagnostics = prognostic_state(sim.diagnostics),
+            output_writers = prognostic_state(sim.output_writers),
+            callbacks = prognostic_state(sim.callbacks),
+            run_wall_time = sim.run_wall_time,
+            align_time_step = sim.align_time_step,
+            verbose = sim.verbose,
+            minimum_relative_step = sim.minimum_relative_step)
 end
 
 function restore_prognostic_state!(sim::Simulation, state)
--- a/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
@@ -483,10 +483,8 @@ const DirectionallyAveragedSmagorinskyFields = NamedTuple{(:νₑ, :Σ, :Σ̄, :
 const LagrangianAveragedSmagorinskyFields = NamedTuple{(:νₑ, :Σ, :Σ̄, :𝒥ᴸᴹ, :𝒥ᴹᴹ, :𝒥ᴸᴹ⁻, :𝒥ᴹᴹ⁻, :previous_compute_time)}
 
 function prognostic_state(cf::DirectionallyAveragedSmagorinskyFields)
-    return (
-        𝒥ᴸᴹ = prognostic_state(cf.𝒥ᴸᴹ),
-        𝒥ᴹᴹ = prognostic_state(cf.𝒥ᴹᴹ),
-    )
+    return (𝒥ᴸᴹ = prognostic_state(cf.𝒥ᴸᴹ),
+            𝒥ᴹᴹ = prognostic_state(cf.𝒥ᴹᴹ))
 end
 
 function restore_prognostic_state!(cf::DirectionallyAveragedSmagorinskyFields, state)
@@ -496,13 +494,11 @@ function restore_prognostic_state!(cf::DirectionallyAveragedSmagorinskyFields, s
 end
 
 function prognostic_state(cf::LagrangianAveragedSmagorinskyFields)
-    return (
-        𝒥ᴸᴹ = prognostic_state(cf.𝒥ᴸᴹ),
-        𝒥ᴹᴹ = prognostic_state(cf.𝒥ᴹᴹ),
-        𝒥ᴸᴹ⁻ = prognostic_state(cf.𝒥ᴸᴹ⁻),
-        𝒥ᴹᴹ⁻ = prognostic_state(cf.𝒥ᴹᴹ⁻),
-        previous_compute_time = cf.previous_compute_time[],
-    )
+    return (𝒥ᴸᴹ = prognostic_state(cf.𝒥ᴸᴹ),
+            𝒥ᴹᴹ = prognostic_state(cf.𝒥ᴹᴹ),
+            𝒥ᴸᴹ⁻ = prognostic_state(cf.𝒥ᴸᴹ⁻),
+            𝒥ᴹᴹ⁻ = prognostic_state(cf.𝒥ᴹᴹ⁻),
+            previous_compute_time = cf.previous_compute_time[])
 end
 
 function restore_prognostic_state!(cf::LagrangianAveragedSmagorinskyFields, state)
--- a/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/catke_vertical_diffusivity.jl
@@ -428,14 +428,12 @@ end
 #####
 
 function prognostic_state(cf::CATKEDiffusivityFields)
-    return (
-        previous_compute_time = cf.previous_compute_time[],
-        previous_velocities = prognostic_state(cf.previous_velocities),
-        Jᵇ = prognostic_state(cf.Jᵇ),
-        κu = prognostic_state(cf.κu),
-        κc = prognostic_state(cf.κc),
-        κe = prognostic_state(cf.κe),
-    )
+    return (previous_compute_time = cf.previous_compute_time[],
+            previous_velocities = prognostic_state(cf.previous_velocities),
+            Jᵇ = prognostic_state(cf.Jᵇ),
+            κu = prognostic_state(cf.κu),
+            κc = prognostic_state(cf.κc),
+            κe = prognostic_state(cf.κe))
 end
 
 function restore_prognostic_state!(cf::CATKEDiffusivityFields, state)
--- a/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/tke_dissipation_vertical_diffusivity.jl
@@ -437,14 +437,12 @@ end
 #####
 
 function prognostic_state(cf::TKEDissipationDiffusivityFields)
-    return (
-        previous_compute_time = cf.previous_compute_time[],
-        previous_velocities = prognostic_state(cf.previous_velocities),
-        κu = prognostic_state(cf.κu),
-        κc = prognostic_state(cf.κc),
-        κe = prognostic_state(cf.κe),
-        κϵ = prognostic_state(cf.κϵ),
-    )
+    return (previous_compute_time = cf.previous_compute_time[],
+            previous_velocities = prognostic_state(cf.previous_velocities),
+            κu = prognostic_state(cf.κu),
+            κc = prognostic_state(cf.κc),
+            κe = prognostic_state(cf.κe),
+            κϵ = prognostic_state(cf.κϵ))
 end
 
 function restore_prognostic_state!(cf::TKEDissipationDiffusivityFields, state)
--- a/src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/ri_based_vertical_diffusivity.jl
@@ -379,12 +379,10 @@ end
 const RiBasedVerticalDiffusivityFields = NamedTuple{(:κc, :κu, :Ri, :previous_compute_time)}
 
 function prognostic_state(closure_fields::RiBasedVerticalDiffusivityFields)
-    return (
-        κc = prognostic_state(closure_fields.κc),
-        κu = prognostic_state(closure_fields.κu),
-        Ri = prognostic_state(closure_fields.Ri),
-        previous_compute_time = closure_fields.previous_compute_time[],
-    )
+    return (κc = prognostic_state(closure_fields.κc),
+            κu = prognostic_state(closure_fields.κu),
+            Ri = prognostic_state(closure_fields.Ri),
+            previous_compute_time = closure_fields.previous_compute_time[])
 end
 
 function restore_prognostic_state!(closure_fields::RiBasedVerticalDiffusivityFields, state)
--- a/src/Utils/schedules.jl
+++ b/src/Utils/schedules.jl
@@ -96,10 +96,8 @@ function schedule_aligned_time_step(schedule::TimeInterval, clock, Δt)
 end
 
 function prognostic_state(schedule::TimeInterval)
-    return (
-        first_actuation_time = schedule.first_actuation_time,
-        actuations = schedule.actuations,
-    )
+    return (first_actuation_time = schedule.first_actuation_time,
+            actuations = schedule.actuations)
 end
 
 function restore_prognostic_state!(schedule::TimeInterval, state)
@@ -267,9 +265,7 @@ function specified_times_str(st)
 end
 
 function prognostic_state(schedule::SpecifiedTimes)
-    return (
-        previous_actuation = schedule.previous_actuation,
-    )
+    return (; previous_actuation = schedule.previous_actuation)
 end
 
 function restore_prognostic_state!(schedule::SpecifiedTimes, state)
@@ -314,10 +310,8 @@ schedule_aligned_time_step(schedule::ConsecutiveIterations, clock, Δt) =
     schedule_aligned_time_step(schedule.parent, clock, Δt)
 
 function prognostic_state(schedule::ConsecutiveIterations)
-    return (
-        parent = prognostic_state(schedule.parent),
-        previous_parent_actuation_iteration = schedule.previous_parent_actuation_iteration,
-    )
+    return (parent = prognostic_state(schedule.parent),
+            previous_parent_actuation_iteration = schedule.previous_parent_actuation_iteration)
 end
 
 function restore_prognostic_state!(schedule::ConsecutiveIterations, state)


src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -261,3 +261,39 @@ initialize!(model::HydrostaticFreeSurfaceModel) = initialize_free_surface!(model
...
+    return (
+        clock = prognostic_state(model.clock),
+        particles = prognostic_state(model.particles),
+        velocities = prognostic_state(model.velocities),
+        tracers = prognostic_state(model.tracers),
+        pressure = prognostic_state(model.pressure),
+        diffusivity_fields = prognostic_state(model.diffusivity_fields),
+        timestepper = prognostic_state(model.timestepper),
+        auxiliary_fields = prognostic_state(model.auxiliary_fields),
+        free_surface = prognostic_state(model.free_surface),
```
```suggestion
        diffusivity_fields = prognostic_state(model.diffusivity_fields),
        timestepper = prognostic_state(model.timestepper),
        free_surface = prognostic_state(model.free_surface),
```

we also need the forcing

glwagner: but pressure is diagnostic by definition and I think we can define auxiliary fields as diagnostic too

glwagner: (at least, it would be challenging/complicated for a user to make an auxiliary field prognostic. it is possible but not natural)

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -320,14 +320,16 @@ function restore_prognostic_state!(model::NonhydrostaticModel, state)
...
     if length(model.tracers) > 0
         restore_prognostic_state!(model.tracers, state.tracers)
     end
 
-    restore_prognostic_state!(model.pressures, state.pressures)
-    restore_prognostic_state!(model.diffusivity_fields, state.diffusivity_fields)
-    restore_prognostic_state!(model.timestepper, state.timestepper)
+    if length(model.closure_fields) > 0
+        restore_prognostic_state!(model.closure_fields, state.closure_fields)
+    end
```
glwagner: should we handle this with dispatch?

glwagner: also for dispatch I think this may need to know the closure as well. This is a unique object that is "managed" by the closure but doesn't store much identifying info. We could also change that design, but might want to dedicate / test in a prior PR

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -320,14 +320,16 @@ function restore_prognostic_state!(model::NonhydrostaticModel, state)
     restore_prognostic_state!(model.clock, state.clock)
     restore_prognostic_state!(model.particles, state.particles)
     restore_prognostic_state!(model.velocities, state.velocities)
+    restore_prognostic_state!(model.pressures, state.pressures)
+    restore_prognostic_state!(model.timestepper, state.timestepper)
 
     if length(model.tracers) > 0
         restore_prognostic_state!(model.tracers, state.tracers)
     end
```
glwagner: it might also be possible to catch this with dispatch on `::NamedTuple{}` ( I think that's the rigfht way to write empty NamedTuple)

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
ali-ramadhan: Thanks! I've removed pressure.

I've kept in the auxiliary fields because while they may not be prognostic, I can think of a few ways a user would use an auxiliary field in a way that it's value cannot be diagnosed from just the current model state. For example, if an auxiliary field accumulates a value or involves a time integral.

Not that I've used an auxiliary field this way but I suppose it's possible?

Will also add in forcings soon.

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
ali-ramadhan: We could! And it would make these functions simpler! Will do this soon as well.

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
ali-ramadhan: For sure. Right now I'm still working on getting all the existing tests to pass, but once they do I want to start testing checkpointing more and more complex simulations. As part of it, we should also test closures that use `model.closure_fields`.

ali-ramadhan:
Looks like tests will all pass :tada: I'll start testing the checkpointing of increasingly complex simulations while iterating on the design! This way we'll be able to weed out most bugs and issues.

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
glwagner: > Not that I've used an auxiliary field this way but I suppose it's possible?

Hmm, thinking about it more, this is easily implementable with a callback. So I agree with you.

src/Models/ShallowWaterModels/shallow_water_model.jl
```diff
@@ -233,3 +234,35 @@ shallow_water_velocities(model::ShallowWaterModel) = shallow_water_velocities(mo
...
+    restore_prognostic_state!(model.clock, state.clock)
+    restore_prognostic_state!(model.solution, state.solution)
+    restore_prognostic_state!(model.velocities, state.velocities)
+    restore_prognostic_state!(model.timestepper, state.timestepper)
+
+    if length(model.tracers) > 0
+        restore_prognostic_state!(model.tracers, state.tracers)
+    end
+
+    if !isnothing(model.closure_fields)
```
glwagner: handle with dispatch?

glwagner:
CATKE is an edge case so it should be tested

ali-ramadhan:
Looking at https://github.com/CliMA/Oceananigans.jl/pull/4005 I should also test that checkpointing works for distributed simulations.

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -259,5 +259,44 @@ validate_momentum_advection(momentum_advection, grid::OrthogonalSphericalShellGr
...
+    end
+
+    if length(model.auxiliary_fields) > 0
+        restore_prognostic_state!(model.auxiliary_fields, state.auxiliary_fields)
+    end
+
+    return model
+end
+
+
```
giordano:
```suggestion
```

ali-ramadhan:
Ok I think this PR is finally ready for review! I went through and added checkpointing support + tests to different time steppers, models, schedules, turbulence closures, etc. but let me know if you feel like I missed something.

This PR has a few breaking changes so we should tag v0.103.0 once it's merged:
1. The `Checkpointer` constructor no longer has the `properties` keyword. `prognostic_state` decides exactly what gets checkpointed.
2. `set!(model, filepath::String)` is now `set!(simulation, pickup)` where `pickup` is `Bool`, `Int`, or `String`.
3. Checkpoint files are not backwards compatible. You will not be able to restore from an old (pre v0.103.0) checkpoint file. Although this has generally always been true.

---

The PR is getting quite large and distributed CI is not working, so I'll test checkpointing support for distributed models in a subsequent PR.

tomchor:
> This leads to the possibility of the user changing times then picking up previous_actuation which can lead to undefined behavior. I think this is fine, because the checkpointer only works assuming you set up the same simulation as the one that was checkpointed.
>
> Checkpointing both times and previous_actuation allows us to check that times is the same when restoring. But I don't think this is the checkpointer's responsibility.

There's been a lot of progress since this PR was open, so I'm wondering if this is still true. I don't oppose this, but I also don't oppose saving non-prognostic information when easy and throwing a warning to the user if we flag differences there. Mistakes happen and I think this sort of user-friendliness can avoid some headache. Again, only when it's to do so though. I definitely don't think `Checkpointer` needs to check every single aspect of a `Simulation` to see if everything matches up.

docs/src/simulations/checkpointing.md
```diff
@@ -59,3 +65,37 @@ run!(simulation, pickup=true)
...
 Or, use `pickup=filepath`, where `filepath` is a string, to pickup from a specific file located
 at `filepath`.
+
+The [`set!`](@ref) function can also be used to restore from a checkpoint without immediately
+running the simulation:
+
+```julia
+set!(simulation, filepath)   # restore from specific file (no Checkpointer required)
+set!(simulation, true)       # restore from latest checkpoint (requires Checkpointer)
+set!(simulation, iteration)  # restore from specific iteration (requires Checkpointer)
```
tomchor: :heart:

docs/src/simulations/checkpointing.md
```diff
@@ -59,3 +65,37 @@ run!(simulation, pickup=true)
...
+For cluster jobs with time limits, use `WallTimeInterval` to checkpoint based on elapsed
+wall-clock time rather than simulation time or iterations:
+
+```julia
+# Checkpoint every 30 minutes of wall-clock time
+Checkpointer(model, schedule=WallTimeInterval(30minute), prefix="checkpoint")
+```
+
+This ensures checkpoints are saved regularly even if individual time steps vary significantly.
+
```
tomchor, line 91: This may be a job for another PR, but should we add option to write a `Checkpoint` when the `wall_time_limit` is exceeded? It might be super useful exactly for this sort of cluster situation such that we don't run any time-step twice regardless of how many `pickups` a simulation ends up needing.

src/Fields/field.jl
```diff
@@ -860,3 +860,18 @@ end
...
 nodes(f::Field; kwargs...) = nodes(f.grid, instantiated_location(f)...; indices=indices(f), kwargs...)
+
+#####
+##### Checkpointing
+#####
+
+function prognostic_state(field::Field)
+    return (
+        data = prognostic_state(field.data),
+    )
```
tomchor:
```suggestion
    return (; data = prognostic_state(field.data))
```

ext/OceananigansNCDatasetsExt.jl
```diff
@@ -1578,4 +1578,38 @@ end
...
+    # Only checkpoint WindowedTimeAverage outputs (which have accumulated state)
+    wta_outputs = NamedTuple(Symbol(name) => prognostic_state(output)
+                             for (name, output) in pairs(writer.outputs)
+                             if output isa WindowedTimeAverage)
+
+    return (
+        schedule = prognostic_state(writer.schedule),
+        part = writer.part,
+        windowed_time_averages = isempty(wta_outputs) ? nothing : wta_outputs,
+    )
```
tomchor: Do we wanna enforce the formatting we usually prefer in Oceananigans? Namely:
```suggestion
    return (schedule = prognostic_state(writer.schedule),
            part = writer.part,
            windowed_time_averages = isempty(wta_outputs) ? nothing : wta_outputs)
```

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
```diff
@@ -393,3 +394,24 @@ for Type in (SplitExplicitFreeSurface,
...
+#####
+
+function prognostic_state(fs::SplitExplicitFreeSurface)
+    return (
+        η = prognostic_state(fs.η),
+        barotropic_velocities = prognostic_state(fs.barotropic_velocities),
+        filtered_state = prognostic_state(fs.filtered_state),
+        timestepper = prognostic_state(fs.timestepper),
+    )
+end
```
tomchor: Same comment here about formatting

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_timesteppers.jl
```diff
@@ -160,3 +162,33 @@ end
...
+    return (
+        ηᵐ   = prognostic_state(ts.ηᵐ),
+        ηᵐ⁻¹ = prognostic_state(ts.ηᵐ⁻¹),
+        ηᵐ⁻² = prognostic_state(ts.ηᵐ⁻²),
+        Uᵐ⁻¹ = prognostic_state(ts.Uᵐ⁻¹),
+        Uᵐ⁻² = prognostic_state(ts.Uᵐ⁻²),
+        Vᵐ⁻¹ = prognostic_state(ts.Vᵐ⁻¹),
+        Vᵐ⁻² = prognostic_state(ts.Vᵐ⁻²),
+    )
+end
```
tomchor: And here (and other places below as well)

src/OutputWriters/jld2_writer.jl
```diff
@@ -335,3 +337,36 @@ function Base.show(io::IO, ow::JLD2Writer)
...
+    if hasproperty(state, :windowed_time_averages) && !isnothing(state.windowed_time_averages)
+        for (name, wta_state) in pairs(state.windowed_time_averages)
+            if haskey(writer.outputs, name) && writer.outputs[name] isa WindowedTimeAverage
+                restore_prognostic_state!(writer.outputs[name], wta_state)
+            end
+        end
+    end
+
+    return writer
+end
```
tomchor: Is this code exactly the same for all `OutputWriter`s? If so, can we write it only once? In the future we might also have other supported formats (like zarr) and it'd be good to unify as much code as possible.

test/test_checkpointer.jl
```diff
@@ -1,258 +1,1854 @@
...
+    elseif schedule_type == :WallTimeInterval
+        @test restored_schedule.interval == original_schedule.interval
+    end
+
+    rm.(glob("$(prefix)_iteration*.jld2"), force=true)
+
+    return nothing
+end
+
+function test_windowed_time_average_checkpointing(arch, WriterType)
```
tomchor, line 1243: We're not testing that the values are correct here, right? Should we?

tomchor: nvmd, just saw you're testing that in the function below! Nice!

glwagner:
> There's been a lot of progress since this PR was open, so I'm wondering if this is still true. I don't oppose this, but I also don't oppose saving non-prognostic information when easy and throwing a warning to the user if we flag differences there. Mistakes happen and I think this sort of user-friendliness can avoid some headache. Again, only when it's to do so though. I definitely don't think `Checkpointer` needs to check every single aspect of a `Simulation` to see if everything matches up.

@tomchor can you specify what you mean by "this" in your comment?

Possibly unrelated, but I'm wondering if we think there would be value in being as strict as humanly possible about the accuracy of picking up from a checkpoint. Otherwise critical operational applications will have to implement their own testing system because they could not rely on the Oceananigans in-built system for it. If there is a need to relax strictness for the purpose of productivity, we might have a checkpointer "mode" (strict=true, false)

test/test_checkpointer.jl
tomchor, line 1: Great work with testing! Pretty comprehensive. My only suggestion here is maybe instead of write two functions like `test_XYZ_hydrostatic()` and `test_XYZ_nonhodrostatic()`, should we write one function called `test_XYZ()` and pass the model as a argument? There might be some `if/else` statements in there, but I feel like this would be an easier test suite to keep track of and maintain in the long run, no?

tomchor:
> > There's been a lot of progress since this PR was open, so I'm wondering if this is still true. I don't oppose this, but I also don't oppose saving non-prognostic information when easy and throwing a warning to the user if we flag differences there. Mistakes happen and I think this sort of user-friendliness can avoid some headache. Again, only when it's to do so though. I definitely don't think `Checkpointer` needs to check every single aspect of a `Simulation` to see if everything matches up.
> 
> @tomchor can you specify what you mean by "this" in your comment?
> 
> Possibly unrelated, but I'm wondering if we think there would be value in being as strict as humanly possible about the accuracy of picking up from a checkpoint. Otherwise critical operational applications will have to implement their own testing system because they could not rely on the Oceananigans in-built system for it. If there is a need to relax strictness for the purpose of productivity, we might have a checkpointer "mode" (strict=true, false)

Sorry, by `this` I meant the general idea in the sentences I quoted (maybe I didn't quote the best ones!). Mainly that `Checkpointer` should assume that the simulation setup is the same and _only_ keep track of prognostic state information. So, to repeat @ali-ramadhan's example: 

> For example, take the SpecifiedTimes schedule. It has two properties times and previous_actuation. Since previous_actuation changes as the simulation runs, only previous_actuation needs to be checkpointed.
>
> This leads to the possibility of the user changing times then picking up previous_actuation which can lead to undefined behavior. I think this is fine, because the checkpointer only works assuming you set up the same simulation as the one that was checkpointed.
>
> Checkpointing both times and previous_actuation allows us to check that times is the same when restoring. But I don't think this is the checkpointer's responsibility.

I'm not opposed to this, but I wonder if it's useful to save `times` and throw an error if the checkpointer realizes that this property changed (which usually is due to human error). This is ofc just one example of something simple that could be checked.

ext/OceananigansNCDatasetsExt.jl
glwagner: I prefer keeping it consistent (and changing it globally if we want) but it is not the most critical thing in the world

src/Fields/field.jl
glwagner: Also I think for `NamedTuple`, let's start with `;` for clarity, eg `(; data = prognostic_state(field.data))`.

Curious --- if there is a prognostic state associated with field boundary conditions, do we have to include? I presume we probalby don't have a test for it. There are lots of edge cases too so I am ok not supporting this in a first pass. This infrastructure is clean and we can keep building it in future PRs.

src/Fields/field.jl
```diff
@@ -860,3 +860,18 @@ end
...
+end
+
+function restore_prognostic_state!(field::Field, state)
+    restore_prognostic_state!(field.data, state.data)
+    return field
+end
```
glwagner:
```suggestion
end

```

src/OutputWriters/jld2_writer.jl
glwagner: if there are small differences, we can open an issue to smooth out those wrinkles between the two --- effectively creating a "convention for output writers" with the benefits mentioned by @tomchor

navidcy:
> The PR is getting quite large and distributed CI is not working, so I'll test checkpointing support for distributed models in a subsequent PR.

There was always a distributed test failing at https://github.com/CliMA/Oceananigans.jl/pull/4005 and never understood why...

glwagner:
I will review again once @tomchor comments have been addressed but overall, I think it is excellent and very close. Most of my thoughts were 1:1 overlap with @tomchor's comments.

docs/src/simulations/checkpointing.md
ali-ramadhan: Just realized that `set!(simulation, true)` looks vague. As in if you saw that line in a script randomly you wouldn't know what it does. So I might add kwargs so it instead looks like

```julia
set!(simulation; filepath="spinup.jld2")
set!(simulation; use_latest_checkpoint=true)
set!(simulation; iteration=12345)
```

docs/src/simulations/checkpointing.md
ali-ramadhan: Definitely a good use case. I was thinking we can do something more general and checkpoint after a simulation stops running (whether it is done or not)?

src/Fields/field.jl
ali-ramadhan: Yeah I think the suggestion is just missing a `;` so I modified it @tomchor and committed.

@glwagner Yeah right now field boundary conditions are not checkpointed. I suppose a use case would be a boundary condition whose values come from an array that changes with time depending on the model's prognostic state. Not that I've seen a setup like this.

We can definitely build this feature in the future. It shouldn't be hard to add.

ext/OceananigansNCDatasetsExt.jl
ali-ramadhan: Yeah I can be a bit sloppy with coding style. I'll make a pass and use a coding style that is consistent with the rest of Oceananigans.jl.

src/OutputWriters/jld2_writer.jl
ali-ramadhan: Good catch. I should be able to combine `prognostic_state` and `restore_prognostic_state!` for multiple writers.

test/test_checkpointer.jl
ali-ramadhan: I have combined some but definitely seeing room to combine more. I agree it'll make it easier to maintain in the future.

src/Fields/field.jl
glwagner: It would only be required for a very particular case in which, eg, an equation is being integrated in time and the result is stored in a boundary condition array. in that case, data inside the boundary condition has a prognostic state.

docs/src/simulations/checkpointing.md
tomchor: Sounds good to me!

docs/src/simulations/checkpointing.md
tomchor: Agreed!

src/Fields/field.jl
glwagner: the vast majority of the time, boundary condition data is diagnostic and computed instantaneously from the prognostic state (for example, bulk-formula-forced ocean simulations in ClimaOcean)

we cannot know when an array is prognostic or diagnostic. so we can conservatively always checkpoint array, or not support it

ali-ramadhan:
Thanks for the reviews! I made a TODO list to tackle and I'll request another round of reviews. Please feel free to add to it.

- [x] Add kwargs to `set!(simulation, ...)`.
- [x] Checkpoint after a simulation stops running (whether it is done or not).
- [x] Use a coding style that is consistent with the rest of Oceananigans.jl.
- [x] Combine `prognostic_state` and `restore_prognostic_state!` for multiple writers.
- [x] Combine test functions where possible, e.g. `test_XYZ_hydrostatic()` and `test_XYZ_nonhodrostatic()` -> `test_XYZ()`.
- [ ] Maybe tackle https://github.com/CliMA/Oceananigans.jl/issues/3206 if it's easy (to make @simone-silvestri happy).

> Possibly unrelated, but I'm wondering if we think there would be value in being as strict as humanly possible about the accuracy of picking up from a checkpoint.

@glwagner Are you suggesting that we checkpoint the diagnostic state so it can be used to make sure the simulation was restored properly? Or are you suggesting checkpointing things like `TimeInterval.interval` (not diagnostic nor prognostic?) to make sure it is the same when restoring from checkpoint?

> There was always a distributed test failing at #4005 and never understood why...

@navidcy Right now it seems like distributed CI even fails to initialize here. I tried looking through Buildkite runs from #4005 and couldn't find a failure with an actual error. Maybe I wouldn't think too much about it until we can run distributed checkpointing tests with/after this PR.

docs/src/simulations/checkpointing.md
glwagner: only qualm is that "false" means nothing.

An altnerative is a symbol, eg 

```julia
set!(simulation, :latest_checkpoint)
```

i do like the kwarg for the other options.

navidcy:
@ali-ramadhan, #5067 should have fixed the distributed CI.

navidcy:
We need to bump minor to 104: https://github.com/CliMA/Oceananigans.jl/blob/141696a97ab59d83f95b6b166aa26a37434d3d2f/Project.toml#L4

ali-ramadhan:
@glwagner @navidcy If we're not in a rush to tag and release v0.104.0 I'll try to get this merged in the next day or two so that we can make one breaking release. I see a couple of other breaking changes lined up.

navidcy:
That'd be sweet. It'd be good if we can include this in the breaking 0.104!

glwagner:
sounds great! amazing work on this!

src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
```diff
@@ -150,3 +152,28 @@ function step_free_surface!(free_surface::ImplicitFreeSurface, model, timesteppe
...
+
+function local_compute_integrated_volume_flux!(∫ᶻQ, velocities, arch)
+
+    foreach(mask_immersed_field!, velocities)
+
+    # Compute barotropic volume flux. Blocking.
+    compute_vertically_integrated_volume_flux!(∫ᶻQ, velocities)
+
+    return nothing
+end
```
simone-silvestri:
```suggestion
```
I think this is not needed anymore

src/Models/HydrostaticFreeSurfaceModels/prescribed_hydrostatic_velocity_fields.jl
```diff
@@ -132,3 +134,10 @@ end
 
 update_state!(model::OnlyParticleTrackingModel, callbacks) =
     [callback(model) for callback in callbacks if callback.callsite isa UpdateStateCallsite]
+
+#####
+##### Checkpointing
+#####
+
+prognostic_state(::PrescribedVelocityFields) = nothing
```
simone-silvestri, line 142: should we save down the prescribed velocities and restore them if they are fields?

src/OutputWriters/checkpointer.jl
```diff
@@ -56,43 +50,17 @@ Keyword arguments
...
-    end
-
-    for p in properties
-        p isa Symbol || error("Property $p to be checkpointed must be a Symbol.")
-        p ∉ propertynames(model) && error("Cannot checkpoint $p, it is not a model property!")
-
-        if (p ∉ required_properties) && has_reference(Function, getproperty(model, p))
-            @warn "model.$p contains a function somewhere in its hierarchy and will not be checkpointed."
-            filter!(e -> e != p, properties)
-        end
```
simone-silvestri, line 90: very nice :)

src/Models/HydrostaticFreeSurfaceModels/prescribed_hydrostatic_velocity_fields.jl
simone-silvestri: scratch it, prescribed velocities are not prognostic

src/TimeSteppers/split_hydrostatic_runge_kutta_3.jl
```diff
@@ -166,3 +168,17 @@ function cache_previous_fields!(model)
...
+        Gⁿ = prognostic_state(timestepper.Gⁿ),
+        Ψ⁻ = prognostic_state(timestepper.Ψ⁻),
+    )
+end
+
+function restore_prognostic_state!(timestepper::SplitRungeKutta3TimeStepper, state)
+    restore_prognostic_state!(timestepper.Gⁿ, state.Gⁿ)
+    restore_prognostic_state!(timestepper.Ψ⁻, state.Ψ⁻)
+    return timestepper
+end
```
simone-silvestri, line 185:
```suggestion
prognostic_state(timestepper::SplitRungeKutta3TimeStepper) = nothing
restore_prognostic_state!(timestepper::SplitRungeKutta3TimeStepper, state) = nothing
```
I think this can be avoided. These quantities are diagnostic because RK3 is self-starting

src/TimeSteppers/TimeSteppers.jl
```diff
@@ -80,4 +80,18 @@ function first_time_step!(model::AbstractModel{<:QuasiAdamsBashforth2TimeStepper
...
+        Gⁿ = prognostic_state(timestepper.Gⁿ),
+        G⁻ = prognostic_state(timestepper.G⁻),
+    )
+end
+
+function restore_prognostic_state!(timestepper::AbstractTimeStepper, state)
+    restore_prognostic_state!(timestepper.Gⁿ, state.Gⁿ)
+    restore_prognostic_state!(timestepper.G⁻, state.G⁻)
+    return timestepper
+end
```
simone-silvestri, line 96:
```suggestion
```
Not all timesteppers have a `Gⁿ` and `G⁻`. I think this should be specialized for the different timesteppers

src/TimeSteppers/split_hydrostatic_runge_kutta_3.jl
simone-silvestri: same thing for the `RungeKutta3` timestepper.

src/TimeSteppers/TimeSteppers.jl
simone-silvestri: also `Gⁿ` is recomputed so we do not need to store it (it saves space to not store it :) )

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
```diff
@@ -393,3 +394,24 @@ for Type in (SplitExplicitFreeSurface,
...
+
+#####
+##### Checkpointing
+#####
+
+function prognostic_state(fs::SplitExplicitFreeSurface)
+    return (
+        η = prognostic_state(fs.η),
+        barotropic_velocities = prognostic_state(fs.barotropic_velocities),
+        filtered_state = prognostic_state(fs.filtered_state),
```
simone-silvestri:
```suggestion
```
This is also not required, the filtered state is computed while timestepping

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
```diff
@@ -393,3 +394,24 @@ for Type in (SplitExplicitFreeSurface,
...
+        timestepper = prognostic_state(fs.timestepper),
+    )
+end
+
+function restore_prognostic_state!(fs::SplitExplicitFreeSurface, state)
+    restore_prognostic_state!(fs.η, state.η)
+    restore_prognostic_state!(fs.barotropic_velocities, state.barotropic_velocities)
+    restore_prognostic_state!(fs.filtered_state, state.filtered_state)
```
simone-silvestri:
```suggestion
```

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_timesteppers.jl
```diff
@@ -160,3 +162,33 @@ end
...
+function restore_prognostic_state!(ts::AdamsBashforth3Scheme, state)
+    restore_prognostic_state!(ts.ηᵐ,   state.ηᵐ)
+    restore_prognostic_state!(ts.ηᵐ⁻¹, state.ηᵐ⁻¹)
+    restore_prognostic_state!(ts.ηᵐ⁻², state.ηᵐ⁻²)
+    restore_prognostic_state!(ts.Uᵐ⁻¹, state.Uᵐ⁻¹)
+    restore_prognostic_state!(ts.Uᵐ⁻², state.Uᵐ⁻²)
+    restore_prognostic_state!(ts.Vᵐ⁻¹, state.Vᵐ⁻¹)
+    restore_prognostic_state!(ts.Vᵐ⁻², state.Vᵐ⁻²)
+    return ts
+end
```
simone-silvestri, line 192: I think also this can be avoided. These variables are defined only within one baroclinic time-step and unless we checkpoint in-between baroptropic substeps we do not need to checkpoint these fields

src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
ali-ramadhan: Nice catch!

src/TimeSteppers/split_hydrostatic_runge_kutta_3.jl
ali-ramadhan: Ah hmm I might be wrong but I thought with step-site callbacks it's technically possible to checkpoint in the middle of an RK3 step (like after stage 1 or stage 2) in which case we would need to checkpoint the state of the RK3 time stepper?

But perhaps this is a bizarre edge case that is best left unsupported.

src/TimeSteppers/TimeSteppers.jl
ali-ramadhan: Yeah true I implemented this for `AbstractTimestepper` since it was the exact same code for QAB2 and RK3. Happy to switch it to back if it makes the code clearer.

Can also just checkpoint `G⁻`!

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
ali-ramadhan: This makes sense to me but when I removed the `filtered_state` from checkpointing the model equality tests failed so I'm putting it back in for now.

ali-ramadhan:
Thanks for merging in `main` and resolving conflicts @giordano! I just cancelled the build since I'm going to push some test fixes soon and didn't want to add to the Buildkite CI load.

I can update all instances of `prognostic_state` and `restore_prognostic_state!` to use `Oceananigans.prognostic_state` and `Oceananigans.restore_prognostic_state!`.

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
simone-silvestri: that is weird, something to investigate then. Probably because the filtered_state is not zeroed out until we actually perform the barotropic stepping, so there is scratch data that remains hanging until the filtered state is reset. I can correct it when we merge this.

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_timesteppers.jl
ali-ramadhan: We now actually test both `ForwardBackwardScheme` and `AdamsBashforth3Scheme` but when removing checkpointing for `AdamsBashforth3Scheme` the tests failed so I'll keep this in for now.

ali-ramadhan:
@simone-silvestri Thanks for pointing out that some stuff doesn't need to be checkpointed. I may incorrectly been saying that they do, but I think I'm not testing very well.

I think checkpointing tests should be restoring from a checkpoint then taking a further number of time steps before checking for model equality. This is because there are some self-starting time steppers that do not need to be checkpointed because once restored they will compute the correct tendencies. But checking for model equality right after picking up will fail because the tendencies will be zero (which is fine, they'll be computed correctly as part of the next time step).

I'll update the tests then test again if we can get away with not checkpointing `AdamsBashforth3Scheme` and `filtered_state`.

ali-ramadhan:
Improving the tests now shows that checkpointing for stateful turbulence closures is actually incorrect. Looking into how to fix this now.

We might be missing some things that need to be checkpointed, but I think the main issue is that `compute_diffusivities!` is called during `initialize!(sim)` after a checkpoint restore, which causes time-dependent updates (e.g. implicit TKE stepping, time averaging) to be applied twice.

glwagner:
> Improving the tests now shows that checkpointing for stateful turbulence closures is actually incorrect. Looking into how to fix this now.
> 
> We might be missing some things that need to be checkpointed, but I think the main issue is that `compute_diffusivities!` is called during `initialize!(sim)` after a checkpoint restore, which causes time-dependent updates (e.g. implicit TKE stepping, time averaging) to be applied twice.

this includes both DynamicSmagorinsky and CATKEVerticalDiffusivity, right?

brainstorming... we could add a kwarg to `update_state!` that's something like `update_state!(model; time_step=false)`.

This flag would then have to be propagated into `compute_diffusivities!`, and we would use it to determine whether or not to update stateful quantities as would be needed after a time-step is complete, versus merely updating auxiliary quantities in place.

I could pursue this in another PR or in this PR if you think appropriate -- let me know.

ali-ramadhan:
Yeah I'm seeing failures for `RiBasedVerticalDiffusivity`, `DynamicSmagorinsky`, `CATKEVerticalDiffusivity`, and `TKEDissipationVerticalDiffusivity`.

I think I've managed to fix `RiBasedVerticalDiffusivity` based on checking whether the time step Δt is zero. I'll see if I can fix the others similarly.

A `time_step=false` kwarg sounds cleaner. I'm happy to implement this. I'll continue trying to fix checkpointing for each closure, then that should provide a clearer picture of how to implement `time_step=false`.

simone-silvestri:
I think we can strive to remove prognostic variables update in `update_state!`. With the timestepping PR, the hydrostatic tendency computations are separated between momentum and tracer so we do not need to time step TKE inside update state because the new velocities can be readily available to compute TKE (I am ready to remove that update as soon as the PR is merged). Maybe something like this can be done also where needed for the Smagorinsky model.

glwagner:
> I think we can strive to remove prognostic variables update in `update_state!`. With the timestepping PR, the hydrostatic tendency computations are separated between momentum and tracer so we do not need to time step TKE inside update state because the new velocities can be readily available to compute TKE (I am ready to remove that update as soon as the PR is merged). Maybe something like this can be done also where needed for the Smagorinsky model.

I think this is a longer effort.

simone-silvestri:
I don't think it would be too difficult, I can try to draw up a plan on how to do it. Do we need to do it for this PR? Or can we merge this PR with the caveat that turbulence closures are not checkpointed correctly (which basically is the same that we have now) and improve on checkpointing in subsequent PRs?

ali-ramadhan:
I've made some progress but it has been difficult. `RiBasedVerticalDiffusivity` works now.

`DynamicSmagorinsky` works but only with QAB2, not RK3. Might need deeper fixes to work with RK3.

I'll see what I can do for CATKE and TKE. But otherwise I've addressed reviewer comments so I might push what I have, mark some tests as broken, then re-request review.

simone-silvestri:
sounds like a good plan, we can open an issue to track the development and document where we have to go next. 
Should we also change the formatting like @tomchor suggested?

ali-ramadhan:
Got `DynamicSmagorinsky` to work now! I'll make a post detailing some changes I made to make turbulence closure checkpointing work.

>  Should we also change the formatting like @tomchor suggested?

Already done! Just haven't pushed it yet to conserve CI resources until I need them.

ali-ramadhan:
Ok so I think checkpointing now works for all turbulence closures!

In general, we shouldn't be calling `initialize!(model)` when restoring from a checkpoint because this could overwrite/corrupt closure fields. And we should use `update_state!(model; compute_tendencies=false)` when restoring from a checkpoint.

I also had to add a very small absolute tolerance to model equality tests when testing checkpointing turbulence closures, usually for fields like e and ε that may accumulate floating point differences due to differing order of operations (reductions?). But it's tiny, like `atol = 1e-20`.

## Smagorinsky closures

`Smagorinsky` and `SmagorinskyLilly` don't need checkpointing.

In `DirectionallyAveragedDynamicSmagorinsky` we technically may not need to checkpoint `𝒥ᴸᴹ` and `𝒥ᴹᴹ` since they're recomputed from `LM` and `MM` but I did it to ensure exact reproducibility.

`LagrangianAveragedDynamicSmagorinsky` needed some changes to be checkpointed because it maintains time history for Lagrangian trajectory averaging. I had to also checkpoint `previous_compute_time` along with `𝒥ᴸᴹ⁻` and `𝒥ᴹᴹ⁻`.

When using a multi-stage timestepper like RK3, I think it should only compute coefficients at the final stage. Lagrangian averaging was being applied 3 times with small fractional Δt values instead of once with the full Δt. Not doing so leaves the closure in an inconsistent state so when you pickup you can't recreate the same sequence of intermediate states. I think the main issue is that `t⁻` is updated at every substep. The checkpoint captures `t⁻` from `stage=1`, but the next computation in the continuous run happens at `stage=2`.

But please let me know if this makes sense for `LagrangianAveragedDynamicSmagorinsky` + RK3. If not, I can revert the changes and we can fix this properly in another PR. This is not an issue with QAB2. cc @tomchor @glwagner @simone-silvestri 

## `RiBasedVerticalDiffusivity`

This was relatively easy to checkpoint. Gotta checkpoint some closure fields but otherwise just needed to keep track of the `previous_compute_time`.

## CATKE

I had to add some checks like making sure we only call `time_step_catke_equation!` and `compute_average_surface_buoyancy_flux!` on new iterations (Δt > 0) and making sure they work with split RK3. I also added a `_skip_next_compute` property to `CATKEDiffusivityFields` because we need to 

## `TKEDissipationVerticalDiffusivity`

Also similar to CATKE, I had to introduce a `previous_compute_time` and `_skip_next_compute`. Should also work once TKE-dissipation supports RK3 (#5127).

simone-silvestri:
If we do not call initialize, then we need to checkpoint also the grid because in case of a variable vertical coordinate this should be set in `restore_prognostic_state!`

ali-ramadhan:
> If we do not call initialize, then we need to checkpoint also the grid because in case of a variable vertical coordinate this should be set in `restore_prognostic_state!`

Ah right for something like `ZStarCoordinate`? I do not test checkpointing for this, so let me add it.

Let me know if you can think of other cases I should be testing with checkpointing!

ali-ramadhan:
@giordano Looks like this PR reduces the number of method ambiguities a little bit.

I assume I should be updating `@test number_of_ambiguities <= 361` to `@test number_of_ambiguities <= 355` so that we are always reducing the number and never increasing it?

giordano:
Yay! Yes, updating the count would be great, although that's going to conflict with #4811 which also reduces the ambiguities a bit.

glwagner:
> Ok so I think checkpointing now works for all turbulence closures!
> 
> In general, we shouldn't be calling `initialize!(model)` when restoring from a checkpoint because this could overwrite/corrupt closure fields. And we should use `update_state!(model; compute_tendencies=false)` when restoring from a checkpoint.
> 
> I also had to add a very small absolute tolerance to model equality tests when testing checkpointing turbulence closures, usually for fields like e and ε that may accumulate floating point differences due to differing order of operations (reductions?). But it's tiny, like `atol = 1e-20`.
> 
> ## Smagorinsky closures
> `Smagorinsky` and `SmagorinskyLilly` don't need checkpointing.
> 
> In `DirectionallyAveragedDynamicSmagorinsky` we technically may not need to checkpoint `𝒥ᴸᴹ` and `𝒥ᴹᴹ` since they're recomputed from `LM` and `MM` but I did it to ensure exact reproducibility.
> 
> `LagrangianAveragedDynamicSmagorinsky` needed some changes to be checkpointed because it maintains time history for Lagrangian trajectory averaging. I had to also checkpoint `previous_compute_time` along with `𝒥ᴸᴹ⁻` and `𝒥ᴹᴹ⁻`.
> 
> When using a multi-stage timestepper like RK3, I think it should only compute coefficients at the final stage. Lagrangian averaging was being applied 3 times with small fractional Δt values instead of once with the full Δt. Not doing so leaves the closure in an inconsistent state so when you pickup you can't recreate the same sequence of intermediate states. I think the main issue is that `t⁻` is updated at every substep. The checkpoint captures `t⁻` from `stage=1`, but the next computation in the continuous run happens at `stage=2`.
> 
> But please let me know if this makes sense for `LagrangianAveragedDynamicSmagorinsky` + RK3. If not, I can revert the changes and we can fix this properly in another PR. This is not an issue with QAB2. cc @tomchor @glwagner @simone-silvestri
> 
> ## `RiBasedVerticalDiffusivity`
> This was relatively easy to checkpoint. Gotta checkpoint some closure fields but otherwise just needed to keep track of the `previous_compute_time`.
> 
> ## CATKE
> I had to add some checks like making sure we only call `time_step_catke_equation!` and `compute_average_surface_buoyancy_flux!` on new iterations (Δt > 0) and making sure they work with split RK3. I also added a `_skip_next_compute` property to `CATKEDiffusivityFields` because we need to
> 
> ## `TKEDissipationVerticalDiffusivity`
> Also similar to CATKE, I had to introduce a `previous_compute_time` and `_skip_next_compute`. Should also work once TKE-dissipation supports RK3 (#5127).

I think this is great!

We could consider a refactor in the future --- for example adding a new function that's something like `time_step_closure_fields!(closure, model)`. With such an interface, we can distinguish between true "auxiliary state updates" (via update_state!) and prognostic state updates.

ali-ramadhan:
@simone-silvestri `ZStarCoordinate` checkpointing works now (and is tested)! I didn't have to checkpoint grids, just `model.vertical_coordinate` and also pass the `grid` to `prognostic_state` and `restore_prognostic_state!`. But if we want to checkpoint grids in the future it'll be easy to do.

I think this PR is ready to be merged now so just re-requesting review. I'd also like to wait to hear from @tomchor about whether the changes I made to `DynamicSmagorinsky` are okay.

docs/src/simulations/checkpointing.md
```diff
@@ -1,12 +1,12 @@
...
-can be restored at any time. This is useful if you'd like to periodically checkpoint when running
+A [`Checkpointer`](@ref) can be used to serialize the simulation state to a file from which
+the simulation can be restored at any time. This is useful if you'd like to periodically checkpoint when running
 long simulations in case of crashes or hitting cluster time limits, but also if you'd like to restore
 from a checkpoint and try out multiple scenarios.
 
-For example, to periodically checkpoint the model state to disk every 1,000,000 seconds of simulation
-time to files of the form `model_checkpoint_iteration12500.jld2` where `12500` is the iteration
+For example, to periodically checkpoint to disk every 1,000,000 seconds of simulation
+time to files of the form `checkpoint_iteration12500.jld2` where `12500` is the iteration
```
tomchor: This sentence is a bit confusing. Maybe the numbers are doing more harm than good here

docs/src/simulations/checkpointing.md
```diff
@@ -22,6 +22,13 @@ simulation = Simulation(model, Δt=1, stop_iteration=8)
 simulation.output_writers[:checkpointer] = Checkpointer(model, schedule=IterationInterval(5), prefix="model_checkpoint")
 ```
 
+For long simulations, use `cleanup=true` to automatically delete old checkpoint files
```
tomchor:
```suggestion
Use `cleanup=true` to automatically delete old checkpoint files
```

docs/src/simulations/checkpointing.md
tomchor: Was this implemented? Should I resolve this comment?

tomchor: Ah, I see it has. I'll resolve this.

src/Models/LagrangianParticleTracking/LagrangianParticleTracking.jl
```diff
@@ -157,4 +158,19 @@ function fetch_output(lagrangian_particles::LagrangianParticles, model)
     return NamedTuple{names}([getproperty(particle_properties, name) for name in names])
 end
 
+# Checkpointing
+
+function prognostic_state(lagrangian_particles::LagrangianParticles)
+    return (
+        properties = prognostic_state(lagrangian_particles.properties),
+    )
```
tomchor:
```suggestion
    return (; properties = prognostic_state(lagrangian_particles.properties))
```

tomchor:
> `Smagorinsky` and `SmagorinskyLilly` don't need checkpointing.
> 
> In `DirectionallyAveragedDynamicSmagorinsky` we technically may not need to checkpoint `𝒥ᴸᴹ` and `𝒥ᴹᴹ` since they're recomputed from `LM` and `MM` but I did it to ensure exact reproducibility.
> 
> `LagrangianAveragedDynamicSmagorinsky` needed some changes to be checkpointed because it maintains time history for Lagrangian trajectory averaging. I had to also checkpoint `previous_compute_time` along with `𝒥ᴸᴹ⁻` and `𝒥ᴹᴹ⁻`.
> 
> When using a multi-stage timestepper like RK3, I think it should only compute coefficients at the final stage. Lagrangian averaging was being applied 3 times with small fractional Δt values instead of once with the full Δt. Not doing so leaves the closure in an inconsistent state so when you pickup you can't recreate the same sequence of intermediate states. I think the main issue is that `t⁻` is updated at every substep. The checkpoint captures `t⁻` from `stage=1`, but the next computation in the continuous run happens at `stage=2`.
> 
> But please let me know if this makes sense for `LagrangianAveragedDynamicSmagorinsky` + RK3. If not, I can revert the changes and we can fix this properly in another PR. This is not an issue with QAB2. cc @tomchor @glwagner @simone-silvestri

I tried to find some info on this online but couldn't. What I can say is that the models (I'm aware of) will usually update the closure every `n` time steps (usually 5), and when comes time to update the closure, they compute the averaging at every substep. Given that, and given the fact that substeps are very short (and perhaps not a "real" solution of the model), I think your changes are okay. Perhaps even an improvement! Great work!

docs/src/simulations/checkpointing.md
ali-ramadhan: Agreed! Changed it to this and will push soon:

```
Here's an example where we checkpoint every 5 iterations to files of the form
`model_checkpoint_iteration5.jld2` (where the iteration number is automatically included in
the filename). This is far more often than appropriate for typical applications: we only do
it here for illustration purposes.
```

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
ali-ramadhan: I'm trying this again now that we've improved the tests.

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_timesteppers.jl
ali-ramadhan: I'm trying this again now that we've improved the tests.

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
ali-ramadhan: @simone-silvestri Tests passing without checkpointing the filtered state so I'll commit this!

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_timesteppers.jl
ali-ramadhan: @simone-silvestri Tests still failed when I removed checkpointing for `AdamsBashforth3Scheme` so I'll keep this one in.

docs/src/simulations/checkpointing.md
ali-ramadhan: This is what I went with:

``````
The [`set!`](@ref) function can also be used to restore from a checkpoint without immediately
running the simulation:

```julia
set!(simulation; checkpoint="path/to/file.jld2")  # restore from specific file
set!(simulation; checkpoint=:latest)              # restore from latest checkpoint (requires Checkpointer)
set!(simulation; iteration=12345)                 # restore from specific iteration (requires Checkpointer)
```
``````

src/OutputWriters/jld2_writer.jl
ali-ramadhan: This has been done!

ext/OceananigansNCDatasetsExt.jl
ali-ramadhan: Went back and fixed these!

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_timesteppers.jl
ali-ramadhan: Done!

src/Models/HydrostaticFreeSurfaceModels/SplitExplicitFreeSurfaces/split_explicit_free_surface.jl
ali-ramadhan: Done!

PR #4905, opened 2025-11-03
Simplify and add tests for `prettysummary(::Function)`

giordano:
Ref: https://github.com/CliMA/Oceananigans.jl/pull/4840#discussion_r2487853689.  This also helps towards #4836.

### DIFF
--- a/src/Utils/prettysummary.jl
+++ b/src/Utils/prettysummary.jl
@@ -5,21 +5,13 @@ prettysummary(x, args...) = summary(x)
 
 function prettysummary(f::Function, showmethods=true)
     ft = typeof(f)
-    if VERSION >= v"1.12.0"
-        name = ft.name.name
-    else
-        mt = ft.name.mt
-        name = mt.name
-    end
+    name = string(Base.nameof(f))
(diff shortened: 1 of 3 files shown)


src/Utils/prettysummary.jl
```diff
@@ -5,21 +5,13 @@ prettysummary(x, args...) = summary(x)
...
     m = n==1 ? "method" : "methods"
-    sname = string(name)
-    isself = isdefined(ft.name.module, name) && ft == typeof(getfield(ft.name.module, name))
-    ns = (isself || '#' in sname) ? sname : string("(::", ft, ")")
     if showmethods
-        return string(ns, " (", "generic function", " with $n $m)")
+        return string(name, " (", "generic function", " with $n $m)")
     else
-        return string(ns)
+        return string(name)
```
giordano, line 14: The output is now slightly different than before for cases where the function is local, but I think the code is a lot simpler now, and the output is actually nicer: for the local function `f4905` in the tests we now get just `f4905` instead of something like `(::f4905)`.

PR #4909, opened 2025-11-05
fix tapering function in diffusive GM

xkykai:
@simone-silvestri @glwagner


simone-silvestri:
To avoid NaNs we probably need something like
```julia
if max_slope == 0 && Sx^2 + Sy^2 == 0 
   return 0
```
(not in pseudocode :smile:)

simone-silvestri:
Or probably not, I don't think we'll ever use `max_slope == 0`

simone-silvestri (approved):
I would test that this does not NaN immediately in a case with an infinite slope before merging it.

xkykai:
Right now it fails on the minimal Abernathey channel case

glwagner:
> Right now it fails on the minimal Abernathey channel case

so we should not merge?

xkykai:
> > Right now it fails on the minimal Abernathey channel case
> 
> so we should not merge?

That was before I put in
```julia
ifelse(bz <= 0, 0, min(one(grid), slope_limiter.max_slope^2 / (Sx^2 + Sy^2)))
```

Just verified this fix against the advective formulation in a Abernathey channel-like setup.

https://github.com/user-attachments/assets/72f1ac4a-8192-469b-9414-5e4e3bee8774

Looks okay and might've even be slightly better than before. Diffusive seems much closer than advective than in the past. Seems like we can merge then!

glwagner:
> > > Right now it fails on the minimal Abernathey channel case
> > 
> > 
> > so we should not merge?
> 
> That was before I put in
> 
> ```julia
> ifelse(bz <= 0, 0, min(one(grid), slope_limiter.max_slope^2 / (Sx^2 + Sy^2)))
> ```
> 
> Just verified this fix against the advective formulation in a Abernathey channel-like setup.
> 
>  check_gm_fix_abernathey_channel_2D.mp4 
> Looks okay and might've even be slightly better than before. Diffusive seems much closer than advective than in the past. Seems like we can merge then!

nice!

is there any way to encode the possible failure in a test? no worries if its too hard

xkykai:
it failed previously when it hit a scenario where NaNs begin to emerge due to interactions between `Inf`s and `NaN`s. Those can emerge for when `bz <= 0`. Do you mean setting up a test case with `bz = 0` and then ensuring it doesn't throw `NaN`s ?

glwagner:
> it failed previously when it hit a scenario where NaNs begin to emerge due to interactions between `Inf`s and `NaN`s. Those can emerge for when `bz <= 0`. Do you mean setting up a test case with `bz = 0` and then ensuring it doesn't throw `NaN`s ?

if you think that is a reasonable test, feel free to add. otherwise merge away

xkykai:
@simone-silvestri @glwagner do you know why this test errors out specifically only on GPUs with `Float32` during `time_step!`?
https://github.com/CliMA/Oceananigans.jl/blob/df4ea86920df85429b7c62e584a2dbe5e83629c3/test/test_gm_infinite_slope.jl

Something about 
```julia
Reason: unsupported dynamic function invocation (call to +)
```
during `∇_dot_qᶜ` in `hydrostatic_free_surface_tracer_tendency`. But what I did seemed pretty innocuous to me, just timestepping a model.

src/TurbulenceClosures/turbulence_closure_implementations/isopycnal_skew_symmetric_diffusivity.jl
```diff
@@ -203,13 +203,10 @@ end
...
-    slope_x = - bx / bz
-    slope_y = - by / bz
+    Sx = - bx / bz
+    Sy = - by / bz
 
-    # in case of a stable buoyancy gradient (bz > 0), the slope is set to zero
-    slope² = ifelse(bz <= 0, zero(grid), slope_x^2 + slope_y^2)
-
-    return min(one(grid), slope_limiter.max_slope^2 / slope²)
+    return ifelse(bz <= 0, 0, min(one(grid), slope_limiter.max_slope^2 / (Sx^2 + Sy^2)))
```
simone-silvestri:
```suggestion
    return ifelse(bz <= 0, zero(grid), min(one(grid), slope_limiter.max_slope^2 / (Sx^2 + Sy^2)))
```
I think this is the reason for the failure. The return was either a float or an Int, which cannot be done on GPUs (you need to have a type=stable value returned)

xkykai:
is this ready to go once all checks pass?

PR #4910, opened 2025-11-05
(0.101.0) Export public Smagorinsky constructors

glwagner:
This PR changes exports for Smagorinsky utilities. Previously we exported the "Coefficient" objects, but the public interface is supposed to be SmagorinskyLilly and DynamicSmagorinsky.

cc @tomchor @giordano

### DIFF
--- a/src/Advection/Advection.jl
+++ b/src/Advection/Advection.jl
@@ -64,6 +64,8 @@ const advection_buffers = [1, 2, 3, 4, 5, 6]
 @inline required_halo_size_y(::AbstractAdvectionScheme{B}) where B = B
 @inline required_halo_size_z(::AbstractAdvectionScheme{B}) where B = B
 
+struct DescreasingOrderAdvectionScheme end
+
 include("centered_advective_fluxes.jl")
 include("upwind_biased_advective_fluxes.jl")
 
--- a/src/Advection/centered_reconstruction.jl
+++ b/src/Advection/centered_reconstruction.jl
@@ -8,14 +8,19 @@ struct Centered{N, FT, CA} <: AbstractCenteredAdvectionScheme{N, FT}
     Centered{N, FT}(buffer_scheme::CA) where {N, FT, CA} = new{N, FT, CA}(buffer_scheme)
 end
 
-function Centered(FT::DataType=Oceananigans.defaults.FloatType; order = 2)
+function Centered(FT::DataType=Oceananigans.defaults.FloatType; 
+                  order = 2,
+                  buffer_scheme = DescreasingOrderAdvectionScheme())
+
     mod(order, 2) != 0 && throw(ArgumentError("Centered reconstruction scheme is defined only for even orders"))
 
     N  = Int(order ÷ 2)
-    if N > 1
-        buffer_scheme = Centered(FT; order=order-2)
-    else
-        buffer_scheme = nothing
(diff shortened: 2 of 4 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.100.7"
+version = "0.101.0"
```
giordano, line 4: How about going to v1 at some point?  😁  At least you'd have more space to distinguish breaking changes/new features/patches, with only two numbers available that becomes more complicated.

glwagner: it's a good idea...

Let me open an issue to discuss.

navidcy: #4911

tomchor:
At the moment I think `closure=DynamicSmagorinsky()` doesn't work because we have to assign the `averaging` keyword. I think if we're only gonna export SmagorinskyLilly and DynamicSmag, then both should work out of the box. Therefore I think `LagrangianAveragedCoefficient()` should be the default averaging for DynamicSmag.

glwagner:
> At the moment I think `closure=DynamicSmagorinsky()` doesn't work because we have to assign the `averaging` keyword. I think if we're only gonna export SmagorinskyLilly and DynamicSmag, then both should work out of the box. Therefore I think `LagrangianAveragedCoefficient()` should be the default averaging for DynamicSmag.

ok, i'll make this change

navidcy (approved):

src/TurbulenceClosures/turbulence_closure_implementations/Smagorinskys/dynamic_coefficient.jl
```diff
@@ -171,9 +171,10 @@ Smagorinsky closure with
 
 `schedule`s other than `IterationInterval(1)` are valid if the coefficient at any particular location
 changes slowly relative to the frequency of recalculation.
-Some published studies compute the dynamic every 5 steps (e.g., [Bou-Zeid et al. (2005)](@cite BouZeid05);
-[Chen et al. 2016](@cite Chen2016); [Salesky et al. (2017)](@cite Salesky2017);
-[Chor et al. 2021](@cite Chor2021)) to balance fidelity with computational cost.
+Some published studies compute the dynamic coefficient every 5 steps
```
tomchor, line 183: I'd actually say that:
```suggestion
Most published studies compute the dynamic coefficient every 5 steps
```
since it's very hard to see a study with dynamic Smag that does any other frequency

glwagner: Aren't there thousands of studies that use dynamic Smag?

glwagner: Bou-Zeid has 853 citations and we don't even have solid info that this is the only Lagrangian-averaged dynamic smag out there...

glwagner: Perhaps the other reason, which I feel pretty strongly about, is that I do not want to be "pressuring" users into using a certain approximation by saying "just so you know, everyone does it this way, not sure why you don't" I don't think it advances the cause of science.

PR #4914, opened 2025-11-06
Extend API to allow passing a buffer scheme to advection schemes

simone-silvestri:
This was not possible at the moment, but it might be useful to test interactions between different schemes at the boundary (and the code change is minimal)

### DIFF
--- a/src/Advection/centered_reconstruction.jl
+++ b/src/Advection/centered_reconstruction.jl
@@ -10,12 +10,12 @@ end
 
 function Centered(FT::DataType=Oceananigans.defaults.FloatType; 
                   order = 2,
-                  buffer_scheme = nothing)
+                  buffer_scheme = DescreasingOrderAdvectionScheme())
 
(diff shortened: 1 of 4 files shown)


src/Advection/centered_reconstruction.jl
```diff
@@ -8,14 +8,19 @@ struct Centered{N, FT, CA} <: AbstractCenteredAdvectionScheme{N, FT}
...
+                  buffer_scheme = nothing)
+
     mod(order, 2) != 0 && throw(ArgumentError("Centered reconstruction scheme is defined only for even orders"))
 
     N  = Int(order ÷ 2)
-    if N > 1
-        buffer_scheme = Centered(FT; order=order-2)
-    else
-        buffer_scheme = nothing
+    if isnothing(buffer_scheme)
```
glwagner: you might want to come up with a name for this scheme. Because I feel like "buffer_scheme=nothing" has an actual meaning 🤔

PR #4921, opened 2025-11-10
Correctly load `_part*.jld2` files in `FieldTimeSeries`

simone-silvestri:
the order of loading was not preserved since, for example, `part10.jld2` was loaded before `part2.jld2`

### DIFF
--- a/src/OutputReaders/field_time_series.jl
+++ b/src/OutputReaders/field_time_series.jl
@@ -451,7 +451,7 @@ end
 function naturalsort(x::Vector{String})
     f = text -> all(isnumeric, text) ? Char(parse(Int, text)) : text
     sorter = key -> join(f(m.match) for m in eachmatch(r"[0-9]+|[^0-9]+", key))
-    sort(x, by=sorter)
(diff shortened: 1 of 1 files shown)


src/OutputReaders/field_time_series.jl
```diff
@@ -446,6 +446,14 @@ function FieldTimeSeries{LX, LY, LZ}(grid::AbstractGrid, times=(); kwargs...) wh
     return FieldTimeSeries(loc, grid, times; kwargs...)
 end
 
+# Function to naturally sort strings containing numbers, code credit:
+# https://discourse.julialang.org/t/sorting-strings-containing-numbers-so-that-a2-a10/5372/28
+function naturalsort(x::Vector{String})
+    f = text -> all(isnumeric, text) ? Char(parse(Int, text)) : text
+    sorter = key -> join(f(m.match) for m in eachmatch(r"[0-9]+|[^0-9]+", key))
+    sort(x, by=sorter)
```
navidcy:
```suggestion
    return sort(x, by=sorter)
```

PR #4924, opened 2025-11-10
Use default fallback for `isapprox` on `Fields`

### LINKED ISSUE #4920: Fix scalar operations in `isapprox` between `Field`
It seems that we currently use broadcasting for `isapprox`. This doesn't seem to hit the right codepaths and ends up invoking scalar operations

https://github.com/CliMA/Oceananigans.jl/blob/bc4299048e26143971808b708feb737f0ca7ad8f/src/Fields/field.jl#L832-L837

Need to dig into this more...

cc @giordano

giordano:
The default fallback should be good enough for getting the right result without having to enable scalar indexing.

Contrary to what I said in #4920, generic fallback of `isapprox` seems to work on `ImmersedBoundaryGrid`, unless I'm missing something obvious about it?   Let me know if the tests can be improved!

Fix #4920.

### DIFF
--- a/test/test_field.jl
+++ b/test/test_field.jl
@@ -496,8 +496,10 @@ end
                 @info "  Testing isapprox on fields [$(typeof(arch)), $FT, $(nameof(typeof(grid)))]..."
                 u = CenterField(grid)
                 v = CenterField(grid)
-                set!(u, FT(1))
-                set!(v, FT(1))
+                set!(u, 1)
+                set!(v, 1)
+                Oceananigans.ImmersedBoundaries.mask_immersed_field!(u, 1)
+                Oceananigans.ImmersedBoundaries.mask_immersed_field!(v, 2)
                 # Make sure the two fields are the same
                 @test isapprox(u, v)
                 @test isapprox(u, v; rtol=0, atol=0)


test/test_field.jl
```diff
@@ -477,6 +479,40 @@ end
...
+            W = 1000.0
+            mountain(x, y) = H * exp(-(x^2 + y^2) / 2W^2)
+            imm_grid = ImmersedBoundaryGrid(rect_grid, GridFittedBottom(mountain))
+
+            for grid in (rect_grid, imm_grid)
+                @info "  Testing isapprox on fields [$(typeof(arch)), $FT, $(nameof(typeof(grid)))]..."
+                u = CenterField(grid)
+                v = CenterField(grid)
+                set!(u, FT(1))
+                set!(v, FT(1))
```
glwagner:
```suggestion
                set!(u, 1)
                set!(v, 1)
                Oceananigans.ImmersedBoundaries.mask_immersed_field!(u, 1)
                Oceananigans.ImmersedBoundaries.mask_immersed_field!(v, 2)
```

This will fill the immersed regions with different values to ensure that they don't pollute the test.

test/test_field.jl
```diff
@@ -477,6 +479,40 @@ end
...
+                # Make sure the two fields are the same
+                @test isapprox(u, v)
+                @test isapprox(u, v; rtol=0, atol=0)
+
+                set!(v, FT(1.1))
+                @test !isapprox(u, v)
+                @test isapprox(u, v; rtol=0.1)
+                @test !isapprox(u, v; atol=2.0)
```
giordano, line 510: Note that this gives a different result than the previous definition, it's the same as:
```julia-repl
julia> u = ones(8, 8, 8);

julia> v = fill(1.1, 8, 8, 8);

julia> all(isapprox(u, v; atol=2.0))
false

julia> all(isapprox.(u, v; atol=2.0))
true
```
But the previous definition was inconsistent with [`isapprox`](https://docs.julialang.org/en/v1/base/math/#Base.isapprox) semantic on arrays.

test/test_field.jl
glwagner: also in my opinion we should use Int when possible (not convert to FT). type conversion should be handled internally and not be the responsibility of user

giordano: The test still passes for me without having to define a custom `isapprox` after this change.

test/test_field.jl
glwagner: what is the difference between the two?

giordano: `isapprox` takes the norm of the whole arguments, elementwise `isapprox` would compare the norms of the individual pairwise arguments.

glwagner: ooh makes sense.

Interesstingly one can also define yet another norm for a `Field` (because we have the concept of a `grid`, so eg smaller cells should contriubte less than larger cells). But we don't really need that...

giordano: `isapprox` allows passing custom norm functions already.

PR #4927, opened 2025-11-11
Avoid type piracy in `OceananigansXESMFext`

simone-silvestri:


simone-silvestri:
Actually I am wrong, there is no type piracy as XESMF does not define a `regrid!` method, but regridding is called by applying the regridder object. I ll close this...

PR #4930, opened 2025-11-12
Change regularize_forcing to materialize_forcing, plus refine model_forcing interface

glwagner:
This PR cleans up the interface for forcing with Models. It's inspired by some similar improvements for Breeze.AtmosphereModel. In particular it more clearly distinguishes between "prognostic fields" and "model fields". Model fields (perhaps not the greatest name) is a superset of prognostic fields and may include selected auxiliary fields (but typically not _all_ auxiliary fields) that the model designer deems could be useful for users within forcing and boundary condition functions.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.101.0"
+version = "0.101.1"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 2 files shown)


simone-silvestri (approved):

giordano:
This breaks Breeze: https://github.com/NumericalEarth/Breeze.jl/blob/44b470a63b625fa277c35107ca6218655bc11f09/src/AtmosphereModels/atmosphere_model.jl#L232

glwagner:
> This breaks Breeze: https://github.com/NumericalEarth/Breeze.jl/blob/44b470a63b625fa277c35107ca6218655bc11f09/src/AtmosphereModels/atmosphere_model.jl#L232

Let's bump the patch after #4936

PR #4931, opened 2025-11-12
Add missing method for `ReactantState` architecture

simone-silvestri:
`cpu_interpolating_time_indices` is missing for a `ReactantState` architecture.
This PR introduces it.
exposed in this error: https://github.com/CliMA/ClimaOcean.jl/actions/runs/19296994671/job/55181367934

I am not sure that what I wrote here is valid though, this assumes that `times` for a `ReactantState` architecture are accessible and stored on the CPU


glwagner (approved):
happy if test pass

PR #4933, opened 2025-11-13
Don't materialize gradients by default

glwagner:
I don't think we want to allocate memory for these by default, since they are not used in the majority of cases (especially the horizontal gradients).

### DIFF
--- a/src/BuoyancyFormulations/buoyancy_force.jl
+++ b/src/BuoyancyFormulations/buoyancy_force.jl
@@ -48,7 +48,7 @@ NonhydrostaticModel{CPU, RectilinearGrid}(time = 0 seconds, iteration = 0)
 └── coriolis: Nothing
 ```
 """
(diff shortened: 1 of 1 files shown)


glwagner:
@giordano though this will resolve the error observed on https://github.com/NumericalEarth/Breeze.jl/pull/155, we probably still want to add gradients for MoistAirBuoyancy, since this will allow us to support more physics/closures

simone-silvestri (commented):
Oh right, I might have left it a default! Maybe in a future PR we can add some default behavior based on the closures passed to the model.

simone-silvestri (approved):

glwagner:
> Oh right, I might have left it a default! Maybe in a future PR we can add some default behavior based on the closures passed to the model.

for sure, it probably makes sense to precompute them for GM. Maybe also the vertical component for the vertical mixing closures and SmagorinskyLilly.

PR #4939, opened 2025-11-14
Refactor the buffer system and add methods for `synchronize_communication!`

simone-silvestri:
I am starting to dissect #4811 now that the formulation is complete and the tests pass, so that we do not have to merge 2000 lines and 107 files altogether.

This is a very simple start -> a quality of life refactor of the buffer system for distributed computations.

### DIFF
--- a/src/DistributedComputations/communication_buffers.jl
+++ b/src/DistributedComputations/communication_buffers.jl
@@ -188,12 +188,12 @@ function y_communication_buffer(arch::Distributed, grid::AbstractGrid{<:Any, TX,
     FT = eltype(data)
     Nx = size(grid, 1)
     if (ranks(arch)[1] == 1) || (TX == RightConnected) || (TX == LeftConnected)
-        send = on_architecture(arch, zeros(FT, H, Tx, Tz))
-        recv = on_architecture(arch, zeros(FT, H, Tx, Tz))
+        send = on_architecture(arch, zeros(FT, Tx, H, Tz))
+        recv = on_architecture(arch, zeros(FT, Tx, H, Tz))
         return OneDBuffer(send, recv)
     else
-        send = on_architecture(arch, zeros(FT, H, Nx, Tz))
-        recv = on_architecture(arch, zeros(FT, H, Nx, Tz))
+        send = on_architecture(arch, zeros(FT, Nx, H, Tz))
+        recv = on_architecture(arch, zeros(FT, Nx, H, Tz))
         return TwoDBuffer(send, recv)
     end
 end
@@ -210,8 +210,8 @@ end
 function y_communication_buffer(arch, grid, data, H, ::MCBC) 
     Tx, _, Tz = size(parent(data))
     FT = eltype(data)
-    send = on_architecture(arch, zeros(FT, H, Tx, Tz))
-    recv = on_architecture(arch, zeros(FT, H, Tx, Tz))
+    send = on_architecture(arch, zeros(FT, Tx, H, Tz))
+    recv = on_architecture(arch, zeros(FT, Tx, H, Tz))
     return OneDBuffer(send, recv)
 end
 


simone-silvestri:
and splitting some tests to speed up the distributed tests...

src/DistributedComputations/communication_buffers.jl
```diff
@@ -50,67 +70,77 @@ CommunicationBuffers(grid, data, ::Nothing) = nothing
...
+y_communication_buffer(arch, grid, data, H, bc) = nothing
 
-function x_communication_buffer(arch::Distributed, grid, data, H, ::DCBC)
-    # Either we pass corners or it is a 1D parallelization in x
-    size_y = arch.ranks[2] == 1 ? size(parent(data), 2) : size(grid, 2)
-    return (send = on_architecture(arch, zeros(eltype(data), H, size_y, size(parent(data), 3))),
-            recv = on_architecture(arch, zeros(eltype(data), H, size_y, size(parent(data), 3))))
+# Either we pass corners or it is a 1D parallelization in x, note that, even for 2D parallelizations, we need a 
+# buffer that includes the corners (i.e. a OneDBuffer) if we are at the edge of the domain in y since we are not
+# passing all the corners in that case
```
glwagner: now that we have AI to make things pretty easy, I think it makes sense to convert all these comments to docstrings. This opens the possibility to getting a gist of the program by looking at the docs

src/DistributedComputations/communication_buffers.jl
```diff
@@ -50,67 +70,77 @@ CommunicationBuffers(grid, data, ::Nothing) = nothing
...
+function x_communication_buffer(arch::Distributed, grid::AbstractGrid{FT, TX, TY}, data, H, ::DCBC) where {FT, TX, TY}
+    if (arch.ranks[2] == 1) || (TY == RightConnected) || (TY == LeftConnected)
+        return OneDBuffer(on_architecture(arch, zeros(eltype(data), H, size(parent(data), 2), size(parent(data), 3))),
+                          on_architecture(arch, zeros(eltype(data), H, size(parent(data), 2), size(parent(data), 3))))
```
glwagner:
```suggestion
        _, Ty, Tz = size(parent(data))
        FT = eltype(data)
        send = on_architecture(arch, zeros(FT, H, Ty, Tz))
        recv = on_architecture(arch, zeros(FT, H, Ty, Tz))
        return OneDBuffer(send, recv)
```

src/DistributedComputations/communication_buffers.jl
```diff
@@ -50,67 +70,77 @@ CommunicationBuffers(grid, data, ::Nothing) = nothing
...
+    else
+        return (send = on_architecture(arch, zeros(eltype(data), H, size(grid, 2), size(parent(data), 3))),
+                recv = on_architecture(arch, zeros(eltype(data), H, size(grid, 2), size(parent(data), 3))))
```
glwagner: this named tuple looks like the same thing as `OneDBuffer`. What's going on here? Do you use the type secretly for dispatch somehow? Do we need to use an explicit type here rather than a named tuple?

A comment might clear this up.

glwagner:
```suggestion
        Tz = size(parent(data), 3)
        Ny = size(grid, 2)
        FT = eltype(data)
        return (send = on_architecture(arch, zeros(FT, H, Ny, Tz)), recv = on_architecture(arch, zeros(FT, H, Ny, Tz))) 
```

glwagner:
I'd love if we can take the opportunity of the refactor to clean up the code and make it readable, eg add comments in mysterious parts, use docstrings, and convert golf'd code to a more evocative style

src/DistributedComputations/communication_buffers.jl
simone-silvestri: right it is only for dispatch, I can add another buffer type: `TwoDBuffer`

simone-silvestri:
> damn that looks good

github copilot at the rescue :)

PR #4940, opened 2025-11-14
Three dimensional `column_depth` methods

simone-silvestri:
for use in `KernelOperation`s


glwagner (approved):

glwagner:
I've always wondered if we should try to figure out an abstraction for the "kernel function configuration"... but this is pretty simple

PR #4943, opened 2025-11-14
Remove integrated areas in `ImplicitFreeSurface` and refactor solver

simone-silvestri:
By using a `KernelOperation`, we can include a mutable area in the solver. 
This is required to have a non-linear implicit free surface and is another step towards #4811 

This PR also removes the extra `barotropic_volume_flux` storage since it is useless: the barotropic velocities can be computed on the fly

### DIFF
--- a/test/test_implicit_free_surface_solver.jl
+++ b/test/test_implicit_free_surface_solver.jl
@@ -67,13 +67,8 @@ function run_implicit_free_surface_solver_tests(arch, grid, free_surface)
     g = Oceananigans.defaults.gravitational_acceleration
     η = model.free_surface.η
 
-    ∫ᶻ_Axᶠᶜᶜ = Field{Face, Center, Nothing}(grid)
-    ∫ᶻ_Ayᶜᶠᶜ = Field{Center, Face, Nothing}(grid)
-
-    vertically_integrated_lateral_areas = (xᶠᶜᶜ = ∫ᶻ_Axᶠᶜᶜ, yᶜᶠᶜ = ∫ᶻ_Ayᶜᶠᶜ)
-
-    compute_vertically_integrated_lateral_areas!(vertically_integrated_lateral_areas)
-    fill_halo_regions!(vertically_integrated_lateral_areas)
+    ∫ᶻ_Axᶠᶜᶜ = KernelFunctionOperation{Face, Center, Nothing}(Oceananigans.Models.HydrostaticFreeSurfaceModels.integrated_x_area, grid)
+    ∫ᶻ_Ayᶜᶠᶜ = KernelFunctionOperation{Center, Face, Nothing}(Oceananigans.Models.HydrostaticFreeSurfaceModels.integrated_y_area, grid)
 
     left_hand_side = ZFaceField(grid, indices = (:, :, grid.Nz + 1))
     implicit_free_surface_linear_operation!(left_hand_side, η, ∫ᶻ_Axᶠᶜᶜ, ∫ᶻ_Ayᶜᶠᶜ, g, Δt)


src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
```diff
@@ -42,6 +42,31 @@ on_architecture(arch, coord::ZStarCoordinate) = ZStarCoordinate(on_architecture(
...
+##### Utilities to compute the vertically integrated ``barotropic'' velocities
+#####
+
+# If U and V are prognostic (for example in `SplitExplicitFreeSurface`), we use them
+@inline barotropic_U(i, j, k, grid, U, u) = @inbounds U[i, j, k]
+@inline barotropic_V(i, j, k, grid, V, v) = @inbounds V[i, j, k]
+
+# If either U or V are not available, we compute them
+@inline function barotropic_U(i, j, k, grid, ::Nothing, u)
+    U = 0
```
glwagner:
```suggestion
    U = zero(grid)
```

src/Models/HydrostaticFreeSurfaceModels/HydrostaticFreeSurfaceModels.jl
```diff
@@ -42,6 +42,31 @@ on_architecture(arch, coord::ZStarCoordinate) = ZStarCoordinate(on_architecture(
...
+    for k in 1:size(grid, 3)
+        @inbounds U += u[i, j, k] * Δzᶠᶜᶜ(i, j, k, grid)
+    end
+    return U
+end
+
+@inline function barotropic_V(i, j, k, grid, ::Nothing, v)
+    V = 0
```
glwagner:
```suggestion
    V = zero(grid)
```

src/Models/HydrostaticFreeSurfaceModels/fft_based_implicit_free_surface_solver.jl
```diff
@@ -101,15 +101,17 @@ function compute_implicit_free_surface_right_hand_side!(rhs, implicit_solver::FF
 
     launch!(arch, grid, :xy,
             fft_implicit_free_surface_right_hand_side!,
-            rhs, grid, g, Lz, Δt, ∫ᶻQ, η)
```
glwagner, line 104: `∫ᶻQ` was a problematic name anyways 😂

src/Models/HydrostaticFreeSurfaceModels/implicit_free_surface.jl
```diff
@@ -100,24 +98,18 @@ function materialize_free_surface(free_surface::ImplicitFreeSurface{Nothing}, ve
...
     return ImplicitFreeSurface(η,
                                gravitational_acceleration,
-                               barotropic_volume_flux,
                                solver,
                                solver_method,
                                free_surface.solver_settings)
 end
 
-build_implicit_step_solver(::Val{:Default}, grid::XYRegularRG, settings, gravitational_acceleration) =
+build_implicit_step_solver(::Val{:Default}, grid::XYRegularStaticRG, settings, gravitational_acceleration) =
```
glwagner: given the confusion discovered about "mutable" vs "generalized" --- which matters here? Does this work for a genearlized coordinate that is static? Or does it also require a non-generalized, AND static?

simone-silvestri: I think the FFT can be used only with a flat bottom right? So it should be valid only for non-generalized and static

glwagner: mm yes.

simone-silvestri: I removed this change before merging. We can deal with it when we enable zstar with implicit free surface and correct the naming for the vertical coordinate

Issue #4944, opened 2025-11-14
`compute!(field, time)` doesn't update status of `field`


erin-atkinson:
```julia
grid = RectilinearGrid(; size=(8, 8, 8), extent=(1, 1, 1), topology=(Periodic, Periodic, Periodic))

a = CenterField(grid)
b = Field(a + a)
c = Field(b + b)

a.status, b.status, c.status
```
```
(nothing, Oceananigans.Fields.FieldStatus{Float64}(0.0), Oceananigans.Fields.FieldStatus{Float64}(0.0))
```
After `compute!(c, 1)`, the fields have the following status:
```
(nothing, Oceananigans.Fields.FieldStatus{Float64}(1.0), Oceananigans.Fields.FieldStatus{Float64}(0.0))
```

It seems that `compute!` in [AbstractOperations/computed_field.jl](https://github.com/CliMA/Oceananigans.jl/blob/d270593c2f28278132b59dcc5fbce76fb8201b7b/src/AbstractOperations/computed_field.jl#L80) only updates the status of the dependencies.

erin-atkinson:
Closing was a mistake...

PR #4945, opened 2025-11-14
Properly set status for computed fields

erin-atkinson:
Fix for #4944

### DIFF
--- a/src/Fields/scans.jl
+++ b/src/Fields/scans.jl
@@ -77,13 +77,14 @@ const ScannedComputedField = Field{<:Any, <:Any, <:Any, <:Scan}
 function compute!(field::ScannedComputedField, time=nothing)
     s = field.operand
     compute_at!(s.operand, time)
-    set_status!(field.status, time)
 
     if s.type isa AbstractReducing
         s.scan!(field, s.operand)
     elseif s.type isa AbstractAccumulating
         s.scan!(field, s.operand; dims=s.dims)
     end
+
+    set_status!(field.status, time)
 
     return field
 end
@@ -306,4 +307,3 @@ end
         @inbounds B[i, j, k] = op(B[i, j, pr], A[i, j, k])
     end
 end
-


glwagner:
Before this change `compute!` is guaranteed to make a computation regardless of `time`. So this changes the behavior because now the computation will only happen if the status check is false (which is the same behavior as `compute_at!`). I _think_ the point is that we do want there to be a difference.

However, it does make sense to update the status in `compute!` I think.

erin-atkinson:
Ah okay, I wondered whether that was the case. The documentation is what needs changing then

src/AbstractOperations/computed_field.jl
```diff
@@ -82,8 +82,22 @@ function compute!(comp::ComputedField, time=nothing)
     compute_at!(comp.operand, time)
 
     # Now perform the primary computation
-    @apply_regionally compute_computed_field!(comp)
+    compute_at!(comp, time)
+
+    return comp
+end
+
+function compute_at!(comp::ComputedField, time)
```
glwagner: is this needed / different than the general version here?

https://github.com/CliMA/Oceananigans.jl/blob/d270593c2f28278132b59dcc5fbce76fb8201b7b/src/Fields/field.jl#L532

glwagner:
> Ah okay, I wondered whether that was the case. The documentation is what needs changing then

it is for sure a little confusing. better docs would be great. I also think it makes sense to set the status. I can't think of why we wouldn't want to.

src/AbstractOperations/computed_field.jl
erin-atkinson: I was a bit confused about this. It was a bit of a circular definition if `compute!` was intended to work like how its docstring implies, so this new definition was needed. If `compute!` should always compute the field then I think it's not necessary.

glwagner:
Note if you set `status` in `compute!` then we won't need this line:

https://github.com/CliMA/Oceananigans.jl/blob/d270593c2f28278132b59dcc5fbce76fb8201b7b/src/Fields/field.jl#L539

erin-atkinson:
I guess it just needs a single line to set the new status. I changed the docstring a little too.

src/AbstractOperations/computed_field.jl
```diff
@@ -75,17 +75,19 @@ end
...
     # First compute `dependencies`:
     compute_at!(comp.operand, time)
 
     # Now perform the primary computation
     @apply_regionally compute_computed_field!(comp)
-
     fill_halo_regions!(comp)
 
+    # Update status
+    isnothing(time) || comp.status.time = time
```
glwagner:
```suggestion
    !isnothing(time) && comp.status isa FieldStatus && comp.status.time = time
```

It is possible to create fields without a status (this is a "feature" which allows people to re-use scratch fields in diagostics systems to save memory)

could also dispatch:

```julia
set_status!(status, time) = nothing
set_status!(status::FieldStatus, time::Number) = status.time = time
```

glwagner (approved):
will merge once tests pass and last comment is addressed! thank you @erin-atkinson !

src/AbstractOperations/computed_field.jl
erin-atkinson: I defined `set_status!` underneath `FieldStatus` and used it.

src/Fields/field.jl
```diff
@@ -508,6 +508,11 @@ end
 FieldStatus() = FieldStatus(0.0)
 Adapt.adapt_structure(to, status::FieldStatus) = (; time = status.time)
 
+set_status!(status, time) = nothing
+
+set_status!(status::FieldStatus, time::Nothing) = nothing
```
navidcy, line 512: Do we need the fallback? It's the same as the other method, isn't it?

```suggestion
set_status!(status::FieldStatus, ::Nothing) = nothing
```

navidcy: I think the `set_status!(status, time) = nothing` method was anyway being overwritten with the `set_status!(status::FieldStatus, time) = status.time = time ` method, right?

erin-atkinson: You are right, I should have removed that when I changed the previous suggestion

EDIT: actually no, we need a fallback for fields without a status (like Gregory mentioned earlier)

navidcy: Gotcha! Agreed.

navidcy:
This seems to fail:

```Julia
julia> grid = RectilinearGrid(size=(2, 2, 2), extent=(1, 1, 1));

julia> c = CenterField(grid);

julia> C = Field(Average(c; dims));

julia> set!(c, 1);

julia> compute_at!(C, 1.0)
1×2×2 Field{Nothing, Center, Center} reduced over dims = (1,) on RectilinearGrid on CPU
├── data: OffsetArray{Float64, 3, Array{Float64, 3}}, size: (1, 2, 2)
├── grid: 2×2×2 RectilinearGrid{Float64, Periodic, Periodic, Bounded} on CPU with 2×2×2 halo
├── operand: Average of 2×2×2 Field{Center, Center, Center} on RectilinearGrid on CPU over dims (1,)
├── status: time=0.0
└── data: 1×6×6 OffsetArray(::Array{Float64, 3}, 1:1, -1:4, -1:4) with eltype Float64 with indices 1:1×-1:4×-1:4

julia> C.status.time
0.0
```

The `C.status.time` should have been 1.0, right?

navidcy:
I think I fixed it with [2c104e0](https://github.com/CliMA/Oceananigans.jl/pull/4945/commits/2c104e00d400f4f74cd79d107b6170e56474fcc7); let's see if tests pass now.

navidcy:
> I think I fixed it with [2c104e0](https://github.com/CliMA/Oceananigans.jl/pull/4945/commits/2c104e00d400f4f74cd79d107b6170e56474fcc7); let's see if tests pass now.

@glwagner if there’s a cleaner solution I’d be keen to hear. But otherwise this should be good to merge!

src/Fields/scans.jl
```diff
@@ -77,6 +77,7 @@ const ScannedComputedField = Field{<:Any, <:Any, <:Any, <:Scan}
 function compute!(field::ScannedComputedField, time=nothing)
     s = field.operand
     compute_at!(s.operand, time)
+    set_status!(field.status, time)
```
glwagner: Should you call this at the end, after the computation is performed? I guess if there is an error, the status will change but the field has not actually been computed

PR #4947, opened 2025-11-15
Fix bug in multi-region `SplitExplicitFreeSurface`

simone-silvestri:
I noticed there was this bug. 
This meant that the correct fill halos was not hit for a cubed sphere, since it requires a tuple of fields with `(Face, Center, <:Any)` and `(Center, Face, <:Any)`. defined here
https://github.com/CliMA/Oceananigans.jl/blob/5bbc2a1df622e233ccd97d8f6a86bba069c3a7c2/src/MultiRegion/cubed_sphere_boundary_conditions.jl#L345-L347

@siddharthabishnu this might be the error that you were searching for? Can you test the baroclinic aquaplanet on this branch? If this fixes you issue can you also add a test?


siddharthabishnu:
> I noticed there was this bug. This meant that the correct fill halos was not hit for a cubed sphere, since it requires a tuple of fields with `(Face, Center, <:Any)` and `(Center, Face, <:Any)`. defined here
> 
> https://github.com/CliMA/Oceananigans.jl/blob/5bbc2a1df622e233ccd97d8f6a86bba069c3a7c2/src/MultiRegion/cubed_sphere_boundary_conditions.jl#L345-L347
> 
> @siddharthabishnu this might be the error that you were searching for? Can you test the baroclinic aquaplanet on this branch? If this fixes you issue can you also add a test?

It might very well be! This is a major bug! Great catch! I clearly remember implementing this correctly earlier, so I must have missed it later while resolving merge conflicts for another major PR. I’ll add a test.

PR #4951, opened 2025-11-17
Run benchmarks on julia 1.12

simone-silvestri:
I want to see if there is a difference from julia 1.10 in kernel execution


glwagner:
what'd you find?

glwagner (approved):

simone-silvestri:
It looks like there is some variance, with a maximum of about 3% in performance. Some kernels are faster, some are slower, but in general, I think the performance remains the same.

PR #4953, opened 2025-11-17
Remove unused cubed spheres module

simone-silvestri:
and the corresponding tests, which are not included in Oceananigans.
The implementation will remain in github history if we need to look back to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,20 +133,20 @@ If you use Oceananigans for your research, teaching, or fun 🤩, everyone in ou
 if you credit Oceananigans by name.
 
 The community has published a number of articles describing the development of Oceananigans,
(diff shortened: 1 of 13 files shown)


navidcy:
I started seeing this DOCUMENTER_KEY error:

https://buildkite.com/clima/oceananigans/builds/26874#019a92c9-65fb-4b90-87b1-d36d50d5faca/38-3394

@ali-ramadhan did something change on the machine the docs are being built?

ali-ramadhan:
@navidcy Nothing should have changed as far as I know. I do see the docs passing on main so I'm curious to know what happened here. In case it's a random/intermittent issue I might first try just restarting the Buildkite Docker container once there are no jobs running.

navidcy (approved):

navidcy (commented):
nothing

Issue #1113, opened 2020-10-29
When multithreading use 4 times more threads for FFTW


ali-ramadhan:
Super easy fix. Suggested by @vchuravy.

Should rerun multithreading benchmarks after this.

glwagner:
See also 

https://github.com/JuliaMath/FFTW.jl/pull/105

and

https://github.com/JuliaMath/FFTW.jl/pull/151

milankl:
The `FFTW.set_num_threads` is a global FFTW configuration that ideally shouldn't be set to avoid affecting other packages that are loaded in the same session also depended on FFTW see https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/919. PR #4958 fixes that

PR #4958, opened 2025-11-18
Oceananigans-local FFTW num threads setting

milankl:
Fixes https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/919


src/Oceananigans.jl
```diff
@@ -118,6 +118,8 @@ export
 using DocStringExtensions
 using FFTW
 
+const FFTW_NUM_THREADS = Ref{Int}(1)
```
simone-silvestri: We have a `Oceananigans.defaults` container for global defaults here:
https://github.com/CliMA/Oceananigans.jl/blob/499ce53cc127cf31590ca410f58ac4489cd4164b/src/Oceananigans.jl#L148-L152

Probably this should be stored there?

milankl:
@glwagner suggested to move `FFTW_NUM_THREADS` into Solvers

milankl:
Related to #1113

glwagner:
> @glwagner suggested to move `FFTW_NUM_THREADS` into Solvers

we could use the "4 * threads" rule as the default. Then there is already a global default for the threads, so no need to add to `Oceananigans.defaults`.

There could be scope _also_ for an addition default, like `fftw_threads`. The main lift is to make it an optoin of the solver though. Adding a property to `Oceananigans.defaults` could be done as a subsequent step, or saved for another PR.

milankl:
It only affects the Solver module though so I agree with your comment that it should be a setting locally in it. I set it up as Ref so one can always manually change it anytime but by default it's set to `4*Threads.nthreads()`. I've also set it up as default argument to the `plan_forward_transform`/`plan_backward_transform` so it's still possible to create other plans ignoring the Solver-global `FFTW_NUM_THREADS` "constant"

glwagner:
The main reason for the `defaults` object is when multiple defaults neeed to be changed at once. This would mainly come up (in the future) for an ocean and atmos model:

```julia
using Breeze, Oceananigans

atmos_model = AtmosphereModel(grid) # uses default threads

pressure_solver = FFTBasedPressureSolver(grid, num_threads=1)
ocean_model = NonhydrostaticModel(; grid, pressure_solver)
```

more compact and simple could be

```julia
using Breeze, Oceananigans
Oceananigans.defaults.FFTW_threads = 1
atmos_model = AtmosphereModel(grid)
ocean_model = NonhydrostaticModel(; grid)
```

glwagner (approved):

glwagner:
I think we can merge this and maybe add a TODO to add this as a default, and/or implement a `num_threads` property for the solvers. As-is though I think this is progress and useful.

PR #4963, opened 2025-11-19
Update CPU architecture reference in OutputReaders.jl

simone-silvestri:
This is another bug found in ClimaOcean. Not sure why the test does not capture it.

### DIFF
--- a/ext/OceananigansReactantExt/OutputReaders.jl
+++ b/ext/OceananigansReactantExt/OutputReaders.jl
@@ -24,7 +24,7 @@ import Oceananigans.OutputReaders: find_time_index, cpu_interpolating_time_indic
 end
 
 function cpu_interpolating_time_indices(::ReactantState, times, time_indexing, t) 
-    cpu_times = on_architecture(CPU(), times)
+    cpu_times = on_architecture(Oceananigans.Architectures.CPU(), times)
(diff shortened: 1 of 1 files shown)


ext/OceananigansReactantExt/OutputReaders.jl
```diff
@@ -24,7 +24,7 @@ import Oceananigans.OutputReaders: find_time_index, cpu_interpolating_time_indic
 end
 
 function cpu_interpolating_time_indices(::ReactantState, times, time_indexing, t) 
-    cpu_times = on_architecture(CPU(), times)
+    cpu_times = on_architecture(Oceananigans.Architectures.CPU(), times)
```
navidcy, line 27: Wait, what's the difference?

simone-silvestri: `CPU` was not defined

navidcy: And it wasn’t breaking?

simone-silvestri: yeah I am not sure if we are doing something weird in ClimaOcean, but I have added a test for fts using Reactant and it was not breaking.

PR #4967, opened 2025-11-20
Clean up; implement `buoyancy_force` interface for TurbulenceClosures

glwagner:
This PR generalizes buoyancy abstractions (and de-conflicts names where needed) to make turbulence closures compatible with Breeze's AtmosphereModel.

### DIFF
--- a/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
+++ b/src/TurbulenceClosures/turbulence_closure_implementations/TKEBasedVerticalDiffusivities/time_step_catke_equation.jl
@@ -113,9 +113,9 @@ function time_step_catke_equation!(model, ::SplitRungeKutta3TimeStepper)
     tracer_index = findfirst(k -> k == :e, keys(model.tracers))
     implicit_solver = model.timestepper.implicit_solver
     stage = model.clock.stage
-    β = (β¹, β², 1)[stage]
-
-    Δt = model.clock.last_Δt / β
+    β = (model.timestepper.β¹, model.timestepper.β², one(model.timestepper.β¹))
+    βn = β[stage]
+    Δt = model.clock.last_Δt / βn
 
     tracers = buoyancy_tracers(model)
     buoyancy = buoyancy_force(model)


src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -257,7 +258,5 @@ validate_momentum_advection(momentum_advection::VectorInvariant, grid::Orthogona
 validate_momentum_advection(momentum_advection, grid::OrthogonalSphericalShellGrid) = error("$(typeof(momentum_advection)) is not supported with $(typeof(grid))")
 
 initialize!(model::HydrostaticFreeSurfaceModel) = initialize_free_surface!(model.free_surface, model.grid, model.velocities)
-
-# return the total advective velocities
 @inline total_velocities(model::HydrostaticFreeSurfaceModel) = model.velocities
-
+buoyancy_force(model::HydrostaticFreeSurfaceModel) = model.buoyancy
```
navidcy:
```suggestion
buoyancy_force(model::HydrostaticFreeSurfaceModel) = model.buoyancy

```

src/Models/Models.jl
```diff
@@ -216,9 +215,9 @@ function required_checkpoint_properties(model::OceananigansModels)
     return properties
 end
 
-# Implementation of a `seawater_density` `KernelFunctionOperation
-# applicable to both `NonhydrostaticModel` and  `HydrostaticFreeSurfaceModel`
+# Implementation of diagnostics applicable to both `NonhydrostaticModel` and  `HydrostaticFreeSurfaceModel`
```
navidcy:
```suggestion
# Implementation of diagnostics applicable to both `NonhydrostaticModel` and `HydrostaticFreeSurfaceModel`
```

src/Models/buoyancy_operation.jl
```diff
@@ -0,0 +1,41 @@
...
+    BuoyancyForce,
+    BuoyancyTracer,
+    AbstractBuoyancyFormulation,
+    ∂z_b,
+    buoyancy_perturbationᶜᶜᶜ
+
+"""
+    buoyancy_operation(model)
+
+Returns a `KernelFunctionOperation` that computes the buoyancy perturbation,
```
navidcy:
```suggestion
Return a `KernelFunctionOperation` that computes the buoyancy perturbation,
```

src/Models/buoyancy_operation.jl
```diff
@@ -0,0 +1,41 @@
...
+
+Returns a `Field` that can `compute!` and store the buoyancy perturbation.
+"""
+buoyancy_field(model) = Field(buoyancy_operation(model))
+
+"""
+    buoyancy_frequency(model)
+
+Returns a `KernelFunctionOperation` that computes the vertical derivative of buoyancy,
+which is also known as the buoyancy frequency.
```
navidcy:
```suggestion
which is also known as the square of the buoyancy frequency.
```

glwagner: good point. maybe we should use a different name...

simone-silvestri (approved):

src/Models/buoyancy_operation.jl
glwagner: or maybe we should get rid of this one

glwagner: I'll leave the name for now but it might make sense to revisit this

src/Models/NonhydrostaticModels/nonhydrostatic_model.jl
```diff
@@ -300,3 +300,4 @@ end
 # return the total advective velocities
 @inline total_velocities(m::NonhydrostaticModel) = sum_of_velocities(m.velocities, m.background_fields.velocities)
 buoyancy_force(model::NonhydrostaticModel) = model.buoyancy
+buoyancy_tracers(model::NonhydrostaticModel) = model.tracers
```
navidcy:
```suggestion
buoyancy_tracers(model::NonhydrostaticModel) = model.tracers

```

src/Models/buoyancy_operation.jl
```diff
@@ -0,0 +1,41 @@
...
+buoyancy_operation(::Nothing, args...) = ZeroField()
+buoyancy_operation(::BuoyancyTracer, grid, tracers) = tracers.b
+
+buoyancy_operation(bm::AbstractBuoyancyFormulation, grid, tracers) =
+    KernelFunctionOperation{Center, Center, Center}(buoyancy_perturbationᶜᶜᶜ, grid, bm, tracers)
+
+"""
+    buoyancy_field(model)
+
+Returns a `Field` that can `compute!` and store the buoyancy perturbation.
```
navidcy:
```suggestion
Return a `Field` that can `compute!` and store the buoyancy perturbation.
```

navidcy:
CATKE timestepping breaks: 

https://buildkite.com/clima/oceananigans/builds/26995#019aa80e-43db-4464-99fd-098f0055f41a/17-489

cc @glwagner

tomchor:
@glwagner sorry, I'm a bit lost here. There used to exist `Oceananigans.BuoyancyFormulations.buoyancy`, which got the buoyancy force calculated from both `SeawaterBuoyancy` or `BuoyancyTracer`. I can't find an equivalent function now. Could you point to it?

A couple of side-notes:
1. I've been meaning to open an issue about the nomenclature of `buoyancy` in models and in the code (basically to differentiate a buoyancy _formulation_ from the buoyancy _force/acceleration_ from the buoyancy _tracer_. While this PR tries to partly address this, I probably will take this PR as a cue to finally open that issue.
2. This came up in the context of https://github.com/tomchor/Oceanostics.jl/pull/216, where I try to update Oceanostics to this new version.
3. What do people think of trying to enforce a rule where we need to give instructions at the top of breaking PR about how to migrate old code (i.e. using deprecated features) to the new version?

glwagner:
> @glwagner sorry, I'm a bit lost here. There used to exist `Oceananigans.BuoyancyFormulations.buoyancy`, which got the buoyancy force calculated from both `SeawaterBuoyancy` or `BuoyancyTracer`. I can't find an equivalent function now. Could you point to it?

I believe this is now either `Oceananigans.Models.buoyancy_operation` or `Oceananigans.Models.buoyancy_field` (I don't think that `buoyancy` consistently previously returned an operation or field but I am not 100% sure).

glwagner:
There are release notes here but they are incomplete as pointed out (that's my fault): https://github.com/CliMA/Oceananigans.jl/commit/13edbd83317b25a3c140e299840b2d43bf5beb44#comments

glwagner:
> What do people think of trying to enforce a rule where we need to give instructions at the top of breaking PR about how to migrate old code (i.e. using deprecated features) to the new version?

The usual way to do this is to include instructions in the release notes.

src/Models/buoyancy_operation.jl
```diff
@@ -0,0 +1,41 @@
...
+`Field` corresponding to the buoyancy perturbation directly, or `ZeroField` if buoyancy is `nothing`.
+"""
+buoyancy_operation(model) = buoyancy_operation(model.buoyancy, model.grid, model.tracers)
+buoyancy_operation(bf::BuoyancyForce, grid, tracers) = buoyancy_operation(bf.formulation, grid, tracers)
+
+buoyancy_operation(::Nothing, args...) = ZeroField()
+buoyancy_operation(::BuoyancyTracer, grid, tracers) = tracers.b
+
+buoyancy_operation(bm::AbstractBuoyancyFormulation, grid, tracers) =
+    KernelFunctionOperation{Center, Center, Center}(buoyancy_perturbationᶜᶜᶜ, grid, bm, tracers)
```
glwagner, line 23: @tomchor this may be what you are looking for

tomchor: Indeed this seems to have worked! Thanks!

Issue #4898, opened 2025-10-31
Lock-Exchange Implementation


surbhidhiman:
**Motivation** 
A potential example of the lock-exchange problem involves two fluids of different densities separated by a vertical barrier on a slope, and the system evolves into an underwater density current. My goal is to use this case to deepen my understanding of computational fluid dynamics, boundary conditions, and density-driven mixing. My ideal outcome is to produce a fully documented example script that meets the standards of the existing examples. 

**Proposed Implementation:** 
- Set up a RectilinearGrid 
     - ImmersedBoundaryGrid to help define the slope in the grid 
     - Topology (Bounded, Flat, Bounded)
- Set up a NonhydrostaticModel with a buoyancy tracer 
     - Boussinesq approximation
- Set up the initial conditions
     - Two fluids of different densities are separated by a vertical barrier that is removed at t=0 (lock exchange)
- Run simulation 

This outline is still a work in progress. I plan to continue refining this implementation and including some interesting new feature, and I will keep this issue updated with my progress. If there’s anything I should consider or any feedback from the community, I’d be grateful to hear it. Thank you!

simone-silvestri:
We already have a lock-exchange setup in the validation case at https://github.com/CliMA/Oceananigans.jl/blob/main/validation/z_star_coordinate/lock_release.jl

Would be quite simple to spruce it up by adding some comments and a nice visualization, and convert it into an example.

surbhidhiman:
> We already have a lock-exchange setup in the validation case at https://github.com/CliMA/Oceananigans.jl/blob/main/validation/z_star_coordinate/lock_release.jl
> 
> Would be quite simple to spruce it up by adding some comments and a nice visualization, and convert it into an example.

@simone-silvestri Thank you for sharing this!

surbhidhiman:
I have also previously discussed this idea with Francis Poulin in [this thread](https://github.com/CliMA/Oceananigans.jl/discussions/4794). My plan is to build on these ideas by enhancing the lock-exchange validation test case to incorporate an immersed boundary with a sloped bottom and a smooth tanh profile to define the initial density interface. I will also include documentation to explain the setup. I will work on this implementation and plan to open a PR soon.

glwagner:
I'd support a hydrostatic example. We need more of those.

glwagner:
Closed by #4968 . thanks @surbhidhiman !

PR #4968, opened 2025-11-20
Lock Exchange Example

surbhidhiman:
Hello Oceananigans community, I am fairly new here and I am interested in building my simulations and contribution skills with the Oceananigans package. I have opened a discussion post [here](https://github.com/CliMA/Oceananigans.jl/discussions/4794) and an issue [here](https://github.com/CliMA/Oceananigans.jl/issues/4898), with idea of implementing a basic lock-exchange problem. My goal is to use this case to deepen my understanding of computational fluid dynamics, boundary conditions, and density-driven mixing. My ideal outcome is to produce a fully documented example script that meets the standards of the existing examples. 

I have opened this PR with a very basic version which includes my current understanding of how to run the simulations. I will happily welcome any and all feedback for any changes or ideas that would help me build on this existing script to make it a worthwhile contribution. So far, I have pulled from the [internal tide example](https://github.com/CliMA/Oceananigans.jl/blob/main/examples/internal_tide.jl) as well as the [lock release validation case](https://github.com/CliMA/Oceananigans.jl/blob/main/validation/z_star_coordinate/lock_release.jl), and I will continue to build on what I have so far. 

Thank you in advance for your time and help!

### DIFF
--- a/examples/lock_exchange.jl
+++ b/examples/lock_exchange.jl
@@ -98,7 +98,7 @@ bᵢ(x, z) = x > L/2 ? 0.01 : 0.06
 set!(model, b=bᵢ)
 
 
-# ## Defining Simulation Timestep
+# ## Construct a Simulation
 #
 # Fast wave speeds make the equations stiff, so the CFL condition restricts the timestep to
 # adequately small values to maintain numerical stability.
@@ -108,8 +108,9 @@ set!(model, b=bᵢ)
 stop_time = 5hours
 simulation = Simulation(model; Δt, stop_time)
 
-# The TimeStepWizard helps ensure stable time-stepping with a Courant-Freidrichs-Lewy (CFL)
-# number of 0.35.
+# The [`TimeStepWizard`](@ref) is incorporated in the simulation via the
+# [`conjure_time_step_wizard!`](@ref) helper function and it ensures stable
+# time-stepping with a Courant-Freidrichs-Lewy (CFL) number of 0.35.
 conjure_time_step_wizard!(simulation, cfl=0.35)
 
 


examples/lock_exchange.jl
```diff
@@ -0,0 +1,63 @@
...
+z_disc  = MutableVerticalDiscretization((-50, 0))
+
+# Initialize the grid 
+underlying_grid = RectilinearGrid(
+                    size = (Nx, Nz),
+                    x = (0, L),
+                    z = z_disc,
+                    topology = (Bounded, Flat, Bounded), 
+                    halo = (5, 5)
+)
```
glwagner:
```suggestion
# Allow for mutable surface height 
z  = MutableVerticalDiscretization((-50, 0))
x = (0, L)

# Initialize the grid 
underlying_grid = RectilinearGrid(size = (Nx, Nz); halo = (5, 5), x, z, topology = (Bounded, Flat, Bounded))
```

examples/lock_exchange.jl
```diff
@@ -0,0 +1,63 @@
+using Oceananigans
```
glwagner, line 35: Can we add a description of what this example hopes to achieve, similar to the other examples?

examples/lock_exchange.jl
```diff
@@ -0,0 +1,63 @@
...
+using Oceananigans.Units
+using Oceananigans.Grids: MutableVerticalDiscretization
+
+
```
glwagner, line 47:
```suggestion
```

examples/lock_exchange.jl
```diff
@@ -0,0 +1,63 @@
...
+bᵢ(x, z) = x > 4kilometers ? 0.06 : 0.01
+set!(model, b=bᵢ)
+
+# Set the timesteps 
+Δt = 5minutes
+stop_time = 3days
+simulation = Simulation(model; Δt, stop_time)
+
+# Run simulation
+run!(simulation)
```
glwagner: The example should conclude with a visualization of the results, including some analysis / physical interpretation of what the results show

examples/lock_exchange.jl
```diff
@@ -0,0 +1,63 @@
...
+    # Closures simplify complex systems by approximating the effects of unresolved scales
+    # ZStarCoordinate allows for the top of the grid to move with the free surface
+    # Runge Kutta good for integrating multiple processes 
+
+model = HydrostaticFreeSurfaceModel(; grid,
+    tracers = (:b,),
+    buoyancy = BuoyancyTracer(),
+    momentum_advection = WENO(order=5), 
+    tracer_advection = WENO(order=7), 
+    closure = (VerticalScalarDiffusivity(ν=1e-4), HorizontalScalarDiffusivity(ν=1.0)),
```
glwagner:
```suggestion
```

I suggest running this as an ILES unless that doesn't work for some reason!

examples/lock_exchange.jl
```diff
@@ -0,0 +1,63 @@
...
+# Initialize the model 
+    # Want to use a hydrostatic model 
+    # Tracers act as markers within the fluid to track movement and dispersion
+    # Weighted Essentially Non-Oscillatory (WENO) methods are useful for capturing sharp changes in density
+    # Closures simplify complex systems by approximating the effects of unresolved scales
+    # ZStarCoordinate allows for the top of the grid to move with the free surface
+    # Runge Kutta good for integrating multiple processes 
+
+model = HydrostaticFreeSurfaceModel(; grid,
+    tracers = (:b,),
```
glwagner:
```suggestion
    tracers = :b,
```

examples/lock_exchange.jl
```diff
@@ -0,0 +1,63 @@
...
+
+# Set initial conditions for lock exchange with different densities 
+bᵢ(x, z) = x > 4kilometers ? 0.06 : 0.01
+set!(model, b=bᵢ)
+
+# Set the timesteps 
+Δt = 5minutes
+stop_time = 3days
+simulation = Simulation(model; Δt, stop_time)
+
```
glwagner, line 110: Can we add a progress function and output, similar to other examples?

surbhidhiman:
@glwagner Thank you very much for your helpful feedback. I've been working on adding some more documentation as well as visualization for the outputs of the simulation. I have attached the animation that is produced here. I plan to continue to add some helpful documentation to explain some of the numerical methods or physics behind the remaining code that is not as commented, and I will commit that as soon as I can. Of the code that is now added, I pulled strongly from some of the existing examples, but please let me know if there is anything else that you would like to see or to change. 

Looking at this result of the simulation, can I ask if the parameters make sense for this use case or if there is too much noise? Also, I ran simulation for 3 days, but would you recommend a different runtime? 
Thank you again for all your help and consideration! 


https://github.com/user-attachments/assets/746bbeb1-77c8-498e-b210-31dca355a876

examples/lock_exchange.jl
surbhidhiman: I have added some more descriptive documentation here and I plan to add some more as well.

examples/lock_exchange.jl
surbhidhiman: For some reason, I am getting an error when I remove these lines and don't import some methods explicitly. May I ask if you have seen that before?

examples/lock_exchange.jl
surbhidhiman: Thank you! I have removed the closure.

examples/lock_exchange.jl
surbhidhiman: I have added a progress function and output visualization which I have attached in another comment below.

surbhidhiman:
Hello @francispoulin @simone-silvestri, may I ask if I can also request a review from you for this PR? I have previously discussed this idea in this [discussion post](https://github.com/CliMA/Oceananigans.jl/discussions/4794) and [issue](https://github.com/CliMA/Oceananigans.jl/issues/4898).

examples/lock_exchange.jl
glwagner: You should not have to import any names explicitly that are not exported at the top-level!

This comment follows a suggestion I made just to remove some white space though.

examples/lock_exchange.jl
```diff
@@ -0,0 +1,284 @@
...
+model = HydrostaticFreeSurfaceModel(; grid,
+    tracers = :b,      
+    buoyancy = BuoyancyTracer(),
+    momentum_advection = WENO(order=5), 
+    tracer_advection = WENO(order=7), 
+    vertical_coordinate = ZStarCoordinate(grid), 
+    free_surface = SplitExplicitFreeSurface(grid; substeps=10), 
+    timestepper = :SplitRungeKutta3 
+)
```
glwagner: I think you need a vertical mixing closure here as well. You can use `CATKEVerticalDiffusivity` or `TKEDissipationVerticalDiffusivity`. (It would be interesting to know if there are any differences between these two, as well).

examples/lock_exchange.jl
```diff
@@ -0,0 +1,284 @@
...
+stop_time = 3days 
+simulation = Simulation(model; Δt, stop_time)
+
+# The TimeStepWizard helps ensure stable time-stepping with a (CFL) number of 0.7. Since the stability region 
+# of the numerical time stepper extends only up to CFL ≈ 1. , we keep the CFL condition at a conservative value 
+# of 0.7 to ensure robust and stable time stepping.
+
+wizard = TimeStepWizard(cfl=0.7, max_change=1.1, max_Δt=1.0)
+simulation.callbacks[:wizard] = Callback(wizard, IterationInterval(10))
+
```
glwagner:
```suggestion
conjure_time_step_wizard!(simulation, cfl=0.7)
```

curious why the max dt?

examples/lock_exchange.jl
```diff
@@ -0,0 +1,284 @@
...
+@info "Animation created in lock_exchange.mp4"
+
+
+# The visualization shows the time evolution of buoyancy b, stratification N², and the
+# velocity fields u′ and w. Initially, the two water masses are separated horizontally
+# by a sharp density interface. As the flow evolves, gravity currents form and the dense
+# fluid moves beneath the lighter fluid. This shows the characteristic transition from
+# horizontal to vertical density separation for the lock exchange problem. 
+
+# ![](lock_exchange.mp4)
```
glwagner:
```suggestion
# ![](lock_exchange.mp4)

```

examples/lock_exchange.jl
```diff
@@ -0,0 +1,284 @@
...
+
+# Set resolution of the simulation grid 
+Nx, Nz = 256, 64
+
+# Set grid size 
+L = 8kilometers   # horizontal length
+H = 50meters      # depth  
+
+# Allow for mutable surface height 
+z_disc  = MutableVerticalDiscretization((-50, 0))
```
glwagner:
```suggestion
z  = MutableVerticalDiscretization((-50, 0))
```

examples/lock_exchange.jl
```diff
@@ -0,0 +1,284 @@
...
+#       boundary conditions are applied as that dimension is not represented in the simulation, while a 
+#       bounded topology represents represents a physical boundary 
+
+underlying_grid = RectilinearGrid(
+                    size = (Nx, Nz),
+                    halo = (5, 5),
+                    x = x,
+                    z = z_disc,
+                    topology = (Bounded, Flat, Bounded)
+)
```
glwagner:
```suggestion
underlying_grid = RectilinearGrid(size=(Nx, Nz); x, z, halo=(5, 5), topology=(Bounded, Flat, Bounded))
```

glwagner:
> @glwagner Thank you very much for your helpful feedback. I've been working on adding some more documentation as well as visualization for the outputs of the simulation. I have attached the animation that is produced here. I plan to continue to add some helpful documentation to explain some of the numerical methods or physics behind the remaining code that is not as commented, and I will commit that as soon as I can. Of the code that is now added, I pulled strongly from some of the existing examples, but please let me know if there is anything else that you would like to see or to change.
> 
> Looking at this result of the simulation, can I ask if the parameters make sense for this use case or if there is too much noise? Also, I ran simulation for 3 days, but would you recommend a different runtime? Thank you again for all your help and consideration!
> 
>  lock_exchange.mp4

Excellent work!

The simulation is interesting. This is a hydrostatic model, so it cannot properly resolve shear instability (but the solution seems to show signs of it). I suggest the following general ideas:

- I think we can get away with coarser resolution
- You definitely need a vertical mixing closure like `closure = CATKEVerticalDiffusivity()`. This will reduce noise, curious to see other effects.
- I don't think we need to run the simulation so long... perhaps just until the initial wave propagates to one side or perhaps slightly longer to see the rebound effects)

Note the example needs to be _very_ cheap -- it will be run literally every single commit made to Oceananigans. There are a lot of those. That's why examples are very critical. 

@simone-silvestri does adaptive time-stepping work for this case? Would be nice if it does.

examples/lock_exchange.jl
francispoulin: Any reason why you are using a cfl parameter of 0.7?  I have heard that 0.3 is enough for stability, I don't know about 0.7.  But if your results are stable, then I pressume that is enough for this case.

examples/lock_exchange.jl
```diff
@@ -0,0 +1,284 @@
...
+
+# ## Run Simulation
+
+run!(simulation)
+@info "Simulation finished. Output saved to $(filename)"
+
+
+# ## Load Saved TimeSeries Values 
+
+u_t  = FieldTimeSeries(filename, "u")
```
francispoulin: u_t makes me think if a time derivative of us, which is not the case.  This is just a time series.  I wonder whether u_series or u_data or something else might be clearer?

examples/lock_exchange.jl
surbhidhiman: As for the max_dt and for the clf parameter of 0.7, I don't have an intuitive reason for those choices. I mainly used those since I saw it used in a different example.  
I will remove the max_dt. For the CLF value, I can try to run it with a value of 0.3 and see what the output is like.

francispoulin: Greg is suggesting an addaptive time stepping, and as long as it is stable then that should be enough.

examples/lock_exchange.jl
glwagner: In most of the examples we use `ut`. Perhaps slightly better without the underscore?

glwagner: Either way we should be consistent.

examples/lock_exchange.jl
```diff
@@ -0,0 +1,284 @@
...
+# using Pkg
+# pkg"add Oceananigans, CairoMakie"
+# ```
+
+
+# ## Import Required Packages 
+
+using Oceananigans
+using Oceananigans.Units: kilometers, meters, seconds, minutes, days, hour
+using Oceananigans.Grids: MutableVerticalDiscretization
```
francispoulin:
```suggestion
using Oceananigans.Grids
```

This worked fine for me.

examples/lock_exchange.jl
surbhidhiman: Thank you! I was running into some instability when I made the edits to remove max dt and adjust clf value, so I have added them back in but increased the max dt slightly.

surbhidhiman:
> > @glwagner Thank you very much for your helpful feedback. I've been working on adding some more documentation as well as visualization for the outputs of the simulation. I have attached the animation that is produced here. I plan to continue to add some helpful documentation to explain some of the numerical methods or physics behind the remaining code that is not as commented, and I will commit that as soon as I can. Of the code that is now added, I pulled strongly from some of the existing examples, but please let me know if there is anything else that you would like to see or to change.
> > Looking at this result of the simulation, can I ask if the parameters make sense for this use case or if there is too much noise? Also, I ran simulation for 3 days, but would you recommend a different runtime? Thank you again for all your help and consideration!
> > lock_exchange.mp4
> 
> Excellent work!
> 
> The simulation is interesting. This is a hydrostatic model, so it cannot properly resolve shear instability (but the solution seems to show signs of it). I suggest the following general ideas:
> 
> * I think we can get away with coarser resolution
> * You definitely need a vertical mixing closure like `closure = CATKEVerticalDiffusivity()`. This will reduce noise, curious to see other effects.
> * I don't think we need to run the simulation so long... perhaps just until the initial wave propagates to one side or perhaps slightly longer to see the rebound effects)
> 
> Note the example needs to be _very_ cheap -- it will be run literally every single commit made to Oceananigans. There are a lot of those. That's why examples are very critical.
> 
> @simone-silvestri does adaptive time-stepping work for this case? Would be nice if it does.

Thank you again for all the helpful feedback! I have made the recommended changes, please let me know if there is anything else I should adjust. Running this example locally on my macbook air with the M2 chip takes only a couple minutes, and the simulation run took somewhere around a minute. This is the output that is produced :) 


https://github.com/user-attachments/assets/2015cb3e-7e22-4b7f-b65e-8bd401d74421

examples/lock_exchange.jl
glwagner: Why do we need this at all? What name is missing?

examples/lock_exchange.jl
```diff
@@ -0,0 +1,281 @@
...
+using Oceananigans.Operators
+
+using Printf
+using CairoMakie
+
+
+# ## Set up 2D Rectilinear Grid
+
+# Set resolution of the simulation grid 
+Nx, Nz = 128, 32
```
glwagner:
```suggestion
Nx, Nz = 128, 64
```

examples/lock_exchange.jl
```diff
@@ -0,0 +1,281 @@
...
+
+
+# Add a slope at the bottom of the grid 
+h_left = -H
+h_right = -25meters
+slope = (h_right - h_left) / L
+bottom(x) = h_left + slope * x
+
+# Use an immersed boundary with grid fitted bottom to describe the sloped bottom of the domain 
+grid = ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(bottom))
```
glwagner:
```suggestion
grid = ImmersedBoundaryGrid(underlying_grid, PartialCellBottom(bottom))
```

glwagner:
Can you plot the TKE `e` instead of `w`?

surbhidhiman:
> Can you plot the TKE `e` instead of `w`?

@glwagner Got it! I have the adjusted visuals here and I have committed the other changes as well. 

I have an unrelated question about the failing tests for the buildkite/oceananigans-distributed and buildkite/oceananigans-distributed/pipeline - is the issue is related to my changes or is there something upstream in the testing pipeline? I’m happy to make any changes if needed. Thank you! 


https://github.com/user-attachments/assets/6bf5e441-b583-4ad4-9474-db7fe0584831

glwagner:
@surbhidhiman the failures are unrelated. The test pipeline is a bit fragile sadly.

Note, you have added a file to the repo (yay!) but it doesn't actually change any of the tests or docs! For that, we have to make it so that your code runs in CI. (We will get to that in a few days probably).

@giordano probably wants to make it so that tests do not run unless the files they pertain to are touched. But this takes some work. It's easier if you build a system that does it from the get-go (and we did not do this when we set up testing many years ago, unfortunately).

simone-silvestri:
> @simone-silvestri does adaptive time-stepping work for this case? Would be nice if it does.

Indeed, it should work also in this case.

giordano:
> @giordano probably wants to make it so that tests do not run unless the files they pertain to are touched. But this takes some work. It's easier if you build a system that does it from the get-go (and we did not do this when we set up testing many years ago, unfortunately).

We do something like that in Yggdrasil, and we use buildkite over there, so we have a template to follow, but the main problem is formalising the logic "the files they pertain to". If you only edit test files that's easy, but if you edit any piece of the source code you probably want to run the entire test suite. Is that still useful to do the filtering at that point?

glwagner:
> > @giordano probably wants to make it so that tests do not run unless the files they pertain to are touched. But this takes some work. It's easier if you build a system that does it from the get-go (and we did not do this when we set up testing many years ago, unfortunately).
> 
> We do something like that in Yggdrasil, and we use buildkite over there, so we have a template to follow, but the main problem is formalising the logic "the files they pertain to". If you only edit test files that's easy, but if you edit any piece of the source code you probably want to run the entire test suite. Is that still useful to do the filtering at that point?

Yeah maybe its not worth the effort, there are PRs that don't touch code (like here) but they aren't common.

glwagner:
@surbhidhiman the example is still running too long. We want a slow motion video of the initial stages of the relaxation, but we don't really care about all the sloshing that happens after

surbhidhiman:
> @surbhidhiman the example is still running too long. We want a slow motion video of the initial stages of the relaxation, but we don't really care about all the sloshing that happens after

@glwagner Got it! How does this look? I can make any other adjustments as needed before pushing the changes. 

https://github.com/user-attachments/assets/5ac5ade6-3c4b-4231-af7b-267d18330b12

glwagner:
> > @surbhidhiman the example is still running too long. We want a slow motion video of the initial stages of the relaxation, but we don't really care about all the sloshing that happens after
> 
> @glwagner Got it! How does this look? I can make any other adjustments as needed before pushing the changes.
> 
>  lock_exchange.mp4

can yuou save data more frequently so the animation is smoother?

surbhidhiman:
> > > @surbhidhiman the example is still running too long. We want a slow motion video of the initial stages of the relaxation, but we don't really care about all the sloshing that happens after
> > 
> > 
> > @glwagner Got it! How does this look? I can make any other adjustments as needed before pushing the changes.
> > lock_exchange.mp4
> 
> can yuou save data more frequently so the animation is smoother?

Yup! I changed save_interval = 10minutes from 30minutes previously and this is the result: 

https://github.com/user-attachments/assets/6e6857f3-bcf2-468e-bb5e-d56928f8f101

glwagner:
how about 2minutes?

examples/lock_exchange.jl
```diff
@@ -0,0 +1,287 @@
...
+# hydrostatic Boussinesq equations: 
+# ```math
+# \frac{\partial \rho}{\partial t} + \nabla \cdot (\rho \textbf{v}) = 0
+# ```
+# where `ρ` represents density. In the case of this model, we use boyancy as a tracer, which relates to the density: 
+# ```math
+# b = -g \frac{\rho - \rho_{0}}{\rho_{0}}
+# ```
+# 
+# [Source URL](https://www.sciencedirect.com/science/article/abs/pii/S0093641322000842)
```
navidcy: let's add a proper BibTeX citation?

glwagner:
@surbhidhiman this is looking really great. do you mind if me and/or @navidcy commit to your branch? We will add the example to the CI pipeline, and then we can generate a preview

examples/lock_exchange.jl
```diff
@@ -0,0 +1,287 @@
...
+# Initialize the grid: 
+#  For the RectilinearGrid, the different parameters represent: 
+#  * size: N dimensional tuple to set number of cells in the x and z directions 
+#  * halo: Padding of grid cells used to exchange boundary information between cells 
+#  * x: Physical size of grid representation in the x direction 
+#  * z: Physical size of grid representation in the z direction. Since MutableVerticalDiscretization is applied 
+#       to z, it allows for the height of the surface to be adjusted during the simulation 
+#  * topology: Describes the boundary structure in each (x, y, z) direction; a flat topology means that no 
+#       boundary conditions are applied as that dimension is not represented in the simulation, while a 
+#       bounded topology represents represents a physical boundary
```
navidcy: We don't need to explain so much for the grid since we have a whole Docs/Grids section.
Let's explain what's new. Perhaps just what the MutableVerticalDiscretization is?

examples/lock_exchange.jl
```diff
@@ -0,0 +1,287 @@
...
+
+model = HydrostaticFreeSurfaceModel(; grid,
+    tracers = (:b, :e),      
+    buoyancy = BuoyancyTracer(),
+    closure = CATKEVerticalDiffusivity(),
+    momentum_advection = WENO(order=5), 
+    tracer_advection = WENO(order=7), 
+    vertical_coordinate = ZStarCoordinate(grid), 
+    free_surface = SplitExplicitFreeSurface(grid; substeps=10), 
+    timestepper = :SplitRungeKutta3
```
navidcy:
```suggestion
```
removing since it's the default, isn't it?

examples/lock_exchange.jl
```diff
@@ -0,0 +1,287 @@
...
+#  * ZStarCoordinate method allows for the top of the grid to move with the free surface
+#  * Runge Kutta method is good for integrating multiple processes 
+
+model = HydrostaticFreeSurfaceModel(; grid,
+    tracers = (:b, :e),      
+    buoyancy = BuoyancyTracer(),
+    closure = CATKEVerticalDiffusivity(),
+    momentum_advection = WENO(order=5), 
+    tracer_advection = WENO(order=7), 
+    vertical_coordinate = ZStarCoordinate(grid),
```
navidcy: isn't this automatically inferred from the grid's z coordinate type?

examples/lock_exchange.jl
```diff
@@ -0,0 +1,287 @@
...
+Δt = 1seconds 
+stop_time = 20hours
+simulation = Simulation(model; Δt, stop_time)
+
+# The TimeStepWizard helps ensure stable time-stepping with a (CFL) number of 0.7. Since the stability region 
+# of the numerical time stepper extends only up to CFL ≈ 1, we keep the CFL condition at a value 
+# of 0.7 to ensure robust and stable time stepping.
+
+clf_value = 0.7
+wizard = TimeStepWizard(cfl=clf_value, max_change=1.1, max_Δt=5.0)
```
navidcy:
```suggestion
wizard = TimeStepWizard(cfl=0.7, max_change=1.1, max_Δt=5.0)
```

examples/lock_exchange.jl
```diff
@@ -0,0 +1,287 @@
...
+#  * u′: deviation of u from average of horizontal u
+#  * w: z component velocity (vertical)
+#  * N²: buoyancy stratification
+
+b = model.tracers.b
+e = model.tracers.e 
+u, v, w = model.velocities
+
+U = Field(Average(u)) 
+u′ = u - U
```
navidcy: why do we subtract the mean U here? does it make sense in this example or is it leftover from the internal tide one?

examples/lock_exchange.jl
```diff
@@ -0,0 +1,287 @@
...
+hm_u = heatmap!(ax_u, u′ₙ; nan_color = :black,
+                colorrange = (-umax, umax), colormap = :magma)
+Colorbar(fig[4, 2], hm_u, label = "m s⁻¹")
+
+ax_N2 = Axis(fig[5, 1]; title = "N² (stratification)", axis_kwargs...)
+hm_N2 = heatmap!(ax_N2, N²ₙ; nan_color = :black, 
+                colorrange = (-0.25N2max, N2max), colormap = :magma)
+Colorbar(fig[5, 2], hm_N2, label = "s⁻²")
+
+display(fig)
```
navidcy: display won't work with CairoMakie; just `fig` will do the job

```suggestion
fig
```

navidcy:
> @surbhidhiman this is looking really great. do you mind if me and/or @navidcy commit to your branch? We will add the example to the CI pipeline, and then we can generate a preview

@surbhidhiman sorry, I [pushed](https://github.com/CliMA/Oceananigans.jl/pull/4968/commits/bbcefb90723b26d7619e523e8d75e17c2cd5bbac) already... oooops! Feel free to undo!

examples/lock_exchange.jl
glwagner: agree

examples/lock_exchange.jl
glwagner: i thought so too. we need to hear from @simone-silvestri on this!

examples/lock_exchange.jl
glwagner: I recommend not putting in max change or max dt. I doubt it is needed, and misleads people reading the code into thinking they need it for their code, too. also the max_change is the default

examples/lock_exchange.jl
glwagner: i dont think that makes sense for this example

examples/lock_exchange.jl
surbhidhiman: @navidcy @glwagner Question about this, from the docs (https://clima.github.io/OceananigansDocumentation/stable/models/models_overview/#Using-the-HydrostaticFreeSurfaceModel) I understand that the 'QuasiAdamsBashforth2TimeStepper' is the default. Am I misunderstanding and should I remove the `timestepper = :SplitRungeKutta3` line anyway?

surbhidhiman:
> > @surbhidhiman this is looking really great. do you mind if me and/or @navidcy commit to your branch? We will add the example to the CI pipeline, and then we can generate a preview
> 
> @surbhidhiman sorry, I [pushed](https://github.com/CliMA/Oceananigans.jl/pull/4968/commits/bbcefb90723b26d7619e523e8d75e17c2cd5bbac) already... oooops! Feel free to undo!

@glwagner @navidcy Thanks again for all your helpful comments, please feel free to make any commits necessary! 

I believe I've made most of the recommended changes in the comments and committed them here. This is the result of the simulation I got below. Please let me know if there is anything else I can change or clean up - thanks again! 


https://github.com/user-attachments/assets/8cb0635c-7288-4afa-9a08-1448a7f92ccf

examples/lock_exchange.jl
navidcy: you are right! my bad..
the default is not the RK3; so leave the line as is!

examples/lock_exchange.jl
```diff
@@ -0,0 +1,281 @@
...
+
+# ## Initialize the Model 
+#
+#  * Want to use a hydrostatic model since horizontal motion may be more significant than verticle motion 
+#  * Tracers act as markers within the fluid to track movement and dispersion
+#  * Weighted Essentially Non-Oscillatory (WENO) methods are useful for capturing sharp changes in density
+#  * ZStarCoordinate method allows for the top of the grid to move with the free surface
+#  * Runge Kutta method is good for integrating multiple processes 
+
+model = HydrostaticFreeSurfaceModel(; grid,
```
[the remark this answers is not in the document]
francispoulin: Out of curiosity, would it be possible to switch this of hydrostatic to nonhydrostatic to see if the results are the same?

navidcy (approved):

navidcy:
If the docs built successfully I'll merge.

Thanks @surbhidhiman!

examples/lock_exchange.jl
```diff
@@ -0,0 +1,237 @@
...
+# temperature, salinity, etc.) that are separated by a ‘lock’ at time ``t=0``, and we
+# calculate the evolution of how these fluids interact as time progresses.
+# This lock exchange implementation can be a representation of scenarios where water of
+# different salinities or temperatures meet and form sharp density gradients.
+# For example, in estuaries or in the Denmark Strait overflow.
+# Solutions of this problem describe how the fluids interact with each other as time evolves
+# and can be described by hydrostatic Boussinesq equations. In this example, we use buoyancy
+# as a tracer; see the [Boussinesq approximation section](@ref boussinesq_approximation) for
+# more details on the Boussinesq approximation. Also, for a detailed explanation of the
+# Boussinesq formulation for buoyant flows, see [Barletta2022Boussinesq](@citet).
```
navidcy: @surbhidhiman I tweaked the text here. I removed the density equation; seemed a bit too detailed. Also the text mentioned hydrostatic but the equation was just the density equation and thus it seems a bit asymmetrical. 

Let me know if you think I missed something in contracting the text here. Thanks!

surbhidhiman: Thank you @navidcy!

glwagner: I suggest removing the reference about the Boussinesq approximation, which I think is overreach for an example (the entire codebase is based on the Boussinesq approximation, not only this example)

examples/lock_exchange.jl
```diff
@@ -175,7 +175,7 @@ n = Observable(1)
 
 title = @lift @sprintf("t = %5.2f hours", times[$n] / hour)
 
-wₙ = @lift wt[$n]
+uₙ = @lift ut[$n]
```
glwagner:
```suggestion
un = @lift ut[$n]
```

I don't like the look of the _n subscripts here but up to you if you want to keep it.

navidcy:
welcome to Oceananigans @surbhidhiman!

PR #4971, opened 2025-11-23
"For developers" section in the documentation + document AbstractModel interface

glwagner:
This PR adds a new "for developers" section in the docs. Inside this docs section we have the contributors guide, as well as the new `AGENTS.md` file (which is copied over inside `make.jl`). I also added a new section on the `AbstractModel` interface which includes an example of implementing a simple model that integrates the Lorenz equations. cc @bgroenks96 . 

This PR is a WIP but I hope to keep working on it. Agents wrote the model interface section so it probably needs to be improved. Feel free to jump in and commit anything.

### DIFF
--- a/docs/src/developer_docs/model_interface.md
+++ b/docs/src/developer_docs/model_interface.md
@@ -354,15 +354,12 @@ We initialize the model with a perturbed state and use a `JLD2Writer` to
 save the solution at regular intervals:
 
 ```@example model_interface
-
-# Create a 1D periodic grid
-grid = RectilinearGrid(size=128, x=(0, 32π), topology=(Periodic, Flat, Flat), halo=4)
+grid = RectilinearGrid(size=256, x=(0, 32π), topology=(Periodic, Flat, Flat), halo=4)
 ks_model = KuramotoSivashinskyModel(grid)
 
 # Initialize with a combination of sinusoidal modes
 set!(ks_model.solution, x -> cos(x/16) * (1 + sin(x/16)))
-
-simulation = Simulation(ks_model; Δt=0.1, stop_time=500, verbose=false)
+simulation = Simulation(ks_model; Δt=0.002, stop_time=60)
 
 simulation.output_writers[:solution] = JLD2Writer(ks_model, (; u=ks_model.solution),
                                                   filename = "ks_solution.jld2",
@@ -396,8 +393,8 @@ ylims!(ax, -4, 4)
 # lines! works directly with Field
 lines!(ax, u_n; linewidth=2, color=:royalblue)
 
-record(fig, "ks_animation.mp4", eachindex(times); framerate=30) do i
-    n[] = i
+record(fig, "ks_animation.mp4", eachindex(times); framerate=12) do nn
+    n[] = nn
 end
 nothing # hide
 ```


glwagner:
@bgroenks96 this PR is ready to review if you want to take a look. I made it so that models do not have to define timestepper.

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+
+When `run!(sim::Simulation)` is called the following high-level sequence occurs:
+
+1. `initialize!(sim.model)` prepares the model state,
+   etc.) and `update_state!(sim.model)` computes any auxiliary tendencies.
+
+2. Time-stepping begins. For every time step, Simulation computes an aligned `Δt`, gathers callbacks
+   that should run inside the model (`ModelCallsite`s), and calls
+   `time_step!(sim.model, Δt; callbacks=model_callbacks)`. Most models will also call `update_state!`
+   at the end of `time_step!`. This ensures that the diagnostic state is current with the prognostic state
```
bgroenks96: auxiliary state? to avoid confusion with simulation diagnostics?

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+   so that output and callbacks can execute correctly on a fully consistent model state.
+
+3. After `time_step!(model, Δt)`, `Simulation` executes diagnostics, output
```
bgroenks96: here I assume that diagnostics refers to those defined by the simulation

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+   writers, and callbacks scheduled on the `TimeStepCallsite`.
+
+Because `Simulation` assumes this protocol, any custom `AbstractModel` should
+implement (or inherit sane fallbacks for) the items listed below.
+
+## Structure and extensions of an `AbstractModel`
+
+### Required properties
+
+- `model.grid`: Simulation uses the grid to determine `architecture(model)`
```
bgroenks96: What was the reasoning here? I thought you said before that not all models should have grids?

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+  resets it via `reset_clock!(model)` when `reset!(sim)` is called.
+
+### Lifecycle hooks
+
+- `update_state!(model, callbacks=[]; compute_tendencies=true)`: invoked by
+  Simulation right after `initialize!` and inside most time steppers. This is
+  where models fill halos, update boundary conditions, recompute auxiliary
+  fields, and run callbacks with an `UpdateStateCallsite`. Implementations
+  typically finish by calling `compute_tendencies!(model, callbacks)` so that
+  any `TendencyCallsite` callbacks can modify tendencies before integration.
```
bgroenks96: is there documentation for this? can we link to it here?

glwagner:
I don't think there's anything controversial here so I will merge when tests pass. Much more work is needed for the developer docs; for example I would also like to add an example of implementing a new turbulence closure.

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+
+- `initialize!(model::AbstractModel)`: called exactly once per `run!` before the
+  first time step.
+
+### Optional integrations
+
+While not required for Simulation itself, the following methods enable the rest
+of the Oceananigans ecosystem to “see” the model:
+
+- `fields(model)` and `prognostic_fields(model)`: return `NamedTuple`s of
```
bgroenks96, line 106: does `fields` include also prognostic fields? or just auxiliary?

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+  `iteration(model)`. Simulation uses it for stop criteria and logging, and
+  resets it via `reset_clock!(model)` when `reset!(sim)` is called.
+
+### Lifecycle hooks
+
+- `update_state!(model, callbacks=[]; compute_tendencies=true)`: invoked by
+  Simulation right after `initialize!` and inside most time steppers. This is
+  where models fill halos, update boundary conditions, recompute auxiliary
+  fields, and run callbacks with an `UpdateStateCallsite`. Implementations
+  typically finish by calling `compute_tendencies!(model, callbacks)` so that
```
bgroenks96: is `compute_tendencies!` a mandatory model method?

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+  and `eltype(model)` via the fallbacks `architecture(model) = model.grid.architecture`
+  and `eltype(model) = eltype(model.grid)`.
+
+- `model.clock :: Clock`: the source of truth for `time(model)` and
+  `iteration(model)`. Simulation uses it for stop criteria and logging, and
+  resets it via `reset_clock!(model)` when `reset!(sim)` is called.
+
+### Lifecycle hooks
+
+- `update_state!(model, callbacks=[]; compute_tendencies=true)`: invoked by
```
bgroenks96, line 65: do callbacks have a particular expected type or pattern?

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+using Oceananigans.TimeSteppers: Clock, tick!
+import Oceananigans.TimeSteppers: update_state!, time_step!
+import Oceananigans.Fields: set!
+using Oceananigans: TendencyCallsite, UpdateStateCallsite
+
+mutable struct LorenzModel{G, C, P, S} <: AbstractModel{Nothing, Nothing}
+    grid :: G
+    clock :: C
+    parameters :: P
+    state :: S
```
bgroenks96, line 160: this looks a bit Terrarium-like, was that on purpose? :P 

I thought the recommended Oceananigans pattern was to have state variables directly defined as fields on the model struct?

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+
+function set!(model::LorenzModel; kw...)
+    for (name, val) in kw
+        set!(model.state[name], val)
+    end
+    return nothing
+end
+
+function update_state!(model::LorenzModel, callbacks = []; compute_tendencies=true)
+    for callback in callbacks
```
bgroenks96: is this loop always present? could we define a wrapper dispatch for `update_state!` that relieves the downstream implemetnations of always needing to write this?

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+        callback.callsite isa UpdateStateCallsite && callback(model)
+    end
+    return nothing
+end
+
+function time_step!(model::LorenzModel, Δt; callbacks = ())
+    model.clock.iteration == 0 && update_state!(model, callbacks; compute_tendencies = false)
+
+    for callback in callbacks
```
bgroenks96: same thing here

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+### Using the model inside Simulation
+
+We set up a `Callback` to record the trajectory at each time step:
+
+```@example model_interface
+lorenz = LorenzModel()
+set!(lorenz, x=1)
+simulation = Simulation(lorenz; Δt=0.01, stop_time=100, verbose=false)
+
+trajectory = NTuple{3, Float64}[]
```
bgroenks96, line 212: is this the recommended way to do this generally? I had already thought about whether we should have an `OutputWriter` that saves things in memory

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+
+While not required for Simulation itself, the following methods enable the rest
+of the Oceananigans ecosystem to “see” the model:
+
+- `fields(model)` and `prognostic_fields(model)`: return `NamedTuple`s of
+  fields so diagnostics, output writers, and NaN checkers know what to touch.
+
+- `default_nan_checker(model)`: customize the `NaNChecker` that Simulation adds
+  by default.
+
```
bgroenks96, line 116: what about `timestepper`? this is optional but nonetheless relevant for some models, no?

glwagner:
I'll address these! Note if you request changes I cannot merge until you approve again!

docs/src/developer_docs/model_interface.md
glwagner: yes, it refers to `simulation.diagnostics`

docs/src/developer_docs/model_interface.md
glwagner:
```suggestion
   at the end of `time_step!`. This ensures that the auxiliary state (including halo regions, closure auxiliary fields, etc) is current with the prognostic state
```

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,227 @@
...
+1. `initialize!(sim.model)` prepares the model state,
+   etc.) and `update_state!(sim.model)` computes any auxiliary tendencies.
+
+2. Time-stepping begins. For every time step, Simulation computes an aligned `Δt`, gathers callbacks
+   that should run inside the model (`ModelCallsite`s), and calls
+   `time_step!(sim.model, Δt; callbacks=model_callbacks)`. Most models will also call `update_state!`
+   at the end of `time_step!`. This ensures that the auxiliary state (including halo regions, closure auxiliary fields, etc) is current with the prognostic state
+   so that output and callbacks can execute correctly on a fully consistent model state.
+
+3. After `time_step!(model, Δt)`, `Simulation` executes diagnostics, output
```
glwagner:
```suggestion
3. After `time_step!(model, Δt)`, `Simulation` executes its diagnostics, output
```

docs/src/developer_docs/model_interface.md
glwagner: that's true. let me look at this.

docs/src/developer_docs/model_interface.md
glwagner: callbacks must be functions of simulation, eg `func(simulation)`. They are documented here: https://clima.github.io/OceananigansDocumentation/stable/simulations/simulations_overview/

docs/src/developer_docs/model_interface.md
glwagner: no.

docs/src/developer_docs/model_interface.md
glwagner: There is documentation here: https://clima.github.io/OceananigansDocumentation/stable/appendix/library/#Oceananigans.Simulations.Callback

Note that the way callbacks are executed via `TendencyCallsite` and `UpdateStateCallsite` is actually a property of the model design. Only `TimeStepCallsite` is owned by `Simulation`.

docs/src/developer_docs/model_interface.md
glwagner: By convention, `fields` should include all of the fields that are available within discrete-form boundary conditions. By convention, this should be a superset of the prognostic fields (sometimes diagnostic fields are also useful for users)

docs/src/developer_docs/model_interface.md
glwagner: a model should implement `timestepper(model)` if there is also a `reset!(timestepper)` function. This only matters if using the `reset!` feature.

docs/src/developer_docs/model_interface.md
glwagner: Ok, it isn't actually needed.

docs/src/developer_docs/model_interface.md
glwagner: That is possible but outside the scope of this PR because it would be a significant refactor for all models. But note that this form for `update_state!` is actually only needed internally within a model --- because it is called by `time_step!(model, dt)`. I believe when `update_state!` is called by simulation, there is no second argument.

docs/src/developer_docs/model_interface.md
glwagner: The problem here is that tendencies may be computed anywhere inside `time_step!`. The idea here is to give users a hook to modify tendencies after they are computed, so the model implementer must put this loop in the right place. The only way to relieve this boilerplate would be to develop a whole infrastructure for time-steppers. I think it is possible but I am not sure the benefits would justify the effort?

glwagner: we can write a function:

```julia
execute_tendency_callbacks!(callbacks) = # etc
```

which reduces this to one line.

docs/src/developer_docs/model_interface.md
glwagner: output writers should be used for `Field` because this hooks into FieldTimeSeries infrastructure. For lightweight diagnostics though, its not necessary to save to disk.

glwagner:
I think we should implement two examples here. One very small model example, LorenzModel, which does not use Field or have a grid. But we should also have a second model implementation that uses Field. This will answer some of @bgroenks96 's questions about output writers and model design principles.

docs/src/developer_docs/model_interface.md
glwagner: there are diverse implementations. but few models have state variables as fields on the model struct.

glwagner: most models are too complicated to bundle every state variable into a single struct; or at least if this were done, it would hinder usability. most complicated models break up the state into different objects ("velocities", "tracers", "closure fields")

glwagner:
@bgroenks96 I addressed your comments and added a second model example (KuramotoSivashinskyModel) to illustrate the difference between using Field or not in a model implementation.

docs/src/developer_docs/model_interface.md
bgroenks96: couldn't you also make `compute_tendencies!` formally part of the interface? or is that an assumption that we don't want to make for all models?

docs/src/developer_docs/model_interface.md
```diff
@@ -0,0 +1,412 @@
...
+using Oceananigans.Models: AbstractModel
+using Oceananigans.Simulations: Simulation, run!
+using Oceananigans.TimeSteppers: Clock, tick!
+using Oceananigans: TendencyCallsite, UpdateStateCallsite
+
+import Oceananigans.TimeSteppers: update_state!, time_step!
+
+mutable struct LorenzModel{FT} <: AbstractModel{Nothing, Nothing}
+    clock :: Clock{FT}
+    σ :: FT
```
bgroenks96: I actually liked it a bit better the way it was before :see_no_evil:, I was only teasing, I didn't mean to suggest that you had to change it :sweat_smile: 

but it's also fine to keep it like this, it's a simple model anyway.

glwagner: heh. I'll merge as-is if tests pass. do you mean, you prefer the design where there is a `state` property, which is a NamedTuple?

bgroenks96: yes and with a separate `parameters` named tuple, but both patterns are fine, up to you.

glwagner: done

PR #4976, opened 2025-11-24
Add adapt method for AMD plus closure coeff for functions

glwagner:
This is needed to use arrays for the closure coefficient on GPU, or functions like the example illustrates (the doctest example does build, but cannot be time-stepped).

Should probably test this.


navidcy (approved):
yes, a test would be lovely

PR #4978, opened 2025-11-24
fjp/fix oscillating flow.jl

francispoulin:
I tried the oscillating_flow.jl example in the validation folder and it didn't run.

I worked on fixing it and came up with this.  The newer names make it much easier.  Thanks to everyone who did that!

Unfortunately, the names of the mp4s have quotes and that should probably be changed.

When people are happy with this change, I am happy to work on the other two.


francispoulin:
Sorry, this is all wrong.  Will fix it.

PR #4979, opened 2025-11-24
Run Aqua tests

giordano:
This is only a beginning, and I'll need help from others, there are literally hundreds of issues to address.  For example https://github.com/CliMA/Oceananigans.jl/blob/81dc6c5db02ba111ffc750999e3ca83402dfebe3/src/Operators/spacings_and_areas_and_volumes.jl#L390-L436 defines some hundreds of methods which expect non-existing metrics to be there, while also exporting them:
```julia-repl
julia> using Oceananigans

julia> grid = RectilinearGrid(size=(4, 4, 4), extent=(1, 1, 1));

julia> Oceananigans.Operators.Az(1, 1, 1, grid, Nothing(), Center(), Face())
ERROR: UndefVarError: `Azᵃᶜᶠ` not defined in `Oceananigans.Operators`
```
`Oceananigans.Operators.Azᵃᶜᶠ` is even exported, same for all the other metrics involving `Nothing`.  Do I understand correctly `Nothing` should just be removed from the lists?

See https://juliatesting.github.io/Aqua.jl/stable/ for how to run the tests locally, for example I started with
```julia-repl
julia> using Aqua, Oceananigans

julia> Aqua.test_all(Oceananigans; ambiguities=false)
```
I skipped the ambiguities at the beginning, because there are 358 of them (yes, you read that right)

PR #4985, opened 2025-11-25
Start running ExplicitImports checks

giordano:
This isn't complete, but like #4979 this is a beginning for more extensive linting checks.  The tests introduced here should pass locally for me (so in principle this PR is good to go as is), but I haven't checked I haven't broken anything in the process, I guess CI will tell us 🙂

I know the PR looks a bit intrusive, but there's really lots of cruft in the code, these checks (together with #4979 and future related PRs) are aimed at making Oceananigans more maintainable in the long run.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -142,6 +142,9 @@ serve(dir="docs/build")
 - Add references to papers from the literature by adding bibtex to `oceananigans.bib`, and then
   a corresponding citation
 - Make use of cross-references with equations
+- In example code, use explicit imports as sparingly as possible. NEVER explicitly import a name that
+  is already exported by the user interface. Always rely on `using Oceananigans` for imports and keep
+  imports clean. Explicit imports should only be used for source code.
 
 ### Writing examples
 - Explain at the top of the file what a simulation is doing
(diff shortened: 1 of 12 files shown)


glwagner:
The changes look great so far!

src/Models/HydrostaticFreeSurfaceModels/hydrostatic_free_surface_model.jl
```diff
@@ -69,7 +67,7 @@ default_free_surface(grid; gravitational_acceleration=defaults.gravitational_acc
                                 clock = Clock{Float64}(time = 0),
                                 momentum_advection = VectorInvariant(),
                                 tracer_advection = Centered(),
-                                buoyancy = SeawaterBuoyancy(eltype(grid)),
+                                buoyancy = nothing,
```
giordano, line 70: This is how you discover the docstring was stale: `SeawaterBuoyancy` is actually unused in this module 🙃

giordano:
CI now seems to be quite happy this is ready for review 🙂

giordano:
I'll mention that this may be slightly breaking for downstream consumers which were importing functions from wrong modules and relying on the fact that other modules were unnecessarily importing those symbols somewhere else.  Using `ExplicitImports` also downstream, as we do in [`Breeze.jl`](https://github.com/NumericalEarth/Breeze.jl) already, would help a lot to clean up the mess

Issue #4812, opened 2025-09-24
Different behavior of `AdvectiveForcing` when tracers encounter a domain boundary and an immersed boundary


tomchor:
When using `AdvectiveForcing` I noticed that when an advected tracer encounters a domain boundary it goes right through it, but when encountering an immersed boundary it accumulates on that boundary. Here's an animated example of a tracer that has a settling velocity in a quiescent 2D domain. The simulation on the left has a flat immersed boundary, while the simulation on the right does not:

https://github.com/user-attachments/assets/98754ce0-7862-4c35-9937-94af72a779f1

Here's the script that I used to generate this: https://pastebin.com/U2jyuGzK

Is this the intended behavior? I'd expect both cases to behave similarly.

simone-silvestri:
I think the difference comes from the fact that we do not fill velocity boundary conditions in advective forcing, so we have a velocity at the boundary, while we mask out immersed fluxes with the `conditional_flux` function. If you pass a velocity field that has correctly filled boundary conditions (i.e. zero at the boundary), you should recover the same behavior (if I have to guess, your advective velocity is a `ConstantField` which does not have BCs.

Now, the question is: do we want to enforce this by always producing a field and filling the halos? If yes I can open a PR.

simone-silvestri:
just as a check, try passing a field that is 
```julia
w₀=ZFaceField(grid; boundary_conditions = FieldBoundaryConditions(grid, (Center(), Center(), Face()); bottom = OpenBoundaryCondition())
set!(w₀, -0.02)
fill_halo_regions!(w₀)
```
and most likely the behavior is the same.

tomchor:
I couldn't make `AdvectiveForcing` work with a `Field`. 

> Now, the question is: do we want to enforce this by always producing a field and filling the halos? If yes I can open a PR.

I do think whatever behavior we'd like to have should be consistent across domain boundaries and immersed boundaries. I think ideally we'd be able to change the behavior with a flag passed on `AdvectiveForcing`. I can see both being useful.

glwagner:
`AdvectiveForcing` is supposed to work with `Field`. Can you open an issue / MWE if that does not work? We have used code like this for biogeochemistry (for this very purpose) and it is supposed to work. I am surprised we do not have a test, but we should.

It is important to support both boundary conditions for `AdvectiveForcing`. One application for "penetrable" boundaries is to allow settling particles to "leave" the domain when they reach the bottom.

[...]

PR #4987, opened 2025-11-25
Allow arbitrary fluxes across immersed boundaries and test this with `AdvectiveForcing`

tomchor:
This PR makes it possible for users to enforce both open and closed boundaries for `AdvectiveForcing` for both immersed and domain boundaries. This is done by removing methods that enforced zero-flux through immersed boundaries. I'm also adding a test for this and a validation script which produces the following animations for open and closed boundaries, respectively:

https://github.com/user-attachments/assets/457feb2e-f5d6-4f69-a02c-bf6ee2970b7c

https://github.com/user-attachments/assets/8dc8844c-6627-484f-9c28-95db7653102e

I also fixed a few errors/typos in other validation scripts.

~Importantly, even after this PR, if a user passes a `Number` to `AdvectiveForcing` the behavior at domain and immersed boundaries remains different. Therefore this does *not* close https://github.com/CliMA/Oceananigans.jl/issues/4812, since a `Field` with proper boundary conditions is required to unify behavior across different boundaries.~

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -107,6 +107,8 @@ the features they describe! Also, if you have developed a new feature in Oceanan
 
 If you have work using Oceananigans that you would like to have listed here, please open a pull request to add it or let us know!
 
+1. Lee, X. K., Ramadhan, A., Souza, A., Wagner, G. L., Silvestri, S., Marshall, J. and Ferrari, R. (2025) [NORi: An ML-augmented ocean boundary layer parameterization](https://doi.org/10.48550/arXiv.2512.04452), _arXiv preprint_, arXiv:2512.04452. DOI: [10.48550/arXiv.2512.04452](https://doi.org/10.48550/arXiv.2512.04452)
+
 1. Brenner, S., Thompson, A., and Gupta, M. (2025) [Patterns of sea ice floes shape ocean turbulence in the marginal ice zone](https://doi.org/10.21203/rs.3.rs-8099873/v1
 ), preprint (Version 1), available at Research Square. DOI: [10.21203/rs.3.rs-8099873/v1](https://doi.org/10.21203/rs.3.rs-8099873/v1)
 
--- a/src/AbstractOperations/binary_operations.jl
+++ b/src/AbstractOperations/binary_operations.jl
@@ -115,7 +115,7 @@ function define_binary_operator(op)
         end
 
         # Numbers are not fields...
-        $op(Lc::Tuple, a::Number, b::Number) = $op(a, b)
+        $op(Lc::Tuple{<:$Location, <:$Location, <:$Location}, a::Number, b::Number) = $op(a, b)
 
         # Sugar for mixing in functions of (x, y, z)
         $op(Lc::Tuple{<:$Location, <:$Location, <:$Location}, f::Function, b::AbstractField) = $op(Lc, FunctionField(location(b), f, b.grid), b)


src/Forcings/advective_forcing.jl
```diff
@@ -1,10 +1,18 @@
...
 using Oceananigans.Utils: sum_of_velocities
+using Oceananigans.BoundaryConditions: OpenBoundaryCondition, FieldBoundaryConditions
+using Oceananigans.ImmersedBoundaries: ImmersedBoundaryCondition
 using Adapt
 
-maybe_constant_field(u) = u
-maybe_constant_field(u::Number) = ConstantField(u)
+maybe_field(u, args...) = u
+
+function maybe_field(u::Number, loc, grid, u_bcs)
```
glwagner: is this different from the function `field`?

tomchor: It is if `u` isn't a Number. (For example and array or a `ConstantField`.)
There's a slight difference for numbers as well since `u_bcs` isn't ` FieldBoundaryConditions`. (I'm trying to keep the interface simple and avoid having the user pass a whole set of `FieldBoundaryConditions` when creating a simple slip velocity forcing.

glwagner: Hmm... is it all simpler if we just convert `Number` to `Field` and then enforce boundary conditions correctly? (Sorry if I am misundersatnding something)

The other question is whether we can generalize `field` to encompass this case.

tomchor: > Hmm... is it all simpler if we just convert `Number` to `Field` and then enforce boundary conditions correctly? (Sorry if I am misundersatnding something)

Don't you need to pass boundary conditions to the `Field` when creating it? Because if so, that's what this code does (I simplified it a bit more if you wanna take another look).

> The other question is whether we can generalize `field` to encompass this case.

That'd definitely simplify things in this particular instance. But I think we'd need to create a `Field(f::AbstractField, args...) = f` method. Would such method ignore boundary conditions, indices, etc. that are passed as `args`? Or would it "copy" `f` and change the it according to `args`? I don't think any of these options is great to be honest. What do you think?

glwagner: > Don't you need to pass boundary conditions to the Field when creating it? Because if so, that's what this code does (I simplified it a bit more if you wanna take another look).

Yes, `Field` need boundary conditions if they are not default. For velocities, I would:

1. Use default velocity boundary conditions if users give a Number (eg impenetrable)
2. Also support explicitly providing `Field` which bundle boundary conditions

tomchor:
Re-centering and updating the discussion around face boundary conditions for `AdvectiveForcing` here. Per https://github.com/CliMA/Oceananigans.jl/issues/4998#issuecomment-3592754986, @glwagner suggested that the BCs for `AdvectiveForcing` be imposed by masking the immersed values to the appropriate value.

This by itself doesn't seem to work. For example, in the example below, `AdvectiveForcing.w` is equal to -0.02 everywhere, including in the immersed solid and domain faces. I'd expect both cases to result in the tracer passing through the boundaries, but instead the tracer accumulates in the immersed boundary:

https://github.com/user-attachments/assets/72b80ba5-ffda-4eac-b5b4-c76a2467cacd

@glwagner did I miss something about your suggestion?

glwagner:
> Re-centering and updating the discussion around face boundary conditions for `AdvectiveForcing` here. Per [#4998 (comment)](https://github.com/CliMA/Oceananigans.jl/issues/4998#issuecomment-3592754986), @glwagner suggested that the BCs for `AdvectiveForcing` be imposed by masking the immersed values to the appropriate value.
> 
> This by itself doesn't seem to work. For example, in the example below, `AdvectiveForcing.w` is equal to -0.02 everywhere, including in the immersed solid and domain faces. I'd expect both cases to result in the tracer passing through the boundaries, but instead the tracer accumulates in the immersed boundary:
> 
>  nonhydrostatic_tracer_circle_2d_compare.mp4 
> @glwagner did I miss something about your suggestion?

You didn't miss anything. This reveals that there is more work to do. Try commenting out these specializations for immersed boundary grid:

https://github.com/CliMA/Oceananigans.jl/blob/d23257c9c9b0bce70923b2e5cdb809885188c02b/src/Advection/immersed_advective_fluxes.jl#L57-L59

these seem to hardcode impenetrability for an immersed boundary grid.

@simone-silvestri may want to comment on this. I think we have discussed about how the conditional advection operators are redundant with masking. If users want to have open immersed boundaries, we will need to stop using conditional advection operators that assume impenetrability on immersed boundaries.

simone-silvestri:
I guess we could just remove these. The flux on the immersed boundaries will be zeroed by masking the velocities.

tomchor:
This does seem to work. Let's see if removing this doesn't break any tests.

src/Advection/immersed_advective_fluxes.jl
```diff
@@ -54,10 +54,6 @@ end
 @inline _advective_momentum_flux_Vw(i, j, k, ibg::IBG, scheme, V, w) = conditional_flux_cff(i, j, k, ibg, zero(ibg), advective_momentum_flux_Vw(i, j, k, ibg, scheme, V, w))
 @inline _advective_momentum_flux_Ww(i, j, k, ibg::IBG, scheme, W, w) = conditional_flux_ccc(i, j, k, ibg, zero(ibg), advective_momentum_flux_Ww(i, j, k, ibg, scheme, W, w))
 
-@inline _advective_tracer_flux_x(i, j, k, ibg::IBG, scheme, U, c) = conditional_flux_fcc(i, j, k, ibg, zero(ibg), advective_tracer_flux_x(i, j, k, ibg, scheme, U, c))
-@inline _advective_tracer_flux_y(i, j, k, ibg::IBG, scheme, V, c) = conditional_flux_cfc(i, j, k, ibg, zero(ibg), advective_tracer_flux_y(i, j, k, ibg, scheme, V, c))
-@inline _advective_tracer_flux_z(i, j, k, ibg::IBG, scheme, W, c) = conditional_flux_ccf(i, j, k, ibg, zero(ibg), advective_tracer_flux_z(i, j, k, ibg, scheme, W, c))
```
glwagner, line 59: I would suggest also removing the momentum fluxes for consistency, although it is not necessary for AdvectiveForcing

simone-silvestri:
I think the momentum fluxes might require a bit more work, since momentum fluxes always require interpolation of both advecting and advected velocities, and we do not have three-dimensional interpolation operators yet, so the velocity interpolated at the face might not be zero, leading to a non-zero flux across the immersed boundary.

src/Advection/immersed_advective_fluxes.jl
tomchor: Based on @simone-silvestri 's comment maybe we should leave this for another PR then?

src/Forcings/advective_forcing.jl
```diff
@@ -1,9 +1,26 @@
...
+maybe_field(u::AbstractField, args...) = u
+
+function maybe_field(u::Number, loc, grid, u_bcs, open_boundaries)
+    immersed_bcs = ImmersedBoundaryCondition(; u_bcs...)
+    bcs = FieldBoundaryConditions(grid, loc; u_bcs..., immersed = immersed_bcs)
+    u_field = Field(loc, grid; boundary_conditions = bcs)
+    set!(u_field, u)
+    fill_halo_regions!(u_field)
+    if open_boundaries
+        mask_immersed_field!(u_field, u)
```
simone-silvestri: Wouldn't this fill `u` over all the immersed boundaries?

tomchor: Yes, but since the interior of the immersed solid isn't used in any calculation I think that's okay, no?

simone-silvestri: I mean, if there is a complex boundary, all the boundaries end up advecting? Both west-facing and east-facing, for example. Isn't this a bit too much of an edge case to enforce with just a single keyword argument?

glwagner: There's only advection across faces that are normal to the velocity component

glwagner: I don't think we should use a keyword argument though, or anything specific to `AdvectiveForcing` at all. I think we should use regular boundary conditions and require users to build fields with those boundary conditions. We then should modify the `Field` constructor so that, if users have requested open immersed boundary conditions, the immersed region is masked.

For this to work generally we will have to fill every time step. But we can support constant ones to start.

glwagner: happy to help with this @tomchor if what I said isn't very clear. The first thing to fix is the issue you raised that Face ImmersedBoundaryCondition are not respected.

This also addresses @simone-silvestri's point because the ImmersedBoundaryCondition is applied to specific facets of the cells.

tomchor: I leaned towards having a boolean keyword argument specifically because `ImmersedBoundaryCondition` isn't respected. I think otherwise the interface becomes an little more unclear. It feels odd to me if a user passes complete `FieldBoundaryConditions` and then the velocities appear masked. There's still the option for the user to pass a `Field` with proper boundary conditions by the way.

> The first thing to fix is the issue you raised that Face ImmersedBoundaryCondition are not respected.

This was already fixed with your suggestion! Now we only gotta discuss the interface.

tomchor:
For the record, as it is, this code has the desired effect (which is also tested in `test_forcings.jl`). This is the result of a simulation with open boundaries:

https://github.com/user-attachments/assets/a51dfe70-be10-407e-9190-d1b9e01f68ab

And the result for closed boundaries:

https://github.com/user-attachments/assets/17155ebd-df6f-4220-9abc-80ff584ec104

src/Forcings/advective_forcing.jl
glwagner: The problem with this solution is that it is not general. We want to have the same interface for boundary conditions for AdvectiveForcing and resolved velocity fields. I don't think we want a special interface just for forcing.

simone-silvestri: I agree that we should probably solve this issue with a PR that modifies the field constructor (and mask immersed field) rather than advective forcing.

tomchor: Yeah the current solution is not general. I do agree with @simone-silvestri that if we want to have something general, then I think this should be another PR dedicated to modifying the field constructor.

src/Forcings/advective_forcing.jl
```diff
@@ -38,17 +63,38 @@ R = mean_particle_radius
...
+
+    if open_boundaries
+        u_bc = OpenBoundaryCondition(u)
+        v_bc = OpenBoundaryCondition(v)
+        w_bc = OpenBoundaryCondition(w)
+    else
+        u_bc = OpenBoundaryCondition(nothing)
+        v_bc = OpenBoundaryCondition(nothing)
+        w_bc = OpenBoundaryCondition(nothing)
+    end
```
glwagner: what will happen if the user provides a field that has open boundary conditions embedded into it? Will this code work correctly?

test/test_forcings.jl
```diff
@@ -252,6 +252,69 @@ function seven_forcings(arch)
...
+        return model
+    end
+
+    # Create models
+    w_settle = -0.01
+    regular_model = build_settling_model(regular_grid, w_settle)
+    immersed_model = build_settling_model(immersed_grid, w_settle)
+
+    regular_initial_integral = Integral(regular_model.tracers.c) |> Field |> deepcopy
+    immersed_initial_integral = Integral(immersed_model.tracers.c) |> Field |> deepcopy
```
glwagner: since the `Integral` is used later, this will be easier to understand if the `Integral` fields are persisted and then used later, and the deepcopy is used to save the initial state which is compared to the later fields.

src/Forcings/advective_forcing.jl
tomchor: If the user provides a field then the field is used as-is.

test/test_forcings.jl
tomchor: Done!

src/Forcings/advective_forcing.jl
glwagner: ah ok great. So we can perhaps revert these changes to the `AdvectiveForcing` constructor and keep the contributions of this PR to the changes to the advection operators (a very significant contribution)! Then we can work on the user interface in a subsequent PR so that users don't have to manually call `mask_immersed_field!` and can use the boundary condition interface instead. What do you think?

glwagner: If I understand correctly what is needed now is something like

```julia
w = ZFaceField(grid)
parent(w) .= 1
```

then `w` will work if used in `AdvectiveForcing`?

simone-silvestri: I think so. In this case the tracer will pass through immersed boundaries since we have eliminated the `_advective_tracer_flux_x` methods.

glwagner: Honestly the above is probably simpler than building an ImmersedBoundaryCondition, if your goal is to have a constant velocity

glwagner: the boundary condition would be needed for time-dependent velocities (and possibly useful for spatially-varying as well)

tomchor: > If I understand correctly what is needed now is something like
> 
> ```julia
> w = ZFaceField(grid)
> parent(w) .= 1
> ```
> 
> then `w` will work if used in `AdvectiveForcing`?

Yes

tomchor: > ah ok great. So we can perhaps revert these changes to the `AdvectiveForcing` constructor and keep the contributions of this PR to the changes to the advection operators (a very significant contribution)! Then we can work on the user interface in a subsequent PR so that users don't have to manually call `mask_immersed_field!` and can use the boundary condition interface instead. What do you think?

Yeah, that sounds good to me!

docs/src/models/forcing_functions.md
```diff
@@ -294,20 +294,22 @@ As an example, consider a model for sediment settling at a constant rate:
 ```jldoctest
 using Oceananigans
 
+grid = RectilinearGrid(size=(1, 1, 1), extent=(1, 1, 1))
+
```
tomchor:
```suggestion
```

docs/src/models/forcing_functions.md
```diff
@@ -294,20 +294,22 @@ As an example, consider a model for sediment settling at a constant rate:
...
 
-sinking = AdvectiveForcing(w=w_sediment)
+sinking = AdvectiveForcing(w=w_sediment; grid)
 
 # output
 AdvectiveForcing:
 ├── u: ZeroField{Int64}
 ├── v: ZeroField{Int64}
-└── w: ConstantField(-0.00352102)
+└── w: 1×1×2 Field{Center, Center, Face} on RectilinearGrid on CPU
```
tomchor:
```suggestion
└── w: ConstantField(-0.00352102)
```

tomchor:
@glwagner @simone-silvestri if tests pass this should be ready (they passed locally for me).

src/Advection/immersed_advective_fluxes.jl
glwagner, line 59: can we also delete the momentum flux functions? I don't want to risk inconsistency between momentum and tracers.

test/test_forcings.jl
```diff
@@ -252,6 +253,88 @@ function seven_forcings(arch)
...
+function test_settling_tracer_comparison(arch; open_bottom=true)
+    """
+    Test that compares settling tracer simulations on regular vs immersed boundary grids.
+    Both should conserve tracer mass and have similar maximum values.
+    """
+
+    Nz = 16
+    Lz = 1
+
+    regular_grid = RectilinearGrid(arch, topology = (Flat, Flat, Bounded), size = (Nz,), z = (-Lz, 0))
```
glwagner:
```suggestion
    regular_grid = RectilinearGrid(arch, topology = (Flat, Flat, Bounded), size = tuple(Nz), z = (-Lz, 0))
```

glwagner: does `size=Nz` work? that would be preferred.

validation/advective_forcing/settling_tracer_blob.jl
```diff
@@ -0,0 +1,151 @@
...
+
+using NCDatasets
+using CairoMakie
+
+function build_grid(use_immersed_grid::Bool;
+    Nx = 16,
+    Nz = 16,
+    Lx = 1,
+    Lz = 1,
+    )
```
glwagner, line 16: hanging `)` violates what we put in AGENTS.md, but its ok here bc its just a validation case. just FYI

tomchor: Changed

test/test_forcings.jl
tomchor: Yup, changed

src/Advection/immersed_advective_fluxes.jl
tomchor: Let's see if tests pass with this change

tomchor: Tests pass! @glwagner  I think this should be gtg

test/test_forcings.jl
```diff
@@ -252,6 +253,88 @@ function seven_forcings(arch)
...
+    w_settle = -0.01
+    regular_model = build_settling_model(regular_grid, w_settle)
+    immersed_model = build_settling_model(immersed_grid, w_settle)
+
+    ∫c_regular = Integral(regular_model.tracers.c) |> Field
+    ∫c_immersed = Integral(immersed_model.tracers.c) |> Field
+
+    regular_initial_integral = ∫c_regular |> deepcopy
+    immersed_initial_integral = ∫c_immersed |> deepcopy
+    @test regular_initial_integral[] == immersed_initial_integral[]
```
glwagner, line 307: this incurs a scalar operation on GPU, but I guess we are globally allowing scalar ops, so its fine 😂

simone-silvestri:
I was looking at this PR again and I still think removing the momentum flux zeroing on immersed boundaries might still be wrong because of the fact that we do not have boundary-aware interpolation. Is this tested? How was it solved?

tomchor:
> I was looking at this PR again and I still think removing the momentum flux zeroing on immersed boundaries might still be wrong because of the fact that we do not have boundary-aware interpolation. Is this tested? How was it solved?

Is was tested with the added validation script and the added test, but that's about it.

simone-silvestri:
Would it be possible to add a test for all the advection schemes that should that independently of the internal solution, when computing the momentum flux on peripheral nodes it is zero?

tomchor:
@simone-silvestri added https://github.com/CliMA/Oceananigans.jl/pull/5402

Can you double check if that's what you meant?

PR #4990, opened 2025-11-26
Problems with TimeStepWizard and ImmersedBoundaryGrid on MetalGPU

### LINKED ISSUE #4969: Run HydrostaticFreeSurfaceModel with ImmersedBoundaryGrid on MetalGPU
Hi,

I've run into a couple of issues while using ```HydrostaticFreeSurfaceModel``` with ```ImmersedBoundaryGrid``` on MetalGPU.
First, using ```PartialCellBottom(depth)``` gives an error ```Reason: unsupported use of double value```. It seems the ```minimum_fractional_cell_height = FT(0.2)```should be specified explicitly
```julia
grid = ImmersedBoundaryGrid(
    underlying_grid,
    PartialCellBottom(depth; minimum_fractional_cell_height = FT(0.2));
    active_cells_map = false,
)
```
And second, the 
```julia
conjure_time_step_wizard!(
    sim,
    cfl = FT(0.6),
    min_change = FT(0.5),
    max_change = FT(1.1),
    min_Δt = FT(1second),
    max_Δt = FT(30seconds),
)
```
fails on MetalGPU. There are two Float64 somewhere
```
Reason: unsupported use of double value
Reason: unsupported use of double value
```
but I haven’t been able to track them down.

BTW, a first estimate of the speedup on a 128×128×32 grid: the GPU is about 20 times faster than a single CPU thread, or about 5 times faster than 6 CPU threads (performance cores on the Apple M3 Pro).

Thank you,

Alexander

alesok:
This PR adds two additional tests to test_metal.jl. The first one, MetalGPU: ImmersedBoundaryGrid, contains a few TODOs and generally works due to some workarounds. The second test, MetalGPU: TimeStepWizard, currently fails.

Fixes #4969


test/test_metal.jl
```diff
@@ -21,12 +21,12 @@ Oceananigans.defaults.FloatType = Float32
...
-                                        tracers = :b,
-                                        momentum_advection = WENO(order=5),
-                                        tracer_advection = WENO(order=5),
-                                        free_surface = SplitExplicitFreeSurface(grid; substeps=60))
+        coriolis=FPlane(latitude=45),
+        buoyancy=BuoyancyTracer(),
+        tracers=:b,
+        momentum_advection=WENO(order=5),
+        tracer_advection=WENO(order=5),
+        free_surface=SplitExplicitFreeSurface(grid; substeps=60))
```
giordano: @alesok thanks for your contribution, but there are a few unrelated changes, here and in `Project.toml`, can you please revert them and keep only the relevant changes (i.e. the new tests)

test/test_metal.jl
```diff
@@ -40,3 +40,127 @@ Oceananigans.defaults.FloatType = Float32
     @test time(simulation) == 3minutes
 end
 
+@testset "MetalGPU: ImmersedBoundaryGrid" begin
+    using Oceananigans
+    using Oceananigans.Units
+    using Oceananigans.BoundaryConditions: fill_halo_regions!
+    using Metal
```
glwagner: these should be placed up at the top of the file

test/test_metal.jl
```diff
@@ -40,3 +40,127 @@ Oceananigans.defaults.FloatType = Float32
...
+    H .= FT(0.0)
+    Δx = Lx / FT(Nx - 1)
+    Δy = Ly / FT(Ny - 1)
+    for i = 1:Nx
+        x = (i - 1) * Δx
+        for j = 1:Ny
+            y = (j - 1) * Δy
+            H[i, j] = depth(x, y)
+        end
+    end
```
glwagner: is it necessary to use this bathymetry? You could write a simple function like

```julia
bathymetry(x, y) = - FT(Lz / 2)
```

the advantage of something simple is that the test is easier to maintain and understand

test/test_metal.jl
```diff
@@ -40,3 +40,127 @@ Oceananigans.defaults.FloatType = Float32
...
+
+    sim = Simulation(model, Δt=5seconds, stop_iteration=20)
+
+    wizard = TimeStepWizard(cfl=FT(0.7), min_Δt=FT(1second), max_Δt=FT(15seconds))
+    # sim.callbacks[:wizard] = Callback(wizard, IterationInterval(5)) #TODO does not work with MetalGPU
+
+    run!(sim)
+
+    @test iteration(sim) == 20
+    @test time(sim) ≥ 100seconds
```
glwagner:
```suggestion
    simulation = Simulation(model, Δt=5seconds, stop_iteration=20)
    run!(simulation)

    @test iteration(simulation) == 20
    @test time(simulation) == 100seconds
```

test/test_metal.jl
```diff
@@ -40,3 +40,127 @@ Oceananigans.defaults.FloatType = Float32
...
+end
+
+@testset "MetalGPU: TimeStepWizard" begin
+    using Oceananigans
+    using Metal
+
+    #! format: off
+    arch = GPU(Metal.MetalBackend()); Oceananigans.defaults.FloatType = Float32
+    #! format: on
+    # arch = CPU()
```
glwagner:
```suggestion
    arch = GPU(Metal.MetalBackend())
    Oceananigans.defaults.FloatType = Float32
```

test/test_metal.jl
```diff
@@ -40,3 +40,127 @@ Oceananigans.defaults.FloatType = Float32
...
+    end
+
+    bathymetry = Field{Center,Center,Nothing}(underlying_grid)
+    set!(bathymetry, coalesce.(H, FT(0.0)))
+    fill_halo_regions!(bathymetry)
+
+    grid = ImmersedBoundaryGrid(
+        underlying_grid,
+        # PartialCellBottom(depth; minimum_fractional_cell_height = FT(0.2)); #TODO does not work in the @testset env
+        PartialCellBottom(bathymetry; minimum_fractional_cell_height=FT(0.2)); #TODO minimum_fractional_cell_height = FT(0.2) is required
```
glwagner: Ok, I didn't appreciate these comments when making my above suggestion. 

Does the error arise on line 70 here?

https://github.com/CliMA/Oceananigans.jl/blob/6b1c59bad589311774d1e9aeb8fb1fbbb4e3f126/src/ImmersedBoundaries/partial_cell_bottom.jl#L70

I see the issue with the minimum fractional cell height which should be easy to fix.

glwagner:
@alesok I think that https://github.com/CliMA/Oceananigans.jl/pull/4990/commits/63e22a4f36f51964a62f7e02e525e63abce1962c should address the issue with having to provide the minimum_fractional_cell_height explictly.

alesok:
Thank you, @glwagner. After a small correction in commit https://github.com/CliMA/Oceananigans.jl/pull/4990/commits/54ff7e3ee334d80714e735fc1e4f40af4339c79d
, the minimum_fractional_cell_height is no longer needed. I have significantly simplified the testset for ImmersedBoundaryGrid, and now there are no TODOs left!

test/test_metal.jl
alesok: I removed these lines. These things are provided in the include("dependencies_for_runtests.jl")

test/test_metal.jl
alesok: Actually use of the function ```depth```in ```PartialCellBottom(depth)``` now works after fix in the ```materialize_immersed_boundary```. So these lines are deleted.

test/test_metal.jl
alesok: Fine with me. We have a separate testset for the TimeStepWizard.

test/test_metal.jl
```diff
@@ -39,4 +40,96 @@ Oceananigans.defaults.FloatType = Float32
...
+        underlying_grid,
+        PartialCellBottom(depth);
+        active_cells_map=false,
+    )
+    @test eltype(grid) == Float32
+
+    cᴾ = FT(4184.0) # J K⁻¹ kg⁻¹, typical heat capacity for fresh water 
+    ρₐ = FT(1.0)    # kg m⁻³, 
+    ρ₀ = FT(1000.0) # kg m⁻³,
+    Qᵀ = Qʰ / (ρ₀ * cᴾ) # K m s⁻¹, surface _temperature_ flux
```
glwagner:
```suggestion
    Qᵀ = FT(0.01)
```

best to simplify the test. 

as a side note (possibly pointless since this may have been AI-generated, but I will say anyways), if one uses the standard TEOS10 equation of state, then the heat capacity _must_ be 3991.87 J / K-kg per the definition of conservative temperature. This is the value on page 147 of the TEOS10 manual: https://www.teos-10.org/pubs/TEOS-10_Manual.pdf

test/test_metal.jl
```diff
@@ -39,4 +40,96 @@ Oceananigans.defaults.FloatType = Float32
...
+    T_bcs = FieldBoundaryConditions(top=FluxBoundaryCondition(Qᵀ))
+
+    Cᴰ = FT(1.3e-3)
+    τₐ = ρₐ * Cᴰ * u10 * abs(u10)
+    Qᵘ = -τₐ / ρ₀ # m² s⁻²
```
glwagner:
```suggestion
    Qᵘ = - FT(1e-4)
```

glwagner:
> Thank you, @glwagner. After a small correction in commit [54ff7e3](https://github.com/CliMA/Oceananigans.jl/commit/54ff7e3ee334d80714e735fc1e4f40af4339c79d) , the minimum_fractional_cell_height is no longer needed. I have significantly simplified the testset for ImmersedBoundaryGrid, and now there are no TODOs left!

excellent! It looks like the test fails though, so I guess we have some puzzle to solve still, right?

alesok:
The MetalGPU: ImmersedBoundaryGrid test passes successfully on my Mac. However, the third test set, MetalGPU: TimeStepWizard, fails. We haven't touched it yet.

test/test_metal.jl
```diff
@@ -39,4 +40,94 @@ Oceananigans.defaults.FloatType = Float32
...
+    @inline function depth(x::FT, y::FT)::FT
+        a² = (FT(0.45) * Lx)^2  # semi-major squared axis
+        b² = (FT(0.45) * Ly)^2  # semi-minor squared axis
+        x₀ = FT(0.5) * Lx
+        y₀ = FT(0.5) * Ly
+        Δx = x - x₀
+        Δy = y - y₀
+        r² = (Δx * Δx) / a² + (Δy * Δy) / b²
+        return r² ≤ FT(1) ? -Lz * (FT(1) - r²) : FT(1)
+    end
```
glwagner:
```suggestion
    depth(x, y) = - Lz / 2
```

navidcy: I agree with the suggestion by @glwagner

test/test_metal.jl
navidcy: I agree

alesok: Sure, we can simplify the test, but I don't think it will make the code clearer. But, it is a matter of taste:)

test/test_metal.jl
alesok: Well, it might be useful to keep a test case with a land inside the modelling domain ...

glwagner: My suggestion keeps the land! It just simplifies the bathymetry (the complexity of the bathymetry is irrelevant for the test)

test/test_metal.jl
```diff
@@ -39,4 +40,91 @@ Oceananigans.defaults.FloatType = Float32
...
     @test time(simulation) == 3minutes
 end
+#! format: on
 
+@testset "MetalGPU: ImmersedBoundaryGrid" begin
+    arch = GPU(Metal.MetalBackend())
+    FT = Oceananigans.defaults.FloatType
+
+    Qʰ = FT(-400.0)  # W m⁻², surface _heat_ flux (negative value - downward!)
+    u10 = FT(5.0)    # Wind m/s
```
glwagner:
```suggestion
```

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
     @test time(simulation) == 3minutes
 end
+#! format: on
 
+@testset "MetalGPU: ImmersedBoundaryGrid" begin
+    arch = GPU(Metal.MetalBackend())
+    FT = Oceananigans.defaults.FloatType
+
+    Lx, Ly, Lz = FT(5000meters), FT(5000meters), FT(20meters)
+    Nx, Ny, Nz = 128, 128, 32
```
glwagner:
```suggestion

    Lx, Ly, Lz = FT(5000meters), FT(5000meters), FT(20meters)
    Nx, Ny, Nz = 16, 16, 8
```

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
+    model = HydrostaticFreeSurfaceModel(;
+        grid,
+        coriolis=FPlane(latitude=60),
+        tracers=(:T, :S),
+        buoyancy=SeawaterBuoyancy(),
+        momentum_advection=WENO(),
+        tracer_advection=WENO(),
+        free_surface=SplitExplicitFreeSurface(grid; substeps=30), # default does not work on MetalGPU
+        closure=ConvectiveAdjustmentVerticalDiffusivity(convective_κz=FT(1.0), background_κz=FT(1e-3)),
+        boundary_conditions=(u=u_bcs, T=T_bcs),
```
glwagner:
```suggestion
        grid,
        coriolis = FPlane(latitude=60),
        tracers = (:T, :S),
        buoyancy = SeawaterBuoyancy(),
        momentum_advection = WENO(),
        tracer_advection = WENO(),
        free_surface = SplitExplicitFreeSurface(grid; substeps=30), # default does not work on MetalGPU
        closure = ConvectiveAdjustmentVerticalDiffusivity(convective_κz=1, background_κz=1e-3),
        boundary_conditions = (u=u_bcs, T=T_bcs),
```

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
+
+    model = HydrostaticFreeSurfaceModel(;
+        grid,
+        coriolis = FPlane(latitude=60),
+        tracers = (:T, :S),
+        buoyancy = SeawaterBuoyancy(),
+        momentum_advection = WENO(),
+        tracer_advection = WENO(),
+        free_surface = SplitExplicitFreeSurface(grid; substeps=30), # default does not work on MetalGPU
+        closure = ConvectiveAdjustmentVerticalDiffusivity(convective_κz=1, background_κz=1e-3),
```
glwagner:
```suggestion
        closure = nothing, # ConvectiveAdjustmentVerticalDiffusivity(convective_κz=1, background_κz=1e-3),
```

glwagner:
FYI I am just experimenting -- I noticed that the TimeStepWizard step is actually passing, but the ImmersedBoundaryGrid test fails. However, this test adds a number of new features in addition to just ImmersedBoundaryGrid. Ultimately maybe we should run this test with and without the immersed boundary part.

alesok:
OK, I don't want to interrupt your work; just a small remark. I noticed that the calculation of Qᵀ was removed and replaced with the assignment Qᵀ = FT(0.01). This value seems quite large and, more importantly, appears to have the wrong sign. Since the initial temperature is zero, after one hour of simulation the temperature near the coast drops below –15. I believe a more appropriate value might be something like Qᵀ = -FT(9.560229e-5).

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
+    end
+
+    grid = ImmersedBoundaryGrid(
+        underlying_grid,
+        PartialCellBottom(depth);
+        active_cells_map=false,
+    )
+    @test eltype(grid) == Float32
+
+    Qᵀ = FT(0.01)
```
glwagner:
```suggestion
    Qᵀ = FT(1e-4)
```

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
+        closure = nothing, # ConvectiveAdjustmentVerticalDiffusivity(convective_κz=1, background_κz=1e-3),
+        boundary_conditions = (u=u_bcs, T=T_bcs),
+    )
+    @test parent(model.velocities.u) isa MtlArray
+    @test parent(model.velocities.v) isa MtlArray
+    @test parent(model.velocities.w) isa MtlArray
+    @test parent(model.tracers.T) isa MtlArray
+    @test parent(model.tracers.S) isa MtlArray
+
+    simulation = Simulation(model, Δt=5seconds, stop_iteration=20)
```
glwagner:
```suggestion
    simulation = Simulation(model, Δt=1seconds, stop_iteration=1)
```

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
+    @test parent(model.velocities.v) isa MtlArray
+    @test parent(model.velocities.w) isa MtlArray
+    @test parent(model.tracers.T) isa MtlArray
+    @test parent(model.tracers.S) isa MtlArray
+
+    simulation = Simulation(model, Δt=1seconds, stop_iteration=1)
+    run!(simulation)
+
+    @test iteration(simulation) == 20
+    @test time(simulation) == 100seconds
```
glwagner:
```suggestion
    @test iteration(simulation) == 1
    @test time(simulation) == 1seconds
```

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
+        closure = nothing, # ConvectiveAdjustmentVerticalDiffusivity(convective_κz=1, background_κz=1e-3),
+        boundary_conditions = (u=u_bcs, T=T_bcs),
+    )
+    @test parent(model.velocities.u) isa MtlArray
+    @test parent(model.velocities.v) isa MtlArray
+    @test parent(model.velocities.w) isa MtlArray
+    @test parent(model.tracers.T) isa MtlArray
+    @test parent(model.tracers.S) isa MtlArray
+
+    simulation = Simulation(model, Δt=1seconds, stop_iteration=1)
```
glwagner:
```suggestion
    simulation = Simulation(model, Δt=1, stop_iteration=1)
```

glwagner:
> OK, I don't want to interrupt your work; just a small remark. I noticed that the calculation of Qᵀ was removed and replaced with the assignment Qᵀ = FT(0.01). This value seems quite large and, more importantly, appears to have the wrong sign. Since the initial temperature is zero, after one hour of simulation the temperature near the coast drops below –15. I believe a more appropriate value might be something like Qᵀ = -FT(9.560229e-5).

No worries -- also you should feel free to commit / work on this as much as you like! I changed the value to 1e-4. Note, we only do 1 iteration, so the actual number is irrelevant (eg we could use "1" for simplicity...). For the test, we merely need to have the pattern right; the numbers do not matter.

test/test_metal.jl
```diff
@@ -10,6 +10,7 @@ Oceananigans.defaults.FloatType = Float32
 # * https://github.com/CliMA/Oceananigans.jl/pull/4124#discussion_r1976449272
 # * https://github.com/CliMA/Oceananigans.jl/pull/4152
 
+#! format: off
```
navidcy: what is this? `#! format: off/on`?

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
 
+@testset "MetalGPU: ImmersedBoundaryGrid" begin
+    arch = GPU(Metal.MetalBackend())
+    FT = Oceananigans.defaults.FloatType
+
+    Lx, Ly, Lz = FT(5000meters), FT(5000meters), FT(20meters)
+    Nx, Ny, Nz = 16, 16, 8
+
+    underlying_grid = RectilinearGrid(
+        arch;
```
navidcy: Not sure what's the intention here.

FT = Oceananigans.defaults.FloatType, but the previous test already changed that to Float32.

Are we meant to give Float64 and see if they are converted to Float32 when arch = GPU(Metal)? Or do we want to give Float32?

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
+        tracer_advection = WENO(),
+        free_surface = SplitExplicitFreeSurface(grid; substeps=30), # default does not work on MetalGPU
+        closure = nothing, # ConvectiveAdjustmentVerticalDiffusivity(convective_κz=1, background_κz=1e-3),
+        boundary_conditions = (u=u_bcs, T=T_bcs),
+    )
+    @test parent(model.velocities.u) isa MtlArray
+    @test parent(model.velocities.v) isa MtlArray
+    @test parent(model.velocities.w) isa MtlArray
+    @test parent(model.tracers.T) isa MtlArray
+    @test parent(model.tracers.S) isa MtlArray
```
navidcy:
```suggestion
    for field in merge(model.velocities, model.tracers)
        @test parent(field) isa MtlArray
    end
```

test/test_metal.jl
```diff
@@ -39,4 +40,88 @@ Oceananigans.defaults.FloatType = Float32
...
+    simulation = Simulation(model, Δt=1, stop_iteration=1)
+    run!(simulation)
+
+    @test iteration(simulation) == 1
+    @test time(simulation) == 1seconds
+end
+
+@testset "MetalGPU: TimeStepWizard" begin
+    arch = GPU(Metal.MetalBackend())
+    FT = Oceananigans.defaults.FloatType
```
navidcy: define FT but not used?

alesok: I am not sure it is needed but FT is used in 
```julia
wizard = TimeStepWizard(cfl=FT(0.7), min_Δt=FT(1), max_Δt=FT(15))
```

test/test_metal.jl
alesok: I think in previous test we set we have 
```julia
using Metal

Oceananigans.defaults.FloatType = Float32
```
but FT is not defined

test/test_metal.jl
alesok: Well, my JuliaFormatter a bit different and in order to keep the original style I turned it off for the first test

test/test_metal.jl
glwagner: Probably best to write `FT = Float32` somewhere, or even just hardcode Float32 here for clarity.

test/test_metal.jl
glwagner: `FT` should not be needed in any constructor. AI tools seem to want to pepper `FT` everywhere (we explicitliy write not to do this in AGENTS.md but it does not always seem to get into the AI's head)

alesok:
I suspect that the issue with using TimeStepWizard on MetalGPU is related to the execution of KernelFunctionOperation. Please review commit [5683c2e](https://github.com/CliMA/Oceananigans.jl/pull/4990/commits/5683c2ebd589c4144928d007fc0e495dabc9c109); I’ve just added two debug print statements.
```julia
function cell_advection_timescale(grid, velocities)
    u, v, w = velocities
    τ = KernelFunctionOperation{Center, Center, Center}(cell_advection_timescaleᶜᶜᶜ, grid, u, v, w)
    println("********** Hello 1")
    min_τ = minimum(τ) #TODO here is a problem on MetalGPU
    println("********** Hello 2")
    return min_τ
end
```
Thank you

src/Advection/cell_advection_timescale.jl
```diff
@@ -13,10 +13,7 @@ is the minimum over all `i, j, k` in the `grid` of
 function cell_advection_timescale(grid, velocities)
     u, v, w = velocities
     τ = KernelFunctionOperation{Center, Center, Center}(cell_advection_timescaleᶜᶜᶜ, grid, u, v, w)
-    println("********** Hello 1")
-    min_τ = minimum(τ) #TODO here is a problem on MetalGPU
-    println("********** Hello 2")
-    return min_τ
+    return minimum(τ) #TODO here is a problem on MetalGPU
```
glwagner: Ah very interesting. We should add a few small test for reductions on Metal then, something like

```julia

grid = RectilinearGrid(arch, size=(4, 4, 4), extent=(1, 1, 1))

# Test reduction of Field
c = CenterField(grid)
set!(c, 1)
@test minimum(c) == 1

# Test reduction of KernelFunctionOperation
add_2(i, j, k, grid, c) = @inbounds c[i, j, k] + 2
add_2_kfo = KernelFunctionOperation{Center, Center, Center}(add_2, grid, c)
@test minimum(add_2_kfo) == 3
```

alesok: Thank you. I will try. Actually I found some problem in fields.jl with masks (see  [0651518](https://github.com/CliMA/Oceananigans.jl/pull/4990/commits/065151876a416d4a00a1afbf74c589d66cfa2042) ). There were some notes
```julia
# TODO make this Float32 friendly
get_neutral_mask(::MinimumReduction) = +Inf
get_neutral_mask(::MaximumReduction) = -Inf
``` 
So I added some conversions, they works but may be sub-optimal. So, now the problem is here:
```julia
       function Base.$(reduction)(f::Function,
                                   c::AbstractField;
                                   condition = nothing,
                                   mask = get_neutral_mask(Base.$(reduction!)),
                                   dims = :)

            mask = convert(eltype(c), mask)
            conditioned_c = condition_operand(f, c, condition, mask)
            T = filltype(Base.$(reduction!), c)
            loc = reduced_location(instantiated_location(c); dims)
            r = Field(loc, c.grid, T; indices=indices(c))
            initialize_reduced_field!(Base.$(reduction!), identity, r, conditioned_c)
            println("Hi there")
            Base.$(reduction!)(identity, interior(r), conditioned_c, init=false) # Here we have a problem on MetalGPU
            println("Hello")
```

Thank you

alesok:
Well, the tests for reductions pass successfully.  See https://github.com/CliMA/Oceananigans.jl/pull/4990/commits/149a9a1b879933ee913c7da6c138a5682b390852

alesok:
I was too quick: tests for reductions work only for very small grids and fail otherwise
```julia
   arch = GPU(Metal.MetalBackend())
    # grid = RectilinearGrid(arch, size=(4, 4, 4), extent=(1, 1, 1)) # OK
    # grid = RectilinearGrid(arch, size=(6, 6, 6), extent=(1, 1, 1)) # OK
    # grid = RectilinearGrid(arch, size=(7, 7, 7), extent=(1, 1, 1)) # OK
    # grid = RectilinearGrid(arch, size=(8, 8, 8), extent=(1, 1, 1)) # OK
    grid = RectilinearGrid(arch, size=(9, 9, 9), extent=(1, 1, 1)) # Error
```

alesok:
The commit [dde00d1](https://github.com/CliMA/Oceananigans.jl/pull/4990/commits/dde00d16cb318a291ce912de122c586714acd6db) addresses some reduction issues on MetalGPU. I have successfully run test_metal on my Mac. Finally! The problem with reductions was that Metal does not support ReshapedArray and we need to materialise the view. Please check the implementation. Thank you

src/Fields/field.jl
```diff
@@ -734,6 +734,24 @@ const Identity = typeof(Base.identity)
...
+"""
+function safe_interior(r::AbstractField)
+    interior_r = interior(r)
+
+    # Only materialize if the field uses Float32 (MetalGPU) and interior_r is a wrapper
+    if eltype(r) == Float32 && parent(interior_r) !== interior_r
+        interior_r = copy(interior_r)
+    end
+    return interior_r
+end
```
glwagner: what's going on here?

in principle you could extend `interior` in the MetalGPU extension, if you did need to.

alesok: Metal does not support ReshapedArray objects produced by reshaping a view, but it does support reshaping a concrete array. Therefore, the view should be materialized first.

I am not very familiar with the overall architecture of Oceananigans, so there is likely a more elegant approach than checking eltype(r) == Float32 to apply this materialization specifically for Metal. The same applies to identifying the most appropriate location in the codebase for this modification.

simone-silvestri: you could do something like this:
```suggestion
safe_interior(r::AbstractField) = safe_interior(architecture(r), r)

# Extended in the OceananigansMetalExt for compatibility with Metal
safe_interior(arch, r) = interior(r)
```

then in the metal extension:
```julia
function FIelds.safe_interior(::MetalGPU, r::AbstractField)
    interior_r = interior(r)

    if parent(interior_r) !== interior_r
        interior_r = copy(interior_r)
    end
    return interior_r
end

Comment (alesok) [reply src/Fields/field.jl]: Thank you, [simone-silvestri](https://github.com/simone-silvestri). Your suggestion is implemented in the [2162215](https://github.com/CliMA/Oceananigans.jl/pull/4990/commits/21622152f610805d4a6e19566a8b667832f49fd2) and it works as expected on my Mac.

Comment (glwagner) [reply src/Fields/field.jl]: There isn't a need for `safe_interior` because we can merely extend `interior` itself

Comment (glwagner) [reply src/Fields/field.jl]: You had me fooled --- you seem quite familiar!

Functions are specialized to types by [multiple dispatch](https://docs.julialang.org/en/v1/manual/methods/). A very simple example:

```julia
f(x::Int) = "I am Int"
f(x::Float64) = "I am float"
```

when you call `f(x)`, the _method_ which is invoked depends on its type.

A Field on MetalGPU has a specific type, which you can leverage to produce specialized behavior. It's because the architecture is contained within the grid.

```julia
const MetalGrid = AbstractGrid{<:Any, <:Any, <:Any, <:Any, <:MetalGPU}
const MetalField = Field{<:Any, <:Any, <:Any, <:Any, <:MetalGrid}

function interior(metal_field::MetalField)
    # etc
end
```

The new method should be implemented in the metal extension:

https://github.com/CliMA/Oceananigans.jl/blob/main/ext/OceananigansMetalExt.jl

@simone-silvestri has a different suggestion, but I am not sure that `safe_interior` is needed. It would better if not, because the code is simpler.

alesok: How to make MetalField visible in both OceananigansMetalExt.jl and field.jl? Where to place these two lines?
```julia
const MetalGrid = AbstractGrid{<:Any, <:Any, <:Any, <:Any, <:MetalGPU}
const MetalField = Field{<:Any, <:Any, <:Any, <:Any, <:MetalGrid}
```

glwagner: You cannot make `Metal` visible in `field.jl`. All metal-specific methods should be implemented in the extension. What are you trying to express?

alesok: I defined 
```julia
const MetalGrid = GD.AbstractGrid{<:Any, <:Any, <:Any, <:Any, <:MetalGPU}
const MetalField = FD.Field{<:Any, <:Any, <:Any, <:Any, <:MetalGrid}

function FD.interior(::MetalGPU, r::MetalField)
    interior_r = interior(r.data, location(r), r.grid, r.indices)
    
    if parent(interior_r) !== interior_r
        interior_r = copy(interior_r)
    end
    return interior_r
end
``` 
in the OceananigansMetalExt.jl and now trying to call it from the field.jl. That what I meant about familiarity with the code base.

glwagner: I don't mean to nitpick I just want to say this because it could be important in the future -- the concept we're wrestling with has to do with how dispatch works, and is not specific to the Oceananigans codebase.

We define methods in `field.jl` that are generic, such as

```julia
function interior(f::Field) end
```

In the extension, we create a _more specific_ method, which gets filed under the name `interior`:

```julia
function FD.interior(::MetalField)
```

Now when you are running code, and you call `interior`, the compiler will search through all of the methods attached to the name `interior`. It will choose the most specific one for the case at hand. If you are using a `MetalField`, and you have successfully created a more specific method for it, then it will call that function --- not the one in `field.jl`.

This is a process of writing generic code in `src`, and then specializing it to more specific cases in the extensions, and it applies to all julia packages.

glwagner: this could help:

https://docs.julialang.org/en/v1/manual/methods/

Also in that vein, i believe we should not define a method `interior(arch, field)`. This is purely a convenience, because the arch is inside the field. The signaturae `interior(field)` should be sufficient.

glwagner: If you have a MetalGPU, you can inspect which method will be called at the REPL:

```julia
julia> @which interior(f)
```

this will print the line nubmer and file in which the function is defined. In this case you it should point to your new definition in the extension. If it does not then you may not have imported the name correctly (or something)

simone-silvestri: I think that would mean that for Metal fields `interior` would always be allocating. We could live with that.

alesok: I am playing with that suggestion:
```julia
function FD.interior(r::MetalField)
    interior_r = interior(r.data, FD.location(r), r.grid, r.indices)
    @show FD.location(r)
    
    if parent(interior_r) !== interior_r
        interior_r = copy(interior_r)
    end
    return interior_r
end
```
but stuck that ```@show FD.location(r)``` gives
```
FD.location(r) = (Nothing, Nothing, Nothing)
```
and ```@show size(interior_r)```gives
```
size(interior_r) = (1, 1, 1)
```

glwagner: those two results are consistent with one another if `r` is a triply reduced field.

Can you paste here what `@show r` produces? `@show parent(r)` will also help and `@show r.grid` perhaps.

glwagner: you can also figure out if you hit this function correctly:

https://github.com/CliMA/Oceananigans.jl/blob/4cc25a3e6f386cc6bfe1deed40f889dc22a31b6c/src/Fields/abstract_field.jl#L33

alesok: Using ```@show r``` in
```julia
function FD.interior(r::MetalField)
    @show r
    interior_r = interior(r.data, FD.location(r), r.grid, r.indices)

    if parent(interior_r) !== interior_r
        interior_r = copy(interior_r)
    end
    return interior_r
end
``` 
gives 
```
Warning: detected a stack overflow; program state may be corrupted, so further execution might be unreliable.
``` 
and long output starting with 
```
  StackOverflowError:
  
  caused by: StackOverflowError:
  
  caused by: StackOverflowError:
  Stacktrace:
       [1] macro expansion
         @ ~/.julia/packages/ObjectiveC/UNTzb/src/syntax.jl:160 [inlined]
       [2] alloc_buffer(dev::Metal.MTL.MTLDeviceInstance, bytesize::Int64, opts::UInt64)
         @ Metal.MTL ~/.julia/packages/Metal/av3If/lib/mtl/buffer.jl:76
       [3] Metal.MTL.MTLBuffer(dev::Metal.MTL.MTLDeviceInstance, bytesize::Int64; storage::Type, hazard_tracking::Metal.MTL.MTLResourceOptions, cache_mode::Metal.MTL.MTLResourceOptions)
         @ Metal.MTL ~/.julia/packages/Metal/av3If/lib/mtl/buffer.jl:26
       [4] MTLBuffer
         @ ~/.julia/packages/Metal/av3If/lib/mtl/buffer.jl:20 [inlined]
       [5] #26
         @ ~/.julia/packages/ObjectiveC/UNTzb/src/foundation.jl:652 [inlined]
       [6] macro expansion
         @ ~/.julia/packages/ObjectiveC/UNTzb/src/foundation.jl:572 [inlined]
       [7] macro expansion
         @ ./lock.jl:376 [inlined]
       [8] ObjectiveC.Foundation.NSAutoreleasePool(f::Metal.var"#26#27"{@Kwargs{storage::DataType}, Metal.MTL.MTLDeviceInstance, Int64, Tuple{}})
         @ ObjectiveC.Foundation ~/.julia/packages/ObjectiveC/UNTzb/src/foundation.jl:564
       [9] macro expansion
         @ ~/.julia/packages/ObjectiveC/UNTzb/src/foundation.jl:651 [inlined]
      [10] macro expansion
         @ ~/.julia/packages/Metal/av3If/src/pool.jl:54 [inlined]
      [11] macro expansion
         @ ./timing.jl:461 [inlined]
      [12] alloc(::Metal.MTL.MTLDeviceInstance, ::Int64; kwargs::@Kwargs{storage::DataType})
         @ Metal ~/.julia/packages/Metal/av3If/src/pool.jl:53
      [13] alloc
         @ ~/.julia/packages/Metal/av3If/src/pool.jl:52 [inlined]
      [14] encode_argument!
         @ ~/.julia/packages/Metal/av3If/src/compiler/execution.jl:260 [inlined]
      [15] macro expansion
         @ ~/.julia/packages/Metal/av3If/src/compiler/execution.jl:234 [inlined]
      [16] macro expansion
         @ ./none:0 [inlined]
      [17] encode_arguments!
         @ ./none:0 [inlined]
      [18] macro expansion
         @ ~/.julia/packages/Metal/av3If/src/compiler/execution.jl:285 [inlined]
      [19] (::Metal.var"#222#223"{Metal.MTL.MTLCommandQueueInstance, Metal.HostKernel{GPUArrays.var"#gpu_fill_kernel!#gpu_fill_kernel!##0", Tuple{KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicCheck, Nothing, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, KernelAbstractions.NDIteration.NDRange{1, KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicSize, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}}}, MtlDeviceArray{Float32, 3, 1}, Float32}}, Tuple{KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicCheck, Nothing, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, KernelAbstractions.NDIteration.NDRange{1, KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicSize, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}}}, MtlArray{Float32, 3, Metal.PrivateStorage}, Float32}})()
         @ Metal ~/.julia/packages/ObjectiveC/UNTzb/src/foundation.jl:644
      [20] macro expansion
         @ ~/.julia/packages/ObjectiveC/UNTzb/src/foundation.jl:572 [inlined]
      [21] macro expansion
         @ ./lock.jl:376 [inlined]
      [22] ObjectiveC.Foundation.NSAutoreleasePool(f::Metal.var"#222#223"{Metal.MTL.MTLCommandQueueInstance, Metal.HostKernel{GPUArrays.var"#gpu_fill_kernel!#gpu_fill_kernel!##0", Tuple{KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicCheck, Nothing, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, KernelAbstractions.NDIteration.NDRange{1, KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicSize, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}}}, MtlDeviceArray{Float32, 3, 1}, Float32}}, Tuple{KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicCheck, Nothing, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, KernelAbstractions.NDIteration.NDRange{1, KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicSize, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}}}, MtlArray{Float32, 3, Metal.PrivateStorage}, Float32}})
         @ ObjectiveC.Foundation ~/.julia/packages/ObjectiveC/UNTzb/src/foundation.jl:564
      [23] (::Metal.HostKernel{GPUArrays.var"#gpu_fill_kernel!#gpu_fill_kernel!##0", Tuple{KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicCheck, Nothing, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, KernelAbstractions.NDIteration.NDRange{1, KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicSize, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}}}, MtlDeviceArray{Float32, 3, 1}, Float32}})(::KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicCheck, Nothing, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, KernelAbstractions.NDIteration.NDRange{1, KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicSize, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}}}, ::Vararg{Any}; groups::Int64, threads::Int64, queue::Metal.MTL.MTLCommandQueueInstance)
         @ Metal ~/.julia/packages/ObjectiveC/UNTzb/src/foundation.jl:643
      [24] (::KernelAbstractions.Kernel{MetalBackend, KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicSize, GPUArrays.var"#gpu_fill_kernel!#gpu_fill_kernel!##0"})(::MtlArray{Float32, 3, Metal.PrivateStorage}, ::Vararg{Any}; ndrange::Int64, workgroupsize::Nothing)
         @ Metal.MetalKernels ~/.julia/packages/Metal/av3If/src/MetalKernels.jl:132
      [25] Kernel
         @ ~/.julia/packages/Metal/av3If/src/MetalKernels.jl:111 [inlined]
      [26] fill!
         @ ~/.julia/packages/GPUArrays/0F4Dn/src/host/construction.jl:22 [inlined]
      [27] zeros
         @ ~/.julia/packages/Metal/av3If/src/array.jl:525 [inlined]
      [28] zeros(::MetalBackend, ::Type{Float32}, dims::Tuple{Int64, Int64, Int64})
         @ Metal.MetalKernels ~/.julia/packages/Metal/av3If/src/MetalKernels.jl:27
      [29] zeros(::MetalBackend, ::Type, ::Int64, ::Vararg{Int64}; kwargs::@Kwargs{})
         @ KernelAbstractions ~/.julia/packages/KernelAbstractions/X5fk1/src/KernelAbstractions.jl:564
      [30] zeros
         @ ~/.julia/packages/KernelAbstractions/X5fk1/src/KernelAbstractions.jl:564 [inlined]
      [31] zeros
         @ ~/models/Oceananigans.jl/src/Grids/zeros_and_ones.jl:8 [inlined]
      [32] new_data(FT::DataType, arch::GPU{MetalBackend}, loc::Tuple{Nothing, Nothing, Nothing}, topo::Tuple{DataType, DataType, DataType}, sz::Tuple{Int64, Int64, Int64}, halo_sz::Tuple{Int64, Int64, Int64}, indices::Tuple{Colon, Colon, Colon})
         @ Oceananigans.Grids ~/models/Oceananigans.jl/src/Grids/new_data.jl:68
      [33] new_data
         @ ~/models/Oceananigans.jl/src/Grids/new_data.jl:73 [inlined]
      [34] maximum(f::Function, c::Field{Nothing, Nothing, Nothing, Nothing, RectilinearGrid{Float32, Periodic, Periodic, Bounded, Oceananigans.Grids.StaticVerticalDiscretization{OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, Float32, Float32}, Float32, Float32, OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, GPU{MetalBackend}}, Tuple{Colon, Colon, Colon}, OffsetArray{Float32, 3, MtlArray{Float32, 3, Metal.PrivateStorage}}, Float32, FieldBoundaryConditions{Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, @NamedTuple{bottom_and_top::Nothing, south_and_north::Nothing, west_and_east::Nothing}, @NamedTuple{bottom_and_top::Tuple{Nothing, Nothing}, south_and_north::Tuple{Nothing, Nothing}, west_and_east::Tuple{Nothing, Nothing}}}, Nothing, Nothing}; condition::Nothing, mask::Float64, dims::Function)
         @ Oceananigans.Fields ~/models/Oceananigans.jl/src/Fields/field.jl:784
      [35] maximum(f::Function, c::Field{Nothing, Nothing, Nothing, Nothing, RectilinearGrid{Float32, Periodic, Periodic, Bounded, Oceananigans.Grids.StaticVerticalDiscretization{OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, Float32, Float32}, Float32, Float32, OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, GPU{MetalBackend}}, Tuple{Colon, Colon, Colon}, OffsetArray{Float32, 3, MtlArray{Float32, 3, Metal.PrivateStorage}}, Float32, FieldBoundaryConditions{Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, @NamedTuple{bottom_and_top::Nothing, south_and_north::Nothing, west_and_east::Nothing}, @NamedTuple{bottom_and_top::Tuple{Nothing, Nothing}, south_and_north::Tuple{Nothing, Nothing}, west_and_east::Tuple{Nothing, Nothing}}}, Nothing, Nothing})
         @ Oceananigans.Fields ~/models/Oceananigans.jl/src/Fields/field.jl:774
      [36] maximum(c::Field{Nothing, Nothing, Nothing, Nothing, RectilinearGrid{Float32, Periodic, Periodic, Bounded, Oceananigans.Grids.StaticVerticalDiscretization{OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, Float32, Float32}, Float32, Float32, OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, OffsetVector{Float32, StepRangeLen{Float32, Float32, Float32, Int64}}, GPU{MetalBackend}}, Tuple{Colon, Colon, Colon}, OffsetArray{Float32, 3, MtlArray{Float32, 3, Metal.PrivateStorage}}, Float32, FieldBoundaryConditions{Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, @NamedTuple{bottom_and_top::Nothing, south_and_north::Nothing, west_and_east::Nothing}, @NamedTuple{bottom_and_top::Tuple{Nothing, Nothing}, south_and_north::Tuple{Nothing, Nothing}, west_and_east::Tuple{Nothing, Nothing}}}, Nothing, Nothing}; kwargs::@Kwargs{})
         @ Oceananigans.Fields ~/models/Oceananigans.jl/src/Fields/field.jl:795
```

glwagner: Try doing this outside of the function. For example, in a script (this can also be typed at the REPL):

```julia
using Oceananigans
using Metal
Oceananigans.defaults.FloatType = Float32
metal = Metal.MetalBackend()
arch = GPU(metal)
grid = RectilinearGrid(arch, size=(8, 8, 8), extent=(1, 1, 1))
c = CenterField(grid)
```

glwagner: This seems to work for me on `main`:

```julia
julia> using Oceananigans

julia> using Metal

julia> Oceananigans.defaults.FloatType = Float32
Float32

julia> metal = Metal.MetalBackend()
MetalBackend()

julia> arch = GPU(metal)
GPU{MetalBackend}(MetalBackend())

julia> grid = RectilinearGrid(arch, size=(8, 8, 8), extent=(1, 1, 1))
8×8×8 RectilinearGrid{Float32, Periodic, Periodic, Bounded} on MetalGPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.125
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=0.125
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=0.125

julia> c = CenterField(grid)
8×8×8 Field{Center, Center, Center} on RectilinearGrid on MetalGPU
├── grid: 8×8×8 RectilinearGrid{Float32, Periodic, Periodic, Bounded} on MetalGPU with 3×3×3 halo
│   └── west: Periodic, east: Periodic, south: Periodic, north: Periodic, bottom: ZeroFlux, top: ZeroFlux, immersed: Nothing
└── data: 14×14×14 OffsetArray(::MtlArray{Float32, 3, Metal.PrivateStorage}, -2:11, -2:11, -2:11) with eltype Float32 with indices -2:11×-2:11×-2:11

julia> interior(c)
8×8×8 view(::MtlArray{Float32, 3, Metal.PrivateStorage}, 4:11, 4:11, 4:11) with eltype Float32:
[:, :, 1] =
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0

[:, :, 2] =
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0

[:, :, 3] =
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0

[:, :, 4] =
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0

[:, :, 5] =
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0

[:, :, 6] =
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0

[:, :, 7] =
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0

[:, :, 8] =
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
```

glwagner:
```julia
julia> Oceananigans.Fields.location(c)
(Center, Center, Center)
```

glwagner: are you certain that you are not in a case where `interior(r)` _should_ return `(Nothing, Nothing, Nothing)`? What is the expected result?

alesok: Here is output on version with interior() extended in the OceananigunsMetalExt
```julia
Precompiling VSCodeServer finished.
  1 dependency successfully precompiled in 6 seconds. 30 already precompiled.
julia> using Oceananigans

julia> using Metal

julia> Oceananigans.defaults.FloatType = Float32
Float32

julia> metal = Metal.MetalBackend()
MetalBackend()

julia> arch = GPU(metal)
GPU{MetalBackend}(MetalBackend())

julia> grid = RectilinearGrid(arch, size=(8, 8, 8), extent=(1, 1, 1))
8×8×8 RectilinearGrid{Float32, Periodic, Periodic, Bounded} on MetalGPU with 3×3×3 halo
├── Periodic x ∈ [0.0, 1.0)  regularly spaced with Δx=0.125
├── Periodic y ∈ [0.0, 1.0)  regularly spaced with Δy=0.125
└── Bounded  z ∈ [-1.0, 0.0] regularly spaced with Δz=0.125

julia> c = CenterField(grid)
Warning: detected a stack overflow; program state may be corrupted, so further execution might be unreliable.
Error showing value of type Field{Center, Center, Center, Nothing, RectilinearGrid{Float32, Periodic, Periodic, Bounded, Oceananigans.Grids.StaticVerticalDiscretization{OffsetArrays.OffsetVect
``` 

But using version with safe_interior the script produces the same output as yours above.

alesok: And no, I am not certain of anything. I suspect that something fishy may be going on when calling
```julia
interior(r.data, FD.location(r), r.grid, r.indices)
``` 
from ```interior(r)``` . In principle, this should be fine due to multiple dispatch, but there may be an unexpected chain of method calls involved. Sorry—I dislike speculating ...

glwagner: You can call that function at the REPL to test it. Definitely, do not speculate. Test.

alesok: I've found that  the issue with adding a method for ```interior``` in ```OceananigansMetalExt``` stems from the fact that Metal requires materialisation of a view in order to perform reduction operations. Implementing materialisation in ```safe_interior``` by creating a copy works correctly for this purpose.

However, once we add another method to the generic ```function interior``` in ```OceananigansMetalExt```, that method is selected not only for reductions but also for many other operations that are intended to mutate the underlying data. In those cases, the mutations are applied to the copy rather than to the original field data, and therefore do not propagate back.

Test example:
```julia
using Oceananigans
using Metal

Oceananigans.defaults.FloatType = Float32
metal = Metal.MetalBackend()
arch = GPU(metal)
grid = RectilinearGrid(arch, size=(8, 8, 8), extent=(1, 1, 1))
c = CenterField(grid)
set!(c, 1)
```
When ```OceananigansMetalExt``` does not define a method for ```interior```, the result is as expected:
```julia
julia> include("my_test.jl")
8×8×8 Field{Center, Center, Center} on RectilinearGrid on MetalGPU
├── grid: 8×8×8 RectilinearGrid{Float32, Periodic, Periodic, Bounded} on MetalGPU with 3×3×3 halo
│   └── west: Periodic, east: Periodic, south: Periodic, north: Periodic, bottom: ZeroFlux, top: ZeroFlux, immersed: Nothing
└── data: 14×14×14 OffsetArray(::MtlArray{Float32, 3, Metal.PrivateStorage}, -2:11, -2:11, -2:11) with eltype Float32 with indices -2:11×-2:11×-2:11
```

But when ```OceananigansMetalExt``` does define a method for ```interior```, the same script yields:
```julia
julia> include("my_test.jl")
Precompiling OceananigansMetalExt finished.
  1 dependency successfully precompiled in 3 seconds
8×8×8 Field{Center, Center, Center} on RectilinearGrid on MetalGPU
├── grid: 8×8×8 RectilinearGrid{Float32, Periodic, Periodic, Bounded} on MetalGPU with 3×3×3 halo
│   └── west: Periodic, east: Periodic, south: Periodic, north: Periodic, bottom: ZeroFlux, top: ZeroFlux, immersed: Nothing
└── data: 14×14×14 OffsetArray(::MtlArray{Float32, 3, Metal.PrivateStorage}, -2:11, -2:11, -2:11) with eltype Float32 with indices -2:11×-2:11×-2:11
``` 
The max=0.0, min=0.0, mean=0.0.

So, I think it is preferable to keep using ```safe_interior``` as in the current commit proposed by simone-silvestri, rather than extending ```interior```. It may also be worth renaming ```safe_interior``` to something like ```copy_interior``` to make explicit that it returns a materialized copy intended specifically for reduction operations.
What do you think?

glwagner: If we have no other choice than to modify the source code for this metal-specific implementation, then so be it. Since `copy_interior` does not always make a copy, you should probably call it "maybe_copy_interior"

alesok: The commit [bea62dc](https://github.com/CliMA/Oceananigans.jl/pull/4990/commits/bea62dc2a3489ccaf8ca350840039dc21a247ff3) passes all the tests and uses "maybe_copy_interior" for reductions. Please check all changes. Thanks

src/Advection/cell_advection_timescale.jl
```diff
@@ -13,7 +13,7 @@ is the minimum over all `i, j, k` in the `grid` of
 function cell_advection_timescale(grid, velocities)
     u, v, w = velocities
     τ = KernelFunctionOperation{Center, Center, Center}(cell_advection_timescaleᶜᶜᶜ, grid, u, v, w)
-    return minimum(τ)
+    return minimum(τ) #TODO here is a problem on MetalGPU
```
glwagner:
```suggestion
    return minimum(τ)
```

src/Fields/field.jl
```diff
@@ -700,7 +700,7 @@ get_neutral_mask(::Union{AllReduction, AnyReduction})  = true
 get_neutral_mask(::Union{SumReduction, MeanReduction}) = 0
 get_neutral_mask(::ProdReduction)    = 1
 
-# TODO make this Float32 friendly
+# TODO(completed): make this Float32 friendly. Masks are converted in appropriated functions
```
glwagner:
```suggestion
```

test/test_metal.jl
glwagner:
```suggestion
```

test/test_metal.jl
```diff
@@ -39,4 +40,102 @@ Oceananigans.defaults.FloatType = Float32
     @test iteration(simulation) == 3
     @test time(simulation) == 3minutes
 end
+#! format: on
```
glwagner:
```suggestion
```

test/test_metal.jl
```diff
@@ -39,4 +40,102 @@ Oceananigans.defaults.FloatType = Float32
...
 
+@testset "MetalGPU: ImmersedBoundaryGrid" begin
+    arch = GPU(Metal.MetalBackend())
+    FT = Oceananigans.defaults.FloatType
```
glwagner:
```suggestion
    FT = Float32
```

test/test_metal.jl
glwagner: @alesok can we make this simplification?

test/test_metal.jl
glwagner: I prefer @navidcy 's suggestion here

glwagner:
This is good to go for me. For sure we need to remove the comments which are now incorrect before merging. I also hope we can merge the suggestions that simplify the tests, but this can also be done in a future PR.

I did not follow _why_ we need `maybe_copy_interior`. @alesok, can you please add a comment that explains why this is necessary? I also wonder if it is needed due to a bug in Metal. I think we can put it into Oceananigans as a hotfix for now. But it could also make sense to raise an issue upstream so that in the long run, we can remove `maybe_copy_interior`.

glwagner:
Possibly we can't build the docs from a fork so we will have to merge and hope they are ok?

navidcy:
Sure.

But the CI / Metal is not passing which I'd say is crucial for this PR?

glwagner:
oh yeah, that's true, I assumed it was. @alesok would you like some help with this?

alesok:
Yes, please. I don't know why the CI/Metal is not passing. The test_metal works well on my mac.
```julia
sokolov@mbp23 Oceananigans.jl % julia --project=. test/test_metal.jl
[2026/01/04 14:26:23.067] INFO  Initializing simulation...
[2026/01/04 14:26:27.239] INFO      ... simulation initialization complete (4.121 seconds)
[2026/01/04 14:26:27.245] INFO  Executing initial time step...
[2026/01/04 14:26:30.268] INFO      ... initial time step complete (3.023 seconds).
[2026/01/04 14:26:30.283] INFO  Simulation is stopping after running for 7.401 seconds.
[2026/01/04 14:26:30.283] INFO  Model iteration 3 equals or exceeds stop iteration 3.
Test Summary:      | Pass  Total   Time
MetalGPU extension |   10     10  38.3s
[2026/01/04 14:26:46.773] INFO  Initializing simulation...
[2026/01/04 14:26:54.147] INFO      ... simulation initialization complete (7.374 seconds)
[2026/01/04 14:26:54.147] INFO  Executing initial time step...
[2026/01/04 14:26:58.529] INFO  Simulation is stopping after running for 0 seconds.
[2026/01/04 14:26:58.529] INFO  Model iteration 1 equals or exceeds stop iteration 1.
[2026/01/04 14:26:58.529] INFO      ... initial time step complete (4.382 seconds).
Test Summary:                  | Pass  Total   Time
MetalGPU: ImmersedBoundaryGrid |    9      9  28.1s
Test Summary:                 | Pass  Total  Time
MetalGPU: test for reductions |    2      2  5.5s
[2026/01/04 14:27:08.967] INFO  Initializing simulation...
[2026/01/04 14:27:12.134] INFO      ... simulation initialization complete (3.167 seconds)
[2026/01/04 14:27:12.134] INFO  Executing initial time step...
[2026/01/04 14:27:13.745] INFO      ... initial time step complete (1.611 seconds).
[2026/01/04 14:27:13.833] INFO  Simulation is stopping after running for 4.862 seconds.
[2026/01/04 14:27:13.833] INFO  Model iteration 20 equals or exceeds stop iteration 20.
Test Summary:            | Pass  Total  Time
MetalGPU: TimeStepWizard |    2      2  9.8s
sokolov@mbp23 Oceananigans.jl % 
```

glwagner:
@alesok are you using Julia 1.12?

alesok:
Yes, 1.12.3
```Julia
julia> Metal.versioninfo()
macOS 26.2.0, Darwin 25.2.0

Toolchain:
- Julia: 1.12.3
- LLVM: 18.1.7

Julia packages:
- Metal.jl: 1.9.1
- GPUArrays: 11.3.2
- GPUCompiler: 1.7.5
- KernelAbstractions: 0.9.39
- ObjectiveC: 3.4.2
- LLVM: 9.4.4
- LLVMDowngrader_jll: 0.6.0+1

1 device:
- Apple M3 Pro 18 GPU cores (80.000 KiB allocated)
```

navidcy:
To fully reproduce the CI, you should do:

```
$ TEST_FILE=test_metal.jl julia --project -e"using Pkg; Pkg.test()"
```

For me that breaks with:

[metal.err.log](https://github.com/user-attachments/files/24423957/metal.err.log)

navidcy:
I tried to find the culprit. The error comes up when the kernel:

```Julia
    launch!(arch, grid, kernel_parameters,
            compute_hydrostatic_free_surface_Gu!, model.timestepper.Gⁿ.u, grid,
            u_kernel_args; active_cells_map)
```
is launched with an immersed grid.

src/Advection/cell_advection_timescale.jl
navidcy: Is the comment still relevant @alesok?

src/Fields/field.jl
```diff
@@ -249,7 +249,7 @@ function Base.similar(f::Field, grid=f.grid)
 end
 
 """
-    offset_windowed_data(data, data_indices, loc, grid, view_indices)
+offset_windowed_data(data, data_indices, loc, grid, view_indices)
```
navidcy: why deleted the 4 spaces that convert the signature in code form?

src/Fields/field.jl
navidcy: I agree, if completed then the whole comments needs to be deleted!

test/test_metal.jl
navidcy: Actually if we remove the FT's from in there the CI passes. This might be telling us something?!?

glwagner:
> I tried to find the culprit. The error comes up when the kernel:
> 
> ```julia
>     launch!(arch, grid, kernel_parameters,
>             compute_hydrostatic_free_surface_Gu!, model.timestepper.Gⁿ.u, grid,
>             u_kernel_args; active_cells_map)
> ```
> 
> is launched with an immersed grid.

it appears to be a type inference error?

navidcy:
Does that smell like related to a method being overwritten instead of added? E.g. what ExplicitImports.jl trying to deal with?

glwagner:
why would it fail when included as a single file but pass when run outside of the test environment? that could suggest that the codepath depends on the test environment specifically somehow.

i'm not sure how it would be due to an internal Oceananigans source code method being overwritten, but have to thikn about it...

navidcy:
Yes, I agree. It feels fishy to me as well.

simone-silvestri:
> To fully reproduce the CI, you should do:
> 
> ```
> $ TEST_FILE=test_metal.jl julia --project -e"using Pkg; Pkg.test()"
> ```
> 
> For me that breaks with:
> 
> [metal.err.log](https://github.com/user-attachments/files/24423957/metal.err.log)

Very weird, I cannot reproduce this error on my mac

alesok:
Can you run ```julia --project=. test/test_metal.jl  ``` without errors?

simone-silvestri:
I ran `$ TEST_FILE=test_metal.jl julia --project -e"using Pkg; Pkg.test()"` without errors, which should be more restrictive.
Anyways I approved the workflow on github actions to run. Let's see what is the outcome there...

navidcy:
Yes, this
```
$ TEST_FILE=test_metal.jl julia --project -e"using Pkg; Pkg.test()"
```
works for me.

navidcy:
All CI passes except the one that matters: Metal.
I don't think there is any point updating the branch to rerun the whole CI if we don't find what's keeping the Metal from passing. Otherwise we are just wasting resources...

I'm a bit on a dead end here though. I really can't understand why locally the tests would pass but on the CI no. Perhaps something to do with the virtual Mac OS X machine? Is it a virtual machine?

glwagner:
The CI has this statement:

```
│ To reproduce this CI run locally run the following from the same repository state on julia version 1.12.4:
│
│ `import Pkg; Pkg.test(;coverage=true, julia_args=["--check-bounds=yes", "--compiled-modules=yes", "--depwarn=yes"], force_latest_compatible_version=false, allow_reresolve=true)`
```

did we try that?

navidcy:
Not me. Let me try now!

navidcy:
Yeap, it fails...

[error.log](https://github.com/user-attachments/files/24538128/error.log)

alesok:
> All CI passes except the one that matters: Metal. I don't think there is any point updating the branch to rerun the whole CI if we don't find what's keeping the Metal from passing. Otherwise we are just wasting resources...
> 
> I'm a bit on a dead end here though. I really can't understand why locally the tests would pass but on the CI no. Perhaps something to do with the virtual Mac OS X machine? Is it a virtual machine?

I am afraid not.  I got the same errors running 
```bash
TEST_FILE=test_metal.jl julia --project -e "import Pkg; Pkg.test()"
``` 
on the local Mac as we have in the CI env. However, it runs correctly using
```bash
julia --project=. test/test_metal.jl 
```

glwagner:
> > All CI passes except the one that matters: Metal. I don't think there is any point updating the branch to rerun the whole CI if we don't find what's keeping the Metal from passing. Otherwise we are just wasting resources...
> > I'm a bit on a dead end here though. I really can't understand why locally the tests would pass but on the CI no. Perhaps something to do with the virtual Mac OS X machine? Is it a virtual machine?
> 
> I am afraid not. I got the same errors running
> 
> ```shell
> TEST_FILE=test_metal.jl julia --project -e "import Pkg; Pkg.test()"
> ```
> 
> on the local Mac as we have in the CI env. However, it runs correctly using
> 
> ```shell
> julia --project=. test/test_metal.jl 
> ```

this is a great clue! it means that you can reproduce the test results. The difference may have to do with either the different compile options that the test environment uses (for example, always checking bounds) or the different test environment / combination of packages between either case.

alesok:
Finally, commit https://github.com/CliMA/Oceananigans.jl/pull/4990/commits/5883cb3e8f9b55107e3660e8e8b7f17089b3c7ec passes all tests. It appears that using an ImmersedBoundaryGrid created with the default setting active_cells_map = false leads to an error in the model. Setting active_cells_map = true resolves the issue.

navidcy:
> Finally, commit [5883cb3](https://github.com/CliMA/Oceananigans.jl/commit/5883cb3e8f9b55107e3660e8e8b7f17089b3c7ec) passes all tests. It appears that using an ImmersedBoundaryGrid created with the default setting `active_cells_map = false` leads to an error in the model. Setting active_cells_map = true resolves the issue.

Interesting! Thanks @alesok!!

@simone-silvestri does this give us an insightful information?

If there is a fundamental reason why `active_cells_map = false` shouldn't work with Metal we can spit out an informative error to users. Or if there is a method we need to add perhaps we do that?

alesok:
Also interesting that using active_cells_map = false passes tests running as
```bash
julia --project=. test/test_metal.jl 
```
but fails with 
```bash
julia --project=. --check-bounds=yes test/test_metal.jl
``` 
in the same way as with 
```bash
TEST_FILE=test_metal.jl julia --project -e "using Pkg; Pkg.test()"
```

simone-silvestri:
> Finally, commit [5883cb3](https://github.com/CliMA/Oceananigans.jl/commit/5883cb3e8f9b55107e3660e8e8b7f17089b3c7ec) passes all tests. It appears that using an ImmersedBoundaryGrid created with the default setting active_cells_map = false leads to an error in the model. Setting active_cells_map = true resolves the issue.

very weird, `active_cells_map = true` adds complexity does not remove it. What is the error?

alesok:
When ```active_cells_map``` is set to ```true```, all tests in ```test_metal.jl``` pass.  
When it is set to ```false```, the tests fail.  

This failure can also be reproduced by running:
```bash
julia --project=. --check-bounds=yes test/test_metal.jl
```

glwagner:
> When `active_cells_map` is set to `true`, all tests in `test_metal.jl` pass. When it is set to `false`, the tests fail.
> 
> This failure can also be reproduced by running:
> 
> ```shell
> julia --project=. --check-bounds=yes test/test_metal.jl
> ```

Happy to always set to `true` in the tests as a stop gap to get this PR in (we can open an issue to indicate the need for further work). We can also extend the constructor to make this default in the extension.

alesok:
I think it's a good suggestion

glwagner:
@alesok tests pass. are you ok if I merge?

glwagner (approved):

alesok:
Yes, It will be fine and we can go further with smaller PRs

PR #4992, opened 2025-11-26
Check undefined exports with Aqua

giordano:
Opening as draft because this fails at the moment.  The test failure is elaborated in https://github.com/CliMA/Oceananigans.jl/pull/4979#issue-3660106097.  @simone-silvestri when you get a chance, could you please help with this one?  I think we need to either not export the non-existing metrics (and the related methods), or define all of them.  In https://github.com/CliMA/Oceananigans.jl/pull/4979#issuecomment-3572316835 you seemed to lean towards the former.

### DIFF
--- a/src/Operators/spacings_and_areas_and_volumes.jl
+++ b/src/Operators/spacings_and_areas_and_volumes.jl
@@ -382,7 +382,7 @@ end
 ##### We also use the function "volume" rather than `V`.
 #####
 
-function interpolate_location(val::Symbol) 
+function location_from_superscript(val::Symbol) 
     if val == :ᶜ 
         return :Center 
     elseif val == :ᶠ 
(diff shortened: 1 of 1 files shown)


simone-silvestri:
Sure, I have some time to do it this weekend. I think it is better to add the methods rather than avoiding the exports unless the burden on precompilation becomes excessive.

simone-silvestri:
Tests should pass now. I have defined 3D and 2D methods for all metrics and operators. I do not think that 0D methods should be defined (i.e. what does it mean to have `δxᵃᵃᵃ`?), so I have always left at least one explicit dimension (depending on the direction of the metric / operator) either center or face. 
I think we can merge this, next we can look at the ambiguities, it's going to be fun 😄

src/AbstractOperations/grid_metrics.jl
giordano, line 1: Ugh, it's a bit unfortunate the number of methods is part of a docstring test.  In principle we could [filter the numbers out](https://documenter.juliadocs.org/stable/man/doctests/#Filtering-Doctests) not to affect the tests, although perhaps it's good to keep those numbers consistent?

simone-silvestri: yeah, I guess we will not change this for a while :). In the end, we do not touch the operators module that often.

PR #4994, opened 2025-11-27
Remove all self-qualified accesses

giordano:
More follow up to #4985.


src/DistributedComputations/distributed_macros.jl
giordano, line 1: Similarly to #4178, simply interpolating the functions is an improvement on its own right: it allows you to use the macros in scopes where the `Oceananigans` module isn't actually available.

glwagner (approved):

PR #4997, opened 2025-11-27
Use explicit imports in more modules

giordano:
Another follow up to #4985.

### DIFF
--- a/src/AbstractOperations/derivatives.jl
+++ b/src/AbstractOperations/derivatives.jl
@@ -1,4 +1,4 @@
-using Oceananigans.Operators: interpolation_code
+using Oceananigans.Operators: Operators, interpolation_code
 
 struct Derivative{LX, LY, LZ, D, A, IN, AD, G, T} <: AbstractOperation{LX, LY, LZ, G, T}
                ∂ :: D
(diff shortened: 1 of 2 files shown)


src/AbstractOperations/AbstractOperations.jl
```diff
@@ -19,7 +19,7 @@ using Oceananigans: location
 using Oceananigans.Fields: instantiated_location
 using Oceananigans.Operators: interpolation_operator
 
-import Adapt
+using Adapt: Adapt, adapt
```
navidcy, line 15: minor question

is this

```Julia
using Adapt: Adapt, adapt
```

just simply exactly equivalent  to


```Julia
using Adapt
using Adapt: adapt
```
?

giordano: No:
```julia
using Adapt
```
brings into scope all symbols exported by `Adapt` (in addition to loading the module)
```julia
using Adapt: Adapt, adapt
```
brings into scope _only_ the listed symbols (`Adapt`, `adapt` in this case)

navidcy: Does this mean that if `foo` is exported by `Adapt` then after

```Julia
using Adapt: Adapt
```

I can call it with 

```Julia
Adapt.foo
```

while after

```Julia
using Adapt
```

I just need to call

```Julia
foo
```

Correct?

giordano: Correct.  The [package documentation](https://juliatesting.github.io/ExplicitImports.jl/stable/) gives some motivation for being explicit.  Also, this enables identifying unused functions/variables, to potentially clean up the code.  Furthermore in previous related PRs I found a few cases of slightly overcomplicated or suboptimal code (nothing dramatic to be fair, but these checks make code simpler and/or more robust overall)

src/AbstractOperations/derivatives.jl
giordano, line 1: It took me a bit to understand what the hell was going on with these `eval`s (😱), but I believe I cracked it: you always want the metrics in `Operators`, so let's just `eval` inside that module and that's it, easy peasy (would be easier peasier without `eval` 😰)

glwagner: I am not sure I understand how `eval` is being used here either.

glwagner: @simone-silvestri ?

giordano: That's quite interesting because you introduced the eval here in e79fc820cf206ce42c042d81d0e63261cebc94f7 🙃

PR #5000, opened 2025-11-29
Add `GPUArraysCore` to list of weakdeps

giordano:
Apparently Julia v1.11 (but not v1.12) requires a package which is both a regular dep and a weak dep to be listed in both sections, currently we get
```
Oceananigans has a malformed Project.toml, the extension package GPUArraysCore is not listed in [weakdeps]
```


giordano:
Oh, Julia v1.12 doesn't like this at all.  How's this supposed to work?!?

navidcy:
Perhaps we can add a compat to preclude Julia v1.11 from Oceananigans?
We never got around to have it working with immersed boundaries...

giordano:
Apparently the error above was fixed in v1.11.1 (upstream patch to Pkg: https://github.com/JuliaLang/Pkg.jl/pull/3865, backport to julia v1.11.1: https://github.com/JuliaLang/julia/pull/56032)

navidcy:
Btw, let's note that this was the 5000th PR! 🥳

PR #5002, opened 2025-11-30
Generalize bounded WENO advection

glwagner:
This generalization is necessary to support "mass advection" of scalars, in which the mass-specific scalar distribution is reconstructed (as opposed to the scalar density). The generalize is needed because th bounded WENO functions do not follow the common form implemented by the other advection schemes.

Note, as far as I can tell, the fact that this is implemented differently is not strictly necessary. However, it avoids repeating some computations? @simone-silvestri if you can confirm this motivation for writing the code this way, I will add a note.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oceananigans"
 uuid = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 authors = ["Climate Modeling Alliance and contributors"]
-version = "0.102.2"
+version = "0.102.3"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/ext/OceananigansNCDatasetsExt.jl
+++ b/ext/OceananigansNCDatasetsExt.jl
@@ -222,7 +222,9 @@ function create_spatial_dimensions!(dataset, dims, attributes_dict; dimension_ty
         dim_name == "" && continue # Don't create anything if dim_name is an empty string
         push!(effective_dim_names, dim_name)
 
-        dim_array = dimension_type.(dim_array) # Transform dim_array to the correct float type
+        # Transform dim_array to the correct float type and ensure it's on the CPU
+        dim_array = collect(dimension_type.(dim_array))
+
         if dim_name ∉ keys(dataset.dim)
             # Create missing dimension
             defVar(dataset, dim_name, dim_array, (dim_name,), attrib=attributes_dict[dim_name]; kwargs...)
--- a/src/BoundaryConditions/boundary_condition.jl
+++ b/src/BoundaryConditions/boundary_condition.jl
@@ -119,6 +119,18 @@ struct MixedCondition{A, B}
     inhomogeneity :: B
 end
 
+# Helper to unwrap Ref values (Ref is not isbits and can't be passed to GPU kernels)
+_unwrap_for_gpu(r::Base.RefValue) = r[]
+_unwrap_for_gpu(x) = x
+
+Adapt.adapt_structure(to, mc::MixedCondition) =
+    MixedCondition(_unwrap_for_gpu(mc.coefficient),
+                   Adapt.adapt(to, mc.inhomogeneity))
+
+on_architecture(to, mc::MixedCondition) =
+    MixedCondition(on_architecture(to, mc.coefficient),
+                   on_architecture(to, mc.inhomogeneity))
+
 """
     MixedBoundaryCondition(coefficient, inhomogeneity=0; kwargs...)
 
--- a/test/test_netcdf_writer.jl
+++ b/test/test_netcdf_writer.jl
@@ -10,7 +10,7 @@ using SeawaterPolynomials.TEOS10: TEOS10EquationOfState
 using SeawaterPolynomials.SecondOrderSeawaterPolynomials: RoquetEquationOfState
 
 using Oceananigans: Clock
-using Oceananigans.Models.HydrostaticFreeSurfaceModels: VectorInvariant
+using Oceananigans.Models.HydrostaticFreeSurfaceModels: VectorInvariant, ImplicitFreeSurface
 using Oceananigans.OutputWriters: trilocation_dim_name
 using Oceananigans.Grids: ξname, ηname, rname, ξnodes, ηnodes
 using Oceananigans.Fields: interpolate!
@@ -2513,18 +2513,22 @@ function test_netcdf_overriding_attributes(arch)
     return nothing
 end
 
-function test_netcdf_free_surface_only_output(arch)
+function test_netcdf_hydrostatic_free_surface_only_output(arch; immersed=false, vertically_stretched=false)
     Nλ, Nφ, Nz = 8, 8, 4
     Hλ, Hφ, Hz = 3, 4, 2
 
-    grid = LatitudeLongitudeGrid(arch;
+    z = vertically_stretched ? [k^2 - 100 for k in 0:Nz] : (-100, 0)
+
+    underlying_grid = LatitudeLongitudeGrid(arch;
         topology = (Bounded, Bounded, Bounded),
         size = (Nλ, Nφ, Nz),
         halo = (Hλ, Hφ, Hz),
         longitude = (-1, 1),
         latitude = (-1, 1),
-        z = (-100, 0)
+        z
     )
+
+    grid = immersed ? ImmersedBoundaryGrid(underlying_grid, GridFittedBottom(-50)) : underlying_grid
 
     model = HydrostaticFreeSurfaceModel(; grid,
         closure = ScalarDiffusivity(ν=4e-2, κ=4e-2),
@@ -2540,7 +2544,9 @@ function test_netcdf_free_surface_only_output(arch)
     )
 
     Arch = typeof(arch)
-    filepath_with_halos = "test_free_surface_with_halos_$Arch.nc"
+    immersed_str = immersed ? "_immersed" : ""
+    stretched_str = vertically_stretched ? "_stretched" : ""
+    filepath_with_halos = "test_free_surface_with_halos_$(Arch)$(immersed_str)$(stretched_str).nc"
     isfile(filepath_with_halos) && rm(filepath_with_halos)
 
     simulation.output_writers[:with_halos] =
@@ -2549,7 +2555,7 @@ function test_netcdf_free_surface_only_output(arch)
             schedule = IterationInterval(1),
             with_halos = true)
 
-    filepath_no_halos = "test_free_surface_no_halos_$Arch.nc"
+    filepath_no_halos = "test_free_surface_no_halos_$(Arch)$(immersed_str)$(stretched_str).nc"
     isfile(filepath_no_halos) && rm(filepath_no_halos)
 
     simulation.output_writers[:no_halos] =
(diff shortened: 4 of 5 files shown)


src/Advection/bounds_preserving_tracer_advection_operators.jl
```diff
@@ -24,91 +23,98 @@ end
...
-                     Axᶠᶜᶜ(i,   j, k, grid) * upwind_biased_product(u[i,   j, k], c₋ᴸ, c₋ᴿ)
-end
-
-@inline function bounded_tracer_flux_divergence_y(i, j, k, grid, advection::BoundsPreservingWENO, v, c)
+    Ax_ρuc⁺ = ℑxᶠᵃᵃ(i+1, j, k, grid, ρ) * Axᶠᶜᶜ(i+1, j, k, grid) * upwind_biased_product(u[i+1, j, k], c₊ᴸ, c₊ᴿ)
+    Ax_ρuc⁻ = ℑxᶠᵃᵃ(i,   j, k, grid, ρ) * Axᶠᶜᶜ(i,   j, k, grid) * upwind_biased_product(u[i,   j, k], c₋ᴸ, c₋ᴿ)
 
-    lower_limit = @inbounds advection.bounds[1]
-    upper_limit = @inbounds advection.bounds[2]
+    return Ax_ρuc⁺ - Ax_ρuc⁻
```
glwagner, line 59: @simone-silvestri this is technically the negative of the flux divergence. Do you understand what is happening here? Is there a cancellation of signs?

src/Advection/bounds_preserving_tracer_advection_operators.jl
```diff
@@ -24,91 +23,98 @@ end
...
     c₋ᴸ = _biased_interpolate_xᶠᵃᵃ(i,   j, k, grid, advection, LeftBias(),  c)
     c₋ᴿ = _biased_interpolate_xᶠᵃᵃ(i,   j, k, grid, advection, RightBias(), c)
 
-    FT = typeof(cᵢⱼ)
+    FT = eltype(c)
     ω̂₁ = convert(FT, _ω̂₁)
     ω̂ₙ = convert(FT, _ω̂ₙ)
     ε₂ = convert(FT, _ε₂)
 
+    cᵢⱼ = c[i, j, k]
```
navidcy: does the @inbounds outside the function propagate inside? Otherwise we need 

```suggestion
    cᵢⱼ = @inbounds c[i, j, k]
```

?

src/Advection/bounds_preserving_tracer_advection_operators.jl
```diff
@@ -24,91 +23,98 @@ end
...
     c₊ᴸ = θ * (c₊ᴸ - cᵢⱼ) + cᵢⱼ
     c₋ᴿ = θ * (c₋ᴿ - cᵢⱼ) + cᵢⱼ
 
-    return @inbounds Axᶠᶜᶜ(i+1, j, k, grid) * upwind_biased_product(u[i+1, j, k], c₊ᴸ, c₊ᴿ) -
-                     Axᶠᶜᶜ(i,   j, k, grid) * upwind_biased_product(u[i,   j, k], c₋ᴸ, c₋ᴿ)
-end
-
-@inline function bounded_tracer_flux_divergence_y(i, j, k, grid, advection::BoundsPreservingWENO, v, c)
+    Ax_ρuc⁺ = ℑxᶠᵃᵃ(i+1, j, k, grid, ρ) * Axᶠᶜᶜ(i+1, j, k, grid) * upwind_biased_product(u[i+1, j, k], c₊ᴸ, c₊ᴿ)
+    Ax_ρuc⁻ = ℑxᶠᵃᵃ(i,   j, k, grid, ρ) * Axᶠᶜᶜ(i,   j, k, grid) * upwind_biased_product(u[i,   j, k], c₋ᴸ, c₋ᴿ)
```
navidcy: Why is there a `ρ` in the variable name?

src/Advection/bounds_preserving_tracer_advection_operators.jl
```diff
@@ -24,91 +23,98 @@ end
...
     c₊ᴸ = θ * (c₊ᴸ - cᵢⱼ) + cᵢⱼ
     c₋ᴿ = θ * (c₋ᴿ - cᵢⱼ) + cᵢⱼ
 
-    return @inbounds Azᶜᶜᶠ(i, j, k+1, grid) * upwind_biased_product(w[i, j, k+1], c₊ᴸ, c₊ᴿ) -
-                     Azᶜᶜᶠ(i, j, k,   grid) * upwind_biased_product(w[i, j, k],   c₋ᴸ, c₋ᴿ)
-end
+    Az_ρwc⁺ = ℑzᵃᵃᶠ(i, j, k+1, grid, ρ) * Azᶜᶜᶠ(i, j, k+1, grid) * upwind_biased_product(w[i, j, k+1], c₊ᴸ, c₊ᴿ)
+    Az_ρwc⁻ = ℑzᵃᵃᶠ(i, j, k,   grid, ρ) * Azᶜᶜᶠ(i, j, k,   grid) * upwind_biased_product(w[i, j, k],   c₋ᴸ, c₋ᴿ)
 
+    return Az_ρwc⁺ - Az_ρwc⁻
```
navidcy: why `ρ`?

src/Advection/bounds_preserving_tracer_advection_operators.jl
giordano: I don't think the two forms are entirely equivalent:
```julia-repl
julia> @inbounds f(x) = x[1]

julia> g(x) = @inbounds x[1]
g (generic function with 1 method)

julia> code_llvm(f, (Vector{Float64},))
```
```llvm
; Function Signature: f(Array{Float64, 1})
;  @ REPL[7]:1 within `f`
define double @julia_f_1367(ptr noundef nonnull align 8 dereferenceable(24) %"x::Array") local_unnamed_addr #0 {
top:
; ┌ @ essentials.jl:964 within `getindex`
; │┌ @ essentials.jl:385 within `checkbounds` @ essentials.jl:381
; ││┌ @ essentials.jl:11 within `length`
; │││┌ @ essentials.jl:10 within `size`
      %"x::Array.size_ptr" = getelementptr inbounds nuw i8, ptr %"x::Array", i64 16
      %"x::Array.size.0.copyload" = load i64, ptr %"x::Array.size_ptr", align 8
; ││└└
    %.not = icmp eq i64 %"x::Array.size.0.copyload", 0
; ││ @ essentials.jl:385 within `checkbounds`
    br i1 %.not, label %L11, label %L15

L11:                                              ; preds = %top
    call void @j_throw_boundserror_1370(ptr nonnull %"x::Array", ptr nocapture nonnull readonly @"_j_const#1") #10
    unreachable

L15:                                              ; preds = %top
; │└
; │ @ essentials.jl:965 within `getindex`
   %memoryref_data = load ptr, ptr %"x::Array", align 8
   %0 = load double, ptr %memoryref_data, align 8
   ret double %0
; └
}
```
```julia-repl
julia> code_llvm(g, (Vector{Float64},))
```
```llvm
; Function Signature: g(Array{Float64, 1})
;  @ REPL[8]:1 within `g`
define double @julia_g_1371(ptr noundef nonnull align 8 dereferenceable(24) %"x::Array") local_unnamed_addr #0 {
top:
; ┌ @ essentials.jl:965 within `getindex`
   %memoryref_data = load ptr, ptr %"x::Array", align 8
   %0 = load double, ptr %memoryref_data, align 8
   ret double %0
; └
}
```
While `f` uses `getelementptr inbounds`, it doesn't remove the error path entirely.  But the body of `f` is the same as the body a function without any `@inbounds` annotation
```julia-repl
julia> h(x) = x[1]
h (generic function with 1 method)

julia> code_llvm(h, (Vector{Float64},))
```
```llvm
; Function Signature: h(Array{Float64, 1})
;  @ REPL[14]:1 within `h`
define double @julia_h_1633(ptr noundef nonnull align 8 dereferenceable(24) %"x::Array") local_unnamed_addr #0 {
top:
; ┌ @ essentials.jl:964 within `getindex`
; │┌ @ essentials.jl:385 within `checkbounds` @ essentials.jl:381
; ││┌ @ essentials.jl:11 within `length`
; │││┌ @ essentials.jl:10 within `size`
      %"x::Array.size_ptr" = getelementptr inbounds nuw i8, ptr %"x::Array", i64 16
      %"x::Array.size.0.copyload" = load i64, ptr %"x::Array.size_ptr", align 8
; ││└└
    %.not = icmp eq i64 %"x::Array.size.0.copyload", 0
; ││ @ essentials.jl:385 within `checkbounds`
    br i1 %.not, label %L11, label %L15

L11:                                              ; preds = %top
    call void @j_throw_boundserror_1636(ptr nonnull %"x::Array", ptr nocapture nonnull readonly @"_j_const#1") #10
    unreachable

L15:                                              ; preds = %top
; │└
; │ @ essentials.jl:965 within `getindex`
   %memoryref_data = load ptr, ptr %"x::Array", align 8
   %0 = load double, ptr %memoryref_data, align 8
   ret double %0
; └
}
```
so I'm not sure the outer `@inbounds` is doing anything at all.  I must say I had never seen it before.

navidcy: `@propagate_inbounds` perhaps?

navidcy: no, I take that back..

giordano: That relies on an outer function calling this one with `@inbounds`, I'm not sure it's what it's wanted.

src/Advection/bounds_preserving_tracer_advection_operators.jl
glwagner: Because it is multiplied by `ρ`

src/Advection/bounds_preserving_tracer_advection_operators.jl
glwagner: Because it is multiplied by `ρ`

src/Advection/bounds_preserving_tracer_advection_operators.jl
glwagner: Umm dammit.

glwagner: We haven't figured out what is best regarding propagate_inbounds / where it is correct to invoke inbounds. The problem with propagate inbounds is that you have to "remember" if there is an inbounds hidden inside somewhere.

giordano: You could do something like
```julia
function f(x)
    @inbounds let
        x[1]
    end
end
```
and that'd work (it'd apply to the entire `let`/`begin` block annotated with `@inbounds`), but the `@inbounds` in front of the function definition doesn't do anything.  But also, `@inbounds` is best used locally, than with "global" annotations 🙂

src/Advection/bounds_preserving_tracer_advection_operators.jl
navidcy: Damn! I missed that ;)

src/Advection/bounds_preserving_tracer_advection_operators.jl
navidcy: I missed it!

src/Advection/bounds_preserving_tracer_advection_operators.jl
glwagner: i switched back to local annotations in https://github.com/CliMA/Oceananigans.jl/pull/5002/commits/c2be67ac32ebad9a5f9fdff0fec567c4cc9e225e

src/Advection/bounds_preserving_tracer_advection_operators.jl
simone-silvestri: Why is it negative? Should it be the actual flux divergence, no? The negative of it is taken in the tendency kernels.

simone-silvestri:
> Note, as far as I can tell, the fact that this is implemented differently is not strictly necessary. However, it avoids repeating some computations? @simone-silvestri if you can confirm this motivation for writing the code this way, I will add a note.

Right, the fact is that we need fluxes at both cell-faces to compute the flux on one face, so we can compute flux at i and i+1 in one go. Otherwise, we would need to recompute both fluxes at i and i+1 two times.

src/Advection/bounds_preserving_tracer_advection_operators.jl
glwagner: these functions should output 

$$ \nabla \cdot \left ( u c \right ) $$

The tendency requires the negative of the flux divergence,

$$ G = - \nabla \cdot \left ( u c \right ) + \cdots $$

simone-silvestri: Hmmm, I am not sure what you mean. Is this the negative of the flux divergence? These functions should indeed output $\nabla \cdot (uc)$ while the negative is taken in the tendency functions
https://github.com/CliMA/Oceananigans.jl/blob/6106d8abd470318dd862dff0abf1e624ea57bd4d/src/Models/NonhydrostaticModels/nonhydrostatic_tendency_kernel_functions.jl#L292

glwagner: Ok, I was confused and now I think this makes sense.

glwagner:
@simone-silvestri do you approve this PR?

glwagner:
I think the test failures are unrelated, so I will merge after this next round

PR #5006, opened 2025-12-01
Move `prettysummary` to Utils

giordano:
Also, move some methods from `Utils` to `Grids`, as they need types or functions defined in that submodule.  Ref: https://github.com/CliMA/ClimaOcean.jl/pull/688#issuecomment-3594140493 and https://github.com/NumericalEarth/Breeze.jl/pull/195#discussion_r2575386683.

I wish I had done this before, but it didn't occur to me that I could just move some methods from `Utils` to `Grids` :disappointed: However this change should be non-breaking for downstream packages, as `prettysummary` is accessible from both `Grids` and `Utils`.

Other functions moved around are

* [`periphery_offset`](https://juliahub.com/ui/Search?type=symbols&q=periphery_offset&t=function)
* [`interior_work_layout`](https://juliahub.com/ui/Search?type=symbols&q=interior_work_layout&t=function) (but this is still nominally in `Utils`, but extended in `Grids`)
* [`longitude_in_same_window`](https://juliahub.com/ui/Search?type=symbols&q=longitude_in_same_window&t=function)
* [`flip_location`](https://juliahub.com/ui/Search?type=symbols&q=flip_location&t=function)
* [`get_longitude_vertices`](https://juliahub.com/ui/Search?type=symbols&q=get_longitude_vertices&t=function)
* [`get_latitude_vertices`](https://juliahub.com/ui/Search?type=symbols&q=get_latitude_vertices&t=function)
* [`get_lat_lon_nodes_and_vertices`](https://juliahub.com/ui/Search?type=symbols&q=get_lat_lon_nodes_and_vertices&t=function)
* [`get_cartesian_nodes_and_vertices`](https://juliahub.com/ui/Search?type=symbols&q=get_cartesian_nodes_and_vertices&t=function)

but according to JuliaHub search (I included the links above) they are used exclusively by Oceananigans, among all packages registered in General, so no downstream package should be affected by these moves.

### DIFF
--- a/src/Grids/grid_utils.jl
+++ b/src/Grids/grid_utils.jl
@@ -437,14 +437,12 @@ end
 ##### Extensions for kernel launching
 #####
 
-import Oceananigans.Utils: periphery_offset
-
-function periphery_offset(loc, grid::AbstractGrid, side::Int)
+function Utils.periphery_offset(loc, grid::AbstractGrid, side::Int)
     T = topology(grid, side)
     N = size(grid, side)
 
-    return periphery_offset(loc, T(), N)
+    return Utils.periphery_offset(loc, T(), N)
 end
 
 # Other cases are already covered by the fallback in Oceananigans.Utils
-periphery_offset(::Face, ::Bounded, N::Int) = ifelse(N > 1, 1, 0)
+Utils.periphery_offset(::Face, ::Bounded, N::Int) = ifelse(N > 1, 1, 0)


src/Grids/grid_utils.jl
```diff
@@ -431,3 +432,59 @@ function add_halos(data::AbstractArray{FT, 2} where FT, loc, topo, sz, halo_sz;
...
+The `workgroup` is a tuple specifying the threads per block in each
+dimension. The `worksize` specifies the range of the loop in each dimension.
+
+Specifying `include_right_boundaries=true` will ensure the work layout includes the
+right face end points along bounded dimensions. This requires the field `location`
+to be specified.
+
+For more information, see: https://github.com/CliMA/Oceananigans.jl/pull/308
+"""
+@inline function Utils.interior_work_layout(grid, workdims::Symbol, (LX, LY, LZ))
```
simone-silvestri: This is quite an important function to be buried in the `grid_utils.jl` file. I would probably separate it out in its own file.

giordano: Suggestions for the name of the file? :grin:

simone-silvestri: Actually, can this not be kept in the Utils module? Note that you are not extending this function with types defined in the Grids module.

giordano: `topology` is defined in `Grids` (and that seems to be a sensible module for this function?).

simone-silvestri: > Actually, can this not be kept in the Utils module? Note that you are not extending this function with types defined in the Grids module.

ok unfortunately not I guess 😞

simone-silvestri: What do you think about keeping the interior_work_layout function in utils, near the kernel launching, and extend `periphery_offset` in the grids module, which is a simpler function?
We could do something like
```julia
ox = periphery_offset(ℓx, grid, 1)
oy = periphery_offset(ℓy, grid, 2)
oz = periphery_offset(ℓz, grid, 3)
```
and then extend `periphery_offset` in grids with 
```julia 
periphery_offset(loc, grid::AbstractGrid, side) = periphery_offset(loc, topology(grid, side), size(grid, side))
```
so that at least we are extending a method with a type defined in the correct module

simone-silvestri: then `periphery_offset` can actually fallback to 0, so that in the `Utils` module we have
```julia
periphery_offset(loc, grid, side) = 0
```

giordano: Sounds good!  Do you want to give it a go yourself? :grinning:

simone-silvestri: Sure

simone-silvestri:
Ups, I think I merged by mistake #4992 in this branch 😅

giordano:
I did some git surgery to revert that :grinning:

PR #5009, opened 2025-12-01
Change default minimum buffer upwind order to 3

glwagner:
Following up on #5003. @simone-silvestri

### DIFF
--- a/src/Advection/bounds_preserving_tracer_advection_operators.jl
+++ b/src/Advection/bounds_preserving_tracer_advection_operators.jl
@@ -10,11 +10,12 @@ const BoundsPreservingWENO = WENO{<:Any, <:Any, <:Any, <:Tuple}
 @inline div_Uc(i, j, k, grid, advection::BoundsPreservingWENO, U, ::ZeroField) = zero(grid)
 
 # Is this immersed-boundary safe without having to extend it in ImmersedBoundaries.jl? I think so... (velocity on immmersed boundaries is masked to 0)
+# For bounds preserving advection, we need fluxes at both cell-faces to compute the flux on one face.
+# So we extend div_Uc in order to compute the fluxes at i and i+1 in one go and avoid recomputation.
 @inline function div_Uc(i, j, k, grid, advection::BoundsPreservingWENO, U, c)
-
-    div_x = bounded_tracer_flux_divergence_x(i, j, k, grid, advection, U.u, c)
-    div_y = bounded_tracer_flux_divergence_y(i, j, k, grid, advection, U.v, c)
-    div_z = bounded_tracer_flux_divergence_z(i, j, k, grid, advection, U.w, c)
+    div_x = bounded_tracer_flux_divergence_x(i, j, k, grid, advection, 1, U.u, c)
+    div_y = bounded_tracer_flux_divergence_y(i, j, k, grid, advection, 1, U.v, c)
+    div_z = bounded_tracer_flux_divergence_z(i, j, k, grid, advection, 1, U.w, c)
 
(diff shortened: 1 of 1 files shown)


simone-silvestri:
I am a bit afraid to change the behavior for the WENO vector invariant here. Maybe we can add a `minimum_buffer_upwind_order = 1` in the `WENOVectorInvariant` constructor to preserve behavior there?

glwagner:
> I am a bit afraid to change the behavior for the WENO vector invariant here. Maybe we can add a `minimum_buffer_upwind_order = 1` in the `WENOVectorInvariant` constructor to preserve behavior there?

Why?

simone-silvestri:
Because we haven't tested it, in general, a centered scheme should be unstable, so we are trading here possible instabilities for less dissipation. The tracer is probably not an issue (also because I have tested it quite a lot now), but the momentum scheme might not be able to hold a centered scheme and might crash all the downstream work, for example, for ClimaOcean users.

glwagner:
> Because we haven't tested it, in general, a centered scheme should be unstable, so we are trading here possible instabilities for less dissipation. The tracer is probably not an issue (also because I have tested it quite a lot now), but the momentum scheme might not be able to hold a centered scheme and might crash all the downstream work, for example, for ClimaOcean users.

Hmm ok, I was wondering if you had specifically seen something. It does make sense to me that this is a bigger issue for tracers than momentum; almost always we use drag laws which means that momentum needs to be mixed. The problem with 1st order upwind is that it will mix away a gradient spuriously against insulating boundaries (even in nearly quiescent flow)

glwagner:
@simone-silvestri I made it so that WENOVectorInvariant goes to first order at the boundary, but plain WENO will limit to Centered(order=2).

simone-silvestri (approved):

glwagner:
@tomchor looks like a test you added has intermittent failures: https://buildkite.com/clima/oceananigans/builds/27382/steps/canvas?jid=019afed3-278c-481d-bc8c-08ed6a1ae16c#019afed3-278c-481d-bc8c-08ed6a1ae16c/17-1274

tomchor:
> @tomchor looks like a test you added has intermittent failures: [buildkite.com/clima/oceananigans/builds/27382/steps/canvas?jid=019afed3-278c-481d-bc8c-08ed6a1ae16c#019afed3-278c-481d-bc8c-08ed6a1ae16c/17-1274](https://buildkite.com/clima/oceananigans/builds/27382/steps/canvas?jid=019afed3-278c-481d-bc8c-08ed6a1ae16c#019afed3-278c-481d-bc8c-08ed6a1ae16c/17-1274)

Hmm, apparently the error isn't the same across runs, so it needs a higher tolerance for the `approx` call. I'm too busy today but I'll open a PR to fix this soon (probably tomorrow). Feel free to make that change yourself if it's easier.

glwagner:
I can fix, just wanted to inform!

test/test_forcings.jl
```diff
@@ -329,7 +329,9 @@ function test_settling_tracer_comparison(arch; open_bottom=true)
         @test (regular_integral[] / regular_initial_integral[]) < 1e-3
         @test (immersed_integral[] / immersed_initial_integral[]) < 1e-3
     else
-        @test regular_integral[] ≈ immersed_integral[] ≈ regular_initial_integral[]
+        # Mass is approximately conserved, with some numerical diffusion
+        @test isapprox(regular_integral[], regular_initial_integral[], rtol=1e-3)
+        @test isapprox(immersed_integral[], immersed_initial_integral[], rtol=1e-3)
```
tomchor, line 334: :heart: thank you! I was about to fix this :)

glwagner: 'course

PR #5010, opened 2025-12-01
Disambiguate binary operations on Numbers

glwagner:
Found this bug when building a BinaryOperation --- would be nice to test this. I believe it requires using `@at` in a BinaryOperation that involves multiplying numbers.

### DIFF
--- a/src/AbstractOperations/binary_operations.jl
+++ b/src/AbstractOperations/binary_operations.jl
@@ -115,7 +115,7 @@ function define_binary_operator(op)
         end
 
         # Numbers are not fields...
-        $op(Lc::Tuple, a::Number, b::Number) = $op(a, b)
+        $op(Lc::Tuple{<:$Location, <:$Location, <:$Location}, a::Number, b::Number) = $op(a, b)
 
         # Sugar for mixing in functions of (x, y, z)
         $op(Lc::Tuple{<:$Location, <:$Location, <:$Location}, f::Function, b::AbstractField) = $op(Lc, FunctionField(location(b), f, b.grid), b)


giordano:
@simone-silvestri this PR alone resolves 5 ambiguities detected by Aqua out of 358.  Only 353 to go 😂

giordano:
BTW, of the 353 remaining ambiguous methods, 100 are in this module alone (`Test.detect_ambiguities(Oceananigans.AbstractOperations; recursive=true)`), maybe there are many similar cases of unbound tuples?

glwagner:
Oh if Aqua can find this without needing an explicit test, that is very nice!

giordano:
You can find all of them with https://juliatesting.github.io/Aqua.jl/stable/ambiguities/#Aqua.test_ambiguities
```julia
using Oceananigans, Aqua
Aqua.test_ambiguities(Oceananigans)
```
But this errors out mercilessly if there are ambiguities. One thing we could do to at least not increase ambiguities in this module is
```julia
using Oceananigans, Test
@test length(Test.detect_ambiguities(Oceananigans.AbstractOperations)) <= 100
```

glwagner:
Can we find ambiguities and eliminate them ?

glwagner:
```
using Oceananigans, Test
@test length(Test.detect_ambiguities(Oceananigans.AbstractOperations)) <= 100
```

^^ @giordano this doesn't require Aqua?

giordano:
> Can we find ambiguities

Yes, as I said above, `Aqua.test_ambiguities(Oceananigans)` lists all of them.

>  and eliminate them ?

That'd be great, only problem is that they're _a lot_.

> ^^ @giordano this doesn't require Aqua?

No, that's just using the `Test` stdlib.  Note we're already using Aqua for other things: https://github.com/CliMA/Oceananigans.jl/blob/0217789f82583f942b77d9a105ebfb24aa2b8ee6/test/test_quality_assurance.jl#L8 but we're skipping the ambiguities check because it'd fail.

glwagner:
ill just merge this for now. there's more work to do.

PR #5015, opened 2025-12-05
Explicit imports in more modules

giordano:
More follow up to #4985.


giordano:
This is good to go, it's big enough 🫠 I know it's not fun to review these PRs, but the good news is that if I imported too much, ExplicitImports would bark at us, the only risk is that I imported too little, but that'd manifest as an undefined variable error (if the offending code is covered by tests)

simone-silvestri:
Some issues with the distributed tests. I ll restart them

giordano:
Can you please restart also the documentation job? As far as I cen tell it only failed to push the docs.

simone-silvestri (approved):

giordano:
https://buildkite.com/clima/oceananigans/builds/27400#019b02a8-5309-4941-8a6a-0e5673dcbe3d/31-37
```
/bin/bash: line 1: cd: /var/lib/buildkite-agent/Oceananigans.jl-27400: No such file or directory
```
Do we need to restart the whole job? :disappointed:

PR #5017, opened 2025-12-05
Make `FieldTimeSeries` work with NetCDF output

tomchor:
This PR implements NetCDF supoprt for `FieldTimeSeries`. In the process it also reorganizes the code in `OceananigansNCDasasetsExt` a bit (but not the logic; I just broke it apart into different files).

Closes https://github.com/CliMA/Oceananigans.jl/issues/3935

To-do list (not all of this for this PR but I'll do what I can):
- [x] Make `FieldTimeSeries` work with `InMemory()` backend
- [ ] Make `OnDisk()` backend work
- [ ] Make `FieldDataset` work
- [ ] Make chunked `FieldTimeSeries` work
- [ ] Allow reading `FieldBoundaryCondition`s from `NetCDF` output
- [ ] Allow for NetCDF output `with_halos` and non-default `indices` (why isn't that allowed?)
- [x] Test output readers with `NetCDFWriter`

Necessary NetCDFWriter enhancements:

- [x] Write `Field` location to NetCDF files along `Field`s
- [x] Write `Field` indices to NetCDF
- [ ] ~Write boundary conditions to NetCDF somehow~


tomchor:
@glwagner This PR will require me to somehow store boundary conditions on the NetCDF files. Can you shed some light on how this is done for JLD2? I remember there are some limitations for this even for with the JLD2Writer, no?

glwagner:
> @glwagner This PR will require me to somehow store boundary conditions on the NetCDF files. Can you shed some light on how this is done for JLD2? I remember there are some limitations for this even for with the JLD2Writer, no?

I'd suggest not supporting that to start since it might be a little complex. Even for JLD2, we cannot support serializing functions. Also, geophysical applications are almost exclusively based on flux boundary conditions which are not useful in post-processing.

tomchor:
I think this PR is ready to review. As you can see from the top comment, there are quite a few things still to support, but I think we should merge this "smaller" NetCDF support first and then I'll open an issue to track all those to-dos and tackle them in different PRs.

ext/OceananigansNCDatasetsExt/OceananigansNCDatasetsExt.jl
```diff
@@ -0,0 +1,84 @@
...
+    NoFileSplitting,
+    update_file_splitting_schedule!,
+    construct_output,
+    time_average_outputs,
+    restrict_to_interior,
+    fetch_output,
+    convert_output,
+    fetch_and_convert_output,
+    show_array_type
+using Oceananigans.OutputReaders: InMemoryFTS, time_indices
```
glwagner, line 61:
```suggestion
    show_array_type
    
using Oceananigans.OutputReaders: InMemoryFTS, time_indices
```

totally trivial comment but can you add some whitespace to this file? I think it makes it more readable. I think its best to have whitespace following multiline code blocks .

ext/OceananigansNCDatasetsExt/grid_reconstruction.jl
```diff
@@ -0,0 +1,268 @@
...
+
+gather_grid_metrics(grid::ImmersedBoundaryGrid, args...) = gather_grid_metrics(grid.underlying_grid, args...)
+
+# TODO: Proper masks for 2D models?
+flat_loc(T, L) = T == Flat ? nothing : L
+
+const PCBorGFBIBG = Union{GFBIBG, PCBIBG}
+
+# For Immersed Boundary Grids (IBG) with either a Grid Fitted Bottom (GFB) or a Partial Cell Bottom (PCB)
+function gather_immersed_boundary(grid::PCBorGFBIBG, indices, dim_name_generator)
```
glwagner, line 192: It's not completely clear to me what the purpose of this function is. It looks like it builds operations for peripheral nodes and bottom height? Why is that the only information that is needed for the immersed boundary? Why just those four locations?

ext/OceananigansNCDatasetsExt/grid_reconstruction.jl
```diff
@@ -0,0 +1,268 @@
...
+const GFBoundaryIBG = ImmersedBoundaryGrid{<:Any, <:Any, <:Any, <:Any, <:Any, <:GridFittedBoundary}
+
+# For Immersed Boundary Grids (IBG) with a Grid Fitted Boundary (also GFB!)
+function gather_immersed_boundary(grid::GFBoundaryIBG, indices, dim_name_generator)
+    op_peripheral_nodes_ccc = KernelFunctionOperation{Center, Center, Center}(peripheral_node, grid, Center(), Center(), Center())
+    op_peripheral_nodes_fcc = KernelFunctionOperation{Face, Center, Center}(peripheral_node, grid, Face(), Center(), Center())
+    op_peripheral_nodes_cfc = KernelFunctionOperation{Center, Face, Center}(peripheral_node, grid, Center(), Face(), Center())
+    op_peripheral_nodes_ccf = KernelFunctionOperation{Center, Center, Face}(peripheral_node, grid, Center(), Center(), Face())
+
+    return Dict("peripheral_nodes" => Field(grid.immersed_boundary.mask; indices),
```
glwagner: This looks like it returns the immersed cells, not the peripheral nodes?

ext/OceananigansNCDatasetsExt/grid_reconstruction.jl
```diff
@@ -0,0 +1,268 @@
...
+    grid_metadata = Dict(:immersed_boundary_type => nothing,
+                         :underlying_grid_type => underlying_grid_type)
+    return underlying_grid_args, underlying_grid_kwargs, immersed_grid_args, grid_metadata
+end
+
+function netcdf_grid_constructor_info(grid::ImmersedBoundaryGrid)
+    underlying_grid_args, underlying_grid_kwargs, immersed_grid_args = constructor_arguments(grid)
+
+    immersed_boundary_type = typeof(grid.immersed_boundary).name.wrapper |> string # Save type of immersed boundary for reconstruction
+    underlying_grid_type   = typeof(grid.underlying_grid).name.wrapper |> string # Save type of underlying grid for reconstruction
```
glwagner: Do we need to build a more codified infrastructure for this? For example:

```julia
netcdf_string(obj) = typeof(obj).name.wrapper |> string # fallback, but can be extended
build_from_netcdf(str, args) = build_from_netcdf(Val(Symbol(str)), args)
```

then for example for a given type we would have something like

```julia
function build_from_netcdf(Val{:PartialCellBottom}, args)
    # etc
    return obj
end
```

This implements an interface that can also be extended for custom types, so its generalizable.

ext/OceananigansNCDatasetsExt/grid_reconstruction.jl
```diff
@@ -0,0 +1,268 @@
...
+
+    elseif immersed_boundary_type == PartialCellBottom
+        bottom_height = Array(ibg_group["bottom_height"])
+        minimum_fractional_cell_height = ibg_group.attrib["minimum_fractional_cell_height"] |> materialize_from_netcdf
+        immersed_boundary = immersed_boundary_type(bottom_height, minimum_fractional_cell_height)
+
+    elseif immersed_boundary_type == GridFittedBoundary
+        mask = Array(ibg_group["mask"])
+        immersed_boundary = immersed_boundary_type(mask)
+
```
glwagner: I think this is fine because the immersed boundary is relatively static, but with an interface like mentioned above we don't need if-statements here

ext/OceananigansNCDatasetsExt/output_readers.jl
```diff
@@ -0,0 +1,212 @@
...
+using Oceananigans.BoundaryConditions: FieldBoundaryConditions
+
+import Oceananigans.Fields: Field
+
+
+#####
+##### FieldTimeSeries from NetCDF
+#####
+
+function FieldTimeSeries_from_netcdf(path::String, name::String;
```
glwagner: Why don't we implement something a little cleaner. We need to change the source code so that a type signififies whether the filename is netcdf or jld2, something like

```julia
struct JLD2Path
    path :: String
end

struct NetCDFPath
    path :: String
end
```

Then the constructor for `FieldTimeSeries(path::String, name::String)` first checks to see what filetype:

```julia
function FieldTimeSeries(path::String, args...; kw...)
    typed_path = if ext(path) == ".jld2"
        JLD2Path(path)
    elseif ext(path) == ".nc"
        NetCDFPath(path)
    end

    return FieldTimeSeries(typed_path, args...; kw...)
end
```

Now, the constructor for `FieldTimeSeries(path::NetCDFPath, args...; kw...)` can be implemented in the extension.

ext/OceananigansNCDatasetsExt/OceananigansNCDatasetsExt.jl
tomchor: Done

src/OutputReaders/field_time_series.jl
```diff
@@ -456,6 +456,107 @@ end
 
 struct UnspecifiedBoundaryConditions end
 
+"""
+    reconstruct_legacy_grid(grid, file, architecture)
+
+Reconstruct a grid from legacy JLD2 output files (prior to Oceananigans 0.95.0)
+that did not serialize grids properly.
+"""
+function reconstruct_legacy_grid(grid, file, architecture)
```
simone-silvestri, line 465: I guess when we launch version 1.0 we can remove this code.

src/OutputReaders/field_time_series.jl
```diff
@@ -718,18 +743,45 @@ function FieldTimeSeries(path::String, name::String;
     return time_series
 end
 
+# Stub function for NetCDF files - will be extended by OceananigansNCDatasetsExt
+FieldTimeSeries_from_netcdf(path::String, args...; kwargs...) = error("Loading FieldTimeSeries from NetCDF files requires NCDatasets")
```
simone-silvestri:
```suggestion
field_time_series_from_netcdf(path::String, args...; kwargs...) = error("Loading FieldTimeSeries from NetCDF files requires NCDatasets")
```

src/OutputReaders/field_time_series.jl
```diff
@@ -718,18 +743,45 @@ function FieldTimeSeries(path::String, name::String;
...
+
+function FieldTimeSeries(path::String, args...; reader_kw = NamedTuple(), kwargs...)
+    path = auto_extension(path, ".jld2") # JLD2 is the default extension
+
+    if endswith(path, ".nc")
```
simone-silvestri: will this work? we are adding a `.jld2` in the line above

src/OutputReaders/field_time_series.jl
```diff
@@ -718,18 +743,45 @@ function FieldTimeSeries(path::String, name::String;
...
+        return FieldTimeSeries_from_netcdf(path, args...; reader_kw, kwargs...)
```
simone-silvestri:
```suggestion
        return field_time_series_from_netcdf(path, args...; reader_kw, kwargs...)
```

src/OutputReaders/field_time_series.jl
tomchor: `auto_extension()` only adds `.jld2` if the file name doesn't have an extension.

src/OutputReaders/set_field_time_series.jl
```diff
@@ -20,11 +20,22 @@ end
 
 find_time_index(time::AbstractTime, file_times, Δt) = findfirst(t -> t == time, file_times)
 
-function set!(fts::InMemoryFTS, path::String=fts.path, name::String=fts.name; warn_missing_data=true)
-    file = jldopen(path; fts.reader_kw...)
+set_from_netcdf!(fts, path::String, args...; kwargs...) = error("Setting FieldTimeSeries from NetCDF files requires NCDatasets")
```
simone-silvestri, line 24: maybe we can add a comment:
```julia
# Extended in the OceananigansNCDatasetsExt
```

src/OutputReaders/set_field_time_series.jl
```diff
@@ -20,11 +20,22 @@ end
...
+
+function set!(fts::InMemoryFTS, path::String=fts.path, args...; kwargs...)
+    if endswith(path, ".jld2")
+        file = jldopen(path; fts.reader_kw...)
+        set!(fts, file, args...; kwargs...)
```
simone-silvestri, line 29:
```suggestion
        set!(fts, file, args...; kwargs...)
        close(file)
```

src/OutputReaders/field_time_series.jl
tomchor: Done

src/OutputReaders/set_field_time_series.jl
tomchor: Good idea. Done

src/OutputReaders/field_time_series.jl
tomchor: Done

simone-silvestri:
I think both the chunked backend and the OnDisk backend will be very very clean to implement with NetCDF, since it supports chunking the underlying data and typically netcdf is a lazy load. Not necessary to implement it here, just wanted to leave this comment.

ext/OceananigansNCDatasetsExt/grid_reconstruction.jl
tomchor: Thanks for catching that! Fixed

ext/OceananigansNCDatasetsExt/grid_reconstruction.jl
tomchor: This function is just a quality-of-life function for NetCDF outputs. It gathers these variables so that they can be written into the NetCDF file for any users using any language to have easy access to. The peripheral node locations that are there here inherited from older code, but I think these are just the most common locations that users typically need (i.e. tracer and velocity nodes). Happy to include all locations though.

ext/OceananigansNCDatasetsExt/output_readers.jl
tomchor: Good idea. Done

tomchor:
All tests are passing except for Aqua's quality control, which fails with

```julia
Aqua: Test Failed at /home/tomas/repos/Oceananigans.jl2/test/test_quality_assurance.jl:11
  Expression: length(detect_ambiguities(Oceananigans; recursive = true)) <= 361
   Evaluated: 364 <= 361
```
~but I can't figure out what ambiguity it is. Anyone know how to get that info out of Aqua?~

nvmd I think I figured it out.

ext/OceananigansNCDatasetsExt/grid_reconstruction.jl
tomchor: Good idea. Done

ext/OceananigansNCDatasetsExt/grid_reconstruction.jl
tomchor: Removed if-else statements in favor of multiple dispatch

tomchor:
@glwagner @simone-silvestri thanks for the comments, I believe I addressed all of them. Could you please re-review? I'd like to merge this before more merge conflicts pop up.

ext/OceananigansNCDatasetsExt/output_readers.jl
```diff
@@ -0,0 +1,233 @@
...
+
+    isnothing(grid) && (grid = reconstruct_grid(file))
+
+
+    isnothing(location) && (location = file[name].attrib["location"] |> materialize_from_netcdf)
+    LX, LY, LZ = location
+    loc = (LX(), LY(), LZ())
+
+    @warn "Reading boundary conditions from NetCDF files is not supported for FieldTimeSeries. Using default FieldBoundaryConditions for `grid` and `location`."
+    boundary_conditions = FieldBoundaryConditions(grid, loc)
```
ali-ramadhan, line 54: This warning might get a bit noisy when loading multiple `FieldTimeSeries` but I can't think of a better way to warn the user.

If `boundary_conditions = FieldBoundaryConditions(grid, loc)` will always run then is there a reason for having the kwarg `boundary_conditions = UnspecifiedBoundaryConditions()`? From reading the code it seems like `UnspecifiedBoundaryConditions()` will never be used?

ext/OceananigansNCDatasetsExt/output_readers.jl
```diff
@@ -0,0 +1,233 @@
...
+
+function find_time_index(t, file_times, Δt)
+    # Find the index in file_times that is closest to t
+    for (i, file_time) in enumerate(file_times)
+        if abs(file_time - t) < Δt / 2
+            return i
+        end
+    end
+    return nothing
+end
```
ali-ramadhan, line 147: This function seems to assume regularly spaced time indices (in which case we can compute the index rather than doing a linear search). Should we be supporting arbitrary time indices? It would still be a linear search. I guess you shouldn't have arbitrary time indices because output writers write output on a schedule.

To be clear, I think this is fine as is! Just curious if anyone sees a case for arbitrary time indices.

tomchor: Yeah you're right. Technically we could have arbitrary time indices with `SpecifiedTimes`, but I think that's usually not used for this end so it's okay to keep it as is for now.

ext/OceananigansNCDatasetsExt/output_readers.jl
tomchor: Yes, you're right. I kept the `boundary_condition` kwargs (as well as some other ones that aren't supported) to keep the same method signature as the JLD2 method.

PR #5021, opened 2025-12-07
Use volume independent residual in `ConjugateGradientPoissonSolver` tolerance calculation

### LINKED ISSUE #5374: Default `reltol` and `abstol` for the `ConjugateGradientPoissonSolver` don't match docstring (and are perhaps too small)
In going over some simulations with @vhi3u I realized that the default `abstol` and `reltol` for the `ConjugateGradientPoissonSolver` is much lower than what is listed in the docstring:

https://github.com/CliMA/Oceananigans.jl/blob/c0ad19bff6651f512bedbb7cd6508376b4504590/src/Solvers/conjugate_gradient_poisson_solver.jl#L80-L102

For reference, for a Float64 `grid` I get `reltol = 5e-18` and `abstol = 2e-14`, which is much lower than the `sqrt(eps(grid)) = 1e-8` that we used to use (and that is listed on the docstring).

Before I open a PR I'd like to ask: is there a reason why we stopped using `sqrt(eps(grid))` and moved towards this more complex default? I tried a few different grid sizes and the current default tolerances seem pretty low...

xkykai:
### Update: closes #5374 

The changes in `abstol` and `reltol` started from https://github.com/CliMA/Oceananigans.jl/pull/4563 when I changed the `ConjugateGradientPoissonSolver` to solve the volume multiplied $V \nabla^2 \phi$ instead of $\nabla^2 \phi$. This is done so that the operator is symmetric under both uniform and non-uniform grids. The requirement for a symmetric operator is so that we can use the conjugate gradient algorithm.

However in scaling the operator the residual is now scaled by the cell volume, therefore a universal `abstol` and `reltol` will not work. I put in a heuristic to account for the cell volume dependence, but that has turned out to take many more iterations than generally needed, especially in cases where cell volumes are large. 

In trying to put in a fix I realized that we can actually compute the convergent criteria by scaling the cell volume back, so that the convergence criteria the same as pre https://github.com/CliMA/Oceananigans.jl/pull/4563, which is universal and non cell-volume dependent. This comes at a cost of only two kernel multiplications as I am scaling and descaling `residual` in place before and after computing `norm` for use with `tolerance`, so no extra memory is required.

This has shown to be working quite well. I've put in tests to ensure that this works for the future as well.

---
fix `reltol` and `abstol` in docstring


src/Solvers/conjugate_gradient_poisson_solver.jl
```diff
@@ -84,8 +84,8 @@ struct DefaultPreconditioner end
 """
     ConjugateGradientPoissonSolver(grid;
                                    preconditioner = DefaultPreconditioner(),
-                                   reltol = sqrt(eps(grid)),
-                                   abstol = sqrt(eps(grid)),
+                                   reltol = min(100 * eps(grid), 100 * eps(grid) * minimum_cell_volume(grid)^2, sqrt(eps(grid))),
+                                   abstol = min(100 * eps(grid), sqrt(eps(grid))),
```
glwagner: can we fix this so its more readable? something like `reltol = default_rel_tol`. And then describe the `reltol` later on in the docstring? Same for abstol.

tomchor (approved):
Nice solution!

xkykai:
incompressibility tests fail after relaxing the tolerance for the lat-lon grid:
https://github.com/CliMA/Oceananigans.jl/blob/c69b6eb282a03fbbf568095e25e4f53a68a2306b/test/test_nonhydrostatic_models.jl#L214-L217

@simone-silvestri should I reduce the default tolerance or relax the test tolerance? right now I am putting the defaults as 
```julia
reltol = sqrt(eps(grid)),
abstol = sqrt(eps(grid)),
```

simone-silvestri (approved):

tomchor:
How off are we from the tolerance? I'd say try decreasing the default abstol and reltol of the solver by a bit first (maybe multiply it by something O(0.1)?) and see if that does the trick

xkykai:
Using the configuration where `reltol = sqrt(eps(grid)), abstol = sqrt(eps(grid)),` the tests give
`tolerance = 2.220446049250313e-15, max w = 3.1676883338604966e-11` for `Float64`.

For `Float32`, we get `tolerance = 1.1920929e-6, max w = 3.8802624e-5`.

Note the `tolerance = 10 * eps(FT)`.

In my opinion for most immersed-boundary flows, `reltol = sqrt(eps(grid)), abstol = sqrt(eps(grid))` should be sufficient as you generally get divergences which are O(1e-9). So if folks are okay I can loosen the tolerance for this test which affects the lat-lon grid.

Alternatively I can also tighten the default tolerance to `reltol = 100 * eps(grid), abstol = 100 * eps(grid)` and change this test tolerance to match that as well.

tomchor:
> In my opinion for most immersed-boundary flows, `reltol = sqrt(eps(grid)), abstol = sqrt(eps(grid))` should be sufficient as you generally get divergences which are O(1e-9). So if folks are okay I can loosen the tolerance for this test which affects the lat-lon grid.

I'm okay with this. Although we should probably get someone else's opinion too. @simone-silvestri @glwagner @navidcy

tomchor:
Almost! Only one test not passing

Issue #4916, opened 2025-11-06
Do we need docs for `OpenBoundaryCondition` with the `PerturbationAdvection` scheme?


glwagner:
@kaiyuan-cheng and I were looking at the docs and we noticed there is no information about using open boundary conditions:

https://clima.github.io/OceananigansDocumentation/stable/model_setup/boundary_conditions/

however, this is an exported feature. Should we add an example there?

As a side note, we are hoping to rinse and repeat the algorithm for Oceananigans.NonhydrostaticModel over at Breeze: https://github.com/NumericalEarth/Breeze.jl

cc @tomchor @jagoosw

tomchor:
> [@kaiyuan-cheng](https://github.com/kaiyuan-cheng) and I were looking at the docs and we noticed there is no information about using open boundary conditions:
> 
> [clima.github.io/OceananigansDocumentation/stable/model_setup/boundary_conditions](https://clima.github.io/OceananigansDocumentation/stable/model_setup/boundary_conditions/)
> 
> however, this is an exported feature. Should we add an example there?

I think that'd be helpful! Otherwise people won't know it's a feature they can use.

> As a side note, we are hoping to rinse and repeat the algorithm for Oceananigans.NonhydrostaticModel over at Breeze: [NumericalEarth/Breeze.jl](https://github.com/NumericalEarth/Breeze.jl?rgh-link-date=2025-11-06T19%3A47%3A25.000Z)

That sounds good, but isn't that more impetus for splitting Oceananigans into different packages, some of which can be used by other models (such as Breeze)?

glwagner:
Not really. Breeze needs to depend on pretty much all of Oceananigans because it also reuses a lot of physics. But there are other applications like ClimaSeaIce that don't need the ocean models.

peitianyi23-sudo:
Really looking forward to some docs about PerturbationAdvection! Could anyone recommend some related papers or references for this method?

tomchor:
@jagoosw was the one that originally implemented it, so I'm not sure if there's a reference for it. That said, there's a short derivation in the docstring if you haven't checked it:

```
help?> PerturbationAdvection
search: PerturbationAdvection

  PerturbationAdvection(FT = defaults.FloatType;
                        outflow_timescale = Inf,
                        inflow_timescale = 0)

[...]

PR #5026, opened 2025-12-09
Adds docs for perturbation advection

jagoosw:
Addresses #4916 

Happy for others to edit but though I should start a draft for this.

@glwagner @tomchor

### DIFF
--- a/docs/src/models/boundary_conditions.md
+++ b/docs/src/models/boundary_conditions.md
@@ -365,7 +365,7 @@ As discussed in [the numerical description of open boundary conditions](@ref num
 on open boundaries to approximate the behaviour of the boundary nodes given the interior state
 and specified external conditions. For example if we want to specify an outflowing boundary
 with a mean velocity ``U=1`` and damp the exiting flow to this speed we can setup a
-[`PertubationAdvection`](@ref) open boundary:
+[`PerturbationAdvection`](@ref) open boundary:
 
 ```jldoctest
 julia> scheme = PerturbationAdvection(; outflow_timescale=10, inflow_timescale=1)


glwagner:
nice, great start! Let's also include an example of using it

jagoosw:
Perhaps I could change the internal waves example to be bounded in x:

https://github.com/user-attachments/assets/22c4faed-b54f-428a-b27e-8b9531629c9f

This just has a PertubationAdvection OBC with the x-phase speed as the value. The black is bounded gray is periodic.

tomchor:
@jagoosw thinking about it now, doesn't the derivation of `PerturbationAdvection` (which assumes an advection around a mean velocity) imply that there needs to be a mean velocity that's larger than the perturbations through the boundary?

glwagner:
@jagoosw I actually meant to put an example on the docs page.

As for a literate example though --- good idea! I propse we add a mountain wave case that also has bathymetry and uses @xkykai's new solver

jagoosw:
> @jagoosw I actually meant to put an example on the docs page.
> 

Oh yes that makes sense, I guess that would go on this page right: https://clima.github.io/OceananigansDocumentation/stable/models/boundary_conditions/#9.-A-random,-spatially-varying,-constant-in-time-temperature-flux-specified-by-an-array

> As for a literate example though --- good idea! I propse we add a mountain wave case that also has bathymetry and uses @xkykai's new solver

Oh yes thats a good idea, if we do this PR first then we can put together the example in a different PR.

glwagner:
> > @jagoosw I actually meant to put an example on the docs page.
> 
> Oh yes that makes sense, I guess that would go on this page right: https://clima.github.io/OceananigansDocumentation/stable/models/boundary_conditions/#9.-A-random,-spatially-varying,-constant-in-time-temperature-flux-specified-by-an-array

yes!

> > As for a literate example though --- good idea! I propse we add a mountain wave case that also has bathymetry and uses @xkykai's new solver
> 
> Oh yes thats a good idea, if we do this PR first then we can put together the example in a different PR.

yeah separate PR is good for sure

jagoosw:
> @jagoosw thinking about it now, doesn't the derivation of `PerturbationAdvection` (which assumes an advection around a mean velocity) imply that there needs to be a mean velocity that's larger than the perturbations through the boundary?

Yes that's true. 

I think when $U\not\gg u\prime$ the terms scale like:
- $Uu\prime/L + u\prime^2/L\sim u^2/L$ for the advection, 
- $\nu u\prime/L^2$ for diffusion, 
- and $u\prime/\tau$ for the damping 
where $L$ is the length scale that the velocity varies over. 

Then the damping term dominates over the diffusion term as long as $\tau\ll L^2/\nu$ which is likely to always be true even (e.g. in the wave field example above if $L\sim\Delta x\approx0.05$ and $\nu=10^{-6}$ then $\tau\ll 2500$) but would need to be checked for specific cases. 

Problems occur when $u\prime^2/L\not\ll u\prime/\tau$ which we can't always know before, then $\tau\not\ll L/u\prime$ then the $u\prime\partial_xu\prime$ term is important. Perhaps we need to switch to using $u\partial_xu$ not $U\partial_xu$ when $(1-\epsilon)U < u < (1+\epsilon)U$ where $\epsilon$ is some factor?

A case I think of where we know $u\prime$ is going to become significant is if we have an oscillating flow like a tide turning, but in that case $L/u$ is likely to be large since $u$ is overall small.

docs/src/models/boundary_conditions.md
```diff
@@ -359,6 +359,25 @@ FluxBoundaryCondition: 16×16 Matrix{Float64}
 
 When running on the GPU, `Q` must be converted to a `CuArray`.
 
+### 10. Open boundary condition with mathcing scheme
+
+As discussed in [here](@ref numerical_bcs) it is often neccessary to specify a matching scheme
```
glwagner:
```suggestion
As discussed in [the numerical description of open boundary conditions](@ref numerical_bcs) it is often neccessary to specify a matching scheme
```

(as a rule, don't write "here" with a link, instead describe the link)

francispoulin:
This all looks greate and I'm happy to read all about it.  Can someone point me to the pretty version of the new docs?

jagoosw:
> This all looks greate and I'm happy to read all about it. Can someone point me to the pretty version of the new docs?

The previews don't seem to be on the deploying(https://clima.github.io/OceananigansDocumentation/previews/PR5026/numerical_implementation/boundary_conditions/index.html ?) but this is the preview I get in VSCode from the markdown:
Screenshot 2025-12-11 at 12 30 21

docs/src/numerical_implementation/boundary_conditions.md
```diff
@@ -222,3 +222,49 @@ are typlically already unphysical and only used in an attempt to allow informati
...
+full equation of motion as gradients across the boundary can not be computed and simply prescribing
+a boundary normal velocity is unphysical and reflects energy leaving the domain [Orlanksi1976](@citep).
+
+### Perturbation advection 
+
+A common method for allowing information to exit is to perform a one-sided advection operation. 
+For example in the [Orlanksi1976](@citet) boundary condition fields are advected out of the domain by a 
+locally determined phase speed. We can show that this is the first order approximation of the full equations
+of motion in the predictor velocity step. Consider a right boundary normal to the `u` velocity 
+component (the east boundary):
```
tomchor: Nice work with the docs and derivation. One comment here is that start (and only mention) Orlanski in the `PerturbationAdvection` section, which isn't Orlanski. So I'd rephrase this to explain `PeturbationAdvection` without mentioning to the Orlanski BC, as it might give the wrong impression that this is somehow a version of Orlanski.

jagoosw: If I understand Orlanski correctly you wouldn't you use `PeturbationAdvection` and set `U` (with `update_boundary_condition!`) to the phase speed?

tomchor: I think Orlanski is a bit trickier than that, but that is part of my point. Namely that I think Orlanski shouldn't be mentioned here at all I think.

docs/src/models/boundary_conditions.md
```diff
@@ -359,6 +359,25 @@ FluxBoundaryCondition: 16×16 Matrix{Float64}
...
+### 10. Open boundary condition with mathcing scheme
+
+As discussed in [the numerical description of open boundary conditions](@ref numerical_bcs) it is often neccessary to specify a matching scheme
+on open boundaries to approximate the behaviour of the boundary nodes given the interior state
+and specificed external conditions. For example if we want to specify an outflowing boundary
+with a mean velocity ``U=1`` and damp the exiting flow to this speed we can setup a 
+`PertubationAdvection` open boundary:
+
+```jldoctest
+julia> scheme = PerturbationAdvection(; outflow_timescale=10, inflow_timescale=1)
```
glwagner, line 371: @jagoosw how did you choose these time-scales? How do you choose the time-scale in general?

docs/oceananigans.bib
```diff
@@ -1095,3 +1095,16 @@ @article {Chor2021
...
+@article{Orlanksi1976,
+	title = {A simple boundary condition for unbounded hyperbolic flows},
+	journal = {Journal of Computational Physics},
+	volume = {21},
+	number = {3},
+	pages = {251-269},
+	year = {1976},
+	issn = {0021-9991},
+	doi = {https://doi.org/10.1016/0021-9991(76)90023-1},
+	url = {https://www.sciencedirect.com/science/article/pii/0021999176900231},
```
navidcy:
```suggestion
	doi = {10.1016/0021-9991(76)90023-1},
```

navidcy: (appears better in the References section like this..)

docs/src/numerical_implementation/boundary_conditions.md
```diff
@@ -222,3 +222,49 @@ are typlically already unphysical and only used in an attempt to allow informati
...
+```math
+    \partial_tu + u\partial_xu + v\partial_yu + w\partial_zu = (\nabla\cdot\vec{\tau})_x + F,
+```
+let ``\vec{u} = \vec{U} + \vec{u}\prime`` with ``\vec{U} = U(x, y, z, t)\hat{x}`` where ``U`` is an externally
+determined ``background" wall normal flow in the proximity of the boundary, and assume that the 
+stress tensor gradient is small,
+```math
+    \partial_tu = -(U + u\prime)\partial_x(U+u\prime) - v\partial_y(U+u\prime) - w\partial_z(U + u\prime) + F,
+```
+then, taking only first order terms:
```
navidcy:
```suggestion
then, taking only first-order terms:
```

docs/src/numerical_implementation/boundary_conditions.md
```diff
@@ -222,3 +222,49 @@ are typlically already unphysical and only used in an attempt to allow informati
...
+A common method for allowing information to exit is to perform a one-sided advection operation. 
+For example in the [Orlanksi1976](@citet) boundary condition fields are advected out of the domain by a 
+locally determined phase speed. We can show that this is the first order approximation of the full equations
+of motion in the predictor velocity step. Consider a right boundary normal to the `u` velocity 
+component (the east boundary):
+```math
+    \partial_tu + u\partial_xu + v\partial_yu + w\partial_zu = (\nabla\cdot\vec{\tau})_x + F,
+```
+let ``\vec{u} = \vec{U} + \vec{u}\prime`` with ``\vec{U} = U(x, y, z, t)\hat{x}`` where ``U`` is an externally
+determined ``background" wall normal flow in the proximity of the boundary, and assume that the
```
navidcy:
```suggestion
determined ``background" wall-normal flow in the proximity of the boundary, and assume that the 
```

navidcy (approved):

navidcy:
@jagoosw I see this outstanding question by @glwagner. Otherwise I think this PR is ready?

glwagner:
I'm merging this!

PR #5038, opened 2025-12-10
Fix flaky test: `test_chunked_abstraction`

### LINKED ISSUE #5037: Flaky test: `test_chunked_abstraction` in `test_output_readers.jl`
This test seems to fail rather frequently but pass when the test group is retried in Buildkite. I would guesstimate a failure rate of 10-15% based on my experience from the past week?

Example of a failure: https://buildkite.com/clima/oceananigans/builds/27439#019b0a2f-4c6a-4737-96d6-d33d62d301b9/35-5093

https://github.com/CliMA/Oceananigans.jl/blob/b1535424fd74f83aa35d1002d3673facc3112036/test/test_output_readers.jl#L413-L431

ali-ramadhan:
`test_chunked_abstraction` will now be passed a filepath that points to a JLD2 that is guaranteed to exist.

Resolves #5037

### DIFF
--- a/test/test_output_readers.jl
+++ b/test/test_output_readers.jl
@@ -410,9 +410,9 @@ function test_field_time_series_reductions(filepath3d, Nt)
     return nothing
 end
 
-function test_chunked_abstraction(filepath)
-    fts = FieldTimeSeries(filepath, "c")
-    fts_chunked = FieldTimeSeries(filepath, "c"; backend=InMemory(2), time_indexing=Cyclical())
+function test_chunked_abstraction(filepath, name)
+    fts = FieldTimeSeries(filepath, name)
+    fts_chunked = FieldTimeSeries(filepath, name; backend=InMemory(2), time_indexing=Cyclical())
 
     for t in eachindex(fts.times)
         fts_chunked[t] == fts[t]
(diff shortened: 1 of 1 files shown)


test/test_output_readers.jl
```diff
@@ -621,8 +621,7 @@ end
 
     @testset "Test chunked abstraction" begin
         @info "  Testing Chunked abstraction..."
-        filepath = "testfile.jld2"
-        test_chunked_abstraction(filepath)
+        test_chunked_abstraction(filepath3d, "T")
```
glwagner, line 624: i guess there is a requirement that `filepath3d` have at least two time points right? so that `InMemory(2)` works?

we could just create a new file for this test, for extra credit

ali-ramadhan: I believe so. Based on the following code in `generate_some_interesting_simulation_data`, `filepath3d` should have 5 time points since `stop_time=2minutes` and `schedule = TimeInterval(30seconds)`.

https://github.com/CliMA/Oceananigans.jl/blob/b1535424fd74f83aa35d1002d3673facc3112036/test/test_output_readers.jl#L43-L67

glwagner: happy to merge this as is

PR #5047, opened 2025-12-12
Fill all the halos for a flux boundary condition

simone-silvestri:
this makes sure that also second derivatives are zero across the boundary and might solve a [this Reactant issue](https://github.com/EnzymeAD/Enzyme-JaX/issues/1740)

This will not work if we have two bounded directions one next to each other, in that case, we need to extend the parameters of the fill halo kernels to encompass also the corner regions. 

Therefore, let's see if this solves the issue otherwise we can extend the parameters.

cc @wsmoses @glwagner


glwagner:
I might be confused but is the change we need to change the size of the kernel, ie change this:

https://github.com/CliMA/Oceananigans.jl/blob/a47754d2526bcebc39bb38cd0b810665ec90c198/src/BoundaryConditions/fill_halo_regions.jl#L139

to something that extends all the way in the xy direction? (the config `:xy` is interpreted as 1:Nx, 1:Ny)

simone-silvestri:
Ideed, for a double bounded domain we should do that, but with this change a bounded + periodic halo corner will be filled because periodic boundaries go pick into the halo.

I ll change the parameters.

PR #5052, opened 2025-12-16
Change default to `hydrostatic_pressure_anomaly = nothing` for vertically periodic topologies

xkykai:
It would be a better default for users if we turn off hydrostatic pressure anomaly by default for vertically periodic topologies. Otherwise, there will be a pressure jump at the periodic "boundary" which is unexpected for most users who intend to run simulations in a vertically periodic setup.
I haven't yet thought of a simulation example where the old default would be useful.


xkykai:
@glwagner can you help rerun the tests or force merge? seems like it errored out due to technical issues

PR #5054, opened 2025-12-17
Show bounds in WENO summary plus validation

glwagner:


src/Advection/weno_reconstruction.jl
```diff
@@ -101,6 +100,11 @@ function WENO(FT::DataType=Oceananigans.defaults.FloatType, FT2::DataType=Float3
 
     mod(order, 2) == 0 && throw(ArgumentError("WENO reconstruction scheme is defined only for odd orders"))
 
+    if !isnothing(bounds)
+        bounds isa Tuple && throw(ArgumentError("bounds must be nothing or a tuple of two values"))
```
navidcy: This reads wrong... It was throwing the ArgumentError if the bounds was actually a Tuple.

How about:

```suggestion
        !(bounds isa Tuple{<:Number, <:Number}) && throw(ArgumentError("bounds must be nothing or a tuple of two values"))
```

glwagner: > This reads wrong... It was throwing the ArgumentError if the bounds was actually a Tuple.

how can we avoid putting the explicit Number?

navidcy: use `NTuple{2}`?

navidcy: [b341adf](https://github.com/CliMA/Oceananigans.jl/pull/5054/commits/b341adf366d65058b4c6b61b7d4764826e33dc9b)

navidcy: It's bit less strict, because if you give a tuple of 2 random types then you won't get the informative error message

glwagner: i'm confused

```julia
julia> b = (1, 2)
(1, 2)

julia> b isa Tuple
true
```

glwagner: oh yeah its supposed to be `||` not `&&`

src/Advection/weno_reconstruction.jl
```diff
@@ -101,12 +100,17 @@ function WENO(FT::DataType=Oceananigans.defaults.FloatType, FT2::DataType=Float3
 
     mod(order, 2) == 0 && throw(ArgumentError("WENO reconstruction scheme is defined only for odd orders"))
 
+    if !isnothing(bounds)
+        !(bounds isa NTuple{2}) && throw(ArgumentError("bounds must be nothing or a tuple of two values"))
```
glwagner:
```suggestion
        bounds isa Tuple || throw(ArgumentError("bounds must be nothing or a tuple of two values"))
```

glwagner: @navidcy lemme know what you think

navidcy:
```Julia
julia> (1, 2) isa NTuple{2}
true

julia> (1, 2, 2) isa NTuple{2}
false

julia> (1, ) isa NTuple{2}
false

julia> (1, 2, 2) isa Tuple
true

julia> (1,) isa Tuple
true
```

NTuple{2} is a subtype of Tuple that only has 2 elements, so how about:

```suggestion
        bounds isa NTuple{2} || throw(ArgumentError("bounds must be nothing or a tuple of two values"))
```

glwagner: deal

navidcy:
Merge?

PR #5058, opened 2025-12-19
Add info to docs about using the TEST_GROUP env variable

glwagner:
This PR will add docs about running the tests (we're starting with a humble comment).

### DIFF
--- a/src/BoundaryConditions/fill_halo_kernels.jl
+++ b/src/BoundaryConditions/fill_halo_kernels.jl
@@ -31,14 +31,9 @@ construct_boundary_conditions_kernels(::Missing, data, grid, loc, indices) = mis
 @inline select_bc(bcs::Tuple) = @inbounds bcs[1]
 @inline select_bc(bcs::Tuple{<:Nothing, <:BoundaryCondition}) = @inbounds bcs[2]
 @inline select_bc(bcs::Tuple{<:BoundaryCondition, <:Nothing}) = @inbounds bcs[1]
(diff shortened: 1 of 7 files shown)


navidcy:
There is some info here that we can use as a starting point: https://clima.github.io/OceananigansDocumentation/stable/contributing/#Setting-up-your-development-environment

glwagner:
Yeah, actually I think all we need is to add another bullet point there about `TEST_GROUP`

navidcy:
Yeah...

docs/src/contributing.md
```diff
@@ -126,13 +126,20 @@ Oceananigans or by requesting something you think is missing.
...
   variable. For example, to run the tests only from the `test_coriolis.jl` file we call:
 
   ```
   $ TEST_FILE=test_coriolis.jl julia --project -e"using Pkg; Pkg.test()"
   ```
 
+  We can also run all the tests within a certain test group, as they are defined in [`test/runtest.jl`](https://github.com/CliMA/Oceananigans.jl/blob/main/test/runtests.jl).
+  For example, to run all unit tests we call:
+
+  ```
```
glwagner:
```suggestion
  ```bash
```

navidcy: done via [f7ff593](https://github.com/CliMA/Oceananigans.jl/pull/5058/commits/f7ff5932ae6d50eaa809614434f90b12bf6a782b)

PR #5062, opened 2025-12-22
Fix immersed cell when bottom = top z

### LINKED ISSUE #5061: `compute_numerical_bottom_height` is incorrect for `PartialCellBottom`
`mask_immersed_field!` does not always mask out the tracer when it should. This happens in particular when setting `bottom = 0.0` with "surface" cells whose top face is also at `z = 0.0`. MWE:

```julia
using Oceananigans
using GLMakie

underlying_grid = RectilinearGrid(
    size = (7, 1, 4),
    x = (0, 1),
    y = (0, 1),
    z = (-1, 0),
)

# Chose some bottom depths to illustrate different cases
bottom = [
    -1.1 # all the cells are wet
    -1.0 # all wet
    -0.5 # half the cells are wet
    -0.2 # partial cell
    -0.0 # all dry
    +0.0 # all dry
    +0.1 # all dry
]

grid = ImmersedBoundaryGrid(
    underlying_grid, PartialCellBottom(bottom);
)

# Make a constant tracer = 0.5 (so that it looks blue on RdBu colormap)
c = CenterField(grid)
c.data .= 0.5

# Mask the immersed field (the culprit)
mask_immersed_field!(c, NaN)

# Plot the tracer
fig, ax, plt = heatmap(c, colormap = :RdBu, colorrange = (-1, 1), nan_color = :black)

# Overlay the bottom topography
Δx = grid.Δxᶜᵃᵃ[1]
for (ix, x) in enumerate(xnodes(grid, Center(), Center(), Center()))
    z = bottom[ix]
    lines!(ax, [x - Δx / 2, x + Δx / 2], [z, z]; color = :red)
    text!(ax, x, z; text = "bottom\nz = $z", color = :red, align = (:center, :bottom), offset = (0, 3))
end

# Adjust axes limits to see text
ax.limits = (0, 1, -1.2, 0.4)

fig
```

Image

briochemc:
Fixes #5061

| Before | After | 
| :-- | :-- |
| Image  | Oceananigans_issue5062 |

### DIFF
--- a/.buildkite/distributed/pipeline.yml
+++ b/.buildkite/distributed/pipeline.yml
@@ -1,7 +1,7 @@
 agents:
   queue: central
   slurm_mem: 8G # Note that the tests run on shared nodes, so limiting the memory usage might help in avoiding long queues
-  modules: climacommon/2024_10_10
+  modules: climacommon/2025_07_30
 
 env:
   JULIA_DEPOT_PATH: "${BUILDKITE_BUILD_PATH}/${BUILDKITE_PIPELINE_SLUG}/depot/default:"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,3 +184,28 @@ jobs:
       - uses: julia-actions/julia-runtest@v1
         env:
           TEST_GROUP: "metal"
+
+  conservative_regridding:
+    name: ConservativeRegridding - Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - '1.10'
+        os:
+          - ubuntu-latest
+        arch:
+          - x64
+    steps:
+      - uses: actions/checkout@v6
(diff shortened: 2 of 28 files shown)


briochemc:
Not sure how/where to put a test for this. I'm happy to reuse the same example as in #5061 but this was mostly to make the plot shown here, so some pointers would be great as to what I should keep/remove, and where to put the example!

briochemc:
PS: I also took the liberty to rename a variable in `src/ImmersedBoundaries/grid_fitted_bottom.jl`. It was called `bottom_cell` but `immersed_cell` is more accurate as it this variable represents whether a cell is immersed or not, rather than whether a cell is the bottom cell.

glwagner:
> Not sure how/where to put a test for this. I'm happy to reuse the same example as in #5061 but this was mostly to make the plot shown here, so some pointers would be great as to what I should keep/remove, and where to put the example!

In your example, there are two cells which you have determined must be immersed; the bug is that they are not. So the test can be to make `ImmersedBoundaryGrid`, and then manually call `immersed_cell` at those indices. They should return `true`:

```julia
# make grid

@test immersed_cell(i1, j1, k1, grid)
```

as an example

navidcy:
I'm merging this. Thanks @briochemc!

PR #5071, opened 2025-12-25
Update coordinate system information in physics documentation

### LINKED ISSUE #1779: Update Coord system and Notation page in docs
I think [Coordinate system and notation](https://clima.github.io/OceananigansDocumentation/stable/physics/notation/) page in the docs needs to be updated. 

First, we don't only use a Cartesian grid anymore (we don't even use the word Cartesian for the `RegularRectilinearGrid` anymore). I'm not sure how you guys wanna adapt this page and notation for the spherical coordinate grid though.

Second, a part of the math isn't rendering and I'm not sure why.

glwagner:
This PR updates the discussion about coordinate systems in the physics documentation.

Although, I think this discussion should be made (and in greater detail) in the `Grids` tutorial. But this page is out of date so it's worth updating it (if we want to keep the physics documentation...)

Closes #1779

Note, I also changed the title of this page so that it does not refer to "notation". I think we want to have a notation page, but it should serve a different purpose and cover both math and unicode symbols, similar to the notation page for Breeze: https://numericalearth.github.io/BreezeDocumentation/stable/appendix/notation/

More generally I think we should emphasize code documentation, and include equations interspersed with code, rather than having "physics" docs that are a bit divorced from the code and numerics.

### DIFF
--- a/docs/src/physics/coordinate_systems.md
+++ b/docs/src/physics/coordinate_systems.md
@@ -6,8 +6,8 @@ The extrinsic coordinate system is either _Cartesian_ or _geographic_
 
 ## Cartesian coordinates
 
-Cartesian coordinates are used by [`RectilinearGrid`](@ref) and provide a local, flat representation
-of space with coordinates ``\boldsymbol{x} = (x, y, z)`` and unit vectors
+Cartesian coordinates are used by [`RectilinearGrid`](@ref) and provide a local, flat (Euclidean)
+representation of space with coordinates ``\boldsymbol{x} = (x, y, z)`` and unit vectors
 ``\boldsymbol{\hat x}``, ``\boldsymbol{\hat y}``, and ``\boldsymbol{\hat z}``.
 By convention, ``\boldsymbol{\hat x}`` points east, ``\boldsymbol{\hat y}`` points north,
 and ``\boldsymbol{\hat z}`` points "upward", opposite to the direction of gravitational acceleration.
@@ -31,11 +31,12 @@ We use ``\boldsymbol{u} = u \boldsymbol{\hat x} + v \boldsymbol{\hat y}`` to den
 
 Geographic (or spherical) coordinates ``(λ, φ, z)`` represent longitude, latitude, and
 the vertical coordinate and are used by [`LatitudeLongitudeGrid`](@ref) and [`OrthogonalSphericalShellGrid`](@ref).
-The corresponding unit vectors are ``\boldsymbol{\hat λ}`` (eastward), ``\boldsymbol{\hat φ}`` (northward),
-and ``\boldsymbol{\hat z}`` (upward, radially outward from the center of the sphere).
+The corresponding unit vectors are ``\boldsymbol{\hat λ}``, ``\boldsymbol{\hat φ}``,
+and ``\boldsymbol{\hat z}`` and at each grid location the point eastward, northward, and
+upward (radially outward from the center of the sphere) respectively.
 
-The velocity components in geographic coordinates are ``(u, v, w)``
-where ``u`` is the eastward velocity, ``v`` is the northward velocity, and ``w`` is the vertical velocity:
+The velocity components in geographic coordinates are ``(u, v, w)`` where ``u`` is the eastward
+velocity, ``v`` is the northward velocity, and ``w`` is the vertical velocity:
 ```math
 \boldsymbol{v} = u \boldsymbol{\hat λ} + v \boldsymbol{\hat φ} + w \boldsymbol{\hat z} \, .
 ```


docs/src/physics/coordinate_systems.md
```diff
@@ -0,0 +1,64 @@
+# Coordinate systems
+
+Every grid in Oceananigans is associated with an _extrinsic_ coordinate system.
+The extrinsic coordinate system is either _Cartesian_ or _geographic_
+(spherical / latitude-longitude).
+
+## Cartesian coordinates
+
+Cartesian coordinates are used by [`RectilinearGrid`](@ref) and provide a local, flat representation
```
navidcy: what does "flat" mean to convey in this context?

navidcy: I mean, I understand what it tries to convey tbh. I'm wondering if "Euclidean" is a better word here. But I'm happy either way!

glwagner: or "rectilinear"?

docs/src/physics/coordinate_systems.md
```diff
@@ -0,0 +1,62 @@
...
+We use ``u``, ``v``, and ``w`` to denote the velocity components in the ``x``, ``y``, and ``z`` directions,
+such that the three-dimensional velocity is
+``\boldsymbol{v} = u \boldsymbol{\hat x} + v \boldsymbol{\hat y} + w \boldsymbol{\hat z}``.
+We use ``\boldsymbol{u} = u \boldsymbol{\hat x} + v \boldsymbol{\hat y}`` to denote the horizontal velocity.
+
+## Geographic coordinates
+
+Geographic (or spherical) coordinates ``(λ, φ, z)`` represent longitude, latitude, and
+the vertical coordinate and are used by [`LatitudeLongitudeGrid`](@ref) and [`OrthogonalSphericalShellGrid`](@ref).
+The corresponding unit vectors are ``\boldsymbol{\hat λ}`` (eastward), ``\boldsymbol{\hat φ}`` (northward),
```
navidcy: Should we clarify "local unit vectors" to convey the fact that in contrast with RectilinearGrid, these unit vectors change with grid location?

docs/src/physics/coordinate_systems.md
glwagner: honestly I am not totally sure what "flat" means here 😂  what is your interpretation?

docs/src/physics/coordinate_systems.md
glwagner: In the extrinsic coordinate system it is grid independent right?

docs/src/physics/coordinate_systems.md
navidcy: It means "not curved". Thus Euclidean.
I'm interpreting it from a differential geometry perspective.

navidcy: Let's not use rectilinear, just because it sounds circular.

docs/src/physics/coordinate_systems.md
navidcy: No. For example, the $\hat{φ}$ unit vector rotates as we go from the equator to the pole! In cartesian coordinates $\hat{φ} = \hat{z}$ at the Equator but at the pole $\hat{φ} \cdot \hat{z} = 0$

navidcy:
I'm happy to merge this as is btw!

docs/src/physics/coordinate_systems.md
glwagner: but that is a property of the coordinate system, not any particular grid, right? What grid would it be a property of?

navidcy: Perhaps I missed what you meant by "grid independent".
I just wanted to point out that the unit vectors rotate as you loop through the `i, j` points in any curvilinear grid.

glwagner: my confusion is that we are not referring to anything discrete (you are referring to `i, j`, of a "grid") in this section of the docs. however, your point holds in a continuous context, too, so I think that is how it should be explained

glwagner: the unit vectors rotate depending on position on the sphere?

navidcy: Yes. They do!

navidcy:
@glwagner have a look at https://github.com/CliMA/Oceananigans.jl/pull/5071/commits/5b4a9349c13d9b7450701b254880722b2af885be and edit or merge if you are happy!

docs/src/physics/coordinate_systems.md
glwagner: That is how I would describe it rather than referring to the grid. The purpose of this section is to introduce coordinate systems that are used in later equations.

docs/src/physics/coordinate_systems.md
glwagner: I don't think it's circular.

I think some clarification is needed. This docs section is defining the coordinate systems which are later used to write governing equations.

It doesn't matter what the grids are / what coordinates they use. It's just cherry on top to mention them here.

Should we remove mention of the grids? It is confusing?

As a further side note, I believe that if RectilinearGrid uses rectilinear coordinates, it is not circular but rather literal. It's like the fact that `Array` represents an array.

navidcy: Go for it

docs/src/physics/coordinate_systems.md
navidcy: Feel free to tweak/edit

PR #5073, opened 2025-12-26
Refactor BoundaryAdjacentMean to AverageBoundaryFlux

glwagner:
@jagoosw this PR refactors your original PR "fix bam". It addresses a few things:

1. In the original implementation of BoundaryAdjacentMean, the flux average is taken over the whole domain, but only the boundary value is used. This PR makes it so that we perform the average over only the slice adjacent to the boundary, using `view`.
2. I precompute and store the area for the flux.
3. I removed the design wherein `BoundaryAdjacentMean` acts like a callable object. Instead, I extend `getbc` to extra the flux value. Also, I noticed that there is no need to store the flux within a Ref. It is fine to store the flux in a 1-element array, which is where the reduction ends up anyways.

I also changed the name to `AverageBoundaryFlux` to emphasize the fact that this is a flux calculation.

I'd also like to mention that on the original PR, I suggested using `Average`. But in the course of this work I realize this is not possible because boundary conditions must be available _before_ a field is constructed. Because of this chicken-egg issue, it is not possible to produce a boundary condition that depends on the average of a Field without a bit of contortion (which goes well beyond merely defining `regularize_field_boundary_conditions` --- it requires intercepting the `Field` constructor itself). For now, I decided it would be best to avoid extending / changing the `Field` constructor.

If you agree with these changes, I will merge this into your PR and continue there. Let me know!

### DIFF
--- a/src/Models/boundary_adjacent_mean.jl
+++ b/src/Models/boundary_adjacent_mean.jl
@@ -9,9 +9,9 @@ import Oceananigans.BoundaryConditions: update_boundary_condition!, getbc, regul
 """
     AverageBoundaryFlux
 
-Computes and stores the area-weighted average of a field on a boundary plane.
+Computes and stores the area-weighted average of a field `f` on a boundary plane.
 
-The average is computed as `∫u dA / ∫dA` where the integral is over the
+The average is computed as `∫f dA / ∫dA` where the integral is over the
 boundary-adjacent plane. The total area `∫dA` is pre-computed during
 boundary condition regularization.
 


test/test_boundary_conditions.jl
```diff
@@ -356,4 +356,54 @@ end
...
         @test all(c.data[1:10, 11, 1:10] .== 2 * mean(c.data[1:10, 10, 1:10]) .- c.data[1:10, 10, 1:10])
     end
+
+    @testset "AverageBoundaryFlux" begin
+        @info "  Testing AverageBoundaryFlux..."
+
+        using Oceananigans.Models: AverageBoundaryFlux
+        using Oceananigans.BoundaryConditions: regularize_boundary_condition, update_boundary_condition!
+
+        grid = RectilinearGrid(size=(4, 4, 4), extent=(1, 1, 1))
```
navidcy, line 366: should we test a lat-lon to ensure the areas are computed correctly?

glwagner: I would hold off because we don't know if this would be used for a hydrostatic open boundary condition

glwagner:
I'm going to merge this, and we can continue on the `fix-bam` PR.

PR #5079, opened 2025-12-27
Combine distributed data automagically in FieldTimeSeries

### LINKED ISSUE #2351: Feature: MPI-aware output writing and `FieldTimeSeries`
One way to support post-processing data generated by distributed runs is to adopt some standard naming scheme for writing output with `MultiArch` (like appending filenames with `rank0`, `rank1`, etc). Then we can implement a "distributed-output-aware" `FieldTimeSeries` that knows how to

1. Reconstruct a global grid given N local grids and
2. Reconstruct global field data given N fields

A related feature is "re-writing" output produced by a distributed simulation. However, I think adding features to `FieldTimeSeries` is actually easier, and might suffice?

Right now, users need to mangle their own names for writing output with `MultiArch`, which violates the rule that the only change users should have to make to run a script on distributed systems is `CPU()` -> `MultiArch(CPU())`.

glwagner:
This PR implements a feature wherein FieldTimeSeries will automatically combine distributed data if it finds files called `filename_rank0`, `filename_rank1`, etc.

A similar output combining was implemented by @taimoorsohail here: https://github.com/taimoorsohail/ocean-ensembles/blob/ts/parallel-run/analysis/combine2d_distributed.jl

Part of this change also effectively requires `FieldTimeSeries` to work when only the prefix of a filename is provided, without the extension `.jld2`. So that is done as well.

Non-combination can still be achieved by passing the specific filename, eg `FieldTimeSeries(filename_rank0, varname)` for the data on rank0.

Closes #2351


src/OutputReaders/combining_field_time_series.jl
```diff
@@ -0,0 +1,408 @@
...
+    return x_offsets, y_offsets
+end
+
+"""
+    collect_global_coordinates(all_ranks, dim, coord_func)
+
+Concatenate coordinate data from all ranks along dimension `dim`.
+`coord_func` extracts coordinates from a grid (e.g., `cpu_face_constructor_x`).
+"""
+function collect_global_coordinates(all_ranks, dim, coord_func)
```
simone-silvestri: maybe here we can generalize the `assemble_coordinate` function in `DistributedComputations` to work without MPI?
I am not sure if it is possible though

src/OutputReaders/combining_field_time_series.jl
```diff
@@ -0,0 +1,408 @@
...
+    return vcat(coords...)
+end
+
+"""
+    reconstruct_global_grid(all_ranks, arch)
+
+Reconstruct a global grid from distributed rank output data.
+This is the offline (file-based) equivalent of `DistributedComputations.reconstruct_global_grid`.
+"""
+function reconstruct_global_grid(all_ranks, arch)
```
simone-silvestri: should this function be imported?

src/OutputReaders/combining_field_time_series.jl
```diff
@@ -0,0 +1,408 @@
...
+    yG = collect_global_coordinates(all_ranks, 2, cpu_face_constructor_y)
+    zG = cpu_face_constructor_z(grid0)
+    
+    return build_global_grid(grid0, arch, FT, topo, (Nx, Ny, Nz), H, xG, yG, zG)
+end
+
+# Grid building dispatched by grid type
+# (Mirrors the pattern in DistributedComputations.reconstruct_global_grid)
+
+function build_global_grid(grid0::RectilinearGrid, arch, FT, topo, N, H, xG, yG, zG)
```
simone-silvestri: here we can probably use the same functions that we have in `DistributedComputations`?

simone-silvestri: probably not, there `reconstruct_global_grid` is dispatched on the grid type...

src/OutputReaders/combining_field_time_series.jl
glwagner: I think we could, but we have to develop a kind "fake distributed context" in which we have all of the architecture for the ranks available, and loop over them, or something?

glwagner: and then we have to dispatch this:

https://github.com/CliMA/Oceananigans.jl/blob/d5cdd2dc29f39d057a8e1e1d9e475384c297b5ed/src/DistributedComputations/partition_assemble.jl#L99

so something like

```julia
allreduce!(obj, op, arch::Distributed) = MPI.Allreduce!(obj, op, arch.communicator)
allreduce!(obj, op, arch::FakeDistributed) = # something else
```

glwagner: It might require something not so dissimilar to what we have for MultiRegion.

glwagner: Actually this hints at the benefits of having a more integrated abstraction for distributed computations that works for both single process situations (MultiRegion) and multiple process situations (Distributed). 

Since we already have infrastructure in place for MultiRegion, it possibly would not be too hard to expand the Distributed abstraction in that direction. But it is well beyond the scope of this PR and I am not sure worth the heavy lifting.

src/OutputReaders/field_time_series.jl
```diff
@@ -498,17 +503,44 @@ function FieldTimeSeries(path::String, name::String;
...
+                                                   boundary_conditions,
+                                                   time_indexing,
+                                                   iterations,
+                                                   times,
+                                                   reader_kw)
+            end
+        end
+        
+        # Look for part1, etc (file splitting)
         start = path[1:end-5]
```
simone-silvestri, line 531: is `*part*` before or after `rank` in the convention?

glwagner: This line actualy isn't changed in this PR -- i think it is for files with output split into "parts" due to a max_filesize

src/OutputReaders/combining_field_time_series.jl
glwagner: I think what we should do is rename this function, so that it cannot be confused with `reconstruct_global_grid(local_distributed_grid)`

src/OutputReaders/combining_field_time_series.jl
glwagner: I'm addressing this by chnaging the function name to make it clear we are proceeding from a file

src/OutputReaders/combining_field_time_series.jl
glwagner: i changed the name of this function to resolve the ambiguity here, since dispatch is probably not appropriate

PR #5080, opened 2025-12-27
Move initialization of JLD2Writer to run!

### LINKED ISSUE #1163: Initialize output before running a simulation, rather than on construction?
A few recent PRs may suggest that separating the _construction_ of an output writer should be separated from output writer "initialization".

Currently both JLD2 and NetCDF output writers open their files and save a bunch of data when they are constructed. For example, 

https://github.com/CliMA/Oceananigans.jl/blob/7b029a75b8e17d3ca818db27eba41fd5d3a0397e/src/OutputWriters/netcdf_output_writer.jl#L300-328

and

https://github.com/CliMA/Oceananigans.jl/blob/7b029a75b8e17d3ca818db27eba41fd5d3a0397e/src/OutputWriters/jld2_output_writer.jl#L170-L181

This initialization can make "pickup" of a simulation annoying. For example, the JLD2 writer will _remove_ an existing file if `force=true`:

https://github.com/CliMA/Oceananigans.jl/blob/7b029a75b8e17d3ca818db27eba41fd5d3a0397e/src/OutputWriters/jld2_output_writer.jl#L172

This is desirable when experimenting with a script and setting up (and thus we usually set `force=true` in examples as a matter of user-friendliness). But if you're picking up a simulation you certainly don't want to set `force=true`, since this will overwrite existing data.

Another example is the warning implemented in #1162, which I suppose one must resign themselves to always receiving if they are picking up a simulation?

Debacles such as these might be avoided if we performed initialization in `run!` rather than when the output writers are created --- because then we can implement special behavior in the case that we are picking up a simulation. More or less I think the point here is that we don't _know_ if we want to create a file until `run!` is called with appropriate arguments. If `run!` is never called, there will not be output (so maybe we don't want to create a file). If `run!` is called and we are not picking up and `force` or `overwrite_existing=true` then we should do that. But if `run!` is called and `pickup=true` we can likely assume that existing files should not be deleted.

### LINKED ISSUE #5085: Make output writers work for models that do not have grids
because an output at the moment requires the model to have a grid. In this case the "grid" of the specific `OceanSeaIceModel` can be the exchange grid I think.

_Originally posted by @simone-silvestri in https://github.com/CliMA/ClimaOcean.jl/issues/602#issuecomment-3694626813_

glwagner:
Closes #1163
Closes #5085

I did this just for JLD2Writer, but we could also add this feature for NetCDFWriter if we want.

### DIFF
--- a/docs/src/models/turbulence_closures.md
+++ b/docs/src/models/turbulence_closures.md
@@ -187,6 +187,23 @@ ConvectiveAdjustmentVerticalDiffusivity{VerticallyImplicitTimeDiscretization}(ba
 ocean turbulence, as described by [Wagner25catke](@citet).
 It uses a prognostic equation for turbulent kinetic energy (the `:e` tracer).
 
+The `minimum_tke` keyword argument (default: `1e-9`) produces background tracer diffusivity
+
+```math
+\kappa_{bg} \approx C^{hi}_c \frac{e^{min}}{N}
+```
+
+and background viscosity
+
+```math
+\nu_{bg} \approx C^{hi}_u \frac{e^{min}}{N}
+```
+
+where ``N`` is the buoyancy frequency.
+By default, ``C^{hi}_c = 0.098`` and ``C^{hi}_u = 0.242`` are parameters of `CATKEMixingLength`.
+This feature may be used to model background mixing due to breaking internal waves
+[Wagner25catke](@citep).
+
 !!! note "HydrostaticFreeSurfaceModel only"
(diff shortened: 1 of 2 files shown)


glwagner:
also addressing https://github.com/CliMA/ClimaOcean.jl/pull/602#issuecomment-3694792192

cc @simone-silvestri

glwagner:
I will merge this now. We can address initializing the NetCDF output writer once various refactoring has gone in for it (eg #5017)

PR #5082, opened 2025-12-27
(0.104.0) Change notation to free_surface.displacement

### LINKED ISSUE #3702: Change free surface displacement name from `η` to `displacement` for `HydrostaticFreeSurfaceModel`
I think it reads better in a script / produces scripts that are more understandable because instead of

```julia
η = model.free_surface.η
```

we would read

```julia
η = model.free_surface.displacement
```

glwagner:
This PR changes 

```julia
η = model.free_surface.η
```

to

```julia
η = model.free_surface.displacement
```

It closes #3702.

I am not sure we should do this. But if we decide not to, we can also just close #3702 rather than keeping it open.

### DIFF
--- a/test/test_split_explicit_free_surface_solver.jl
+++ b/test/test_split_explicit_free_surface_solver.jl
@@ -106,13 +106,13 @@ clock = Clock(time=0)
             sefs = SplitExplicitFreeSurface(substeps = 200, averaging_kernel = constant_averaging_kernel)
             sefs = materialize_free_surface(sefs, velocities, grid)
 
-            sefs.η .= 0
+            sefs.displacement .= 0
 
             @testset "Averaging / Do Nothing test " begin
                 state = sefs.filtered_state
                 U, V  = sefs.barotropic_velocities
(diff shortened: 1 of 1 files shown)


navidcy:
Let’s do it!

navidcy (approved):
this is breaking change probably...
user scripts that save η won't work...

navidcy:
This might be ready. Merge if CI passes?

test/test_split_explicit_free_surface_solver.jl
```diff
@@ -176,7 +175,7 @@ clock = Clock(time=0)
                 state = sefs.filtered_state
                 U, V = sefs.barotropic_velocities
                 η̅, U̅, V̅ = state.η̅, state.U̅, state.V̅
```
navidcy, line 177: within the `filtered_state` there is still `η̅`
do we want to change that to `filtered_displacement`?

glwagner: i say no because it appears alongside U, V.

